PR #31, opened 2015-03-04
adding stdout and stderr to  resolve_interpreted_items for webview 

Raisins:
This allows the worksheet view on the web to display stdout and error.

### DIFF
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -718,8 +718,27 @@ class LocalBundleClient(BundleClient):
                 pass
             else:
                 raise UsageError('Invalid display mode: %s' % mode)
-
+            # assign the interpreted from the processed data
             item['interpreted'] = data
+
+            # we need to get check if this is a run and we can get teh stdout and stderr
+            if 'bundle_info' in item.keys():  # making sure this is a bundle, not markdown or something else
+                for info in item['bundle_info']:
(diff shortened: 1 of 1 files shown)


percyliang:
Could we make fetching stdin/stderr a separate API call, so that this part can load asynchronously?  Otherwise, there's going to be a large blocking bottleneck where we have to fetch all the files for all the elements on a worksheet all in one go.

Raisins:
I originally had it doing that on the web view backend. On some more complex worksheets it would be many many calls over the internet and RPC with network lag it can take a very long time for even the page to render then due to all the RPC calls that need to be placed. 

Alternatively.  I had it doing a separate API call to fetch the info (a RPC call) for each item for the bundle info for sidebar when the bundle was focused in the interface but this cause other problems due to rapid J/K presses, many requests coming back at the same time or out of order due to network lag. Made the entire page feel slower and added more complexity to worksheet interface for the web.

Weighing the options, I went for updating the one call for only the web interface and the code does checks to make sure the item is a run and it also has stdout or stderr before going to fetch the files. Since the cli is already running on that system it's just a file lookup not an RPC+network it is much faster the the other methods I tried even on larger worksheets.

percyliang:
Thanks for the reply.  One question: is the worksheet viewable (in some
crude form) before the call to resolve_interpreted_items?  If so, then it's
probably fine, since the lag for getting the information for the side panel
is tolerable.

Issue #103, opened 2015-08-12
symlinks


percyliang:
Don't cat symlinks, but make sure that symlinks are preserved properly when uploading/downloading a bundle.

PR #114, opened 2015-08-22
treat symlinks as opaque entities

percyliang:
Don't follow them on cat, head, ls, or download.
Improve rendering of file listing.


percyliang:
Closes #103

Issue #97, opened 2015-08-12
deleting queued and running jobs


percyliang:
Sometimes things die and job states remain as 'queued', and they're not deletable.  Create some mechanism to force delete queued or running jobs that have gone stale.

PR #129, opened 2015-09-04
allow deleting queued/running bundles using '-f' (if bundle is dead, …

percyliang:
…but its state is not updated)


percyliang:
Fixes #97

PR #143, opened 2015-09-25
Fetch bundle uuids from the client in batches

### LINKED ISSUE #66: when reference many bundles (^1-10), use worksheet.get_bundle_uuids
Currently, in bundle_cli.py, we're getting one bundle at a time, which is really slow.

For example, in do_rm_command(), we are calling
bundle_uuids = [worksheet_util.get_bundle_uuid(client, worksheet_uuid, bundle_spec) for bundle_spec in args.bundle_spec]
This should be replaced with one call.
Most commands (e.g., kill, add, perm) should be changed.

kashizui:
Fixes #66.

@pbhatnagar3 please review :)

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -173,12 +173,20 @@ def get_bundle_uuids(client, worksheet_uuid, bundle_specs):
     Important difference from client.get_bundle_uuids: if all bundle_specs are already
     uuids, then just return them directly.  This avoids an extra call to the client.
     '''
-    bundle_specs = [bundle_spec.strip() for bundle_spec in bundle_specs]
+    bundle_uuids = {}
+    unresolved = []
+    for spec in bundle_specs:
+        spec = spec.strip()
+        if spec_util.UUID_REGEX.match(spec):
+            bundle_uuids[spec] = spec
+        else:
+            unresolved.append(spec)
 
-    if all(spec_util.UUID_REGEX.match(bundle_spec) for bundle_spec in bundle_specs):
-        return bundle_specs
+    # Resolve uuids with a batch call to the client and update dict
+    bundle_uuids.update(zip(unresolved, client.get_bundle_uuids(worksheet_uuid, unresolved)))
 
-    return client.get_bundle_uuids(worksheet_uuid, bundle_specs)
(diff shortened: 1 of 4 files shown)


codalab/lib/worksheet_util.py
```diff
@@ -167,24 +167,33 @@ def request_lines(worksheet_info, client):
...
+    unresolved = []
+    for bundle_spec in bundle_specs:
+        bundle_spec = bundle_spec.strip()
+
+        if spec_util.UUID_REGEX.match(bundle_spec):
+            resolved.append(bundle_spec)
+        else:
+            unresolved.append(bundle_spec)
+
+    return resolved + client.get_bundle_uuids(worksheet_uuid, unresolved)
```
percyliang: Don't we need to return the uuids in the same order as the bundle_specs?
This code seems might permute things.

kashizui:
To address the ordering problem, simplified the logic to using the batch call on all of the provided bundle_specs if at least one of them need resolving (i.e. we will need to call the client anyway). Does this make sense?

percyliang:
Could do that, but I think it's not much harder to make a call on only the
unresolved names, build a dict from names to uuids, and then take another
pass over the array of bundle_specs.

 -Percy

kashizui:
Got it-- does this implementation of the dict idea make sense?

Stephen

kashizui:
(still pending testing, please don't merge yet)

percyliang:
Structure looks good to me!

kashizui:
OK to merge!

this works as expected
https://gist.github.com/kashizui/f5f4d111845ca7fd3802

percyliang:
Great!  Could you distill this into a test and add it to test-cli.py?
Also, make sure you test passing in a combination of uuids and names (with
repeated specs).  It suffices just to do it for cl info -f uuid and check
the output.

kashizui:
Will do—what do you mean by repeated specs?

percyliang:
cl info a a a 0x23... a 0x23...

kashizui:
Added a test for the batch call, and also updated test-cli.py to work on Mac, and also to continue to the remaining tests even if one fails.

percyliang:
LGTM

PR #150, opened 2015-09-29
Refactor test-cli.py

kashizui:
In general trying to make writing new tests in test-cli.py cleaner and more robust. Figured this is a good investment as we put more work into the cli.
- Decorator for test functions
- All actions now done on temporary worksheets that are cleaned up
  automatically at the end of test modules
- If an exception is raised or an assertion fails during a test, all created resources are
  cleaned up before moving on to the next test, to avoid leaving workspace in an
  incomplete state
- Context manager for test modules to manage isolated test environments
  and handle exceptions
- Improved feedback for module selection

example output:
https://gist.github.com/kashizui/04d277c8afd8de2c5498

documentation to come

### DIFF
--- a/test-cli.py
+++ b/test-cli.py
@@ -2,9 +2,13 @@
 '''
 Tests all the CLI functionality end-to-end.
 
-Currently, the tests will operate on your current worksheet.  In theory, it
-shouldn't mutate anything, but this is not guaranteed, and you should run this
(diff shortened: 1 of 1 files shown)


percyliang:
This is really nice - thanks for doing this!
LGTM

kashizui:
All tests passing, merging now.

PR #156, opened 2015-10-05
fix #145 cl edit opens a editor when no additional args are supplied

### LINKED ISSUE #145: cl edit <bundle>
Should popup an editor, but doesn't currently.  Note that if arguments are specified to edit, then there should be no popup. [30m]

pujun-ai:
#145

cl edit opens an editor when no additional args are supplied. Also added logic for handling the use case where the new and the old values are the same, the editor does not open

```
cl edit a.txt -n a.txt
```

now does not open the editor. 
image

and when their are no arguments supplied:
image

@percyliang @kashizui please review.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -694,13 +694,16 @@ class BundleCLI(object):
 
         metadata = info['metadata']
         new_metadata = copy.deepcopy(metadata)
+        is_new_metadata_updated = False
         if args.name:
             new_metadata['name'] = args.name
+            is_new_metadata_updated = True
         if args.description:
             new_metadata['description'] = args.description
+            is_new_metadata_updated = True
 
         # Prompt user for all information
-        if not self.headless and metadata == new_metadata:
+        if not is_new_metadata_updated and not self.headless and metadata == new_metadata:
             new_metadata = metadata_util.request_missing_metadata(bundle_subclass, args, new_metadata)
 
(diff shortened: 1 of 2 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -694,13 +694,16 @@ def do_edit_command(self, argv, parser):
...
         if args.name:
             new_metadata['name'] = args.name
+            is_new_metadata_updated = True
         if args.description:
             new_metadata['description'] = args.description
+            is_new_metadata_updated = True
 
         # Prompt user for all information
-        if not self.headless and metadata == new_metadata:
+        if not is_new_metadata_updated and not self.headless and metadata == new_metadata:
```
percyliang, line 706: Why doesn't the metadata == new_metadata condition cover this case?

pujun-ai, line 706: if the old and new name are the same, then the condition metadata == new_metaddata evaluates to True and the editor is opened.

percyliang, line 706: So why do you need to add is_new_metadata_updated?

pujun-ai, line 706: is_new_metadata_updated serves like a flag. Whenever anything in the new_metadata is changed, it is set to True, thus avoiding opening the editor.

If this variable were absent, then in the use case where old and the new value are the same, the editor would open (even when the new value was passed with cl edit command)

PR #161, opened 2015-10-10
delete temporary bundles all at once and dedup

percyliang:
test-cli.py was failing on 'detach' because it was trying to delete a bundle more than once.  So this change dedups.  And just to streamline things, might as well delete all bundles at once.
@kashizui

### DIFF
--- a/test-cli.py
+++ b/test-cli.py
@@ -116,8 +116,9 @@ class ModuleContext(object):
             self.bundles.extend(run_command([cl, 'ls', worksheet, '-u']).split())
             run_command([cl, 'wrm', '--force', worksheet])
 
-        for bundle in self.bundles:
-            run_command([cl, 'rm', '--force', bundle])
+        # Delete all bundles (dedup first)
+        if len(self.bundles) > 0:
+            run_command([cl, 'rm', '--force'] + list(set(self.bundles)))
 
         # Do not reraise exception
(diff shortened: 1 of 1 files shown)


kashizui:
LGTM!

percyliang:
Was the detach test not failing for you?  That would be weird.

kashizui:
No, it wasn't failing for me, not sure why. :/ but looks like a good change to do anyway.

https://gist.github.com/kashizui/599ab6f710f2f24d6408

percyliang:
The issue is that when you do:
cl rm --force 0xa5542cf7826d4c59b89c726169b4d0ea
cl rm --force 0xa5542cf7826d4c59b89c726169b4d0ea
in local mode, then the second statement doesn't throw an error since
you're the codalab (root) user, whereas when when connected to a cl server,
the second one fails because it tries to get the permissions of a bundle
and fails:
User pliang(3) does not have sufficient permissions on bundle
0xa5542cf7826d4c59b89c726169b4d0ea (have none, need all).

I guess this says that it's useful to run the tests not as a non-root
user.  In general, our testing of the permissions systems, etc. is lacking.

kashizui:
That makes a lot of sense. Just added a comment to test-cli to reflect this
in the meantime before we have permissions testing.

PR #166, opened 2015-10-12
Upload fix for action-bar

pujun-ai:
@percyliang @kashizui please review

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -322,6 +322,21 @@ class BundleCLI(object):
         self.hack_formatter(parser)
         return parser
 
+    def get_missing_metadata(self, bundle_subclass, args, initial_metadata=None):
+        '''
+        Return missing metadata for bundles by either returning default metadata values or
+        pop up an editor and request that data from the user.
+        '''
+        if not initial_metadata:
+            initial_metadata = {
+                spec.key: getattr(args, metadata_util.metadata_key_to_argument(spec.key,))
+                for spec in bundle_subclass.get_user_defined_metadata()
+            }
+        if not getattr(args, 'edit', True):
+            return metadata_util.fill_missing_metadata(bundle_subclass, args, initial_metadata)
+        else:
+            return metadata_util.request_missing_metadata(bundle_subclass, initial_metadata)
+
     #############################################################################
     # CLI methods
     #############################################################################
@@ -502,7 +517,7 @@ class BundleCLI(object):
             bundle_uuid = worksheet_util.get_bundle_uuid(client, worksheet_uuid, args.base)
             info = client.get_bundle_info(bundle_uuid)
             metadata = info['metadata']
-        metadata = metadata_util.request_missing_metadata(bundle_subclass, args, initial_metadata=metadata)
+        metadata = self.get_missing_metadata(bundle_subclass, args, initial_metadata=metadata)
         # Type-check the bundle metadata BEFORE uploading the bundle data.
         # This optimization will avoid file copies on failed bundle creations.
         # pass in a null owner to validate. Will be set to the correct owner in the client upload_bundle below.
@@ -636,7 +651,7 @@ class BundleCLI(object):
 
         client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
         targets = self.parse_key_targets(client, worksheet_uuid, args.target_spec)
-        metadata = metadata_util.request_missing_metadata(MakeBundle, args)
+        metadata = self.get_missing_metadata(MakeBundle, args)
         print client.derive_bundle('make', targets, None, metadata, worksheet_uuid)
 
     # After running a bundle, we can wait for it, possibly observing it's output.
@@ -676,7 +691,7 @@ class BundleCLI(object):
         client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
         args.target_spec, args.command = cli_util.desugar_command(args.target_spec, args.command)
         targets = self.parse_key_targets(client, worksheet_uuid, args.target_spec)
-        metadata = metadata_util.request_missing_metadata(RunBundle, args)
+        metadata = self.get_missing_metadata(RunBundle, args)
         uuid = client.derive_bundle('run', targets, args.command, metadata, worksheet_uuid)
         print uuid
         self.wait(client, args, uuid)
@@ -705,7 +720,7 @@ class BundleCLI(object):
 
         # Prompt user for all information
         if not is_new_metadata_updated and not self.headless and metadata == new_metadata:
-            new_metadata = metadata_util.request_missing_metadata(bundle_subclass, args, new_metadata)
+            new_metadata = self.get_missing_metadata(bundle_subclass, args, new_metadata)
 
         if metadata != new_metadata:
             client.update_bundle_metadata(bundle_uuid, new_metadata)
--- a/codalab/lib/metadata_util.py
+++ b/codalab/lib/metadata_util.py
@@ -60,17 +60,12 @@ def add_edit_argument(parser):
     )
 
 
-def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+def fill_missing_metadata(bundle_subclass, args, initial_metadata):
     '''
-    For any metadata arguments that were not supplied through the command line,
-    pop up an editor and request that data from the user.
+    Return metadata for bundles by filling in the missing metadata with default values.
+    args: Namespace object created from attributes parsed out of the command line. See
+        `argparse` for more information
     '''
-    if not initial_metadata:
-        initial_metadata = {
-          spec.key: getattr(args, metadata_key_to_argument(spec.key,))
-          for spec in bundle_subclass.get_user_defined_metadata()
-        }
-
     # Fill in default values for all unsupplied metadata keys.
     new_initial_metadata = {}
     for spec in bundle_subclass.get_user_defined_metadata():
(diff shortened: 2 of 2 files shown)


codalab/lib/metadata_util.py
```diff
@@ -60,7 +60,7 @@ def add_edit_argument(parser):
     )
 
 
-def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+def request_missing_metadata(bundle_subclass, args, initial_metadata=None, bypass_editor= False):
```
kashizui: nit: Avoid spaces around `=` for keyword arguments.

codalab/lib/metadata_util.py
```diff
@@ -82,7 +82,7 @@ def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
 
     # If args.edit exists (when doing 'cl edit'), then we want to show
     # the editor.
-    if not getattr(args, 'edit', True):
+    if getattr(args, 'edit', bypass_editor):
```
kashizui: I'm not sure this logic is right: if 'edit' is in the args, this would end up returning early without showing the editor, regardless of the value of `bypass_editor`.

codalab/lib/metadata_util.py
```diff
@@ -60,7 +60,7 @@ def add_edit_argument(parser):
     )
 
 
-def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+def request_missing_metadata(bundle_subclass, args, initial_metadata=None, bypass_editor= False):
     '''
```
kashizui: It might help in the future to add a brief comment about how `bypass_editor` affects behavior.

codalab/lib/metadata_util.py
kashizui: It might also help to add a brief comment about how `bypass_editor` should affect behavior here.

percyliang: As I mentioned in the codalab repo, we should split this function into two:
- create_default_metadata: does everything before the test for editing
- request_missing_metadata: only does what the function name claims :-)

pujun-ai:
Split the request_missing_metadata(...) into two functions:
fill_missing_metadata(...),
request_missing_metadata(...)

 and now the codalab-web-interface calls only fill_missing_metadata(...)

@percyliang @kashizui ready to review

codalab/lib/metadata_util.py
```diff
@@ -78,12 +77,16 @@ def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
...
-    # the editor.
+
+def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+    '''
+    For any metadata arguments that were not supplied through the command line,
+    pop up an editor and request that data from the user.
+    '''
     if not getattr(args, 'edit', True):
-        return initial_metadata
+        return fill_missing_metadata(bundle_subclass, args, initial_metadata)
```
kashizui: Shouldn't you still be calling `fill_missing_metadata` regardless of this condition?

kashizui:
Very clean! See note above, I may be misinterpreting the diff

codalab/lib/metadata_util.py
pujun-ai: Oh yeah. This is what I was thinking:

When there are no args supplied with edit eg. `cl edit b.txt`, the condition `not getattr(args, 'edit', True)` will evaluate to False and the editor will be shown, which is the intended behavior.

When args are supplied with edit eg. `cl edit b.txt -n c.txt`, request_missing_metadata method would never be called (since `is_new_metadata_updated` in bundles_cli: do_cl_edit will be True)

When args are supplied with edit but with the same name e.g. `cl wedit c.txt -n c.txt`, again 
`is_new_metadata_updated` will evaluate to True and will prevent request_missing_metadata method from being called. 

However, let's look at some other command. Let's say when we call `cl run 'echo hello'`, the condition `not getattr(args, 'edit', True)` will evaluate to True and the editor will not be shown, which is the intended behavior. :)

codalab/lib/metadata_util.py
```diff
@@ -78,12 +77,16 @@ def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
         if not new_initial_metadata[spec.key]:
             default = MetadataDefaults.get_default(spec, bundle_subclass, args)
             new_initial_metadata[spec.key] = default
-    initial_metadata = new_initial_metadata
+    return new_initial_metadata
 
-    # If args.edit exists (when doing 'cl edit'), then we want to show
-    # the editor.
+
+def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
```
percyliang: I'd actually remove args from here entirely and do this if statement in bundle_cli where it is more transparent what's going on.

pujun-ai:
moved the condition, that checks if fill_missing_metadata or request_missing_metadata needs to be called to bundles_cli. @percyliang @kashizui please review.

#### Testing

All tests pass. And cl upload from the aciton bar is working.

codalab/lib/bundle_cli.py
```diff
@@ -322,6 +322,10 @@ def create_parser(self, command):
         self.hack_formatter(parser)
         return parser
 
+    def get_missing_metadata(self, bundle_subclass, args, initial_metadata=None):
+        return metadata_util.fill_missing_metadata(bundle_subclass, args, initial_metadata) if not getattr(args, 'edit', True) \
```
percyliang: Make this not a one-liner for readability

codalab/lib/metadata_util.py
```diff
@@ -78,13 +77,14 @@ def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
...
             new_initial_metadata[spec.key] = default
-    initial_metadata = new_initial_metadata
+    return new_initial_metadata
 
-    # If args.edit exists (when doing 'cl edit'), then we want to show
-    # the editor.
-    if not getattr(args, 'edit', True):
-        return initial_metadata
 
+def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
```
percyliang: args is no longer needed here

pujun-ai:
args is one of the arguments of request_missing_metadata because its needed to call `get_default_initial_metadata()`. Since `request_missing_metadata` and `fill_missing_metadata` were originally part of the same method, the use case `initial_metadata=None` was handled. However, when we separated the two, we had to write a new method `get_default_initial_metadata(..)` and call it from the two methods. (otherwise the command `cl upload dataset a.txt --edit` while doing manual testing was crashing). I had not accounted for this in the earlier code review. Have updated the code to reflect this. Sorry for the confusion.

codalab/lib/metadata_util.py
```diff
@@ -60,16 +60,19 @@ def add_edit_argument(parser):
     )
 
 
-def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+def get_default_initial_metadata(bundle_subclass, args):
```
percyliang: Comment what this method does

codalab/lib/metadata_util.py
```diff
@@ -60,16 +60,19 @@ def add_edit_argument(parser):
...
-    For any metadata arguments that were not supplied through the command line,
-    pop up an editor and request that data from the user.
+    Fill in the missing metadata with default values and return modified metadata
     '''
     if not initial_metadata:
-        initial_metadata = {
-          spec.key: getattr(args, metadata_key_to_argument(spec.key,))
-          for spec in bundle_subclass.get_user_defined_metadata()
-        }
+        initial_metadata = get_default_initial_metadata(bundle_subclass, args)
```
percyliang: Move this logic to get_missing_metadata in bundle_cli, so that initial_metadata is never None

codalab/lib/metadata_util.py
```diff
@@ -60,16 +60,19 @@ def add_edit_argument(parser):
...
+        spec.key: getattr(args, metadata_key_to_argument(spec.key,))
+        for spec in bundle_subclass.get_user_defined_metadata()
+    }
+
+
+def fill_missing_metadata(bundle_subclass, args, initial_metadata=None):
     '''
-    For any metadata arguments that were not supplied through the command line,
-    pop up an editor and request that data from the user.
+    Fill in the missing metadata with default values and return modified metadata
```
percyliang: Make this comment more precise

codalab/lib/metadata_util.py
```diff
@@ -78,15 +81,18 @@ def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
...
 
+def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+    '''
+    For any metadata arguments that were not supplied through the command line,
+    pop up an editor and request that data from the user.
+    '''
     # Construct a form template with the required keys, prefilled with the
     # command-line metadata options.
+    if not initial_metadata:
+        initial_metadata = get_default_initial_metadata(bundle_subclass, args)
```
percyliang: Same as above, don't allow initial_metadata to be None

pujun-ai:
@kashizui @percyliang ready for review.

codalab/lib/metadata_util.py
```diff
@@ -60,17 +60,10 @@ def add_edit_argument(parser):
     )
 
 
-def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
+def fill_missing_metadata(bundle_subclass, args, initial_metadata):
     '''
-    For any metadata arguments that were not supplied through the command line,
-    pop up an editor and request that data from the user.
+    Return metadata for bundles by filling in the missing metadata with default values.
```
percyliang, line 65: Mention what args is

codalab/lib/metadata_util.py
```diff
@@ -78,13 +71,14 @@ def request_missing_metadata(bundle_subclass, args, initial_metadata=None):
...
+    return new_initial_metadata
 
-    # If args.edit exists (when doing 'cl edit'), then we want to show
-    # the editor.
-    if not getattr(args, 'edit', True):
-        return initial_metadata
 
+def request_missing_metadata(bundle_subclass, initial_metadata):
+    '''
+    For any metadata arguments that were not supplied through the command line,
```
percyliang: Delete "For any ... command line,"

percyliang:
Only two small comments on comments - otherwise LGTM!

kashizui:
LGTM

Issue #67, opened 2015-04-21
support multithreading


percyliang:
cl server is currently single-threaded.  Make this multi-threaded so one user doesn't block another.

percyliang:
First attempt in the multithreading branch.  Some weird things happening though.

1) Run two copies of the following:
while true; do cl search size=.sort-; done
The results are non-deterministic.

2) If run one copy of the above and then run ./test-cli.py, then it fails at a random place...

The problem is probably in the treatment of sqlalchemy in bundle_model.py.

Raisins:
I am getting back the correct results while running `while true; do cl search size=.sort-; done` 
Also working fine is running 
`for i in `seq 1 200`; do cl search size=.sort-; done`
in one window while running
`for i in `seq 1 20`; do python test-cli.py upload1; done` 
in another console at the same time is not causing any problems.

What weird results or errors are you seeing.

percyliang:
Try running it for longer - it sometimes takes a few minutes for bad things
to happen.
Make sure you're using mysql with a good number of bundles in the system.
./test-cli.py just crashes, and the 'cl search' command just returns
incorrect results (largest bundle changes in about 10% of the cases).

Raisins:
Made a copy of all the worksheets/bundles on testevn. I have run `cl search ...` across 3 windows and with mysql as the backend. Left it running for 30 min and again for an hour. While I working on worksheets and using them normally as well. I am not seeing any errors or weirdities across both web and cli side.

Raisins:
Make copy of codalab test

kashizui:
Just to confirm, all the existing code is already in codalab/codalab-cli/multithreading branch on github?

percyliang:
Yes

kashizui:
I'm unable to reproduce any errors by running the tests above on the multithreading branch (which i've rebased again)

```
while true; do cl search size=.sort-; done &
while true; do cl search size=.sort-; done &
./test-cli.py all
```

I'm on MySQL, though maybe have at most a dozen bundles on my system at the moment.

kashizui:
Same results after uploading a bunch more bundles.

percyliang:
I'm still getting this issue on Ubuntu 14.04 with MySQL, when I run
the two in different windows:

while true; do cl search size=.sort-; done
while true; do cl search size=.sort-; done

You have to watch it for about 30 seconds before it happens.

I only have 133 bundles of varying size from 4 bytes to 1G.

[...]

PR #171, opened 2015-10-16
Enable multithreading on FileServer

kashizui:
Use `SocketServer.ThreadingMixin` to make `FileServer` spawn a new thread for each RPC request. 

The spawned threads will share the same instance of `FileServer`, `LocalBundleClient`, `BundleModel`, and `BundleStore`. These have been verified to be thread-safe when working with the filesystem (mostly thanks to immutability of bundles), and also do not share any mutable state data in process memory. Concerns with concurrent interactions with the database will be addressed in a different patch.

See discussion on #67 

hereby recommending this to be merged with master

@percyliang @pbhatnagar3


percyliang:
LGTM

PR #191, opened 2015-10-29
renaming run bundles and home worksheets

percyliang:
- set default run bundle names to be more succinct (e.g., run-java)
- change home worksheet to home-<username>, enforce always reserved for the user

@kashizui please review


kashizui:
LGTM!

PR #200, opened 2015-11-02
Fix alias command and prompt docker image on config init

### LINKED ISSUE #196: cl alias broken
cl alias throws an error currently.

Also, during the interactive setup, we should ask:
- Do you want to connect to a codalab server (in which case, we don't want to use MockAuthHandler)?
- Do you want to set up docker (in which case docker_image + q needs to be added)?

kashizui:
Fixes #196 
(happy 200th pull request 🎉)

@percyliang @pbhatnagar3

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -34,6 +34,7 @@ import sys
 import time
 import psutil
 import tempfile
+import textwrap
 from distutils.util import strtobool
 
 from codalab.client import is_local_address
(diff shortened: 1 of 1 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -209,6 +208,8 @@ def init_config(self, dry_run=False):
 online to set up execution with Docker:
 https://github.com/codalab/codalab/wiki/Dev_CodaLab%20CLI%20Execution%20in%20Docker
 """
+            if prompt_bool("Would you like to configure a Docker image to use?", default=False):
```
percyliang: Is the indentation off here?

percyliang:
LGTM

kashizui:
Added prompts for OAuth setup and made indentation clearer.

@percyliang

percyliang:
LGTM again - thanks!

PR #204, opened 2015-11-04
Limit -w flag on cl add to destination worksheet

### LINKED ISSUE #201: Improve worksheet manipulation commands
cl add ^ -w worksheet

doesn't work and prints out

UsageError: bundle spec None doesn't match (want index 1 out of 0 bundles)

kashizui:
Now relative references to bundles in `cl add` refer to the _current_ worksheet, not the worksheet specified by the `-w` flag.

For example,
`cl add ^ -w target-worksheet`
will add the last bundle on the current worksheet to `target-worksheet`.

Previously, it would try to add the last bundle on `target-worksheet` to itself.

Fixes #201 
See discussion on the issue.


kashizui:
not the desired solution

PR #209, opened 2015-11-07
Allow force rm of bundle on multiple worksheets

### LINKED ISSUE #208: rm with force should allow deleting bundles from multiple worksheets

kashizui:
Fixes #208 

@percyliang @pbhatnagar3


percyliang:
LGTM

PR #210, opened 2015-11-08
Use placeholder for info of private bundle

kashizui:
@percyliang @pbhatnagar3


kashizui:
wrong target branch, oops

Issue #202, opened 2015-11-04
Replace private/secret bundles with empty placeholders when retrieving info


kashizui:
rather than silently filtering them out

this causes a 500 server error on the website when trying to view bundle info of a private bundle (rather than providing any useful feedback)

kashizui:
@percyliang is this desired behavior?

I noticed that the `get_bundle_infos` function on the client just silently filters out unreadable bundles.

percyliang:
That was the intended behavior (to prevent leaking private information).
But we should make sure that the downstream code is aware of this contract
of only getting readable bundles.

kashizui:
Should we instead throw a `PermissionError` or pretend that it doesn't exist?

percyliang:
We still want to display a worksheet with a private bundle though.

Although it's weird if we just have missing bundles.

It might be cleaner then to always return an object for each bundle, even
the private ones, but just have no information on it other than the uuid
(which does reveal some information, but hopefully it's okay).

kashizui:
OK, re-orienting this issue.

kashizui:
This got a little more complicated than expected, since a lot of code depends on the bundle_info fields being populated with valid/non-empty values.

@percyliang Would it be better to modify all the code reading these fields to be able to handle these empty placeholders, or some how hack an empty bundle that would work with all existing code? (or other ideas)

percyliang:
Hm...I was was afraid that was going to happen.

I think it would be good to just be able to work with bundle_infos which
are {'uuid': ...} everywhere without putting fake info, etc.

What are some places where you need this?  Some helper functions might make
this less annoying?  At least a is_valid_bundle()?

kashizui:
So as you said we should make the code everywhere robust enough to handle incomplete bundle_infos (with only 'uuid' key) as necessary.

In addition, perhaps we can make a subclass of Bundle that is a PrivateBundle to serve as a placeholder?

I was also thinking it might be good to make the bundle_info dict creation a method of the Bundle classes...

percyliang:
On Sat, Nov 7, 2015 at 2:48 PM, Stephen Koo notifications@github.com
wrote:

> So as you said we should make the code everywhere robust enough to handle
> incomplete bundle_infos (with only 'uuid' key) as necessary.
> 
> Well, not universally maybe.  I think code that deals with bundle_info in
> the context of a worksheet should handle incomplete bundle_infos.  But code
> that just handles bundle_infos in general may not need to.

[...]

PR #211, opened 2015-11-08
Use placeholder for info of private bundle 

kashizui:
@percyliang @pbhatnagar3

### DIFF
--- a/test-cli.py
+++ b/test-cli.py
@@ -117,7 +117,7 @@ class ModuleContext(object):
             run_command([cl, 'wrm', '--force', worksheet])
 
         # Delete all bundles (dedup first)
-        if len(set(self.bundles)) > 0:
+        if len(self.bundles) > 0:
             run_command([cl, 'rm', '--force'] + list(set(self.bundles)))
 
         # Do not reraise exception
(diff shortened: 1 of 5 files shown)


test-cli.py
```diff
@@ -117,7 +117,7 @@ def __exit__(self, exc_type, exc_value, tb):
             run_command([cl, 'wrm', '--force', worksheet])
 
         # Delete all bundles (dedup first)
-        if len(self.bundles) > 0:
+        if len(set(self.bundles)) > 0:
```
percyliang: why set?

percyliang:
This makes me a bit nervous - probably good to double check everything (does the website have to do know about private bundles too now?) - but LGTM.

test-cli.py
kashizui: ..not actually necessary, you're right.

kashizui:
Tested the website, private bundles don't crash the website (see screenshot). The `PrivateBundle` subclass is just an easy way to tell the rest of the codebase that a private bundle doesn't have any metadata fields.

There are definitely better ways to do this, but until I can be more confident about how we'd want to refactor a lot of things, I think this is the smoothest solution. I do feel like I'm being significantly hampered by how the codebase is now (for example, permissions checks are only applied when one remembers to apply them, rather than there being just a few checkpoints that guarantee permission safety), and hopefully I'll be able to address these soon.

screen shot 2015-11-07 at 5 33 14 pm

kashizui:
This fixes #202 
and also fixes codalab/codalab#1371

percyliang:
Ok, sounds good!

Issue #212, opened 2015-11-08
program versus dataset bundles


percyliang:
Currently, we have a distinction between program and dataset bundles.  But in practice, this is not really a meaningful distinction (for example, what are libraries or models - are they programs or datasets?).  So let's get rid of this.  For now, make everything a dataset bundle...

Interface change:

```
cl upload file
```

rather than

```
cl upload dataset|program file
```

Issue #152, opened 2015-10-01
cl upload does not handle filenames with spaces correctly


kashizui:
When I run `cl upload dataset ~/Cloud/.sync/Archive/Documents/SCOr/05-31-15\ SCOr.mp4 -n scorvideo`
I get

```
BundleStore.upload: copying /Users/sckoo/Cloud/.sync/Archive/Documents/SCOr/05-31-15 SCOr.mp4 to /Users/sckoo/.codalab/temp/bbbd76f2a6d249e6ae3d0a3e15448627rsync: link_stat "/Users/sckoo/Cloud/.sync/Archive/Documents/SCOr/05-31-15" failed: No such file or directory (2)
rsync: link_stat "/Users/sckoo/Developer/codalab-cli/SCOr.mp4" failed: No such file or directory (2)
rsync error: some files could not be transferred (code 23) at /SourceCache/rsync/rsync-45/rsync/main.c(992) [sender=2.6.9]
 UsageError: Unable to copy /Users/sckoo/Cloud/.sync/Archive/Documents/SCOr/05-31-15 SCOr.mp4 to: /Users/sckoo/.codalab/temp/bbbd76f2a6d249e6ae3d0a3e15448627
```

Maybe this has something to do with how arguments are passed to rsync?

kashizui:
Actually this might just be me not knowing how quotes work in the shell, will close asap if that's the case...

kashizui:
Also fails even if the command itself doesn't contain the spaces, but there are spaces in the current working directory:

```
sckoo@hermes-2:~/Cloud/Documents/Reading/Course Notes$ cl upload dataset Complex1.pdf
BundleStore.upload: copying /Users/sckoo/Cloud/Documents/Reading/Course Notes/Complex1.pdf to /Users/sckoo/.codalab/temp/752b1f5428314156bf8f6db6ff3b2e48rsync: link_stat "/Users/sckoo/Cloud/Documents/Reading/Course" failed: No such file or directory (2)
rsync: link_stat "/Users/sckoo/Cloud/Documents/Reading/Course Notes/Notes/Complex1.pdf" failed: No such file or directory (2)
rsync error: some files could not be transferred (code 23) at /SourceCache/rsync/rsync-45/rsync/main.c(992) [sender=2.6.9]
[...]

Issue #21, opened 2015-01-05
allow cl upload urls with ? in the url


Raisins:
Ticket https://github.com/codalab/codalab/issues/783 was fixed on the front end but the backend rejects it because the name comes from the url when creating an uploaded bundle. 

You also get an error if doing it via command line
`cl upload http://pastebin.com/raw.php?i=V9UPN45u

Solution: strip characters in the URL that are not allowed in names.

PR #213, opened 2015-11-10
Improve documentation and upload

percyliang:
- Change 'cl upload dataset/program file' to 'cl upload file'
- Change 'cl ls worksheet' to 'cl ls -w worksheet'
- Add 'cl help -v' to show all help
- Added documentation for commands and made it more consistent
- When uploading bundle with weird name, gets sanitized to default name instead of throwing an error
- Fix quoting bug in copying files
  @kashizui please review

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -3,10 +3,16 @@ BundleCLI is a class that provides one major API method, do_command, which takes
 a list of CodaLab bundle system command-line arguments and executes them.
 
 Each of the supported commands corresponds to a method on this class.
-This function takes an argument list and an ArgumentParser and does the action.
+This function takes an argument list and does the action.
 
-  ex: BundleCLI.do_command(['upload', 'program', '.'])
-   -> BundleCLI.do_upload_command(['program', '.'], parser)
+For example:
+
+  cl upload foo
+
+results in the following:
+
+  BundleCLI.do_command(['upload', 'foo'])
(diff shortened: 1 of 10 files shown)


percyliang:
Fixes #212 #152 #21

codalab/lib/bundle_cli.py
```diff
@@ -695,11 +734,9 @@ def do_upload_command(self, args):
             raise UsageError('Nothing to upload')
 
         # Pull out the upload bundle type from the arguments and validate it.
-        if args.bundle_type not in UPLOADED_TYPES:
-            raise UsageError('Invalid bundle type %s (options: [%s])' % (
-              args.bundle_type, '|'.join(sorted(UPLOADED_TYPES)),
-            ))
-        bundle_subclass = get_bundle_subclass(args.bundle_type)
+        # Note: only allow dataset bundles (eventually deprecate the program bundle and just have )
+        bundle_type = 'dataset'
```
kashizui, line 738: Should we potentially just set this to `'uploaded'` (and add the appropriate fields in `UploadedBundle` to make it a concrete type)?

kashizui:
LGTM!

codalab/lib/bundle_cli.py
percyliang, line 738: Eventually, but for now, everything in the system is a dataset or program
bundle_type, so let's leave it that way to not break anything.

PR #226, opened 2015-11-14
Only open bundle in UI when cl info with no flags

### LINKED ISSUE #222: cl info -f ... should not trigger opening a open UI action

kashizui:
Fixes #222 

@percyliang


percyliang:
LGTM

PR #227, opened 2015-11-15
edit worksheet to empty behavior changes

percyliang:
if user edits worksheet to empty, old behavior was cancel edit; new behavior is to delete everything in worksheet; this is because we don't have '// Editing ...' comments which make it harder to delete everything in a worksheet.
@kashizui


kashizui:
LGTM

PR #229, opened 2015-11-15
quote arguments to dispatch command

### LINKED ISSUE #183: quote arguments to dispatch-q and dispatch-torque

percyliang:
Patches security hole.  Fixes #183 
@kashizui

### DIFF
--- a/scripts/dispatch-torque.py
+++ b/scripts/dispatch-torque.py
@@ -59,12 +59,12 @@ if mode == 'start':
     if args.request_priority != None:
         resource_args.extend(['-p', args.request_priority])
 
-    stdout = get_output(ssh(['/usr/local/bin/qsub', '-o', '/dev/null', '-e', '/dev/null', '-q', 'nlp'] + resource_args + [args.script]))
+    stdout = get_output(ssh(['/usr/bin/env', 'qsub', '-o', '/dev/null', '-e', '/dev/null', '-q', 'nlp'] + resource_args + [args.script]))
     handle = stdout.strip()
     response = {'raw': stdout, 'handle': handle}
 elif mode == 'info':
     handles = sys.argv[2:]  # If empty, then get info about everything
-    stdout = get_output(ssh(['/usr/bin/qstat', '-f'] + handles))
+    stdout = get_output(ssh(['/usr/bin/env', 'qstat', '-f'] + handles))
     response = {'raw': ''}  # Suppress output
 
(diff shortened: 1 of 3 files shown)


scripts/dispatch-torque.py
```diff
@@ -42,28 +47,25 @@ def get_output(command):
...
+        resource_args.extend(['-l', 'mem=%d' % int(args.request_memory)])
     if args.request_queue != None:
-        resource_args += ' -q %s' % args.request_queue
+        resource_args.extend(['-q', args.request_queue])
     if args.request_priority != None:
-        resource_args += ' -p %s' % args.request_priority
+        resource_args.extend(['-p', args.request_priority])
 
-    stdout = get_output('qsub -o /dev/null -e /dev/null%s %s' % (resource_args, args.script))
+    stdout = get_output(ssh(['/usr/local/bin/qsub', '-o', '/dev/null', '-e', '/dev/null', '-q', 'nlp'] + resource_args + [args.script]))
```
kashizui: I know specifying absolute path is necessary for `shell=False`, but is it safe to assume that `qsub` lives here? Do you know if `/usr/bin/env qsub` would work here?

scripts/dispatch-torque.py
```diff
@@ -42,28 +47,25 @@ def get_output(command):
...
     handle = stdout.strip()
     response = {'raw': stdout, 'handle': handle}
 elif mode == 'info':
     handles = sys.argv[2:]  # If empty, then get info about everything
-    list_args = ''
-    if len(handles) > 0:
-        list_args += ' ' + ' '.join(handles)
-    stdout = get_output('qstat -f%s' % list_args)
-    response = {'raw': ''}
+    stdout = get_output(ssh(['/usr/bin/qstat', '-f'] + handles))
```
kashizui: same as above

scripts/dispatch-torque.py
```diff
@@ -120,7 +122,7 @@ def get_output(command):
     handle = sys.argv[2]
     response = {
         'handle': handle,
-        'raw': get_output('qdel %s' % handle)
+        'raw': get_output(ssh(['/usr/bin/qdel', handle]))
```
kashizui: same as above

percyliang: Thanks - switched to /usr/bin/env

kashizui:
LGTM

Issue #232, opened 2015-11-16
support GPUs in docker


percyliang:
When start docker container, need to attach the GPU devices, like this:

```
sudo docker run -ti --device /dev/nvidia0:/dev/nvidia0 --device /dev/nvidiactl:/dev/nvidiactl --device /dev/nvidia-uvm:/dev/nvidia-uvm tleyden5iwx/ubuntu-cuda /bin/bash
```

Source: http://stackoverflow.com/questions/25185405/using-gpu-from-a-docker-container

That way, programs can access the GPUs.

percyliang:
Fixed by #234

PR #234, opened 2015-11-16
allow docker container to access host's GPUs

percyliang:
Addresses #232 
Of course, the docker image needs to the right version of CUDA installed matching the kernel of the host.  For example, one might use: https://hub.docker.com/r/codalab/tensorflow-cuda7.0-352.39/
@kashizui


kashizui:
LGTM if it works!

from `docker help run` I see

```
  --device=[]                     Add a host device to the container
```

though the equals sign is probably optional?

PR #244, opened 2015-11-19
Dict key should be str fix

pujun-ai:
@percyliang @kashizui please review. I have changed the index to be strings so that the RPC does not freak out in  certain cases. Also modified how   `interpreted_items_raw_items_map` is made to improve navigation and jumping around modes in ace.


percyliang:
Please document that the maps keys are not strings...give an example.

LGTM

PR #248, opened 2015-11-22
Sanitize display contents

kashizui:
Binary and other un-encodable lines get replaced.

Fixes codalab/codalab#1423


kashizui:
wrong target branch

PR #249, opened 2015-11-22
Sanitize display contents

kashizui:
Binary and other un-encodable lines get replaced.

Fixes codalab/codalab#1423

@percyliang please review

### DIFF
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -474,13 +474,16 @@ class LocalBundleClient(BundleClient):
     # Maximum number of bytes to read per line requested
     MAX_BYTES_PER_LINE = 128
 
-    def head_target(self, target, max_num_lines):
+    def head_target(self, target, max_num_lines, sanitize=False):
         max_total_bytes = None if max_num_lines is None else max_num_lines * self.MAX_BYTES_PER_LINE
         check_bundles_have_read_permission(self.model, self._current_user(), [target[0]])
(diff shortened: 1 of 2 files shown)


codalab/client/local_bundle_client.py
```diff
@@ -958,7 +961,7 @@ def resolve_interpreted_items(self, interpreted_items):
                 if 'type' not in info:
                     data = None
                 elif info['type'] == 'file':
-                    data = self.head_target(data, int(properties.get('maxlines', 10)))
+                    data = self.head_target(data, int(properties.get('maxlines', 10)), sanitize=True)
```
percyliang: Not related to this PR, but make 10 a CONSTANT

codalab/client/local_bundle_client.py
```diff
@@ -474,7 +474,7 @@ def cat_target(self, target, out):
     # Maximum number of bytes to read per line requested
     MAX_BYTES_PER_LINE = 128
 
-    def head_target(self, target, max_num_lines):
+    def head_target(self, target, max_num_lines, sanitize=False):
```
percyliang: Not sure 'sanitize' the right variable here.
From this code, it's not clear that what is going on is that '<binary>' is returned if it's a binary file and otherwise, we return the original string.

kashizui:
@percyliang please review changes

percyliang:
LGTM!

PR #250, opened 2015-11-22
Go home after wrm-ing current worksheet

kashizui:
Fixes codalab/codalab#1425

@percyliang please review

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1879,8 +1879,7 @@ class BundleCLI(object):
 
         if delete_current:
             # Go to home worksheet
-            client, _ = self.parse_client_worksheet_uuid('')
-            return self.change_current_worksheet(client, None, verbose=True)
+            return self.change_current_worksheet(current_client, None, verbose=True)
 
     @Commands.command(
         'wadd',
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1857,9 +1857,24 @@ class BundleCLI(object):
         ),
     )
     def do_wrm_command(self, args):
+        delete_current = False
         for worksheet_spec in args.worksheet_spec:
             client, worksheet_uuid = self.parse_client_worksheet_uuid(worksheet_spec)
+            if (client, worksheet_uuid) == self.manager.get_current_worksheet_uuid():
+                delete_current = True
             client.delete_worksheet(worksheet_uuid, args.force)
+
+        if delete_current:
+            # Go to home worksheet
+            client, _ = self.parse_client_worksheet_uuid('')
+            home_worksheet_uuid = client.get_worksheet_uuid(None, '')
+
+            if self.headless:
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1857,10 +1857,25 @@ def do_wls_command(self, args):
         ),
     )
     def do_wrm_command(self, args):
+        delete_current = False
         for worksheet_spec in args.worksheet_spec:
             client, worksheet_uuid = self.parse_client_worksheet_uuid(worksheet_spec)
+            if (client, worksheet_uuid) == self.manager.get_current_worksheet_uuid():
```
percyliang: Call get_current_worksheet_uuid outside the for loop - saves extra calls (just in case).

codalab/lib/bundle_cli.py
```diff
@@ -1857,10 +1857,25 @@ def do_wls_command(self, args):
...
+                delete_current = True
             client.delete_worksheet(worksheet_uuid, args.force)
 
+        if delete_current:
+            # Go to home worksheet
```
percyliang, line 1881: Can we just call do_work_command to not duplicate code?

percyliang:
Note that wrm'ing the home worksheet when you're on it is possible.  But don't know what you would do in that case...

codalab/lib/bundle_cli.py
kashizui, line 1881: We can, but it would involve artificially creating an `args` object. What do you think?

kashizui:
It will just create the worksheet again in the call to
client.get_worksheet_uuid... perhaps not desired behavior?

percyliang:
It means people can't really delete their worksheet, which is slightly weird, but I think okay.

codalab/lib/bundle_cli.py
percyliang, line 1881: Can we refactor a bit to avoid duplicate code and fake args?

kashizui, line 1881: Sure—will take a stab at it.

kashizui:
@percyliang please review changes

codalab/lib/bundle_cli.py
```diff
@@ -1857,10 +1869,19 @@ def do_wls_command(self, args):
...
+        current_client, current_worksheet = self.manager.get_current_worksheet_uuid()
         for worksheet_spec in args.worksheet_spec:
             client, worksheet_uuid = self.parse_client_worksheet_uuid(worksheet_spec)
+            if (client, worksheet_uuid) == (current_client, current_worksheet):
+                delete_current = True
             client.delete_worksheet(worksheet_uuid, args.force)
 
+        if delete_current:
+            # Go to home worksheet
+            client, _ = self.parse_client_worksheet_uuid('')
```
percyliang: why not use current_client?

percyliang:
LGTM

codalab/lib/bundle_cli.py
kashizui: Nice catch.

PR #256, opened 2015-11-23
Handle unicode properly

kashizui:
I can now print Chinese characters, while hiding binary content (as in a PDF or executable).

I misunderstood slightly how Unicode worked the first time around, sorry about that!

@percyliang please review


percyliang:
Thanks, LGTM!

PR #257, opened 2015-11-23
update worksheet_util for ace-navigation

pujun-ai:
@percyliang @kashizui This is almost finalized but would still say that is a work in progress since I have not fully tested out the `% search` and `% w_search`.

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -620,11 +620,18 @@ def interpret_items(schemas, items):
 
     # Mapping from worksheet_info['raw'] index to |(focusIndex, subFocusIndex)| based on interpreted_items.
     # Used to intelligently focus on the worksheet_item where the cursor was present when exiting the editor.
-    raw_items_interpreted_items_map = {}
+    raw_interpreted_map = {}
 
     # Reverse mapping from interpreted_items indexes to worksheet_info['raw'] indexes.
     # Used to intelligently place the cursor at the focused worksheet item while entering the 'edit' mode.
-    interpreted_items_raw_items_map = {}
+    interpreted_raw_map = {}
+
+    # A list to store |((focus_index, sub_focus_index), end_index)| which is used to populate raw_interpreted_map.
+    raw_interpreted_list = []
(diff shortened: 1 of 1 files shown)


codalab/lib/worksheet_util.py
```diff
@@ -589,6 +589,12 @@ def interpret_items(schemas, items):
     # Used to intelligently place the cursor at the focused worksheet item while entering the 'edit' mode.
     interpreted_items_raw_items_map = {}
 
+    raw_items_interpreted_items_list = []
```
percyliang: Document the high-level logic and the invariants.

codalab/lib/worksheet_util.py
```diff
@@ -606,13 +612,13 @@ def get_schema(args):  # args is a list of schema names
     def is_missing(info):
         return 'metadata' not in info
 
-    def flush_bundles():
+    def flush_bundles(start_index, focus_index, sub_focus_index):
         """
```
percyliang, line 655: Document what this function returns

percyliang:
This would be a good function to have a unit test for.

Also, how about calling it interpreted_to_raw instead of interpreted_items_raw_items_map to make things more concise?

pujun-ai:
@percyliang yeah good point. I have renamed the maps.

codalab/lib/worksheet_util.py
```diff
@@ -749,11 +761,22 @@ def raise_usage_error():
...
 
     def get_command(value_obj):  # For directives only
         return value_obj[0] if len(value_obj) > 0 else None
 
-    last_was_empty_line = False
+    def update_interpreted_raw_map(start_index, end_index, focus_index, sub_focus_index=-1):
+        interpreted_raw_map[str(focus_index) + ',' + str(sub_focus_index)] = end_index
+        raw_interpreted_list.append(((focus_index, sub_focus_index), end_index))
+
+    def update_indexes(end_index, focus_index):
```
kashizui: nit: This function has no side-effects, so `updated_indices` or `next_indices` might make more sense.

kashizui:
LGTM

PR #262, opened 2015-11-26
Overhaul of the uploading/downloading system.

percyliang:
Note: changes the BundleClient interface, so users need to git pull on the codalab-cli.
- Switched over to tar.gz instead of zip to allow streaming.  This way, when we
  download a large bundle, stuff starts showing up on the client immediately.
- Minimize the amount of copying to temporary files and zipping before even
  starting to copy.
- Support `cl upload https://path-to-git-repo --git`
- Support `cl upload https://path-to-archive`, where archive is one of
  {.tar.gz, .tgz, .tar.bz2, .zip} and the archive gets automatically unpacked.
- Created more comprehensive tests for uploading/downloading (e.g., symlinks,
  funny characters, etc.)
- Created tests that allow testing copying bundles across instances.

@kashizui please review


codalab/server/bundle_rpc_server.py
```diff
@@ -97,39 +99,37 @@ def function_to_register(*args, **kwargs):
         for command in RemoteBundleClient.SERVER_COMMANDS:
             self.register_function(wrap(self, command), command)
 
-    def upload_bundle_zip(self, file_uuid, construct_args, worksheet_uuid, follow_symlinks, add_to_worksheet):
+    def finish_upload_bundle(self, file_uuids, unpack, info, worksheet_uuid, add_to_worksheet):
```
kashizui, line 102: great name change!

codalab/lib/zip_util.py
```diff
@@ -7,83 +7,95 @@
...
+    # Unpack to a temporary location.
+    tmp_path = tempfile.mkdtemp('-zip_util.unpack')
+    if isinstance(source, basestring):
+        source_path = source
+        if source_path.endswith('tar.gz') or source_path.endswith('tgz'):
+            subprocess.call(['tar', 'xfz', source_path, '-C', tmp_path])
+        elif source_path.endswith('tar.bz2'):
+            subprocess.call(['tar', 'xfj', source_path, '-C', tmp_path])
+        elif source_path.endswith('zip'):
+            subprocess.call(['unzip', '-q', source_path, '-d', tmp_path])
```
kashizui: is this safe against zip bombs? I'm unfamiliar with them @pbhatnagar3

codalab/lib/zip_util.py
```diff
@@ -7,83 +7,95 @@
...
+    """
+    Unpack the archive |source_path| to |dest_path|.
+    Note: |source| can be a file handle or a path.
+    """
+    # Unpack to a temporary location.
+    tmp_path = tempfile.mkdtemp('-zip_util.unpack')
+    if isinstance(source, basestring):
+        source_path = source
+        if source_path.endswith('tar.gz') or source_path.endswith('tgz'):
+            subprocess.call(['tar', 'xfz', source_path, '-C', tmp_path])
```
kashizui: Should we check for exit codes here?

codalab/lib/bundle_cli.py
```diff
@@ -1068,7 +1103,7 @@ def do_rm_command(self, args):
         if args.dry_run:
             print >>self.stdout, 'This command would permanently remove the following bundles (not doing so yet):'
             bundle_infos = client.get_bundle_infos(deleted_uuids)
-            bundle_info_list = [bundle_infos[uuid] for uuid in deleted_uuids]
+            bundle_info_list = [bundle_infos[uuid] for uuid in deleted_uuids for uuid in bundle_infos]
```
kashizui: `uuid` references clashing

codalab/lib/bundle_cli.py
```diff
@@ -1091,7 +1126,7 @@ def do_rm_command(self, args):
         arguments=(
             Commands.Argument('keywords', help='Keywords to search for.', nargs='+'),
             Commands.Argument('-a', '--append', help='Append these bundles to the current worksheet.', action='store_true'),
-            Commands.Argument('-u', '--uuid-only', help='Print only uuids.', action='store_true'),
+            Commands.Argument('-u', '--uuid_only', help='Print only uuids.', action='store_true'),
```
kashizui, line 1129: Why use underscore instead of dash here? (and elsewhere)

kashizui:
My confidence in reviewing such a huge diff is not great, but mostly looks good to me!

codalab/lib/zip_util.py
percyliang: Added a TODO

codalab/lib/zip_util.py
percyliang: good point - added

codalab/lib/bundle_cli.py
percyliang: oops, second for is supposed to be an if

codalab/lib/bundle_cli.py
percyliang, line 1129: Wanted to be consistent everywhere.  I'd actually prefer dashes everywhere, but was held up because the metadata specs automatically generates options with underscores.  We could easily change that though.  Would have to update the docs too.  Let's save for a different PR.

PR #268, opened 2015-11-28
Support for graphing + redoing interpret_items.

percyliang:
Details:
- Added support for graphing (just pass TSV back to frontend).
- Renamed things in interpret_items to make it more readable.
- Rewrote read_lines with truncation to make it correct when truncation level is small.
- Rewrote logic for performing mapping between raw and interpreted indices to make things simpler.
- Directive properties is now 'width=100 height=100' rather than 'width=100,height=100'.
  @pbhatnagar3 please review


pujun-ai:
This is great :) . LGTM :+1:

PR #269, opened 2015-11-29
Support resources accounting / limits.

percyliang:
- Added a table to keep of per-usage time/disk usage/quotas.
- When running a job, enforce memory, disk, and network usage (via bash).
- Added fields for started / last_updated for run bundles.
- Generic support for boolean metadata fields (e.g., request_network).
- 'cl uedit' to edit quotas.
- 'cl status' to see what the existing usage is.
- Deprecate support for non-docker work manager.
- Work manager now requires max_request_time/memory/disk to enforce control.
- Save request_time/memory/disk into integers rather than strings.
- Formatting sizes is now '3k' instead of '3K'.
- Save default request_\* into the bundle for explicitness.
- Update the q script to allow starting workers that have maxTime.
- Added a test to check if resource constraints are working properly.
  @kashizui please review


codalab/lib/formatting.py
```diff
@@ -125,9 +129,7 @@ def parse_duration(s):
         return n * 60 * 60 * 24
     if unit == 'y':
         return n * 60 * 60 * 24 * 365
-    # If error, ignore unit
-    print >>sys.stderr, 'Warning: invalid unit in ', s
-    raise n
+    raise BaseException('Invalid duration: %s, expected <number>[<s|m|h|d|y>]' % s)
```
kashizui: nit: Perhaps `ValueError` more precise?

kashizui:
Again self-confidence not high here but LGTM AFAIK!

PR #270, opened 2015-12-01
use dashes instead of underscores (e.g., '--dry-run' rather than '--dry_run')

percyliang:
@kashizui please review


kashizui:
LGTM

PR #271, opened 2015-12-01
Autocomplete for aliases and target specs

### LINKED ISSUE #223: better autocomplete for 'cl run'
When type ':' tab, should complete with possible bundles.
When begin the command, then complete with possible dependency keys (the bundle name if the key is empty).
When type 'main:: tab, should complete with worksheets on instance main.

kashizui:
Partially fixes #223 

@percyliang


codalab/lib/completers.py
```diff
@@ -67,44 +114,55 @@ def __call__(self, prefix, action=None, parsed_args=None):
...
+        ])
+
+        if subpath is None:
+            # then suggest completions for bundle_spec
+            return (suggestion_format.format(b) for b in BundlesCompleter(self.cli)(bundle_spec, action, parsed_args))
+        else:
+            # then suggest completions for subpath
+            target = self.cli.parse_target(client, worksheet_uuid, bundle_spec)
+            info = client.get_target_info(target, 0)
+            if info['type'] == 'directory':
```
percyliang: 'type' might not exist for non-existent paths

codalab/lib/completers.py
```diff
@@ -1,39 +1,84 @@
...
+def require_not_headless(completer):
+    """
+    Given a completer, return a CodaLabCompleter that will only call the
+    given completer if the client is not headless.
+    """
+    class SafeCompleter(CodaLabCompleter):
+        def __call__(self, *args, **kwargs):
+            if self.cli.headless:
+                return ()
+            elif inspect.isclass(completer):
```
percyliang: Do we really need to multiplex between a class and an instance?  Why not have the caller add '()' if necessary?

percyliang:
LGTM

codalab/lib/completers.py
kashizui: No reason-- fixing now

PR #274, opened 2015-12-02
Fix clashing argument destinations

### LINKED ISSUE #263: cl events doesn't work due to argument parsing
`cl events` doesn't work properly (returns empty) because args.command gets set to 'events' for some reason instead of None.  Seems to be a problem with the command-line arguments parsing.

kashizui:
`cl events` now works on python 2.7.6

fixes #263 

@percyliang please review


percyliang:
LGTM

PR #282, opened 2015-12-09
Downgrade line mapping assertion to warning

kashizui:
This fixes the unrecoverable error when worksheet has something like `% display bad`.

In this case, the line mapping still works regardless, but it may cause some inconsistent mappings in the frontend editor in the future if the worksheet contains a lot of errors. I'm claiming that this isn't too bad a problem to have until we can have more time to clean up the worksheet interpretation, since jumping to the wrong line with a bad worksheet isn't life or death.

@percyliang @pbhatnagar3 please review

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -29,6 +29,7 @@ See get_worksheet_lines for documentation on the specification of the directives
 import copy
 import os
 import re
+import sys
 import types
 import yaml
 import json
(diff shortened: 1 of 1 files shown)


codalab/lib/worksheet_util.py
```diff
@@ -938,8 +939,10 @@ def get_command(value_obj):  # For directives only
         finally:
             last_was_empty_line = new_last_was_empty_line
 
+    if len(raw_to_interpreted) != len(raw_items):
```
percyliang, line 943: Put a TODO here, because this really shouldn't happen...we can fix it after the demo.

percyliang:
LGTM

PR #283, opened 2015-12-09
Change argument order for tar archiving

kelvinguu:
Options must be specified before the source file. This should allow the --exclude options to be processed correctly.


kashizui:
Really sorry to make you do this, but is it possible for you to re-submit this pull request to be merged into the `develop` branch rather than `master`? We need to figure out how to make this default...

kelvinguu:
Ok, sounds good. I'll make a new pull request, and might as well try to fix the folder-in-folder problem as well before I do that.

Issue #243, opened 2015-11-18
cl upload -x seems to be broken (exclude pattern)


kelvinguu:
Example:

``` bash
$ cl upload -x '*.ipynb' src
UsageError: Nothing to upload
```

If I change the argument order:

``` bash
$ cl upload src -x '*.ipynb'
```

`src` is uploaded, but ipynb files aren't excluded.

kashizui:
@kelvinguu is this happening while connected to a remote bundle service (e.g. codalab.stanford.edu)?

kashizui:
Never mind, doesn't look like it really matters, but still good to know just in case—thanks in advance!

kelvinguu:
Ah yes, it was on codalab.stanford.edu. Thanks!

kashizui:
Looks like the first case didn't work because the `-x` flag eats all the following arguments—should this be undesirable behavior, or just better documented?

I'm unable to reproduce the second case—my files are excluded as specified. The CLI simply makes a command to `rsync`, with the arguments to `-x` passed as `--exclude` arguments, and I know `rsync` sometimes has different behavior between OS distros. Could you maybe send me your `rsync` version number and/or manpage?

kashizui:
I'm unable to reproduce the second case—my files are excluded as specified. Any thoughts on why it might not be working for you?

kashizui:
The CLI simply makes a command to `rsync`, with the arguments to `-x` passed as `--exclude` arguments, and I know `rsync` sometimes has different behavior between OS distros. Could you maybe send me your `rsync` version number and/or manpage?

kashizui:
@kelvinguu are you still getting this issue? There have been some recent improvements (on the `develop` branch of the CLI) in the uploading procedures. I'm currently still unable to reproduce your problem.

percyliang:
We will push develop to master tonight and then @kelvinguu can try it out
again

kelvinguu:
Thanks! I'll be sure to try it.

kelvinguu:
@kashizui, thanks, in my latest attempts, it works fine now.

kashizui:
Awesome, thanks for your help @kelvinguu.

kelvinguu:
Unfortunately, this problem came back. Here's what I see now:

``` bash
$ cl upload src -x '*.ipynb'
tar: --exclude=*.ipynb: Cannot stat: No such file or directory
tar: Error exit delayed from previous errors.
```

[...]

PR #286, opened 2015-12-20
Change argument order for tar archiving

kelvinguu:
For some versions of tar, options must be specified before the source file. This should allow the --exclude options to be processed correctly.

I verified that this works on Mac OS X Yosemite, and on CentOS 6.7.


percyliang:
Fixes #243

Issue #264, opened 2015-11-26
autocomplete for docker containers


percyliang:
If type `--request-docker-image`, then should autocomplete with all the docker containers that are in the system.  This way, it's easy for people to choose from the popular docker containers (e.g., for TensorFlow).

The list of docker containers should be gotten dynamically by calling `docker search`.  Perhaps this should be cached.

a10y:
We're using argparse so should be able to use this: https://pypi.python.org/pypi/argcomplete/1.0.0

kashizui:
We are currently using argcomplete—you can take a look at `completers.py` (and `bundle_cli.py`) for how to add more autocompletion functionality.

a10y:
@kashizui thanks!

a10y:
@percyliang @kashizui did you mean `docker search`, which searches DockerHub? Or should we complete off of the images that are available on the worker machine?

percyliang:
Yes, we should use 'docker search', which searches DockerHub (or maybe they
have a more direct API?)  Docker images that don't exist on the local
machines will be pulled automatically from the hub.

a10y:
@percyliang sounds good. is the auto-pulling already implemented or is that going to be another change?

percyliang:
docker already does auto-pulling

a10y:
Implemented in https://github.com/codalab/codalab-cli/pull/290

PR #290, opened 2015-12-21
#264: Docker completion

a10y:
Implements #264 

**NOTE**: Upfront, should say that this doesn't currently filter images that start with `prefix`, which I actually think might be helpful because you can just type ex. `python` and see all completions that include the prefix. Is a departure from the current behavior with the static list though.

Things that could use review:
- **import-in-function-body syntax**: not sure if it's necessary but Popen isn't used anywhere else in completers so I figured it would avoid misleading people by placing it at the top. Could be moved.
- **General strategy of wrapping calls to the docker binary as opposed to using the Docker API**: I tried using docker-py, but there were some complications with SSL that after pouring ~1 hour in didn't get me anywhere. If anyone has a strong opinion we could discuss this, but it seems like using the Remote API might be a little more difficult than it's worth, and relying on the `docker` binary being installed on each worker might be the better route.

All comments welcome!


percyliang:
Thanks, Andrew.

I think using calling 'docker search' is good.  I was hoping there would be a flag to have it output JSON so you don't have to parse it using regex's (speaking of which, we should do that for `cl`), but this is fine.  Perhaps document a bit more and show an example of what the output of docker search looks like.

Popen is fine.

I would create a separate Docker class with a 'search' command, so we can test this functionality independent of command-line completion.

Edge case: If you do a completion on 'codalab/', then things fail because `docker search` treats the trailing slash specially somehow.  This should be fixed.

In general, if there are errors here, we probably should have a way of showing them (just dump to stderr, maybe).  Currently, things just fail silently, which is always bad.

a10y:
@percyliang 

> I would create a separate Docker class with a 'search' command, so we can test this functionality independent of command-line completion.

Will do, I'll push another commit to this branch with the refactor

> Edge case: If you do a completion on 'codalab/', then things fail because docker search treats the trailing slash specially somehow. This should be fixed.

I can't seem to reproduce this, currently on OS X I can list 4 docker containers after typing `codalab/`:
screen shot 2015-12-21 at 00 22 36

> In general, if there are errors here, we probably should have a way of showing them (just dump to stderr, maybe). Currently, things just fail silently, which is always bad.

Cool. I silenced stderr because I figured that we wouldn't want stderr of the raw commands spamming you while you're tabbing and e.g. offline. What should the error include in it? Just the returncode of the process, or all of stderr, etc?

a10y:
Made the changes, tested the stderr output by turning off wifi and trying to perform tab completion, gives this dump (but configurable)

screen shot 2015-12-21 at 01 05 59

codalab/lib/docker_util.py
```diff
@@ -0,0 +1,37 @@
...
+
+    This captures the beginning of the line up until the first whitespace character, i.e.
+    the first column of the output, corresponding to the image tag.
+    """
+    DOCKER_SEARCH_TAG_REGEX = re.compile(r'^(?P<tag>\S+)\s+')
+
+    @classmethod
+    def search(cls, term, failure_cb=None):
+        """
+        Performs`docker search <term>`. Returns tuple of image tags matching the search term.
```
percyliang: missing space before `

codalab/lib/docker_util.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Compiled regular expression to parse the stdout of `docker search` calls.
+
+    This captures the beginning of the line up until the first whitespace character, i.e.
+    the first column of the output, corresponding to the image tag.
+    """
+    DOCKER_SEARCH_TAG_REGEX = re.compile(r'^(?P<tag>\S+)\s+')
+
+    @classmethod
+    def search(cls, term, failure_cb=None):
```
percyliang: nit: rename 'term' to 'keyword' to be consistent with name of such variables in bundle_model.py

codalab/lib/completers.py
```diff
@@ -173,6 +174,15 @@ def __call__(self, prefix, action=None, parsed_args=None):
...
+    """
+    IMAGE_TAG_REGEX = re.compile(r'^(?P<tag>\S+)\s+')
 
+    """
+    Completes names of Docker images available local to the user of the cl script, which should search the
+    """
+    def __call__(self, prefix, action=None, parsed_args=None):
+        def handle_err(errcode, stderr):
+            warn("Error(%d): %s" % (errcode, stderr))
+        return Docker.search(prefix, handle_err)
```
percyliang: missing newline

percyliang:
% docker -v
Docker version 1.7.1, build 786b29d/1.7.1
% docker search codalab/
Invalid repository name (), only [a-z0-9-_.] are allowed
Perhaps we need to hack things and strip off '/'?

docker has pretty weird behavior (which we might not be able to do anything about):
jagupard11:~% docker search codalab/p
NAME      DESCRIPTION   STARS     OFFICIAL   AUTOMATED
jagupard11:~% docker search codalab/py
NAME                            DESCRIPTION   STARS     OFFICIAL   AUTOMATED
codalab/python                                0  
codalab/python-cuda7.0-352.39                 0                    

Other than this, the PR looks good!

a10y:
@percyliang 

Implemented the style fixes.

After some experiments, it seems that when you supply multiple space-separated words to `docker search` it does a boolean OR query, which will return more results from DockerHub than we actually need. This is a minor optimization but instead I have it just trim the search query to the position of the first '/' character, and then rely on argcomplete's default validation filtering to filter that down to the results that are prefixed with what the user actually entered. Tested this and it seems to yield what's desired for the queries that were failing before.

**EDIT**: Once this has cleared review, do you prefer that commits are squashed before merging?

codalab/lib/docker_util.py
```diff
@@ -0,0 +1,37 @@
...
+
+    This captures the beginning of the line up until the first whitespace character, i.e.
+    the first column of the output, corresponding to the image tag.
+    """
+    DOCKER_SEARCH_TAG_REGEX = re.compile(r'^(?P<tag>\S+)\s+')
+
+    @classmethod
+    def search(cls, keyword, failure_cb=None):
+        """
+        Performs `docker search <term>`. Returns tuple of image tags matching the search term.
```
percyliang: term => keyword

codalab/lib/completers.py
```diff
@@ -173,6 +174,18 @@ def __call__(self, prefix, action=None, parsed_args=None):
 
 
 class DockerImagesCompleter(CodaLabCompleter):
+    """
+    Matches the first column image tag output in 'docker search'
+    """
+    IMAGE_TAG_REGEX = re.compile(r'^(?P<tag>\S+)\s+')
+
+    """
+    Completes names of Docker images available local to the user of the cl script, which should search the
```
percyliang: incomplete sentence

percyliang:
Modulo trivial comments, LGTM (go ahead and merge)!

We've been just keeping the full history (the philosophy is to err on the side of keeping around information, figuring that if we wanted to make things look nice, we can do that post-hoc).  But don't feel especially strongly about this, so open to arguments the other way too.

a10y:
Got it, merged it with the full history!

Issue #291, opened 2015-12-24
Move to using docker-py


a10y:
After doing a little more digging, I found this issue in docker-py: https://github.com/docker/docker-py/issues/465

This way, we can use docker-py directly instead of having to call out to the `docker` binary installed on the worker.

I'll have a PR in soon that accomplishes this for #264, and then for #285 I'll use docker-py as well.

PR #292, opened 2015-12-24
Change over use of docker binary in Docker to use docker-py

a10y:
PR for #291 

@percyliang

### DIFF
--- a/codalab/lib/completers.py
+++ b/codalab/lib/completers.py
@@ -185,6 +185,8 @@ class DockerImagesCompleter(CodaLabCompleter):
     def __call__(self, prefix, action=None, parsed_args=None):
         def handle_err(exc):
             warn("Error: %s" % exc)
+        if prefix == "":
+            prefix = "codalab"
         first_slash = prefix.find('/')
         trimmed_prefix = prefix[0:first_slash] if first_slash >= 0 else prefix
         return Docker.search(trimmed_prefix, handle_err)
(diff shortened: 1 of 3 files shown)


a10y:
Note: I've only tested this on docker 1.9.1, and seems to be working fine. I'm not certain if the features of the Docker Remote API are different enough across the recent versions of docker that it would lock us into using the most recent.

requirements.txt
```diff
@@ -1,7 +1,16 @@
...
+pathtools==0.1.2
+psutil==3.3.0
+python-editor==0.5
 PyYAML==3.11
+requests==2.9.0
+six==1.10.0
 SQLAlchemy==1.0.8
-alembic==0.8.2
 watchdog==0.8.3
-# Use sckoo@stanford.edu's patched argcomplete, until pull request has been accepted
```
percyliang, line 5: Keep comment

percyliang:
One tangential request: if user types `--request-docker-image <tab>`, then we should complete with things that begin with `codalab/...`, since novice users will want sane defaults here.  But if the user starts typing something that doesn't start with codalab, we can still do the completion as is done now.

a10y:
Last commit in, this PR should be GTG

codalab/lib/completers.py
```diff
@@ -183,8 +183,10 @@ class DockerImagesCompleter(CodaLabCompleter):
     Completes names of Docker images available on DockerHub.
     """
     def __call__(self, prefix, action=None, parsed_args=None):
-        def handle_err(errcode, stderr):
-            warn("Error(%d): %s" % (errcode, stderr))
+        def handle_err(exc):
+            warn("Error: %s" % exc)
```
percyliang, line 187: nit: use single quotes

percyliang:
LGTM!

PR #310, opened 2016-01-13
read sender/receiver email from website-config.json

percyliang:
@andreweduffy please review


a10y:
One thing is that SMTP_SSL never seems to be used, is it supposed to be configurable? Otherwise LGTM

percyliang:
Removed SMTP_SSL

PR #314, opened 2016-01-23
Created CircleCI configuration

a10y:
#275
- config includes upgrading pip so that it has newer version of setuptools needed to install mock

### DIFF
--- a/scripts/ci-tests.sh
+++ b/scripts/ci-tests.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# Setup environment to run tests
+cd $(dirname $0)/..
+export PATH="$HOME/codalab-cli/codalab/bin:$PATH"
+
+
+# Run all tests save for resources and write, which depend on Docker.
+./venv/bin/python test-cli.py unittest basic upload1 upload2 upload3 download refs rm make run worksheet worksheet_search worksheet_tags freeze detach perm search kill mimic status events batch copy
+
--- a/scripts/ci-tests.sh
+++ b/scripts/ci-tests.sh
@@ -7,5 +7,5 @@ export PATH="$HOME/codalab-cli/codalab/bin:$PATH"
 
 # Run all tests save for resources and write, which depend on Docker.
 # TODO: Add support for docker tests
-./venv/bin/python test-cli.py unittest basic upload1 upload2 upload3 download refs rm make run worksheet worksheet_search worksheet_tags freeze detach perm search kill mimic status events batch copy
+./venv/bin/python test-cli.py no-docker
 
--- a/scripts/ci-tests.sh
+++ b/scripts/ci-tests.sh
@@ -6,5 +6,6 @@ export PATH="$HOME/codalab-cli/codalab/bin:$PATH"
 
 
 # Run all tests save for resources and write, which depend on Docker.
+# TODO: Add support for docker tests
 ./venv/bin/python test-cli.py unittest basic upload1 upload2 upload3 download refs rm make run worksheet worksheet_search worksheet_tags freeze detach perm search kill mimic status events batch copy
 
--- a/test-cli.py
+++ b/test-cli.py
@@ -230,8 +230,10 @@ class TestModule(object):
         print '============= SUMMARY'
         if failed:
             print 'Tests failed: %s' % ', '.join(failed)
+            return False
         else:
             print 'All tests passed.'
+            return True
 
 ############################################################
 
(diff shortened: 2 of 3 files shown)


a10y:
@percyliang Example of the config running [here](https://circleci.com/gh/codalab/codalab-cli/14)

percyliang:
Great.  Could we have the test run `./test-cli.py all`?

a10y:
@percyliang 

I still cannot quite track down all the issues with running with q/docker, so I decided to leave it out for now. This runs all but the resources and write tests.

test-cli.py
```diff
@@ -472,10 +474,10 @@ def test(ctx):
     ctx.collect_worksheet(wuuid)
     # Add tags
     run_command([cl, 'wedit', wname, '--tags', 'foo', 'bar', 'baz'])
-    check_contains(['Tags: \\[\'foo\', \'bar\', \'baz\'\\]'], run_command([cl, 'ls', '-w', wuuid]))
+    check_contains(['Tags: \\[u\'foo\', u\'bar\', u\'baz\'\\]'], run_command([cl, 'ls', '-w', wuuid]))
```
percyliang: Hm, is this going to break on other versions of Python?  For example, the previous test passed for me, so it probably won't pass now.

test-cli.py
```diff
@@ -420,7 +422,7 @@ def test(ctx):
     check_contains(['run "echo hello"'], run_command([cl, 'info', '-f', 'args', uuid]))
     check_equals('hello', run_command([cl, 'cat', uuid+'/stdout']))
     # block
-    uuid2 = check_contains('hello', run_command([cl, 'run', 'echo hello', '--tail'])).split('\n')[0]
+    uuid2 = check_contains('hello', run_command([cl, 'run', 'echo hello', '--tail']))
```
percyliang: Why?  The output is uuid on the first line, 'hello' on the second, no?

scripts/ci-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Setup environment to run tests
+cd $(dirname $0)/..
+export PATH="$HOME/codalab-cli/codalab/bin:$PATH"
+
+
+# Run all tests save for resources and write, which depend on Docker.
+./venv/bin/python test-cli.py unittest basic upload1 upload2 upload3 download refs rm make run worksheet worksheet_search worksheet_tags freeze detach perm search kill mimic status events batch copy
```
percyliang: Could we tag the various tests with 'docker' or 'network', so we can just do:

```
python test-cli.py all no-docker
```

percyliang:
@andreweduffy leaving out docker is fine for now, but add a TODO so we can include it later.

test-cli.py
a10y: When I run (both locally on Ubuntu 14.04 w/Python 2.7.5) and on Circle it says that it needs the u's. Not sure why it would output unicode strings now and not before

percyliang: Perhaps it's better to change the CLI code to output tags.join(' ') so that
the output looks better and we don't have to worry about this issue?

test-cli.py
a10y: It seems like uuid2 isn't being used at all, so I think I was actually going to go ahead and delete the assignment as well but forgot about it. Would you rather I leave it in?

percyliang: If it's not used just delete it

test-cli.py
a10y: Yep, that's pretty easy, I'll put that in a different commit then.

test-cli.py
```diff
@@ -420,7 +430,7 @@ def test(ctx):
     check_contains(['run "echo hello"'], run_command([cl, 'info', '-f', 'args', uuid]))
     check_equals('hello', run_command([cl, 'cat', uuid+'/stdout']))
     # block
-    uuid2 = check_contains('hello', run_command([cl, 'run', 'echo hello', '--tail'])).split('\n')[0]
+    check_contains('hello', run_command([cl, 'run', 'echo hello', '--tail']))
```
a10y, line 433: @percyliang removed uuid2 here

a10y:
@percyliang Is this clear to merge?

scripts/ci-tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Setup environment to run tests
+cd $(dirname $0)/..
+export PATH="$HOME/codalab-cli/codalab/bin:$PATH"
+
+
+# Run all tests save for resources and write, which depend on Docker.
+# TODO: Add support for docker tests
+./venv/bin/python test-cli.py unittest basic upload1 upload2 upload3 download refs rm make run worksheet worksheet_search worksheet_tags freeze detach perm search kill mimic status events batch copy
```
percyliang: use no-docker

test-cli.py
```diff
@@ -472,10 +482,10 @@ def test(ctx):
     ctx.collect_worksheet(wuuid)
     # Add tags
     run_command([cl, 'wedit', wname, '--tags', 'foo', 'bar', 'baz'])
-    check_contains(['Tags: \\[\'foo\', \'bar\', \'baz\'\\]'], run_command([cl, 'ls', '-w', wuuid]))
+    check_contains(['Tags: \\[u\'foo\', u\'bar\', u\'baz\'\\]'], run_command([cl, 'ls', '-w', wuuid]))
     # Modify tags
     run_command([cl, 'wedit', wname, '--tags', 'bar', 'foo'])
-    check_contains(['Tags: \\[\'bar\', \'foo\'\\]'], run_command([cl, 'ls', '-w', wuuid]))
+    check_contains(['Tags: \\[u\'bar\', u\'foo\'\\]'], run_command([cl, 'ls', '-w', wuuid]))
```
percyliang: These still need to be replaced with spaces: 'Tags: bar foo'

percyliang:
LGTM modulo comments

Issue #38, opened 2015-03-26
MySQL Error when installing CodaLab


ErickWa:
Hi Guys,

I was doing a fresh install of CodaLab into Ubuntu Server, and when running ./setup.sh, I received the following error message:

![image](https://cloud.githubusercontent.com/assets/4361560/6838518/f8838636-d317-11e4-8c67-253b43c1618d.png)

Conversely, I saw that in the release notes the setup.sh script was recently updated to install MySQL by default.  Has this change not been integrated somehow?

percyliang:
Ok, I removed MySQL from the initial install dependencies (#39), so it
should work now.

 -Percy

PR #319, opened 2016-01-28
Print usage for headless CLI

kashizui:
Fixes codalab/codalab-worksheets#38

@percyliang @klopyrev please review


percyliang:
LGTM

PR #323, opened 2016-01-31
Set up a Bottle server and add some example APIs.

klopyrev:
@kashizui @percyliang

### DIFF
--- a/codalab/bin/cl.py
+++ b/codalab/bin/cl.py
@@ -9,20 +9,22 @@ import subprocess
 from watchdog.observers import Observer
 from watchdog.events import FileSystemEventHandler
 
+from codalab.lib.bundle_cli import BundleCLI, Commands
 from codalab.lib.codalab_manager import CodaLabManager
 
 class ClFileWatcherEventHandler(FileSystemEventHandler):
     SERVER_PROCESS = None
 
-    def __init__(self):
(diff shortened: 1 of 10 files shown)


percyliang:
Overall looks good - thanks!  Just to confirm, the changes to cl.py aren't related to the REST API and don't affect the old functionality, right?

One issue I'd like to bring up is that many users who are not interested in running their own servers install the CLI, so we should allow their install process to be as breezy as possible.  bottle adds more dependencies.  Could we split requirements.txt into requirements-client.txt and requirements-server.txt and modify setup.sh to take an argument [client|server] to install the right dependencies?

klopyrev:
Yes, the changes to cl.py don't affect the old functionality.

I modified the setup script.

percyliang:
LGTM from me, but let's hear what @kashizui says

klopyrev:
@percyliang I made a change to what I think looks cleaner. I'm now saving the bundle model and bundle store in the local thread-local variable. Let me know what you think.

percyliang:
Looks good to me.

On Mon, Feb 1, 2016 at 3:43 PM, Konstantin Lopyrev <notifications@github.com

kashizui:
Looks great! Excited to get cracking on this.
I really like the look of Bottle.

PR #328, opened 2016-02-03
Calling cl in scripts will now use the correct session

ppasupat:
This patch modifies the method that identifies the CodaLab session based on the PID.

The process name when running a script (Python, Ruby, Bash, ...) is the filename and not `python`, `ruby`, etc., which confuses the method. Using the first argument of the command line instead of the process name solves the problem.


percyliang:
LGTM - thanks!

percyliang:
Actually, can you make the PR into the develop branch?

ppasupat:
It looks like the pull request target branch cannot be changed. Will make a new pull request.

PR #329, opened 2016-02-03
Stopped error email notification from Django

ChengshuLi:


percyliang:
1) We want to keep these emails so we know when bad stuff happens!  The change we want to make here is to make error_type more coarse grained by replacing uuids with 'uuid'.  That way, there won't be a flood of emails.

2) The Django emails are not being generated by monitor.py but by something in codalab-worksheets.  Those are the ones we want to stop.

percyliang:
Why is this closed?  Don't we want this change?

Issue #299, opened 2015-12-30
store bundles based on uuid rather than data_hash


percyliang:
Original motivation of storing by data_hash is that there might be two bundles with the exact data_hash, and we would save on that.  In practice, this basically never happens, and if it does happen, we should encourage people to use the same bundle!

Doing this would simplify the logic and mapping between logical bundles and physical storage, and allow us to reason about each bundle independently.

This issue should be done jointly with #296 

Requires bundle store migration.

PR #331, opened 2016-02-07
#299 Store bundles by UUID instead of Digest

a10y:
@percyliang 

I've tested functionality both locally and using `cl server`, and have tested the migration script by creating a bunch of bundles from the same data, changed them to be stored de-duped by their `data_hash`, and then run `./scripts/migrate-hash-uuid.py`.

Steps:
1. Create a bunch of bundles:
   
   ```
   cl up alembic.ini
   cl up alembic.ini
   cl up README.md
   cl up alembic.ini
   ```
2. Run `cl info` to find the `data_hash`, then do `mv ~/.codalab/data/$uuid ~/.codalab/data/$data_hash`. There should only end up being 2 entries in `~/.codalab/data`.
3. Run `./scripts/migrate-hash-uuid.py`. You should see that there are 4 things in `~/.codalab/data`, and the directory `~/.codalab/migration-temp` is still intact with the old tree structure in case you want to restore (script prints a message telling you how to do so).

### DIFF
--- a/codalab/lib/bundle_store.py
+++ b/codalab/lib/bundle_store.py
@@ -94,74 +94,81 @@ class BundleStore(object):
         '''
         to_delete = []
 
-        # Create temporary directory as a staging area and put everything there.
-        temp_path = tempfile.mkdtemp('-bundle_store_upload')
-        temp_subpaths = []
+        # If just a single file, set the final path to be equal to that file
+        single_file = len(sources) == 1
+
+        final_path = os.path.join(self.data, uuid)
+        if os.path.exists(final_path):
+            # Already exists, just delete it
+            print >> sys.stderr, 'Error, path %s already present in bundle store' % final_path
+            return (None, None)
+        # Only make if not there
+        elif not single_file:
+            path_util.make_directory(final_path)
+
+        # Paths to resources
+        subpaths = []
+
         for source in sources:
             # Where to save |source| to (might change this value if we unpack).
-            temp_subpath = os.path.join(temp_path, os.path.basename(source))
+            if not single_file:
+                subpath = os.path.join(final_path, os.path.basename(source))
+            else:
+                subpath = final_path
+
             if remove_sources:
                 to_delete.append(source)
             source_unpack = unpack and zip_util.path_is_archive(source)
 
+            if source_unpack:
+                # Load the file into the bundle store under the given path
+                subpath += zip_util.get_archive_ext(source)
+
             if path_util.path_is_url(source):
                 # Download the URL.
-                print_util.open_line('BundleStore.upload: downloading %s to %s' % (source, temp_path))
+                print_util.open_line('BundleStore.upload: downloading %s to %s' % (source, subpath))
                 if git:
-                    file_util.git_clone(source, temp_subpath)
+                    file_util.git_clone(source, subpath)
                 else:
-                    file_util.download_url(source, temp_subpath, print_status=True)
+                    file_util.download_url(source, subpath, print_status=True)
                     if source_unpack:
-                        zip_util.unpack(temp_subpath, zip_util.strip_archive_ext(temp_subpath))
-                        path_util.remove(temp_subpath)
-                        temp_subpath = zip_util.strip_archive_ext(temp_subpath)
+                        zip_util.unpack(subpath, zip_util.strip_archive_ext(subpath))
+                        path_util.remove(subpath)
+                        subpath = zip_util.strip_archive_ext(subpath)
                 print_util.clear_line()
             else:
                 # Copy the local path.
                 source_path = path_util.normalize(source)
                 path_util.check_isvalid(source_path, 'upload')
 
-                # Recursively copy the directory into a new BundleStore temp directory.
-                print_util.open_line('BundleStore.upload: %s => %s' % (source_path, temp_subpath))
+                # Recursively copy the directory into the BundleStore
+                print_util.open_line('BundleStore.upload: %s => %s' % (source_path, subpath))
                 if source_unpack:
-                    zip_util.unpack(source_path, zip_util.strip_archive_ext(temp_subpath))
-                    temp_subpath = zip_util.strip_archive_ext(temp_subpath)
+                    zip_util.unpack(source_path, zip_util.strip_archive_ext(subpath))
+                    subpath = zip_util.strip_archive_ext(subpath)
                 else:
                     if remove_sources:
-                        path_util.rename(source_path, temp_subpath)
+                        path_util.rename(source_path, subpath)
                     else:
-                        path_util.copy(source_path, temp_subpath, follow_symlinks=follow_symlinks, exclude_patterns=exclude_patterns)
+                        path_util.copy(source_path, subpath, follow_symlinks=follow_symlinks, exclude_patterns=exclude_patterns)
                 print_util.clear_line()
 
-            temp_subpaths.append(temp_subpath)
+            subpaths.append(subpath)
 
-        # If exactly one source, then upload that directly.
-        if len(temp_subpaths) == 1:
-            to_delete.append(temp_path)
-            temp_path = temp_subpaths[0]
-
-        # Multiplex between uploading a directory and uploading a file here.
-        # All other path_util calls will use these lists of directories and files.
-        if os.path.isdir(temp_path):
-            dirs_and_files = path_util.recursive_ls(temp_path)
+        dirs_and_files = None
+        if os.path.isdir(final_path):
+            dirs_and_files = path_util.recursive_ls(final_path)
         else:
-            dirs_and_files = ([], [temp_path])
+            dirs_and_files = [], [final_path]
 
-        # Hash the contents of the temporary directory, and then if there is no
-        # data with this hash value, move this directory into the data directory.
-        print_util.open_line('BundleStore.upload: hashing %s' % temp_path)
-        data_hash = '0x%s' % (path_util.hash_directory(temp_path, dirs_and_files),)
+        # Hash the contents of the bundle directory. Update the data_hash attribute
+        # for the bundle
+        print_util.open_line('BundleStore.upload: hashing %s' % final_path)
+        data_hash = '0x%s' % (path_util.hash_directory(final_path, dirs_and_files))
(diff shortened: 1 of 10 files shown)


codalab/lib/canonicalize.py
```diff
@@ -115,12 +115,7 @@ def get_target_path(bundle_store, model, target):
     Return the on-disk location of the target (bundle_uuid, subpath) pair.
     """
     (uuid, path) = target
-    bundle = model.get_bundle(uuid)
-    if not bundle.data_hash:
-        # Note that the bundle might not be ready, but return the location anyway to the temporary directory.
-        bundle_root = get_current_location(bundle_store, uuid)
```
percyliang, line 121: Currently, while the bundle is running, data_hash = null, and we access the temp directory for the bundle.  This logic is still necessary, I think.

But actually, we could just not use the temporary directory at all and do everything directly in the data directory, including uploads (in which case the status should be 'uploading', analogous to 'running').  This would require a few more changes in other parts of the code.  I think it should be fine.

a10y, line 121: @percyliang Are you suggesting doing that?

percyliang, line 121: Thinking about this some more:

When the job is being run, there are other files that created too in temp: <uuid>.status, etc.  Do we want these hanging out in the bundle store?  The logic is certainly simpler that way, and when you rm a bundle, it could delete <uuid>.\* rather than currently having a slew of temp files just hanging out there.  (Some of these shouldn't actually be deleted because their states are still 'running' but their workers died, so their data is stuck in temp.)

On the other hand, the bundle store would be constantly changing.  I think this is okay though, because I think the contents of a bundle (being run or being uploaded) should be accessible as is, and we should just use the data_hash and state flag to denote whether a bundle's contents have actually finished changing and become immutable.  We can even do fancy things like partial uploads or depend on a running bundle later.

So in short, let's just use the data directory, not the temp directory everywhere (need to change this in upload logic too).  In general, copying things to a temp directory first and then moving atomically is good (how browser downloads are handled), but this is because there's no natural bit that tells you whether a file is complete, but here we rely on the DB (and should make sure that logic is sound).

Do you see any problems with this?

a10y, line 121: That sounds pretty good, I'll look into that and see if I run into any problems

a10y:
@percyliang 

This should be ready for review. It includes removal of use of the temp directory for uploading and RunBundles

codalab/objects/work_manager.py
```diff
@@ -292,15 +292,8 @@ def update_running_bundle(self, status):
                     print >>sys.stderr, 'Worker.finalize_bundle: installing (copying) dependencies to %s (MakeBundle)' % temp_dir
                     bundle.install_dependencies(self.bundle_store, self.get_parent_dict(bundle), temp_dir, copy=True)
 
-                # Note: uploading will move temp_dir to the bundle store.
-                (data_hash, bundle_store_metadata) = self.bundle_store.upload(sources=[temp_dir],
```
percyliang, line 296: Just to confirm: temp_dir is now the actual bundle directory, right?  Probably would be good to note that.

codalab/lib/bundle_store.py
```diff
@@ -94,74 +81,79 @@ def upload(self, sources, follow_symlinks, exclude_patterns, git, unpack, remove
         '''
         to_delete = []
 
-        # Create temporary directory as a staging area and put everything there.
-        temp_path = tempfile.mkdtemp('-bundle_store_upload')
-        temp_subpaths = []
+        # If just a single file, set the final path to be equal to that file
+        single_file = len(sources) == 1
```
percyliang: Replace with `single_path` (doesn't have to be file, could be directory)

codalab/lib/bundle_store.py
```diff
@@ -94,74 +81,79 @@ def upload(self, sources, follow_symlinks, exclude_patterns, git, unpack, remove
...
+
+        final_path = os.path.join(self.data, uuid)
+        if os.path.exists(final_path):
+            raise UsageError('Error, path %s already present in bundle store' % final_path)
```
percyliang: Don't need 'Error, '

codalab/objects/work_manager.py
a10y, line 296: Yep. I'll go and change the name

codalab/lib/bundle_store.py
```diff
@@ -94,74 +81,79 @@ def upload(self, sources, follow_symlinks, exclude_patterns, git, unpack, remove
...
         for source in sources:
             # Where to save |source| to (might change this value if we unpack).
-            temp_subpath = os.path.join(temp_path, os.path.basename(source))
+            if not single_file:
+                subpath = os.path.join(final_path, os.path.basename(source))
+            else:
+                subpath = final_path
+
             if remove_sources:
                 to_delete.append(source)
```
percyliang, line 105: I don't think we need `to_delete` anymore.  This was when we were first uploading to a temporary directory and then moving it to the bundle store (see remote_bundle_client's upload method).  But now, uploading should just go directly to the bundle store.  This might require testing the upload / download / add bundle functionality carefully.

codalab/objects/work_manager.py
a10y, line 296: Should we rename temp_dir in the metadata as well or just leave it?

codalab/lib/bundle_store.py
a10y, line 105: I think we'd need it for `remove_sources`, unless removing a source means just renaming it into the bundle store

a10y, line 105: Actually that's not quite as safe because if upload fails for some reason, we probably don't want the source to get deleted.

a10y, line 105: So in essence, I'm saying that we should leave in `to_delete` so that we can batch delete all of the sources at the end of the upload

percyliang, line 105: `to_delete` was only to delete the source in the `temp_dir`, never an original source.  The way it worked before was when you use RemoteBundleClient to upload: you'd copy everything first to a temp_dir, and then call LocalBundelClient.upload_bundle, which would "upload" it to the bundle store with the remove_sources=True, which really meant move it to the `data` directory.

But now, we should just upload directly into the data directory, since we have the uuid (yes!).  This requires modifying RemoteBundleClient.upload_bundle to not use open_temp_file, but to add a command to bundle_rpc_server called open_bundle_for_writing, which (after checking permissions, etc.) allows us to upload directly to the bundle store.

Now that I think about it, it might be good to defer this to another PR since it's going to be somewhat involved...

codalab/objects/work_manager.py
percyliang, line 296: Leave it for now.  In a sense, we don't need temp_dir anymore.  Plus, Konstantin's changes will make obviate this stuff.

codalab/lib/bundle_store.py
a10y, line 105: Okay, so I'll save that for another PR and leave it out of this one.

percyliang:
This is a big change because it breaks backward compatibility.  Let's use a new directory 'bundles' rather than 'data', so that if someone does a pull and tries to run the new code, it will give them an error message (since 'bundles' doesn't exist) and ask them to run the migration script.  This seems safer than just not being able to find your bundles all of the sudden in 'data'.

Also, the migration script currently copies bundles, which is going to take a long time...(~10 hours).  I think it's better to use 'mv'.  But let's have a dry-run mode (the default), which just prints out the list of all command that will be executed (and make sure all the files exist, etc.) before executing it.

a10y:
@percyliang I agree with the first comment, however I believe we need to be making copies because we're now taking a single data_hash that potentially maps to multiple bundles and we're forking it off into multiple storage units under separate subdirectories.

percyliang:
Fair point.  Almost all bundles have distinct data_hashes...so let's do the
copy if there are more than one but move if there's exactly one.

codalab/objects/work_manager.py
```diff
@@ -292,15 +292,8 @@ def update_running_bundle(self, status):
...
-                                                                              follow_symlinks=False,
-                                                                              exclude_patterns=None,
-                                                                              git=False,
-                                                                              unpack=False,
-                                                                              remove_sources=True)
-                db_update['data_hash'] = data_hash
-                metadata.update(bundle_store_metadata)
+                db_update['data_hash'] = path_util.hash_path(temp_dir)
```
a10y, line 295: @percyliang This was changed to support MakeBundles (which can create files) and RunBundles (which always create directories). You can now hash them both and not care what you're hashing.

a10y:
@percyliang That should be it for this PR.

scripts/migrate-hash-uuid.py
```diff
@@ -0,0 +1,71 @@
...
+    bundles_with_hash = model.batch_get_bundles(data_hash=data_hash)
+    # We'd prefer renaming bundles to making copies, but because we are converting from deduplicated storage
+    # we need to make sure that we only perform renames if we map 1:1 UUID->Hash.
+    rename_allowed = len(bundles_with_hash) <= 1
+    for bundle in bundles_with_hash:
+        uuid = bundle.uuid
+        copy_location = os.path.join(FINAL_LOCATION, uuid)
+
+        if rename_allowed:
+            print >> sys.stderr, 'Moving Bundle 0x%s with data_hash 0x%s to %s' % (uuid, data_hash, copy_location)
```
percyliang: Can we print out the actual mv or cp -a command (in a way that can be piped to the shell if needed)?
And let's just execute those commands directly.  There should be as tight of a connection between the dry run and the actual run as possible.

percyliang:
Note: there might be items in temp that haven't gotten moved over (should be fixed by #341).
When we deploy, we need to mv the things from temp into the bundle store.

scripts/migrate-hash-uuid.py
a10y: @percyliang That makes sense, see latest commit

percyliang:
Remove the comment "Migration complete! ...has been left with your old data" because that ain't true anymore.  You might as well delete the 'data' directory if migration succeeded.

Otherwise, LGTM!

PR #333, opened 2016-02-09
Fix Worksheets#85, Dashboard Link

ChengshuLi:
@percyliang Updated my changes. Feel free to have a look.

### DIFF
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -168,8 +168,9 @@ class LocalBundleClient(BundleClient):
 
     def get_worksheet_uuid(self, base_worksheet_uuid, worksheet_spec):
         if worksheet_spec == '' or worksheet_spec == worksheet_util.DASHBOARD:         
-            worksheet_spec = spec_util.dashboard()
-            return self.new_worksheet(worksheet_spec, True)
+            # create the home page for the current user before returning the dashboard
+            self.new_worksheet(spec_util.home_worksheet(self._current_user_name()), True)
+            return self.new_worksheet(spec_util.dashboard(), True)
         else:
             return canonicalize.get_worksheet_uuid(self.model, base_worksheet_uuid, worksheet_spec)
 
@@ -769,8 +770,8 @@ class LocalBundleClient(BundleClient):
         # If trying to set the name to a home worksheet, then it better be
         # user's home worksheet.
         username = self._current_user_name()
-        if spec_util.is_dashboard(name) and spec_util.dashboard() != name:
-            raise UsageError('Cannot create %s because this is the Codalab Dashboard' % name)
+        if spec_util.is_home_worksheet(name) and spec_util.home_worksheet(username) != name:
+            raise UsageError('Cannot create %s because this is potentially the home worksheet of another user' % name)
         try:
             self.get_worksheet_uuid(None, name)
             exists = True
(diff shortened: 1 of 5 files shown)


codalab/lib/worksheet_util.py
```diff
@@ -40,6 +40,7 @@
 
 # Special characters to point to worksheets
 HOME_WORKSHEET = '/'
+DASHBOARD = 'Dashboard'
```
percyliang: lowercase

codalab/lib/spec_util.py
```diff
@@ -89,10 +89,14 @@ def create_default_name(bundle_type, raw_material):
...
-
 def home_worksheet(username):
     return 'home-' + username
 
-
 def is_home_worksheet(name):
     return name.startswith('home-')
+
+def dashboard():
+    return 'dashboard'
```
percyliang: use DASHBOARD variable

codalab/client/local_bundle_client.py
```diff
@@ -804,15 +803,23 @@ def new_worksheet(self, name, ensure_exists):
...
 
         # Make worksheet publicly readable by default
         self.set_worksheet_perm(worksheet.uuid, self.model.public_group_uuid, 'read')
-
+        if spec_util.is_dashboard(name):
+            self.populate_dashboard(worksheet)
         return worksheet.uuid
 
+    def populate_dashboard(self, worksheet):
+        lines = [line.rstrip() for line in open('../objects/dashboard.ws').readlines()]
```
percyliang: I don't think this will access the right file in general.  'cl' could be run from anywhere...

One compromise is to have a single python file dashboard_lines.py which contains 
DASHBOARD_LINES = """
...
"""
and just load that.

codalab/client/local_bundle_client.py
```diff
@@ -167,10 +167,10 @@ def get_bundle_target(self, target):
         return (self.model.get_bundle(bundle_uuid), subpath)
 
     def get_worksheet_uuid(self, base_worksheet_uuid, worksheet_spec):
-        if worksheet_spec == '' or worksheet_spec == worksheet_util.HOME_WORKSHEET:
-            # Default worksheet name: take the username.
-            worksheet_spec = spec_util.home_worksheet(self._current_user_name())
-            return self.new_worksheet(worksheet_spec, True)
+        if worksheet_spec == '' or worksheet_spec == worksheet_util.DASHBOARD:
```
percyliang: This should still be HOME_WORKSHEET....logic shouldn't change in this function.

ChengshuLi: Just wanna clarify. Do we want our landing page to be the dashboard or the home worksheet? I think I personally prefer the former.

percyliang: The landing page on the website is the dashboard, but that's taken care of
already when you have name=dashboard.

On the command-line, when you do cl work main::, it should take you to your
home directory in analog with what happens on the shell.

ChengshuLi: I see. I will clean up the logic here

ChengshuLi:
@percyliang

ChengshuLi:
@percyliang    Fixed what you said in your comments :)

codalab/client/local_bundle_client.py
```diff
@@ -168,9 +168,15 @@ def get_bundle_target(self, target):
 
     def get_worksheet_uuid(self, base_worksheet_uuid, worksheet_spec):
         if worksheet_spec == '' or worksheet_spec == worksheet_util.HOME_WORKSHEET:
-            # Default worksheet name: take the username.
             worksheet_spec = spec_util.home_worksheet(self._current_user_name())
             return self.new_worksheet(worksheet_spec, True)
+        elif spec_util.is_dashboard(worksheet_spec):
+            home_worksheet_spec = spec_util.home_worksheet(self._current_user_name())
```
percyliang: Why create home worksheet when people ask for dashboard?  I don't think should have it.

codalab/client/local_bundle_client.py
```diff
@@ -796,7 +801,8 @@ def new_worksheet(self, name, ensure_exists):
             except UsageError:
                 pass
 
-        self.ensure_unused_worksheet_name(name)
+        if not spec_util.is_dashboard(name):
```
percyliang: Why is dashboard exempt?

if someone does 'cl new dashbaord' we don't want to create one if it exists.

codalab/client/local_bundle_client.py
```diff
@@ -804,15 +810,24 @@ def new_worksheet(self, name, ensure_exists):
             'title': None,
             'frozen': None,
             'items': [],
-            'owner_id': self._current_user_id()
+            'owner_id': '0' if spec_util.is_dashboard(name) else self._current_user_id()
```
percyliang: Don't use 0, use root_user_id

codalab/client/local_bundle_client.py
```diff
@@ -928,7 +943,8 @@ def update_worksheet_items(self, worksheet_info, new_items):
         last_item_id = worksheet_info['last_item_id']
         length = len(worksheet_info['items'])
         worksheet = self.model.get_worksheet(worksheet_uuid, fetch_items=False)
-        check_worksheet_has_all_permission(self.model, self._current_user(), worksheet)
+        if not spec_util.is_dashboard(worksheet_info['name']):
```
percyliang: Why all these checks?

We don't want any user modifying the dashboard...

percyliang:
thanks.  Some more questions/comments @ChengshuLi

codalab/client/local_bundle_client.py
ChengshuLi: Yeah this is definitely not ideal. I put in these checks because the first user who visit the dashboard needs to create it as a root user, but he doesn't have the permission to populate the dashboard after creating it. So shall we make the creation of the dashboard part of the configuration process, not specific to any user? I think that might make more sense. If so, could you point me to some files that I should look at to achieve that?  @percyliang

percyliang: I think we should just create the dashboard under the first user who
accesses it, not root.  It's up to the admin of the local instance to visit
the dashboard first.

ChengshuLi: Is there a reason we wanna do this rather than making it part of the configuration process? I felt like counting on the admin being the first to visit the dashboard is not super robust.

percyliang: There's no good place to put this in the configuration process right now.
 ./setup.sh runs but doesn't assume there's a DB yet.  This will be
improved once we have OAuth on the bundle service.

In any case, let's have dashboard be created on the fly still.  The
configuration will then be just 'cl work dashboard'.  Also, if someone
upgrading forgets to run the config script, then things won't break with
errors, which is always good.

ChengshuLi:
@percyliang .

Fixed your comments. But I am struggling to find the best place to create the home sheet for a new user. Do you have any advice?

codalab/client/local_bundle_client.py
```diff
@@ -167,12 +167,16 @@ def get_bundle_target(self, target):
         return (self.model.get_bundle(bundle_uuid), subpath)
 
     def get_worksheet_uuid(self, base_worksheet_uuid, worksheet_spec):
-        if worksheet_spec == '' or worksheet_spec == worksheet_util.HOME_WORKSHEET:
```
percyliang, line 170: Just leave this logic there as before.

codalab/client/local_bundle_client.py
```diff
@@ -772,9 +776,8 @@ def ensure_unused_worksheet_name(self, name):
         username = self._current_user_name()
         if spec_util.is_home_worksheet(name) and spec_util.home_worksheet(username) != name:
             raise UsageError('Cannot create %s because this is potentially the home worksheet of another user' % name)
-
         try:
-            self.get_worksheet_uuid(None, name)
+            self.ensure_worksheet_exist(None, name)
```
percyliang: Don't do this - the purpose of this function is to ensure the worksheet doesn't exists, not create one.

codalab/lib/spec_util.py
```diff
@@ -9,6 +9,7 @@
   UsageError,
 )
 
+DASHBOARD = 'dashboard'
```
percyliang: rename DASHBOARD_WORKSHEET

codalab/client/local_bundle_client.py
ChengshuLi, line 170: The previous logic won't work. It will lead to infinite recursion when it comes to dashboard. So my current the logic is the following:

Try to get worksheet uuid and return it.
catch error (it means the worksheet does not exist):
if the requested worksheet is home_worksheet or dashboard, create it
otherwise, just raise that error. 

We can also talk in person :)

codalab/client/local_bundle_client.py
ChengshuLi: That's actually my intention. Sorry for the bad naming. 

ensure_worksheet_exist will throw an error if the worksheet doesn't exist 
It won't create a new one.

Maybe I should rename it to something else?

codalab/client/local_bundle_client.py
percyliang, line 170: You still want to intercept '' and HOME_WORKSHEET and handle them specially as before - these should not get passed into canonicalize.get_worksheet_uuid.

Then the dashboard creation can be in the UsageError as you have it.

codalab/client/local_bundle_client.py
percyliang: Still don't know why this needs to be changed.

ChengshuLi:
Thanks for the comments. Made some changes and feel free have a look. @percyliang 

Still not sure where is the best place to create home-worksheet for a new user though.

codalab/client/local_bundle_client.py
```diff
@@ -167,12 +167,30 @@ def get_bundle_target(self, target):
...
     def get_worksheet_uuid(self, base_worksheet_uuid, worksheet_spec):
+        """
+        Return the uuid of the specified worksheet if it exists.
+        If not, create a new worksheet if the specified worksheet is home_worksheet or dashboard. Otherwise, throw an error.
+        """
         if worksheet_spec == '' or worksheet_spec == worksheet_util.HOME_WORKSHEET:
-            # Default worksheet name: take the username.
-            worksheet_spec = spec_util.home_worksheet(self._current_user_name())
-            return self.new_worksheet(worksheet_spec, True)
+                worksheet_spec = spec_util.home_worksheet(self._current_user_name())
```
percyliang: remove extra indent

percyliang:
LGTM

PR #334, opened 2016-02-13
Make tag output comma-separated in CLI

a10y:
@percyliang

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1237,7 +1237,7 @@ class BundleCLI(object):
         fields = [
             ('Worksheet', self.worksheet_str(worksheet_info)),
             ('Title', formatting.verbose_contents_str(worksheet_info['title'])),
-            ('Tags', worksheet_info['tags']),
+            ('Tags', ' '.join(worksheet_info['tags'])),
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1237,7 +1237,7 @@ def _worksheet_description(self, worksheet_info):
         fields = [
             ('Worksheet', self.worksheet_str(worksheet_info)),
             ('Title', formatting.verbose_contents_str(worksheet_info['title'])),
-            ('Tags', worksheet_info['tags']),
+            ('Tags', ', '.join(worksheet_info['tags'])),
```
percyliang: Minor: I think joining on a single space would be nicer.

a10y: Oh okay, is there no idea of multi-word tags in codalab?

percyliang: Yes (although we don't enforce it, but we should - add an issue?)  I think
it's pretty normal to have tags without spaces though.

a10y: Are you saying that keeping tags to single-words is not enforced? I can issue that

percyliang: Right

a10y: #336 Just registered

percyliang:
LGTM

Issue #339, opened 2016-02-16
timing out bundles prints 00:00:00 and doesn't clean up


percyliang:
1) Minor: in work_manager.py,

```
failure_msg = 'No response from worker in %s' % time.strftime('%H:%M:%S', time.gmtime(update_timeout))
```

Should use formatting.duration_str for consistency.

2) Major: setting the status of a bundle to FAILED doesn't clean up temp files, do proper accounting, etc.  We need to call update_running_bundle to use the same code path as if a run were actually finishing.

PR #341, opened 2016-02-16
 #339: timing out bundles prints 00:00:00 and doesn't clean up

a10y:
@percyliang The only thing that doesn't get cleaned up is the <uuid>.sh script, should that be getting deleted or is that collected by something else?

### DIFF
--- a/codalab/objects/work_manager.py
+++ b/codalab/objects/work_manager.py
@@ -23,6 +23,7 @@ from codalab.common import (
 )
 from codalab.lib import (
   canonicalize,
+  formatting,
   path_util,
 )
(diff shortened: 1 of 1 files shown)


percyliang:
.sh should be deleted (hm, strange that the .status, .cid get deleted though) - this should all be done in finalize_bundle.

a10y:
The reason it's happening for LocalMachine is because when you kill the work-manager, `self.temp_dir` and `self.bundle` are erased from the machine. I'm not sure if there's a way to fix this for LocalMachine, but it's going to be gone anyway with the new BundleService correct? I'm working on fixing up RemoteMachine.

PR #350, opened 2016-02-22
Changes to REST server setup.

klopyrev:
A few changes to the REST server setup, including:

1) Monkey patching before importing Bottle. This ensures that the Bottle thread locals work correctly.
2) Calling codalab_manager.model() and thus creating the SQL engine only on the first request. This is a hack to work around the requirement that any SQL engine objects are creating after forking (here the forking happening in the Gunicorn code)
3) Working around some logic in Bottle that returns a complex HTML error page on errors. With the new solution the error page simply contains the error text. This ensures that any code calling the REST API can display errors the way it wants.
4) Rename workers to processes for clarity.
5) Work around a bug in Gunicorn that tries to read all command line arguments, even when it's called as a library.

@percyliang @kashizui

### DIFF
--- a/codalab/bin/cl.py
+++ b/codalab/bin/cl.py
@@ -78,10 +78,10 @@ def do_server_command(bundle_cli, args):
             '--watch', help='Restart the server on code changes.',
             action='store_true'),
         Commands.Argument(
-            '-w', '--workers',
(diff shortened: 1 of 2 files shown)


percyliang:
LGTM

PR #351, opened 2016-02-22
Fix help in the web UI.

klopyrev:
This fix doesn't show help for the server and rest-server commands in the web UI, which is the correct behavior since you can't run a server from the web UI.

Fixes https://github.com/codalab/codalab-worksheets/issues/103

@percyliang


percyliang:
LGTM

Issue #345, opened 2016-02-21
cl upload on multiple files adds ".tar.gz" to each uploaded file.


ppasupat:
To reproduce, run

```
cl up -L mydir1 mydir2 myfile
```

where `mydir1`, `mydir2` and `myfile` are existing directories and files.

The directories and files in the top level of the uploaded bundle will have `.tar.gz` affixed to the end.

PR #354, opened 2016-02-24
 #345: cl upload on multiple files adds ".tar.gz" to each uploaded file.

a10y:
@percyliang 

Fix for #345 

The reason for the bug was because bundles that are just single files need to be named `CODALAB_HOME/bundles/UUID.extension`, so I was simply adding `.extension` to the end of `subpath`. However, I forgot subpath is the name of the sourcefile for non-single file uploads, so we were uploading `thing.zip` and it was being placed in `CODALAB_HOME/bundles/UUID/thing.zip.zip` and unpacking to `CODALAB_HOME/bundles/UUID/thing.zip`. This ensures that we only append the archive extension if we're uploading a single file.

### DIFF
--- a/codalab/lib/bundle_store.py
+++ b/codalab/lib/bundle_store.py
@@ -105,7 +105,7 @@ class BundleStore(object):
                 to_delete.append(source)
             source_unpack = unpack and zip_util.path_is_archive(source)
 
-            if source_unpack:
+            if source_unpack and single_path:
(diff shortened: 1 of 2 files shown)


percyliang:
LGTM

Shouldn't test-cli.py have caught this?

a10y:
Hmm looks like there aren't any tests that target uploading multiple archives in one command, let me go ahead and add one before merging.

PR #355, opened 2016-02-25
Add --allow-failed-dependencies flag.

klopyrev:
@percyliang


codalab/bundles/runnable_bundle.py
```diff
@@ -0,0 +1,40 @@
+'''
+RunnableBundle is an abstract Bundle supertype for bundles that need to be
```
percyliang: I like the refactoring, but this name is confusing (that runnable includes make and run).  Let's call it DerivedBundle.

percyliang:
LGTM modulo renaming

codalab/bundles/runnable_bundle.py
klopyrev: Done.

PR #356, opened 2016-02-26
Initialize root logger

### LINKED ISSUE #180: Instrument bundle service with better logging
Set up a logger through the python `logging` module, and start putting in helpful log statements at the appropriate log levels.

Design decisions:
- Where should log messages go (e.g. when using the `cl` command)?
- Where should logger configuration go? (assuming `~/.codalab/config.json`)
  etc.

kashizui:
Set up logging.

After this, instead of print statements, we should add

```
import logging
log = logging.getLogger(__name__)
```

at the top of each module, and use the standard log calls thereafter:

```
log.info("just info")
log.debug("verbose info")
log.warning("probably ok")
log.error("uh oh")
```

Logs will also be collected at the default location `$CODALAB_HOME/codalab.log`; this path, as well as log levels for both these recorded logs and the console logs can be configured.

@percyliang @klopyrev @andreweduffy @ChengshuLi Does anyone have any feedback on best practices here?

Starts to fix #180

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -29,6 +29,7 @@ include any server configuration.
 '''
 import getpass
 import json
+import logging
 import os
 import sys
 import time
@@ -131,6 +132,9 @@ class CodaLabManager(object):
             return x
         self.config = replace(self.config)
 
+        # Initialize logging
+        self.init_logger()
+
         # Read state file, creating if it doesn't exist.
         if not os.path.exists(self.state_path):
             write_pretty_json({
@@ -142,10 +146,10 @@ class CodaLabManager(object):
         self.clients = {}  # map from address => client
 
     def init_config(self, dry_run=False):
-        '''
+        """
         Initialize configurations.
         TODO: create nice separate abstraction for building/modifying config
-        '''
+        """
         print_block(r"""
            ____          _       _            _
          / ____|___   __| | __ _| |     T T  | |__
@@ -184,6 +188,11 @@ class CodaLabManager(object):
                     'verbose': 1,
                     'dispatch_command': "python $CODALAB_CLI/scripts/dispatch-q.py",
                 }
+            },
+            'logging': {
+                'file_path': "{0.codalab_home}/codalab.log",
+                'console_level': "INFO",
+                'file_level': "DEBUG",
             }
         }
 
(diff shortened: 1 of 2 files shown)


a10y:
One thing I've heard a lot is that it's best to make the logs structured (e.g. JSON) and then have a separate CLI for pretty-printing for human use. There's a Node library called [Bunyan that exemplifies this](https://github.com/trentm/node-bunyan). Basically you make sure that your logs are stored at rest in some structured format so that anyone can write tools to parse the logs. Bunyan ships a default CLI with the library that does all of the pretty things.

It's helpful for building monitoring tools.

One article: http://kartar.net/2015/12/structured-logging/

An example Python library that does this http://www.structlog.org/en/stable/

We could build our own, it wouldn't be too much lift to build one that does some simple JSON logging, and since we're starting fresh we might as well try and do best practices :)

codalab/lib/codalab_manager.py
```diff
@@ -232,6 +243,34 @@ def init_config(self, dry_run=False):
...
+        Initialize and configure the root logger.
+        Every module should get its own logger instance:
+            import logging
+            log = logging.getLogger(__name__)
+        """
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
+
+        # create file handler which logs even debug messages
+        log_file = self.config.get('logging', {}).get('file_path', os.path.join(self.codalab_home, 'codalab.log'))
```
klopyrev, line 257: Keep lines to 80 characters.

codalab/lib/codalab_manager.py
```diff
@@ -184,6 +190,11 @@ def init_config(self, dry_run=False):
                     'verbose': 1,
                     'dispatch_command': "python $CODALAB_CLI/scripts/dispatch-q.py",
                 }
+            },
+            'logging': {
+                'file_path': "$CODALAB_HOME/codalab.log",
+                'console_level': "INFO",
+                'file_level': "DEBUG",
```
klopyrev, line 197: DEBUG would probably create too much crap in the log file. DEBUG is meant for debugging, not production use. I would default to INFO.

codalab/lib/codalab_manager.py
percyliang, line 257: Being religious about the 80 character policy can sometimes make code even harder to read, I don't think we need to adhere to that.

That being said, one could definitely split this up by defining a default_file_path variable if one wanted.

codalab/lib/codalab_manager.py
```diff
@@ -232,6 +243,34 @@ def init_config(self, dry_run=False):
...
+        """
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
+
+        # create file handler which logs even debug messages
+        log_file = self.config.get('logging', {}).get('file_path', os.path.join(self.codalab_home, 'codalab.log'))
+        fh = logging.FileHandler(log_file, 'a')
+        fh.setLevel(self.config.get('logging', {}).get('file_level', 'DEBUG').upper())
+
+        # create console handler with a potentially different log level
```
percyliang, line 261: Capitalize the first word of comments.

codalab/lib/codalab_manager.py
```diff
@@ -232,6 +243,34 @@ def init_config(self, dry_run=False):
...
 
+    def init_logger(self):
+        """
+        Initialize and configure the root logger.
+        Every module should get its own logger instance:
+            import logging
+            log = logging.getLogger(__name__)
+        """
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
```
percyliang, line 254: why set to DEBUG here?  add comment?

percyliang:
LGTM

Note that we also log activity to the DB, so we can query that easily.  So keeping the file logging simple is good for now.

kashizui:
note: use `logging.config.dictConfig` to standardize logging config.
https://docs.python.org/2/library/logging.config.html

Issue #346, opened 2016-02-21
BundleStore Sanity Check Script


a10y:
Add a script under `scripts/sanity-check_bundlestore.py` that:
- Checks to see if there are any bundles on disk that are not in the database. These need to be relocated to some sort of trash directory
- For each bundle, make sure its input dependencies don't show up as directories inside of the bundle.  (Currently, this should be done only for `ready` or `failed` bundles.)
- No files aside from ones matching the UUID pattern should exist in the `bundles/` directory.
- For bundles that have a `UUID.cid` file or `UUID.status/` directory should not have state `ready` or `failed`.

a10y:
Since we're going to make this a command on the CLI, it might be better to add this as a method inside of MultiDiskBundleStore, and then hook this into bundle_cli

percyliang:
Yes.

PR #360, opened 2016-03-01
Bundle Health / Sanity Check

a10y:
@percyliang I also went ahead and added a rule that will delete the `.sh` files if its corresponding bundle is marked ready or failed. Everything else is as specified in #346

### DIFF
--- a/codalab/lib/bundle_store.py
+++ b/codalab/lib/bundle_store.py
@@ -1,11 +1,12 @@
 import os
-import time
+import re
 import sys
+import time
 
 from .hash_ring import HashRing
 
 from codalab.lib import path_util, file_util, print_util, zip_util
-from codalab.common import UsageError
+from codalab.common import UsageError, State
 
 def require_partitions(f):
     """Decorator added to MulitDiskBundleStore methods that require a disk to
@@ -35,6 +36,17 @@ class BundleStoreCleanupMixin(object):
         Cleanup a given bundle. If dry_run is True, do not actually
         delete the bundle from storage.
         """
+        pass
+
+class BundleStoreHealthCheckMixin(object):
+    """
+    This mixin defines functionality on a BundleStore that supports some sort of health-check mechanism.
+
+    Health check is an intentionally broad term that leaves its definition up to the interpretation of each
+    BundleStore. Note that this method IS allowed to perform operations destructive to objects stored in the bundle
+    store, i.e. this is not an idempotent operation, and calling this method should be done with care.
+    """
+    def health_check(self):
         pass
 
 class BaseBundleStore(object):
@@ -75,7 +87,7 @@ class BaseBundleStore(object):
         pass
 
 
-class MultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin):
+class MultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin, BundleStoreHealthCheckMixin):
     """
     A MultiDiskBundleStore is responsible for taking a set of locations and load-balancing the placement of
     bundle data between the locations. It accomplishes this goal using a consistent hash ring, a technique
@@ -382,4 +394,77 @@ class MultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin):
         print >>sys.stderr, "cleanup: data %s" % absolute_path
         if not dry_run:
             path_util.remove(absolute_path)
+
+
+    def health_check(self, model):
+        """
+        MultiDiskBundleStore.health_check(): In the MultiDiskBundleStore, bundle contents are stored on disk, and
+        occasionally the disk gets out of sync with the database, in which case we make repairs in the following ways:
+
+            1. Creates a trash directory
+            2. Moves the contents of bundles with corresponding UUID not in the database to trash
+            3. Any files that don't begin with a UUID string should be moved to trash
+            4. For each bundle marked READY or FAILED, ensure that its dependencies are not located in the bundle
+               directory.
+            5. Bundles that are stored in <UUID>.cid or <UUID>.status/ should be marked as READY or FAILED
+            6. <UUID>.sh and <UUID>-internal.sh should be gone for bundles marked READY or FAILED
+        """
+        # Create a trash directory
+        trash_dir = os.path.join(self.codalab_home, 'trash')
+        path_util.make_directory(trash_dir)
+
+        def _trash_bundle(bundle_loc, trash_loc):
+            print >> sys.stderr, 'Moving %s to trash...' % bundle_loc
+            trash_path = os.path.join(trash_loc, os.path.basename(bundle_loc))
+            path_util.rename(bundle_loc, trash_path)
+
+        # Scan files in the data directory for every partition, moving bundles that are marked as garbage to the trash
+        partitions, _ = path_util.ls(self.partitions)
+        trash_count = 0
+        for partition in partitions:
+            partition_path = os.path.join(self.partitions, partition, MultiDiskBundleStore.DATA_SUBDIRECTORY)
+            for entry in reduce(lambda d,f: d + f, path_util.ls(partition_path)):
+                bundle = os.path.join(partition_path, entry)
+                if self.__check_should_trash(bundle, model):
+                    _trash_bundle(bundle, trash_dir)
+                    trash_count += 0
+
+        print >> sys.stderr, 'Moved %d objects to the trash at %s' % (trash_count, trash_dir)
+
+
+    def __check_should_trash(self, bundle_path, model):
+        """Checks a path to a bundle to see if it needs to be trashed.
+        """
+
+        UUID_REGEX = re.compile(r'^(0x[0-9a-z]{32})')
+        file_name = os.path.basename(bundle_path)
+        uuid_match = UUID_REGEX.match(file_name)
+        if not uuid_match:
+            return True
+
+        uuid = uuid_match.groups()[0]
+
+        # Check if bundle is in the database.
+        try:
+            bundle_db = model.get_bundle(uuid)
+        except:
+            return True
+
+        if bundle_db.state in [State.READY, State.FAILED]:
+            # This is the set of paths that should be gone if the bundle is either READY or FAILED
+            # If these paths are still around then something has gone wrong and we should move the bundle to the trash.
+            dep_paths = [
+                    os.path.join(bundle_path, dep.child_path)
+                    for dep in bundle_db.dependencies
+                  ]
+            for path in dep_paths:
+                if os.path.exists(path):
+                    return True
+
+            # Bundles marked as READY or FAILED should not end with .cid, .status or .sh
+            if bundle_path.endswith('.cid') or bundle_path.endswith('.status') or bundle_path.endswith('.sh'):
+                return True
+
+        return False
+
 
--- a/codalab/lib/bundle_store.py
+++ b/codalab/lib/bundle_store.py
@@ -352,7 +352,7 @@ class MultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin, BundleStore
             new_partition = self.ring.get_node(bundle)
             relocations[bundle] = os.path.join(self.partitions, new_partition)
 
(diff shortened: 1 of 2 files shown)


codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+        """
+        MultiDiskBundleStore.health_check(): In the MultiDiskBundleStore, bundle contents are stored on disk, and
+        occasionally the disk gets out of sync with the database, in which case we make repairs in the following ways:
+
+            1. Creates a trash directory
+            2. Moves the contents of bundles with corresponding UUID not in the database to trash
+            3. Any files that don't begin with a UUID string should be moved to trash
+            4. For each bundle marked READY or FAILED, ensure that its dependencies are not located in the bundle
+               directory.
+            5. Bundles that are stored in <UUID>.cid or <UUID>.status/ should be marked as READY or FAILED
```
percyliang: What do you mean 'stored in <UUID>.cid'?  If the bundle is READY or FAILED, then those files should be moved to trash.

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+            6. <UUID>.sh and <UUID>-internal.sh should be gone for bundles marked READY or FAILED
```
percyliang: Say more accurately, moved to trash.

5 and 6 (and perhaps 4) can be consolidated into: if bundle is done, then make sure it doesn't have random junk associated with it.

percyliang:
Another thing we should do: after cleaning up, if the data_hash doesn't exist for a bundle, then compute the data_hash.

Also add a more expensive flag that goes through and makes sure all the data_hashes are accurate.

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+        """
+        # Create a trash directory
+        trash_dir = os.path.join(self.codalab_home, 'trash')
```
percyliang: Let's not move things into a trash directory because this might involve transferring between disks.  Instead, prints out 'rm -r ...' on the things that should be deleted it, and have a force mode that actually executes these commands.

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+        file_name = os.path.basename(bundle_path)
+        uuid_match = UUID_REGEX.match(file_name)
+        if not uuid_match:
+            return True
+
+        uuid = uuid_match.groups()[0]
+
+        # Check if bundle is in the database.
+        try:
+            bundle_db = model.get_bundle(uuid)
```
percyliang: Is this slow with 10K bundles?  If so, we might consider batching these calls...

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+                bundle = os.path.join(partition_path, entry)
+                if self.__check_should_trash(bundle, model):
+                    _trash_bundle(bundle, trash_dir)
+                    trash_count += 0
+
+        print >> sys.stderr, 'Moved %d objects to the trash at %s' % (trash_count, trash_dir)
+
+
+    def __check_should_trash(self, bundle_path, model):
+        """Checks a path to a bundle to see if it needs to be trashed.
```
percyliang: bundle_path is misleading - this is actually any path (possibly ending with '.cid')

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+
+        # Check if bundle is in the database.
+        try:
+            bundle_db = model.get_bundle(uuid)
+        except:
+            return True
+
+        if bundle_db.state in [State.READY, State.FAILED]:
+            # This is the set of paths that should be gone if the bundle is either READY or FAILED
+            # If these paths are still around then something has gone wrong and we should move the bundle to the trash.
```
percyliang: Shouldn't delete the whole bundle!  Should just delete the dependencies inside the bundle.

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,76 @@ def cleanup(self, uuid, dry_run):
...
+            dep_paths = [
+                    os.path.join(bundle_path, dep.child_path)
+                    for dep in bundle_db.dependencies
+                  ]
+            for path in dep_paths:
+                if os.path.exists(path):
+                    return True
+
+            # Bundles marked as READY or FAILED should not end with .cid, .status or .sh
+            if bundle_path.endswith('.cid') or bundle_path.endswith('.status') or bundle_path.endswith('.sh'):
```
percyliang: If there are other things in the store with <UUID>.<random-ext>, then we should print a warning (but don't delete it just to be safe, since later we'll have .tar.gz).

a10y:
@percyliang If there's a bundle found that has a bad hash, should it delete it, update the data_hash stored in the database or just print a warning?

a10y:
I just have it print a warning now but that can be changed, all other functionality is present.

a10y:
(Fixed a comment)

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,130 @@ def cleanup(self, uuid, dry_run):
...
+                trash_count += 1
+                _delete_path(to_delete)
+
+
+            # Check for each bundle if we need to compute its data_hash
+            data_hash_recomputed = 0
+
+            for bundle in bundle_paths:
+                uuid = _get_uuid(bundle)
+                bundle_model = db_bundle_by_uuid[uuid]
```
percyliang: I get a KeyError when there's a directory with a uuid, but is not in the DB.  This should be treated as to_delete.

percyliang:
Print the warning for now.  Have a command-line option that fixes the data
hashes.

codalab/lib/bundle_store.py
a10y: Yeah that was my mistake, I forgot that bundles not in the db will not necessarily have been deleted by this line, only if the -f flag had been used. Should be fixed now.

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,132 @@ def cleanup(self, uuid, dry_run):
...
+            except:
+                return None
+
+        def _is_bundle(path):
+            """Returns whether the given path is a bundle directory/file"""
+            return _get_uuid(path) == os.path.basename(path)
+
+        def _check_bundle_paths(bundle_paths, db_bundle_by_uuid):
+            """
+            Takes in a list of bundle paths, and returns a list of paths and subpaths that need to be removed.
```
percyliang: Document what db_bundle_by_uuid is

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,132 @@ def cleanup(self, uuid, dry_run):
...
+                _delete_path(to_delete)
+
+
+            # Check for each bundle if we need to compute its data_hash
+            data_hash_recomputed = 0
+
+            for bundle in bundle_paths:
+                uuid = _get_uuid(bundle)
+                bundle_model = db_bundle_by_uuid.get(uuid, None)
+                if bundle_moel == None:
```
percyliang: typo

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,132 @@ def cleanup(self, uuid, dry_run):
...
+                trash_count += 1
+                _delete_path(to_delete)
+
+
+            # Check for each bundle if we need to compute its data_hash
+            data_hash_recomputed = 0
+
+            for bundle in bundle_paths:
+                uuid = _get_uuid(bundle)
+                bundle_model = db_bundle_by_uuid.get(uuid, None)
```
percyliang: let's just call this bundle instead of bundle_model to be consistent

a10y:
Let me integrate that and also fix a few other things, I'll push again soon...

a10y:
Okay, that should wrap it up.

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,138 @@ def cleanup(self, uuid, dry_run):
...
+                            model.update_bundle(bundle, db_update)
+                    elif compute_data_hash and data_hash != bundle.data_hash:
+                        data_hash_recomputed += 1
+                        print >> sys.stderr, 'Bundle %s should have data_hash %s, actual digest is %s' % (bundle_path, bundle.data_hash, data_hash)
+                        if repair_hashes and force:
+                            db_update = dict(data_hash=data_hash)
+                            model.update_bundle(bundle, db_update)
+
+
+        print >> sys.stderr, 'Deleted %d objects from the bundle store' % trash_count
```
percyliang: Only print this if --force; print different message otherwise

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,138 @@ def cleanup(self, uuid, dry_run):
...
+            3. Deletes any files not beginning with UUID string.
+            4. For each bundle marked READY or FAILED, ensure that its dependencies are not located in the bundle
+               directory. If they are then delete the dependencies.
+            5. For bundle <UUID> marked READY or FAILED, <UUID>.cid or <UUID>.status, or the <UUID>(-internal).sh files
+               should not exist.
+        |force|: Perform any destructive operations on the bundle store the health check determines are necessary. False by default
+        |compute_data_hash|: If True, compute the data_hash for every single bundle ourselves and see if it's consistent with what's in
+                             the database. False by default.
+        """
+        UUID_REGEX = re.compile(r'^(0x[0-9a-z]{32})')
```
percyliang: Can't we use the one in spec_util?

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,138 @@ def cleanup(self, uuid, dry_run):
...
+
+        def _delete_path(loc):
+            cmd = 'rm -r %s' % loc
```
percyliang: Quote this just in case

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,138 @@ def cleanup(self, uuid, dry_run):
...
+                    elif '.' in path:
+                        print >> sys.stderr, 'WARNING: File %s is likely junk.' % path
+            return to_delete
+
+
+        partitions, _ = path_util.ls(self.partitions)
+        trash_count = 0
+
+        for partition in partitions:
+            partition_path = os.path.join(self.partitions, partition, MultiDiskBundleStore.DATA_SUBDIRECTORY)
```
percyliang, line 483: Print out 'Looking for trash in partition ...'

codalab/lib/bundle_store.py
```diff
@@ -383,3 +395,138 @@ def cleanup(self, uuid, dry_run):
...
+                _delete_path(to_delete)
+            for to_delete in _check_other_paths(other_paths, db_bundle_by_uuid):
+                trash_count += 1
+                _delete_path(to_delete)
+
+
+            # Check for each bundle if we need to compute its data_hash
+            data_hash_recomputed = 0
+
+            for bundle_path in bundle_paths:
```
percyliang, line 508: Print out 'checking data_hash....'

codalab/lib/bundle_store.py
a10y: UUID_REGEX in spec_util assumes that the entire string is the UUID, here we just wanted to do regex matching to see if the filename begins with a full uuid. I can use spec_util.UUID_STR to build this regex though

codalab/lib/bundle_store.py
a10y: I felt like having these summary statistics might be nice as well for the dry-run, but I can also hide them

percyliang: Summary statistics are good - just don't say 'Deleted' when they haven't
been deleted.  Print 'use --force to actually delete'.

a10y:
@percyliang LGTY?

percyliang:
LGTM

Issue #359, opened 2016-02-29
Create ~/.codalab/misc_temp automatically


kashizui:
I get this error when running `cl rest-server` if `~/.codalab/misc_temp` directory doesn't exist yet:

```
➜  cl rest-server                                                                                                                                             16-02-28 17:44:02
Bottle v0.12.9 server starting up (using GunicornServer(workers=1, worker_class='gevent'))...
Listening on http://127.0.0.1:8080/
Hit Ctrl-C to quit.

[2016-02-28 17:44:15 -0800] [32790] [INFO] Starting gunicorn 19.4.5
[2016-02-28 17:44:15 -0800] [32790] [INFO] Listening at: http://127.0.0.1:8080 (32790)
[2016-02-28 17:44:15 -0800] [32790] [INFO] Using worker: gevent
[2016-02-28 17:44:15 -0800] [32790] [INFO] Unhandled exception in main loop:
Traceback (most recent call last):
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/arbiter.py", line 180, in run
    self.manage_workers()
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/arbiter.py", line 482, in manage_workers
    self.spawn_workers()
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/arbiter.py", line 548, in spawn_workers
    self.spawn_worker()
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/arbiter.py", line 502, in spawn_worker
    self.cfg, self.log)
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/workers/async.py", line 24, in __init__
    super(AsyncWorker, self).__init__(*args, **kwargs)
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/workers/base.py", line 55, in __init__
    self.tmp = WorkerTmp(cfg)
  File "/Users/sckoo/Developer/codalab-cli/venv/lib/python2.7/site-packages/gunicorn/workers/workertmp.py", line 23, in __init__
    fd, name = tempfile.mkstemp(prefix="wgunicorn-", dir=fdir)
[...]

PR #361, opened 2016-03-01
 #359: Force initialization of bundle store

a10y:


a10y:
@percyliang Take a look at latest version, @kashizui might want to try this just to sanity check me. We're just adding a call to bundle_store() so nothing should break because of this

percyliang:
LGTM

Issue #362, opened 2016-03-02
work_manager not prepending '0x' to data_hash


a10y:
When UUID-based storage was being written, I moved the data_hash calculation into work_manager, but forgot to prepend the '0x' to data_hash. Currently all bundles that were created since the upgrade will have a malformed data_hash string, which needs to be fixed by prepending '0x' to them all.

kashizui:
Work manager is no longer used.

PR #363, opened 2016-03-02
 #362: Fix issue with work_manager not appending '0x' to data_hash

a10y:
@percyliang This will have affected any bundles created since the UUID-based storage upgrade occurred over a week ago. Do you want me to check-in a fix-it script as well or should I just go in and do the fix on the two instances?


percyliang:
LGTM

As for the fix, let's just fix it with your bs-check-health script, which has the option of recomputing all the data hashes.

a10y:
Yes, this bug was caught with very convenient timing :)

PR #364, opened 2016-03-02
Set user on oauth protected views

kashizui:
Sets `local.user` for views protected by the `require_oauth` decorator.

Also create new `use_oauth` decorator for views where an oauth token is optional. `local.user` will be `None` if no valid token is provided.

@klopyrev


klopyrev:
Feel free to drop this request. I'll send out a pull request that subsumes this code.

PR #367, opened 2016-03-03
Add script to create root user.

klopyrev:
Add a script to create the root user, one that has ID "0", matching root_user_id of the codalab manager.

Also, some minor style changes.

@percyliang @kashizui


percyliang:
LGTM

PR #368, opened 2016-03-03
Some changes to REST setup.

klopyrev:
1. Place all REST endpoints behind /rest/. The idea is that we'll see up reverse proxying so that it forwards requests, such as https://worksheets.codalab.org/rest/xxx to localhost:2900/rest/xxxx. The Bottle side change is needed so that redirection works properly (e.g. when during account sign-up, the redirect goes to https://worksheets.codalab.org/rest/account/success, instead of https://worksheets.codalab.org/account/success.
2. Don't monkey patch when debugging rest-server, since we're not using gevent. This change is needed so that you can connect the Python remote debugger to the Gunicorn process.
3. Return a non-HTML error response for Ajax calls.
4. Minor style clean-up.
5. Increase Gunicorn timeout.
6. Remove unnecessary /status API.

@percyliang @kashizui

### DIFF
--- a/codalab/bin/cl.py
+++ b/codalab/bin/cl.py
@@ -94,6 +94,10 @@ def do_rest_server_command(bundle_cli, args):
     if args.watch:
         run_server_with_watch()
     else:
+        if not args.debug:
+            # gevent.monkey.patch_all() needs to be called before importing
+            # bottle.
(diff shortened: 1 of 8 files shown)


percyliang:
LGTM

PR #369, opened 2016-03-03
Changes to auth setup

klopyrev:
Specifically,

1) When verifying the access token in the request, don't check request.forms since that screws up the input stream and breaks file uploading.
2) For all requests check either the cookie or access token by default. Populate request.user, instead of local.user.
3) Force account verification for most APIs via a plugin.

@kashizui @percyliang

### DIFF
--- a/codalab/server/authenticated_plugin.py
+++ b/codalab/server/authenticated_plugin.py
@@ -1,0 +1,54 @@
+from bottle import (
+  abort,
+  httplib,
+  redirect,
+  request,
+  url,
+)
+
+
+class AuthenticatedPlugin(object):
+    """
+    Fails the request if the user is not authenticated (i.e. request.user hasn't
+    been set).
+
+    This method redirects the user to login unless the request is an AJAX request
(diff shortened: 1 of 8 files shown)


codalab/server/authenticated_plugin.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class AuthenticatedPlugin(object):
+    """
+    Fails the request if the user is not authenticated (i.e. request.user hasn't
+    been set).
+
+    This method redirects the user to login unless the request is an AJAX request
+    (i.e. has X-Requested-With header set to XMLHttpRequest). The worker system
+    sets this header as well and so should any other code that doesn't want
```
percyliang: It's not just the worker system...it's also the CLI in the future.  In the documentation, let's be more explicit (bulleted list) about all the use cases.

percyliang:
LGTM

codalab/server/authenticated_plugin.py
klopyrev: Done.

PR #370, opened 2016-03-03
Fix DJANGO startup.

klopyrev:
Currently, the Django server fails to start-up, because codalab_manager.py has a dependency on bottle.py through server_util and bottle.py is not available in the codalab-worksheets virtual environment. Break the dependency to fix Django.

@percyliang @kashizui


percyliang:
LGTM

PR #373, opened 2016-03-04
Add a RestOAuth handler

klopyrev:
Add a RestOAuth handler that uses the REST server to generate and validate tokens, and that uses the users database for the get_users call. When we launch the new OAuth set up we'll need to update config.json to 

```
    "auth": {
        "class": "RestOAuthHandler"
    },
```

Note that with this setup token generation works the same as before. Token validation, however, is different. Since the REST server by default checks for a valid access token for every REST request to validate the token I've simply added an API that doesn't do much, aside from requiring request.user to be set. Finally, get_users doesn't get users from the OAuth server, instead using the local database, which is shared by the bundle service and the REST bundle service.

This change requires https://github.com/codalab/codalab-cli/pull/369 to be merged first.

Additionally, add a LocalUserAuthHandler that will be used to construct a LocalBundleClient to handle the legacy API calls migrated from Django to the Bottle app.

@percyliang @kashizui

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1468,23 +1468,40 @@ class BundleModel(object):
         :param username: username or email of user to fetch
         :return: User object, or None if no matching user.
         """
+        user_ids = None
+        usernames = None
+        if user_id is not None:
+            user_ids = [user_id]
+        if username is not None:
+            usernames = [username]
+        result = self.get_users(user_ids, usernames)
+        if result:
+            return result[0]
+        return None
+
+    def get_users(self, user_ids=None, usernames=None):
+        """
+        Get users.
+
+        :param user_ids: user ids of users to fetch
+        :param usernames: usernames or emails of users to fetch
+        :return: list of matching User objects
+        """
         clauses = []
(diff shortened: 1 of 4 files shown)


klopyrev:
Ugh, this change is broken. Don't review it yet.

klopyrev:
Nevermind, please review this change.

codalab/model/bundle_model.py
```diff
@@ -1468,23 +1468,40 @@ def get_user(self, user_id=None, username=None):
...
+            rows = connection.execute(select([
                 cl_user
-            ]).where(and_(*clauses)).limit(1)).fetchone()
-
-        if row is None or not row.is_active:
-            return None
+            ]).where(and_(*clauses))).fetchall()
 
-        return User(row)
+        return [User(row) for row in rows]
```
percyliang, line 1504: Weird, we weren't returning an array before...?

percyliang:
LGTM

codalab/model/bundle_model.py
klopyrev, line 1504: I updated the method to get multiple users at a time.

PR #374, opened 2016-03-05
Add versions of all used REST APIs in Django.

klopyrev:
The changes are what's needed to finish the migration of the Django REST APIs to the Bottle server. Specifically, I've simply copied the code from codalab-worksheets repo with some minor modification.

The copied code is as follows:

codalab-worksheets/codalab/apps/authenz/oauth.py get_user_token
  ---> codalab-cli/codalab/model/bundle_model.py find_oauth2_token
  ---> codalab-cli/codalab/rest/legacy.py RemoteBundleService._get_user_token

codalab-worksheets/codalab/apps/web/bundles.py BundleService
  ---> codalab-cli/codalab/rest/legacy.py BundleService and RemoteBundleService

codalab-worksheets/codalab/apps/common/worksheet_utils get_worksheets
  ---> codalab-cli/codalab/rest/legacy.py get_sample_worksheets

codalab-worksheets/codalab/apps/web/views.py WorksheetLandingView
  ---> codalab-cli/codalab/rest/legacy.py get_worksheets_landing

codalab-worksheets/codalab/apps/api/views/worksheets_views.py WorksheetsCommandApi, WorksheetContentApi, BundleContentApi, BundleUploadApi, BundleInfoApi
  ---> codalab-cli/codalab/rest/legacy.py post_worksheets_command, get_worksheet_content, post_worksheet_content, get_bundle_content, post_bundle_upload, get_bundle_info, post_bundle_info

The bundle download API is a new, fully REST version since I already had that for the worker code.

Finally, some minor changes include:
- Fix the Bottle template loading logic. It didn't work before unless you ran cl rest-server for the correct directory.
- Increase the cookie expiration time to 30 days. After 30 days the user has to sign in to CodaLab again.
- Add a CSS file to the account API that is used in the front-end to show / hide buttons and display the user name, depending on whether the user is logged in or not. It's definitely kind of hacky, but the alternatives are no better. Specifically, there are two alternatives:
  --- We fetch the user using Ajax in the Javascript app and then update all the fields. This is not that great, because that results in weird flickering and other stuff as the page loads. Either it shows first with no buttons, or it shows first with all buttons. Either way, visually, it doesn't look good.
  --- Before rendering the template, we have Django send the cookie over to the REST server to interpret it and return the user. Then, we do the normal template logic. This is not good, since then Django is not just a dumb static file server.

@percyliang @kashizui

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1727,6 +1727,20 @@ class BundleModel(object):
 
         return OAuth2Token(self, **row)
 
+    def find_oauth2_token(self, client_id, user_id, expires_after):
+        with self.engine.begin() as connection:
+            row = connection.execute(
+                select([oauth2_token])
+                    .where(and_(oauth2_token.c.client_id == client_id,
+                                oauth2_token.c.user_id == user_id,
+                                oauth2_token.c.expires > expires_after))
+                    .limit(1)).fetchone()
+
+        if row is None:
+            return None
+
+        return OAuth2Token(self, **row)
+
     def save_oauth2_token(self, token):
         with self.engine.begin() as connection:
             result = connection.execute(oauth2_token.insert().values(token.columns))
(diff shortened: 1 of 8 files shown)


percyliang:
LGTM

kashizui:
genius!

PR #376, opened 2016-03-07
Fix path to bundle store in monitor script

a10y:


percyliang:
Let's merge this into master for a hotfix.  develop has a lot of changes - let's not deploy that yet.

PR #377, opened 2016-03-07
Few changes to server setup

klopyrev:
1. Get rid of gevent completely. gevent is a non-preemptive worker type. When a request is blocked on some I/O in order for a thread switch to happen the blocked thread must yield. Unfortunately, the MySQL-python library uses an external C library which doesn't get monkey patched. We could switch to pymysql. However, even worse it turns out that gevent doesn't work well with the standard file library. Notably, a file upload blocks other requests from being handled for some periods of time. We're now using gthread, a worker that uses a thread pool.
2. Add a name method to user.py. The reason it is needed is to support using the new user object with the old check_bundles_have_read_permission, check_bundles_have_all_permission, etc. method calls.
3. Don't log requests to some endpoints, such as /oauth2, since there are too many of them and they are not interesting. We won't be logging requests from the workers either.
4. Fix the error handler. It needs to be added to default_app(), not root_app.

@percyliang @kashizui

### DIFF
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -1,4 +1,4 @@
 bottle==0.12.9
-gevent==1.0.2
+futures==3.0.5
 gunicorn==19.4.5
 oauthlib==1.0.3
--- a/codalab/bin/cl.py
+++ b/codalab/bin/cl.py
@@ -84,6 +84,11 @@ def do_server_command(bundle_cli, args):
                  'CPUs.',
             type=int, default=1),
         Commands.Argument(
+            '-t', '--threads',
+            help='Number of threads to use. The server will be able to handle '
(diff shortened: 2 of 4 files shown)


kashizui:
LGTM

requirements-server.txt
```diff
@@ -1,4 +1,4 @@
 bottle==0.12.9
-gevent==1.0.2
+futures==3.0.5
```
percyliang, line 2: Is this used somewhere?

percyliang:
LGTM

requirements-server.txt
klopyrev, line 2: Yes, required by gthread.

PR #380, opened 2016-03-08
fix a bug for worksheets#63

ChengshuLi:
Add fetch_extra flag to get_user_info.

@percyliang Feel free to have a look
@andreweduffy did you change the user table? What is the field is_superuser? Is it the same as is_root_user? Root_user_id can be configured in config.json so it doesn't necessarily have to be '0'.


percyliang:
LGTM

PR #382, opened 2016-03-09
Use refresh token even when access token expires. That's what refresh…

klopyrev:
… tokens are for. Otherwise, we're asking for a password every hour.

@percyliang


percyliang:
LGTM

PR #383, opened 2016-03-09
Client library for talking to Docker to be used by the worker

klopyrev:
@percyliang

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -1,0 +1,287 @@
+from contextlib import closing
+import httplib
+import json
+import logging
+import os
+import socket
+import ssl
+import subprocess
+import sys
+
+
+logger = logging.getLogger(__name__)
+
+
+def wrap_exception(message):
+    def decorator(f):
+        def wrapper(*args, **kwargs):
+            try:
+                return f(*args, **kwargs)
+            except DockerException as e:
+                raise DockerException, \
+                    DockerException(message + ': ' + e.message), \
+                    sys.exc_info()[2]
+            except (httplib.HTTPException, socket.error) as e:
+                raise DockerException, \
+                    DockerException(message + ': ' + str(e)), \
+                    sys.exc_info()[2]
+        return wrapper
+    return decorator
+
+
+class DockerException(Exception):
+    def __init__(self, message):
+        super(DockerException, self).__init__(message)
+
+
+class DockerClient(object):
+    """
+    Methods for talking to Docker.
+    """
+    def __init__(self):
+        self._docker_host = os.environ.get('DOCKER_HOST') or None
+        if self._docker_host:
+            self._docker_host = self._docker_host.replace('tcp://', '')
+
+        cert_path = os.environ.get('DOCKER_CERT_PATH') or None
+        if cert_path:
+            self._ssl_context = ssl.create_default_context(
+                cafile=os.path.join(cert_path, 'ca.pem'))
+            self._ssl_context.load_cert_chain(
+                os.path.join(cert_path, 'cert.pem'),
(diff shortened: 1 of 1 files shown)


worker/docker_client.py
```diff
@@ -0,0 +1,287 @@
...
+                os.path.join(cert_path, 'key.pem'))
+            self._ssl_context.check_hostname = False
+
+        # Test to make sure that a connection can be established.
+        try:
+            self.test()
+        except DockerException:
+            print >> sys.stderr, """
+On Linux, a valid Docker installation should create a Unix socket at
+//var/run/docker.sock.
```
percyliang: Extra /?

worker/docker_client.py
```diff
@@ -0,0 +1,287 @@
...
+            if create_image_response.status != 200:
+                raise DockerException(create_image_response.read())
+
+            # Wait for the download to finish.
+            while True:
+                loop_callback()
+                response = None
+                line = ''
+                while True:
+                    ch = create_image_response.read(1)
```
percyliang, line 130: What's going on here?  This looks kind of weird/hacky - you have to keep on parsing something as JSON?  You can't just read the entire line and parse it as JSON?  Please document.

worker/docker_client.py
```diff
@@ -0,0 +1,287 @@
...
+            'BASHRC=$(pwd)/.bashrc',
+            'cd %s' % docker_bundle_path,
+            # Run as the user that owns the bundle directory. That way
+            # any new files are created as that user and not root.
+            'U_ID=$(stat -c %%u %s)' % docker_bundle_path,
+            'G_ID=$(stat -c %%g %s)' % docker_bundle_path,
+            'sudo -u \\#$U_ID -g \\#$G_ID -n bash -c ' +
+            '"[ -e $BASHRC ] && . $BASHRC; "' +
+            '"export HOME=%s; "' % docker_bundle_path +
+            '\'(%s) >stdout 2>stderr\'' % command.replace('\'', '\'"\'"\''),
```
percyliang, line 167: Does this work for any command with any sort of existing quotes?  This always makes me nervous...

worker/docker_client.py
```diff
@@ -0,0 +1,287 @@
...
+                    break
+                if 'error' in response:
+                    raise DockerException(response['error'])
+
+    @wrap_exception('Unable to start Docker container')
+    def start_container(self, bundle_path, uuid, command, docker_image,
+                        request_network, dependencies):
+        # Set up the command.
+        docker_bundle_path = '/' + uuid
+        docker_command = [
```
percyliang: rename to 'docker_commands', since it's an array

percyliang:
Does this code only work with certain versions of Docker?  If so, we should check the docker version and throw an error if we don't have the right version.

worker/docker_client.py
klopyrev: Done.

worker/docker_client.py
klopyrev, line 130: Added more documentation. It's just one way of parsing things.

worker/docker_client.py
klopyrev, line 167: Added more documentation. I tested this fairly well. For example, including things like:
-  Running multiple commands a time
- Running echo $HOME -> returns  /0x??????????????
- Running echo '$HOME' -> returns $HOME

worker/docker_client.py
klopyrev: Done.

percyliang:
LGTM

klopyrev:
Done.

klopyrev:
Also, I added another commit that fixes two small bugs.

PR #384, opened 2016-03-09
Bundle upload method.

klopyrev:
Add a method for uploading a bundle. Currently, this version is used by the workers to upload data to a bundle that already exists. For uploading new bundles from the website, I recommend adding in query parameters specifying how to create the bundle (e.g. worksheet_uuid, etc)

@percyliang @andreweduffy

### DIFF
--- a/codalab/rest/bundle.py
+++ b/codalab/rest/bundle.py
@@ -5,9 +5,10 @@ import re
 import shutil
 import subprocess
 import tarfile
+import traceback
 import zlib
 
-from bottle import abort, get, local, put, request, response
+from bottle import abort, get, HTTPError, local, put, request, response
 
 from codalab.common import PermissionError, UsageError
 from codalab.lib import path_util, spec_util
@@ -91,6 +92,16 @@ def get_blob(uuid, path=''):
 @put('/bundle/<uuid:re:%s>/contents/blob/' % spec_util.UUID_STR,
      apply=AuthenticatedPlugin())
 def put_blob(uuid):
+    """
+    API to upload the contents of a bundle. This endpoint replaces the current
+    contents with those uploaded.
+
+    If the Content-Disposition header contains the filename and the filename
+    ends with .tar.gz the bundle will be a directory with the extracted
+    contents. If the filename ends with just .gz the bundle will be a file with
+    the contents. When the upload happens from the browser, the filename should
+    be set to the filename on the user's computer.
+    """
     try:
         check_bundles_have_all_permission(local.model, request.user, [uuid])
     except PermissionError as e:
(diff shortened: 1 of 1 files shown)


codalab/rest/bundle.py
```diff
@@ -77,3 +86,81 @@ def get_blob(uuid, path=''):
     response.set_header('Content-Encoding', encoding or 'identity')
 
     return fileobj
+
+
+@put('/bundle/<uuid:re:%s>/contents/blob/' % spec_util.UUID_STR,
+     apply=AuthenticatedPlugin())
+def put_blob(uuid):
+    try:
```
percyliang: Write high-level comments on what this function does.

codalab/rest/bundle.py
```diff
@@ -77,3 +86,81 @@ def get_blob(uuid, path=''):
...
+        local.bundle_store.cleanup(uuid, dry_run=False)
+        bundle_update = {
+            'data_hash': None,
+        }
+        local.model.update_bundle(bundle, bundle_update)
+
+    # Figure out what kind of upload it is.
+    tar_archive = False
+    single_file_archive = False
+    if 'Content-Disposition' in request.headers:
```
percyliang, line 128: Do we always want to unpack tar.gz and gz?  What if we have a bundle whose contents is a tar.gz?  Should the query params specify?

In general, what do we expect filename to be set to?  Something that's related to the bundle name?  Please document.

codalab/rest/bundle.py
klopyrev: Done.

codalab/rest/bundle.py
klopyrev, line 128: I have made the function work like the existing upload mechanism. See what happens when you try to upload a tar archive. I don't see the point of adding a query parameter until someone starts using it. Otherwise, it's just dead code. When Andrew does the migration to this new API, he's not going to use the query parameter. These is simply no place to specify for the user to specify it.

I've documented how the Content-Disposition header should be populated. For reference, when you currently try to upload a file, here's what gets sent in the request (not including the actual binary data). Note the filename in the first section. That is the filename on my computer.

------WebKitFormBoundaryBDtYlm2Ql0Cde31F
Content-Disposition: form-data; name="file"; filename="bundle.tar.gz"
Content-Type: application/gzip

------WebKitFormBoundaryBDtYlm2Ql0Cde31F
Content-Disposition: form-data; name="bundle_type"

dataset
------WebKitFormBoundaryBDtYlm2Ql0Cde31F
Content-Disposition: form-data; name="worksheet_uuid"

0x13894eb2358b437eb2df8bd371599fd1
------WebKitFormBoundaryBDtYlm2Ql0Cde31F--

klopyrev:
I added a bit more bullet proofing.

codalab/rest/bundle.py
percyliang, line 128: Currently, when a user uploads a directory from the CLI, it passes it to tar | gzip and the contents are streamed to the server.  Will this still work without the filename?

klopyrev, line 128: If we migrate that code to use this code, we can make it work.

klopyrev, line 128: Actually, I'm going to drop this pull request for now. I realize that I still need to implement a lot of downloading logic for CLI commands. I'm planning to rewrite all the downloading logic throughout the codalab-cli repo over the next 4 days. It's all a gigantic mess of overly generic methods.

PR #385, opened 2016-03-11
Add Javascript that sets worksheet / bundle title.

klopyrev:
@percyliang


percyliang:
Can you put a comment in there saying why we have to do this hack?

Otherwise, LGTM

klopyrev:
Done.

PR #388, opened 2016-03-11
Fix create default clients script.

klopyrev:
Don't delete default clients if they already exist. There could be oauth tokens for them and the delete will fail. No databases out there should currently have a row that's not correct, so updating is not necessary.

@percyliang @kashizui


percyliang:
LGTM

PR #390, opened 2016-03-12
Clean-up FileServer.

klopyrev:
Clean-up some weirdness with the BundleRPCServer. No idea why the XML RPC logic is split up among 2 different files. Here, I move it all into a single file: BundleRPCServer. The FileServer just does its file stuff.

@percyliang


percyliang:
This is all going to be scrapped in favor of the REST API, so why bother cleaning it up?

LGTM

klopyrev:
It makes it easier for me to migrate to the new worker system. I need to repurpose this class somewhat in a later CL and it'll be easier to understand without all this XML RPC crap in the FileServer class.

PR #392, opened 2016-03-13
Clean-up of download and some upload code.

klopyrev:
This change introduces a download manager that all requests that access bundle content need to go through. In a future CL, this download manager will support getting the data from the worker.

Additionally, other related changes are:

1) Fix all places where files are accessed to check type. Before, there was pretty bad handling of things like symlinks, directories where files are expected, etc.
2) Fix upload code to not leave temporary files when uploads are interrepted.
3) Fix the get_info method (that indexes the contents of a path) to handle symlinks better. Note that isfile returns True for symlinks).
4) Don't follow symlinks where you shouldn't. Ensure all paths are inside the bundle (i.e. no more cl run "cat "%blah//etc/passwd%")
5) Add support for unzipping .gz archives. Switch to using them to transferring single files, as opposed to tar archives.
6) Fix the download REST API. There were some bugs there before.
7) Move a bunch of logic into the worker directory. The reason for doing that is to make the worker code self contained and not depend on other codalab code.

Note, this change is a bit invasive, especially in deprecating some XML RPC APIs. However, it should be pretty easy to migrate and is necessary since with downloading from the worker we can't support things like "tell()" and "seek()".

@percyliang

### DIFF
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -3,10 +3,13 @@ LocalBundleClient is BundleClient implementation that interacts directly with a
 BundleStore and a BundleModel. All filesystem operations are handled locally.
 """
 import base64
+from contextlib import closing
+from cStringIO import StringIO
 import copy
 import datetime
 import os
 import re
+import shutil
 import sys
 import yaml
 
@@ -29,7 +32,6 @@ from codalab.lib import (
     canonicalize,
     path_util,
     worksheet_util,
-    file_util,
     spec_util,
     formatting,
 )
@@ -50,6 +52,7 @@ from codalab.objects.permission import (
 from codalab.model.tables import (
     GROUP_OBJECT_PERMISSION_READ,
 )
+from worker.file_util import un_tar_gzip_directory
 
 
 def authentication_required(func):
@@ -66,10 +69,11 @@ def authentication_required(func):
 
 
 class LocalBundleClient(BundleClient):
-    def __init__(self, address, bundle_store, model, auth_handler, verbose):
+    def __init__(self, address, bundle_store, model, download_manager, auth_handler, verbose):
         self.address = address
         self.bundle_store = bundle_store
         self.model = model
+        self.download_manager = download_manager
         self.auth_handler = auth_handler
         self.verbose = verbose
 
@@ -152,23 +156,6 @@ class LocalBundleClient(BundleClient):
     def search_bundle_uuids(self, worksheet_uuid, keywords):
         keywords = self.resolve_owner_in_keywords(keywords)
         return self.model.search_bundle_uuids(self._current_user_id(), worksheet_uuid, keywords)
-
-    # Helper
-    def get_target_path(self, target):
-        check_bundles_have_read_permission(self.model, self._current_user(), [target[0]])
-        path = canonicalize.get_target_path(self.bundle_store, self.model, target)
-        if os.path.islink(path):
-            raise UsageError('Following symlink disallowed: %s/%s' % (target[0], target[1]))
-        if not os.path.exists(path):
-            # Too stringent, maybe hasn't been created yet.
-            #raise UsageError('Target does not exist: %s/%s' % (target[0], target[1]))
-            return None
-        return path
-
-    # Helper
-    def get_bundle_target(self, target):
-        (bundle_uuid, subpath) = target
-        return (self.model.get_bundle(bundle_uuid), subpath)
 
     def get_worksheet_uuid(self, base_worksheet_uuid, worksheet_spec):
         """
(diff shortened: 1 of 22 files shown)


klopyrev:
Will fix the test.

klopyrev:
Test should be fixed now.

codalab/client/local_bundle_client.py
```diff
@@ -530,21 +511,75 @@ def select_unreadable_worksheets(uuids):
 
         return bundle_dict
 
-    # Return information about an individual target inside the bundle.
+    def check_download_permission(self, target):
```
percyliang: Why do we need a separate method?  If one can read the entire bundle iff one can read the target.

If you want a separate method, rename to check_target_has_read_permission.

codalab/client/local_bundle_client.py
```diff
@@ -530,21 +511,75 @@ def select_unreadable_worksheets(uuids):
...
+        """
+        self._do_download_file(target, out_fileobj=out)
+
+    def _do_download_file(self, target, out_path=None, out_fileobj=None):
+        self.check_download_permission(target)
+        with closing(self.download_manager.stream_file(target[0], target[1], gzipped=False)) as fileobj:
+            if out_path is not None:
+                with open(out_path, 'wb') as out:
+                    shutil.copyfileobj(fileobj, out)
+            else:
```
percyliang: Replace this line with 'if out_fileobj is not None' to be semantically more correct

codalab/lib/bundle_cli.py
```diff
@@ -1491,53 +1515,73 @@ def follow_targets(self, client, bundle_uuid, subpaths):
...
-            if info['state'] in (State.READY, State.FAILED): break
+            # The run finished and we read all the data.
+            if run_finished:
+                break
 
-            # Sleep if nothing happened
+            # If we got no data this time around, check less often.
             if not change:
-                time.sleep(period)
                 period = min(backoff*period, max_period)
```
percyliang: If change, then set backoff = 1

worker/file_util.py
```diff
@@ -0,0 +1,187 @@
...
+    Raises an IOError if the archive is not valid.
+    """
+    class UnGzipStream(object):
+        def __init__(self, fileobj):
+            self._fileobj = fileobj
+            self._decoder = zlib.decompressobj(16 + zlib.MAX_WBITS)
+            self._buffer = ''
+            self._finished = False
+    
+        def read(self, num_bytes=None):
```
percyliang, line 85: Why not use 'gunzip' and just stream it like we do for gzip?  There an asymmetry here.  (Same for tar above).

worker/file_util.py
```diff
@@ -0,0 +1,187 @@
...
+    to be truncated, places truncation_text at the truncation point.
+    """
+    assert(num_head_lines > 0 or num_tail_lines > 0)
+
+    def ensure_ends_with_newline(lines):
+        if lines and not lines[-1].endswith('\n'):
+            lines[-1] += '\n'
+    
+    file_size = os.stat(file_path).st_size
+    with open(file_path) as fileobj:
```
percyliang: head_lines and tail_lines should end at clean line boundaries, but this seems like the last and first lines will be incomplete, respectively in this implementation

percyliang:
This is great, thanks. @klopyrev

worker/file_util.py
klopyrev, line 85: Here's the comment from the code:

```
# Note, that we don't use gzip.GzipFile or the gunzip shell command since
# they require the input file-like object to support either tell() or
# fileno(). Our version requires only read() and close().
```

Same for tar. Subprocess requires fileno() which is not available in the RPCFileHandle.

worker/file_util.py
klopyrev: I thought about this part. The problem is that, at least for the tail lines, it's not easy to determine if the last line is truncated or the user simply didn't add a newline character to the end of the file. I don't want to be presumptuous and just assume that it's the first case, and remove the last line.

This isn't a problem for the head lines since we can tell if they are truncated. I just added the logic to remove the last head line if it doesn't end with a newline.

FYI, the Linux tail command seems to somehow figure out if the last line is truncated or has no newline character. I think it looks at whether any processes have the file open or something.

codalab/lib/bundle_cli.py
klopyrev: The point of the back-off was to poll the file at the rate that the file is changing. 
Anyway, I just got rid of it entirely. It's not that expensive to check for any new file contents every second.

codalab/client/local_bundle_client.py
klopyrev: Done.

codalab/client/local_bundle_client.py
klopyrev: The reason for the separate method is because I need to check permission in bundle_rpc_server and I don't want to access client.model and client._current_user() from there.

worker/file_util.py
percyliang, line 85: I see

worker/file_util.py
percyliang: I was pointing out a different issue, which is that the first line in the
tail_lines could be truncated because you just start reading from some
arbitrary position at the end of the file, which would not in general line
up with line boundaries.

klopyrev:
Also, fixed tests.

percyliang:
LGTM

PR #393, opened 2016-03-14
Bundle contents index.

klopyrev:
Add an index of the bundle contents. Once the worker system is launched, this index will be used, instead of calling path_util.get_info. The worker updates this index pretty often, so it should be fresh.

I'm also moving the get_info method to worker.file_util since it'll be used by the worker.

Add a script to index the current contents of the bundle store. This will be used to launch the new worker system.

@percyliang 

Note: since this change depends on the download_manager change, I've done a trick where I made the pull requests again the download_manager branch in codalab/codalab-cli that is identical to the one in klopyrev/codalab-cli. Once the download_manager change is submitted, I'll fix the pull request to be for the codalab/codalab-cli develop branch.

### DIFF
--- a/alembic/versions/2016031317_add_bundle_contents_index_40d61632fd13.py
+++ b/alembic/versions/2016031317_add_bundle_contents_index_40d61632fd13.py
@@ -1,0 +1,22 @@
+"""Add bundle contents index
+
+Revision ID: 40d61632fd13
+Revises: 58eccccb346d
+Create Date: 2016-03-13 17:00:23.055571
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '40d61632fd13'
+down_revision = '58eccccb346d'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # bundle_contents_index automatically added
+    pass
+
+def downgrade():
+    op.drop_table('bundle_contents_index')
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -341,6 +341,12 @@ class CodaLabManager(object):
         info['disk_quota'] = formatting.parse_size(info['disk_quota'])
         return info
 
+    def launch_new_worker_system(self):
+        try:
+            return self.config['workers']['launch_new_worker_system']
+        except KeyError:
+            return False
+
     @cached
     def model(self):
         """
@@ -365,7 +371,7 @@ class CodaLabManager(object):
 
     @cached
     def download_manager(self):
-        return DownloadManager(self.bundle_store())
+        return DownloadManager(self.launch_new_worker_system(), self.model(), self.bundle_store())
 
     def auth_handler(self, mock=False):
         '''
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -342,10 +342,9 @@ class CodaLabManager(object):
         return info
 
     def launch_new_worker_system(self):
-        try:
-            return self.config['workers']['launch_new_worker_system']
-        except KeyError:
-            return False
+        # TODO: This flag and all code in the False code path of this flag will
+        # get deleted once the new worker system is launched.
+        return self.config['workers'].get('launch_new_worker_system', False)
 
     @cached
     def model(self):
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -1,7 +1,6 @@
 import os
 
 from codalab.common import UsageError
-from codalab.lib import path_util
 from worker.download_util import get_and_check_target_path, get_target_path
 from worker import file_util
 
@@ -18,7 +17,9 @@ class DownloadManager(object):
     TODO(klopyrev): Worker code in a future pull request.
     """
 
-    def __init__(self, bundle_store):
+    def __init__(self, launch_new_worker_system, bundle_model, bundle_store):
+        self._launch_new_worker_system = launch_new_worker_system
+        self._bundle_model = bundle_model
         self._bundle_store = bundle_store
 
     def get_target_info(self, uuid, path, depth):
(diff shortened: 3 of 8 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -341,6 +341,12 @@ def default_user_info(self):
         info['disk_quota'] = formatting.parse_size(info['disk_quota'])
         return info
 
+    def launch_new_worker_system(self):
+        try:
+            return self.config['workers']['launch_new_worker_system']
```
percyliang: Do self.config['workers'].get('...', False) instead

alembic/versions/2016031317_add_bundle_contents_index_40d61632fd13.py
```diff
@@ -0,0 +1,22 @@
+"""Add bundle contents index
```
percyliang, line 1: Can we call this something other than '*_index', which should be reserved for names of indexes in the table?
I propose bundle_contents_metadata.

codalab/lib/codalab_manager.py
```diff
@@ -341,6 +341,12 @@ def default_user_info(self):
         info['disk_quota'] = formatting.parse_size(info['disk_quota'])
         return info
 
+    def launch_new_worker_system(self):
```
percyliang, line 344: rename this from 'new_worker' to just 'worker' and document what this flag means

codalab/lib/download_manager.py
```diff
@@ -26,11 +27,19 @@ def get_target_info(self, uuid, path, depth):
         Returns information about an individual target inside the bundle, or
         None if the target doesn't exist.
         """
-        bundle_path = self._bundle_store.get_bundle_location(uuid)
-        final_path = get_target_path(bundle_path, path)
-        if not os.path.exists(final_path):
-            return None
-        return path_util.get_info(final_path, depth)
+        if not self._launch_new_worker_system:
+            bundle_path = self._bundle_store.get_bundle_location(uuid)
```
percyliang: put a note that this will be deprecated

codalab/lib/codalab_manager.py
percyliang, line 344: I suppose if this is just a temporary flag that will be removed and always be true eventually, then we should write that and call it whatever we want.

codalab/model/bundle_model.py
```diff
@@ -687,6 +693,64 @@ def remove_data_hash_references(self, uuids):
         with self.engine.begin() as connection:
             connection.execute(cl_bundle.update().where(cl_bundle.c.uuid.in_(uuids)).values({'data_hash': None}))
 
+    def update_bundle_contents_index(self, uuid, index):
+        rows = []
```
percyliang: Document carefully what index is - what are the fields and give a concrete example.
(or reference the method in file_util)

codalab/model/bundle_model.py
```diff
@@ -687,6 +693,64 @@ def remove_data_hash_references(self, uuids):
...
+        def recurse(entry, parent_path):
+            entry['bundle_uuid'] = uuid
+            entry['path'] = os.path.join(parent_path, entry['name'])
+            del entry['name']
```
percyliang, line 709: Why delete this (same for entry['contents'])?  This doesn't seem like it's used later and this is a temporary variable anyway.

codalab/model/bundle_model.py
```diff
@@ -687,6 +693,64 @@ def remove_data_hash_references(self, uuids):
...
+        index_copy = copy.deepcopy(index)
+        if index_copy['name'] != uuid:
+            raise UsageError(
+                'Valid indices have the directory with the name as the UUID '
+                'being the top-level directory.')
+        # Clear the top-level name to save space in the database.
+        index_copy['name'] = ''
+        recurse(index_copy, '/')
+        with self.engine.begin() as connection:
+            connection.execute(cl_bundle_contents_index.delete().where(
```
percyliang, line 726: Add comment: delete old and add new

percyliang, line 726: I'm a bit worried about bundles that have lots and lots of files...this might get pretty slow...
We don't have a cap on the number of files (maybe we should at least for intermediate updates).  In the past, a user has created 100K files.

codalab/model/tables.py
```diff
@@ -74,6 +74,20 @@
   sqlite_autoincrement=True,
 )
 
+# Stores information about the files, directories and links stored in the
+# bundle.
+bundle_contents_index = Table(
```
percyliang, line 79: As before, rename to bundle_contents_metadata

codalab/model/tables.py
```diff
@@ -74,6 +74,20 @@
...
+  'bundle_contents_index',
+  db_metadata,
+  Column('bundle_uuid', String(63), ForeignKey(bundle.c.uuid), nullable=False),
+  Column('path', Text, nullable=False),
+  Column('type', String(63), nullable=False),
+  Column('size', Integer, nullable=False),
+  Column('perm', Integer, nullable=False),
+  Column('link', Text, nullable=True),
```
percyliang, line 87: Document what each of these fields are (for 'type', give the possible values)

General comment: Now that 'type' is more offiical/permanent, just like for permissions, let's define constants FILETYPE_FILE, FILETYPE_DIRECTORY, and FILETYPE_LINK instead of using 'file', 'directory', and 'link' throughout the code.

worker/file_util.py
```diff
@@ -7,6 +7,79 @@
...
+                index_contents(os.path.join(path, file_name), depth - 1)
+                for file_name in os.listdir(path)]
+    return result
+
+
+def restrict_contents_index(index, path, depth=None):
+    """
+    Given an index generates a restricted version.
+    
+    First, it finds the entry for the given path. If no entry is found, None is
```
percyliang: say 'entry in the index'

worker/file_util.py
```diff
@@ -7,6 +7,79 @@
...
+    returned.
+
+    Then, if depth is not None, it filters out any entries more than depth levels
+    deep. Depth 0, for example, means only the top-level entry is included, and
+    no contents. Depth 1 means the contents of the top-level are included, but
+    nothing deeper.
+    """
```
percyliang, line 53: Can we make this method not mutable index in place?  Not having side-effects generally is cleaner/safer.  If there's a good reason for not changing it, then at least document that index is mutated.

percyliang:
@klopyrev see comments

alembic/versions/2016031317_add_bundle_contents_index_40d61632fd13.py
klopyrev, line 1: I don't see the word index as causing confusion. It's going to take a bunch of time to switch all my code to use the new name. Unless, you're really against using the word index, I'm going to leave it as is.

The name index, in general, is appropriate for what this is. It's a table listing the contents of what's on the file system.

codalab/model/bundle_model.py
klopyrev, line 709: It's not a temporary variable. Note the rows.append statement.

codalab/model/bundle_model.py
klopyrev, line 726: Done for the comment.

Re, the size of the index: You have a good point. I'll put it on my TODO list for next quarter.

codalab/model/tables.py
klopyrev, line 79: See comment above.

codalab/model/tables.py
klopyrev, line 87: These are well documented in index_contents.

Also, I'm going to push back on this request. It's going to take too much of my time, and I don't see it as being worth it.

worker/file_util.py
klopyrev, line 53: Done.

worker/file_util.py

codalab/model/bundle_model.py
klopyrev: Added a pointer to the file_util file.

codalab/lib/download_manager.py
klopyrev: Done. Added to codalab_manager.py.

codalab/lib/codalab_manager.py
klopyrev, line 344: Added a TODO to remove the flag.

codalab/lib/codalab_manager.py
klopyrev: Done.

percyliang:
LGTM

PR #394, opened 2016-03-14
Fix the arbitrary handling of errors.

klopyrev:
This change adapts exceptions to HTTP errors, as appropriate. No, we don't need to catch all exceptions in the REST APIs, and return appropriate HTTP error codes. This is done automatically.

Also, the returned error message doesn't need to be JSON. It can just be the error string.

@percyliang


codalab/common.py
```diff
@@ -49,6 +57,36 @@ class PermissionError(UsageError):
...
+# Listed in order of most specific to least specific.
+http_codes_and_exceptions = [
+    (httplib.FORBIDDEN, PermissionError),
+    (httplib.UNAUTHORIZED, AuthorizationError),
+    (httplib.NOT_FOUND, NotFoundError),
+    (httplib.BAD_REQUEST, UsageError),
+]
+
+
+def ExceptionToHttpError(e):
```
percyliang: Style: do exception_to_http_error

codalab/common.py
```diff
@@ -49,6 +57,36 @@ class PermissionError(UsageError):
...
+    """
+    Returns the appropriate HTTP error code and message for the given exception.
+    """
+    for known_code, exception_type in http_codes_and_exceptions:
+        if isinstance(e, exception_type):
+            return known_code, e.message
+    return httplib.INTERNAL_SERVER_ERROR, e.message
+
+
+def HttpErrorToException(code, message):
```
percyliang: same comment as above

percyliang:
Can you make sure that some places which depend on catching UsageError's now work with the new exception class?

Otherwise, LGTM

codalab/common.py
klopyrev: Done, thanks for catching that.

codalab/common.py
klopyrev: Done.

klopyrev:
NotFoundError is a subclass of UsageError, so there shouldn't be any issues.

PR #395, opened 2016-03-14
Even better handling of symlinks.

klopyrev:
We really should allow the user to cat a symlink and see the contents of where it points to. However, if it points to outside the bundle, that should be prevented.

This change makes the handling of symlinks are follows:
- If the symlink points to outside the bundle, return a permission error.
- If the symlink is broken, return a not found error.
- Otherwise, return the contents of the symlink.

@percyliang

### DIFF
--- a/worker/download_util.py
+++ b/worker/download_util.py
@@ -1,24 +1,30 @@
+import httplib
 import os
 
 
-def is_valid_bundle_path(bundle_path, path):
+def get_and_check_real_target_path(bundle_path, uuid, path):
     """
-    Checks to ensure that the given path is inside the bundle, after resolving
-    any links.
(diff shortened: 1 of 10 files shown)


klopyrev:
Also, this fixes CircleCI tests.

worker/download_util.py
```diff
@@ -1,24 +1,30 @@
+import httplib
 import os
 
 
-def is_valid_bundle_path(bundle_path, path):
+def get_and_check_real_target_path(bundle_path, uuid, path):
     """
-    Checks to ensure that the given path is inside the bundle, after resolving
-    any links.
+    Checks that the path is not a symlink pointing outside the bundle and not
```
percyliang, line 7: Put a note that this is really important to get right for security issues.

PR #399, opened 2016-03-15
Fix host.

klopyrev:
Make sure you get the forwarded host name from Nginx. It sets X-Forwarded-Host to what the correct hostname should be, not Host.

@kashizui


kashizui:
LGTM

PR #401, opened 2016-03-15
Worker model

klopyrev:
This change includes:

1) Worker database tables. Note, they include some details of scheduling and dependency management, which will be sent in a later CL.
2) Code for workers to register themselves with the bundle service.
3) Code for workers to receive messages and send replies to them.

@percyliang

### DIFF
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -18,6 +18,7 @@ from sqlalchemy.types import (
   Float,
   Enum,
 )
+from sqlalchemy.sql.schema import ForeignKeyConstraint
 
 db_metadata = MetaData()
 
@@ -326,4 +327,59 @@ chat = Table(
   Column('worksheet_uuid', String(63), nullable=True), # What is the id of the worksheet that the sender is on? 
   Column('bundle_uuid', String(63), nullable=True), # What is the id of the bundle that the sender is on? 
   sqlite_autoincrement=True,
+)
+
+# Store information about workers.
+worker = Table(
+  'worker',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), primary_key=True, nullable=False),
+  Column('worker_id', String(127), primary_key=True, nullable=False),
+
+  Column('slots', Integer, nullable=False),  # Number of additional bundles the worker can run.
+  Column('checkin_time', DateTime, nullable=False),  # When the worker last checked in with the bundle service.
+  Column('socket_id', Integer, nullable=False),  # Socket ID worker listens for messages on.
+)
+
+# Store information about all sockets currently allocated to each worker.
+worker_socket = Table(
+  'worker_socket',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
+  Column('worker_id', String(127), nullable=False),
+  # No foreign key constraint on the worker table so that we can create a socket
+  # for the worker before adding the worker to the worker table.
+
+  Column('socket_id', Integer, primary_key=True, nullable=False),
+)
+
+# Store information about the bundles currently running on each worker.
+worker_run = Table(
+  'worker_run',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
+  Column('worker_id', String(127), nullable=False),
+  ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
+
+  Column('socket_id', Integer, nullable=False),  # The socket ID of the worker, to send messages on.
+
+  Column('run_uuid', String(63), ForeignKey(bundle.c.uuid), nullable=False),
+  Index('uuid_index', 'run_uuid'),
+)
+
+# Store information about the dependencies available on each worker.
+worker_dependency = Table(
+  'worker_dependency',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
+  Column('worker_id', String(127), nullable=False),
+  ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
+
+  # No foreign key here, since we don't have any logic to clean-up bundles that
+  # are deleted.
+  Column('dependency_uuid', String(63), nullable=False),
 )
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -364,8 +364,6 @@ worker_run = Table(
   Column('worker_id', String(127), nullable=False),
   ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
 
-  Column('socket_id', Integer, nullable=False),  # The socket ID of the worker, to send messages on.
-
   Column('run_uuid', String(63), ForeignKey(bundle.c.uuid), nullable=False),
   Index('uuid_index', 'run_uuid'),
 )
--- a/codalab/model/worker_model.py
+++ b/codalab/model/worker_model.py
@@ -25,7 +25,7 @@ class WorkerModel(object):
     2) It is used for communication with the workers. This communication happens
        through Unix domain sockets stored in a special directory. This class
        provides methods to allocate sockets (i.e. figure out unique paths in the
-       socket directory), clean-up sockets (i.e. delete the socket files),
+       socket directory), clean up sockets (i.e. delete the socket files),
        listen on these sockets for messages and send messages to these sockets.
     """
     def __init__(self, engine, socket_dir):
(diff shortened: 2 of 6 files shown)


codalab/model/tables.py
```diff
@@ -327,3 +328,58 @@
...
+# Store information about the bundles currently running on each worker.
+worker_run = Table(
+  'worker_run',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
+  Column('worker_id', String(127), nullable=False),
+  ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
+
+  Column('socket_id', Integer, nullable=False),  # The socket ID of the worker, to send messages on.
```
percyliang: Why is socket_id a field of worker_socket and worker_run when it can be gotten from worker?  Shouldn't this only be in one table

codalab/model/tables.py
```diff
@@ -327,3 +328,58 @@
...
+  'worker_dependency',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
+  Column('worker_id', String(127), nullable=False),
+  ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
+
+  # No foreign key here, since we don't have any logic to clean-up bundles that
+  # are deleted.
+  Column('dependency_uuid', String(63), nullable=False),
```
percyliang, line 382: If this dependency exists, do we assume the entire bundle exists or just a subpath?

codalab/model/worker_model.py
```diff
@@ -0,0 +1,322 @@
...
+class WorkerModel(object):
+    """
+    Manages the worker, worker_dependency and worker_socket tables. This class
+    serves 2 primary functions:
+ 
+    1) It is used to add, remove and query information about workers.
+    2) It is used for communication with the workers. This communication happens
+       through Unix domain sockets stored in a special directory. This class
+       provides methods to allocate sockets (i.e. figure out unique paths in the
+       socket directory), clean-up sockets (i.e. delete the socket files),
```
percyliang: "clean up"

codalab/model/worker_model.py
```diff
@@ -0,0 +1,322 @@
...
+            conn.execute(cl_worker_dependency.delete()
+                         .where(and_(cl_worker_dependency.c.user_id == user_id,
+                                     cl_worker_dependency.c.worker_id == worker_id)))      
+            conn.execute(cl_worker.delete()
+                         .where(and_(cl_worker.c.user_id == user_id,
+                                     cl_worker.c.worker_id == worker_id)))
+            
+    def get_workers(self):
+        """
+        Returns information about all the workers in the database.
```
percyliang: Document that it returns a map indexed by (user_id, worker_id)

codalab/model/worker_model.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        def do(sock):
+            sock.settimeout(timeout_secs)
+            try:
+                conn, _ = sock.accept()
+                conn.settimeout(None)  # Need to remove timeout before makefile.
+                return conn.makefile('rb')
+            except socket.timeout:
+                return None
+        if sock is None:
```
percyliang: Can restructure to make it simpler:

```
if sock is None: sock = self.start_listening(socket_id)
```

Then, get rid of the do function.

codalab/model/worker_model.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        self._cleanup_socket(socket_id)
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        sock.bind(self._socket_path(socket_id))
+        sock.listen(0)
+        return sock
+
+    def get_stream(self, socket_id, timeout_secs, sock=None):
+        """
+        Receives a single message on the socket with the given ID and returns a
```
percyliang: I find it a bit confusing that both socket_id and sock are arguments...how should this function be called?

percyliang: Why can't we just have this method always just take a sock, and it's up to the caller to call start_listening on socket_id?

codalab/model/worker_model.py
```diff
@@ -0,0 +1,322 @@
...
+                                     .where(cl_worker_run.c.run_uuid == uuid)).fetchone()
+            precondition(row, 'Trying to find worker for bundle that is not running.')
+            if row:
+                return {
+                    'user_id': row.user_id,
+                    'worker_id': row.worker_id,
+                    'socket_id': row.socket_id,
+                }
+
+    def allocate_socket(self, user_id, worker_id, conn=None):
```
percyliang: It seems like the only place this is called is when 'conn' is not None, so why not just require it to be specified?

codalab/model/tables.py
```diff
@@ -327,3 +328,58 @@
...
+)
+
+# Store information about all sockets currently allocated to each worker.
+worker_socket = Table(
+  'worker_socket',
+  db_metadata,
+
+  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
+  Column('worker_id', String(127), nullable=False),
+  # No foreign key constraint on the worker table so that we can create a socket
```
percyliang, line 352: Why can't we just create the worker entry before worker_socket (we can just fill in slots, checkin_time, socket_id with null-ish values)?

percyliang:
@klopyrev cool stuff - please take a look at the comments

codalab/model/tables.py
klopyrev, line 352: Then, I have to make socket_id nullable. That's no better than leaving out the foreign key constraint. Either way works, so I'm leaving it as is.

codalab/model/tables.py
klopyrev: Yeah, it really doesn't need to be in worker_run. That was premature optimization to avoid having to read 2 tables in get_bundle_worker. Not really a concern and the code is easier to understand without socket_id in worker_run, so I just got rid of it.

worker_socket contains all the sockets that a worker has, both ones it listens on and ones it replies to. The table exists for 2 purposes: to allocate unique IDs and to store those allocated so that they can be cleaned-up later.

codalab/model/tables.py
klopyrev, line 382: Currently, it's the whole bundle. I'll work on making it more granular next quarter.

codalab/model/worker_model.py
klopyrev: Done.

codalab/model/worker_model.py
klopyrev: It doesn't return a map, it returns a list. The code should be pretty clear about what's in that list, as well, so it's self documenting.

codalab/model/worker_model.py
klopyrev: Crap, I was supposed to close the socket in one of those branches. That's why it's like that.

I restructured the code per your comment below, so not a concern.

codalab/model/worker_model.py
klopyrev: Done. Yes, it's cleaner that way.

codalab/model/worker_model.py
klopyrev: This method is public (doesn't have the _ prefix). It's called in code that makes read requests to the worker. Those don't need to do other database requests in the same transaction.

codalab/model/worker_model.py
percyliang: I would still write a sentence saying what the type of the return value is - it's a simple thing to do.

codalab/model/worker_model.py
```diff
@@ -0,0 +1,320 @@
...
+                                     .where(cl_worker_run.c.run_uuid == uuid)).fetchone()
+            precondition(row, 'Trying to find worker for bundle that is not running.')
+            worker_row = connection.execute(cl_worker.select()
+                                            .where(and_(cl_worker.c.user_id == row.user_id,
+                                                        cl_worker.c.worker_id == row.worker_id))).fetchone()
+            return {
+                'user_id': worker_row.user_id,
+                'worker_id': worker_row.worker_id,
+                'socket_id': worker_row.socket_id,
+                }
```
percyliang: style: unindent }

codalab/model/tables.py
percyliang, line 382: Please add a comment to that effect.

percyliang:
LGTM modulo comments

codalab/model/tables.py
klopyrev, line 382: I don't see a point in doing that. It should be pretty obvious that it's whole dependencies and not just subpaths.

codalab/model/worker_model.py
klopyrev: Done.

codalab/model/worker_model.py
klopyrev: Done.

PR #404, opened 2016-03-16
Refactoring uploading code.

klopyrev:
Refactor uploading logic to:

1) Support streaming archives in and add an API to be used by the worker for updating the contents of a bundle. Most of the same logic can be used for the upload REST API.
2) Minimize the use of temporary files.
3) Use the same logic for updating bundles as well for uploading new bundles.
4) Clean up better on failed uploads.

Also, I centralize things more in the upload manager class.

@andreweduffy @percyliang


codalab/lib/upload_manager.py
```diff
@@ -0,0 +1,187 @@
...
+    
+        If |sources| contains one source, then the bundle contents will be that source.
+        Otherwise, the bundle contents will be a directory with each of the sources.
+        Exceptions:
+        - If |git|, then each source is replaced with the result of running 'git clone |source|'
+        - If |unpack| is True or a source is an archive (zip, tar.gz, etc.), then unpack the source.
+        """
+        bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
+        try:
+            path_util.make_directory(bundle_path)
```
percyliang, line 41: But the bundle could be a file, not a directory - how does that work?

codalab/lib/upload_manager.py
```diff
@@ -0,0 +1,187 @@
...
+        if simplify_archive:
+            self._simplify_archive(dest_path)
+
+    def _unpack_fileobj(self, source_filename, source_fileobj, dest_path, simplify_archive):
+        zip_util.unpack(zip_util.get_archive_ext(source_filename), source_fileobj, dest_path)
+        if simplify_archive:
+            self._simplify_archive(dest_path)
+
+    def _simplify_archive(self, path):
+        if not os.path.isdir(path):
```
percyliang, line 121: Add comment 'Modify |path| in place: if |path| is a directory containing exactly one file/directory that's not ignored, then replace path with that file/directory.'

codalab/lib/upload_manager.py
```diff
@@ -0,0 +1,187 @@
...
+            }
+            self._bundle_model.update_bundle(bundle, bundle_update)
+
+        self._bundle_model.update_user_disk_used(bundle.owner_id)
+
+    def has_contents(self, bundle):
+        return os.path.exists(self._bundle_store.get_bundle_location(bundle.uuid))
+
+    def cleanup_existing_contents(self, bundle):
+        self._bundle_store.cleanup(bundle.uuid, dry_run=False)
```
percyliang, line 190: When is this called?  If we're deleting the bundle, then why update the bundle metadata?

codalab/lib/zip_util.py
```diff
@@ -51,43 +40,48 @@ def strip_archive_ext(path):
...
+        except (tarfile.TarError, IOError):
+            raise UsageError('Invalid archive upload.')
+        finally:
+            if close_source:
+                source.close()
     else:
-        path_util.rename(tmp_path, dest_path)
+        delete_source = False
+        try:
+            # unzip doesn't accept input from standard input, so we have to save
```
percyliang, line 73: That's unfortunate.  The CLI and worker both use tar, right?  I think zip is only used when people upload from the website.

codalab/client/local_bundle_client.py
```diff
@@ -255,34 +256,17 @@ def upload_bundle(self, sources, follow_symlinks, exclude_patterns, git, unpack,
...
-            data_hash = None
-        if construct_args.get('data_hash', data_hash) != data_hash:
-            print >>sys.stderr, 'ERROR: provided data_hash doesn\'t match: %s versus %s' %\
-                                (construct_args.get('data_hash'), data_hash)
-        construct_args['data_hash'] = data_hash
-
-        # Set the owner
+        construct_args['uuid'] = spec_util.generate_uuid()
         construct_args['owner_id'] = self._current_user_id()
+        construct_args['data_hash'] = None  # Will be set later.
```
percyliang: Is this explicitly needed?  The default should be none.

codalab/lib/upload_manager.py
percyliang, line 41: I see - you have simplify_directory below - make a note of this.

percyliang:
Just a few minor comments.

Handling all the compression types (gz/tar/zip/bzip2) with formats (url, git) with files versus directories does become a bit of a bear, but maybe that's just unavoidable at this point.

LGTM

codalab/lib/upload_manager.py
klopyrev, line 190: It's called in the worker code when uploading a new version of the bundle contents. If the upload fails, that code tries to not leave the metadata in an inconsistent state.

codalab/lib/zip_util.py
klopyrev, line 73: Yup, only used for uploads from the website.

codalab/lib/upload_manager.py
klopyrev, line 41: Done.

codalab/lib/upload_manager.py
klopyrev, line 121: Done.

codalab/client/local_bundle_client.py
klopyrev: Done.

klopyrev:
Yup, it's quite a mess. I tried to make the code more readable, but it's still a mess.

PR #407, opened 2016-03-17
Obsolete PR: Worker scheduling

klopyrev:


codalab/lib/bundle_cli.py
```diff
@@ -1434,6 +1434,7 @@ def wrap(string):
         ),
     )
     def do_cat_command(self, args):
+        # TODO: Real?? fail_if_headless? What about wait.
         self._fail_if_headless('cat')  # Files might be too big
```
percyliang, line 1438: We can get rid of this and have people use at their own risk.

codalab/worker/bundle_manager.py
```diff
@@ -0,0 +1,401 @@
...
+
+        if 'default_docker_image' not in config:
+            print >> sys.stderr, 'workers config missing default_docker_image'
+            exit(1)
+        self._default_docker_image = config['default_docker_image']
+
+        if 'torque' in config:
+            assert(self._worker_model.shared_file_system)
+            self._use_torque = True
+            self._torque_host = config['torque']['host']
```
percyliang, line 34: Maybe make this more explicit - 'ssh_host' - this is simply where we ssh into to run commands.

percyliang:
The main thing that I'd react against is the strong dependence on Torque.  The way the code was set up before provided decoupling with the 'dispatch-torque' or 'dispatch-q' commands.  Different groups who might want to use CodaLab will have their own setups.  In fact the NLP group will probably switch from Torque to Slurm at some point.  So it would be good to have something a bit more decoupled / general if possible.

codalab/lib/bundle_cli.py
klopyrev, line 1438: This TODO wasn't supposed to make it into the PR. =)

Yeah, I think it's worth it removing the check. I don't buy the "files might be too big" comment.

codalab/worker/bundle_manager.py
klopyrev, line 34: Done.

klopyrev:
Added a subclass hierarchy. Will update the PRs in a bit.

PR #408, opened 2016-03-17
The rest of the worker system code

klopyrev:
@percyliang


klopyrev:
@percyliang You can play around with the code at http://klopyrev.net:8000. I'll leave that running until I'm back from vacation.

You should have lots of time to review. I'm not planning on doing much for the next 1.5 weeks.

codalab/bundles/run_bundle.py
```diff
@@ -39,11 +39,14 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('time_user', float, 'Amount of time (seconds) by user.', generated=True, formatting='duration'))
     METADATA_SPECS.append(MetadataSpec('time_system', float, 'Amount of time (seconds) by the system.', generated=True, formatting='duration'))
     METADATA_SPECS.append(MetadataSpec('memory', float, 'Amount of memory (bytes) used by this run.', generated=True, formatting='size'))
+    # TODO(klopyrev): These statistics are wrong in the old worker system and are not reported in the new worker system.
+    #                 These metadata fields should be removed.
```
percyliang, line 43: Where will they be reported then?

codalab/client/local_bundle_client.py
```diff
@@ -320,7 +322,15 @@ def kill_bundles(self, bundle_uuids):
         """
         check_bundles_have_all_permission(self.model, self._current_user(), bundle_uuids)
         for bundle_uuid in bundle_uuids:
-            self.model.add_bundle_action(bundle_uuid, BundleAction.kill())
+            if not self.launch_new_worker_system:
+                self.model.add_bundle_action(bundle_uuid, BundleAction.kill())
```
percyliang, line 302: Put a comment somewhere that once we switch over to the new worker system, all this code under launch_new_worker_system will be deleted.

codalab/client/local_bundle_client.py
```diff
@@ -332,7 +342,36 @@ def write_targets(self, targets, string):
...
+                self.model.add_bundle_action(bundle_uuid, BundleAction.write(subpath, string))
+            else:
+                worker_message = {
+                    'type': 'write',
+                    'uuid': bundle_uuid,
+                    'subpath': subpath,
+                    'string': string,
+                }
+                action_string = BundleAction.write(subpath, string)
+                self._do_bundle_action(bundle_uuid, worker_message, action_string)
```
percyliang, line 332: Why does action_string need to have subpath and string inside it, since worker_message already has everything.

Can we make this just self.do_bundle_action(worker_message)?

codalab/model/bundle_model.py
```diff
@@ -591,6 +591,61 @@ def batch_update_bundles(self, bundles, update, condition=None):
...
+                    'remote': hostname,
+                    'started': start_time,
+                    'last_updated': start_time,
+                },
+            }
+            self.update_bundle(bundle, bundle_update, connection)
+
+        self.update_events_log(
+            user_id=bundle.owner_id,
+            user_name=None,  # Don't know
```
percyliang, line 699: This was there before because the user_name was not in the same DB as user_id, but now we can get rid of this entirely.

codalab/rest/legacy.py
```diff
@@ -173,13 +175,13 @@ def get_bundle_file_contents(self, uuid):
         Otherwise, get stdout and stderr.
         For each file, return a truncated version.
         """
-        def get_summary(info, name):
+        def get_summary(download_manager, info, name):
             if info['type'] == 'file':
                 TRUNCATION_TEXT = (
                     '\n'
                     '... Truncated. Click link above to see full file. ...\n'
```
percyliang, line 183: Change to simply '(truncated)' since 'link' only makes sense in the context of the particular GUI.

worker/dependency_manager.py
```diff
@@ -0,0 +1,104 @@
...
+        self._dependencies = {}
+
+    def dependencies(self):
+        """
+        Returns the UUIDs of all the dependencies stored on the worker.
+        """
+        with self._lock:
+            return self._dependencies.keys()
+
+    def add_dependee(self, uuid, dependee_uuid):
```
percyliang: I think 'dependee' is confusing (and I think it actually means the opposite of what you want).  Use 'child' instead.

worker/file_util.py
```diff
@@ -272,6 +199,20 @@ def ensure_ends_with_newline(lines, remove_line_without_newline=False):
     
     return ''.join(lines)
 
+def get_path_size(path, ignore_paths=[]):
```
percyliang: ignore_names might be better because these are names of directories inside path, not absolute paths (also to be consistent with other parts of the code)

worker/run.py
```diff
@@ -0,0 +1,399 @@
...
+                    self._worker.remove_dependee(dep['parent_uuid'], self._uuid)
+                # Docker creates files for each mounted volume. Delete them.
+                child_path = os.path.realpath(
+                    os.path.join(self._bundle_path, dep['child_path']))
+                if child_path.startswith(self._bundle_path):
+                    remove_path(child_path)
+
+            if not self._worker.shared_file_system:
+                logger.debug('Uploading results for run with UUID %s', self._uuid)
+                self._safe_update_run_status('Uploading results')
```
percyliang, line 362: We should make a more deliberate stance whether the run_status is going to be more like an enum (small set of values, like state) or an open-ended string.  Using caps and spaces suggests the latter.  Might make sense for it to be the former though (because we have failure_message for details).  If so, perhaps making it 'uploading-results'.

codalab/bundles/run_bundle.py
klopyrev, line 43: They won't be reported at all. We don't have a good way of measuring disk I/O.

codalab/client/local_bundle_client.py
klopyrev, line 302: Done. Such a comment already exists in codalab_manager.py

codalab/client/local_bundle_client.py
klopyrev, line 332: Then I have to put an if statement that checks the action type.

codalab/model/bundle_model.py
klopyrev, line 699: What do you mean by this? As in: get rid of what?

codalab/rest/legacy.py
klopyrev, line 183: This method is only used in that GUI. I could try to make it more general by accepting a truncation string, but I don't really see the point.

worker/dependency_manager.py
klopyrev: Done.

worker/file_util.py

worker/run.py
klopyrev, line 362: It's an open ended string meant to help the user understand what a running bundle is doing. I don't really want to make State more granular. Thus, I added this.

klopyrev, line 362: Also, failure_message is used only when a failure actually occurs I guess it makes sense to simplify all the status fields (state, run_status, failure_message). Perhaps that can be done later, time-permitting. For now, this run_status is pretty useful.

codalab/bundles/run_bundle.py
percyliang, line 43: docker supposedly has a mechanism for getting this information.  You're saying that that information is inaccurate?  If they will be correct in the future, we don't need to necessarily remove these fields.

codalab/client/local_bundle_client.py
percyliang, line 332: You mean to generate the action_string?  It'd be cleaner to make the action_string just the json serialized version of worker_message.

codalab/model/bundle_model.py
percyliang, line 699: Remove the user_name field from the events_log DB.  Because we can just look up the user_name in the user table (which didn't exist before).

codalab/rest/legacy.py
percyliang, line 183: I'm not suggesting making it an argument.  I'm just saying simplifying the string to:

```
... (truncated) ...
```

worker/run.py
percyliang, line 362: My question still stands: is run_status going to be a small set of values or an open-ended string?

codalab/bundles/run_bundle.py
klopyrev, line 43: This information is not accurate for mounted volumes. I don't see it useful to keep the fields around. We can add them back in if Docker does ever fix it, although I have my doubts that they even can.

codalab/client/local_bundle_client.py
klopyrev, line 332: This information is displayed to the user. I find the BundleAction format to be more readable.

codalab/model/bundle_model.py
klopyrev, line 699: I will add that to my TODO list.

codalab/rest/legacy.py
klopyrev, line 183: It's helpful to have some message telling the user where they can find the non-truncated version.

worker/run.py
klopyrev, line 362: I don't really understand the question. It is an open-ended, user-readable string that has a small set of values. What is your suggestion and justification for that suggestion?

klopyrev:
This is ready to go now.

worker/bundle_service_client.py
```diff
@@ -0,0 +1,247 @@
...
+        Returns a file-like object and a file name.
+        """
+        response = self._make_request(
+            'GET', '/bundle/' + uuid + '/contents/blob/' + path,
+            headers={'Accept-Encoding': 'gzip'}, return_response=True)
+        match = re.match('filename="(.*)"',
+                         response.headers['Content-Disposition'])
+        return (response, match.group(1))
+
+    def _make_request(self, method, url, query_params=None, headers={}, data=None,
```
kashizui: `headers={}`
Default arguments are only evaluated once when the function is defined so changes to this `{}` will accidentally persist for future calls.
http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

codalab/model/bundle_model.py
```diff
@@ -593,6 +593,149 @@ def batch_update_bundles(self, bundles, update, condition=None):
                 return success
         return True
 
+    def set_waiting_for_torque_bundle(self, bundle, torque_job_handle):
```
percyliang: Please rename this and related things to `waiting_for_worker_startup` and remove all references to `torque`.

codalab/worker/bundle_manager.py
```diff
@@ -0,0 +1,370 @@
...
+        self._upload_manager = codalab_manager.upload_manager()
+        
+        self._exiting_lock = threading.Lock()
+        self._exiting = False
+
+        self._make_uuids_lock = threading.Lock()
+        self._make_uuids = set()
+
+        if 'default_docker_image' not in config:
+            print >> sys.stderr, 'workers config missing default_docker_image'
```
percyliang: Give more details on how to fix this (which file, which field).

percyliang:
`cl bundle-manager` should print something out when it starts up (e.g., important fields) like `cl work-manager` so it's clear it's not just hanging.  Same for `worker.py`.

percyliang:
Is there any documentation on how to actually run the worker?

codalab/model/tables.py
```diff
@@ -348,7 +336,10 @@
   Column('user_id', String(63), ForeignKey(user.c.user_id), primary_key=True, nullable=False),
   Column('worker_id', String(127), primary_key=True, nullable=False),
 
-  Column('slots', Integer, nullable=False),  # Number of additional bundles the worker can run.
+  Column('tag', Text, nullable=True),  # Tag that allows for scheduling runs on specific workers.
```
percyliang, line 339: Make `String(63)` in the spirit of worksheet_tag

codalab/worker/torque_bundle_manager.py
```diff
@@ -0,0 +1,204 @@
...
+            script_args = [
+                '--bundle-service-url',  self._torque_bundle_service_url,
+                '--password-file', self._torque_password_file,
+                '--shared-file-system',
+            ]
+            
+            script_env = {
+                'LOG_DIR': self._torque_log_dir,
+                'WORKER_CODE_DIR': self._torque_worker_code_dir,
+                  # -v doesn't work with spaces, so we have to hack it.
```
percyliang: extra leading space

codalab/worker/torque_bundle_manager.py
```diff
@@ -0,0 +1,204 @@
...
+                'WORKER_ARGS': '|'.join(script_args),
```
percyliang: This is kind of yucky.

worker/run.py
```diff
@@ -0,0 +1,402 @@
...
+        self._worker.finish_run(self._uuid)
+
+    def _check_and_report_resource_utilization(self):
+        new_metadata = {}
+
+        # Get wall clock time.
+        new_metadata['time'] = time.time() - self._start_time
+        if (self._resources['request_time'] and
+            new_metadata['time'] > self._resources['request_time']):
+            self.kill('Out of time')
```
percyliang: Rename to 'Time limit %s exceeded' % formatting.duration_str(...)'

worker/run.py
```diff
@@ -0,0 +1,402 @@
...
+
+        # Get memory, time_user and time_system.
+        new_metadata.update(self._docker.get_container_stats(self._container_id))
+        if (self._resources['request_memory'] and
+            'memory' in new_metadata and
+            new_metadata['memory'] > self._resources['request_memory']):
+            self.kill('Out of memory')
```
percyliang: Rename to 'Memory limit %s exceeded' % ....

worker/run.py
```diff
@@ -0,0 +1,402 @@
...
+
+        # Get disk utilization.
+        with self._disk_utilization_lock:
+            new_metadata['data_size'] = self._disk_utilization
+        if (self._resources['request_disk'] and
+            new_metadata['data_size'] > self._resources['request_disk']):
+            self.kill('Out of disk')
```
percyliang: Rename to 'Disk limit %s exceeded' % ...

Giving the bound is useful.

worker/worker.py
```diff
@@ -0,0 +1,244 @@
...
+            try:
+                with closing(self._bundle_service.get_code()) as code:
+                    remove_path(worker_dir)
+                    un_tar_directory(code, worker_dir, 'gz')
+                    break
+            except Exception:
+                traceback.print_exc()
+                time.sleep(1)
+
+        exit(123)
```
percyliang, line 244: We can have this system-exec the python script directly to obviate the need for worker.sh.

percyliang:
@klopyrev Done giving comments for now.

Are there any tests (something in the spirit of test-cli.py) for all this code?  There are so many cases that it would be nice to have things be more systematic.

worker/worker.py
klopyrev, line 244: It doesn't work. I tried. Although, the reason for why it doesn't work escapes me (I should have documented it. =( )

klopyrev:
No tests. I plan to add some.

codalab/worker/torque_bundle_manager.py
klopyrev: What's your point?

codalab/model/tables.py
klopyrev, line 339: What if a user, for some weird reason, specifies a tag longer than 63 characters and then tries to use that tag in the --request-queue flag?

codalab/worker/bundle_manager.py
klopyrev: Done.

worker/bundle_service_client.py
klopyrev: Wow, good catch.

codalab/model/bundle_model.py
klopyrev: Done.

codalab/worker/torque_bundle_manager.py
klopyrev: Done.

worker/run.py
klopyrev: Done.

worker/run.py
klopyrev: Done.

worker/run.py
klopyrev: Done.

klopyrev:
I also plan to add documentation.

klopyrev:
Finished addressing your comments. @percyliang

codalab/model/tables.py
percyliang, line 339: Why would someone need to do that?

In fact, we want to restrict worksheet tags to not have arbitrary characters in them.

codalab/worker/torque_bundle_manager.py
```diff
@@ -0,0 +1,213 @@
...
+        This method implements a state machine. The states are:
+
+        Staged, no job_handle, no worker, no worker_run DB entry:
+            Ready to run on a CodaLab-owned Torque worker or on a user-owned
+            worker.
+
+        For user-owned workers, the states are the same as used in the
+        DefaultBundleManager:
+
+        Starting, no job_handle, has worker_run DB entry:
```
percyliang: to make it clearer, change to 'STARTING: no job_handle ...' (same for all other states).

codalab/worker/torque_bundle_manager.py
```diff
@@ -0,0 +1,213 @@
...
+            Run message sent, waiting for the run to start.
+        Running, no job_handle, has worker_run DB entry:
+            Worker reported that the run has started.
+        Ready / Failed, no job_handle, no worker_run DB entry:
+            Finished.
+
+        For CodaLab-owned Torque workers:
+
+        Waiting for Torque, has job_handle, worker starting, no worker_run DB entry:
```
percyliang: should be 'waiting for worker startup' now

percyliang:
Almost there - just the one issue regarding tags and another comment about
comments. @kopyrev

On Mon, May 2, 2016 at 3:18 PM, Konstantin Lopyrev <notifications@github.com

klopyrev:
Addressed last comments.

klopyrev:
Subsumed by https://github.com/codalab/codalab-cli/pull/428

Issue #252, opened 2015-11-22
CodaLab REST API


percyliang:
This is a big change, but will make it easier to develop other clients that access the bundle service programmatically.

PR #409, opened 2016-03-29
Create user API

kashizui:
Implements the users API.
- Fetch and update the authenticated user at `/rest/user`.
- Fetch single users at `/rest/users/<id>`.
- List/search users at `/rest/users`
  - Supports `?filter[user_name]=name1,name2` and ?filter[email]=email1,email2`query params

Also adds unique constraints on username and email fields, and updates the migrate-users script to work with Django app backed by sqlite.

Contributes to #252


codalab/lib/jsonapi.py
```diff
@@ -0,0 +1,21 @@
...
+
+FIELDSET_KEY_PATTERN = r'fields\[(\w+)\]'
+
+
+def parse_fieldsets(query):
+    fieldsets = {}
+    for key, value in query.items():
+        m = re.match(FIELDSET_KEY_PATTERN, key)
+        if m is None:
+            print key
```
kashizui: delete

kashizui:
@klopyrev @percyliang please review

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+
+
+@get('/user', apply=[JsonApiPlugin(), AuthenticatedPlugin()])
+def fetch_authenticated_user():
+    """Fetch authenticated user."""
+    fieldset = request.jsonapi.fields.get('users', None)
+    return UserSchema(only=fieldset).dump(request.user).data
+
 
+@route('/user', method=['PUT', 'PATCH'], apply=[JsonApiPlugin(), AuthenticatedPlugin()])
```
klopyrev: Why PUT?

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+def update_authenticated_user():
+    """Update one or multiple fields of the authenticated user."""
+    # Load update request data
+    try:
+        user_info, errors = UserSchema(strict=True).load(request.json, partial=True)
+    except ValidationError as e:
+        response.status = '403 Forbidden'
```
klopyrev: abort(httplib.FORBIDDEN, e.messages)

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
 
-# TODO(skoo): user creation endpoint first
-@get('/users')
+    # Patch in user_id manually (do not allow requests to change id)
+    user_info['user_id'] = request.user.user_id
+
+    # Ensure that user name is not taken
+    if (user_info.get('user_name', request.user.user_name) != request.user.user_name and
+        local.model.user_exists(user_info['user_name'], None)):
+        response.status = '403 Forbidden'
```
klopyrev: You have wildly different error handling code. In one place you just set status, in another you raise a UsageError, in a third (below) you return an HTTPResponse with code 404. Please be consistent. Personally, I would do away with the JSON API errors return value, and simply use HTTP Error codes and body. See these two pull requests:

https://github.com/codalab/codalab-cli/pull/394
https://github.com/codalab/codalab-worksheets/pull/120

codalab/rest/users.py
klopyrev: Is 403 really the right error code? Read its description carefully: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+
+@get('/user', apply=[JsonApiPlugin(), AuthenticatedPlugin()])
+def fetch_authenticated_user():
+    """Fetch authenticated user."""
+    fieldset = request.jsonapi.fields.get('users', None)
+    return UserSchema(only=fieldset).dump(request.user).data
+
 
+@route('/user', method=['PUT', 'PATCH'], apply=[JsonApiPlugin(), AuthenticatedPlugin()])
+def update_authenticated_user():
```
klopyrev: Seems like you should be checking PUBLIC_USER_FIELDS

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+        return formatting.duration_str(value)
+
+    def _deserialize(self, value, attr, data):
+        return formatting.parse_duration(value)
+
+
+class UserSchema(Schema):
+    id = fields.String(dump_only=True, attribute='user_id')
+    user_name = fields.String()
+    email = fields.Email(dump_only=True)
```
klopyrev, line 40: Why can't you change the email?

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+    request.user = local.model.get_user(request.user.user_id)
+
+    # Return updated user
+    return fetch_authenticated_user()
+
+
+@get('/users/<id>', apply=JsonApiPlugin())
+def fetch_user(id):
+    """Fetch a single user."""
+    user = local.model.get_user(id)
```
klopyrev: Just use request.user?

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+    if request.user is None or request.user.user_id != id:
+        if fieldset is None:
+            fieldset = PUBLIC_USER_FIELDS
+        else:
+            fieldset &= set(PUBLIC_USER_FIELDS)
+
+    return UserSchema(only=fieldset).dump(user).data
+
+
+@route('/users/<id>', method=['PUT', 'PATCH'], apply=JsonApiPlugin())
```
klopyrev: Why PUT?

codalab/rest/users.py
```diff
@@ -1,10 +1,133 @@
...
+def update_user(id):
```
klopyrev: Why is this API needed?

codalab/server/json_api_plugin.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+FIELDSET_KEY_PATTERN = r'fields\[(\w+)\]'
+FILTER_KEY_PATTERN = r'filter\[(\w+)\]'
+
+
+def parse_includes(query):
+    """Includes currently not supported."""
+    if 'include' in query:
+        raise HTTPError('400 Bad Request')
```
klopyrev: abort(httplib.BAD_REQUEST), here an elsewhere.

codalab/server/json_api_plugin.py
```diff
@@ -0,0 +1,110 @@
+"""
```
klopyrev: This JSON API stuff seems like overkill. Just my personal opinion.

codalab/rest/users.py
kashizui: You are right, I should not be using `PUT`.

codalab/rest/users.py
kashizui: I agree, 403 is the wrong one to use. Which one would you suggest? I can't seem to find one that fits.. are we supposed to invent our own?

codalab/rest/users.py
kashizui, line 40: I disabled it for now until I've implemented the re-verification process (changing email should require verification of the new email address.)

codalab/rest/users.py
kashizui: This `PUT` will also be removed.

codalab/rest/users.py
kashizui: Simply for convenience on the client side, but I'm happy to remove it if you think it's unnecessary.

codalab/rest/users.py
kashizui: Oops should be `400 Bad Request`!

codalab/rest/users.py
kashizui: This is great, I will incorporate these patterns.

codalab/server/json_api_plugin.py
kashizui: I'll take a look at simplifying this.

codalab/server/json_api_plugin.py
kashizui: Would you suggest abort or `raise UsageError`?

codalab/rest/users.py
klopyrev: Yes, thanks.

codalab/rest/users.py
klopyrev: Yeah, it's just confusing.

codalab/server/json_api_plugin.py
klopyrev: Evaluate whether you think it's worth it. Are you planning on migrating the rest of the code to use the JSON API? If so, you're welcome to do it. If not, then here it's just overkill that makes the code hard to read unless you're pretty familiar with the JSON API.

Take for example the fieldset stuff. Is it really necessary? What's wrong with just sending all the data. I can see it being useful with some worksheet stuff where reading the extra info requires an extra database call. Even then, at this point it's probably premature optimization.

kashizui:
@klopyrev Stripped down the implementation considerably, please take a look when you get the chance.

codalab/server/json_api_plugin.py
kashizui: You're right, I've decided against implementing anything more than necessary at this moment until we need it.

codalab/rest/users.py
```diff
@@ -1,10 +1,109 @@
...
+
+    Takes the following query parameters:
+        filter[user_name]=name1,name2,...
+        filter[email]=email1,email2,...
+
+    Fetches all users that match any of these usernames or emails.
+    """
+    # Combine username and email filters
+    usernames = set(request.query.get('filter[user_name]', '').split(','))
+    usernames |= set(request.query.get('filter[email]', '').split(','))
```
klopyrev, line 108: If I query with filter[user_name]=blah this generates the set ['', 'blah']. The empty string for some reason matches all names. Thus, this code is broken.

Please test your code. Reviewers shouldn't generally have to do that for you.

codalab/rest/users.py
```diff
@@ -1,10 +1,109 @@
...
+    # Return updated user
+    return fetch_authenticated_user()
+
+
+@get('/users/<id>', apply=JsonApiPlugin())
+def fetch_user(id):
+    """Fetch a single user."""
+    user = local.model.get_user(id)
+    if user is None:
+        abort(httplib.NOT_FOUND)
```
klopyrev: Can you add a string message? Otherwise, it says Unknown Error.

codalab/rest/users.py
```diff
@@ -1,10 +1,109 @@
...
+    if (user_info.get('user_name', request.user.user_name) != request.user.user_name and
+        local.model.user_exists(user_info['user_name'], None)):
+        abort(httplib.BAD_REQUEST, "User name %s is already taken." % user_info['user_name'])
+
+    # Update user
+    local.model.update_user_info(user_info)
+    request.user = local.model.get_user(request.user.user_id)
+
+    # Return updated user
+    return fetch_authenticated_user()
```
klopyrev: This doesn't need to be a separate function.

codalab/server/json_api_plugin.py
```diff
@@ -0,0 +1,15 @@
...
+from bottle import response
+
+
+class JsonApiPlugin(object):
+    api = 2
+
+    def apply(self, callback, route):
+        def wrapper(*args, **kwargs):
+            response.content_type = 'application/vnd.api+json'
```
klopyrev: This isn't necessary since by default if you return a dict the response content type is set to application/json by the JsonPlugin in bottle. In fact, this setting is override by that plugin.

codalab/rest/users.py
kashizui, line 108: I'm not able to reproduce this bug:
http://45.55.25.29:8000/rest/users?filter%5Buser_name%5D=sckoo
returns only the one matching user. 

But in fact there are bugs I didn't catch when I last tested it since `http://45.55.25.29:8000/rest/users` doesn't return any users.. I will do more thorough testing this time, sorry about that.

klopyrev, line 108: I have an idea of why it's a bug. create-root-user.py creates a user with
no email and the empty string matches that. You shouldn't include the empty
string.

By the way, this API is not used...

kashizui, line 108: Ah the root user. I will make sure to discard the empty string.

This was part of Percy's requested features (as was the sparse fieldsets thing) but maybe as we discussed earlier, we can remove this since we should only implement things as we need them?

kashizui, line 108: I certainly don't mind fewer things to maintain/test...

kashizui:
Addressed @klopyrev 's comments. @percyliang should we hold off on implementing a way to query for the users list until we actually need to use it? I implemented a basic one but as @klopyrev pointed out it's not being used right now (and nor is the `/rest/users/<id>` endpoint).

percyliang:
I think it's worth having the users list and being able to get info on
individual users.  It's useful for admin purposes, for example.

kashizui:
OK, then this is tested and ready for review @klopyrev .

klopyrev:
LGTM

PR #411, opened 2016-04-04
Fix config error found during local development setup.

klopyrev:
During local development we always want credential verification to go through the REST server.

@kashizui


kashizui:
LGTM

PR #414, opened 2016-04-08
Log last login

### LINKED ISSUE #410: Log last login
The `last_login` field of the users table is not actually being filled in. Update the authentication plugins to update it.

kashizui:
Update user last_login field on login.

(For now using UTC like date_joined, but need to address time zones in the future. Also updated codalab/codalab-worksheets#134 to indicate that dates are UTC as a stopgap.)

Fixes #410 

@klopyrev

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1820,6 +1820,15 @@ class BundleModel(object):
         user_info['time_used'] += amount
         self.update_user_info(user_info)
 
+    def update_user_last_login(self, user_id):
+        """
+        Update user's last login date to now.
+        """
+        self.update_user_info({
+            'user_id': user_id,
+            'last_login': datetime.datetime.utcnow(),
+        })
+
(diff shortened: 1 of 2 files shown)


codalab/model/bundle_model.py
```diff
@@ -1824,6 +1824,14 @@ def increment_user_time_used(self, user_id, amount):
         user_info['time_used'] += amount
         self.update_user_info(user_info)
 
+    def update_user_last_login(self, user_id):
+        '''
+        Update user's last login date to now.
+        '''
+        user_info = self.get_user_info(user_id)
```
klopyrev: Why are you making this extra database call?

If you call update with user_id and new last_login value it should do the update correctly.

klopyrev:
LGTM, except for 1 comment.

codalab/model/bundle_model.py
kashizui: Good point, fixing.

PR #415, opened 2016-04-10
Fixed: Can't see private bundle 

ChengshuLi:
Let me know if this is the correct behavior. I actually haven't seen any private bundle before I broke the code. Sorry for the delay.

@percyliang @klopyrev

### DIFF
--- a/codalab/rest/legacy.py
+++ b/codalab/rest/legacy.py
@@ -25,7 +25,7 @@ from bottle import (
 from codalab.bundles import get_bundle_subclass
 from codalab.client.local_bundle_client import LocalBundleClient
 from codalab.client.remote_bundle_client import RemoteBundleClient
-from codalab.common import State, UsageError, PermissionError
+from codalab.common import State, UsageError
 from codalab.lib import (
   bundle_cli,
   file_util,
@@ -146,12 +146,9 @@ class BundleService(object):
                 elif isinstance(item['bundle_info'], dict):
                     infos = [item['bundle_info']]
                 for bundle_info in infos:
-                    try:
+                    if bundle_info['bundle_type'] != 'private':
                         target_info = self.get_top_level_contents((bundle_info['uuid'], ''))
                         bundle_info['target_info'] = target_info
-                    except PermissionError, e:
-                        # bundle_info will show a private bundle
-                        pass
                     try:
                         if isinstance(bundle_info, dict):
                             worksheet_util.format_metadata(bundle_info.get('metadata'))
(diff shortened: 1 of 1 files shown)


percyliang:
I think this is catching the error too deep down.  In general, you want to preserve the error as much as possible (know that it's a PermissionsError rather than 'None', which might because the file doesn't exist).  Can you move the error handling more towards the client side?

ChengshuLi:
@percyliang Feel free to have a look.

codalab/rest/legacy.py
```diff
@@ -203,11 +207,14 @@ def get_summary(info, name):
...
-            for item in info['contents']:
-                item['size_str'] = formatting.size_str(item['size'])
-        return info
+        try:
+            info = self.client.get_target_info(target, 1)
+            if info is not None and info['type'] == 'directory':
+                for item in info['contents']:
+                    item['size_str'] = formatting.size_str(item['size'])
+            return info
+        except PermissionError, e:
```
percyliang: 'try' not necessary here because you're just throwing back the exception...

codalab/rest/legacy.py
```diff
@@ -446,7 +453,10 @@ def post_worksheet_content(uuid):
 @get('/api/bundles/content/<uuid:re:%s>/<path:path>/' % spec_util.UUID_STR)
 def get_bundle_content(uuid, path=''):
     service = BundleService()
-    info = service.get_top_level_contents((uuid, path))
+    try:
```
percyliang: try/except needed?

codalab/rest/legacy.py
```diff
@@ -471,7 +481,11 @@ def get_bundle_info(uuid):
     bundle_info = service.get_bundle_info(uuid)
     if bundle_info is None:
         abort(httplib.NOT_FOUND, 'The bundle is not available')
-    bundle_info.update(service.get_bundle_file_contents(uuid))
+    try:
```
percyliang: try/except needed?

klopyrev:
There's a much better and easier to understand way to make this change:
1) Revert all the current changes.
2) Check whether the bundle is private and don't make the call to get_top_level_contents if it is. To check, simplify check: bundle_info['bundle_type'] != 'private'

If not that, then I agree with Percy: You only need the try / except in a single place: around the get_top_level_contents call. Adding it to all the other places just makes the code confusing as hell.

ChengshuLi:
@percyliang  @klopyrev 

Followed Konstantin's comment

codalab/rest/legacy.py
```diff
@@ -146,8 +146,9 @@ def decode_lines(interpreted):
                 elif isinstance(item['bundle_info'], dict):
                     infos = [item['bundle_info']]
                 for bundle_info in infos:
-                    target_info = self.get_top_level_contents((bundle_info['uuid'], ''))
-                    bundle_info['target_info'] = target_info
+                    if bundle_info['bundle_type'] != 'private':
```
percyliang: Use PrivateBundle.BUNDLE_TYPE

ChengshuLi:
@percyliang

ChengshuLi:
@percyliang

ChengshuLi:
Also, how about private worksheet? Should it behave the same? Now it gives a red alert message if you try to access a private worksheet.

percyliang:
That's fine.  If we want to do that, that should be a different PR.

percyliang:
LGTM

PR #417, opened 2016-04-20
Ensure libcuda is picked up correctly.

klopyrev:
It turns out that unless the libcuda.so.352.63 file is included tensorflow doesn't pick up libcuda correctly. I swear this worked before.

Additionally, call ldconfig at the beginning for good measure. It's not strictly required for tensorflow, but it's probably helpful. That comes at the cost of making the file system not read-only. However, that's a non-issue since volumes are generally mounted as read-only so any changes are within the container only. Most files within the container are also owned by root and the user's command is not run as root.

@percyliang

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -65,20 +65,27 @@ to run the worker from the Docker shell.
 """
             raise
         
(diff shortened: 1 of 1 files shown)


percyliang:
LGTM

PR #422, opened 2016-04-24
Fix unicode support for user fields

kashizui:
Make SQLAlchemy correctly convert between Python unicode objects and utf-8 encoded strings in the database for most user text fields, and validate other fields (i.e. user_name) to constrict its character set.

@percyliang @klopyrev


kashizui:
merging hotfix

PR #424, opened 2016-04-27
Ensure messages to the worker are not lost

klopyrev:
Add a mechanism to ensure that the worker actually gets all the messages that are sent to it. Previously, if you started running a bundle and then started clicking on it like a million times trying to read the contents while it was running some of the read messages would be lost and there would be a timeout. This makes the sending even more reliable than before.

@percyliang

### DIFF
--- a/codalab/model/worker_model.py
+++ b/codalab/model/worker_model.py
@@ -199,6 +199,8 @@ class WorkerModel(object):
         sock.listen(0)
         return sock
 
+    ACK = 'a'
+
     def get_stream(self, sock, timeout_secs):
         """
         Receives a single message on the given socket and returns a file-like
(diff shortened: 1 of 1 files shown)


codalab/model/worker_model.py
```diff
@@ -209,6 +209,10 @@ def get_stream(self, sock, timeout_secs):
         sock.settimeout(timeout_secs)
         try:
             conn, _ = sock.accept()
+            # Send Ack. This helps protect from messages to the worker being
+            # lost due to spuriously accepted connections when the socket
+            # file is deleted.
+            conn.sendall('a')
```
percyliang: Can you define `ACK = 'a'` and replace `1` with `len(ACK)` below to more explicitly connect these two parts of the code semantically?

klopyrev: Done.

percyliang:
LGTM

PR #425, opened 2016-04-28
[WIP] Finish REST API

kashizui:
Contributes to #252 

REST-ified
- [x] gls
- [x] gnew
- [x] grm
- [x] ginfo
- [x] uadd
- [x] urm
- [x] search (Sunday 5/1)
- [x] info, metadata only (Sunday 5/1)
- [x] edit (Monday 5/2)
- [x] rm (Monday 5/2)
- [x] perm (Monday 5/2)
- [x] chown (Monday 5/2)
- [x] run (Tuesday 5/3)
- [x] make (Wednesday 5/4)
- [x] mimic, shadow-only (Wednesday 5/4)
- [x] macro, shadow-only (Thursday 5/5)
- [x] wait (Friday 5/6)
- [x] download (Saturday 5/7)
- [x] cat (Saturday 5/7)
- [x] info, with contents (Sunday 5/8)
- [x] [web bundle details page](https://github.com/codalab/codalab-worksheets/pull/162) (Monday 5/9)
- [ ] upload (Thursday 5/12)
- [ ] ls (Friday 5/13)
- [ ] work
- [ ] print
- [ ] new
- [ ] wedit
- [ ] wls
- [ ] wrm
- [ ] add
- [ ] detach
- [ ] wadd
- [ ] wperm
- [ ] kill
- [ ] write
- [ ] logout
- [ ] all tests passing

### DIFF
--- a/codalab/client/rest_client.py
+++ b/codalab/client/rest_client.py
@@ -1,0 +1,94 @@
+from contextlib import closing
+import httplib
+import json
+import urllib
+import urllib2
+import urlparse
+
+
+class RestClientException(Exception):
+    """
+    Exception raised by the RestClient methods on error. If
+    client_error is False, the failure is caused by a server-side error and
+    can be retried.
+    """
+
+    def __init__(self, message, client_error):
+        super(RestClientException, self).__init__(message)
+        self.client_error = client_error
+
+
+class RestClient(object):
+    """
+    Generic class for communicating with a REST API authenticated by OAuth.
+    """
+
+    def __init__(self, base_url):
+        self._base_url = base_url
+
+    def _get_access_token(self):
+        """
+        Should be implemented to return a valid access token.
+        """
+        raise NotImplementedError
+
+    def _make_request(self, method, path, query_params=None, headers=None,
+                      data=None, return_response=False, authorized=True):
+        if headers is None:
+            headers = {}
+
+        if authorized:
+            headers['Authorization'] = 'Bearer ' + self._get_access_token()
+
+        if data is not None and isinstance(data, dict):
+            headers['Content-Type'] = 'application/json'
+            data = json.dumps(data)
+        headers['X-Requested-With'] = 'XMLHttpRequest'
+        if query_params is not None:
+            path = path + '?' + urllib.urlencode(query_params)
+        request = urllib2.Request(self._base_url + path, data=data,
+                                  headers=headers)
+        request.get_method = lambda: method
+        if return_response:
+            return urllib2.urlopen(request)
+        with closing(urllib2.urlopen(request)) as response:
+            if response.headers.get('Content-Type') == 'application/json':
+                response_data = response.read()
+                try:
+                    return json.loads(response_data)
+                except ValueError:
+                    raise RestClientException(
+                        'Invalid JSON: ' + response_data, False)
+
+    def _upload_with_chunked_encoding(self, method, url, query_params, fileobj):
+        # Start the request.
+        parsed_base_url = urlparse.urlparse(self._base_url)
+        if parsed_base_url.scheme == 'http':
+            conn = httplib.HTTPConnection(parsed_base_url.netloc)
+        else:
+            conn = httplib.HTTPSConnection(parsed_base_url.netloc)
+        with closing(conn):
+            conn.putrequest(
+                method, parsed_base_url.path + url + '?' + urllib.urlencode(
+                    query_params))
+
+            # Set headers.
+            conn.putheader('Authorization', 'Bearer ' + self._get_access_token())
+            conn.putheader('Transfer-Encoding', 'chunked')
+            conn.putheader('X-Requested-With', 'XMLHttpRequest')
+            conn.endheaders()
+
+            # Use chunked transfer encoding to send the data through.
+            while True:
+                to_send = fileobj.read(16 * 1024)
+                if not to_send:
+                    break
+                conn.send('%X\r\n%s\r\n' % (len(to_send), to_send))
+            conn.send('0\r\n\r\n')
+
+            # Read the response.
+            response = conn.getresponse()
+            if response.status != 200:
+                raise RestClientException(
+                    httplib.responses[response.status] + ' - ' + response.read(),
+                    response.status >= 400 and response.status < 500)
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -10,6 +10,7 @@ import httplib
 # http://semver.org/
 CODALAB_VERSION = '0.1.7'
 
+
 class IntegrityError(ValueError):
     """
     Raised by the model when there is a database integrity issue.
@@ -56,6 +57,7 @@ class PermissionError(UsageError):
     Raised when access to a resource is refused because the user does not have
     necessary permissions. Similar to HTTP status 403.
     """
+
 
 # Listed in order of most specific to least specific.
 http_codes_and_exceptions = [
--- a/codalab/rest/bundle.py
+++ b/codalab/rest/bundle.py
@@ -1,15 +1,546 @@
 import httplib
 import mimetypes
+import re
+import os
+import sys
+import time
 
-from bottle import abort, get, local, request, response
+from bottle import abort, get, post, delete, local, request, response
+from marshmallow import (
+    ValidationError,
+    Schema as PlainSchema,
+    validate,
+    validates_schema,
+)
+from marshmallow_jsonapi import Schema, fields
 
-from codalab.lib import spec_util, zip_util
-from codalab.objects.permission import check_bundles_have_read_permission
+from codalab.bundles import BUNDLE_SUBCLASSES, get_bundle_subclass
+from codalab.common import UsageError, State
+from codalab.lib import spec_util, zip_util, worksheet_util
+from codalab.lib.server_util import (
+    bottle_patch as patch,
+    json_api_include,
+    json_api_meta,
+    query_get_bool,
+    query_get_list,
+    query_get_type,
+)
+from codalab.lib.spec_util import validate_uuid, validate_child_path
+from codalab.objects.permission import (
+    check_bundles_have_all_permission,
+    check_bundles_have_read_permission,
+    check_worksheet_has_all_permission,
+    PermissionSpec,
+)
+from codalab.rest.users import UserSchema
+from codalab.rest.util import (
+    check_worksheet_not_frozen,
+    get_bundle_infos,
+    resolve_bundle_specs,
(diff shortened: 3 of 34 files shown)


codalab/rest/json_api_client.py
```diff
@@ -0,0 +1,238 @@
...
+                    sys.exc_info()[2]
+            except (urllib2.URLError, httplib.HTTPException, socket.error) as e:
+                raise JsonApiException, \
+                    JsonApiException(message + ': ' + str(e), False), \
+                    sys.exc_info()[2]
+        return wrapper
+    return decorator
+
+
+def unpack_json(raw_json):
```
percyliang: I was under the impression that marshmallow did this for you...

codalab/rest/groups.py
```diff
@@ -0,0 +1,217 @@
...
+        local.model.update_user_in_group(membership['user_id'], membership['group_uuid'], membership['is_admin'])
+    else:
+        membership = membership_updates
+        membership = local.model.add_user_in_group(membership['user_id'], membership['group_uuid'], membership['is_admin'])
+
+    return GroupMembershipSchema().dump(membership).data
+
+
+@delete('/group-memberships/<membership_id>', apply=AuthenticatedPlugin())
+def delete_group_membership(membership_id):
```
percyliang: Is delete with a single membership_id standard?

I thought that with the JSON API, you would be sending some sort of JSON message to delete.

What if you wanted to delete multiple memberships in batch?

Don't you also need to check that the user has permissions to delete this?

kashizui: > Is delete with a single membership_id standard?
> I thought that with the JSON API, you would be sending some sort of JSON message to delete.
> What if you wanted to delete multiple memberships in batch?

Yes but we can support bulk deletes with requests like this:

```
DELETE /articles HTTP/1.1
Content-Type: application/vnd.api+json; ext=bulk
Accept: application/vnd.api+json; ext=bulk

{
  "data": [
    { "type": "articles", "id": "1" },
    { "type": "articles", "id": "2" }
  ]
}
```

Perhaps we can simplify and only allow bulk deletes?

> Don't you also need to check that the user has permissions to delete this?

The `need_admin=True` forces a permission check--it's not very clear though to be sure, I just copied the `get_group_info` function from `LocalBundleClient`.

codalab/rest/json_api_client.py
kashizui: I experimented with deserializing on the client side with marshmallow, but I was struggling to think of a good way to represent, for example, the username of the owner on the client side. If we deserialize back into the original group_info, we get:

```
{
    "memberships": [3, 6, 7],
    "name": "ok-group",
    "owner_id": "6f2e70e0fed2487185fe0f21fffde98f",
    "uuid": "0x4612d33b6709410895b2bee028e05880"
}
```

We also get the owner object and each of the membership objects included as part of the compound response, but in order to look up the owner's username, we would have to manually search through these included objects to find the user matching the owner_id. This search procedure also would not generalize to other resource types (e.g. searching through bundle permissions).

I instead opted to make the CLI/client side ignore any prior knowledge about the structure of the internal DB dicts, and instead just use these client-side JSON API attributes as is, and do a little bit of processing work to wire up the relationships. So instead we get this:

```
{
    "id": "0x4612d33b6709410895b2bee028e05880",
    "memberships": [
        {
            "group": {
                "id": "0x4612d33b6709410895b2bee028e05880"
            },
            "id": 3,
            "is_admin": true,
            "user": {
                "affiliation": "SCOr",
                "date_joined": "Mon Mar 28 22:49:17 2016",
                "email": "sckoo@cs.stanford.edu",
                "first_name": "\u9756\u5d34",
                "id": "6f2e70e0fed2487185fe0f21fffde98f",
                "last_name": "St\u00e9ph\u00e5n",
                "url": "http://sckoo.net/though",
                "user_name": "sckoo"
            }
        },
        {
            "group": {
                "id": "0x4612d33b6709410895b2bee028e05880"
            },
            "id": 6,
            "is_admin": true,
            "user": {
                "affiliation": null,
                "date_joined": "Fri Apr  8 20:53:24 2016",
                "email": "newuser@gmail.com",
                "first_name": null,
                "id": "fbeb0aebabb84da8b844ac9fba4fe95a",
                "last_name": null,
                "url": null,
                "user_name": "newuser"
            }
        },
        {
            "group": {
                "id": "0x4612d33b6709410895b2bee028e05880"
            },
            "id": 7,
            "is_admin": true,
            "user": {
                "affiliation": null,
                "date_joined": "Fri Apr  8 21:03:21 2016",
                "email": "newuser2@gmail.com",
                "first_name": null,
                "id": "9af28fa6a8e140e6a50561f3fcdb040e",
                "last_name": null,
                "url": null,
                "user_name": "newuser2"
            }
        }
    ],
    "name": "ok-group",
    "owner": {
        "affiliation": "SCOr",
        "date_joined": "Mon Mar 28 22:49:17 2016",
        "email": "sckoo@cs.stanford.edu",
        "first_name": "\u9756\u5d34",
        "id": "6f2e70e0fed2487185fe0f21fffde98f",
        "last_name": "St\u00e9ph\u00e5n",
        "url": "http://sckoo.net/though",
        "user_name": "sckoo"
    }
}
```

Now we can directly get the owner's username with `group['owner']['user_name']`. And we also don't need to choose the right deserializer for each of the objects in response, and we don't have to assume anything about how the objects are internally represented in the DB model. The `unpack_json` operation is the same for all responses.

Now I'm defining the function of marshmallow as: transforming db dicts into the JSON API format when formulating responses, and **validating** input data from JSON API requests and transforming them back into directly usable db dicts. I was able to completely factor out the Group ORMObject, and I'm confident I can factor out the Bundle ORMObjects with this as well. Marshmallow is in this case `not` being used on the client side.

percyliang: I think that makes sense.

codalab/rest/groups.py
percyliang: What code actually gets invoked when you make your bulk delete request?  Simplifying sounds good.

Regarding `need_admin=True`, but one shouldn't need admin access to modify one's own permissions...(perhaps I'm misinterpreting something).

kashizui: I haven't written that code yet, but I believe bulk delete code exists for bundles if not for groups.

What modifications to one's own permissions can one do without admin access?

codalab/rest/json_api_client.py
kashizui: If there's anything that seems funny though please let me know lest I screw something up! :)

codalab/rest/groups.py
percyliang: Never mind, you mean admin of the group.

kashizui:
I've decided that it's been causing too much trouble to delete existing/redundant code as I go, since the dependencies are still too intricate not to break thing unintentionally. I'll go back and delete everything once the entire REST API is ready, but in the meanwhile please bear with me, there will be a lot of additional code that is mostly adapted from existing code.

kashizui:
@percyliang should I finish out the REST API on this one PR, or try to separate into two+ PRs?

kashizui:
@klopyrev I've cleaned up the PR, let me know if there's anything else that might be unclear/unnecessary

codalab/client/rest_client.py
```diff
@@ -0,0 +1,94 @@
...
+    client_error is False, the failure is caused by a server-side error and
+    can be retried.
+    """
+
+    def __init__(self, message, client_error):
+        super(RestClientException, self).__init__(message)
+        self.client_error = client_error
+
+
+class RestClient(object):
```
klopyrev, line 21: Put this file into a separate pull request.

codalab/common.py
```diff
@@ -57,12 +58,20 @@ class PermissionError(UsageError):
     necessary permissions. Similar to HTTP status 403.
     """
 
+
+class InternalError(StandardError):
```
klopyrev: PreconditionViolation already does this. No need for this extra class.

codalab/client/rest_client.py
kashizui, line 21: As in, keep my separate JsonApiClient implementation for now in this PR? (I was hoping to use the upload_with_chunked_encoding method, so I should just make a copy of it for now?)

codalab/rest/bundle.py
```diff
@@ -1,15 +1,546 @@
...
+
+    check_bundles_have_read_permission(local.model, request.user, [uuid])
+    return {
+        'data': local.download_manager.get_target_info(uuid, path, depth)
+    }
+
+
+@get('/bundles/<uuid:re:%s>/contents/blob/' % spec_util.UUID_STR, apply=AuthenticatedPlugin())
+@get('/bundles/<uuid:re:%s>/contents/blob/<path:path>' % spec_util.UUID_STR, apply=AuthenticatedPlugin())
+def fetch_contents_blob(uuid, path=''):
```
klopyrev, line 543: Put this change as well as all the places that use this API into a separate pull request.

codalab/rest/bundle.py
```diff
@@ -1,15 +1,546 @@
...
+    if depth < 0:
+        abort(httplib.BAD_REQUEST, "Depth must be at least 0")
+
+    check_bundles_have_read_permission(local.model, request.user, [uuid])
+    return {
+        'data': local.download_manager.get_target_info(uuid, path, depth)
+    }
+
+
+@get('/bundles/<uuid:re:%s>/contents/blob/' % spec_util.UUID_STR, apply=AuthenticatedPlugin())
```
klopyrev, line 541: The worker calls this method and you haven't updated it to rename the API. Do you really need to rename the API? If you submit this after I launch the new worker system renaming it would be bad.

kashizui, line 541: Hm yeah that's why I asked you about this yesterday. What are your thoughts? We could 1) keep this as the singular "bundle" and just deal with it being different from the other, 2) change all the bundle methods to be singular, 3) change ALL the methods to be singular, 4) do this renaming in a separate PR that is merged ASAP.

codalab/rest/bundle.py
kashizui, line 543: Are you referring to the route name change, or the code for handling the extra query parameters? The latter is required to remove the CLI's dependence on LocalBundleClient.

kashizui:
Split out into multiple pull-requests:
#431 
#432 
#433 
#434 
and more

codalab/rest/bundle.py
kashizui, line 541: @klopyrev what were your thoughts on this? I'd like to do my best to keep you unblocked for deploying your worker system.

PR #426, opened 2016-04-29
Ensure more than 10 jobs can run at the same time.

klopyrev:
Fix bug that causes many of the jobs running on codalab.stanford.edu to get stuck in QUEUED state. Not sure who's bright idea it was to use this search_bundle_uuids method.

@percyliang


klopyrev:
I deployed this to the codalab.stanford.edu as a quick fix. It's extremely annoying.

percyliang:
Ah, that's what was going on - thanks for fixing!

LGTM

PR #429, opened 2016-05-11
Limit rate of requests in rest server

kashizui:
Limits unauthenticated users (identified by IP) to 60 requests per hour, and authenticated users to 5,000 requests per hour.

Uses the same HTTP header conventions as [the GitHub API](https://developer.github.com/v3/#rate-limiting).

### DIFF
--- a/codalab/server/rest_server.py
+++ b/codalab/server/rest_server.py
@@ -1,7 +1,8 @@
-from httplib import INTERNAL_SERVER_ERROR, BAD_REQUEST
+from httplib import INTERNAL_SERVER_ERROR, BAD_REQUEST, FORBIDDEN
 import os
 import re
 import sys
+import threading
 import time
 import traceback
 
@@ -16,6 +17,7 @@ from bottle import (
     install,
     local,
     request,
+    response,
     run,
     static_file,
 )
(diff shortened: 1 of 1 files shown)


codalab/server/rest_server.py
```diff
@@ -41,6 +43,77 @@
...
+
+            # Update count and prevent integer overflow
+            self._counts[user_id] = (min(count, limit), first_time)
+
+            # Returning remaining count and (truncated) remaining time
+            return limit - count, time_remaining
+
+
+class RateLimitPlugin(object):
+    """
```
klopyrev, line 79: Make a note that this is per-process. We usually run with multiple processes.

codalab/server/rest_server.py
```diff
@@ -41,6 +43,77 @@
 ]
 
 
+class RequestCounter(object):
```
klopyrev, line 46: I have no idea why you're doing this. If someone really wanted to DDOS us they'd get around your REMOTE_ADDR check.

Do you expect some existing code to send a ton of requests?

codalab/server/rest_server.py
```diff
@@ -41,6 +43,77 @@
...
+                user_id = request.environ.get('REMOTE_ADDR')
+                limit = self.unauth_limit
+
+            remaining, reset = self.request_counter.count(user_id, limit)
+            response.set_header('X-RateLimit-Limit', limit)
+            response.set_header('X-RateLimit-Remaining', remaining)
+            response.set_header('X-RateLimit-Reset', reset)
+            if remaining < 0:
+                # Cannot use abort, since it ignores the headers on response
+                response.status = FORBIDDEN
```
klopyrev, line 110: Code 429 seems more appropriate.

codalab/server/rest_server.py
```diff
@@ -41,6 +43,77 @@
...
+    """
+    Limit the request rate.
+    Request counts are associated with user for authenticated requests,
+    and with IP address for unauthenticated requests. Different limits can be
+    configured for these two cases.
+    Sets the appropriate headers following the GitHub API's conventions.
+    """
+    api = 2
+
+    def __init__(self, window_seconds=60*60, auth_limit=5000, unauth_limit=60):
```
klopyrev, line 88: 60 requests per hour for unauthorized seems too low. What if someone is browsing through the website, looking at the different bundles. I can see them hit 60 requests in a few minutes. With 4 processes, that's 240. Still, a few minutes of  heavy browsing would hit that.

klopyrev, line 88: A worker can easily hit the 5000 per hour limit. I don't support this change at all. I'd just not do this change and worry about too many requests when we start seeing too many requests.

kashizui, line 88: Ah that's fair. That and the multiprocess thing and the weak REMOTEADDR check I think are the nails in the coffin for this one.

kashizui:
Pigeonholing until the need arises, in which case this will require a much more robust implementation...

PR #431, opened 2016-05-12
1. Add RestClient and JsonApiClient

kashizui:
Factor out general methods from BundleServiceClient into RestClient and
add a new JsonApiClient to make calls to JSON API endpoints.

@klopyrev

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -1,0 +1,376 @@
+import httplib
+import socket
+import sys
+import urllib2
+
+from codalab.common import http_error_to_exception, UsageError
+from worker.rest_client import RestClient, RestClientException
+
+
+def wrap_exception(message):
+    def decorator(f):
+        def wrapper(*args, **kwargs):
+            try:
+                return f(*args, **kwargs)
+            except urllib2.HTTPError as e:
+                # Translate known errors to the standard CodaLab errors
+                exc = http_error_to_exception(e.code, e.read())
+                # All standard CodaLab errors are subclasses of UsageError
+                if isinstance(exc, UsageError):
+                    raise exc.__class__, exc, sys.exc_info()[2]
+                else:
+                    # Shunt other exceptions into one class
+                    raise JsonApiException, \
+                        JsonApiException(message.format(*args, **kwargs) +
+                                         ': ' + httplib.responses[e.code] +
+                                         ' - ' + e.read(),
+                                         400 <= e.code < 500), \
+                        sys.exc_info()[2]
+            except RestClientException as e:
+                raise JsonApiException, \
+                    JsonApiException(message.format(*args, **kwargs) +
+                                     ': ' + e.message, e.client_error), \
+                    sys.exc_info()[2]
+            except (urllib2.URLError, httplib.HTTPException, socket.error) as e:
+                raise JsonApiException, \
+                    JsonApiException(message.format(*args, **kwargs) +
+                                     ': ' + str(e), False), \
+                    sys.exc_info()[2]
+        return wrapper
+    return decorator
+
+
+class JsonApiException(RestClientException):
+    """
+    Exception raised by the JsonApiClient methods on error. If
+    client_error is False, the failure is caused by a server-side error and
+    can be retried.
+    """
+
+
+class JsonApiRelationship(object):
+    """
+    Placeholder for a relationship to another resource.
+    Used to build requests to create or update a resource.
+
+    For example, to update the owner of a bundle:
+        client.update('bundles', {
+            'id': '0x7d67f3e0fda249e5b0531670f473c04f',
+            'owner': JsonApiRelationship('users', owner_id)
+        })
+    """
+    def __init__(self, type_, id_):
+        self.type_ = type_
+        self.id_ = id_
+
+
+class JsonApiClient(RestClient):
+    """
+    Simple JSON API client.
+    """
+    def __init__(self, address, authenticate):
+        self._authenticate = authenticate
+
+        base_url = address + '/rest'
+        super(JsonApiClient, self).__init__(base_url)
+
+    def _get_access_token(self):
+        return self._authenticate(self)
+
+    @staticmethod
+    def _get_resource_path(resource_type, resource_id=None):
+        """
+        Build and return API path.
+        """
+        if resource_id is None:
+            return '/' + resource_type
+        else:
+            return '/' + resource_type + '/' + str(resource_id)
+
+    @staticmethod
+    def _pack_params(params):
+        """
+        Process lists into comma-separated strings, and booleans into 1/0.
+        """
+        if params is None:
+            return None
+
+        result = {}
+        for k, v in (params.iteritems() if isinstance(params, dict) else params):
+            if isinstance(v, list):
+                result[k] = ','.join(map(unicode, v))
+            elif isinstance(v, bool):
+                result[k] = int(v)
+            else:
+                result[k] = v
+        return result
+
+    @staticmethod
+    def _unpack_document(document):
+        """
+        Unpack a JSON API document into a plain dict, with the relationship keys
+        wired up to the 'included' resources. Supports multiple levels of
+        relationships. Relationships that do not have a corresponding object in the
+        'included' objects will be filled in with a dict that only has the 'id'
+        key. For example:
+
+         >> unpack_document({
+                'data': {
+                    'id': '123',
+                    'type': 'bundles',
+                    'attributes': {
+                        'name': 'hello'
+                    },
+                    'relationships': {
+                        'owner': {
+                            'data': {
+                                'id': '345',
+                                'type': 'users'
+                            }
+                        },
+                        'parent': {
+                            'data': {
+                                'id': '567',
+                                'type': 'bundles'
+                            }
+                        }
+                    },
+                },
+                'included': [
+                    {
+                        'type': 'users',
+                        'id': '345',
+                        'attributes': {
+                            'name': 'percy',
+                            'affiliation': 'stanford'
+                        }
+                    }
+                ]
+            })
+
+         => {
+                'owner': {
+                    'id': '345',
+                    'name': 'percy',
+                    'affiliation': 'stanford'
+                },
+                'parent': {
+                    'id': '567'
+                },
+                'id': '123',
+                'name': 'hello'
+            }
+
+        If the document does not follow the expected JSON API format, this
+        method throws a JsonApiException.
+        If relationships do not contain resource linkages (the 'data' item),
+        this method also throws a JsonApiException.
+
+        :param document: the JSON-API payload as a dict
+        :return: unpacked resource info as a dict
+        """
+        def unpack_linkage(linkage):
+            # Return recursively unpacked object if the data was included in the
+            # document, otherwise just return the linkage object
+            if linkage is None or (linkage['type'], linkage['id']) not in included:
+                return linkage
+            else:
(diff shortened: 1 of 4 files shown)


codalab/client/rest_client.py
```diff
@@ -0,0 +1,94 @@
...
+
+class RestClient(object):
+    """
+    Generic class for communicating with a REST API authenticated by OAuth.
+    """
+
+    def __init__(self, base_url):
+        self._base_url = base_url
+
+    def _get_access_token(self):
```
klopyrev: Add a comment that the implementation should be thread-safe.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
+from collections import defaultdict
+import urllib2
+
+from codalab.client.bundle_client import BundleClient
+from codalab.client.rest_client import RestClient
+from codalab.common import http_error_to_exception
+
+
+# TODO(sckoo): deal with more error cases
```
klopyrev: Please do this part properly. Exception handling is important. See for example my implementation of wrap_exception.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+        client.update('bundles', uuid, data={
+            'owner': JsonApiRelationship('users', owner_id)
+        })
+    """
+    def __init__(self, type_, id_):
+        self.type_ = type_
+        self.id_ = id_
+
+
+class JsonApiClient(RestClient, BundleClient):
```
klopyrev: Why is this a BundleClient?

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    Simple JSON API client.
+    """
+    def __init__(self, address, auth_handler, authenticate):
+        self.address = address
```
klopyrev: Not used.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+        self.auth_handler = auth_handler
```
klopyrev: Not used.

klopyrev: I shouldn't have to track these down. Please pay more attention to writing clean code.

kashizui: These are used by the CodalabManager to authenticate, I'll add comments to clarify.

codalab/client/json_api_client.py
kashizui: I'm mixing in BundleClient in order inherit its `login` method, in order to fit in with how CodalabManager authenticates a client.

codalab/client/json_api_client.py
kashizui: Same as below.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+        else:
+            path = self._get_resource_path(resource_type, data['id'])
+        params = self._pack_params(params)
+        data = pack_document(data, resource_type)
+        return unpack_document(
+            self._make_request('PATCH', path, query_params=params, data=data))
+
+    @wrap_exception
+    def delete(self, resource_type, resource_id, params=None):
+        if isinstance(resource_id, list):
```
klopyrev: I have no idea why you're supporting both ways of deleting things.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+
+    # Include meta
+    # (Warning: this may overwrite meta present at the resource object level.)
+    if 'meta' in document:
+        result['meta'] = document['meta']
+
+    return result
+
+
+def pack_document(objects, type_):
```
klopyrev: This packing and unpacking of documents seems like premature optimization... Just my two cents.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+        Request to update a resource or resources.
+
+        :param resource_type: resource type as string
+        :param data: update dict or list of update dicts, update dicts
+        must contain 'id' fields specifying the resources to update
+        :param params: dict of query parameters
+        :return:
+        """
+        if isinstance(data, list):
+            path = self._get_resource_path(resource_type)
```
klopyrev: What use case does this fit?

codalab/client/json_api_client.py
kashizui: You're right it would probably best would just be to always do multiple-delete. I haven't implemented bulk-delete for some of the API yet so I will need to make sure to do that in the next PRs.

codalab/client/json_api_client.py
kashizui: I use every feature I implemented here in the subsequent PRs. I kind of built this up incrementally as I needed it.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+                for linkage in relationship['data']:
+                    if linkage['id'] in included[linkage['type']].keys():
+                        obj[key].append(unpack_object(included[linkage['type']][linkage['id']]))
+                    else:
+                        obj[key].append({'id': linkage['id']})
+            else:
+                linkage = relationship['data']
+                if linkage is None:
+                    obj[key] = None
+                elif linkage['id'] in included[linkage['type']]:
```
klopyrev: This is copied code.

codalab/client/json_api_client.py
kashizui: `cl chown` updates the owner of potentially multiple bundles, and so this allows a batch update. But as you pointed out with delete, I can simplify this down to always assuming batch operations.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+            if isinstance(relationship['data'], list):
+                obj[key] = []
+                for linkage in relationship['data']:
+                    if linkage['id'] in included[linkage['type']].keys():
+                        obj[key].append(unpack_object(included[linkage['type']][linkage['id']]))
+                    else:
+                        obj[key].append({'id': linkage['id']})
+            else:
+                linkage = relationship['data']
+                if linkage is None:
```
klopyrev: Use same control flow structure as below for 'data':

data = relationship['data']
if isinstance(..., list):
elif instance(..., dict):
elif None??

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+        obj['id'] = obj_data['id']
+
+        if 'meta' in obj_data:
+            obj['meta'] = obj_data['meta']
+
+        for key, relationship in obj_data.get('relationships', {}).iteritems():
+            if isinstance(relationship['data'], list):
+                obj[key] = []
+                for linkage in relationship['data']:
+                    if linkage['id'] in included[linkage['type']].keys():
```
klopyrev: No need for .keys

codalab/client/json_api_client.py
kashizui: I'll look into refactoring it.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,299 @@
...
+                        obj[key].append(unpack_object(included[linkage['type']][linkage['id']]))
+                    else:
+                        obj[key].append({'id': linkage['id']})
```
klopyrev: Why no type in this case? Just include the whole object.

codalab/client/json_api_client.py
kashizui: Good catch!

klopyrev: You're relying on defaultdict behavior here... That's kind of weird. I'd use a tuple: type, id

codalab/client/json_api_client.py
klopyrev: I don't understand this logic. Please leave it out of the pull request and bring it in once you needed it. I'd like to see the code that uses it.

codalab/client/json_api_client.py
kashizui: @klopyrev Should I make a `JsonApiException` like you did with `BundleServiceException`, or should I try to translate errors to the standard CodaLab errors in `codalab.common`?

kashizui:
@klopyrev Addressed all comments

kashizui:
Found a bug: this causes the worker to throw this error:

```
  File "dev/codalab-cli/worker/main.py", line 15, in <module>
    from bundle_service_client import BundleServiceClient
  File "/Users/sckoo/Developer/codalab-cli/worker/bundle_service_client.py", line 13, in <module>
    from codalab.client.rest_client import RestClient, RestClientException
ImportError: No module named codalab.client.rest_client
```

will have to address this tomorrow.

klopyrev:
Yeah, the worker can't depend on the codalab libraries.

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,369 @@
...
+
+def wrap_exception(message):
+    def decorator(f):
+        def wrapper(*args, **kwargs):
+            try:
+                return f(*args, **kwargs)
+            except urllib2.HTTPError as e:
+                # Translate known errors to the standard CodaLab errors
+                exc = http_error_to_exception(e.code, e.read())
+                if isinstance(exc, UsageError):
```
klopyrev, line 19: Why just this type? What about permissions errors? Why do we even need to translate?

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,369 @@
...
+        included = {
+            (resource['type'], resource['id']): resource
+            for resource in document.get('included', [])
+        }
+
+        # Build result info dict
+        try:
+            data = document.get('data', None)
+            if isinstance(data, list):
+                result = [unpack_object(d) for d in document['data']]
```
klopyrev: s/document['data']/data/

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,369 @@
...
+            elif isinstance(data, dict):
+                result = unpack_object(document['data'])
```
klopyrev: s/document['data']/data/

codalab/client/json_api_client.py
```diff
@@ -0,0 +1,369 @@
...
+        :param params: dict of query parameters
+        :return: the updated object(s)
+        """
+        result = self._unpack_document(
+            self._make_request(
+                method='PATCH',
+                path=self._get_resource_path(resource_type),
+                query_params=self._pack_params(params),
+                data=self._pack_document(data, resource_type)))
+        return result if isinstance(data, list) else result[0]
```
klopyrev, line 350: This line seems broken.

codalab/client/json_api_client.py
klopyrev, line 19: Add some more comments if you keep it this way.

codalab/client/rest_client.py
```diff
@@ -0,0 +1,95 @@
...
+            while True:
+                to_send = fileobj.read(16 * 1024)
+                if not to_send:
+                    break
+                conn.send('%X\r\n%s\r\n' % (len(to_send), to_send))
+            conn.send('0\r\n\r\n')
+
+            # Read the response.
+            response = conn.getresponse()
+            if response.status != 200:
```
klopyrev: This will not get properly translated into a UsageError, even though it has an HTTP error code and it should.

kashizui:
@klopyrev Finished addressing comments
- Moved RestClient into worker directory
- `RestClient._upload_with_chunked_encoding` throws `HTTPError`
- Added comments clarifying the "translation" of errors
- `s/document['data']/data`
- Clarified confusing line in `JsonApiClient.update`

klopyrev:
LGTM

PR #432, opened 2016-05-12
[Invalid] Add groups API

kashizui:
And use the new API in the CLI commands for groups.
Includes some hacks to allow use of JsonApiClient and *BundleClient side-by-side, that will be cleaned up in future PR when the latter is no longer used.
- gls
- gnew
- grm
- ginfo
- uadd
- urm

@klopyrev

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -3,7 +3,11 @@ import socket
 import sys
 import urllib2
 
-from codalab.common import http_error_to_exception, UsageError
+from codalab.client.bundle_client import BundleClient
+from codalab.common import (
+    http_error_to_exception,
+    UsageError,
+)
 from worker.rest_client import RestClient, RestClientException
 
 
@@ -64,12 +68,18 @@ class JsonApiRelationship(object):
         self.id_ = id_
 
 
-class JsonApiClient(RestClient):
+class JsonApiClient(RestClient, BundleClient):
     """
     Simple JSON API client.
     """
-    def __init__(self, address, authenticate):
+    def __init__(self, address, auth_handler, authenticate):
         self._authenticate = authenticate
+
+        # Used by CodaLabManager to look up the associated server address
+        self.address = address
+
+        # Used by BundleClient.login(), which is called by CodaLabManager._authenticate()
+        self.auth_handler = auth_handler
 
         base_url = address + '/rest'
         super(JsonApiClient, self).__init__(base_url)
--- a/codalab/rest/groups.py
+++ b/codalab/rest/groups.py
@@ -1,0 +1,167 @@
+"""
+Worksheets REST API Groups Views.
+"""
+import httplib
+
+from bottle import abort, get, delete, post, request, local
+from marshmallow_jsonapi import Schema, fields
+
+from codalab.lib.server_util import (
+    json_api_include,
+    query_get_list,
+)
+from codalab.lib.spec_util import validate_uuid, validate_name
+from codalab.rest.users import UserSchema
+from codalab.rest.util import get_group_info, ensure_unused_group_name
+from codalab.server.authenticated_plugin import AuthenticatedPlugin
+from codalab.server.json_api_plugin import JsonApiPlugin
+
+#############################################################
+#  GROUP DE/SERIALIZATION AND VALIDATION SCHEMAS
+#############################################################
+
+
+class GroupSchema(Schema):
+    id = fields.String(validate=validate_uuid, attribute='uuid')
+    name = fields.String(required=True, validate=validate_name)
+    user_defined = fields.Bool(dump_only=True)
+    owner = fields.Relationship(include_data=True, type_='users', attribute='owner_id')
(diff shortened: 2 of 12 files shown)


kashizui:
Hm I will need to look into how to make the tests pass on circleCI...

codalab/server/rest_server.py
```diff
@@ -135,6 +136,9 @@ def wrapper(*args, **kwargs):
                 code, message = exception_to_http_error(e)
                 if code == INTERNAL_SERVER_ERROR:
                     traceback.print_exc()
+                    # Internal error details should not be echoed to client
```
klopyrev: Why not? Seems like that would be very useful to us since then the user can report the exact error.

codalab/server/rest_server.py
```diff
@@ -135,6 +136,9 @@ def wrapper(*args, **kwargs):
...
+                    message = 'Server encountered unexpected error while ' \
```
klopyrev: Why is this change part of this pull request?

codalab/rest/groups.py
```diff
@@ -0,0 +1,167 @@
...
+            None
+        )
+    else:
+        groups = local.model.batch_get_all_groups(
+            None,
+            {'owner_id': request.user.user_id, 'user_defined': True},
+            {'user_id': request.user.user_id}
+        )
+
+    for group in groups:
```
klopyrev: I don't see why you're fetching this data in cases where include doesn't include group-memberships and users.

codalab/rest/groups.py
```diff
@@ -0,0 +1,167 @@
...
+def fetch_group(group_spec):
+    """Fetch a single group."""
+    group = get_group_info(group_spec, need_admin=False)
+    document = GroupSchema().dump(group).data
+    include_group_relationships(document, [group])
+    return document
+
+
+@get('/groups', apply=[AuthenticatedPlugin(), JsonApiPlugin()])
+def fetch_groups():
```
klopyrev, line 60: Wow, I have no idea why you're putting in all this time to make this code so generic and flexible. Why support have include-memberships and users be False?

codalab/rest/groups.py
```diff
@@ -0,0 +1,167 @@
...
+    name = fields.String(required=True, validate=validate_name)
+    user_defined = fields.Bool(dump_only=True)
+    owner = fields.Relationship(include_data=True, type_='users', attribute='owner_id')
+    memberships = fields.Relationship(dump_only=True, many=True, type_='group-memberships', include_data=True)
+
+    class Meta:
+        type_ = 'groups'
+
+
+class GroupMembershipSchema(Schema):
```
klopyrev, line 35: This group membership relationship seems like overkill if the only reason you need it for is the is_admin field. How about the GroupSchema has three relationships on user: owner, admins and members?

codalab/client/json_api_client.py
```diff
@@ -62,13 +63,19 @@ def __init__(self, type_, id_):
...
-class JsonApiClient(RestClient):
+class JsonApiClient(RestClient, BundleClient):
     """
     Simple JSON API client.
     """
-    def __init__(self, address, authenticate):
+    def __init__(self, address, auth_handler, authenticate):
         self._authenticate = authenticate
 
+        # Used by CodaLabManager to look up the associated server address
```
klopyrev, line 78: This logic is very convoluted. Please refactor the codaLabManager or whatever so that you don't have to make the JsonApiClient a BundleClient. That makes no sense to me whatsoever. Move the login method if necessary.

Also, split out the change into a separate pull request.

kashizui, line 78: I agree, I'll take some time today to clean this up.

codalab/server/rest_server.py
kashizui: Sorry, it slipped in when I was doing the massive rebasing to split the original PR.

codalab/server/rest_server.py
kashizui: That's a good point-- I was worried about creating security issues by exposing internal error details, but is that perhaps not a concern since everything is open-source anyways?

codalab/server/rest_server.py
kashizui: I'll remove this entirely as per your above comment.

codalab/rest/groups.py
kashizui, line 35: That's a great idea, I'll try it out today.

`POST /groups/<id>/relationships/admins` to add an admin
`POST /groups/<id>/relationships/members` to add a member

If adding someone as admin who is already a member, just silently handle upgrading the `is_admin` flag.

codalab/rest/groups.py
kashizui, line 60: I think I kind of got it in my head early on to create a generic API that can be used by third-party developers.. this has become much less feasible as time gets shorter. We don't have time to test all these generic code paths.

I'll strip this down and leave further development for next year?

codalab/rest/groups.py
kashizui, line 35: @klopyrev Should members be a superset of admins, or disjoint?

PR #433, opened 2016-05-12
Add basic Bundles API

kashizui:
And update CLI to use new API for basic bundle operations, with no
support for fetching target (file) data or metadata in this pull request:
- [x] search (Sunday 5/1)
- [x] info, metadata only (Sunday 5/1)
- [x] edit (Monday 5/2)
- [x] rm (Monday 5/2)
- [x] perm (Monday 5/2)
- [x] chown (Monday 5/2)
- [x] run (Tuesday 5/3)
- [x] make (Wednesday 5/4)
- [x] mimic, shadow-only (Wednesday 5/4)
- [x] macro, shadow-only (Thursday 5/5)

Maintains support for old versions of the client (leaves LocalBundleClient intact).

PR #434, opened 2016-05-12
Add bundle contents API

### LINKED ISSUE #335: upload files directly to the bundle store
Currently, RemoteBundleClient.upload_bundle first (via open_temp_file) copies things into a temporary directory, which is then copied into data.  But let us copy directly into data directory.  More specifically:

1) When user wants to upload something, we should immediately create a bundle with a uuid, add it to the current worksheet, etc., and set its state to a new state 'uploading'.
2) Then, RemoteBundleClient.upload_bundle should upload directly to the bundle data directory (we can even support partial uploads now!), updating the data_size incrementally.
3) When uploading is done, we compute the data_hash, and change the bundle state to 'ready', preventing further modifications.

For 2), we should think about what this will look like under the new REST API.

kashizui:
And update the CLI to use the new API.
- [x] wait (Friday 5/6)
- [x] download (Saturday 5/7)
- [x] cat (Saturday 5/7)
- [x] info, with contents (Sunday 5/8)
- [x] worksheet_util.interpret_file_genpath
- [x] upload
- [x] optimize and clean up upload 
- [x] fix CI tests
- [x] squash and rebase

Fixes #335, #495

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -1,6 +1,7 @@
 import httplib
 import socket
 import sys
+import urllib
 import urllib2
 
 from codalab.common import (
@@ -10,6 +11,7 @@ from codalab.common import (
     UsageError,
 )
 from worker.rest_client import RestClient, RestClientException
+from worker.file_util import un_gzip_stream
 
 
 def wrap_exception(message):
@@ -468,3 +470,72 @@ class JsonApiClient(RestClient):
                 path=self._get_resource_path('user'),
                 query_params=self._pack_params(params),
                 data=self._pack_document(data, 'users')))
+
+    @wrap_exception('Unable to fetch contents info of bundle {1}')
+    def fetch_contents_info(self, bundle_id, target_path='', depth=0):
+        request_path = '/bundles/%s/contents/info/%s' % \
+                       (bundle_id, urllib.quote(target_path))
+        response = self._make_request('GET', request_path,
+                                      query_params={'depth': depth})
+        return response['data']
+
+    @wrap_exception('Unable to fetch contents blob of bundle {1}')
+    def fetch_contents_blob(self, bundle_id, target_path='', range_=None,
+                            head=None, tail=None):
+        """
+        Returns a file-like object for the target on the given bundle.
+
+        :param bundle_id: id of target bundle
+        :param target_path: path to target in bundle
+        :param range_: range of bytes to fetch
+        :param head: number of lines to summarize from beginning of file
+        :param tail: number of lines to summarize from end of file
+        :return: file-like object containing requested data blob
+        """
+        request_path = '/bundles/%s/contents/blob/%s' % \
+                       (bundle_id, urllib.quote(target_path))
+        headers = {'Accept-Encoding': 'gzip'}
+        if range_ is not None:
+            headers['Range'] = 'bytes=%d-%d' % range_
+        params = {}
+        if head is not None:
+            params['head'] = head
+        if tail is not None:
+            params['tail'] = tail
+        response = self._make_request('GET', request_path, headers=headers,
+                                      query_params=params, return_response=True)
+
+        if response.headers.get('Content-Encoding') == 'gzip':
+            return un_gzip_stream(response)
+        return response
+
+    @wrap_exception('Unable to upload contents of bundle {1}')
+    def upload_contents_blob(self, bundle_id, fileobj=None, params=None,
+                             progress_callback=None):
+        """
+        Uploads the contents of the given fileobj as the contents of specified
+        bundle.
+
+        :param bundle_id: the id of the target bundle
+        :param fileobj: file-like object containing the data to upload
+        :param params: dict of query parameters
+        :param progress_callback: function that will be called periodically
+                                  with the number of bytes uploaded so far
+        :return: None
+        """
+        request_path = '/bundles/%s/contents/blob/' % bundle_id
+        params = params or {}
+        params['finalize'] = True
+        params = self._pack_params(params)
+        if fileobj is None:
+            self._make_request(
+                method='PUT',
+                path=request_path,
+                query_params=params)
+        else:
+            self._upload_with_chunked_encoding(
+                method='PUT',
+                url=request_path,
+                query_params=params,
+                fileobj=fileobj,
+                progress_callback=progress_callback)
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -743,13 +743,8 @@ class BundleCLI(object):
         if args.alias:
             address = self.manager.apply_alias(args.alias)
             self.manager.logout(address)
-            # TODO(sckoo): clean up use_rest hack when REST API migration complete
-            self.manager.logout(self.manager.derive_rest_address(address))
         else:
             client = self.manager.current_client()
-            self.manager.logout(client.address)
-            # TODO(sckoo): clean up use_rest hack when REST API migration complete
-            client = self.manager.current_client(use_rest=True)
             self.manager.logout(client.address)
 
     @Commands.command(
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -856,6 +856,7 @@ class BundleCLI(object):
             Commands.Argument('-x', '--exclude-patterns', help='Exclude these file patterns.', nargs='*'),
             Commands.Argument('-g', '--git', help='Path is a git repository, git clone it.', action='store_true'),
             Commands.Argument('-p', '--pack', help='If path is an archive file (e.g., zip, tar.gz), keep it packed.', action='store_true', default=False),
+            Commands.Argument('-z', '--force-compression', help='Always use compression (this may speed up single-file uploads over a slow network).', action='store_true', default=False),
             Commands.Argument('-w', '--worksheet-spec', help='Upload to this worksheet (%s).' % WORKSHEET_SPEC_FORMAT, completer=WorksheetsCompleter),
         ) + Commands.metadata_arguments([UploadedBundle] + [get_bundle_subclass(bundle_type) for bundle_type in UPLOADED_TYPES])
         + EDIT_ARGUMENTS,
@@ -865,11 +866,14 @@ class BundleCLI(object):
         if self.headless and not args.path and args.contents is None:
             return ui_actions.serialize([ui_actions.Upload()])
 
-        # TODO(sckoo): clean up use_rest hack when REST API migration complete
-        client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec, use_rest=True)
+        if args.contents is None and not args.path:
+            raise UsageError("Nothing to upload.")
 
         if args.contents is not None and args.path:
-            raise UsageError("Upload does not support mixing content strings and local files")
+            raise UsageError("Upload does not support mixing content strings and local files.")
+
+        # TODO(sckoo): clean up use_rest hack when REST API migration complete
+        client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec, use_rest=True)
 
         # Build bundle info
         metadata = self.get_missing_metadata(UploadedBundle, args, initial_metadata={})
@@ -922,7 +926,8 @@ class BundleCLI(object):
                 zip_util.pack_files_for_upload(
                     sources, should_unpack=(not args.pack),
                     follow_symlinks=args.follow_symlinks,
-                    exclude_patterns=args.exclude_patterns)
+                    exclude_patterns=args.exclude_patterns,
+                    force_compression=args.force_compression)
 
             print >>self.stderr, 'Uploading %s (%s) to %s' %\
                                  (filename, new_bundle['id'], client.address)
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -21,6 +21,8 @@ import inspect
 import itertools
 import os
 import shlex
+import shutil
+from StringIO import StringIO
 import sys
 import time
 import tempfile
@@ -77,6 +79,8 @@ from codalab.lib.completers import (
 from codalab.lib.bundle_store import (
     MultiDiskBundleStore
 )
+from codalab.lib.print_util import FileTransferProgress
+from worker.file_util import un_tar_directory
 
 # Formatting Constants
 GLOBAL_SPEC_FORMAT = "[<alias>::|<address>::](<uuid>|<name>)"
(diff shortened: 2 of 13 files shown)


kashizui:
Still needs to be rebased on #433

codalab/rest/bundle.py
```diff
@@ -33,21 +567,34 @@ def get_blob(uuid, path=''):
...
             # Let's gzip to save bandwidth. The browser will transparently decode
             # the file.
             filename = filename + '.gz'
-            fileobj = local.download_manager.stream_file(uuid, path, gzipped=True)
+            gzipped = True
+
+        if byte_range and (head_lines or tail_lines):
+            abort(httplib.BAD_REQUEST, 'Head and range not supported on the same request.')
+        elif byte_range:
+            start, end = byte_range
```
klopyrev: nit: Range queries with start and length are easier to understand since you don't have to specify whether end is inclusive or exclusive.

codalab/rest/bundle.py
```diff
@@ -33,21 +567,34 @@ def get_blob(uuid, path=''):
...
+            fileobj = local.download_manager.read_file_section(uuid, path, start, end - start + 1, gzipped)
+        elif head_lines or tail_lines:
+            fileobj = local.download_manager.summarize_file(uuid, path, head_lines, tail_lines, 128, None, gzipped)
```
klopyrev: Don't hard code the maximum line length.

klopyrev: You could give it a default, though. Although, this is minor.

codalab/rest/bundle.py
klopyrev: Nevermind.

klopyrev:
Generally, looks pretty good. Don't forget that you also need to change worksheet_util.interpret_file_genpath, since that makes some calls to get_target_info and cat_target.

klopyrev:
I only looking at the downloading stuff, not the other bundle stuff, though.

kashizui:
@percyliang Finished with upload! I still need to fix the CircleCI test, but everything else is working and ready to be reviewed.

(The issue with the CircleCI test is that the RESTified commands, in particular `cl upload`, no longer work with the `local::` CodaLab instance, as we're now phasing out local instances. But this makes the `wadd` test a bit harder to implement, since we have to run multiple instances of the REST server. I'll give it a shot though soon.)

kashizui:
@percyliang Fixed the test! All ready for review.
Required some major refactoring, which might be easier to pick apart by viewing that commit alone.

kashizui:
@percyliang I'm seeing a 40% improvement in speed for uploading a ~70mb directory on wstest, and a whopping 12x improvement in download speed. I've deployed this branch to wstest, so you can check out this branch and try it out yourself.

percyliang:
@kashizui : great news about the download speed; upload speed is still worrisome though.  Konstantin was getting ~60 MB/sec, whereas even on localhost, I'm only getting ~6 MB/sec.  (I'd test on localhost, not wstest, to make sure Internet speed is not the bottleneck factor.)

kashizui:
Ah ok I was wondering what numbers Konstantin had gotten. I'll look into
this more later this week..

codalab/client/json_api_client.py
```diff
@@ -468,3 +470,61 @@ def update_authenticated_user(self, data, params=None):
...
+                                      query_params=params, return_response=True)
+
+        if response.headers.get('Content-Encoding') == 'gzip':
+            return un_gzip_stream(response)
+        return response
+
+    @wrap_exception('Unable to upload contents of bundle {1}')
+    def upload_contents_blob(self, bundle_id, fileobj=None, params=None,
+                             progress_callback=None):
+        request_path = '/bundles/%s/contents/blob/' % bundle_id
```
percyliang: Please document the arguments here, especially since the uploading logic has so many cases.

codalab/lib/bundle_cli.py
```diff
@@ -728,15 +732,24 @@ def do_status_command(self, args):
 
     @Commands.command(
         'logout',
-        help='Logout of the current session.',
+        help='Logout of the current session, or a specific instance.',
+        arguments=(
+            Commands.Argument('alias', help='Alias or URL of instance from which to logout. Default is the current session.', nargs='?'),
```
percyliang, line 738: name of argument would make more sense as 'instance' rather than 'alias'

codalab/lib/bundle_cli.py
```diff
@@ -801,61 +814,71 @@ def do_upload_command(self, args):
...
-            worksheet_uuid=worksheet_uuid,
-            add_to_worksheet=True,
-        )
-        print >>self.stdout, uuid
+        # Create bundle
+        new_bundle = client.create('bundles', bundle_info, params={
+            'worksheet': worksheet_uuid
+        })
+
+        if args.contents is not None:
```
percyliang: Document more explicitly the ways people can upload by numbering them
1) Contents
2) URL
3) Local paths (usually one)

codalab/lib/print_util.py
```diff
@@ -12,3 +15,45 @@ def clear_line(f=sys.stderr):
...
+        :param bytes_total: Number of bytes total to transfer, or None if unknown
+        """
+        self.f.write('\r')
+        self.f.write(msg)
+        if bytes_total is None:
+            self.f.write(formatting.size_str(bytes_done))
+        else:
+            self.f.write(formatting.ratio_str(formatting.size_str, bytes_done, bytes_total))
+        speed = float(bytes_done) / (time.time() - self.start_time)
+        self.f.write(' [%sbps]' % formatting.size_str(speed))
```
percyliang: This prints out 'mbps' which is not what we want - I'd go with MB/sec explicitly (don't need formatting.size_str, since the dynamic range here isn't going to be that big)

codalab/lib/bundle_cli.py
```diff
@@ -886,16 +909,20 @@ def do_download_command(self, args):
...
             raise UsageError('Target doesn\'t exist.')
-        if target_info['type'] == 'directory':
-            client.download_directory(target, final_path)
-        elif target_info['type'] == 'file':
-            client.download_file(target, final_path)
-        elif target_info['type'] == 'link':
+        if target_info['type'] == 'link':
             raise UsageError('Downloading symlinks is not allowed.')
 
+        # TODO(sckoo): print progress
```
percyliang: Is this easy to just do in this PR?

codalab/lib/bundle_cli.py
```diff
@@ -1542,11 +1568,13 @@ def print_target_info(self, client, target, decorate, maxlines=10, fail_if_not_e
...
-                import base64
-                for line in client.head_target(target, maxlines):
-                    print >>self.stdout, formatting.verbose_contents_str(base64.b64decode(line)),
+                with closing(client.fetch_contents_blob(target[0], target[1], head=maxlines)) as contents:
+                    self.stdout.write(contents.read())
+                    self.stdout.flush()
             else:
-                client.cat_target(target, self.stdout)
+                with closing(client.fetch_contents_blob(target[0], target[1])) as contents:
+                    self.stdout.write(contents.read())
```
percyliang: We should stream the contents through using copyfileobj...otherwise if you cat a large file, it waits for everything to be downloaded before printing anything out.

On a related note, we should support 'head' and 'tail' in addition to 'cat'.  Add as another PR.

codalab/lib/zip_util.py
```diff
@@ -85,3 +91,78 @@ def unpack(ext, source, dest_path):
         finally:
             if delete_source:
                 path_util.remove(source)
+
+
+def pack_files_for_upload(sources, should_unpack, follow_symlinks, exclude_patterns):
+    """
+    Create a single flat tarfile containing all the sources.
+
+    Caller is responsible for closing the returned file-like object.
```
percyliang: say fileobj explicitly

codalab/lib/zip_util.py
```diff
@@ -85,3 +91,78 @@ def unpack(ext, source, dest_path):
...
+            resolved_source = os.path.realpath(source)
+            if not os.path.exists(resolved_source):
+                raise UsageError('Broken symlink')
+        elif os.path.islink(source):
+            raise UsageError('Not following symlinks.')
+        return resolved_source
+
+    sources = map(resolve_source, sources)
+
+    # For efficiency, return singular files and directories directly
```
percyliang: singular ==> single

codalab/worker/bundle_manager.py
```diff
@@ -111,6 +112,10 @@ def _stage_bundles(self):
         bundles_to_fail = []
         bundles_to_stage = []
         for bundle in bundles:
+            # Ignore non-derived (i.e. executable) bundles
+            if not isinstance(bundle, DerivedBundle):
```
percyliang: non-derived is not-executable (comment is ambiguous)

Do we need this statement?  Edit: I understand why now, but will not be necessary if we have an 'uploading' state.

test-cli.py
```diff
@@ -742,65 +769,93 @@ def stress(use_time, request_time, use_memory, request_memory, use_disk, request
...
+
+        # Restore original config
+        os.environ['CODALAB_HOME'] = old_home
+
+        # Switch to new host and log in to cache auth token
+        subprocess.call([cl, 'logout', remote_worksheet[:-2]])
+        os.environ['CODALAB_USERNAME'] = test_username
+        os.environ['CODALAB_PASSWORD'] = test_password
+        subprocess.call('cl work ' + remote_worksheet, shell=True)
+        subprocess.call('cl gls > /dev/null', shell=True)
```
percyliang: Why this command?

test-cli.py
```diff
@@ -742,65 +769,93 @@ def stress(use_time, request_time, use_memory, request_memory, use_disk, request
 @TestModule.register('copy')
 def test(ctx):
     '''Test copying between instances.'''
-    # Figure out the current instance
+    # Figure out the current instance and use it as the 'remote' instance
```
percyliang, line 778: Cool, testing on remote is good if we're going to deprecate local::

test-cli.py
```diff
@@ -742,65 +769,93 @@ def stress(use_time, request_time, use_memory, request_memory, use_disk, request
...
+    remote_worksheet = 'http://localhost:3800::'
+
+    # Start auxiliary servers for the new auxiliary host
+    os.environ['PYTHONUNBUFFERED'] = '1'  # prevents stalling when waiting for output
+    server_proc = subprocess.Popen([cl, 'server'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    rest_server_proc = subprocess.Popen([cl, 'rest-server'], stderr=subprocess.PIPE)
+
+    # TODO(sckoo): refactor test-cli to allow each module to define cleanup method
+    try:
+        wait_until_substring(server_proc.stdout, 'BundleRPCServer serving')
```
percyliang: Write a comment to get rid of BundleRPCServer later

percyliang:
cl upload with multiple files doesn't work, I get:
  File "/home/pliang/research/codalab/codalab-cli/codalab/lib/zip_util.py", line 147, in should_exclude
    return any(fnmatch(fn, p) for p in exclude_patterns)
TypeError: 'NoneType' object is not iterable

percyliang:
- Having the bundle show up before uploading is complete is great.  I wonder if it's good to have a new state 'uploading' rather than leaving things as 'created'?  It is easier from the user perspective.
- Note: from the website, however, the bundle doesn't show up until uploading finishes (and there's no progress bar).  I guess that's another PR...

percyliang:
LGTM modulo comments (mostly asking for documentation and fast upload)

kashizui:
@percyliang: I can get 60+ mbps if I disable on-the-fly compression. So it becomes a tradeoff depending on whether or not the network is a bottleneck for the user. `worksheets-test` for example seems to be capped at 3mbps for upload. Perhaps we should make it an option to the user whether or not to always force compression?

percyliang:
@kashizui do you mean 60 MB/sec?

Is this the situation?
localhost: compression is 60, no compression is 6
wstest: compression is ?, no compression is ?
Can't we just check if we're connecting to localhost and switch between the two?

Also, I'm really surprised compression means a factor of 10x - what is going on there?

kashizui:
Right, to recap:
localhost: compression is 6 MB/s, **no** compression is 60 MB/s
wstest: compression is 700 KB/s, **no** compression is also 700 KB/s

Check if we're connecting to localhost and/or stanford.edu? (I guess that's also localhost with the ssh tunnel..)

I'll look into why gzip might be 10x slower, definitely doesn't seem right given it's standard on a lot of HTTP services. It may be that we're using the command-line utility that's writing to a pipe, from which we're streaming the compressed data. That's a lot of read and writes that may or may not be avoidable.

kashizui:
Using zlib directly, I'm still getting a maximum of 6 MB/s at the default compression level (6). Using the fastest compression level (1) I can get up to 20+ MB/s but not any faster than that. The data doesn't get compressed very much though.

percyliang:
> wstest: compression is 700 KB/s, no compression is also 700 KB/s
> It seems like compression is just strictly worse than no compression (on
> wstest it's equal, and on localhost, it's worse).  Is there any reason to
> have it then?

kashizui:
Yes, exactly in the case where the network speed is the bottleneck,
compression is better, because you can end up transferring 60 MB rather
than 160 MB (as was the case for the glove vectors I was uploading as a
test.)

percyliang:
Hm, I guess minimally it should be an option (can it just be client-side)?

kashizui:
Yeah I agree, should minimally be an option. The question is whether compression should be default, or whether there's a reasonable heuristic to choose compression automatically.

Yes fortunately this decision is completely up to the client: the API accepts both compressed and uncompressed and can handle them appropriately.

codalab/client/json_api_client.py
kashizui: I will document the arguments here, but more importantly the query parameters are already documented in `rest/bundle.py`.

codalab/lib/bundle_cli.py
kashizui: I thought it would be hard but I just did it :)

test-cli.py
kashizui: Just the first rest-ified command I could think of. I'll give a shot at fixing the double-login thing now, but if it doesn't go well I'll make sure to document this line better.

kashizui:
@percyliang Addressed all comments. Significant changes:
- Fixed double login issue
- Added --force-compression flag, with no compression on single files as the default, but included note in documentation that using the flag can help with uploads over slow networks.
- Fixed multiple-upload bug and added CLI test for that case.
- Added a new UPLOADING bundle state
- Print progress for downloads as well

kashizui:
@percyliang If you think these decisions are OK, we are ready to merge.

percyliang:
This is great - thanks!  Let's merge!

PR #438, opened 2016-05-18
Fix inconsistency bug.

klopyrev:
Small hack to work around inconsistencies in NFS. It seems that file contents stay consistent. Where NFS is not consistent is in directory contents. There, it keeps a cache. Thus, any new directories created in the bundles/ directory don't appear for a bit. This change adds a loop that waits for the directory to appear.

Note: there could still be inconsistencies in the files seen inside the bundle when measuring the size of the bundle. Newly created files may not appear. However, that occurs only if the bundle server read the contents of the directory before (i.e. it has them cached) and that was very recently. Thus, it's not very likely that that would happen, unless the user is refreshing the page constantly to see the files in his running bundle. Even if it does happen, unless the running bundle creates a gigantic file very quickly the resulting bundle size measurement shouldn't be off by much. Data hash would obviously be very wrong, but we're not even using that anymore.

@percyliang

### DIFF
--- a/codalab/rest/worker.py
+++ b/codalab/rest/worker.py
@@ -4,6 +4,7 @@ import httplib
 import json
 import os
 import subprocess
+import time
 
 from bottle import abort, get, local, post, put, request, response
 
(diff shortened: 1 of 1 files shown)


percyliang:
LGTM

PR #440, opened 2016-05-19
Fallback to sysctl when os.sysconf('SC_PHYS_PAGES') fails on OS X

kashizui:
Allows worker to run on OS X.

Referenced this thread for solution: https://github.com/BD2KGenomics/toil/issues/587

@klopyrev Not sure what you think is the best style here (would you prefer `_get_memory_bytes` be defined somewhere else)?

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1,4 +1,5 @@
 from contextlib import closing
+from subprocess import check_output
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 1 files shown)


worker/worker.py
```diff
@@ -91,7 +92,14 @@ def _should_run(self):
         with self._runs_lock:
             if self._runs:
                 return True
-        return False 
+        return False
```
kashizui, line 95: Sorry about this, my editor automatically strips whitespace from the ends of lines.

klopyrev:
LGTM, thanks for fixing this.

Issue #104, opened 2015-08-12
make sure that run -t, wait -t work


percyliang:
Currently, some output gets cut off.  Make sure we preserve the full contents.

percyliang:
Can't seem to reproduce it (on codalab.org)...

kashizui:
Is this still a problem?

percyliang:
Last I checked it seemed okay, but can you double check?

The cases to worry about is when you call cl at different points during the
execution of the run and if the run outputs a lot of text (especially in a
bursty way).

percyliang:
Seems to work, closing for now.

PR #442, opened 2016-05-19
Refresh bundles automatically

ChengshuLi:
Fixes codalab/codalab-worksheets#104

@percyliang

### DIFF
--- a/codalab/rest/legacy.py
+++ b/codalab/rest/legacy.py
@@ -1,5 +1,5 @@
 """
-Legacy REST APIs moved from the codalab-worksheets Django REST server. 
+Legacy REST APIs moved from the codalab-worksheets Django REST server.
 """
 import base64
 from contextlib import closing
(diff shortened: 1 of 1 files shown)


codalab/rest/legacy.py
```diff
@@ -208,10 +211,10 @@ def get_and_update_summary(download_manager, item, name):
                 with info_lock:
                     info[name] = result
 
-            read_threads = []            
+            read_threads = []
             for item in info['contents']:
                 name = item['name']
-                if name in ['stdout', 'stderr'] and (item['type'] == 'file' or item['type'] == 'link'):
+                if name in ['stdout', 'stderr'] or (item['type'] == 'file' or item['type'] == 'link'):
```
percyliang, line 217: why this change?

percyliang:
@ChengshuLi could you explain what this PR is doing?

ChengshuLi:
@percyliang 
I want to periodically update run bundle, including all the embedded file, so I need to fetch the file contents for that.

Here is an example.

% schema foo
% add myoutput  /output
% display table foo
[run run-while : while true; do echo 1 >> output; sleep 0.2; done]

In this case, I need to update the /output file in the bundle so I need to fetch its content

percyliang:
Did you check that there's not a performance hit?

codalab/rest/legacy.py
```diff
@@ -184,7 +187,7 @@ def get_summary(download_manager, info, name):
                     '\n')
                 contents = download_manager.summarize_file(
                     uuid, name,
-                    num_head_lines=50, num_tail_lines=50, max_line_length=128,
+                    num_head_lines=1000, num_tail_lines=0, max_line_length=128,
```
klopyrev, line 190: Why are you doing this?

klopyrev, line 190: It's going to be a lot of data for a running bundle.

ChengshuLi:
Close comment: I completely re-did this PR.

PR #447, opened 2016-05-20
[DEFUNCT] Create JsonApiClient in CodaLabManager

kashizui:
Split from #432

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -147,6 +147,7 @@ class CodaLabManager(object):
         self.state = read_json_or_die(self.state_path)
 
         self.clients = {}  # map from address => client
+        self.rest_clients = {}
 
     def init_config(self, dry_run=False):
         '''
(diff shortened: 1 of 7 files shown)


kashizui:
@klopyrev

codalab/lib/codalab_manager.py
```diff
@@ -497,6 +517,69 @@ def client(self, address, is_cli=True):
     def cli_verbose(self):
         return self.config.get('cli', {}).get('verbose')
 
+    def get_access_token(self, address, auth_handler):
```
klopyrev, line 555: I don't think you need to copy this code. Here's what I propose:

1) Get rid of the BundleClient login method entirely. It doesn't need to exist, and all it does is just call get_access_token.
2) Change authenticate to take in an address and auth_handler.
3) Use a lambda in both places where _authenticate needs to be called to pass in the address and auth_handler.

Am I missing something?

codalab/lib/codalab_manager.py
```diff
@@ -454,23 +455,42 @@ def root_user_id(self):
...
         '''
-        if address in self.clients:
-            return self.clients[address]
-        # if local force mockauth or if local server use correct auth
-        if is_local_address(address):
+        client_cache = self.rest_clients if use_rest else self.clients
+        if address in client_cache:
+            return client_cache[address]
+        if use_rest:
+            # FIXME(sckoo): Temporary hack
```
klopyrev: Let's set up things like we discussed in our chat.

codalab/lib/codalab_manager.py
kashizui, line 555: The terrible part is when client is a RemoteBundleClient:
RemoteBundleClient.login() actually makes an RPC call to the LocalBundleClient.login() on the server, to get the token from the auth handler on the server. I wasn't sure how to redesign this without wasting too much time on code that will be defunct soon anyway.

kashizui, line 555: I was thinking exactly what you were thinking, but ran into this issue.

kashizui, line 555: The main barrier to solving this was that the auth handler needed the address of the REST server (since that's where the oauth endpoints are defined) in order to contact the server directly for authentication (rather than through this ugly RPC call). But maybe we can use the same solutions we just discussed on Slack?

kashizui, line 555: I'm not sure this is worth fixing at this point.

kashizui:
@klopyrev 

- Introduced hardcoded translations for REST API address
- I'm going to leave the authentication procedure as it is if that's OK, since I think it will be more work than it's worth to rewrite stuff to work better with RemoteBundleClient and LocalBundleClient.

PR #449, opened 2016-05-23
Use a different way of disabling network in Docker container.

klopyrev:
This way is as far as I can tell the same as the old way where --net=none was specified to docker run. gethostbyname seems to work with this way, but not the way I had it before.

@percyliang


percyliang:
LGTM

PR #450, opened 2016-05-24
Get dependencies working properly for the NLP cluster.

klopyrev:
It seems that Docker can't modify the bundle path to add the special dependency files on the NLP cluster. Work around this by mounting dependencies in a different location inside the container and adding symlinks to point to that container. Also, fix remove_path to work with symlinks.

@percyliang


percyliang:
LGTM

PR #451, opened 2016-05-24
Always route get_target_info calls to the worker.

klopyrev:
Change the routing for get_target_info on the NLP cluster. See comment in code for reason. This should make it appear that the run starts faster and the dependencies won't appear as symlinks like: b1 -> /0xasdfasdfad_dependencies/b1.

@percyliang


percyliang:
LGTM

PR #454, opened 2016-05-26
Refactor _authenticate to support the new REST changes.

klopyrev:
Avoids copied code.

@kashizui


kashizui:
Wow I never thought of using duck typing in this way... Awesome.
LGTM

PR #455, opened 2016-05-26
Clean-up the duplicate creation of symlinks when the Docker image needs downloading

klopyrev:
Since create is called twice when the image is not present, there was a call to create the symlink twice which causes a File exists error.

Note, this design makes more sense, since we create the symlinks in the same file as we delete them.


klopyrev:
@percyliang

percyliang:
LGTM

PR #456, opened 2016-05-26
Fix bug in deleting the data of dependencies.

klopyrev:
Fix bug that deletes the data of dependencies, even if recursive is not specified.

@percyliang 

FYI: relevant_data_hashes is not used, so I just deleted those lines.


percyliang:
LGTM

PR #457, opened 2016-05-26
Check for correct ack in worker model.

klopyrev:
Actually check for the correct ACK string on the worker socket. On older Mac OS (as demonstrated by Eric), the socket implementation is so broken spurious bytes are sent in addition to spurious connections being accepted.

There's probably something that I'm missing about all this, but this implementation is reliable enough.

@percyliang


percyliang:
LGTM

PR #459, opened 2016-05-26
Check permission of dependencies.

klopyrev:
Check that the user has permission to read a bundle that the user depends on.

@percyliang @kashizui


kashizui:
LGTM

PR #463, opened 2016-05-31
Keep track of maximum amount of memory used by a run.

klopyrev:
This is useful for figuring out how much memory a run used at its maximum point.

@percyliang


percyliang:
cool - LGTM

PR #465, opened 2016-06-01
Fix uedit

### LINKED ISSUE #464: cl uedit bug
`cl uedit -u <user> -d 300g` either is an apparent no-op or throws a Duplicate Row exception.

In the no-op case, what's happening is that a new user row gets added to the DB where user_id is actually the username, which is a bug.

This probably happened when we ported the DBs over.

kashizui:
Fixes #464 

@percyliang

### DIFF
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -1339,7 +1339,8 @@ class LocalBundleClient(BundleClient):
     def update_user_info(self, user_info):
         user_id = self._current_user_id()
         is_root = (user_id == self.model.root_user_id)
-        is_user = (user_id == user_info['user_id'])
+        if 'user_id' not in user_info:
+            user_info['user_id'] = user_id
         if is_root:
             # TODO: in the future, allow user to update, but only in limited ways
             self.model.update_user_info(user_info)
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2293,12 +2293,15 @@ class BundleCLI(object):
         Edit properties of users.
         """
         client = self.manager.current_client()
-        user_info = client.get_user_info(args.user_id)
+        user_update = {}
         if args.time_quota is not None:
-            user_info['time_quota'] = formatting.parse_duration(args.time_quota)
+            user_update['time_quota'] = formatting.parse_duration(args.time_quota)
         if args.disk_quota is not None:
-            user_info['disk_quota'] = formatting.parse_size(args.disk_quota)
-        client.update_user_info(user_info)
+            user_update['disk_quota'] = formatting.parse_size(args.disk_quota)
+        if args.user_id is not None:
+            user_update['user_id'] = args.user_id
+        if user_update:
+            client.update_user_info(user_update)
 
     @Commands.command(
         'reset',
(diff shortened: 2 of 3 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -2293,12 +2293,15 @@ def do_uedit_command(self, args):
...
-        user_info = client.get_user_info(args.user_id)
+        user_update = {}
         if args.time_quota is not None:
-            user_info['time_quota'] = formatting.parse_duration(args.time_quota)
+            user_update['time_quota'] = formatting.parse_duration(args.time_quota)
         if args.disk_quota is not None:
-            user_info['disk_quota'] = formatting.parse_size(args.disk_quota)
-        client.update_user_info(user_info)
+            user_update['disk_quota'] = formatting.parse_size(args.disk_quota)
+        if args.user_id is not None:
```
percyliang, line 2301: why would user_id be None?

codalab/client/local_bundle_client.py
```diff
@@ -1339,7 +1339,8 @@ def get_user_info(self, user_id, fetch_extra=False):
     def update_user_info(self, user_info):
         user_id = self._current_user_id()
         is_root = (user_id == self.model.root_user_id)
-        is_user = (user_id == user_info['user_id'])
+        if 'user_id' not in user_info:
```
percyliang, line 1342: why would user_id be None?  If it is, how does it make sense to update?  Can you explain what the problem was before that's getting fixed?

codalab/lib/bundle_cli.py
kashizui, line 2301: `-u` is currently specified as an optional command-line argument, and I'm assuming that we want to update the current logged-in user if not specified. Should we make it required?

percyliang, line 2301: It can be optional, but then current_user_id needs to be put in user_update, no?

codalab/client/local_bundle_client.py
kashizui, line 1342: If `user_id` is not specified, we assume that it is the current user.

The original problem was that `LocalBundleClient.get_user_info()` would create a new user with the given user_id if one did not exist yet.  (This is especially problematic if the username is specified at the command-line instead of the user ID, creating a user with that username as the user ID.) This is no longer necessary after the user DB migration so I've removed the `get_user_info()` call.

codalab/lib/bundle_cli.py
kashizui, line 2301: The bundle_cli doesn't have direct access to the current_user_id, so I'm leaving that to the `update_user_info` function on the server side.

kashizui, line 2301: This also prevents forgery, though we currently only allow root to changes things anyway.

kashizui:
We also don't currently support specifying users by username, only user ID, which is inconvenient. This is made easier by the REST API and I will include this change in the upcoming REST patches. Created #466

codalab/client/local_bundle_client.py
percyliang, line 1342: Got it.

codalab/lib/bundle_cli.py
percyliang, line 2301: Eventually, we want users to be able to change some of their own fields like `email`.  Just not `disk_quota`.

kashizui, line 2301: I'll include that in my new issue 👍

kashizui:
Should be merged after #468

PR #470, opened 2016-06-06
Fix copy bundle.

klopyrev:
1) Data hash was removed as a construct argument for upload bundles (i.e. program and dataset).
2) Don't generate a new UUID when a bundle is copied. Note, the calling code also checks that the UUID doesn't exist.

@percyliang


percyliang:
LGTM

PR #471, opened 2016-06-06
Send less load from workers to REST server.

klopyrev:
1) Check-in blocks on worker socket for 5 seconds, instead of 2. That makes shutting down the worker longer, since it needs to wait for the check-in call to return. Yet, it sends less traffic to the REST server. Responsiveness should be the same or better.
2) Report resource utilization every 5 seconds, instead of every half a second.
3) Minor changes to debug log messages.

@percyliang


percyliang:
LGTM

PR #472, opened 2016-06-06
Move install_dependencies logic into bundle manager

klopyrev:
With the new worker system this was the last place install_dependencies was used. It will be removed in a future PR.

This also fixes: cl make passwd:0x????????//etc/passwd

@percyliang


percyliang:
LGTM

PR #476, opened 2016-06-06
Remove the oauth handler that was used when authentication went through the Django server.

klopyrev:
@percyliang


codalab/lib/codalab_manager.py
```diff
@@ -403,11 +403,9 @@ def auth_handler(self, mock=False):
 
         if mock or handler_class == 'MockAuthHandler':
```
percyliang, line 404: Can we get rid of this too, since now you have to connect to the rest server to do anything, so MockAuthentation doesn't seem like it's useful to have.

percyliang:
LGTM

codalab/lib/codalab_manager.py
klopyrev, line 404: It's still used by local cl commands, like cl events.

PR #480, opened 2016-06-07
Switch CircleCI tests to use Docker.

klopyrev:
Includes a few changes:
1) Switch to the new worker system.
2) Actually use Docker and enable the 2 disabled tests. Unfortunately, the out of memory test needs to be disabled.
3) Use MySQL. The new worker system doesn't work too well with Sqlite3 due to concurrency issues.
4) Have the tests use a remote connection (RemoteBundleClient), instead of local. This also enables the copy test.
5) Fix the copy test. Copying bundles without data wasn't working.
6) Rewrite the stress test in Perl. I want to use a slim Docker image since downloading the image in the test takes time. The ubuntu:14.04 image is good, but it doesn't have Python.


percyliang:
LGTM

klopyrev:
I haven't actually finished this PR. Just testing for now.

klopyrev:
@percyliang This is ready now. Take a look.

scripts/stress-test.pl
```diff
@@ -0,0 +1,24 @@
+use constant BYTES_IN_MB => 1024 * 1024;
```
percyliang, line 1: Why rewrite in perl?

percyliang:
LGTM for real

scripts/stress-test.pl
klopyrev, line 1: It's in the description. Basically, so I can use the ubuntu:14.04 image for the tests, and not download some other much larger image.

PR #481, opened 2016-06-07
Get single dependency make working again.

klopyrev:
The old code handled this case in a very hidden way, so I missed it.

@percyliang


percyliang:
LGTM
can you make sure ./test-cli.py still passes?

PR #487, opened 2016-06-09
Work around NFS cache issues.

klopyrev:
@percyliang 

This change makes it take longer to start, but when it finishes running things work better.

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -329,6 +329,12 @@ class BundleManager(object):
         """
         if self._model.set_starting_bundle(bundle, worker['user_id'], worker['worker_id']):
             workers.set_starting(bundle.uuid, worker)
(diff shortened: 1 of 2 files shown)


codalab/worker/bundle_manager.py
```diff
@@ -329,6 +329,12 @@ def _try_start_bundle(self, workers, worker, bundle):
         """
         if self._model.set_starting_bundle(bundle, worker['user_id'], worker['worker_id']):
             workers.set_starting(bundle.uuid, worker)
+            if self._worker_model.shared_file_system and worker['user_id'] == self._model.root_user_id:
```
percyliang, line 332: Why does the user have to be root here?
This doesn't parallel the code in run.py.

klopyrev, line 332: Only root-owned workers run with a shared file system. This excludes user-owned workers.

percyliang:
LGTM

PR #492, opened 2016-06-11
(Yet another) attempt to get GPUs working. Doesn't work on jagupard3, which was down when I tested the last time.

klopyrev:
This set up is more consistent with how libcuda is set up natively. It helps a bit on jagupard3 given that libcuda.so is now correctly loaded (https://codalab.stanford.edu/bundles/0x7828b37fb2ff4b3aa803bec2eaf0baf0/), but it has other problems with GPUs. Hopefully, with this more correct set up libraries that I haven't tested with are more likely to work correctly.

@percyliang

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -66,28 +66,19 @@ On Mac, DOCKER_HOST and optionally DOCKER_CERT_PATH should be defined. You need
 to run the worker from the Docker shell.
 """
             raise
-        
-        # Find libcuda. We use ldconfig to find directories where libcuda is
-        # stored and then pick up all libcuda.* files. If we use just ldconfig
-        # we miss files like libcuda.so.352.63.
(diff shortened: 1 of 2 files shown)


percyliang:
There's been a lot of tweaks to this GPU code...can you explain a bit more what you're trying to accomplish and what settings this works for and what settings it doesn't?

klopyrev:
Simply trying to replicate the CUDA set up inside the Docker container as much as possible. Works on all jagupard machines https://codalab.stanford.edu/worksheets/0x988ce7fbfcc3493d8e8e201a43a77755/, except for jagupard3. You tell me where it doesn't work. For more information, see https://github.com/NVIDIA/nvidia-docker and the report I sent you a while back.

percyliang:
LGTM

PR #493, opened 2016-06-11
Ignore maximum resource limits on user-owned workers.

klopyrev:
@percyliang

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -381,6 +381,11 @@ class BundleManager(object):
 
         # Parse |request_string| using |to_value|, but don't exceed |max_value|.
         def parse_and_min(to_value, request_string, default_value, max_value):
+            # On user-owned workers, ignore the maximum value. Users are free to
+            # use as many resources as they wish on their own machines.
+            if worker['user_id'] != self._model.root_user_id:
+                max_value = None
+            
             # Use default if request value doesn't exist
             if request_string:
                 request_value = to_value(request_string)


codalab/worker/bundle_manager.py
```diff
@@ -381,6 +381,11 @@ def _construct_run_message(self, worker, bundle):
 
         # Parse |request_string| using |to_value|, but don't exceed |max_value|.
         def parse_and_min(to_value, request_string, default_value, max_value):
+            # On user-owned workers, ignore the maximum value. Users are free to
+            # use as many resources as they wish on their own machines.
+            if worker['user_id'] != self._model.root_user_id:
```
percyliang, line 386: We still need a limit - users don't want to get their machines accidentally hosed, so I think --max-memory should still take effect.

klopyrev, line 386: There are really low limits set up on worksheets.codalab.org. For example, the memory limit is 10gb. This limit basically disallows running any jobs on worksheets.codalab.org that use more than 10gb, even on a user-owned machine. That doesn't seem right. If a user wants to run a job that uses 100gb, we should allow it on their own machine.

percyliang, line 386: I think the solution is to have people just specify `--max-memory`.  The run creation interface can be improved to make this more apparent.  But resource limits are resource limits; if the user machine has higher resource limits, then this can be exploited by having a larger --max-memory.  Just ignoring resource constraints doesn't seem good.

klopyrev, line 386: You definitely need to spend more time trying to understand the code that is sent your way. You definitely can ask for explanations if you don't have time to understand the code. It's a bit frustrating when you assume that the code works in a certain way and give comments based on that assumption, when even spending another minute or two would show that things are different than you think.

There are 3 limits:
1) A user specified limit, controlled with the --request-<resource-type> flag. This is what I think you you mean by --max-memory in your comment.
2) A default limit that is specified in the CodaLab config. In fact, this isn't used anywhere. There are no defaults.
3) A maximum limit that is specified in the CodaLab config. The user can't use more than this maximum, even if they specify more with the --request-<resource-type> flag. worksheets.codalab.org has maximums specified in codalab-worksheets/codalab/config/templates/bundle_server_config.json.

What this change is doing is making the maximum not apply when users are running bundles on their own machines. It doesn't make sense to enforce this maximum. If a user does want to specify a limit, they can still do that with the --request-<resource-type> flag. Since no defaults are actually set in our configs, if they user doesn't specify a limit after this change there will be no limit on resource usage on user-owned workers. Does that make sense?

percyliang, line 386: Thanks for the explanation.

percyliang:
LGTM

PR #510, opened 2016-07-03
Fix kill cli test

kashizui:
Originally would fail if bundle took too long to start running,
or if the 'actions' field contained Unicode strings (which
is the case for older versions of the bundle service).

@percyliang


kashizui:
Oops I didn't need this.

PR #529, opened 2016-07-11
Add color to setup.sh

kdexd:
This PR introduces colored output by `setup.sh` for better view of progress. The final result looks like this:
image


kdexd:
I am unable to figure out the reason of build failures. It works fine locally:
image

kashizui:
Thanks for your contribution! I have no idea why the tests are failing on your PR. I'll take a closer look at it when I get the chance...

kashizui:
The tip of master was actually broken earlier today (though with a different error than what you were getting), so you may want to rebase on the latest changes on master.

kdexd:
@kashizui Sure I'll try a rebase. Also, if I make future PRs, shall I base them on `develop` or `master` ?  
(I usually adopt `master-develop` branching infrastructure for my own projects as well :smile: )

kdexd:
Looks like the builds fail due to my changes, and only my changes in this PR.
image
I cannot reproduce this failure locally, everything builds successfully. I'll try some manual debugging.

kdexd:
Hi @kashizui,
The mechanism I used earlier ( using `tput` ) works only when `stdout` is directed to a bash terminal, which isn't the case with Circle CI builds I suppose. Now I am using ANSI escape characters for the color, and to escape them properly, output is done by `echo -e`.
Builds are restored now.

kdexd:
Circle CI output is also colored :smiley:
image

kashizui:
Great, this looks very cute! And yes, it would be great if you could actually base this off of `develop`, then we can go ahead and merge it in.

kdexd:
@kashizui I think once a PR is made, I will not be able to change the HEAD of it, this change will still go into `master` which includes 13 extra commits unrelated to the purpose of this PR. I usually do not believe in committing anything to master directly. All changes are good to be merged to master through `develop`, but this is a small enhancement which is not likely going to break anything on master.

I will take care of branching out from `develop` and choosing merge HEAD to be `develop` from next PRs, but in my opinion, due to 13 extra commits of `develop`, this is best to be based on `master`. Although if you are fine with the current situation, you can surely merge ! Do let me know if I should rebase it back to `master`.

kashizui:
Please rebase on `develop` and submit a new PR that is based off of `master`. I will close this PR for now, sorry for the inconvenience! We're working on developing a better policy for development so that we can make `develop` the default branch in the future.
By the way it looks like the tests are failing now because you've rebased on an outdated tip of `develop`.

kdexd:
@kashizui So here it is:
One commit in a branch out of `develop`, to be merged back into `develop`. This is fine, right ?

kashizui:
You will need to create a new PR, unfortunately GitHub doesn't allow you to change the target branch of a PR once you've created it.

kdexd:
Yes right. I will create one in a moment.

PR #530, opened 2016-07-11
Allow commas to be used in API parameters

ppasupat:
Commas are used in some cl commands (e.g., `cl search owner=tom,jerry`). Currently commas are disallowed since they are used to join parameters in the API request. The following escape scheme is introduced to solve the problem:
- Escape `,` as `\C` and `\` as `\B`
- Join the strings as usual with `,`

There is no ambiguity: in the resulting string, `,` is only used to separate items, and `\` can only be followed by `B` or `C`. The conversion can also be done using simple string replacement.

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -108,16 +108,21 @@ class JsonApiClient(RestClient):
         if params is None:
             return None
 
+        def _escape_param(param):
+            if isinstance(param, bool):
+                return int(param)
+            if isinstance(param, (int, long, float)):
+                return param
+            if isinstance(param, (str, unicode)):
+                return unicode(param).replace('\\', '\\B').replace(',', '\\C')
+            raise NotImplementedError("Data type {} is not supported.".format(type(param)))
+
         result = {}
         for k, v in (params.iteritems() if isinstance(params, dict) else params):
             if isinstance(v, list):
-                result[k] = ','.join(
-                        unicode(x).replace('\\', '\\B').replace(',', '\\C')
-                        for x in v)
-            elif isinstance(v, bool):
-                result[k] = int(v)
+                result[k] = ','.join(unicode(_escape_param(x)) for x in v)
             else:
-                result[k] = v
+                result[k] = _escape_param(v)
         return result
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


kashizui:
Thanks Ice! By the way what's the inspiration for your choice of encoding scheme?

Also, could you update `test_pack_params` in `tests.client.json_api_client_test.JsonApiClientTest` to test your new feature? The tests currently expect `NotImplementedError` to be thrown.

kashizui:
Sorry, the build error is actually because the tip of master was broken, I just pushed a fix so it should work after you rebase again on master.

ppasupat:
I edited the tests. I had to do a similar encoding for putting a list in a TSV file before, so I just ported that idea over.

percyliang:
While Ice's scheme is simple, it is non-standard, which is bad I think if people want to use our REST API.  I think it's perhaps better to use another way to encode a list rather than using commas.  What about encoding the list as JSON?

kashizui:
This might be a good reference for best practices (I should have read this earlier):
http://stackoverflow.com/questions/2602043/rest-api-best-practice-how-to-accept-list-of-parameter-values-as-input

The correct way to handle multiple values is to use the same query parameter key multiple times:

```
GET /rest/bundles/?specs[]=bundle1&specs[]=bundle2
```

and then on the server-side, use the `getall` method:

```
print request.query.getall('specs[]')
>> ['bundle1', 'bundle2']
```

(The extra `[]` at the end of the parameter key seems to be convention, though we may or may not choose to use it. A potential problem with it is that it actually has to be URL-encoded to be `%5B%5D`, which is kind of ugly, but it's nice that it makes explicit that this parameter key expects potentially multiple values.)

ppasupat:
I agree with Percy. I will try to implement the best practice and submit a new pull request.

kashizui:
@ppasupat By best practice do you mean JSON-encoding or the redundant query parameters described in the StackOverflow post?

ppasupat:
Using redundant query parameters seems to be the standard, and this can be done by returning a list of pairs (e.g., `[('key', 'a'), ('key', 'b'), ...]`) instead of a dict -- `urllib.urlencode` will properly convert it to `key=a&key=b...`. 

Since most of the times parameters are not lists, how about we use the version without `[]`?

kashizui:
I'm good with that!

PR #536, opened 2016-07-15
Add items only after bundle copied during wadd

### LINKED ISSUE #504: Improve robustness around invalid bundle references in worksheet items
When `cl wadd` is run and Ctrl-C is pressed, this results in an invalid bundle being created.

kashizui:
Fixes #504 

This prevents items with invalid bundle references from being created.

@percyliang

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2429,14 +2429,17 @@ class BundleCLI(object):
         dest_worksheet_info = dest_client.get_worksheet_info(dest_worksheet_uuid, True)
         dest_items = [] if args.replace else dest_worksheet_info['items']
 
-        # Save all items.
-        dest_client.update_worksheet_items(dest_worksheet_info, dest_items + source_items)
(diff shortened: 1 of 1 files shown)


percyliang:
@kashizui Copying the items before the bundle was deliberate so that people could see (and edit) the metadata before the contents arrive (just like how we're handling uploading now).  I think we should just have more robust logic to handle invalid bundle references in general, which could arise from other situations too.

kashizui:
@percyliang Got it-- what are your initial thoughts on what this more robust logic would look like, e.g. in the case of a canceled wadd?

percyliang:
In general, any reference to an invalid uuid should just show up gracefully
as <invalid>, which seems to already happen.

 -Percy

kashizui:
I see, so is the issue is void then?

percyliang:
I think it sometimes crashes the page though, especially when we have table
schemas (we're just not very consistent about it).  Need better tests!

PR #541, opened 2016-07-25
Accept user name and affiliation while sign up.

kdexd:
This PR is in conjunction to codalab/codalab-worksheets#201 and aims to fix codalab/codalab-worksheets#200 . Changes here reflect acceptance of first and last names, affiliation while sign up and populate the database accordingly.

### DIFF
--- a/codalab/rest/account.py
+++ b/codalab/rest/account.py
@@ -66,8 +66,11 @@ def do_signup():
     success_uri = request.forms.get('success_uri')
     error_uri = request.forms.get('error_uri')
     username = request.forms.get('username')
-    password = request.forms.get('password')
     email = request.forms.get('email')
+    first_name = request.forms.get('first_name')
+    last_name = request.forms.get('last_name')
+    password = request.forms.get('password')
+    affiliation = request.forms.get('affiliation')
 
     errors = []
     if request.forms.get('confirm_password') != password:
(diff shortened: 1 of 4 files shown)


kdexd:
The tests were an expected failure. I'll modify the method calls accordingly to make the tests pass.

kashizui:
This looks good, thanks!
 This isn't a big deal, but could you double-check what happens if the user doesn't fill out the optional Affiliation field? I would rather it stores None/NULL rather than empty string.

kdexd:
Since the field exists and is not set, it returns an empty string instead of null value. A null value is obtained on GET response if no such input field exists. I'd probably add a check that:

``` python
if affiliation and len(affiliation) is 0:
    affiliation = None
```

kashizui:
That would be great, and perhaps prefer `if not affiliation:` for conciseness?
Otherwise LGTM, will merge after update.

kdexd:
Oh yes ! I just realized that there are many other quantities `foo` which evaluate to `True` for `not foo` :smile: 
image

kashizui:
Thanks for your contribution, I will test and merge this later this week.

kdexd:
That's great - this was a very easy fix hence I could intuitively solve it by tracing back the origins of certain method calls. I am halfway through the server setup and hopefully that should be done by tomorrow, I'll be able to fiddle with a little bigger issues in my own development environment.

This platform is amazing, I am continuing this bug hunt for sometime more, cheers ! :smiley:

codalab/rest/account.py
```diff
@@ -101,7 +108,9 @@ def do_signup():
         })
```
kashizui, line 107: You need to provide the new fields to this redirect call as well. This allows the sign up page to display the already-filled values again when it displays an error. (If you test it, you will see that only email and username currently gets filled out again after an error, such as mismatched passwords, occurs.)

kashizui:
Functionality looks fine while testing, except for my one comment. Please squash your commits into a single commit when you're ready to merge.

kdexd:
I guess this PR has no more additions pending. :smile:

Issue #265, opened 2015-11-26
cl edit should support tags


percyliang:
Currently, we can tag worksheets by using `cl wedit --tags`.

Do the same thing for `cl edit --tags`.

Look in `bundle_cli.py`.

kashizui:
Fixed by #543

PR #543, opened 2016-07-30
Add --tags option in cl edit, similar to cl wedit.

kdexd:
This PR introduces a quick fix to #265. Mechanism of adding tags in `cl edit` is similar to `cl wedit`. This new argument is tested as well in `test-cli.py`.


kashizui:
LGTM, thanks!

Issue #515, opened 2016-07-04
Prepend 0x to user ids


kashizui:
Should prepend 0x to user ids, since they are hex. I accidentally omitted it when migrating the users from Django.
This may be tricky to do in production because user ids are used as foreign keys.

kashizui:
http://stackoverflow.com/questions/13606469/cannot-change-column-used-in-a-foreign-key-constraint

Write a script to do this

kdexd:
Hi @kashizui:
This solution of Stack Overflow can do the task of prepending '0x' to existing users of the platform. But for the new users, I guess this change has to be made atleast at one place ( [model/bundle_model.py](https://github.com/codalab/codalab-cli/blob/master/codalab/model/bundle_model.py#L1692) ). This is the entry point for any new user, so I guess it will be simply:

```
user_id = user_id or '0x%s' % uuid.uuid4().hex
```

Not yet searched other places to modify (if need be). Thoughts ?

kashizui:
That's correct, that's the easy part, I think that's all we need to do for new users.

kdexd:
Shall I submit a patch now or wait for the existing User IDs to be modified first ? I guess the SQL query can be structured in a way to avoid prepending `0x` to keys which already have it.

kdexd:
Also I think that even a single SQL query on the machine holding this database will do the task within seconds, it is quicker and a script might not be required :smile:

kashizui:
The reason we should write a script is that we have multiple servers that we need to deploy to (e.g. http://worksheets.codalab.org and an internal CodaLab instance on the Stanford network), and we want the procedure to be exactly the same across all instances. This would also allow us to test the procedure beforehand, because most likely there will be bugs or wrong assumptions.

I'm fine with merging the change you specified above first: we should be able to write SQL queries that can distinguish between the old broken IDs and the new "fixed" IDs. There's an additional slight complication that there are also old user IDs that are simple integers (e.g. `127`), which was our original format before I decided to start using UUIDs. I'm not sure what to do with those yet. My instinct is to generate new UUIDs for these users, but I haven't entirely thought through the possible consequences of that.

kdexd:
Ah I see, then I agree that a script would be more suitable. UUID4's have a fixed length of 32 and probably integers cast as string will not be that long. So one possibility can be:
- First iteration, strings which are 32 characters in size must be prepended with `0x`.
[...]

PR #550, opened 2016-08-01
Prepend '0x' to user ids while sign up.

kdexd:
This PR introduces a partial fix to #515. New users from now on will have a 32 character long hexadecimal UUID4 string ( `0x<uuid4>` ) as user id. Prepending `0x` to user id of already existing users is kept out of scope of this PR.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1708,7 +1708,7 @@ class BundleModel(object):
         """
         with self.engine.begin() as connection:
             now = datetime.datetime.utcnow()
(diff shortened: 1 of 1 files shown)


percyliang:
LGTM

kdexd:
- A small note as a reminder:
  _Modify existing user IDs accordingly._
  Directed discussion at concerned issue: #515

( @kashizui )

kashizui:
Probably should have used `spec_util.generate_uuid()`, I'll update this when I write the script to modify existing user IDs too.

PR #551, opened 2016-08-01
Changed the encoding of _pack_params to list of tuples to handle commas

ppasupat:
Changed according to the suggestions in #530

### DIFF
--- a/tests/client/json_api_client_test.py
+++ b/tests/client/json_api_client_test.py
@@ -12,32 +12,29 @@ from codalab.common import PreconditionViolation
 
 class JsonApiClientTest(unittest.TestCase):
     def test_pack_params(self):
-        self.assertDictEqual(JsonApiClient._pack_params({
+        self.assertItemsEqual(JsonApiClient._pack_params({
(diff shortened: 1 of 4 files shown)


tests/client/json_api_client_test.py
```diff
@@ -12,32 +12,29 @@
 
 class JsonApiClientTest(unittest.TestCase):
     def test_pack_params(self):
-        self.assertDictEqual(JsonApiClient._pack_params({
+        self.assertItemsEqual(JsonApiClient._pack_params({
```
kashizui, line 15: Perhaps wrap both arguments to this assertion with `set()` since the order is just an implementation detail?

ppasupat, line 15: `assertItemsEqual(a, b)` checks whether `sorted(a) == sorted(b)`, so the ordering is already discarded.

kashizui, line 15: Oh good point, I should not review code from my phone.

kashizui:
LGTM

Issue #336, opened 2016-02-13
Enforce Single-Words for Tags


a10y:
Currently we don't enforce tags are single-words, but if we print them out on the command line as a set of space-separated characters it makes sense to check for this.

kashizui:
This applies for both bundle and worksheet tags.
This should be enforced on the server side.

karandesai-96:
Since merge of #543 , `cl edit <uuid> --tags a b c` results in three different tags. In my opinion, through a web interface, there shouldn't be a warning on spaces, but it should automatically treat the space separated entries as different tags. If this is not done, then anyone might be able to add only one tag through web interface  as I just checked, old tags are deleted before adding new entries whie updating metadata.

karandesai-96:
@kashizui What are your thoughts on this ? I have found the place where change has to be made for worksheet metadata. I can sure go ahead and submit a patch to knock off this issue :smile:

yshen-google:
I tried both ways (changing tags on UI and changing tabs through CLI) and my observations are following:
## Worksheet
### UI
In Worksheet UI, space is recognized as special character where it separates tags.
### CLI
In Worksheet CLI, space is recognized as special character where it separates tags when running `cl wedit --tags=<tags>`
## Bundle
### UI
In Bundle UI, space is recognized as special character where it separates tags and in CLI it is displayed with "|" to separate tags visually.
### CLI
In Bundle CLI, space is *not* recognized as special character and it is treated as a single tag.

There is inconsistency in bundle, and my proposal is as following:
1. Treat space as special character in Bundle for both CLI and UI to separate tags.
2. Remove the display of "|" in Bundle CLI when showing multiple tags.
3. Add support text to inform users of special character space.

The result will be using space to separate tags and if a user wants some tag like "the best", the user may need to modify it to something like "TheBest".

@percyliang WDYT? If the plan sounds good to you, I can proceed with PRs.

percyliang:
Yes, I think this is a good plan.  Space should be treated as a delimiter and disallowed in tags.

yshen-google:
A small correction to my comment above, it seems for CLI, both bundle and worksheet don't treat space as special delimiter for multiple tags. I will make the change for both.

PR #557, opened 2016-08-05
Split space separated words in metadata 'tag' entry to individual tags.

kdexd:
This PR is a fix for #336. Providing space separated tags through CLI automatically considered them as individual tags. With this PR, space separated words entered through web interface would be separated into individual tags.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -742,6 +742,14 @@ class BundleModel(object):
         # Apply the column and metadata updates in memory and validate the result.
         metadata_update = update.pop('metadata', {})
         bundle.update_in_memory(update)
+
+        # Make individual tags out of space separated words
+        if 'tags' in metadata_update:
+            for index, tag in enumerate(metadata_update['tags']):
+                metadata_update['tags'][index] = u'-'.join(
+                    word for word in tag.strip().split() if word != u''
+                )
+
         for (key, value) in metadata_update.iteritems():
             bundle.metadata.set_metadata_key(key, value)
         bundle.validate()
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -847,7 +847,10 @@ class LocalBundleClient(BundleClient):
             elif key == 'title':
                 metadata[key] = value
             elif key == 'tags':
-                metadata[key] = value
+                if isinstance(value, str):
+                    metadata[key] = value.split()
+                else:
+                    metadata[key] = value
             elif key == 'freeze':
                 metadata['frozen'] = datetime.datetime.now()
             else:
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1227,6 +1227,11 @@ class BundleCLI(object):
             is_new_metadata_updated = True
         if args.tags:
             new_metadata['tags'] = args.tags
+            for index, tag in enumerate(args.tags):
+                args.tags[index] = u'-'.join(
+                    word for word in tag.strip().split() if word != u''
+                )
+            new_metadata['tags'] = args.tags
             is_new_metadata_updated = True
 
         # Prompt user for all information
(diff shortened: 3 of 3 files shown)


codalab/model/bundle_model.py
```diff
@@ -742,6 +742,11 @@ def update_bundle(self, bundle, update, connection=None):
         # Apply the column and metadata updates in memory and validate the result.
         metadata_update = update.pop('metadata', {})
         bundle.update_in_memory(update)
+
+        # Make individual tags out of space separated words
+        if metadata_update.has_key('tags'):
+            metadata_update['tags'] = metadata_update['tags'].split()
+
```
kdexd: Hi @kashizui:
Please have a look, I think worksheet should work, but I am a little confused about the bundle tags as they were not explicitly mentioned like worksheet. I did some back tracing and checked out bundle metadata schema. I had an intuition that this method is the point where manipulation done, would set things right further. I'm not completely certain.

kdexd:
I have a little doubt here. Why is a bundle tag a part of bundle metadata, while worksheet tag is declared separately in `tables.py` ?

kdexd:
I am unable to understand the cause of test failure.

kashizui:
> Providing space separated tags through CLI automatically considered them as individual tags. 

The CLI doesn't automatically consider them separate tags if you use quotes:

```
$ cl edit mybundle --tags "one tag" two tags
```

> Why is a bundle tag a part of bundle metadata, while worksheet tag is declared separately in `tables.py`?

This is because worksheets don't have metadata like bundles do, and we added tags as a feature after we designed the tables.

> I am unable to understand the cause of test failure.

Are you able to run `test-cli.py` on your own machine to debug?

kashizui:
I'll have more time to actually look at your code later this week.

codalab/model/bundle_model.py
kashizui: Just a guess, but this will probably throw an `AttributeError` since `metadata_update['tags']` should already be a `list`

kashizui:
@karandesai-96 You should test your changes on your own local machine to debug, it's easier than trying to read the logs on CircleCI. Let me know if you need any help setting up.

https://github.com/codalab/codalab-worksheets/wiki/Server-Setup

karandesai-96:
@kashizui Indeed I agree, I was waiting for the weekend to spin up a local instance altogether, since I plan to continue on more contributions. Now that the weekend is on, I am about to initiate the procedure. I will mail you with the queries if any. Thanks !

On Aug 6, 2016 12:13 PM, Stephen Koo <notifications@github.com> wrote:

@karandesai-96<https://github.com/karandesai-96> You should test your changes on your own local machine to debug, it's easier than trying to read the logs on CircleCI. Let me know if you need any help setting up.

https://github.com/codalab/codalab-worksheets/wiki/Server-Setup

kdexd:
@kashizui:
`"abc def       " foo "bar"`: `abc-def`, `foo`, `bar`

kdexd:
All spaces are stripped and replaced by single hyphens. Web interface allows comma separated values for multiple tags. The formatting is done at three places for web interface as well as _CLI_. Thoughts ?

kashizui:
- I would prefer throwing an error if a tag doesn't meet our requirements (e.g. has spaces), then users can react by using hyphens if they so choose. @percyliang ?
- Instead of writing duplicate code, let's use `spec_util.check_name()` to validate the format of a tag. Then we might want to add a way to require this validation in the metadata spec for "tags" (in `bundles/named_bundle.py`) rather than manually in these locations.

kdexd:
Hi @kashizui @percyliang, to what decision have we reached here so far ? My project has ended, and now I'm free - so I'll complete this open PR first before I start something else. :smile:

percyliang:
I don't think spaces should be automatically stripped and replaced by
hyphens, nor should we use commas.

This should be as simple / consistent as possible:
cl edit --tags a b c # creates three tags
cl edit --tags "a b" c # throws an error
On the web UI, if someone enters "a b c" into the tags, then this should
create three tags (split automatically on spaces).
All render of lists of tags should be by spaces.

@karandesai-96 sorry for the slow response

kdexd:
Hi @percyliang sounds good ! I will work on this functionality next week.

percyliang:
@karandesai-96 Just checking in on this...

kdexd:
@percyliang Oops, this went out of my attention after the recent examinations in my institute ! I'll pay attention to it this weekend.

Issue #539, opened 2016-07-23
Ubuntu 16.04


percyliang:
After installation, each type one types 'cl ...' into the web terminal, the rest server prompts for a username / password.

kashizui:
@percyliang I'm not sure I can reproduce this, I'm only getting asked for user/pass on the first command. On Ubuntu 16.04.1:
```
ubuntu@ubuntu16:~/codalab-cli$ cl work localhost::
Requesting access at http://localhost:2800
Username: codalab
Password:
Switched to worksheet http://localhost:2800::home-codalab(0x2f128e1182bc425da6dddc4888c5819a).
ubuntu@ubuntu16:~/codalab-cli$ cl ls
### Worksheet: http://localhost:2800::home-codalab(0x2f128e1182bc425da6dddc4888c5819a)
### Title: <none>
### Tags:
### Owner: codalab(0)
### Permissions: public(0x682b58):read
ubuntu@ubuntu16:~/codalab-cli$ cl ls
### Worksheet: http://localhost:2800::home-codalab(0x2f128e1182bc425da6dddc4888c5819a)
### Title: <none>
### Tags:
### Owner: codalab(0)
### Permissions: public(0x682b58):read
ubuntu@ubuntu16:~/codalab-cli$ cl work localhost::
Switched to worksheet http://localhost:2800::home-codalab(0x2f128e1182bc425da6dddc4888c5819a).
ubuntu@ubuntu16:~/codalab-cli$ cl ls
### Worksheet: http://localhost:2800::home-codalab(0x2f128e1182bc425da6dddc4888c5819a)
### Title: <none>
### Tags:
### Owner: codalab(0)
### Permissions: public(0x682b58):read
ubuntu@ubuntu16:~/codalab-cli$
```

kashizui:
@percyliang I'm not able to reproduce this, I'm only getting asked for user/pass twice on Ubuntu 16.04.1 amd64, which is expected on the `master` branch. This issue is fixed on `develop` but hasn't been merged/deployed yet.

percyliang:
The problem is not with typing cl into the shell.  The problem is when you
try commands into the web terminal.  To be more specific:
I type 'cl run date' into the web terminal and then the rest server asks me
for a user name/password.  Then the command gets executed.  If I do this
again, same thing.

I'm seeing this on both develop and master.

kashizui:
[...]

PR #559, opened 2016-08-07
Remove is_cli check on client caching

### LINKED ISSUE #553: web terminal doesn't work (perhaps due to Ubuntu 16.04)
All CLI commands work, but when `cl rm ^` or `cl run date` are typed into the web terminal, `Not authorized` is printed out, even when logged in.  `cl ls` does work though.  Noticed this after upgrading to Ubuntu 16.04.

kashizui:
This should fix #553 and #539, pending further testing. The clients passed into the temporary CodaLabManager were not being used by the web terminal due to this `is_cli` check, which led to the instantiation of *Clients that unauthenticated, prompting credentials or directly throwing "Not authorized" errors.

Also need to remember why I introduced this is_cli check to begin with--I should have documented that better..

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -39,6 +39,7 @@ from distutils.util import strtobool
 from urlparse import urlparse
 
 from codalab.client import is_local_address
+from codalab.client.json_api_client import JsonApiClient
 from codalab.common import UsageError, PermissionError, precondition
 from codalab.server.auth import User
 from codalab.lib.bundle_store import (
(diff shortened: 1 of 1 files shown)


kashizui:
The original commit: 63f52538f0483d9be35c7cb703789fde4897cba9

> Only use client cache when not CLI
> 
> Prevents bug with clashing JsonApiClients and BundleClients in the client cache when they have the same address (as is the case for the Stanford CodaLab instance behind the SSH tunnel). This is only temporary and can be reverted when the REST migration is complete.

Well this is a doozy.

kashizui:
Replaced the original is_cli check with a requirement that the cached client is the correct class (matching the use_rest flag). This should now fix the cache issues and avoid key collision for those using the SSH tunnel to connect to the Stanford instance.

kashizui:
@percyliang

PR #574, opened 2016-08-20
Rate limit error admin notifications

### LINKED ISSUE #567: Cap emails sent through notifications
Rate-limit emails sent through admin notifications.

kashizui:
Adds rate limit on admin notifications.

Limit is hard-coded at 6 per hour for now (one message every 10 minutes), but can add ability to configure this in the future.

Merging into master for quick deployment.

Resolves #567 

@percyliang

### DIFF
--- a/codalab/lib/server_util.py
+++ b/codalab/lib/server_util.py
@@ -20,24 +20,41 @@ class RateLimitExceededError(Exception):
     pass
 
 
-def rate_limited(max_calls, per_seconds):
+def rate_limited(max_calls_per_hour):
+    """
+    Parameterized decorator for rate-limiting a function.
+
+    A running count of remaining calls allowed is kept for the last hour.
+    Every call beyond this limit will raise a RateLimitExceededError.
+    """
     def decorate(func):
         lock = threading.Lock()
         state = {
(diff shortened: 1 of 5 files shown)


codalab/lib/server_util.py
```diff
@@ -13,6 +16,37 @@
 from codalab.common import precondition
 
 
+class RateLimitExceededError(Exception):
+    pass
+
+
+def rate_limited(max_calls, per_seconds):
+    def decorate(func):
```
percyliang, line 30: Please document this function.

Also, why do we need two parameters instead of a single `max_calls_per_second`?  I find `max_calls` kind of hard to reason about due to the `min` statement below.

percyliang:
@kashizui LGTM modulo that one comment

codalab/lib/server_util.py
kashizui, line 30: I agree, one parameter would be easier to reason about. How about `max_calls_per_hour`, since I can't imagine we would want more than one message per second?

percyliang, line 30: That sounds good!

kashizui:
Merged back into develop too

PR #580, opened 2016-08-29
allow connecting to any URL that looks like http://.../bundleservice

percyliang:
allow connecting to any URL that looks like http://.../bundleservice, not just worksheets.codalab.org/bundleservice

@kashizui merging to master to take immediate effect


kashizui:
LGTM from my phone

PR #216, opened 2015-11-10
support wsearch directives

percyliang:
@kashizui please review

PR #587, opened 2016-09-28
Escape filenames when gzipping directories

kashizui:
Fixes codalab/codalab-worksheets#216

Merging into master as hotfix.

@percyliang


percyliang:
Yikes, should make sure this stuff is secure
LGTM

kashizui:
I'm adding a test or two before merging to make sure this doesn't happen again.

PR #589, opened 2016-10-05
RESTify bundle actions

### LINKED ISSUE #507: Add bundle actions API
RESTify `kill` and `write` commands.

Should be fairly straightforward.

kashizui:
Decided to implement this first before the worksheets API (which is already underway) so that I could be free to clean up deprecated code while working on the worksheets API.

Fixes #507

### DIFF
--- a/codalab/client/local_bundle_client.py
+++ b/codalab/client/local_bundle_client.py
@@ -274,54 +274,11 @@ class LocalBundleClient(BundleClient):
 
     @authentication_required
     def kill_bundles(self, bundle_uuids):
-        """
-        Send a kill command to all the given bundles.
-        """
-        check_bundles_have_all_permission(self.model, self._current_user(), bundle_uuids)
-        for bundle_uuid in bundle_uuids:
-            worker_message = {
-                'type': 'kill',
-                'uuid': bundle_uuid,
-            }
-            action_string = BundleAction.kill()
-            self._do_bundle_action(bundle_uuid, worker_message, action_string)
+        raise UsageError("API Updated: Please upgrade to the latest version of the CLI.")
 
     @authentication_required
     def write_targets(self, targets, string):
-        """
-        Write targets by sending a command to all the given bundles in the targets.
-        """
-        bundle_uuids = [bundle_uuid for (bundle_uuid, subpath) in targets]
-        check_bundles_have_all_permission(self.model, self._current_user(), bundle_uuids)
-        for bundle_uuid, subpath in targets:
-            if not re.match('^\w+$', subpath):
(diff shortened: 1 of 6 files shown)


kashizui:
Please review

@percyliang  ( @yashsavani  )

codalab/rest/bundle_actions.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    actions = BundleActionSchema(
+        strict=True, many=True,
+    ).load(request.json).data
+
+    check_bundles_have_all_permission(local.model, request.user, [a['uuid'] for a in actions])
+
+    for action in actions:
+        bundle = local.model.get_bundle(action['uuid'])
+        if bundle.state != State.RUNNING:
```
percyliang, line 37: I think we want to be able to queue up actions on a job that's just starting (in staged, for example)...maybe we want to kill it immediately after creating it before it gets the status of 'running'.

codalab/client/local_bundle_client.py
```diff
@@ -274,54 +274,11 @@ def _bundle_inherit_workheet_permissions(self, bundle_uuid, worksheet_uuid):
...
-        """
-        check_bundles_have_all_permission(self.model, self._current_user(), bundle_uuids)
-        for bundle_uuid in bundle_uuids:
-            worker_message = {
-                'type': 'kill',
-                'uuid': bundle_uuid,
-            }
-            action_string = BundleAction.kill()
-            self._do_bundle_action(bundle_uuid, worker_message, action_string)
+        raise UsageError("API Updated: Please upgrade to the latest version of the CLI.")
```
percyliang: Do we need this?  Wouldn't it be better to have a generic way whereby a server can tell the client the version number, and if this is higher than the client, then it prints out a warning, and recommend the client to upgrade.

percyliang (approved):
Didn't test it, but LGTM modulo comments

codalab/rest/bundle_actions.py
kashizui: I agree, but in our current worker system, a worker has to be online and assigned to run the bundle before we can send a message. In this function, the following call to `get_bundle_worker` will fail.

Queuing up messages will require some significant redesign -- I think that's not high priority right now.

codalab/client/local_bundle_client.py
kashizui: I can implement that with the REST API calls (I'll do so in another PR), but not sure how to do that with XMLRPC. 

I just wanted to delete these functions so that I wouldn't duplicate code, but I'll revert these changes and leave them be until we remove the entire LocalBundleClient.

PR #590, opened 2016-10-06
Cleanup CLI dependencies on LocalBundleClient

kashizui:
In prep for worksheet API changes
- Move user commands into its own section
- Remove dependencies from the administrative CLI commands on
  LocalBundleClient
- Close loophole that allowed BundleStore commands to be run from the
  web terminal
- Remove 'local' requirement on administrative commands, only require
     that the CLI is not headless (i.e. not the web terminal)

@percyliang Please review

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -38,6 +38,7 @@ from codalab.bundles import (
 from codalab.bundles.make_bundle import MakeBundle
 from codalab.bundles.uploaded_bundle import UploadedBundle
 from codalab.bundles.run_bundle import RunBundle
+from codalab.client import is_local_address
 from codalab.common import (
     precondition,
     State,
@@ -735,7 +736,7 @@ class BundleCLI(object):
         )
     )
     def do_logout_command(self, args):
-        self._fail_if_headless('logout')
+        self._fail_if_headless(args)
         if args.alias:
             address = self.manager.apply_alias(args.alias)
             self.manager.logout(address)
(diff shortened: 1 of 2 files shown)


percyliang:
I think some of the admin commands we really want to be local because they're kind of dangerous.  Also, with the REST API, these will not not be available remotely anyway.

kashizui:
I think that is the point -- when fully migrated to REST, there will be no concept of a "local" client. The client will always be connected to a "remote" host over HTTP. So I'm removing the "local" check because there won't be a local client anymore, and the following code would no longer make sense:

```
    client = self.manager.current_client()  # after next week, this will only return a REST client
    query_info = {      
        'user': args.user, 'command': args.match_command, 'args': args.args, 'uuid': args.uuid,     
        'count': args.count, 'group_by': args.group_by      
    }       
    info = client.get_events_log_info(query_info, args.offset, args.limit)
```

kashizui:
Perhaps we should alternatively think about how to protect CodaLab from errant use of administrative commands. Maybe it should be part of a different CLI (e.g. `cl-admin`) or it should require admin credentials on the local instance.

percyliang:
Yes, I think once REST is in, then the admin commands could be a separate
script probably.

kashizui:
Oops, in case it was confusing, in this PR I followed the example of `cl bs-ls-partitions` and use the `model` and `bundle_store` directly regardless of which instance we're connected to.  And `BundleRPCServer` also does not accept administrative commands.

These changes help me clear the way to make the worksheets API cleaner.

kashizui:
Thinking about it more, I agree we should put the "local" checks back, since a user might try to run an admin command against a remote instance, without realizing that it will inadvertantly modify their own local instance.

kashizui:
@percyliang Added back the "local" checks, please take another look!

percyliang:
LGTM

PR #592, opened 2016-10-07
Compare versions across REST API

### LINKED ISSUE #544: Inform user that a new version of the CLI is available
Nice feature to have.

CLI should periodically check (once a day when a cl command is run?) if there are updates available, then inform the user once.

kashizui:
Notify CLI user when a new version is available.

Fixes #544

@percyliang

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -508,7 +508,7 @@ class CodaLabManager(object):
         else:
             from codalab.client.remote_bundle_client import RemoteBundleClient
 
-            client = RemoteBundleClient(address, lambda a_client: self._authenticate(auth_cache_key, a_client), self.cli_verbose)
+            client = RemoteBundleClient(address, lambda a_client: self._authenticate(auth_cache_key, a_client), self.check_version, self.cli_verbose)
             self._authenticate(auth_cache_key, client)
 
         # Cache and return client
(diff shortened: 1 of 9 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -605,6 +612,25 @@ def set_current_worksheet_uuid(self, address, worksheet_uuid):
             if 'worksheet_uuid' in session: del session['worksheet_uuid']
         self.save_state()
 
+    def check_version(self, server_version):
+        # Enforce checking version at most once every 24 hours
+        now = datetime.datetime.now()
+        last_check_dt = formatting.parse_datetime(self.state.get('last_check_version', formatting.datetime_str(datetime.datetime.utcfromtimestamp(0))))
```
percyliang: Replace last_check_version with something like last_check_version_datetime for clarity

codalab/lib/codalab_manager.py
```diff
@@ -605,6 +612,25 @@ def set_current_worksheet_uuid(self, address, worksheet_uuid):
...
+        if (now - last_check_dt) < datetime.timedelta(days=1):
+            return
+        self.state['last_check_version'] = formatting.datetime_str(now)
+        self.save_state()
+
+        # Print notice if server version is newer
+        if server_version.split('.') > CODALAB_VERSION.split('.'):
```
percyliang: Don't you need to convert to integers? Otherwise 11 will come before 9....

percyliang (approved):
Cool!  LGTM modulo comments

percyliang:
Tests are failing...

codalab/lib/codalab_manager.py
kashizui: Yikes, good catch

kashizui:
Added support for XMLRPC as well.

PR #593, opened 2016-10-10
get emails for mailing list.

yashsavani:


percyliang:
I don't think this is quite what we want...
1) If we want a command-line tool, then we should be able to do this through cl uinfo (which should be able to do much more than just print out email addresses)
2) It would be much more useful to be able to do it from the website (some admin console), where you specify the message, and it sends it out to everyone (recording which messages have been sent so we can do things incrementally).

PR #594, opened 2016-10-10
Schema bug

### LINKED ISSUE #326: Backslashes in table schema regexp are removed after saving
To reproduce, add a table schema

```
% schema hello
% add host remote s/\..*/
```

The second line should strip `juke29.stunford.edu` into `juke29`. However, after saving, the schema is changed into

```
% schema hello
% add host remote s/..*/
```

which replace everything into an empty string. Quoting does not help, and this bug affects both the online editor and the command line editor.

However, the following is fine:

```
% schema hello
% add host remote "s/\..*/ | s/hello/world"
```

With multiple filters, the slash is not removed when saved.

yashsavani:
fixed #326 by putting the string into the database without compiling it using shlex


percyliang:
Please explain more what problem this PR is supposed to solve.

I don't think you can just separate items and directive_items, because the order in which the two are interleaved matters for interpretation.

percyliang:
Code LGTM, but I haven't tested the code or taken a careful look at corner cases (tests also seem to be failing, not sure if due to your PR).  Could use another pair of eyes (@kashizui)

kashizui:
Just took a look--

I think we can create a cleaner solution than adding a new 'stringified_items' list parallel to the existing 'items' list.

## Background

We have basically two canonical formats for worksheet items at the moment:

**db-style:** `(bundle_uuid, subworksheet_uuid, value, type)`

This is what the model methods use, and what `worksheet_util.convert_item_to_db` returns. Consists of just UUIDs for the bundle and subworksheets (with `None` for null references) and a string `value`.

**info-style:** `(bundle_info, subworksheet_info, value_obj, type)`

This is what the interpretation functions use, and what `LocalBundleClient._convert_items_from_db` returns. Consists of "info" `dict`s for the bundles and subworksheets, and most importantly, the `value_obj` is a string for markup items, and _a tokenized list for directive items_.

## The Problem

As #326 pointed out, backslashes are being lost in directives when editing and saving worksheets. The problem is that when you save a worksheet, the lines will be parsed into **info-style** items (see `worksheet_util.parse_worksheet_form`). Then these items will be converted into **db-style** items before being saved into the database (see `LocalBundleClient.update_worksheet_items`).  During this process, directive items are tokenized using `shlex.split`, which will interpret any of the backslashes and quoting (and thus dropping them from the original token), after which they are _rejoined_ using plain spaces before being added to the database. Thus when they are retokenized during editing, we have already lost information.

## The Solution

We should change the **info-style** items to use plain strings for directives as well. (The `value_obj` should never be a list.) **The directive strings should only be tokenized during interpretation, and these tokens should never be rejoined and used elsewhere.** The plain directive strings that the user types in should be saved verbatim into the database, and only tokenized on an as-needed basis.

Steps:
1. Make sure that any function that produces **info-style** items does not tokenize the directives.
2. Update any of the worksheet interpretation code that uses **info-style** items to tokenize the directives on an as-needed basis.

Thoughts?

@yashsavani @percyliang

percyliang:
Thanks, Stephen!  Yeah, this is definitely better than having two parallel items.

Actually, how about we actually just apply shlex.split as usual, and just make sure that when we display it as raw lines, we escape backslashes so that the information is preserved.  Basically, formatting.string_to_tokens and formatting.tokens_to_string should be inverses (modulo canonicalizations), which they are not right now.  I like this solution better because logically, we the directive _is_ a list of tokens, and not being able to store it like that is a bit unsettling.

kashizui:
I found a good candidate for the inverse of `shlex.split`:
 [pipes.quote](https://docs.python.org/2/library/pipes.html#pipes.quote)

It correctly handles the token in #326 

```
>>> pipes.quote('s/\..*/')
"'s/\\..*/'"
>>> print pipes.quote('s/\..*/')
's/\..*/'
```

So we can do

```
# formatting.py
import pipes

def tokens_to_string(tokens):
    return ' '.join(map(pipes.quote, tokens))
```

percyliang:
Nice!

kashizui:
@yashsavani To summarize, it looks like the only thing we need to change is the implementation of the `formatting.tokens_to_string` function.

PR #596, opened 2016-10-16
fix pull request #592, pass in correct dummy check_version lambda fun…

ChengshuLi:
fix pull request #592, pass in correct dummy check_version lambda function

@kashizui


kashizui (approved):
Was this the only place that needed this fix?

ChengshuLi:
@kashizui  yes it was the only place.

PR #597, opened 2016-10-17
Make directive de/serialization proper inverses

### LINKED ISSUE #326: Backslashes in table schema regexp are removed after saving

yashsavani:
Fixes #326 by using `pipes.quotes` in the `tokens_to_string` function of `formatting.py` so that quotes are maintained in the representation.

### DIFF
--- a/codalab/lib/formatting.py
+++ b/codalab/lib/formatting.py
@@ -5,6 +5,7 @@ Provides basic formatting utilities.
 import datetime
 import json
 import shlex
+import pipes
 
 from worker import formatting as worker_formatting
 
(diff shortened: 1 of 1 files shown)


percyliang:
LGTM

kashizui (approved):
Did you confirm that this fix indeed keeps the backslash now?

kashizui:
@yashsavani Once you have confirmed that this PR fixes the example error in #326, you can go ahead and merge it in.

PR #205, opened 2015-11-06
Improve worksheet item manipulation commands

kashizui:
- Remove `cl wadd`
- Add only text items or subworksheets through `cl add`
- Create requirements file for running unit tests
- Fix test-cli issues

Fixes #201

@percyliang @pbhatnagar3

PR #600, opened 2016-10-24
Tell user to click Upload button

kashizui:
...when we receive a bare `cl upload` command on the web terminal.

We do this browsers do not allow JavaScript code to directly
open a file dialog outside of a user-initiated event. Because
there is a network call in the chain of events from typing in
`cl upload` in the web terminal to receiving an Upload "UI action"
from the server, the browser considers the final callback to be
script-initiated, not directly user-initiated.

Resolves codalab/codalab-worksheets#205

@percyliang @yashsavani

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -858,11 +858,9 @@ class BundleCLI(object):
         + EDIT_ARGUMENTS,
     )
     def do_upload_command(self, args):
-        # If headless and no args provided, request an Upload dialog on the front end.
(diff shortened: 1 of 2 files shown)


percyliang:
I think at this point, we should just not "support" a bare `cl upload`, and just print out a usage error like what happens on the CLI.  The bare `cl upload` was created in an era when we didn't have upload buttons, but it seems kind of useless now.

kashizui:
I see -- that makes sense.

PR #602, opened 2016-10-27
Don't copy uuids in mimic

kashizui:
Original problem:

> The only thing I found is that I think cl mimic is broken:
> 
> ```
> cl up a.txt
> cl up b.txt
> cl run input:a.txt 'wc -l input'
> cl mimic a.txt b.txt
> This creates a bundle that still uses a.txt instead of b.txt...
> ```

The source of the problem was newly allowing users to specify their own UUIDs for new bundles, for the use cases of `cl add` and `cl wadd`. Before this, `cl mimic` relied on the assumption that the server would ignore any specified UUIDs in new bundles. But once the restriction was eased, the mimic operation accidentally used the old UUIDs when creating the new mimicked bundles, causing the server to do nothing. Thus `cl mimic a.txt b.txt` above actually just duplicated the _item_ without creating a new bundle.

@percyliang

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1856,8 +1856,13 @@ class BundleCLI(object):
                                        for dep in old_info['dependencies'])
             if lone_output or downstream_of_inputs:
                 # Now create a new bundle that mimics the old bundle.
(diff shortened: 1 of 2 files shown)


percyliang:
LGTM

PR #603, opened 2016-10-27
Don't simplify direct uploads of directories

### LINKED ISSUE #578: when upload directory with single file, it gets rid of that directory and just uploads that file
Instead, the directory + file structure should be kept.  This seems to have been introduced recently.

@kashizui Do you know when this came about?

kashizui:
The intermediate archive created during `cl upload DIRECTORY`
should not be "simplified."

Added a test case for this in upload1 and fixes another
test case that didn't expect the new dir2 in the test files.

Fixes #578


kashizui:
Whoops -- so test-cli failed because it had exactly this test case: It uploads a tar.gz archive that contains only one directory, and it expects the resulting bundle to be flattened so that the directory constitutes to the root of the bundle.

percyliang:
If you do `cl upload foo`, where `foo` is a directory, it seems like the bundle should be called foo and in it should be the contents (which could be one file).  The problem is that the current behavior is that the contents of the bundle just become of the contents of the bundle.

It should still be the case that if foo.tar.gz contains one directory foo, the bundle contents should be set to the contents of the directory, as the test demands.

kashizui:
So, the client should tell the server whether or not to "simplify" the archive. I'll try that out and add a test case for this as well.

kashizui:
@percyliang Please review new changes!

tests/files/dir2/the-only-file
```diff
@@ -0,0 +1,9 @@
+This is the only file in dir2.
+`cl upload dir2` should NOT collapse the resulting bundle into just this file.
+This
+file
+has
+a
+lot
+of
+lines.
```
percyliang: add newline?

codalab/lib/bundle_cli.py
```diff
@@ -942,7 +942,7 @@ def do_upload_command(self, args):
             sources = [path_util.normalize(path) for path in args.path]
 
             print >>self.stderr, "Preparing upload archive..."
-            upload_blob, filename, total_bytes, unpack_at_server =\
+            upload_blob, filename, total_bytes, unpack_at_server, simplify_at_server =\
```
percyliang: 5-tuples don't seem great since they're more error prone...maybe consider using a dict?  I don't feel super strongly about it, but just wanted to bring up the issue.

percyliang (approved):
LGTM modulo the 5-tuple issue

Issue #573, opened 2016-08-18
ibdata1 growing too fast


kashizui:
https://www.percona.com/blog/2013/08/20/why-is-the-ibdata1-file-continuously-growing-in-mysql/

I have a suspicion that worker check-ins or something are creating never-closed transactions, etc. (maybe related to #571)

kashizui:
ibdata1 file has grown from zero to 7.3gb on wstest in the last two weeks, this is probably an unreasonable amount? I'm not actually very familiar with InnoDB, this requires more research.

kashizui:
Running the innochecksum program provided here 
http://bugs.mysql.com/bug.php?id=57611

I get this on wsprod:

```
codalabwsprod:~% sudo ./innochecksum /codalabdata/mysql/ibdata1
bytes read (8192) doesn't match universal page size (16384)
0   bad checksum
36121   FIL_PAGE_INDEX
9727    FIL_PAGE_UNDO_LOG
10  FIL_PAGE_INODE
38  FIL_PAGE_IBUF_FREE_LIST
202241  FIL_PAGE_TYPE_ALLOCATED
98  FIL_PAGE_IBUF_BITMAP
137 FIL_PAGE_TYPE_SYS
2   FIL_PAGE_TYPE_TRX_SYS
3   FIL_PAGE_TYPE_FSP_HDR
102 FIL_PAGE_TYPE_XDES
1338485 FIL_PAGE_TYPE_BLOB
0   FIL_PAGE_TYPE_ZBLOB
0   other
219 max index_id
undo type: 786 insert, 8941 update, 0 other
undo state: 0 active, 150 cached, 337 to_free, 1561 to_purge, 0 prepared, 7679 other
#pages  index_id
1   1
1   2
1   3
1   4
1   5
1   11
1   12
1   13
1   14
1   1562
83  1563
17  1564
18  1565
61  1568
28  1569
418 1570
321 1571
152 1572
21367   1573
392 1574
1246    1575
256 1576
601 1577
425 1578
1   1579
1   1580
1   1581
1   1582
30  1583
15  1584
18  1585
5   1586
3   1587
3   1588
5   1592
3   1593
1   1594
1   1595
149 1596
11  1597
72  1598
69  1599
1   1600
1   1601
1   1602
14  1700
1   2258
1   2259
1   2260
1   2261
1   2262
1   2264
1   2265
1   2266
1   2269
1   2270
1   2272
3   2273
1   2274
1   2275
1   2276
[...]

PR #607, opened 2016-11-16
Fix the worker checkin deadlock

### LINKED ISSUE #571: Frequent MySQL errors on worker check-in
Getting a MySQL deadlock error on worksheets.codalab.org every few hours at irregular intervals when handling a worker check-in.

Error report:

```
Error on request by codalab(0):

POST /workers/clws-w2(29917)/checkin

Query params:


Form params:


JSON body:
{
    "cpus": 4,
    "dependencies": [
        [
            "0x1efccd6e618f4318a0a0ab4b3efea94d",
            ""
        ],
        [
            "0x48d7ed4d68ca447ab0cc1edef00164e6",
            ""
        ],
<truncated>
        [
            "0x3cd3bd8c01554e9189f49dd7f8398d14",
            ""
        ]
    ],
    "memory_bytes": 14707765248,
    "slots": 4,
    "tag": null,
    "version": 6,
    "will_upgrade": false
}

  File "/home/azureuser/codalab-cli/codalab/server/rest_server.py", line 136, in wrapper
  File "/home/azureuser/codalab-cli/codalab/server/json_api_plugin.py", line 16, in wrapper
  File "/home/azureuser/codalab-cli/codalab/server/authenticated_plugin.py", line 34, in wrapper
  File "/home/azureuser/codalab-cli/codalab/rest/workers.py", line 41, in checkin
    request.json['dependencies'])
  File "/home/azureuser/codalab-cli/codalab/model/worker_model.py", line 64, in worker_checkin
    cl_worker_dependency.c.worker_id == worker_id)))
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1991, in execute
    return connection.execute(statement, *multiparams, **params)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 914, in execute
    return meth(self, multiparams, params)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/sql/elements.py", line 323, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1010, in _execute_clauseelement
    compiled_sql, distilled_params
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1146, in _execute_context
    context)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/util/compat.py", line 199, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1139, in _execute_context
    context)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/default.py", line 450, in do_execute
    cursor.execute(statement, parameters)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 205, in execute
    self.errorhandler(self, exc, value)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
OperationalError: (_mysql_exceptions.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction') [SQL: u'DELETE FROM worker_dependency WHERE worker_dependency.user_id = %s AND worker_dependency.worker_id = %s'] [parameters: ('0', 'clws-w2(29917)')]
```

kashizui:
- Use the worker_dependency table like a key-value store with serialized blobs, using advice from here: https://www.percona.com/live/mysql-conference-2013/sessions/practical-tips-using-mysql-scalable-key-value-store
- Also minimize tablespace fragmentation by mark workers as inactive instead of deleting all its corresponding rows.
- Also break ties among workers in job scheduling with a random key to avoid unevenly distributing jobs. This had led to workers 3-6 to dying on prod due to #608.

Tested on wstest and currently deployed to worksheets.codalab.org to see if it solves the deadlock problem once and for all.

Fixes #571 

@percyliang

### DIFF
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -11,13 +11,14 @@ from sqlalchemy import (
 )
 from sqlalchemy.types import (
   BigInteger,
-  Integer,
-  String,
-  Text,
   Boolean,
   DateTime,
-  Float,
   Enum,
+  Float,
+  Integer,
+  LargeBinary,
+  String,
+  Text,
 )
 from sqlalchemy.sql.schema import ForeignKeyConstraint
 
@@ -332,6 +333,7 @@ worker = Table(
   Column('memory_bytes', BigInteger, nullable=False),  # Total memory of worker.
   Column('checkin_time', DateTime, nullable=False),  # When the worker last checked in with the bundle service.
   Column('socket_id', Integer, nullable=False),  # Socket ID worker listens for messages on.
+  Column('is_active', Boolean, nullable=False),  # True if worker is considered active and waiting for jobs.
 )
 
 # Store information about all sockets currently allocated to each worker.
@@ -364,13 +366,10 @@ worker_run = Table(
 worker_dependency = Table(
   'worker_dependency',
   db_metadata,
-  Column('user_id', String(63), ForeignKey(user.c.user_id), nullable=False),
-  Column('worker_id', String(127), nullable=False),
+  Column('user_id', String(63), ForeignKey(user.c.user_id), primary_key=True, nullable=False),
+  Column('worker_id', String(127), primary_key=True, nullable=False),
   ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
 
-  # No foreign key here, since we don't have any logic to clean-up bundles that
-  # are deleted.
-  Column('dependency_uuid', String(63), nullable=False),
-  Column('dependency_path', Text, nullable=False),
+  Column('dependencies', LargeBinary, nullable=False),
   sqlite_autoincrement=True,
 )
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -9,6 +9,7 @@ from sqlalchemy import (
   Table,
   UniqueConstraint,
 )
+from sqlalchemy.dialects.mysql import MEDIUMBLOB
 from sqlalchemy.types import (
   BigInteger,
   Boolean,
@@ -370,6 +371,6 @@ worker_dependency = Table(
   Column('worker_id', String(127), primary_key=True, nullable=False),
   ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
 
-  Column('dependencies', LargeBinary, nullable=False),
+  Column('dependencies', MEDIUMBLOB, nullable=False),
   sqlite_autoincrement=True,
 )
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -369,6 +369,8 @@ worker_dependency = Table(
   Column('worker_id', String(127), primary_key=True, nullable=False),
   ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
 
+  # Serialized list of dependencies for the user/worker combination.
+  # See WorkerModel for the serialization method.
   Column('dependencies', LargeBinary, nullable=False),
   sqlite_autoincrement=True,
 )
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -9,7 +9,6 @@ from sqlalchemy import (
   Table,
   UniqueConstraint,
 )
-from sqlalchemy.dialects.mysql import MEDIUMBLOB
 from sqlalchemy.types import (
   BigInteger,
   Boolean,
@@ -371,6 +370,6 @@ worker_dependency = Table(
   Column('worker_id', String(127), primary_key=True, nullable=False),
   ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
 
-  Column('dependencies', MEDIUMBLOB, nullable=False),
+  Column('dependencies', LargeBinary, nullable=False),
   sqlite_autoincrement=True,
 )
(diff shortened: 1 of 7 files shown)


kashizui:
Looks like the default limit is 1mb
http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

Not sure how upload sidestepped this issue.

codalab/model/tables.py
```diff
@@ -364,13 +366,10 @@
...
-  Column('worker_id', String(127), nullable=False),
+  Column('user_id', String(63), ForeignKey(user.c.user_id), primary_key=True, nullable=False),
+  Column('worker_id', String(127), primary_key=True, nullable=False),
   ForeignKeyConstraint(['user_id', 'worker_id'], ['worker.user_id', 'worker.worker_id']),
 
-  # No foreign key here, since we don't have any logic to clean-up bundles that
-  # are deleted.
-  Column('dependency_uuid', String(63), nullable=False),
-  Column('dependency_path', Text, nullable=False),
+  Column('dependencies', LargeBinary, nullable=False),
```
percyliang, line 374: Document what is in here?

codalab/model/worker_model.py
```diff
@@ -45,83 +45,127 @@ def worker_checkin(self, user_id, worker_id, tag, slots, cpus, memory_bytes, dep
...
-                                     cl_worker_dependency.c.worker_id == worker_id)))      
-            conn.execute(cl_worker.delete()
-                         .where(and_(cl_worker.c.user_id == user_id,
-                                     cl_worker.c.worker_id == worker_id)))
+
+    def cleanup_inactive_workers(self):
+        """
+        Deletes inactive workers and all their associated data from the database
+        as well as the socket directory. Should only be run once in a while,
+        e.g. at most once a week.
```
percyliang: Maybe explain why only at most once a week - to reduce fragmentation, etc.

codalab/worker/bundle_manager.py
```diff
@@ -222,15 +223,15 @@ def _make_bundle(self, bundle):
             with self._make_uuids_lock:
                 self._make_uuids.remove(bundle.uuid)
 
-    def _cleanup_dead_workers(self, workers, callback=None):
+    def _deactivate_dead_workers(self, workers, callback=None):
```
percyliang: Probably better to call 'nonresponsive' instead of 'dead' since they haven't been marked with some status yet

codalab/worker/bundle_manager.py
```diff
@@ -313,12 +314,12 @@ def _filter_and_sort_workers(self, workers_list, bundle):
                 return []
 
         # Sort according to the number of dependencies available, breaking
-        # ties by the number of free slots.
+        # ties by the number of free slots and then by a random key.
```
percyliang: Explain why random key.

percyliang (approved):
Part of me still feels a bit uneasy about 1) packing all the dependencies into a blob and 2) requiring workers not to be deleted too frequently.  I feel like there's a lot of implicit knowledge that's not represented in this PR.  Might be worth writing this down as comments...

kashizui:
Will do, I'll make sure to flesh out all of my reasoning in the comments with the code.

kashizui:
I might do some profiling of this while it is deployed as well to make sure it's performant.

percyliang:
Good idea.  Can we do this on wstest for a bit?  You can point some of the
prod workers to test temporarily...

kashizui:
After making sure it worked on wstest, it is actually currently deployed on wsprod, since it had many more bundles/dependencies and more likely to incur the deadlock.

kashizui:
~~Performed some queries of the checkin runtime before and after I deployed this branch on 11/16:~~

```
>> select AVG(duration) from event where date < "2016-11-16" and date > "2016-11-11" and command like "%checkin"
5.019094776022793

>> select AVG(duration) from event where date > "2016-11-16" and command like "%checkin"
5.030078791620384
```

~~We can see that performance of the command is not significantly impacted by this change, this terrible 5-second runtime is probably dominated by the time it takes to transfer the whole dependency list over the network and then maybe to parse the large JSON. The urgency of #608 is clear.~~

**EDIT: I'm an idiot, these calls take 5 seconds because the worker is long-polling for messages from the server (at intervals of 5 seconds).**

Given new findings for the disk space issue ( #573 ) I will roll back the changes requiring workers not to be deleted too frequently, since that's probably a premature optimization. I will keep the blob storage method since it appears to have fixed the deadlock issue, and I will document it better.

kashizui:
Did some quick and dirty profiling of the checkin and get_workers commands after deploying this change, and found that get_workers takes on average 5-10ms (including deserialization), and checkin takes 10-50 ms (including serialization).

Also did some profiling of these commands on master: get_workers 10-50ms, checkin a whopping 200-400ms.

Serialized blobs gets us a 10x speedup.

Issue #508, opened 2016-07-02
Document REST API


kashizui:
As the API begins to stabilize, fully document their parameters and behavior.

Open for discussion: how do we document them in the code comments and then generate useful API documentation we can publish online?

kashizui:
Assigning only as P2 since we're the only clients of the API at the moment.

kashizui:
Consider using http://apidocjs.com/

kashizui:
@percyliang @yashsavani 
Began hammering this out:
https://github.com/codalab/codalab-worksheets/wiki/REST-API-Reference

I might consider autogenerating portions of this from our code since keeping it in sync with our codebase is going to be a pain.

kashizui:
Done

PR #612, opened 2016-11-23
Generate REST API docs from code

kashizui:
Generate REST API documentation from our Marshmallow schemas, Bottle configurations, and docstrings. The idea is that most the information is basically already in the code, just need to extract and format it. General solutions available online such as Sphinx and Swagger are not immediately compatible with our libraries and also require significant additional instrumentation.

Started a barebones skeleton that extracts all the information and just dumps it to stdout. Goal is to output hyperlinked Markdown documents, but keep everything as simple as possible.

Addresses #508


kashizui:
Preview:
https://github.com/codalab/codalab-cli/blob/kashizui/gendoc/docs/rest/index.md

percyliang:
Do you think we should make all the docs on one page or otherwise make
things more searchable?

kashizui:
I think one page is a good idea, unless all of our documentation is in some unified searchable place.

PR #613, opened 2016-11-23
Diff dependencies

kashizui:
The checkin procedure is:

1. Worker computes hash of its current dependencies
2. Worker computes "diff" between current dependencies and last-checked-in dependencies.
3. Worker sends hash of its last-checked-in dependencies and the patch computed in step (2).
4. Server compares hash with its last received hash. If it matches, apply patch to its copy of the worker's dependencies and respond with success. If it doesn't match, return a message requesting the worker to revert to the old protocol and resend its entire set of dependencies.

Note that changes to server-side API are backwards-compatible for old workers until they self-update.

TODO:

- [ ] Decompose the code a bit to clean up logic
- [ ] Add dependencies_hash column to worker (or worker_dependency)
- [ ] Implement WorkerModel.
- [ ] Enable some event logging for checkins and dependencies
- [ ] Test!

Address #573

### DIFF
--- a/worker/dependency_diff.py
+++ b/worker/dependency_diff.py
@@ -1,0 +1,36 @@
+"""
+Shared utility functions to hash, diff, and patch dependency lists.
+
+Dependency lists are assumed to be lists of 2-tuples.
+These implementations currently focus on readability rather than performance,
+so there is a lot of room for optimization.
+
+Future changes to the hashing scheme need not be backwards-compatible, since
+it is very unlikely to get a hash collision between the new and old schemes,
+forcing the dependencies to the fully resynced.
+"""
+import hashlib
+
+
+def hash_dependencies(dependencies):
+    dependencies.sort()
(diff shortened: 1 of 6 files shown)


worker/dependency_diff.py
```diff
@@ -0,0 +1,36 @@
...
+    h = hashlib.md5()
+    h.update(compact)
+    return h.hexdigest()
+
+
+def diff_dependencies(old, new):
+    old = set(old)
+    new = set(new)
+    return {
+        '+': list(new - old),
```
percyliang, line 27: Should we maybe sort these lists to have more reproducibility/consistency?

percyliang:
Note that in the future, workers will send up more information (e.g., list of docker containers on hand), so just wanted to check whether we need to design things any differently to handle that case...

kashizui:
I didn't think of that -- will have to rethink this a bit to make sure it's just general enough to handle those future changes.

PR #616, opened 2016-11-29
Fix state logic for uploading bundles

### LINKED ISSUE #615: Bundle copying loses some files
From @robinjia

> I found another cl add issue: some files just disappear altogether.
> 
> I did `cl add -d` from https://worksheets.codalab.org/bundles/0xb1054602f4d74bef91ceb949387e2180/ to create https://codalab.stanford.edu/bundles/0xb1054602f4d74bef91ceb949387e2180/ .  You'll note that there are many fewer output files on the second bundle than the first.

> I tried repeatedly doing
> 
> cl add bundle main::0xb1054602f4d74bef91ceb949387e2180 .
> 
> followed by
> 
> cl rm ^1
> 
> Most of the time, the added bundle is missing some files; exactly what's missing is non-deterministic.  The '-d' flag doesn't seem to matter.

kashizui:
- Add wait_for_upload flag to `POST /bundles`
  - Update copy_bundles to set wait_for_upload=1.
  - Prevents race condition between upload and job scheduling when copying run bundles between instances.
- Add state_on_success parameter to `PUT /bundles/<uuid>/contents/blob`
  - Update copy_bundles to set state_on_success
  - Properly copies over final states of run bundles
- Add finalize_on_failure parameter to `PUT /bundles/<uuid>/contents/blob`
  - Workers don't want to finalize bundles on upload failures since it will retry, but the CLI does want to finalize bundles, so the flag helps to disambiguate.

Fixes #615

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -226,10 +226,10 @@ def _update_bundles():
     # Check permissions
     bundle_uuids = [b['uuid'] for b in bundle_updates]
     check_bundles_have_all_permission(local.model, request.user, bundle_uuids)
+    bundles = local.model.batch_get_bundles(uuid=bundle_uuids)
 
     # Update bundles
-    for update in bundle_updates:
-        bundle = local.model.get_bundle(update.pop('uuid'))
+    for bundle, update in zip(bundles, bundle_updates):
         local.model.update_bundle(bundle, update)
 
     # Get updated bundles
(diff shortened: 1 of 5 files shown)


kashizui:
@percyliang please review

percyliang:
How does the original state of the run bundle get set after copying is done?

kashizui:
What do you mean by original state?

percyliang:
The state (failed or ready) of the run bundle that you're copying on the
source instance.

kashizui:
That is a good point, I forgot about copying bundle state.

I'm thinking that it would make the most sense to allow user to specify the
'state' field when creating the bundle (it is currently ignored). But at
the same time, a user shouldn't be able to arbitrarily create a bundle that
is already in the RUNNING state.
There is also the issue that a bundle will temporarily be in UPLOADING
state until it is finalized with its final state.

The easiest solution would probably be to allow the user to pass 'ready' or
'failed' to the upload API to finalize the bundle with the correct state.
Does that sound reasonable?

percyliang:
Yeah, that sounds reasonable.

kashizui:
@percyliang ready for second review!

codalab/rest/bundles.py
```diff
@@ -373,14 +380,24 @@ def _update_bundle_contents_blob(uuid):
         simplify - (optional) 1 if the uploaded file should be 'simplified' if
                    it is an archive, or 0 otherwise, default is 1
                    (See UploadManager for full explanation of 'simplification')
-        finalize - (optional) 1 if this should be considered the final version
-                   of the bundle contents and thus mark the bundle as 'ready'
-                   when upload is complete and 'failed' if upload fails, or 0 if
-                   should allow future updates, default is 0
+        finalize_on_failure - (optional) 1 if bundle state should be set to
```
percyliang: Are these actually 1s and 0s are trues and falses?

percyliang: In particular, above you have things like True ('1').

percyliang:
LGTM

codalab/rest/bundles.py
kashizui: Fixed this to be more consistent.

kashizui:
Not sure why the CircleCI check says pending, but the tests did succeed.

PR #624, opened 2016-12-25
Fix empty-relationship check

### LINKED ISSUE #619: cl ginfo on public group crashes
`KeyError: user_name`

kashizui:
The bug resulted from a change that broke the assumption that null relationships were falsey.

Fixes #619

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -73,7 +73,9 @@ class JsonApiRelationship(dict):
         self.type_ = type_
         self.id_ = id_
         dict.__init__(self, *args)
-        # Allow attributes to override the type and id keys in the dict
+        # Allow the JSON API resource object "attributes" [not Python instance
+        # attributes] to override the type and id keys in the dict.
+        # The actual type and id will still be accessible as instance attrs.
         self.setdefault('type', type_)
         self.setdefault('id', id_)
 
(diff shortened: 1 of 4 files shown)


codalab/client/json_api_client.py
```diff
@@ -78,13 +78,21 @@ def __init__(self, type_, id_, *args):
...
-        """Serialize into relationship linkage dict for JSON API requests."""
-        return {
-            'data': {
-                'type': self.type_,
-                'id': self.id_,
+        """
+        Serialize into relationship linkage dict for JSON API requests.
+        Empty relationships should be properly serialized as a null linkage.
+        """
+        if self:
```
percyliang: This looks strange...how can `self` be None?

percyliang (commented):
Are there any other cases where a to-one relationship needs to have a none value?

codalab/lib/bundle_cli.py
```diff
@@ -2589,7 +2589,7 @@ def do_ginfo_command(self, args):
         group = client.fetch('groups', args.group_spec)
 
         members = []
-        # group['owner'] may be None (i.e. for the public group)
+        # group['owner'] may be a falsey null-relationship (i.e. for the public group)
```
percyliang, line 2592: Does the Javascript client need to do anything special for this?

kashizui: No, for the JavaScript client it should just be null. Also implementation for JavaScript should be a lot simpler since the scope for updating attributes/relationships will be a lot smaller, at least at first. 

I do want to think about how to make this a little simpler though, the JsonApiClient stuff seems like it's starting to get a bit unwieldy and maybe it can be much simplified with some thought. I'll stew on this for a bit longer.

codalab/client/json_api_client.py
kashizui: `self` is falsey if it's an empty relationship, as defined in JsonApiClient.__nonzero__. This might be clearer if I revert back to having a separate subclass for empty relationships.

kashizui:
@percyliang I cleaned up the logic a bit.

percyliang:
LGTM

PR #626, opened 2016-12-29
Support nvidia-docker

### LINKED ISSUE #591: libnvidia-fatbinaryloader.so not found with tensorflow docker image
When running with `tensorflow/tensorflow` docker image from docker hub, I got the following error message

```
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcublas.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcudnn.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcufft.so locally
I tensorflow/stream_executor/dso_loader.cc:102] Couldn't open CUDA library libcuda.so.1. LD_LIBRARY_PATH: 
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:160] hostname: 6b9e3874a772
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:185] libcuda reported version is: Not found: was unable to find libcuda.so DSO loaded into this program
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:356] driver version file contents: """NVRM version: NVIDIA UNIX x86_64 Kernel Module  367.27  Thu Jun  9 18:53:27 PDT 2016
GCC version:  gcc version 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.3) 
"""
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:189] kernel reported version is: 367.27.0
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:1077] LD_LIBRARY_PATH: 
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:1078] failed to find libcuda.so on this system: Failed precondition: could not dlopen DSO: libcuda.so.1; dlerror: libnvidia-fatbinaryloader.so.367.27: cannot open shared object file: No such file or directory
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcurand.so locally
```

The codalab command is 

```
cl run --request-queue jag --request-docker-image tensorflow/tensorflow:gpu -t "python -c 'import tensorflow; print tensorflow.__version__'"
```

kashizui:
Use nvidia-docker-plugin if available to set up the container creation request for GPU support. Otherwise, fall back to original hacks for GPU support.

Motivation for using nvidia-docker as opposed to other custom fixes:
https://github.com/NVIDIA/nvidia-docker/wiki/Why%20NVIDIA%20Docker

For more information about what nvidia-docker provides:
https://github.com/NVIDIA/nvidia-docker/wiki/Internals

Fixes #591

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -71,7 +71,7 @@ to run the worker from the Docker shell.
         try:
             self._libcuda = None
             for lib in subprocess.check_output(['/sbin/ldconfig', '-p']).split('\n'):
-                if 'x86-64' in lib and lib.endswith('libcuda.so'):
+                if 'x86-64' not in lib and lib.endswith('libcuda.so'):
                     self._libcuda = os.path.realpath(lib.split(' => ')[-1])
         except OSError:
             # ldconfig isn't available on Mac OS X. Let's just say that we
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -136,7 +136,7 @@ nvidia-docker-plugin not available, defaulting to basic GPU support.
         devices = [{
             "PathOnHost": device_path,
             "PathInContainer": device_path,
-            "CgroupPermissions": "mrw",  # FIXME: is this correct?
+            "CgroupPermissions": "mrw",
         } for device_path in cli_args['Devices']]
 
         # Set configurations in request json
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -94,14 +94,8 @@ No ldconfig found. Not loading libcuda libraries.
 nvidia-docker-plugin not available, defaulting to basic GPU support.
 """
             self._use_nvidia_docker = False
-            # DEBUG
-            with open('/tmp/sckoo_test_output', 'wb') as outfile:
-                outfile.write(e.message + '\n')
         else:
             self._use_nvidia_docker = True
-            # DEBUG
-            with open('/tmp/sckoo_test_output', 'wb') as outfile:
-                outfile.write('nvidia-docker-plugin available')
 
     def _create_nvidia_docker_connection(self):
         return httplib.HTTPConnection("localhost:3476")
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -44,11 +44,10 @@ class DockerClient(object):
     -----------
     DockerClient tries its best to support runs that require GPUs.
     During initialization, DockerClient checks to see if nvidia-docker-plugin
-    is available by contacting the REST API at the address specified by the
-    environment variable `NV_HOST`, or the default `localhost:3476`. If the
-    plugin is available, then DockerClient will query the REST API for
-    information about the volumes and devices that it should specify in the
-    Docker container creation request.
+    is available on the host machine by contacting the REST API at the default
+    address `localhost:3476`. If the plugin is available, then DockerClient will
+    query the REST API for information about the volumes and devices that it
+    should specify in the Docker container creation request.
 
     If the plugin is not available, we do a bit of manual work to attempt to
     support GPU jobs. In particular, DockerClient will query `ldconfig` to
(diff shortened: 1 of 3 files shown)


worker/docker_client.py
```diff
@@ -86,6 +86,20 @@ def __init__(self):
...
+        except DockerException as e:
+            print >> sys.stderr, """
+nvidia-docker-plugin not available, defaulting to basic GPU support.
+"""
+            self._use_nvidia_docker = False
+        else:
+            self._use_nvidia_docker = True
+
+    def _create_nvidia_docker_connection(self):
+        return httplib.HTTPConnection("localhost:3476")
```
kashizui: Place address in string constant

worker/docker_client.py
```diff
@@ -94,6 +108,45 @@ def _create_connection(self):
...
+            cli_response = conn.getresponse()
+            if cli_response.status != 200:
+                raise DockerException(cli_response.read())
+            cli_args = json.loads(cli_response.read())
+
+        # Build device jsons
+        devices = [{
+            "PathOnHost": device_path,
+            "PathInContainer": device_path,
+            "CgroupPermissions": "mrw",  # FIXME: is this correct?
```
kashizui: yes this is correct

kashizui:
Waiting for feedback on how this works. (thread in #591)

percyliang:
Cool...

1) Per our previous discussion, isn't NVIDIA docker a temporary solution?  I guess the amount of code here isn't too much, so I guess it's fine.

2) Could you explain what nvidia-docker provides over docker?  It seems like we have to do less hacking, which is good.  But in terms of functionality, it seems like we were able to run some GPU jobs with docker, but just not all of them?

kashizui:
Working on this helped me understand what nvidia-docker does even better, and I think I could potentially replicate its functionality. Doing so will allow us to support GPUs on any machine without installing nvidia-docker. 

The problem that nvidia-docker is trying to solve is, allowing Docker images/containers to be agnostic to its hardware re: GPUs. The naive solution would be to install all the NVIDIA drivers in the Docker image. But this will lock the image in with a specific machine configuration (e.g. NVIDIA driver/hardware version). `nvidia-docker` instead finds all the necessary driver files on the host machine, creates hard links to all of them in a Docker volume, and mounts the volume in the container.
https://github.com/NVIDIA/nvidia-docker/wiki/NVIDIA-driver

I'm still stumped as to why our current solution doesn't work, but nvidia-docker does...  I'm trying more experiments to figure  it out. Maybe something involving version mismatches, or needing to mount more than just libcuda? Also not sure why it worked before but not now.

Notes for myself on what the final solution will have to do:

- Ensure that `/dev/nvidia-uvm` is available by calling `nvidia-modprobe -u -c=0`
- Parse the output of `ldconfig -p` to get all the paths of the necessary driver files ([according to logic here](https://github.com/NVIDIA/nvidia-docker/blob/93bb65de7fc349e6de9f27abdaa75875f5572b17/tools/src/nvidia/volumes.go#L173-L211)) and mount them in the container somehow ([example code that creates links in a Docker volume](https://github.com/NVIDIA/nvidia-docker/blob/93bb65de7fc349e6de9f27abdaa75875f5572b17/tools/src/nvidia/volumes.go#L219-L275)). Creating a Docker volume would probably be the best solution (and is more efficient for long-lived workers) since it is probably a bad idea to mount so many volume bindings.

kashizui:
@percyliang do you know what kind of GPU jobs the worker has supported in the past?

kashizui:
Wow the problem was staring me in the eye the entire time (in fact it was the title of the original issue but for some reason I had disregarded it...). It looks like it was because libcuda depends on libnvidia-fatbinaryloader, which was not loaded in our original solution.

```
codalab@jagupard8:~% readelf -d /usr/lib/x86_64-linux-gnu/libcuda.so

Dynamic section at offset 0x71c588 contains 27 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libnvidia-fatbinaryloader.so.375.20]
```

worker/docker_client.py
```diff
@@ -40,10 +40,15 @@ class DockerClient(object):
     """
     Methods for talking to Docker.
```
percyliang, line 41: Add some more high-level documentation on trying to use nvidia-docker and what our docker solution is.  Say something about trying to use GPUs.

Separately, on the Wiki, we should add some instructions on how to install/use nvidia-docker.

worker/docker_client.py
```diff
@@ -40,10 +40,15 @@ class DockerClient(object):
...
     """
+    LIBCUDA_DIR = '/usr/lib/x86_64-linux-gnu/'
+
     def __init__(self):
         self._docker_host = os.environ.get('DOCKER_HOST') or None
+        self._nvidia_docker_host = os.environ.get('NVIDIA_DOCKER_HOST', 'localhost:3476')
```
percyliang: Why the asymmetry (we have a default NVIDIA_DOCKER_HOST but not a DOCKER_HOST)?

percyliang (approved):
LGTM modulo comments

worker/docker_client.py
kashizui: Because the nvidia-docker-plugin server address has a default (https://github.com/NVIDIA/nvidia-docker/wiki/nvidia-docker-plugin) but the docker daemon address doesn't have a default and is often different across hosts

PR #628, opened 2017-01-08
Fix for displaying [Object object] for empty files

maxwang7:
Issue description: https://github.com/codalab/codalab-worksheets/issues/222

The problem I found was that in the interpret_file_gen() function, an info dictionary is built with information that is later used to form the output string, but when the info dictionary is an empty dictionary it is not properly processed. So, I simply set info to None if it is an empty dictionary.

Some things I would like comments on:
- [ ] Should I set info to None at the position I have it now or instead after line 470? I see arguments as to why that would be the more appropriate place for this code.
- [ ] Should I unit test this change? The test coverage on this file doesn't seem to cover this function yet: https://github.com/codalab/codalab-cli/blob/f33ff172db28f592aac97cf75c0911040ac3bf22/tests/lib/worksheet_util_test.py.

@percyliang @kashizui

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -443,7 +443,9 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if all('\t' in x for x in contents):
+            if isinstance(contents, list) and len(contents) == 0:
+              info = ''
+            elif all('\t' in x for x in contents):
                 # Tab-separated file (key\tvalue\nkey\tvalue...)
                 info = {}
                 for x in contents:
@@ -460,8 +462,6 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                     except:
                         # Plain text file
                         info = ''.join(contents)
-            if not info: # info is an empty dictionary
-                info = None
         else:
             info = None
         target_cache[target] = info
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -460,6 +460,8 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                     except:
                         # Plain text file
                         info = ''.join(contents)
+            if not info: # info is an empty dictionary
+                info = None
         else:
             info = None
         target_cache[target] = info
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -443,7 +443,7 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if isinstance(contents, list) and len(contents) == 0:
+            if not contents: # contents is empty list
               info = ''
             elif all('\t' in x for x in contents):
                 # Tab-separated file (key\tvalue\nkey\tvalue...)
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -443,7 +443,7 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if not contents: # contents is empty list
+            if len(contents) == 0:
               info = ''
             elif all('\t' in x for x in contents):
                 # Tab-separated file (key\tvalue\nkey\tvalue...)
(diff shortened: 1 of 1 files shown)


maxwang7:
I didn't make a time estimate beforehand, it probably would've been 2-3 hours, but my work up to this point took 5-6 hours. A lot of that was becoming acquainted with the code base.

codalab/lib/worksheet_util.py
```diff
@@ -460,6 +460,8 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                     except:
                         # Plain text file
                         info = ''.join(contents)
+            if not info: # info is an empty dictionary
+                info = None
```
percyliang: In general, we want `info` to preserve information about the file, so don't want to just set it to None.  What's happening here is that `contents = []`.  In this case, we shouldn't be interpreting this as a TSV file, but a plain text file, which means, info should be set to the empty string.  I would add this as the first condition before line 446.  Python makes it very easy to be sloppy with types; it's good to always think about the semantics - difference between `{}, [], '', None`.

percyliang (changes requested):
Great - should be a simple change.

@kashizui I think we can get rid of lines 430, 432-434, right?

kashizui:
@percyliang We actually can't yet delete those lines since it is used by the legacy api. There may be a way to untangle it with some proper thought though.

kashizui (commented):
I agree w/ Percy's comments.
Maybe it makes sense to require a file be nonempty to be considered a TSV file?

maxwang7:
@percyliang @kashizui Thank you for the comments! I've made some changes that I believe are according to what was discussed. Some questions:
- I'm not understanding where the code checks for CSV or TSV files -- I've grep'd the repo and haven't found any useful keywords. Could someone explain to me how the code base checks for those things?
- Is the code I have written the proper we to type check for a Python list? I'm not sure whether to use isinstance() or `variable == []`.

codalab/lib/worksheet_util.py
```diff
@@ -443,7 +443,9 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if all('\t' in x for x in contents):
+            if isinstance(contents, list) and len(contents) == 0:
+              info = ''
+            elif all('\t' in x for x in contents):
```
kashizui, line 448: This is where the code checks that the file is a TSV

codalab/lib/worksheet_util.py
```diff
@@ -443,7 +443,9 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if all('\t' in x for x in contents):
+            if isinstance(contents, list) and len(contents) == 0:
```
kashizui: `contents` should be guaranteed to be a list here, so no need to check its class.

kashizui (changes requested):

codalab/lib/worksheet_util.py
kashizui: Just `if len(contents) == 0` or even just `if not contents` should be fine!

maxwang7:
@kashizui Code has been updated. I would like to write a test for all of this, to do so, would I need to mock out everything that's used in interpret_file_genpath() though?

codalab/lib/worksheet_util.py
```diff
@@ -443,7 +443,9 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if all('\t' in x for x in contents):
+            if not contents: # contents is empty list
+              info = ''
```
kashizui, line 447: use 4 spaces for indentation

kashizui (approved):
LGTM modulo comment

kashizui:
@maxwang7 Yeah the way it is written now will make it hard to test since you would have to mock out the client. But it's only two method calls (get_target_info and head_target) so if it's easy to write up a test very quickly feel free to do so, but not necessary for now.

codalab/lib/worksheet_util.py
```diff
@@ -443,7 +443,7 @@ def interpret_file_genpath(client, target_cache, bundle_uuid, genpath, post):
                 contents = client.head_target(target, MAX_LINES)
                 contents = map(base64.b64decode, contents)
 
-            if isinstance(contents, list) and len(contents) == 0:
+            if not contents: # contents is empty list
```
percyliang: I would write `len(contents) == 0` and remove the comment...  Better to have code be more explicit and self-documenting.

codalab/lib/worksheet_util.py
percyliang: You should make your editor default to 4 for Python.   Make sure you don't add tabs.  Indentation is really important to get right for Python!

maxwang7:
I'm going to go ahead and merge. Tests will be written in a separate commit.

Issue #235, opened 2015-11-16
timeout jobs


percyliang:
Jobs that are 'running' and 'queued' for more than a month (or haven't been updated for more than 1 day) should be marked 'failed'.

In `work_manager.py`, there is a TODO for this.  Call search_bundle_uuids to get all queued/running bundles and check the `last_updated` field of the bundle.

PR #629, opened 2017-01-09
Support datetime serialization in JSON responses

kashizui:
Fixes codalab/codalab-worksheets#235 (stats on worksheets.codalab.org doesn't load)

Also ensure that future JSON serialization errors send us emails.

### DIFF
--- a/codalab/rest/legacy.py
+++ b/codalab/rest/legacy.py
@@ -733,7 +733,7 @@ def get_worksheet_content(uuid):
     service = BundleService()
     bundle_uuids = request.query.getall('bundle_uuid')
     res = service.full_worksheet(uuid, bundle_uuids)
-    print >>sys.stderr, formatting.pretty_json(res)
+    print >>sys.stderr, res
     return res
 
 
--- a/codalab/rest/legacy.py
+++ b/codalab/rest/legacy.py
@@ -731,7 +731,6 @@ def post_worksheets_command():
     return json.dumps(result)
 
 
-import sys
 @get('/api/worksheets/<uuid:re:%s>/' % spec_util.UUID_STR)
 def get_worksheet_content(uuid):
     service = BundleService()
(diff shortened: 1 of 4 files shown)


kashizui:
@percyliang I don't really like my current solution, where do you think would be the best place to make sure that 'frozen' is properly serialized?

codalab/rest/legacy.py
```diff
@@ -145,6 +145,16 @@ def search_worksheets(keywords, worksheet_uuid=None):
...
+        # Hack: Make sure worksheet infos are JSON serializable.
+        # Since this method is only used for worksheet interpretation (i.e.
+        # resolving wsearch directives), these worksheet infos are never
+        # serialized properly through WorksheetSchema or the Worksheet
+        # ORMObject, and so the `frozen` attribute will cause a serialization
+        # error when we attempt to serialize the interpretation results into
+        # JSON.
+        for info in results:
+            if isinstance(info['frozen'], datetime):
+                info['frozen'] = str(info['frozen'])
```
percyliang: I wouldn't use str()...use isoformat().

I think the `BundleModel` should be responsible for returning things back which are JSON serializable.  This method should not be looking at the internals of `results`.

percyliang (changes requested):

kashizui:
+1 on isoformat().

I generally agree, but BundleModel.search_worksheets() is also used by code
that depends on datetime fields being actual datetime objects, so I had
avoided modifying the model method.

percyliang:
We could just convert the isoformat string back to a datetime when we need it, although that's a bit backwards...

Alternatively, we can have a general utility function that converts any JSON object with datetimes (and perhaps other non-serializable types) into a serializable JSON?

kashizui:
@percyliang Pushed new changes that handles all datetime objects in JSON response objects. The `default` function handles any nonstandard objects that the JSON encoder encounters and cannot serialize.

kashizui:
@percyliang reminder to review again

percyliang (approved):
Nice LGTM

PR #631, opened 2017-01-09
Increase request body size limit to 8 MiB

### LINKED ISSUE #627: worker checkin request too large
After a while, when the number of bundles for a worker grows to 15G or so and state.json is 353K, then worker is unable to checkin:

File "/usr/lib/python2.7/urllib2.py", line 527, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
BundleServiceException: Unable to check in with bundle service: Request Entity Too Large - Request to large

and jobs cannot be run on that worker anymore.

kashizui:
Fixes #627


percyliang (approved):
LGTM

PR #638, opened 2017-01-15
Worker now prints message on reconnect (#535)

### LINKED ISSUE #535: Worker should print message on reconnect
A worker should print a message when it successfully reconnects and checks in with the API server after exceptional conditions.

Current behavior:
Worker prints `BundleServiceException: Unable to check in with bundle service` repeatedly until reconnected.

Proposed behavior:
Worker additionally prints something like `Reconnected!` upon successful check-in after a failed check-in.

fabeschan:
A worker should print a message when it successfully reconnects and
checks in with the API server after exceptional conditions.

Current behavior:
Worker prints BundleServiceException: Unable to check in with bundle
service repeatedly until reconnected.

Proposed behavior:
Worker additionally prints something like Connected! upon successful
check-in

Fixes #535

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -49,10 +49,11 @@ class Worker(object):
         # when the Run class calls finish_run.
         self._runs_lock = threading.Lock()
         self._runs = {}
- 
+
         self._exiting_lock = threading.Lock()
         self._exiting = False
         self._should_upgrade = False
+        self._last_checkin_successful = False
 
     def run(self):
         if not self.shared_file_system:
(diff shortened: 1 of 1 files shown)


worker/worker.py
```diff
@@ -61,7 +62,12 @@ def run(self):
         while self._should_run():
             try:
                 self._checkin()
+                if self._last_checkin_successful == False:
```
percyliang: nit: pep8 says that this should probably be `not self._last_checkin_successful`

(Though they recommend doing the same for non-booleans, which I don't like.)

percyliang (approved):
LGTM modulo nit

worker/worker.py
fabeschan: Done! Will merge after CI run

Issue #224, opened 2015-11-12
run with defaults


percyliang:
When do 'cl run ... --defaults bundle', should inherit the dependencies and request\* parameters of the bundle.  The bundle acts as an environment which makes it to run a bunch of commands using the same environment without retyping everything.  Think more carefully about this design decision.

percyliang:
Introduces too much complexity.  These things should be handled outside CodaLab.

PR #643, opened 2017-01-21
Fix exitcode 127 error

kashizui:
Update docker container preamble commands to use chown instead of sudo to ensure proper file permissions. Most newer docker images don't have the sudo command, causing a 'command not found' error.

Also added an additional docker request to fetch container logs at the end of a run to check for any error messages that were previously missed (such as 'sudo: command not found').

Fixes codalab/codalab-worksheets#224

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -272,7 +272,7 @@ No ldconfig found. Not loading libcuda libraries.
             + '\'(%s) >stdout 2>stderr\'' % command.replace('\'', '\'"\'"\''),
             # Ensure that any created files are owned by the user/group that
             # owns the bundle directory, not root.
-            'chown -R $U_ID:$G_ID %s' % docker_bundle_path,
+            #'chown -R $U_ID:$G_ID %s' % docker_bundle_path,
         ]
 
         # Set up the volumes.
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -386,8 +386,11 @@ No ldconfig found. Not loading libcuda libraries.
             
             inspect_json = json.loads(inspect_response.read())
             if not inspect_json['State']['Running']:
-                logs = self._get_logs(container_id)
-                return (True, inspect_json['State']['ExitCode'], logs or None)
+                # If the logs are nonempty, then something might have gone
+                # wrong with the commands run before the user command,
+                # such as ldconfig.
+                failure_msg = self._get_logs(container_id) or None
+                return (True, inspect_json['State']['ExitCode'], failure_msg)
             return (False, None, None)
 
     @wrap_exception('Unable to delete Docker container')
@@ -400,6 +403,7 @@ No ldconfig found. Not loading libcuda libraries.
                 raise DockerException(delete_response.read())
 
     def _get_logs(self, container_id):
+        """Read stdout and stderr of the given container."""
         with closing(self._create_connection()) as conn:
             conn.request('GET', '/containers/%s/logs?stdout=1&stderr=1' % container_id)
             logs_response = conn.getresponse()
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -43,27 +43,21 @@ class DockerClient(object):
 
     GPU Support
     -----------
-    DockerClient tries its best to support runs that require GPUs.
+    DockerClient relies on nvidia-docker-plugin for runs that require GPUs.
     During initialization, DockerClient checks to see if nvidia-docker-plugin
     is available on the host machine by contacting the REST API at the default
     address `localhost:3476`. If the plugin is available, then DockerClient will
     query the REST API for information about the volumes and devices that it
     should specify in the Docker container creation request.
 
-    If the plugin is not available, we do a bit of manual work to attempt to
-    support GPU jobs. In particular, DockerClient will query `ldconfig` to
-    see if libcuda is available on the host machine, then if it is available,
-    manually mount `libcuda.so` and NVIDIA character devices in the containers.
-    Many GPU jobs will require more than just libcuda, so it is recommended that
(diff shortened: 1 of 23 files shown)


worker/docker_client.py
```diff
@@ -174,23 +174,24 @@ def start_container(self, bundle_path, uuid, command, docker_image,
...
-            # any, should happen when bash executes it. Note, since the user's
-            # command can have single quotes we need to escape them.
-            '"[ -e $BASHRC ] && . $BASHRC; "' +
-            '"cd %s; "' % docker_bundle_path +
-            '"export HOME=%s; "' % docker_bundle_path +
-            '\'(%s) >stdout 2>stderr\'' % command.replace('\'', '\'"\'"\''),
+            # appears in single quotes since we do not. The expansion there,
+            # if any, should happen when bash executes it. Note, since the
+            # user's command can have single quotes we need to escape them.
+            'bash -c '
```
percyliang, line 224: Not having to do `sudo` is actually nicer since some docker images assume that you're running as root, so this is a good change.

worker/docker_client.py
```diff
@@ -174,23 +174,24 @@ def start_container(self, bundle_path, uuid, command, docker_image,
...
+            + '"[ -e $BASHRC ] && . $BASHRC; "'
+            + '"cd %s; "' % docker_bundle_path
+            + '"export HOME=%s; "' % docker_bundle_path
+            + '\'(%s) >stdout 2>stderr\'' % command.replace('\'', '\'"\'"\''),
+            # Ensure that any created files are owned by the user/group that
+            # owns the bundle directory, not root.
+            'chown -R $U_ID:$G_ID %s' % docker_bundle_path,
```
percyliang: Will this command always get executed?  (I think so.)

Has this been tested on the NLP machines with the shared filesystem?  I wonder if we'll always have permissions to do this `chown`...  If this fails, can we make the error go to stderr?  Actually same for the `. $BASHRC` command above...

percyliang (approved):
LGTM but just want to really make sure that things work before we push this

worker/docker_client.py
kashizui: You were right, the chown fails on the NLP machines.
https://codalab.stanford.edu/worksheets/0x69c74999b0e8413eb93d1b0f5ddb6b3d/

As you can see in that bundle, the error gets collected into the `failure_message` field, and same for any resulting error from `. $BASHRC`.

kashizui: Are you saying that if the `chown` fails, we should just pipe the error into `/dev/null`?

percyliang: Definitely not!  We need to be able to change the permissions on the NLP machines to the desired user/group or else the CodaLab user outside docker (not running as root) won't be able access/delete those bundles.

I guess it's a bit strange that 'root' doesn't actually have a superset of the privileges of the user running docker.  Not sure what the best solution here is...

kashizui: I had an earlier solution that created the Docker container to run as the proper user/group, but then running `ldconfig` didn't work because it required root permissions. And `su` complained that it has to be run "in a terminal." Still trying to untangle this..

percyliang:
Another option is to see if we can run CodaLab on the NLP machines with
more root-like permission - probably have to ping sysadmins about that.

kashizui:
I'm not sure if that will necessarily help, since the problem isn't that Docker itself lacks permissions to modify the bundle files.
One solution I'm thinking of is just not to run `ldconfig` inside the container -- if we remove the GPU hacks and only rely on `nvidia-docker`, then we don't need to run `ldconfig`, which requires root.
Then we can directly set the user/group of the process inside the container when we create it using the Docker API without any hacks.

percyliang:
Sounds good to me.  How are our GPU hacks (for linking libraries into
docker) different from `nvidia-docker`'s though?

kashizui:
Main difference is that the hack only mounts a tiny subset of the nvidia libraries (libcuda only), while nvidia-docker does it for all nvidia libraries available on the host, builds a Docker volume for the libraries, creating all the necessary hard links, and even excluding libraries known to cause problems.

percyliang:
Ah, cool, thanks.  Btw, did we get nvidia-docker installed on all the jags?

kashizui:
Yep with a couple exceptions, I forgot to forward you the email.

kashizui:
Removed libcuda hacks and tested it on a test CodaLab instance on NLP server. Works with anaconda image, and runs as the current user and group without error.

```
bundle_type               : run
uuid                      : 0x29e6032baac34d8fa7e7ad4fe9c7c972
data_hash                 : 0x5a4b5e843062f04564d35179243a6eb1e3baadc9
state                     : ready
command                   : id
owner                     : codalab(0)
name                      : run-id
created                   : 2017-01-21 23:46:32
data_size                 : 4.0k
allow_failed_dependencies : False
request_docker_image      : continuumio/anaconda:latest
request_cpus              : 0
request_gpus              : 0
request_priority          : 0
request_network           : False
time                      : 1.2s
time_user                 : 0.0s
time_system               : 0.0s
memory                    : 112k
memory_max                : 112k
started                   : 2017-01-21 23:46:36
last_updated              : 2017-01-21 23:46:38
run_status                : Finished
exitcode                  : 0
job_handle                : 1146928.scail.stanford.edu
remote                    : john9.stanford.edu
children:
host_worksheets:
  home-codalab(0x17d5d92d660c44d89a17441ef36266d9)
permission: all
group_permissions:
  public(0x5940c2):read
=== contents preview ===
name    perm  size
------------------
stderr  0644     0
stdout  0644    33
=== stdout preview ===
uid=16356 gid=20004 groups=20004
=== stderr preview ===
```

kashizui:
Reopening as #647 to merge into develop instead (which has the nvidia-docker support).

PR #645, opened 2017-01-22
Increase default soft timeout to 15s in monitor.py

### LINKED ISSUE #635: fix monitor.py so we don't get so many emails
Either figure out why some runs time out (what fraction?) and fix that or raise the time limit if we think it's benign.

fabeschan:
Fixes #635 

This should get rid of ~90% of the timeout emails we get.


percyliang:
LGTM

fabeschan:
Some api call failed to connect then another succeeded. I think I should try triggering the build again. @percyliang can you give me permissions on CircleCI to do that?

PR #647, opened 2017-01-22
Run Docker container as correct user/group.

### LINKED ISSUE #537: docker environment variables not visible to command
Docker images currently can define environment variables, but these are not picked up by CodaLab because we run commands inside `sudo bash -c ...` (see docker_client.py).  The problem is if we don't do sudo, then the files generated by the commands might not be readable by the outside world.  Though this might be okay, as long as we then change the permissions on everything once the run finishes.  The big picture is that any docker container people build should work in CodaLab without any special considerations.

kashizui:
- Remove use of `sudo`, and replace with an explicit user/group specification in the container creation request. `sudo` is not available on all Docker images (i.e. it is not bundled with the ubuntu base image), so calls to `sudo` can cause an exitcode 127 error. Use of `sudo` also bulldozed environment variable defined in docker images (#537)
- Fetch container logs at the end of a run to check for any error messages that were previously missed (such as 'sudo: command not found'), will speed up debugging in the future.
- Remove libcuda hacks to avoid call to `ldconfig`.
- Separate out 'default' set of test-cli modules.

Fixes codalab/codalab-worksheets#224
Fixes #537 

Created new PR to replace #643 and merge into develop instead.

Tested it on a test CodaLab instance on NLP server. Works with anaconda image, and runs as the current user and group without error.

```
bundle_type               : run
uuid                      : 0x29e6032baac34d8fa7e7ad4fe9c7c972
data_hash                 : 0x5a4b5e843062f04564d35179243a6eb1e3baadc9
state                     : ready
command                   : id
owner                     : codalab(0)
name                      : run-id
created                   : 2017-01-21 23:46:32
data_size                 : 4.0k
allow_failed_dependencies : False
request_docker_image      : continuumio/anaconda:latest
request_cpus              : 0
request_gpus              : 0
request_priority          : 0
request_network           : False
time                      : 1.2s
time_user                 : 0.0s
time_system               : 0.0s
memory                    : 112k
memory_max                : 112k
started                   : 2017-01-21 23:46:36
last_updated              : 2017-01-21 23:46:38
run_status                : Finished
exitcode                  : 0
job_handle                : 1146928.scail.stanford.edu
remote                    : john9.stanford.edu
children:
host_worksheets:
  home-codalab(0x17d5d92d660c44d89a17441ef36266d9)
permission: all
group_permissions:
  public(0x5940c2):read
=== contents preview ===
name    perm  size
------------------
stderr  0644     0
stdout  0644    33
=== stdout preview ===
uid=16356 gid=20004 groups=20004
=== stderr preview ===
```

Also confirmed that environment variables defined in Docker images are now visible:
```
bundle_type               : run
uuid                      : 0xff9b96b8e61943afbef1da2c548e4274
data_hash                 : <none>
state                     : running
command                   : bash -c "echo $ENVTEST"
owner                     : codalab(0)
name                      : run-bash
created                   : 2017-01-22 01:52:47
data_size                 : 4.0k
allow_failed_dependencies : False
request_docker_image      : sckoo/envtest:latest
request_cpus              : 0
request_gpus              : 0
request_priority          : 0
request_network           : False
time                      : 8.7s
time_user                 : 0.0s
time_system               : 0.0s
memory                    : 268k
memory_max                : 268k
started                   : 2017-01-22 01:52:52
last_updated              : 2017-01-22 01:53:01
run_status                : Running
job_handle                : 1147404.scail.stanford.edu
remote                    : john9.stanford.edu
children:
host_worksheets:
  home-codalab(0x17d5d92d660c44d89a17441ef36266d9)
permission: all
group_permissions:
  public(0x5940c2):read
=== contents preview ===
name    perm  size
------------------
stderr  0644     0
stdout  0644     5
=== stdout preview ===
GOOD
=== stderr preview ===
```

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -319,10 +319,16 @@ No ldconfig found. Not loading libcuda libraries.
                 'PathInContainer': device,
                 'CgroupPermissions': 'mrw'})
 
+        # Get user/group that owns the bundle directory
+        bundle_stat = os.stat(bundle_path)
+        uid = bundle_stat.st_uid
+        gid = bundle_stat.st_gid
+
         # Create the container.
         create_request = {
             'Cmd': ['bash', '-c', '; '.join(docker_commands)],
             'Image': docker_image,
+            'User': '%s:%s' % (uid, gid),
             'HostConfig': {
                 'Binds': volume_bindings,
                 'Devices': devices,
@@ -415,7 +421,12 @@ No ldconfig found. Not loading libcuda libraries.
                 # wrong with the commands run before the user command,
                 # such as ldconfig.
                 _, stderr = self._get_logs(container_id)
-                return (True, inspect_json['State']['ExitCode'], stderr or None)
+                # Strip non-ASCII chars since failure_message is not Unicode
+                if len(stderr) > 0:
+                    failure_msg = stderr.encode('ascii', 'ignore')
+                else:
+                    failure_msg = None
+                return (True, inspect_json['State']['ExitCode'], failure_msg)
             return (False, None, None)
 
     @wrap_exception('Unable to delete Docker container')
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -44,20 +44,17 @@ class DockerClient(object):
 
     GPU Support
     -----------
-    DockerClient tries its best to support runs that require GPUs.
+    DockerClient relies on nvidia-docker-plugin for runs that require GPUs.
     During initialization, DockerClient checks to see if nvidia-docker-plugin
     is available on the host machine by contacting the REST API at the default
     address `localhost:3476`. If the plugin is available, then DockerClient will
     query the REST API for information about the volumes and devices that it
     should specify in the Docker container creation request.
 
-    If the plugin is not available, we do a bit of manual work to attempt to
-    support GPU jobs. In particular, DockerClient will query `ldconfig` to
-    see if libcuda is available on the host machine, then if it is available,
-    manually mount `libcuda.so` and NVIDIA character devices in the containers.
-    Many GPU jobs will require more than just libcuda, so it is recommended that
-    you install nvidia-docker on the host machines of workers that should
-    support GPU jobs.
+    If you want your worker to support GPU jobs, make sure to install
+    nvidia-docker on the host machine. You can find the project and installation
+    instructions on GitHub:
+    https://github.com/NVIDIA/nvidia-docker
 
     DockerClient will read the CUDA_VISIBLE_DEVICES environment variable and
     only mount the GPU device corresponding to the indices listed in the
@@ -66,9 +63,6 @@ class DockerClient(object):
     # Where to look for nvidia-docker-plugin
     # https://github.com/NVIDIA/nvidia-docker/wiki/nvidia-docker-plugin
     NV_HOST = 'localhost:3476'
-
-    # Where to mount libcuda inside the container
-    LIBCUDA_DIR = '/usr/lib/x86_64-linux-gnu/'
 
     def __init__(self):
         self._docker_host = os.environ.get('DOCKER_HOST') or None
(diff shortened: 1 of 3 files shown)


kashizui:
@percyliang please take a quick second look!

worker/docker_client.py
```diff
@@ -265,65 +223,43 @@ def start_container(self, bundle_path, uuid, command, docker_image,
...
+        # Get user/group that owns the bundle directory
+        # Then we can ensure that any created files are owned by the user/group
+        # that owns the bundle directory, not root.
+        bundle_stat = os.stat(bundle_path)
+        uid = bundle_stat.st_uid
+        gid = bundle_stat.st_gid
 
         # Create the container.
         create_request = {
             'Cmd': ['bash', '-c', '; '.join(docker_commands)],
```
percyliang, line 249: We have a `bash -c` here and also inside `docker_commands`...can this be combined into one?

percyliang (changes requested):
Nice, this is much cleaner!

worker/docker_client.py
```diff
@@ -265,65 +223,43 @@ def start_container(self, bundle_path, uuid, command, docker_image,
...
-            # command can have single quotes we need to escape them.
-            '"[ -e $BASHRC ] && . $BASHRC; "' +
-            '"cd %s; "' % docker_bundle_path +
-            '"export HOME=%s; "' % docker_bundle_path +
-            '\'(%s) >stdout 2>stderr\'' % command.replace('\'', '\'"\'"\''),
+            # appears in single quotes since we do not. The expansion there,
+            # if any, should happen when bash executes it. Note, since the
+            # user's command can have single quotes we need to escape them.
+            'bash -c '
+            + '"[ -e $BASHRC ] && . $BASHRC; "'
```
fabeschan: Is it still necessary to source `.bashrc` if it fixes #537 ?

worker/docker_client.py
```diff
@@ -409,8 +345,17 @@ def check_finished(self, container_id):
             
             inspect_json = json.loads(inspect_response.read())
             if not inspect_json['State']['Running']:
-                return (True, inspect_json['State']['ExitCode'], None)
-            return (False, None, None)   
+                # If the logs are nonempty, then something might have gone
+                # wrong with the commands run before the user command,
+                # such as bash or cd.
```
fabeschan, line 345: Is it feasible to instead check the exitcode rather than checking if logs are empty? https://docs.docker.com/engine/reference/run/#/exit-status

fabeschan (changes requested):

worker/docker_client.py
kashizui: I think so -- I'm working on and testing out a simplified version of this.

worker/docker_client.py
kashizui: You're right, this is no longer necessary. I'll remove this, and also make sure to update [our Torch image](https://github.com/codalab/codalab-cli/tree/master/docker/torch) to use the `ENV` directive rather than use the bashrc hack.

percyliang: The `codalab/ubuntu:1.9` relies in `.bashrc`, so I'd probably not remove it yet unless we rebuild that image.  In general, we need to redo our docker images with Dockerfile, but that's a separate issue.

worker/docker_client.py
kashizui: We do already check the exitcode elsewhere -- I added this in in order to capture errors that were previous hidden (for example the cryptic "Exit code 127" would have been easier to debug if we had been checking the logs of the container, which would have told us that the `sudo` command was not found).

However, I think I might remove this, now that I'm working on simplifying the setup commands (i.e. no more sourcing `bashrc`, or `ldconfig` and `stat`).

worker/docker_client.py
kashizui: Ah good to know.

kashizui:
Next steps:

- [ ] Remove extra level of `bash -c` indirection
- [ ] Consider removing log reading?
- [ ] Research and implement solution to Docker volume file permissions issue (goal: allow container to run as the user defined in the image [default: root], and at the same time allow bundle directory to be writable by the container, and its new contents readable outside of the container).

kashizui:
Don't worry about user, just check it in

worker/docker_client.py
kashizui: Done!

kashizui:
@percyliang @fabeschan Addressed comments, please take another look!

worker/docker_client.py
```diff
@@ -224,19 +224,10 @@ def start_container(self, bundle_path, uuid, command, docker_image,
...
-            # since we want environment variables to be expanded. The last
-            # appears in single quotes since we do not. The expansion there,
-            # if any, should happen when bash executes it. Note, since the
-            # user's command can have single quotes we need to escape them.
-            'bash -c '
-            + '"[ -e $BASHRC ] && . $BASHRC; "'
-            + '"cd %s; "' % docker_bundle_path
-            + '"export HOME=%s; "' % docker_bundle_path
-            + '\'(%s) >stdout 2>stderr\'' % command.replace('\'', '\'"\'"\''),
+            # TODO: Remove when codalab/torch uses ENV directives
```
percyliang, line 227: Oh, we're so close to having this be clean!

But I think this is not going to work for the codalab/ubuntu:1.9 image either, which relies on sourcing .bashrc before cd'ing into docker_bundle_path...

Probably need to have to source .bashrc that exists in several possible directories...
Can you try this out on several docker images to make sure that they all work?

This should be part of our test - even if it is not a formal process...just have a list of docker images that we commit ourselves to checking when we make changes like this.

kashizui: Yikes good to know. I just opened up the codalab/ubuntu:1.9 image and the bashrc file is also stored at ` /user/.bashrc`. 

I did actually test on this image before pushing, but I would need to devise more substantial tests (e.g. programs that use Scala) in order to test whether or not the environment variables work..

kashizui:
@percyliang Works with the `codalab/ubuntu:1.8` image! 

```
cl run -t 'echo $SCALA_HOME'                                                                                                                               17-01-30 18:41:18
0xa4c3496fb9ea45e19ae79738909a8a2b
/opt/scala/current
```

kashizui:
@percyliang Added tests for the bashrc issue, please review

percyliang (approved):
LGTM

Does this docker test run by default?  I could imagine it being kind of suboptimal to be a default, as it requires pulling docker images (we might add more images later).

kashizui:
I agree it could cause issues -- I was worried it wasn't even going to pass in case of disk space issues on CircleCI.

We could split the tests up into a 'default' set and an 'all' superset, then just run 'default' on CircleCI, and run 'all' manually when necessary (e.g. modifying worker code).

percyliang:
default + all sounds good

PR #649, opened 2017-01-23
Show docker image digest (497)

### LINKED ISSUE #497: don't allow docker images without version numbers
--request-docker-image ubuntu or ubuntu:latest shouldn't be allowed because this does not nail down a particular docker images.

Of course, these docker images are simply tags, which can get deleted, but at least this is moving in the right direction.

Better yet, maybe should have Dockerfiles be a more essential part of the story.

fabeschan:
fixes #497

### DIFF
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -23,7 +23,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS = list(DerivedBundle.METADATA_SPECS)
     # Note that these are strings, which need to be parsed
     # Request a machine with this much resources and don't let run exceed these resources
-    METADATA_SPECS.append(MetadataSpec('request_docker_image', basestring, 'Which docker image (e.g., codalab/ubuntu:1.9) or digest we wish to use.', completer=DockerImagesCompleter))
+    METADATA_SPECS.append(MetadataSpec('request_docker_image', basestring, 'Which docker image (either tag or digest, e.g., codalab/ubuntu:1.9) we wish to use.', completer=DockerImagesCompleter))
     METADATA_SPECS.append(MetadataSpec('request_time', basestring, 'Amount of time (e.g., 3, 3m, 3h, 3d) allowed for this run.', formatting='duration'))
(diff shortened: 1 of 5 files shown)


percyliang:
As mentioned in our other discussion, I think `request_docker_image` should be arbitrary, but `docker_image` should be the SHA256.

codalab/bundles/run_bundle.py
```diff
@@ -23,7 +23,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS = list(DerivedBundle.METADATA_SPECS)
     # Note that these are strings, which need to be parsed
     # Request a machine with this much resources and don't let run exceed these resources
-    METADATA_SPECS.append(MetadataSpec('request_docker_image', basestring, 'Which docker image (e.g., codalab/ubuntu:1.9) we wish to use.', completer=DockerImagesCompleter))
+    METADATA_SPECS.append(MetadataSpec('request_docker_image', basestring, 'Which docker image (e.g., codalab/ubuntu:1.9) or digest we wish to use.', completer=DockerImagesCompleter))
```
percyliang: Be more specific - "which docker image, either tag or digest, to use"

codalab/lib/bundle_cli.py
```diff
@@ -1899,6 +1899,10 @@ def recurse(old_bundle_uuid):
                         new_metadata['name'] = new_output_name + '-' + \
                                                old_info['metadata']['name']
 
+                # Put docker_image in requested_docker_image if it is present and this is a run bundle
```
percyliang: Document more based on our discussion: say that by default, we copy request_docker_image, but this is inappropriate because tags can point to different digest (give an example).

percyliang (approved):
LGTM - just add a few more comments - it's always good to capture our thought process in code, or some of these decisions will be a mystery two years from now.

codalab/bundles/run_bundle.py
fabeschan: Done!

codalab/lib/bundle_cli.py
fabeschan: Done!

PR #650, opened 2017-01-23
First dab at stress-test.py

### LINKED ISSUE #632: Stress testing
@percyliang:
>  We should probably have a much more aggressive stress test of CodaLab - have 20 jobs all try to bombard it with random things, so we can even run into out of disk failures, etc.

fabeschan:
Fixes #632

### DIFF
--- a/stress-test.py
+++ b/stress-test.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python
+
+import os, sys
+import datetime
+from collections import defaultdict
+from smtplib import SMTP
+from email.mime.text import MIMEText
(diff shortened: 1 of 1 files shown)


percyliang:
This has a lot of similarity with `monitor.py`...
1) The stress test doesn't actually need to have all the same scaffolding code since it's just going to be run manually.  The main thing is just having the right set of commands.  Even the threading is just a nice to have - you can always just launch 10 copies of the script.
2) If we did need the scafolding, we should really do this inside monitor.py or otherwise have more code sharing.

Issue #246, opened 2015-11-21
editing metadata of worksheet doesn't throw not authenticated error


kashizui:

kashizui:
Not the real issue.

Created new one:
https://github.com/codalab/codalab/issues/1429

PR #653, opened 2017-01-25
Alphabetize Permissions for legacy and REST APIs

maxwang7:
Fixes https://github.com/codalab/codalab-worksheets/issues/209 along with https://github.com/codalab/codalab-worksheets/pull/246.

Adds sorting functionality at the server-layer for group permissions for both bundles and worksheets. Implemented in both the REST and legacy APIs.

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -120,6 +120,7 @@ def build_bundles_document(bundle_uuids):
     # Include permissions
     for bundle in bundles:
         json_api_include(document, BundlePermissionSchema(), bundle['group_permissions'])
+        bundle['group_permissions'] = sorted(bundle['group_permissions'], key=lambda x: x['group_name'])
 
     # Include child bundles
     children_uuids = set(c['uuid'] for bundle in bundles for c in bundle['children'])
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -120,7 +120,6 @@ def build_bundles_document(bundle_uuids):
     # Include permissions
     for bundle in bundles:
         json_api_include(document, BundlePermissionSchema(), bundle['group_permissions'])
-        bundle['group_permissions'] = sorted(bundle['group_permissions'], key=lambda x: x['group_name'])
 
     # Include child bundles
     children_uuids = set(c['uuid'] for bundle in bundles for c in bundle['children'])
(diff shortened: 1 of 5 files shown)


percyliang (commented):
Can the sorting be done no in /rest/* but in bundle_model.py when you construct the worksheet_info and bundle_info objects?  That way, you only need the logic in one place.

Alternatively, this can be done on the frontend right before we display.

In general, putting conversion code at one of the two ends is better than in the middle, because then you just have to reason about one format most of the time.

codalab/rest/bundles.py
```diff
@@ -120,6 +120,7 @@ def build_bundles_document(bundle_uuids):
     # Include permissions
     for bundle in bundles:
         json_api_include(document, BundlePermissionSchema(), bundle['group_permissions'])
+        bundle['group_permissions'] = sorted(bundle['group_permissions'], key=lambda x: x['group_name'])
```
kashizui: This won't get included in the response, this line should go above the line that precedes it.

kashizui (changes requested):
Looking through the code now, it looks like the place to do this is in `BundleModel.batch_get_group_permissions()`. You might even be able to offload the sorting to MySQL in the query. 😛

codalab/rest/bundles.py
maxwang7: Ah, good to know. If I make the change to the BundleModel query though, I won't need this line anymore, right?

kashizui: That's right sorry I made it confusing

maxwang7:
I added an order_by statement in the batch_get_group_permissions() file. I want to do some more testing before committing these changes because something's not working locally for me.

Also, from my reading of the code, both the legacy API and the REST API use the same model, can someone confirm that? If so, I'll remove the other custom sorting functionality that I've written.

kashizui:
yep they use the same model

maxwang7:
@kashizui I've tested the changes and they're working in my test environment. I've also verified that my test environment is taking up the newest changes. When this is OK'd, I'll merge the corresponding commit on CodaLab worksheets.

codalab/rest/bundles.py
kashizui: Don't forget to delete this line!

kashizui (approved):
LGTM modulo comment

PR #660, opened 2017-01-27
Support Local Development

fabeschan:
hi

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -83,6 +83,10 @@ from codalab.lib.bundle_store import (
 from codalab.lib.print_util import FileTransferProgress
 from worker.file_util import un_tar_directory
 
+from codalab.lib.spec_util import generate_uuid
+from worker.docker_client import DockerClient
+from worker.file_util import remove_path
+
 # Formatting Constants
 GLOBAL_SPEC_FORMAT = "[<alias>::|<address>::](<uuid>|<name>)"
 ADDRESS_SPEC_FORMAT = "(<alias>|<address>)"
@@ -100,6 +104,10 @@ BUNDLE_COMMANDS = (
     'upload',
     'make',
     'run',
+    'run-local',
+    'run-image',
+    'commit-image',
+    'push-image',
     'edit',
     'detach',
     'rm',
@@ -484,12 +492,8 @@ class BundleCLI(object):
         bundle_uuid = BundleCLI.resolve_bundle_uuid(client, worksheet_uuid, bundle_spec)
         return (bundle_uuid, subpath)
 
-    def parse_key_targets(self, client, worksheet_uuid, items):
-        """
-        Helper: items is a list of strings which are [<key>]:<target>
-        """
+    def parse_target_specs(self, items):
         targets = []
-        # Turn targets into a dict mapping key -> (uuid, subpath)) tuples.
         for item in items:
             if ':' in item:
                 (key, target) = item.split(':', 1)
@@ -498,6 +502,18 @@ class BundleCLI(object):
             else:
                 # Provide syntactic sugar for a make bundle with a single anonymous target.
                 (key, target) = ('', item)
+
+            targets.append((key, target))
+        return targets
+
+    def parse_key_targets(self, client, worksheet_uuid, items):
+        """
+        Helper: items is a list of strings which are [<key>]:<target>
+        """
+        targets = []
+        # Turn targets into a dict mapping key -> (uuid, subpath)) tuples.
+
+        for key, target in self.parse_target_specs(items):
             if key in targets:
                 if key:
                     raise UsageError('Duplicate key: %s' % (key,))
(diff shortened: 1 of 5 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+            Commands.Argument('-w', '--worksheet-spec', help='Operate on this worksheet (%s).' % WORKSHEET_SPEC_FORMAT, completer=WorksheetsCompleter),
+        ) + Commands.metadata_arguments([RunBundle]) + EDIT_ARGUMENTS + WAIT_ARGUMENTS,
+    )
+    def do_run_local_command(self, args):
+        client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
+        args.target_spec, args.command = cli_util.desugar_command(args.target_spec, args.command)
+        targets = self.parse_key_targets(client, worksheet_uuid, args.target_spec)
+        metadata = self.get_missing_metadata(RunBundle, args)
+
+        docker_image = metadata.get('request_docker_image', 'ubuntu:14.04')
```
percyliang: I think If people are doing run-local, the `--request-docker-image` option should be required (they should be explicit about what they're doing.)  99% of the invocations will not be with `ubuntu:14.04`.

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+        if not docker_image:
+            docker_image = 'ubuntu:14.04'
+
+        uuid = generate_uuid()
+        bundle_path = '/opt/{}'.format(uuid)
```
percyliang: Should not be `/opt`, but should be a flag that's passed in by the user.  Maybe default to the current directory?

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+        container_id = dc.start_container(bundle_path, uuid, command, docker_image, request_network, dependencies)
+        print >>self.stdout, '===='
+        print >>self.stdout, 'ContainerID: ', container_id
+        print >>self.stdout, 'Local Bundle ID: ', uuid
+        print >>self.stdout, 'You can find local bundle contents in: ', bundle_path
+        print >>self.stdout, '===='
+
+    @Commands.command(
+        'run-image',
+        help='Create a bundle by running a program bundle on an input bundle.',
```
percyliang: Needs to be updated to be more specific.

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
         self.wait(client, args, new_bundle['uuid'])
 
     @Commands.command(
+        'run-local',
+        help='Run a run bundle locally.',
```
percyliang: Explain much more.

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+                    ' '.join(['bash', '-c', '; '.join(docker_commands)])
+            )
+        os.system(cli_command)
+
+    @Commands.command(
+        'commit-image',
+        help='Create a bundle by running a program bundle on an input bundle.',
+        arguments=(
+            Commands.Argument('--container', help='Container to commit.', required=True),
+            Commands.Argument('--repository', help='Repository to commit to. E.g: codalabtest-on.azurecr.io/ubuntu', required=True),
```
percyliang: Use dockerhub as an example

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+                    ' '.join([ '-v {}'.format(v) for v in volume_bindings ]),
+                    ' '.join([ '--device {}'.format(d) for d in devices ]),
+                    docker_image,
+                    ' '.join(['bash', '-c', '; '.join(docker_commands)])
+            )
+        os.system(cli_command)
+
+    @Commands.command(
+        'commit-image',
+        help='Create a bundle by running a program bundle on an input bundle.',
```
percyliang: More explanation.

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+            Commands.Argument('--repository', help='Repository to commit to. E.g: codalabtest-on.azurecr.io/ubuntu', required=True),
+        )
+    )
+    def do_commit_image_command(self, args):
+        cli_command = 'docker commit {} {}'.format(args.container, args.repository)
+        os.system(cli_command)
+
+    @Commands.command(
+        'push-image',
+        help='Create a bundle by running a program bundle on an input bundle.',
```
percyliang: Update

codalab/lib/bundle_cli.py
```diff
@@ -1192,6 +1208,119 @@ def do_run_command(self, args):
...
+        request_network = None
+        dependencies = [
+            (u'{}'.format(key), u'/{}_dependencies/{}'.format(uuid, key)) for key, target in self.parse_target_specs(args.target_spec)
+        ]
+
+        # Set up a directory to store the bundle.
+        remove_path(bundle_path)
+        os.mkdir(bundle_path)
+
+        for path1, path2 in dependencies:
```
percyliang: use `name` and `child_path` (I think) to maintain consistency with code elsewhere.

percyliang (changes requested):
This functionality is great!

- In general, more documentation would be nice.
- Can you update test-cli.py to test this functionality (which is also nice since it tells us what commands to use).
- When we add new CLI commands, we should really think through whether this is the best possible design and usage.  We should have several people test it and give comments before committing to it.
- In particular, I'm worried that we're adding too many functions at the top-level.  In some sense, `run-local` and `run-image` behave a lot like `run`, so those could be flags inside `run`.  Of course that has the danger of making `run` too bloated.  But this is something we should think about more and discuss.
- Usability: Before and after each of the commands runs, there should be more stuff printed to stdout saying what's going on and what to do next.
- Also, please use more meaningful commit messages.

fabeschan:
Hi @percyliang, thanks for looking through the changes. These are all good points, and the requested changes all make sense. However like you suggested let's hold a deeper discussion, think things through and settle into a more refined design before committing work to this.

Perhaps I can put together a google doc and we can take down some notes during upcoming meetings?

percyliang:
Sure.  I think we should aim to wrap this up this week (don't think we're
that far off).  And while we should think about it, I think there's no
substitute for rolling it out and getting people to try it (as long as it's
understood that this is a beta feature, we can still change the interface
after it goes into develop).

fabeschan:
https://docs.google.com/document/d/1qELOtPOU6vbA2FtgOMmuBmRrVDGnNRszIgMW8s3XZl4/edit#heading=h.uuvhzjo8mk0y

codalab/lib/bundle_cli.py
fabeschan: Would it be better for bundle paths to be set in config? Having local bundles in one directory could make periodic clean-ups easier. If so, what path would be a good default?

codalab/lib/bundle_cli.py
fabeschan: OK. I will require request-docker-image when running in local mode

fabeschan:
> In some sense, run-local and run-image behave a lot like run, so those could be flags inside run.

It makes sense to absorb run-local into run with a flag, but I'm feeling a bit queasy doing the same for run-image, which is meant for editing images (with some bells added to facilitate testing) but not for producing run bundles (with their contents).

I would prefer to keep them separate commands, but that's just my opinion. I'm not sure which way is best

fabeschan:
> Usability: Before and after each of the commands runs, there should be more stuff printed to stdout saying what's going on and what to do next.

Added more instructions to stdout where appropriate. Please review and let me know if there are more things I can mention @percyliang

fabeschan:
> update test-cli.py to test this functionality

Not exactly sure how to do this since some of the commands drop the user to a shell inside a container. Are there ways to run commands in the inner shell with python?

codalab/lib/bundle_cli.py
fabeschan: @percyliang

PR #663, opened 2017-02-03
Add stress-test.py

kashizui:
Original PR #650


fabeschan:
Requesting preliminary review @percyliang

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+            return False
+
+class UploadBalloons(object):
+
+    def upload_balloon(self):
+        ''' tell codalab to download a balloon, aka a zip file that becomes big when unzipped '''
+
+        self.cli.do_command([
+            'upload',
+            'http://cs246h.stanford.edu/map_only_data.zip' # 250MB per
```
percyliang: Is this a stable link?

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+            return False
+
+class CreateWorksheets(object):
+    def __init__(self, cli, n, multithread=True):
+        self.cli = cli
+        self.n = n
+
+    def __enter__(self):
+        def create_worksheet(i):
+            ws = 'gregwasher91-test-worksheet-{}'.format(i)
```
percyliang: Avoid hardcoding...create an uuid when the script is run to use as a prefix to every name that's used.

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+        ], cli=False)
+
+    def __init__(self, cli):
+        self.cli = cli
+
+    def __enter__(self):
+        try:
+            c = 0
+            while True:
+                self.upload_balloon()
```
percyliang, line 104: Can we print out the status as we go rather than only at the end (same applies to other places)?

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+        print 'Listing ~{} bundles took {}s'.format(len(output['refs']), t.interval)
+
+    with UploadMemoryHog(cli):
+        bundle = cli.do_command([
+            'run',
+            ':memory_hog.py',
+            '---',
+            'timeout 300s python memory_hog.py 500 30'
+        ], cli=False)
+
```
percyliang: Add as TODOs the list of things that we still need to test.

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+        if eType:
+            return False
+
+if __name__ == '__main__':
+    logging.basicConfig(level='INFO')
+
+    cli = BundleCLI(CodaLabManager(), stdout=StringIO(), stderr=StringIO())
+    # Make sure we can connect (might prompt for username/password)
+    try:
+        cli.do_command(['work', 'https://worksheets-test.codalab.org/::gregwasher91-worksheetname'], cli=False)
```
percyliang: Don't use this...require that whoever is running the script already has it pointed at the right place.  I would print out the current worksheet/instance and ask people to confirm (since the stress test can be pretty disruptive).

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+    except Exception, e:
+        logger.error(e)
+        sys.exit(1)
+
+    with CreateWorksheets(cli, n):
```
percyliang: Have command-line arguments so we can specify which tests we run as well as `n`...e.g., `python stress-test.py --tests CreateWorksheets -n 300`

stress-test.py
```diff
@@ -0,0 +1,145 @@
...
+        def create_worksheet(i):
+            ws = 'gregwasher91-test-worksheet-{}'.format(i)
+            try:
+                cli.do_command(['new', ws], cli=False)
+            except:
+                return None
+            return ws
+
+        if self.multithread:
+            array = [ i for i in range(n) ]
```
percyliang: I'd not use extra spaces around `[` and `]` to be consistent with the rest of the codebase...

percyliang (changes requested):
Good start.

Can you also report on the results of the stress test?

PR #665, opened 2017-02-03
Alphabetize Permissions for legacy and REST APIs

kashizui:
Fixes codalab/codalab-worksheets#209 along with codalab/codalab-worksheets#246.

Adds sorting functionality at the server-layer for group permissions for both bundles and worksheets. Implemented in both the REST and legacy APIs.

Original PR #653


percyliang:
LGTM

PR #666, opened 2017-02-03
Support Local Development

kashizui:
Original PR #660

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -83,6 +83,10 @@ from codalab.lib.bundle_store import (
 from codalab.lib.print_util import FileTransferProgress
 from worker.file_util import un_tar_directory
 
+from codalab.lib.spec_util import generate_uuid
+from worker.docker_client import DockerClient
+from worker.file_util import remove_path
+
 # Formatting Constants
 GLOBAL_SPEC_FORMAT = "[<alias>::|<address>::](<uuid>|<name>)"
 ADDRESS_SPEC_FORMAT = "(<alias>|<address>)"
@@ -100,6 +104,10 @@ BUNDLE_COMMANDS = (
     'upload',
     'make',
     'run',
+    'run-local',
+    'run-image',
+    'commit-image',
+    'push-image',
     'edit',
     'detach',
     'rm',
@@ -484,12 +492,8 @@ class BundleCLI(object):
         bundle_uuid = BundleCLI.resolve_bundle_uuid(client, worksheet_uuid, bundle_spec)
         return (bundle_uuid, subpath)
 
-    def parse_key_targets(self, client, worksheet_uuid, items):
-        """
-        Helper: items is a list of strings which are [<key>]:<target>
-        """
+    def parse_target_specs(self, items):
         targets = []
-        # Turn targets into a dict mapping key -> (uuid, subpath)) tuples.
         for item in items:
             if ':' in item:
                 (key, target) = item.split(':', 1)
@@ -498,6 +502,18 @@ class BundleCLI(object):
             else:
                 # Provide syntactic sugar for a make bundle with a single anonymous target.
                 (key, target) = ('', item)
+
+            targets.append((key, target))
+        return targets
+
+    def parse_key_targets(self, client, worksheet_uuid, items):
+        """
+        Helper: items is a list of strings which are [<key>]:<target>
+        """
+        targets = []
+        # Turn targets into a dict mapping key -> (uuid, subpath)) tuples.
+
+        for key, target in self.parse_target_specs(items):
             if key in targets:
                 if key:
                     raise UsageError('Duplicate key: %s' % (key,))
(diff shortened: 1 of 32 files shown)


fabeschan:
Currently local bundles are stored in `/opt/`. Would it be better for bundle paths to be set in config? Having local bundles in one directory could make periodic clean-ups easier. If so, what path would be a good default?

kashizui:
Since codalab installations should already own the directory pointed at by $CODALAB_HOME [default: ~/.codalab] we can probably use that?

kashizui:
You can get this directory using `self.manager.codalab_home`, so maybe `os.path.join(self.manager.codalab_home, 'local_bundles')` or something like that

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
-            self.derive_bundle(RunBundle.BUNDLE_TYPE, args.command, targets, metadata),
-            params={'worksheet': worksheet_uuid},
+
+        if args.local:
+            docker_image = metadata.get('request_docker_image', None)
+            if docker_image is None:
+                raise UsageError('--request-docker-image [docker-image] must be specified when running in local mode')
+
+            uuid = generate_uuid()
+            bundle_path = '/opt/{}'.format(uuid)
```
percyliang: Use `$CODALAB_HOME/temp` instead of /opt

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+            remove_path(bundle_path)
+            os.mkdir(bundle_path)
+
+            for dependency_path, docker_dependency_path in dependencies:
+                child_path = os.path.join(bundle_path, dependency_path)
+                os.symlink(docker_dependency_path, child_path)
+
+            dc = DockerClient()
+            container_id = dc.start_container(bundle_path, uuid, command, docker_image, request_network, dependencies)
+            print >>self.stdout, '===='
```
percyliang, line 1243: Can we print this information before we run the job?

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+                self.derive_bundle(RunBundle.BUNDLE_TYPE, args.command, targets, metadata),
+                params={'worksheet': worksheet_uuid},
+            )
+
+            print >>self.stdout, new_bundle['uuid']
+            self.wait(client, args, new_bundle['uuid'])
+
+    @Commands.command(
+        'edit-image',
+        help='Drop to an interactive container within an image as a bash shell to allow edits.',
```
percyliang: Change to "Start an interactive shell with an image to allow edits to that image."

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+        arguments=(
+            Commands.Argument('target_spec', help=ALIASED_TARGET_SPEC_FORMAT, nargs='*', completer=TargetsCompleter),
+            Commands.Argument('--request-docker-image', help='The docker image to edit', required=True),
         )
+    )
+    def do_edit_image_command(self, args):
+        docker_image = args.request_docker_image
+
+        uuid = generate_uuid()
+        bundle_path = '/opt/{}'.format(uuid)
```
percyliang: Refactor so these paths only appear once (both for `run-local` and this function).

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+                    ' '.join([ '--device {}'.format(d) for d in devices ]),
+                    docker_image,
+                    ' '.join(['bash', '-c', '; '.join(docker_commands)])
+            )
+
+        print >>self.stdout, '===='
+        print >>self.stdout, 'About to enter container' # todo: provide container id
+        print >>self.stdout, 'Once you are happy with the changes, please exit the container (ctrl-D)'
+        print >>self.stdout, 'and commit your changes to a new image by running:'
+        print >>self.stdout, '"cl commit-image --container [container-id] --image-tag [image-tag]"'
```
percyliang: Put the command not in quotes but indented (like in markdown)

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+        print >>self.stdout, 'If you are happy with the changes, please commit your changes to a new image'
+        print >>self.stdout, 'by running "cl commit-image --container [container-id] --image-tag [image-tag]"'
+        print >>self.stdout, '===='
+
+    @Commands.command(
+        'commit-image',
+        help='Create a bundle by running a program bundle on an input bundle.',
+        arguments=(
+            Commands.Argument('--container', help='Container to commit.', required=True),
+            Commands.Argument('image_tag', help='Image-tag to commit to. E.g: codalabtest-on.azurecr.io/ubuntu'),
```
percyliang: Replace `Image-tag` with `Image tag` (same below)

codalab/lib/bundle_cli.py
```diff
@@ -1170,26 +1185,134 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+        remove_path(bundle_path)
+        os.mkdir(bundle_path)
+
+        for dependency_path, docker_dependency_path in dependencies:
+            child_path = os.path.join(bundle_path, dependency_path)
+            os.symlink(docker_dependency_path, child_path)
+
+        dc = DockerClient()
+        extra_args = ['-it']
+        volume_bindings = dc.get_volume_bindings(bundle_path, uuid, command, docker_image, request_network, dependencies)
```
percyliang: All this logic I think should go in DockerClient.  Otherwise, we're exposing too much of the docker internals.

percyliang (changes requested):
Cool, I think we're getting close.  @kashizui should probably take a closer look at the handling of the docker stuff after you address these comments.

codalab/lib/bundle_cli.py
fabeschan: This information is printed out pretty much at the same time the container is started. So it is effectively printed before we run the job.

Furthermore, we cannot get the container id before we start the container.

codalab/lib/bundle_cli.py
fabeschan: Done

codalab/lib/bundle_cli.py
fabeschan: Done

codalab/lib/bundle_cli.py
```diff
@@ -1173,26 +1188,129 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+
+        dc = DockerClient()
+        cli_command = dc.create_cli_command(bundle_path, uuid, command, docker_image, request_network, dependencies, ['-it'])
+
+        print >>self.stdout, '===='
+        print >>self.stdout, 'About to enter container' # todo: provide container id
+        print >>self.stdout, 'Once you are happy with the changes, please exit the container (ctrl-D)'
+        print >>self.stdout, 'and commit your changes to a new image by running:'
+        print >>self.stdout, ''
+        print >>self.stdout, '\tcl commit-image --container [container-id] [image-tag]'
```
kashizui: Print some information on how to get the container id, like 

`Your command prompt should look like root@<CONTAINERID>:~#`

though this doesn't apply generally. Maybe you could use a Docker API call to get the ID of the container that you just started?

codalab/lib/bundle_cli.py
```diff
@@ -1173,26 +1188,129 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
+        print >>self.stdout, ''
+        print >>self.stdout, '===='
+
+    @Commands.command(
+        'commit-image',
+        help='Create a bundle by running a program bundle on an input bundle.',
+        arguments=(
+            Commands.Argument('--container', help='Container to commit.', required=True),
```
kashizui: Why use a named argument if it's required anyway? (Maybe could save some typing if it were just `commit-image <container> <imagetag>`?)

kashizui:
Taking this branch for a spin. Some notes:

- It would be great if `cl help` has a new section with "Commands for Docker images" to separate the new commands from the rest.
- General usage of Docker looks fine -- I understand the rationale for using the Docker CLI rather than the API for all cases here, and glad that start_container() was usable when possible.
- I'm getting this error when trying local run, have you seen this before @fabeschan ?

```
➜  cl run --local :alembic.ini 'cat alembic.ini'                                           17-02-22 12:46:30

nvidia-docker-plugin not available, no GPU support on this worker.

  File "/Users/sckoo/Developer/codalab-cli/codalab/bin/../../codalab/bin/cl.py", line 116, in <module>
    cli.do_command(sys.argv[1:])
  File "/Users/sckoo/Developer/codalab-cli/codalab/lib/bundle_cli.py", line 722, in do_command
  File "/Users/sckoo/Developer/codalab-cli/codalab/lib/bundle_cli.py", line 704, in <lambda>
  File "/Users/sckoo/Developer/codalab-cli/codalab/lib/bundle_cli.py", line 1226, in do_run_command
    container_id = dc.start_container(bundle_path, uuid, command, docker_image, request_network, dependencies)
  File "/Users/sckoo/Developer/codalab-cli/worker/docker_client.py", line 22, in wrapper
    return f(*args, **kwargs)
  File "/Users/sckoo/Developer/codalab-cli/worker/docker_client.py", line 336, in start_container
    raise DockerException(start_response.read())
worker.docker_client.DockerException: Unable to start Docker container: {"message":"oci runtime error: container_linux.go:247: starting container process caused \"exec: \\\"bash\\\": executable file not found in $PATH\"\n"}
```

worker/docker_client.py
```diff
@@ -228,8 +228,36 @@ def download_image(self, docker_image, loop_callback):
                     pass
                 loop_callback(status)
 
-    @wrap_exception('Unable to start Docker container')
-    def start_container(self, bundle_path, uuid, command, docker_image,
+    def create_cli_command(self, bundle_path, uuid, command, docker_image,
```
kashizui: nit: Something a bit more specific like `create_cli_run_command`

worker/docker_client.py
```diff
@@ -228,8 +228,36 @@ def download_image(self, docker_image, loop_callback):
...
+
+        return 'docker run {} {} {} {} {} {} {}'.format(
+                ' '.join(extra_args),
+                ' '.join([ '-v {}'.format(v) for v in volume_bindings ]),
+                '-w="{}"'.format(docker_bundle_path),
+                '', #'--user={}:{}'.format(uid, gid),
+                '-e {}="{}"'.format('HOME', docker_bundle_path),
+                docker_image,
+                command
+        )
```
kashizui: We might have to be careful about quoting here (what happens if a dependency has a space in it?)

worker/docker_client.py
```diff
@@ -228,8 +228,36 @@ def download_image(self, docker_image, loop_callback):
...
+
+    def get_docker_commands(self, bundle_path, uuid, command, docker_image,
```
kashizui: Use leading underscore `_get_docker_commands` to make clear that it's a private method.

worker/docker_client.py
```diff
@@ -239,13 +267,28 @@ def start_container(self, bundle_path, uuid, command, docker_image,
             '[ -e /user/.bashrc ] && . /user/.bashrc',
             '(%s) >stdout 2>stderr' % command,
         ]
+        return docker_commands
+
+    def get_volume_bindings(self, bundle_path, uuid, command, docker_image,
```
kashizui: Same as above, use leading underscore.

codalab/lib/bundle_cli.py
```diff
@@ -1173,26 +1188,129 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
...
-        targets = self.parse_key_targets(client, worksheet_uuid, args.target_spec)
         metadata = self.get_missing_metadata(RunBundle, args)
-        new_bundle = client.create(
-            'bundles',
-            self.derive_bundle(RunBundle.BUNDLE_TYPE, args.command, targets, metadata),
-            params={'worksheet': worksheet_uuid},
+
+        if args.local:
+            docker_image = metadata.get('request_docker_image', None)
+            if docker_image is None:
```
kashizui: This might need to be a general falsey check, not a `None check: the above call `metadata.get()` will return an empty string if a Docker image is not provided because of how `get_missing_metadata()` works.

kashizui: This was why I was getting the error that I just reported.

codalab/lib/bundle_cli.py
fabeschan: I've considered using the Docker API to get the ID somehow but it seems very hacky and not robust. Perhaps I'll just print a helper message like you suggest up there.

codalab/lib/bundle_cli.py
fabeschan: OKay, makes sense

worker/docker_client.py
fabeschan: done

worker/docker_client.py
fabeschan: removed quotes

codalab/lib/bundle_cli.py
fabeschan: good suggestion. done.

codalab/lib/bundle_cli.py
fabeschan: Good catch btw, fixed.

worker/docker_client.py
fabeschan: done

fabeschan:
@kashizui updated the PR. Please review again!

worker/docker_client.py
kashizui: Did you forget to push the latest changes?

kashizui: Woops I see them now, GitHub didn't refresh I guess.

codalab/lib/bundle_cli.py
kashizui: Nice solution, pretty smooth.

kashizui:
Apparently I can't officially "approve" this PR on GitHub since I created it, but here is my LGTM 👍🏼

kashizui:
@fabeschan Just a heads up, I just pushed an additional commit on top of this branch to make the `cl help` display better.

PR #668, opened 2017-02-03
Generate REST API reference

kashizui:
Minimum viable product for REST API docs.

The script `scripts/gen-rest-docs.py` generates a Markdown file at `docs/rest.md`. We can link to this document from the Wiki.

Current look:
https://github.com/codalab/codalab-cli/blob/kashizui/gendoc/docs/rest.md

The docstrings of the REST functions themselves should be fleshed out slowly over time. That will also better inform whether or not we want some special formatting for the route documentations (e.g. for the query parameters, etc.)

Addresses #508


percyliang:
LGTM

Let's leave an issue open for fleshing out the REST API.

We should link to rest.md from the Wiki too.

PR #675, opened 2017-02-06
Help button

maxwang7:
Issue: https://github.com/codalab/codalab-worksheets/issues/251
Corresponding PR in codalab-worksheets: https://github.com/codalab/codalab-worksheets/pull/258

Creates a REST API endpoint that sends a support email.

These are features that I considered "extensions" -- let me know if any of them should be implemented for this PR:

- [ ] Creating a new SQL table that records all the messages that have been sent to us.
- [ ] Sending a "message received" email to the message sender's email.

### DIFF
--- a/codalab/rest/help.py
+++ b/codalab/rest/help.py
@@ -27,4 +27,3 @@ def fetch_help():
         recipient=support_email,
         sender=user_email
     )
-    return '{}'
--- a/codalab/rest/help.py
+++ b/codalab/rest/help.py
@@ -16,15 +16,16 @@ from codalab.server.authenticated_plugin import (
 
 @post('/help/', apply=AuthenticatedPlugin(), skip=UserVerifiedPlugin)
 def fetch_help():
-    # import pdb; pdb.set_trace()
     user = AuthenticatedUserSchema().dump(request.user).data
-    email = user['data']['attributes']['email']
+    support_email = local.config['support_email']
     username = user['data']['attributes']['user_name']
     message = request.json['message']
+    user_email = user['data']['attributes']['email']
+    print "********" + user_email
 
     local.emailer.send_email(
-        subject="",
-        body=template('help_message_to_codalab_body', username=username, email=email, message=message),
-        recipient=email,
+        subject="Message from %s" % user_email,
+        body=template('help_message_to_codalab_body', username=username, email=support_email, message=message, sender=user_email),
+        recipient=support_email,
     )
     return '{}'
--- a/codalab/rest/help.py
+++ b/codalab/rest/help.py
@@ -16,15 +16,15 @@ from codalab.server.authenticated_plugin import (
 
 @post('/help/', apply=AuthenticatedPlugin(), skip=UserVerifiedPlugin)
 def fetch_help():
-    user = AuthenticatedUserSchema().dump(request.user).data
     support_email = local.config['support_email']
-    username = user['data']['attributes']['user_name']
+    username = request.user.user_name
     message = request.json['message']
-    user_email = user['data']['attributes']['email']
+    user_email = request.user.email
 
     local.emailer.send_email(
         subject="Message from %s" % user_email,
-        body=template('help_message_to_codalab_body', username=username, email=support_email, message=message, sender=user_email),
+        body=template('help_message_to_codalab_body', username=username, email=support_email, message=message),
         recipient=support_email,
+        sender=user_email
     )
     return '{}'
--- a/codalab/rest/help.py
+++ b/codalab/rest/help.py
@@ -1,3 +1,4 @@
+import sys
 import httplib
 
 from bottle import abort, get, request, local, post, template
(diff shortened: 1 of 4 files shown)


codalab/rest/help.py
```diff
@@ -0,0 +1,30 @@
...
+def fetch_help():
+    user = AuthenticatedUserSchema().dump(request.user).data
+    support_email = local.config['support_email']
+    username = user['data']['attributes']['user_name']
+    message = request.json['message']
+    user_email = user['data']['attributes']['email']
+
+    local.emailer.send_email(
+        subject="Message from %s" % user_email,
+        body=template('help_message_to_codalab_body', username=username, email=support_email, message=message, sender=user_email),
```
maxwang7: For some reason setting the sender parameter isn't working for me -- it doesn't change who the sender is when I print it out. From my reading of [the email code](https://github.com/codalab/codalab-cli/blob/7e075ff3c4e3f4dc82ffb770382fc68feb90b0c2/codalab/lib/emailer.py#L84), though, it seems like it should work, even on the console.

codalab/rest/help.py
```diff
@@ -0,0 +1,30 @@
...
+    UserSchema,
+)
+from codalab.server.authenticated_plugin import (
+    AuthenticatedPlugin,
+    UserVerifiedPlugin,
+)
+
+@post('/help/', apply=AuthenticatedPlugin(), skip=UserVerifiedPlugin)
+def fetch_help():
+    support_email = local.config['support_email']
```
kashizui: What should happen if 'support_email' isn't defined in the config? Or should we require that 'support_email' always be defined?

codalab/rest/help.py
```diff
@@ -0,0 +1,30 @@
...
+    message = request.json['message']
+    user_email = request.user.email
+
+    local.emailer.send_email(
+        subject="Message from %s" % user_email,
+        body=template('help_message_to_codalab_body', username=username, email=support_email, message=message),
+        recipient=support_email,
+        sender=user_email
+    )
+    return '{}'
```
kashizui: Probably just don't return anything, unless you intend on adding data in the future. The frontend can just check the status code for success/failure.

If you want to return JSON data, you can just

    return {}

without the quotes.

kashizui (commented):
Looks pretty good! Is this still a work in progress?

codalab/rest/help.py
maxwang7: How have we dealt with similar kinds of configuration issues? I think we should require that it's always defined.

codalab/rest/help.py
maxwang7: Ah, that sounds good, I just won't return then

maxwang7:
Also, no, it's not meant to be a work in progress... unless you think there are more things that I should work on :)

codalab/rest/help.py
maxwang7: Just fixed this in my latest commit

codalab/rest/help.py
kashizui: Usually we try to provide a sane default when something is not defined, or just let it through with a warning message, unless it makes no sense to do so.

Example with the `admin_email` config:
https://github.com/codalab/codalab-cli/blob/master/codalab/server/rest_server.py#L200-L205

The issue is that requiring the config may cause a deployment to crash if we just forgot to update the config on the server. In this case it should be pretty isolated and not cause much problems though. Maybe we can just let this throw an exception to let us know we forgot to configure a support email, or throw your own exception to give a more helpful error message.

codalab/rest/help.py
```diff
@@ -0,0 +1,29 @@
+import httplib
+
+from bottle import abort, get, request, local, post, template
+
+from codalab.lib.spec_util import NAME_REGEX
+from codalab.lib.server_util import bottle_patch as patch
+from codalab.rest.schemas import (
+    AuthenticatedUserSchema,
+    USER_READ_ONLY_FIELDS,
+    UserSchema,
```
kashizui: clean up unused imports

kashizui (changes requested):

codalab/rest/help.py
maxwang7: Done, thanks!

kashizui:
- i don't think there's a need for a sql table since we won't want to build an interface for viewing the table haha
- as long as the UI provides feedback on whether or not the message went through, it should be fine

maxwang7:
Re SQL Table: Yeah, I was thinking that we may want to show users the messages that they've sent previously, and it may be a good next step to making this interface a more full-featured chat application. But I think that's a conversation for another day.

Re Message receive email: Ok, that seems reasonable.

kashizui:
If we want to make a more full-featured support ticket thing we might as well integrate with Zendesk.
Chat is probably a different story

codalab/rest/help.py
maxwang7: Ok, well why don't I just follow the pattern we've already established, and I'll just print off a message saying that the support_email needs to be configured, and then print of the email I would've sent to the console.

maxwang7: Ok, changes made!

kashizui (changes requested):
lgtm

do you know why the rest docs are showing up in the diff? They should be master already afaik..

kashizui (approved):

maxwang7:
I'm not sure why or how they ended up here... I think I might've pull'd and they ended up in my this branch somehow. And sounds good, I'll merge.

maxwang7:
Test this using SendGrid

Issue #568, opened 2016-08-13
Clean up Docker images on worker automatically


kashizui:
Periodically clean up the stalest images on the worker, impose a disk usage limit?

percyliang:
Workers should clear their cached bundles periodically and docker images too.

percyliang:
Docker containers currently are a big problem.

kashizui:
Another postportem from @percyliang 

> The number of bundles on the worker machines exceeded some number, so we were getting a lot of 'Request Too Large' errors.  Seems like three things went wrong:
> - We didn't get notified about this problem (we need to change this somehow).
> - We should be able to handle larger requests (thought we fixed this).
> - We shouldn't let the number of bundles fill up in the first place (this is on the list of TODOs).
> 
> I manually cleared out all the scratch directories on the worker machines, and things are back running now.
> 
>  -Percy

kashizui:
Easy solution: Periodically clear all worker caches.

Better solution: Workers should keep track of when dependencies and docker images were last used and clean up stales ones when disk utilization reaches a certain threshold.

kashizui:
Actually, dependencies are already cleaned up in a separate thread when they exceed a certain amount of disk usage, and containers are cleaned up automatically after runs. The main problem is Docker images. Going to narrow the scope of this issue.

kashizui:
Currently kind of stuck on how to estimate how much disk space the Docker images are currently using (so that my cleanup thread can know when to start cleaning up old images) @fabeschan any idea?

kashizui:
One way appears to be something like `du -s /var/lib/docker/aufs/diff` but that does not appear to generalize to Mac. Maybe we just forgo Mac support on this.

percyliang:
What about just deleting images that haven't been used in a while?

kashizui:
@percyliang I'm fine with that, but we may still run into disk full problems in the case where users suddenly use a bunch of different images in a short period of time. Thoughts?

percyliang:
We can just look at the usage of the disk and remove old images until the
usage is below some threshold?

kashizui:
That sounds good. So we specify a disk to monitor, and as long as the disk usage is above a certain threshold (should this be by percentage or say, leave a buffer of 5gb?), we keep removing the stalest images.

percyliang:
Something like that...  Does our bundle cleanup logic work the same way?

Stalest implies you can get the 'last used' date of an image, not 'last
created' - is that possible?

[...]

PR #676, opened 2017-02-06
 Clean up Docker images on worker automatically

kashizui:
Removes the stalest docker images when the disk is almost full (less than 1 gb free).

Addresses #568

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -441,4 +441,10 @@ class BundleManager(object):
         resources['request_network'] = (bundle.metadata.request_network or
                                         self._default_request_network)
 
+        resources['request_cpus'] = (bundle.metadata.request_cpus or
+                                        self._default_request_cpus)
+
+        resources['request_gpus'] = (bundle.metadata.request_gpus or
+                                     self._default_request_gpus)
+
         return message
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -441,10 +441,4 @@ class BundleManager(object):
         resources['request_network'] = (bundle.metadata.request_network or
                                         self._default_request_network)
 
-        resources['request_cpus'] = (bundle.metadata.request_cpus or
-                                        self._default_request_cpus)
-
-        resources['request_gpus'] = (bundle.metadata.request_gpus or
-                                     self._default_request_gpus)
-
         return message
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -221,7 +221,7 @@ nvidia-docker-plugin not available, no GPU support on this worker.
         # once we are sure every worker has updated docker
         try:
             return info['RepoDigests'][0]
-        except KeyError, IndexError:
+        except (KeyError, IndexError):
             return ''
 
     @wrap_exception('Unable to remove Docker image')
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -60,6 +60,7 @@ class DockerClient(object):
     # Where to look for nvidia-docker-plugin
     # https://github.com/NVIDIA/nvidia-docker/wiki/nvidia-docker-plugin
     NV_HOST = 'localhost:3476'
+    MIN_API_VERSION = '1.17'
 
     def __init__(self):
         self._docker_host = os.environ.get('DOCKER_HOST') or None
@@ -166,7 +167,8 @@ nvidia-docker-plugin not available, no GPU support on this worker.
                 version_info = json.loads(version_response.read())
             except:
                 raise DockerException('Invalid version information')
-            if version_info['ApiVersion'] < '1.17':
+            if map(int, version_info['ApiVersion'].split('.')) < \
+                    map(int, self.MIN_API_VERSION.split('.')):
                 raise DockerException('Please upgrade your version of Docker')
 
     @wrap_exception('Unable to get disk usage info')
@@ -398,12 +400,12 @@ nvidia-docker-plugin not available, no GPU support on this worker.
 
     @wrap_exception('Unable to start Docker container')
     def start_container(self, bundle_path, uuid, command, docker_image,
-                        request_network, request_gpus, dependencies):
-        docker_commands = self._get_docker_commands(bundle_path, uuid, command, docker_image,
-                                                    request_network, dependencies)
+                        request_network, dependencies):
+        docker_commands = self._get_docker_commands(
+            bundle_path, uuid, command, docker_image, request_network, dependencies)
 
-        volume_bindings = self._get_volume_bindings(bundle_path, uuid, command, docker_image,
-                                                    request_network, dependencies)
+        volume_bindings = self._get_volume_bindings(
+            bundle_path, uuid, command, docker_image, request_network, dependencies)
 
         # Get user/group that owns the bundle directory
         # Then we can ensure that any created files are owned by the user/group
@@ -427,7 +429,8 @@ nvidia-docker-plugin not available, no GPU support on this worker.
             # This can cause problems if users expect to run as a specific user
             'User': '%s:%s' % (uid, gid),
         }
-        # TODO: Allocate the specified number of GPUs and isolate
+
+        # TODO: Allocate the requested number of GPUs and isolate
         if self._use_nvidia_docker:
             self._add_nvidia_docker_arguments(create_request)
         if not request_network:
@@ -489,16 +492,6 @@ nvidia-docker-plugin not available, no GPU support on this worker.
             pass
 
         return stats
-
-    @wrap_exception('Unable to get Docker info')
-    def get_root_dir(self):
-        with closing(self._create_connection()) as conn:
-            conn.request('GET', '/info')
-            info_response = conn.getresponse()
-            if info_response.status != 200:
-                raise DockerException(info_response.read())
-            info = json.load(info_response)
-            return info['DockerRootDir']
 
     @wrap_exception('Unable to kill Docker container')
     def kill_container(self, container_id):
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -169,16 +169,50 @@ nvidia-docker-plugin not available, no GPU support on this worker.
             if version_info['ApiVersion'] < '1.17':
                 raise DockerException('Please upgrade your version of Docker')
 
-    @wrap_exception('Unable to fetch Docker image metadata')
-    def inspect_image(self, request_docker_image):
-        logger.debug('Fetching Docker image metadata for %s', request_docker_image)
+    @wrap_exception('Unable to get disk usage info')
+    def get_disk_usage(self):
+        """
+        Return the total amount of disk space used by Docker images, and the
+        amount that can be reclaimed by removing unused Docker images, in bytes.
+
+        Emulates computation of the RECLAIMABLE field in the output for
+        `docker system df`.
+
+        Original implementation:
+        https://github.com/docker/docker/blob/ea61dac9e6d04879445f9c34729055ac1bb15050/cli/command/formatter/disk_usage.go#L197-L214
+        """
         with closing(self._create_connection()) as conn:
-            conn.request('GET', '/images/%s/json' % request_docker_image)
+            conn.request('GET', '/system/df')
+            df_response = conn.getresponse()
+            if df_response.status != 200:
+                raise DockerException(df_response.read())
+            df_info = json.loads(df_response.read())
+        total = df_info['LayersSize']
+        used = 0.
+        for image in df_info['Images']:
+            if image['Containers'] > 0:
+                if image['VirtualSize'] == -1 or image['SharedSize'] == -1:
+                    continue
+                used += image['VirtualSize'] - image['SharedSize']
+        reclaimable = total - used
+        return total, reclaimable
+
+    def _inspect_image(self, image_name):
+        """
+        Get raw image info JSON.
+        :param image_name: id, tag, or repo digest
+        """
+        logger.debug('Fetching Docker image metadata for %s', image_name)
+        with closing(self._create_connection()) as conn:
+            conn.request('GET', '/images/%s/json' % image_name)
             create_image_response = conn.getresponse()
             if create_image_response.status != 200:
                 raise DockerException(create_image_response.read())
-            contents = json.loads(create_image_response.read())
+            return json.loads(create_image_response.read())
 
+    @wrap_exception('Unable to fetch Docker image metadata')
+    def get_image_repo_digest(self, request_docker_image):
+        info = self._inspect_image(request_docker_image)
         # the following try-except clause is intentional:
         # old docker versions may have an empty entry under RepoDigests
         # older docker versions may not even have an entry under RepoDigests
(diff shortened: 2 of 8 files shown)


kashizui:
@percyliang please review!

worker/docker_image_manager.py
```diff
@@ -0,0 +1,135 @@
...
+logger = logging.getLogger(__name__)
+
+
+class DockerImageManager(object):
+    """
+    Manages the Docker images available on the worker, and ensures that they
+    do not fill up the disk.
+
+    Warning: Assumes that Docker images are stored on the same disk as the
+    work directory specified by the `workdir` parameter. If this is not the
```
percyliang: Why not just always look at /var/lib/docker?

worker/docker_image_manager.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _should_stop_cleanup(self):
+        with self._lock:
+            return self._stop_cleanup
+
+    def _get_free_bytes(self):
+        """
+        Return bytes free on disk containing state file.
+        """
+        fd = os.open(self._state_file, os.O_RDONLY)
```
percyliang: Here, could just use /var/lib/docker directly, no?

worker/docker_image_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        while not self._should_stop_cleanup():
+            # Start disk usage reduction loop
+            # Iteratively try to remove the oldest images until the disk usage
+            # is within the limit.
+            try:
+                while True:
+                    free_bytes = self._get_free_bytes()
+                    logging.debug('disk free: %f gb (min %f gb)' %
+                                  (free_bytes / 1e9, self._min_free_bytes / 1e9))
```
percyliang: Use formatting.size_str?

worker/run.py
```diff
@@ -191,6 +192,11 @@ def update_status_and_check_killed(status):
                     self._container_id = do_start()
                 else:
                     raise
+
+            self._image_manager.touch_image(self._resources['docker_image'])
```
percyliang: Isn't this the tag?  Would be more stable to store the digest.

percyliang (commented):
If there are already a lot of docker images, and we start running this code, `last_used` will not be filled in the state.  How do we have them default to the current time?

worker/docker_image_manager.py
kashizui: That doesn't work on Mac -- but like I mentioned in the issue thread I have no qualms about forgoing Mac support for this feature.

worker/docker_image_manager.py
kashizui: see discussion above

worker/run.py
kashizui: I originally used the digest then discovered that you can't remove images by their digest..

Maybe I'll see if I can do a reverse lookup by digest to a get a name that I can delete the image by when I need to.

worker/docker_image_manager.py
kashizui: Shall we go with Linux-only?

percyliang: Is it just a different path?

percyliang:
I worry a bit about having the separate last_used state and having it go stale - docker images might be deleted which are still in last_used, and we should clean them up.  If we delete the images-state.json file, things should default to something sane?

Also, probably want to not delete images that were created very recently?  The worker could be running on a machine that has a lot of other stuff going on (user's worker), and we don't want to just randomly delete his docker images.  Maybe autocleanup should be an option which is off by default?

worker/docker_image_manager.py
kashizui: Actually I just realized it's simply impossible on Mac, so there's no reason to support it anyway.
Basically, Docker for Mac doesn't free up disk space when you delete images, until you restart the whole thing. It has something to do with the fact that it's running in a VM.

I'll just point at `/var/lib/docker/aufs/diff` (which holds the images) then!

kashizui: Actually that will require root permissions. Hm.

kashizui: Oops that isn't necessary, I can just point `statvfs()` to `/var/lib/docker`, and it looks like you don't need root permissions to do that (just tried it on one of the Azure vms).

kashizui:
Next todos:

- [x] make images autocleanup OFF by default
- [x] look into tracking images by repo digest / actual ID
- [x] Use formatting.size_str
- [x] Point `statvfs()` to `/var/lib/docker` (can look up this directory with `docker info`, but this will be false for Mac.. how to disable this autoclean for Mac entirely?)

kashizui:
@percyliang addressed all comments, please take another look!

worker/docker_client.py
```diff
@@ -267,7 +272,10 @@ def start_container(self, bundle_path, uuid, command, docker_image,
             # This can cause problems if users expect to run as a specific user
             'User': '%s:%s' % (uid, gid),
         }
-        if self._use_nvidia_docker:
+        # Using nvidia-docker when GPU is not needed can have unintended
+        # side effects, so we only do so when GPUs are requested.
+        # TODO: Allocate the specified number of GPUs and isolate
+        if request_gpus > 0 and self._use_nvidia_docker:
```
percyliang: Is this the behavior we want?  It is extra burden on the users.  We don't have to specify request_cpus > 0 to get CPUs.

What are the unintended side effects?

worker/docker_image_manager.py
```diff
@@ -0,0 +1,140 @@
...
+        self._cleanup_thread.join()
+        self._cleanup_thread = None
+
+    def _should_stop_cleanup(self):
+        with self._lock:
+            return self._stop_cleanup
+
+    def _get_free_bytes(self):
+        """
+        Return bytes free on disk containing state file.
```
percyliang: Update comment

worker/docker_image_manager.py
```diff
@@ -0,0 +1,140 @@
...
+                            self._save_state()
+                            logging.info('removed image %s' % image_id)
+                    else:
+                        # Break out of the loop when disk usage is normal.
+                        break
+            except Exception:
+                # Print the error then go to sleep and try again next time.
+                traceback.print_exc()
+
+            # Sleep for 10 seconds, allowing interruptions every second.
```
percyliang: Update comment from `10`.

worker/docker_image_manager.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        logging.debug('Image cleanup thread started')
+        while not self._should_stop_cleanup():
+            # Start disk usage reduction loop
+            # Iteratively try to remove the oldest images until the disk usage
+            # is within the limit.
+            try:
+                while True:
+                    free_bytes = self._get_free_bytes()
+                    logging.debug('disk free: %s (min %s)' %
```
percyliang: Isn't this going to be a lot of output?  Why not put inside the `if`?

worker/docker_image_manager.py
```diff
@@ -0,0 +1,140 @@
...
+                            logging.info('removed image %s' % image_id)
+                    else:
+                        # Break out of the loop when disk usage is normal.
+                        break
+            except Exception:
+                # Print the error then go to sleep and try again next time.
+                traceback.print_exc()
+
+            # Sleep for 10 seconds, allowing interruptions every second.
+            for _ in xrange(0, self._cleanup_sleep_seconds):
```
percyliang: Do we need this logic, as opposed to just sleeping 1 second in the main loop?  All we have to do is check free space, which should be fast.

worker/docker_image_manager.py
```diff
@@ -0,0 +1,140 @@
...
+    The last-used dates are only tracked for images on which
+    `DockerImageManager.touch_image()` has been called. All other images will
+    be left untouched by the cleanup thread.
+
+    DependencyManager should be instantiated before DockerImageManager, to
+    ensure that the work directory already exists.
+    """
+    STATE_FILENAME = 'images-state.json'
+
+    def __init__(self, docker, work_dir, min_free_bytes=1e9,
```
percyliang: I'd give more than 1GB...some images can be more than 1GB.  I'd say 10GB, and maybe allow this flag to be passed in from the config file.  Perhaps there should be one parameter specified in the config file with a default of 10GB that is passed to both auto bundle cleanup and docker image cleanup.  So we just have to reason about how much free space we want to maintain on disk for the worker.  (Although docker and bundles could in principle live on different disks, but really they should be in the same place.)

percyliang (commented):
Mostly looks good.  Maybe @fabeschan can try it out after comments are addressed.

worker/docker_client.py
kashizui: I don't know any specific examples, but according to [the nvidia-docker documentation](https://github.com/NVIDIA/nvidia-docker/wiki/Image-inspection):

> Mounting user-level driver libraries and device files clobbers the environment of the container, it should be done only when the container is running a GPU application. 

I have noticed that most users on the NLP instance have included the `--request-gpus` flag on their GPU jobs, so it might not be so bad? But I'm not opposed to removing this check until someone complains about their container being clobbered. It might never happen since nvidia-docker only modifies `/usr/local/nvidia` and the device files as far as I know.

worker/docker_image_manager.py
kashizui: I guess not!

worker/docker_image_manager.py
kashizui: 10GB and config file parameter sound good, and uniform cleanup logic would be great.

I think we might need a bit more than just specifying 10GB to each cleanup thread though: when the disk has less than 10 GB left, how do they decide whether to delete an image or a bundle? That's why I was hoping for some way to query for the amount of disk space used by all the Docker images, so that we could just set independent limits on both bundles and images, such that the sum of the limits is less than the size of the disk.

But since that was impossible, I decided to use a disk free parameter, which means that while the dependency manager caps the amount of disk space used by bundles, the image manager independently ensures that images never fill up the remainder of the disk completely.

kashizui: Proposal: Maybe you can specify a `max_docker_images_cache_size`, and we will maintain at least `DISK_SIZE - (max_docker_images_cache_size + max_work_dir_size)` free on the disk. This has the potentially confusing side-effect of sometimes exceeding the `max_docker_images_cache_size` when the dependency cache is not full, but at least disk usage will not exceed `(max_docker_images_cache_size + max_work_dir_size)`. @percyliang thoughts?

kashizui: hm actually not sure this idea is sound

kashizui: The problem being that the actual allowed size of the image cache may effectively be smaller, if there are other files on the same disk besides bundles and images.

fabeschan:
Do you see any potential issues if multiple image managers (from multiple workers) are sharing the same docker service (docker socket)?

worker/run.py
fabeschan: Just a clarification:
you can't remove images by digest on the REST api, but it is possible on the cli

kashizui:
I believe a new file descriptor/port gets allocated for each connection, so should be fine!

kashizui:
I'll rebase and merge if this looks good? This functionality is disabled by default, so we can roll it out slowly manually to the workers and see how well it works in practice.

kashizui:
Still need to make changes to allow this to work on our prod workers. Calling `os.fstatvfs()` on the docker directory fails because worker process does not have permissions to read the docker directory.

kashizui:
Made some final discoveries and fixes:

- I figured out how to query Docker for the actual disk space used by Docker images, and so I switched back to enforcing a *max* disk space usage (instead of min free disk space).
- Now tracking images by repo digest instead of image ID.
- Included a minor fix for #702 
I'll test on wstest asap.

@percyliang @fabeschan please review!

worker/docker_client.py
```diff
@@ -439,6 +490,16 @@ def get_container_stats(self, container_id):
...
+    @wrap_exception('Unable to get Docker info')
+    def get_root_dir(self):
+        with closing(self._create_connection()) as conn:
+            conn.request('GET', '/info')
+            info_response = conn.getresponse()
+            if info_response.status != 200:
+                raise DockerException(info_response.read())
+            info = json.load(info_response)
+            return info['DockerRootDir']
+
```
kashizui: delete this

worker/docker_client.py
```diff
@@ -377,6 +427,7 @@ def start_container(self, bundle_path, uuid, command, docker_image,
             # This can cause problems if users expect to run as a specific user
             'User': '%s:%s' % (uid, gid),
         }
+        # TODO: Allocate the specified number of GPUs and isolate
```
kashizui: maybe remove these changes

codalab/worker/bundle_manager.py
```diff
@@ -441,4 +441,10 @@ def parse_and_min(to_value, request_string, default_value, max_value):
         resources['request_network'] = (bundle.metadata.request_network or
                                         self._default_request_network)
 
+        resources['request_cpus'] = (bundle.metadata.request_cpus or
+                                        self._default_request_cpus)
+
+        resources['request_gpus'] = (bundle.metadata.request_gpus or
+                                     self._default_request_gpus)
+
```
kashizui: are these necessary?

fabeschan: probably not (but I can't say for sure)

worker/docker_client.py
```diff
@@ -171,27 +169,79 @@ def test(self):
             if version_info['ApiVersion'] < '1.17':
```
fabeschan: careful, lexical comparisons have bitten me before (and that occurred precisely when I was comparing docker versions)

fabeschan: i.e. probably breaks when it returns '9' or something like that

worker/docker_client.py
```diff
@@ -171,27 +169,79 @@ def test(self):
...
         # old docker versions may have an empty entry under RepoDigests
         # older docker versions may not even have an entry under RepoDigests
         # this handles both cases
         # the goal is to eventually remove this try-except clause completely,
         # once we are sure every worker has updated docker
         try:
-            return contents['RepoDigests'][0]
-        except KeyError, IndexError:
+            return info['RepoDigests'][0]
+        except (KeyError, IndexError):
```
fabeschan, line 226: Thanks for fixing this for me 👍

worker/main.py
```diff
@@ -32,6 +32,14 @@
...
                         help='Maximum size of the temporary bundle data '
                              '(e.g., 3, 3k, 3m, 3g, 3t).')
+    parser.add_argument('--max-images-size', type=str, default='10g',
+                        help='Maximum amount of disk space to use to store '
+                             'Docker images (e.g., 3, 3k, 3m, 3g, 3t). '
+                             'The --remove-stale-images flag must be specified '
+                             'in order to enforce this limit.')
+    parser.add_argument('--remove-stale-images', action='store_true',
+                        help='Whether to remove Docker images that have not '
+                             'been used recently when the disk is almost full.')
```
maxwang7: Add a check after parsing that enforces that if `--max-images-size` option is present, then `--remove-stale-images` must also be present. Print out helpful error message if that doesn't happen.

worker/worker.py
```diff
@@ -34,10 +33,11 @@ class Worker(object):
         4) Upgrading the worker.
```
maxwang7, line 34: Should we add a comment describing DockerImageManager?

fabeschan (changes requested):
LGTM other than the docker version thingy. nice job

worker/main.py
maxwang7: Change to clearer wording (--remove-stale-images):

`Whether to keep track of Docker images used by CodaLab and to remove those images that have not been used recently when the disk is almost full. Also see --max-images-size.`

codalab/worker/bundle_manager.py
maxwang7: Are you referring to just the CPU and GPU flags or just this set of code in general?

fabeschan (approved):

maxwang7 (approved):

worker/docker_client.py
kashizui: done!

worker/main.py
kashizui: Simplified this even further to a single argument.

```
  --max-image-cache-size SIZE
                        Limit the disk space used to cache Docker images for
                        worker jobs to the specified amount (e.g. 3, 3k, 3m,
                        3g, 3t). If the limit is exceeded, the least recently
                        used images are removed first. Worker will not remove
                        any images if this option is not specified.
```

worker/worker.py
kashizui: done!

Issue #517, opened 2016-07-04
Create a pip package


kashizui:
`pip install codalab-cli`

maxwang7:
Instructions: https://packaging.python.org/distributing/

TODOs:
- [ ] create setup and initialization files
  - [ ] choose a versioning scheme
  - [ ] Run setup commands and place the `cl` inside the user's search paths
- [ ] Package project as a Pure Python Wheel
- [ ] Upload project to PyPi
- [ ] Test installation on different machines using pip and easy_install

fabeschan:
@maxwang7 This link has been useful for my packages, see if it helps you. IMO it explains the process  in a much clearer manner:

http://python-packaging.readthedocs.io/en/latest/index.html

maxwang7:
Thank you Fabian!

maxwang7:
Thank you Fabian!

maxwang7:
Some questions:
For the package name, should we use a hyphenated (codalab-cli) or underscore (codalab_cli) version?
Are all the requirements we need inside of requirements.txt?
Are we using semantic versioning? If so, is 0.0.1 the correct first version label?
Are we using Percy's name and email address?
Should we create a CodaLab PyPI account?

kashizui:
I'm actually not sure, what is the canonical form for pip packages?
All the rquirements we need to run the CLI client should be in requirements.txt.
We are using semver, and our current version number is defined in `codalab-cli/codalab/common.py`
re: name/address/account, I don't know -- @percyliang?

percyliang:
Can we use codalab.worksheets@gmail.com?

kashizui:
package name: use hyphen codalab-cli
Name: CodaLab

maxwang7:
Some notes:

To get all the packages, use setuptools' find_packages() tool: http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages

There are a lot of different recommendations on how to add a command line tool:
- This tutorial uses a new __main__.py file and a console entrypoint that hooks into that file: https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/
- Virtualenv seems to take a similar approach: https://github.com/pypa/virtualenv/blob/master/setup.py
- Ok, this is the recommended approach that's also given in setuptools' documentation: https://packaging.python.org/distributing/#scripts

maxwang7:
[...]

PR #677, opened 2017-02-07
pip install

maxwang7:
Issue: #517 

Tested successfully on the following Dockerized Ubuntu versions: 12.04, 14.04, 16.04.

Apologies for getting in this PR late. I had to wrestle with some PyPI and setup.py idiosyncracies that weren't expected.

### DIFF
--- a/codalab/bin/cl.py
+++ b/codalab/bin/cl.py
@@ -109,11 +109,20 @@ def do_bundle_manager_command(bundle_cli, args):
 
     manager.run(args.sleep_time)
 
-
-if __name__ == '__main__':
+def main():
     cli = BundleCLI(CodaLabManager())
     try:
         cli.do_command(sys.argv[1:])
     except KeyboardInterrupt:
         print 'Terminated by Ctrl-C'
         sys.exit(130)
+
+if __name__ == '__main__':
+    main()
+
+#    cli = BundleCLI(CodaLabManager())
+#    try:
+#        cli.do_command(sys.argv[1:])
+#    except KeyboardInterrupt:
+#        print 'Terminated by Ctrl-C'
+#        sys.exit(130)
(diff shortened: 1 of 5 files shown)


codalab/bin/cl.py
```diff
@@ -109,11 +109,20 @@ def do_bundle_manager_command(bundle_cli, args):
...
+
+if __name__ == '__main__':
+    main()
+
+#    cli = BundleCLI(CodaLabManager())
+#    try:
+#        cli.do_command(sys.argv[1:])
+#    except KeyboardInterrupt:
+#        print 'Terminated by Ctrl-C'
+#        sys.exit(130)
```
kashizui: remove comments?

setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup, find_packages
+from setuptools.command.install import install
+import os
+
+setup(name='codalab-cli',
+    version='0.2.1',
```
kashizui: Do you think there's any way we could link this with the CODALAB_VERSION defined in `codalab.common`, to avoid bringing the variables we have to update to 3?

kashizui (commented):
Just curious, how does

kashizui:
Oops ignore my unfinished comment.
Looks good, modulo a couple comments.

setup.py
maxwang7: Ok, I did some digging, what Django does is they store the version number in [`django/__init__.py`](https://github.com/django/django/blob/master/django/__init__.py) (like we do) and then reads that number from [`setup.py`](https://github.com/django/django/blob/master/setup.py#L32). If that sounds good to you, I'll go ahead and make that change.

codalab/bin/cl.py
maxwang7: Yes, thanks!

maxwang7: Removed

maxwang7: Ok, removed!

setup.py
kashizui: Sounds good to me

kashizui (approved):

PR #682, opened 2017-02-09
Collect stderr in monitor

kashizui:
Use Popen.communicate() so we can get both stdout and stderr, and as separate strings.

Previously, the monitor error notification emails did not include stderr in the output dump.


percyliang (approved):
Good call.  LGTM

PR #684, opened 2017-02-12
Competition evaluation script

kashizui:


kashizui:
Implementation is basically complete! I'll demo it at our next meeting.

kashizui:
For the leaderboard itself, I'll plan on putting together a simple JavaScript page that will load the bundles tagged with the leaderboard tag.

kashizui:
Also needs more documentation on what it does exactly.

kashizui:
Ready for review @maxwang7 @percyliang

scripts/competitiond.py
```diff
@@ -0,0 +1,580 @@
...
+    3. Tag the resulting test run bundle with {predict.tag}, untagging
+       any previous test run bundles for the same submitter.
+    4. Run {evaluate} command with the test run bundle.
+    5. Tag the resulting evaluation bundle with {evaluate.tag}, untagging
+       any previous evaluation bundles for the same submitter.
+
+All bundles created by this daemon are added to {log_worksheet_uuid}.
+Each user will be limited to {max_submissions_per_day}.
+
+Example config file:
```
kashizui: Need to update this to match latest changes

codalab/lib/bundle_util.py
```diff
@@ -26,3 +31,278 @@ def bundle_to_bundle_info(model, bundle):
     result['args'] = worksheet_util.interpret_genpath(result, 'args')
 
     return result
+
+
+def mimic_bundles(client,
```
kashizui, line 36: This is just copied from `bundle_cli.py`, with some additional arguments.

codalab/rest/bundles.py
```diff
@@ -133,9 +133,9 @@ def _create_bundles():
     """
     Bulk create bundles.
 
-    |worksheet_uuid| - The parent worksheet of the bundle, add to this worksheet
-                       if not detached or shadowing another bundle. Also used
-                       to inherit permissions.
+    |worksheet| - The uuid of the parent worksheet of the bundle, add to this
```
kashizui: Should add more REST documentation while I'm at it.

scripts/competitiond.py
kashizui: done

codalab/lib/codalab_manager.py
```diff
@@ -362,7 +362,7 @@ def rest_oauth_handler(self):
         from codalab.server.auth import RestOAuthHandler
         address = 'http://%s:%d' % (self.config['server']['rest_host'],
                                     self.config['server']['rest_port'])
-        return RestOAuthHandler(address, self.model())
+        return RestOAuthHandler(address)
```
maxwang7, line 365: Why is the model parameter being removed?

scripts/competitiond.py
```diff
@@ -0,0 +1,580 @@
...
+import re
+import signal
+import sys
+import time
+import traceback
+from collections import defaultdict
+
+from marshmallow import Schema, fields, ValidationError
+
+sys.path.append('.')
```
maxwang7, line 94: What's the purpose of this line? How would the behavior be different without it? Is it to avoid using a system version of Codalab? And if so, why is using the system version of Codalab bad?

scripts/competitiond.py
```diff
@@ -0,0 +1,580 @@
...
+        submission_ids = set()
+        for predict_bundle in last_tests:
+            submit_info = json.loads(predict_bundle['metadata']['description'])
+            timestamp = predict_bundle['metadata']['created']
+            submission_ids.add(submit_info['bundle_id'])
+            submission_times[submit_info['submitter_id']].append(timestamp)
+
+        # Drop submission if user has exceeded their quota
+        now = time.time()
+        one_day_ago = now - DAY_SECONDS
```
maxwang7: Do we want to make the `DAY_SECOND` variable user configurable? For example, running a competition where users can submit once a week.

scripts/competitiond.py
```diff
@@ -0,0 +1,580 @@
...
+        }
+
+        # Do dry run to check if the submission has the right dependencies.
+        # If the submission bundle is not mimicked (i.e. not in the mimic plan),
+        # that means that none of its ancestors are in the set of bundles that
+        # we are trying to replace.
+        if find_mimicked(mimic_bundles(dry_run=True, **mimic_args)) is None:
+            logger.info(
+                "Submission {uuid} by {owner[user_name]} is missing "
+                "expected dependencies.".format(**submit_bundle))
```
maxwang7, line 391: Didn't know you could put strings across multiple lines like that. Cool!

codalab/lib/codalab_manager.py
kashizui: It's not actually being used by the class -- a relic from darker times.

kashizui: Forgot to remove it in the refactor.

scripts/competitiond.py
kashizui: I also saw someone else do it and thought it was incredible

scripts/competitiond.py
kashizui: - Purpose is to add current working directory to the PYTHONPATH so that we can import the codalab libraries. I copied this over from the other scripts in the directory since we used to rely on this trick before we had pip install. Now I realize it might not be necessary.
- As you surmised, it would use the version of CodaLab in the repo rather than the system version.
- I wasn't purposefully trying to avoid the system version.
- But now that I think about it, I think avoiding the system version is a good idea because then script will always be written to rely on version of CodaLab that is in that specific commit. (Otherwise we will have to have some way to specify which version of CodaLab this script relies on, since might use code that was only introduced in v0.2.4 for example.)

scripts/competitiond.py
kashizui: Good idea

kashizui:
@maxwang7 Addressed all comments, please take another look!

scripts/competitiond.py
maxwang7: OK, that makes sense since users will be git cloning the repo to run this script.

maxwang7 (approved):

PR #685, opened 2017-02-13
fixes escaping problem and uses the correct email as the user email

maxwang7:
Notes on escaping: The bottle template engine allows strings to not be escaped by prepending a `!` to the expression. See documentation here: https://bottlepy.org/docs/dev/stpl.html

The email address was changed from `support_email` to `user_email`


percyliang (approved):
LGTM

PR #686, opened 2017-02-13
List CLI commands when `cl` executed with no arguments

maxwang7:
I thought that this method of checking the number of arguments when argparse errors was the best method to give a help message to users. There are some other proposed ways of doing so on this StackOverflow post: http://stackoverflow.com/questions/4042452/display-help-message-with-python-argparse-when-script-is-called-without-any-argu. Let me know if you think one is better than the other.


kashizui (approved):
I agree this is the better approach since we already override `error()`.

PR #687, opened 2017-02-15
adds user's name to email

maxwang7:
Also changes the location of `support_email` in `config.json` (puts it under server)

### DIFF
--- a/codalab/rest/help.py
+++ b/codalab/rest/help.py
@@ -14,7 +14,7 @@ from codalab.server.authenticated_plugin import (
 def fetch_help():
     message = request.json['message']
 
-    if local.config and 'support_email' not in local.config['server']:
+    if 'server' in local.config and 'support_email' not in local.config['server']:
         print >>sys.stderr, 'Warning: No support_email configured, so no email sent.'
         print >>sys.stderr, 'User\'s message: %s' % message
         return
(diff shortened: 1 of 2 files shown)


kashizui (approved):
Most people don't provide their real names, in which case their real_name will show up as `None None` -- maybe want to add a bit of logic to handle this, but not necessary since this email is only sent to us.

maxwang7:
Ok, I'll add that logic and merge

codalab/rest/help.py
```diff
@@ -13,18 +13,20 @@
 @post('/help/', apply=AuthenticatedPlugin(), skip=UserVerifiedPlugin)
 def fetch_help():
     message = request.json['message']
-    if 'support_email' not in local.config:
+
+    if local.config and 'support_email' not in local.config['server']:
```
kashizui: when will config be falsey?

kashizui: do you mean `if 'server' in local.config ...`?

maxwang7: Right... would it make sense to check for local.config['server']?

maxwang7: Or is that first check just not necessary?

kashizui: I have a similar check elsewhere, so I think it's good to have for now.

maxwang7: Ok, sounds good

PR #689, opened 2017-02-17
Adds unicode support in support emails

maxwang7:
Tested locally with SendGrid. Will test again on wstest.

### DIFF
--- a/codalab/lib/emailer.py
+++ b/codalab/lib/emailer.py
@@ -34,7 +34,8 @@ class SMTPEmailer(Emailer):
         self.port = port
         self.use_tls = use_tls
 
-    def send_email(self, subject, body, recipient, sender=None, mime_type='plain'):
+    def send_email(self, subject, body, recipient, sender=None,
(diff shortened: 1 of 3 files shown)


codalab/lib/emailer.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, host, user, password, default_sender, server_email, port=587,
         self.port = port
         self.use_tls = use_tls
 
-    def send_email(self, subject, body, recipient, sender=None, mime_type='plain'):
+    def send_email(self, subject, body, recipient, sender=None,
+                   mime_type='plain', charset='us-ascii'):
```
kashizui, line 38: add a line to the docstring below for how to use this kwarg

kashizui (approved):
lgtm modulo comment

codalab/lib/emailer.py
maxwang7: Added, as well as a docstring for the `mime_type` keyword argument

Issue #671, opened 2017-02-05
Make all groups publicly visible


maxwang7:
Everyone can know a group exists

Questions

Where are the permissions enforced for this right now?

How do we want to expose the list of groups?

Look in get_groups_info (thanks Stephen!) [in this file](https://github.com/codalab/codalab-cli/blob/f33ff172db28f592aac97cf75c0911040ac3bf22/codalab/rest/util.py#L151-L161). Pretty sure we can do something like setting `group_info` to `None`, but do some more digging on where else `get_group_info` is being used.

Time estimate: 4 hours

maxwang7:
- Where are the permissions enforced for this right now?

- How do we want to expose the list of groups?

maxwang7:
Following the proposed solution in the issue description, here are descriptions of the uses of `get_group_info`. They are all in rest/groups.py:

- fetch_group(): Allows the user to fetch a group. --> Our goal with this issue is to allow fetch_group() to fetch any group.
- delete_groups(): Uses `get_group_info` as a way of checking the user's permissions. Changing `get_group_info`'s behavior would effect this function (causing a major security hole -- anyone could delete anyone else's group!).
- add_group_members_helper(): Uses `get_group_info` as a way to fetch the group, but only if the user is an admin of the group. Changing `get_group_info` should not effect this method since this function needs admin privileges in the group.
- delete_group_members(): Same as with add_group_members_helper(). Uses `get_group_info` as a way to fetch the group, but only if the user is an admin of the group. Changing `get_group_info` should not effect this method since this function needs admin privileges in the group.

In summary, it looks like there will need to be some functionality that support the use case in `delete_groups` so that an arbitrary user isn't allowed to delete a group that s/he doesn't belong to. I'm going to add a keyword argument `access_all_groups` that defaults to False. Setting it to True gives the caller access to all groups. The only place where I set it to True will be for the fetch_group() function. This way I can ensure that no previous functionality is broken from this change.

---

Testing

General testing:

- Test `cl ginfo <group_spec>` by using the `group_spec` of a group that doesn't belong to the logged in user. Expected result: The group's information should still be successfully shown.
- Test `cl grm <group_spec>` by using a `group_spec` of a group that doesn't belong to the logged in user. Expected result: The group should not be deleted.

bundles:

[...]

PR #690, opened 2017-02-20
Make all groups public

maxwang7:
See issue for sanity check tests.
Issue: #671 

Note: marshmallow library was added because I experienced bugs with a higher version number (2.10). Downgrading to 2.6.1 fixed my problems.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1395,12 +1395,6 @@ class BundleModel(object):
             if user_id is None:
                 # Not logged in: include only public group
                 group_restrict = (table.c.group_uuid == self.public_group_uuid)
-            elif user_id != self.root_user_id:
-                # Logged in but not root: include only public group and groups that user_id is in
-                group_restrict = or_(
-                    table.c.group_uuid == self.public_group_uuid,
-                    table.c.group_uuid.in_(select([cl_user_group.c.group_uuid]).where(cl_user_group.c.user_id == user_id))
-                )
             else:
                 # Logged in as root: include all groups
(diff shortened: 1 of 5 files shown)


percyliang:
At a high-level looks reasonable to me, but this is the type of thing we really want to double check.  Can @kashizui take a detailed look?

kashizui:
Testing on my own machine now

kashizui:
Almost everything looks good, tried all the test cases you described in #671 and more.
Just one minor thing I noticed:
When I view a bundle that I have `none` permissions on, I can still see all the group permissions defined on that bundle. I think unless I have at least `read` permissions on it, I shouldn't be able to see any group permissions on the bundle (as with any of the other attributes).

```
➜  cl info -v 0xd24cd0
bundle_type               : private
uuid                      : 0xd24cd01b5d954f75bc554dbe2e0acde6
data_hash                 : <none>
state                     : <none>
command                   : <none>
owner                     : <unknown>
name                      : <private>
children:
host_worksheets:
  home-codalab(0x72a6926d82dd4b9d98f05d8c566ebdaf)
permission: none
group_permissions:
  cs224n(0x53976c):all
```

This resulted from our decision to remove one of the conditions in `batch_get_group_permissions()`. I just pushed changes to fix this, by adding a check in `codalab.rest.util.get_bundle_infos()` and in `codalab.rest.worksheets.get_worksheet_info()` that only fetches or includes the group permissions if the user has at least read permission on the object.

kashizui (approved):
LGTM @maxwang7 have a look at the changes I made and go ahead and merge if you agree with them.

PR #691, opened 2017-02-20
Update `cl` list commands

maxwang7:
Uses do_command to call do_help_command
Issue: #686

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -176,7 +176,7 @@ class CodaLabArgumentParser(argparse.ArgumentParser):
     def error(self, message):
         # Adapted from original:
         # https://hg.python.org/cpython/file/2.7/Lib/argparse.py
-        if len(sys.argv) == 1 or (len(sys.argv) == 2 and (sys.argv[1] == '--help' or sys.argv[1] == '-h')):
+        if len(sys.argv) == 1 or (len(sys.argv) == 2 and sys.argv[1] in ['--help', '-h']):
             self.cli.do_command(['help'])
             self.exit(2)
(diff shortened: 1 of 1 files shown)


kashizui:
This ends up breaking functionality for `cl <command> -h` or `cl <command> --help`, which should show just the help for a single command. What about just updating line 180 to call `do_command(...)`?

kashizui (changes requested):
This ends up breaking functionality for `cl <command> -h` or `cl <command> --help`, which should show just the help for a single command. What about just updating line 180 to call `do_command(...)`?

maxwang7:
I added an extra check so that `cl --help` and `cl -h` would both produce the help information as well. It feels a bit hacky but I couldn't find anywhere else that seemed more appropriate to add this code.

time: 40 minutes (wow, a lot longer than expected...)

codalab/lib/bundle_cli.py
```diff
@@ -176,8 +176,8 @@ def print_help(self, out_file=None):
     def error(self, message):
         # Adapted from original:
         # https://hg.python.org/cpython/file/2.7/Lib/argparse.py
-        if len(sys.argv) == 1:
-            self.print_help()
```
percyliang, line 180: Can you delete `print_help` now?

percyliang (approved):
Make sure all the variants of --help, help, -h with and without other commands work.

maxwang7:
Confirmed that all variants work. I tried these:

- `cl`
- `cl -h`
- `cl --help`
- `cl help`
- `cl add --help`
- `cl add -h`

kashizui (approved):

PR #695, opened 2017-02-24
Add GPU worker support

fabeschan:
- [x]  detect number of gpus on worker

- [x] add gpu info to checkin request

- [x] migrate database (add gpu column to worker table)

- [x] add functionality to filter workers based on requested_gpus


worker/worker.py
```diff
@@ -104,13 +107,24 @@ def _get_memory_bytes(self):
             # Fallback to sysctl when os.sysconf('SC_PHYS_PAGES') fails on OS X
             return int(check_output(['sysctl', '-n', 'hw.memsize']).strip())
 
+    def _get_gpu_count(self):
+        if not self._docker._use_nvidia_docker:
+            return 0
+
+        container_id = self._docker.run_nvidia_smi('-L', 'gcr.io/tensorflow/tensorflow:latest-gpu')
```
kashizui: I think we can use a more basic image than the tensorflow one? Maybe one of these:
https://hub.docker.com/r/nvidia/cuda/

worker/docker_client.py
```diff
@@ -229,6 +229,35 @@ def download_image(self, docker_image, loop_callback):
                 loop_callback(status)
 
     @wrap_exception('Unable to start Docker container')
+    def run_nvidia_smi(self, args, docker_image):
```
kashizui: I think I can see why but wondering what your reasoning was for running 'nvidia-smi' inside Docker rather than directly?

kashizui (changes requested):
Overall looks pretty good, and it works!

worker/docker_client.py
fabeschan: I think in the future we're going to deploy workers as docker containers, in which case running `nvidia-smi` inside docker would not require us to have two different versions of worker images (one for gpu, one not) to deploy.

worker/worker.py
fabeschan: That's a great suggestion.

worker/docker_client.py
fabeschan: Done!

kashizui: Ah that's very inception-y to think about haha.

kashizui (approved):

worker/docker_client.py
fabeschan: ;)

Issue #700, opened 2017-03-14
Automatically pull nvidia/cuda:8.0-runtime


hhexiy:
Just got this error upon job startup. Tried on Jag8 and Jag15 and the result is the same.
```
DockerException: Unable to start Docker container: {"message":"No such image: nvidia/cuda:8.0-runtime"} 
```
See complete failure message here: https://codalab.stanford.edu/bundles/0x12fe95498133408cbc205c7afee9d99e/

kashizui:
Sorry @hhexiy, we had some unexpected incompatibilities in our last deploy earlier today. I just reverted back to an earlier version now until we have a fix. Let me know if it still doesn't work.

hhexiy:
working now. thanks!

kashizui:
@fabeschan Could you update our worker code to automatically pull this image? This is causing issues for people who are running their own workers as well. We have existing code to automatically pull images, and we can factor it out as "ensure_docker_image" or something.

fabeschan:
PR is merged, should be live on next deploy.

PR #704, opened 2017-03-15
Fix GPU worker checkin issues

### LINKED ISSUE #702: RepoDigests 'list index out of range' error
This is because `RepoDigests` is an empty list for images that were pulled before Docker repository v2 was released.

The offending code in `worker.docker_client`:
```
    def get_image_repo_digest(self, request_docker_image):
        logger.debug('Fetching Docker image metadata for %s', request_docker_image)
        with closing(self._create_connection()) as conn:
            conn.request('GET', '/images/%s/json' % request_docker_image)
            create_image_response = conn.getresponse()
            if create_image_response.status != 200:
                raise DockerException(create_image_response.read())
            contents = json.loads(create_image_response.read())
        return contents.get('RepoDigests', [''])[0]
```

@fabeschan Let's make this more robust. This will be a problem for anyone upgrading their CodaLab instance/worker. Maybe at the minimum, some logic to either fall back to empty string `''` or the image tag if `RepoDigests` is an empty list?

fabeschan:
Fixes #702 and #700

### DIFF
--- a/worker/docker_client.py
+++ b/worker/docker_client.py
@@ -316,9 +316,11 @@ nvidia-docker-plugin not available, no GPU support on this worker.
                 def update_status(status):
                     logger.info('Downloading Docker image for running nvidia-smi: ' + status)
                 self.download_image(docker_image, update_status)
-                return _start_nvidia_smi()
+                container_id = _start_nvidia_smi()
             else:
                 raise
+
+        return container_id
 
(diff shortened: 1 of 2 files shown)


kashizui (approved):
Looks pretty good to me, just wondering if it's possible to share/reuse more code for creating/starting containers?

PR #711, opened 2017-03-24
MultiDiskBundleStore

fabeschan:
No Hash ring

Balanced store because it uses the Balance algorithm: pick whatever partition that has the largest free space


fabeschan:
Thinking about making the hashring class a subclass of it

codalab/lib/bundle_store.py
```diff
@@ -80,6 +60,336 @@ def reset(self):
         """
         pass
 
+class BalancedMultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin, BundleStoreHealthCheckMixin):
+    """
+    responsible for taking a set of locations and load-balancing the placement of
```
percyliang: capitalize

codalab/lib/bundle_store.py
```diff
@@ -80,6 +60,336 @@ def reset(self):
...
+
+    def reset(self):
+        """
+        Delete all stored bundles and then recreate the root directories.
+        """
+
+        pass
+
+        '''
+        # Do not run this function in production!
```
percyliang: Can we just remove the `reset` function?

codalab/lib/bundle_store.py
```diff
@@ -80,6 +60,336 @@ def reset(self):
...
+        partition_abs_path = os.path.join(self.partitions, partition)
+
+        try:
+            print partition_abs_path
+            path_util.check_isvalid(partition_abs_path, 'rm-partition')
+        except:
+            print >> sys.stderr, "Partition with name '%s' does not exist. Run `cl ls-partitions` to see a list of mounted partitions." % partition
+            sys.exit(1)
+
+        print >> sys.stderr, "Unlinking partition %s from CodaLab deployment..." % partition
```
percyliang: Do we need to check that the partition we're removing is empty or something?

codalab/lib/bundle_store.py
```diff
@@ -80,6 +60,336 @@ def reset(self):
...
+            disk = None
+            for n in self.nodes: # go through every partition
+                bundle_path = os.path.join(self.partitions, n, BalancedMultiDiskBundleStore.DATA_SUBDIRECTORY, uuid)
+                if os.path.exists(bundle_path):
+                    disk = n
+
+            if disk is None:
+                # return disk with largest free space
+                disk = max(self.nodes, key=lambda x:
+                        self.get_node_avail_frac(os.path.join(self.partitions, x, MultiDiskBundleStore.DATA_SUBDIRECTORY))
```
percyliang: Why wouldn't you return the disk with the max absolute space?  Seems like that's more sensible...

percyliang (changes requested):
Great.  If this version works well enough, let's just get rid of the hash ring version and call `BalancedMultiDiskBundleStore` `BalancedMultiDiskBundleStore` instead.  Less code to maintain.

Can you check how the performance suffers from having, say 8 disks when you print out a worksheet?

codalab/lib/bundle_store.py
fabeschan: Yup, I was considering that as well

codalab/lib/bundle_store.py
fabeschan: I think that would be unnecessary. Since it doesn't do anything destructive to bundles, if someone removes a partition by mistake, all he/she has to do is to re-add it.

codalab/lib/bundle_store.py
fabeschan: I thought I had a good reason to do that... guess not. I'll return the one with the max absolute space

codalab/lib/bundle_store.py
fabeschan: okay

fabeschan:
Yup, I'll set up a server with 8 disks to test it out this weekend

fabeschan:
@percyliang I ran cl `print` on a worksheet with 550 bundles across 8 disks/partitions
fresh, empty cache: 1.02s
subsequence runs with warm cache: 0.82s

I'm not sure if `cl print` actually needs to access the bundle store however, so I also loaded the worksheet on a browser. Seems to take approx the same time to load in both situations.

codalab/lib/bundle_store.py
percyliang: Perhaps we should print out a warning...since stuff in the DB doesn't get removed...

codalab/lib/bundle_store.py
```diff
@@ -74,11 +54,320 @@ def get_bundle_location(self, data_hash):
         """
         pass
 
-    def reset(self):
+class BalancedMultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin, BundleStoreHealthCheckMixin):
```
percyliang: Why don't we just replace the hash ring implementation with `BalancedMultiDiskBundleStore`?

codalab/lib/bundle_store.py
fabeschan: That's a good point, I'll print out a warning.

codalab/lib/bundle_store.py
fabeschan: Alright, I was planning to do that once we switched over, but I'll do it in this PR directly.

codalab/lib/bundle_cli.py
```diff
@@ -2780,8 +2780,8 @@ def do_reset_command(self, args):
         self._fail_if_not_local(args)
         if not args.commit:
             raise UsageError('If you really want to delete EVERYTHING, use --commit')
-        print >>self.stdout, 'Deleting entire bundle store...'
-        self.manager.bundle_store().reset()
+        #print >>self.stdout, 'Deleting entire bundle store...'
```
percyliang: Remove comments?

codalab/lib/bundle_store.py
```diff
@@ -102,16 +92,40 @@ def __init__(self, codalab_home):
...
+        st = os.statvfs(node)
+        free = st.f_bavail * st.f_frsize
+        return free
+
     @require_partitions
     def get_bundle_location(self, uuid):
         """
-        get_bundle_location: Perform a lookup in the hash ring to determine which disk the bundle is stored on.
+        get_bundle_location: look for bundle in the cache, or if not in cache, go through every partition.
+        If not in any partition, return disk with largest free space
```
percyliang: add period

codalab/lib/bundle_store.py
```diff
@@ -74,25 +52,37 @@ def get_bundle_location(self, data_hash):
...
-        pass
-
-
 class MultiDiskBundleStore(BaseBundleStore, BundleStoreCleanupMixin, BundleStoreHealthCheckMixin):
     """
-    A MultiDiskBundleStore is responsible for taking a set of locations and load-balancing the placement of
-    bundle data between the locations. It accomplishes this goal using a consistent hash ring, a technique
-    discovered by Karger et al. in 1997.
+    Responsible for taking a set of locations and load-balancing the placement of
+    bundle data between the locations. Builds an LRU cache of bundle locations over time.
```
percyliang: Add longer description about how we have multiple disks and how the bundles can be distributed in an arbitrary way; and how a cache is built up for efficiency.

percyliang (approved):
Let's put this on wstest and bang on it for a while.

codalab/lib/bundle_store.py
```diff
@@ -102,16 +92,40 @@ def __init__(self, codalab_home):
...
-        disk = self.ring.get_node(uuid)
+        if uuid in self.lru_cache:
+            disk = self.lru_cache.pop(uuid)
+        else:
+            disk = None
+            for n in self.nodes: # go through every partition
+                bundle_path = os.path.join(self.partitions, n, MultiDiskBundleStore.DATA_SUBDIRECTORY, uuid)
+                if os.path.exists(bundle_path):
+                    disk = n
+
```
percyliang: add `break`

codalab/lib/bundle_cli.py
fabeschan: done

codalab/lib/bundle_store.py
fabeschan: done.

codalab/lib/bundle_store.py
fabeschan: done.

codalab/lib/bundle_store.py
fabeschan: good catch; done

fabeschan:
@percyliang FYI this patch has been deployed on wstest

percyliang:
Cool.  I created a new partition, uploaded a bundle and moved it to the new
partition, did a `killall -HUP supervisord` to restart things.  Seems to
work, but perhaps should test it a bit more rigorously (in particular, I
haven't checked whether it always chooses the freest partition) before
deploying on prod.

PR #712, opened 2017-03-26
remove listing all files of bundles when loading worksheet

percyliang:
Prevent slowness due to a bundle having a huge number of files.

This doesn't seem to be used anywhere else.


kashizui (approved):
Hm I haven't seen this used either.

PR #729, opened 2017-04-19
Add sanity check for competition script

kashizui:
Also factor out code for creating temporary CL instance.


test-cli.py
```diff
@@ -998,6 +1061,76 @@ def test(ctx):
...
 
 
+@TestModule.register('competition')
+def test(ctx):
+    """
+    Sanity-check the competition script.
+    """
+    submit_tag = 'submit'
+    eval_tag = 'eval'
+    with temp_instance() as remote:
```
percyliang, line 1071: Why do the competition tests need their own temp instance?

percyliang (approved):
Nice refactor.  LGTM

test-cli.py
kashizui: To avoid any clashing in the tags namespace, and also it's harder to keep track of what bundles get created by the competition script (esp if there are any bugs) so it's easier to just do it in a temp instance and then nuke the whole thing.

fabeschan (approved):
lgtm

PR #730, opened 2017-04-19
Revise local-dev cli help messages and disable push-image, preferring docker push instead

fabeschan:
No need for a simple wrapper for pushing an image, people should learn docker anyway


kashizui (approved):
lgtm

PR #732, opened 2017-04-21
Make compression API saner

### LINKED ISSUE #708: Codalab automatically uncompressing files
It appears that codalab automatically uncompresses files:
```
$ cl upload -p foo.tgz
$ cl run :foo.tgz "tar ztf foo.tgz" -n foo
```
the `run` command fails:
```
$ cl cat foo/stderr
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

kashizui:
Fixes #708

Problem: Conflation of a "transport" layer detail with "application"
layer detail, namely compression, since we compress bundles on the fly
during transfer. (Usage of "transport" and "application" are only by
analogy and don't actually correspond to the OSI stack.)

Bug #1: bundles that are gzip archives accidentally get unzipped on
worker
- Tarred directories currently indicated by filename
  (Content-Disposition)
- not safe when the file itself is a unextracted tar gz, should probably
  be some additional header
- SOLUTION: Target-Type header indicates directory|file, and so if a
  bundle is an archive and should be kept packed on the worker, this can
  be indicated by "Target-Type: file".

Bug #2: bundles that are ZIP archives
- ZIP archives are already compressed, so server skips the compression
  worker client doesn't check Content-Encoding header and always assumes
  bundles are gzipped, so attempts to gunzip the ZIP archive and fails
- SOLUTION: Worker decompresses streams ONLY if Content-Encoding
  indicates so.

Server Policy:
- Single-file bundles (including packed archives of directories):
    - Content-Disposition: filename=[bundle name or target filename]
        - regardless of content encoding, since that is akin to a
          "transport" layer detail, while filename is more an
          "application" layer detail
    - Content-Type: [guess based on filename]
        - make sure NOT to user mimetypes.guess_type, which will ignore
          compression file extensions like ".gz" in determining mimetype
          (considering it to be the "encoding" instead)
    - Content-Encoding:
        - gzip if client accepts gzip
            - Should ALWAYS compress for simplicity, since the case
              where a bundle is already compressed is rare.
            - Helps disambiguate the case where
            - We also have no control over compression of the user
        - identity otherwise
            - send raw file, archive or not, if client doesn't accept
              gzip
    - Target-Type: file
- Directories:
    - Content-Disposition: filename=[bundle name or directory
      name].tar.gz
        - Always send a gzipped archive
    - Content-Type: application/gzip
    - Content-Encoding: identity
    - Target-Type: directory
        - Need some way to indicate to CodaLab-specific clients that it
          is a directory, to differentiate it from bundles that are
          packed archives, since the filename will be the same. (This
          does not matter to browsers, which for all extents and
          purposes should treat directory downloads as just a gzip
          archive. Hence the filename definition above.)

Client Policy:
- Content-Disposition:
    - CLI (and browser) should use filename when saving to disk
    - Worker should not need to know the filename.
- Content-Type:
    - only matters for the browser or other clients that care about how
      to render the contents (for example, display contents as a new tab
      for text, or trigger a download for binary data?)
- Content-Encoding:
    - gzip: unzip on the fly, transparently to the user of the client
      class/library. The actual object is the unzipped stream.
    - identity: return raw stream
- Target-Type:
    - file: application treats returned stream (potentially after
      decoding) as a file [the file itself might be an archive but that
      should be irrelevant to the application]
    - directory: application treats returned stream as a ".tar.gz" file
      to be unpacked as a directory

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -569,12 +569,8 @@ class JsonApiClient(RestClient):
             params['head'] = head
         if tail is not None:
             params['tail'] = tail
-        response = self._make_request('GET', request_path, headers=headers,
-                                      query_params=params, return_response=True)
-
-        if response.headers.get('Content-Encoding') == 'gzip':
-            return un_gzip_stream(response)
-        return response
+        return self._make_request('GET', request_path, headers=headers,
+                                  query_params=params, return_response=True)
 
     @wrap_exception('Unable to upload contents of bundle {1}')
     def upload_contents_blob(self, bundle_id, fileobj=None, params=None,
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -341,7 +341,7 @@ def _fetch_bundle_contents_info(uuid, path=''):
       "data": {
           "name": "<name of file or directory>",
           "link": "<string representing target if file is a symbolic link>",
-          "type": "<file|directory>",
+          "type": "<file|directory|link>",
           "size": <size of file in bytes>,
           "contents": {
             "name": ...,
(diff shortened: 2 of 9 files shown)


kashizui:
TODO:

- [x] Update REST API docs to reflect these new standardized details
- [x] Bump worker version
- [x] Test migration (done, tested on wstest)

kashizui:
Tested and ready for review @percyliang @fabeschan

fabeschan (approved):

PR #743, opened 2017-05-02
Improve Torque scheduler

kashizui:
Following Jimmy's advice:

- use `-k n` instead of `-e /dev/null -o /dev/null` to disable recording stdout/stderr streams (we redirect them manually to log files in NFS in `worker.sh`, so we don't need to capture stdout/stderr through PBS)
- set workdir for torque job to `/tmp` to avoid chdir permissions problems. the job doesn't use the working directory anyway
- allow configurable throttle on `qsub` requests, since torque scheduler sometimes has trouble keeping up with many consecutive requests.
- fire `qsub` requests in a different thread to avoid impact on latency of other operations

@percyliang please review

### DIFF
--- a/codalab/worker/torque_bundle_manager.py
+++ b/codalab/worker/torque_bundle_manager.py
@@ -23,6 +23,7 @@ class TorqueBundleManager(BundleManager):
         self._torque_bundle_service_url = torque_config['bundle_service_url']
         self._torque_password_file = torque_config['password_file']
         self._torque_log_dir = torque_config['log_dir']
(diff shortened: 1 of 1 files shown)


percyliang (approved):

fabeschan (approved):

kashizui:
Tested and it worked, but realized that sleeping in the main thread is a bad idea.

kashizui:
Also should address in the PR:

- ~cleaning up worker jobs that no longer have bundles associated with them~
- ~making sure resource management works~

kashizui:
Tested successfully

Issue #309, opened 2016-01-13
make monitor script use website-config.json to determine address to send email from


percyliang:

percyliang:
Fixed.

PR #744, opened 2017-05-03
Clean up legacy

### LINKED ISSUE #633: remove legacy bundle service

kashizui:
- [x] (Merge with existing legacy cleanup branch w/e that is doing)
- [x] Move everything out of BundleService into proper modules in the rest package (web terminal)
- [x] Keep existing legacy APIs used by the frontend in one neat package that can be slowly deprecated
- [x] Grep for all the legacy kwargs and TODOs and FIXMEs
- [x] replace RPC calls in Cli
- [x] Simplify the multiple (cyclical) levels of indirection in worksheet interpretation
- [x] Remove base64 encoded fields that are immediately decoded.
- [x] check that cat_target file stuff is necessary
- [x] better error handling
- [x] regenerate REST docs
- [x] remove already migrated APIs
- [x] remove deprecated routes
- [x] address comments
- [x] human-readable REST module names in docs
- [x] fill out some more rest docs

Fixes #633


kashizui:
@percyliang basically complete and ready for review

codalab/rest/cli.py
```diff
@@ -0,0 +1,164 @@
...
+from codalab.lib import (
+    bundle_cli,
+    worksheet_util,
+)
+from codalab.lib.codalab_manager import CodaLabManager
+from codalab.objects.oauth2 import OAuth2Token
+
+
+@post('/cli/command')
+@post('/api/worksheets/command/')  # DEPRECATED ROUTE
```
percyliang: Do we still need to keep this route?

codalab/rest/cli.py
```diff
@@ -0,0 +1,164 @@
...
+def post_worksheets_command():
+    # TODO(klopyrev): The Content-Type header is not set correctly in
+    # editable_field.jsx, so we can't use request.json.
+    data = json.loads(request.body.read())
+
+    if data.get('raw_command', None):
+        data['command'] = worksheet_util.get_worksheet_info_edit_command(data['raw_command'])
+
+    if not data.get('worksheet_uuid', None) or not data.get('command', None):
```
percyliang: In all of this, don't need the `None` argument to get since that's the default.

codalab/rest/cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@post('/cli/command')
+@post('/api/worksheets/command/')  # DEPRECATED ROUTE
+def post_worksheets_command():
+    # TODO(klopyrev): The Content-Type header is not set correctly in
+    # editable_field.jsx, so we can't use request.json.
+    data = json.loads(request.body.read())
+
+    if data.get('raw_command', None):
```
percyliang: `raw_command in data`?

codalab/rest/cli.py
```diff
@@ -0,0 +1,164 @@
...
+        data['command'] = worksheet_util.get_worksheet_info_edit_command(data['raw_command'])
```
percyliang: This logic is kind of strange: why is raw_command only for edits?  The name wouldn't suggest that...

docs/rest.md
```diff
@@ -10,15 +10,18 @@ subject to change at any time. Feedback through our GitHub issues is appreciated
   - [Bundle Permissions API](#bundle-permissions-api)
   - [Oauth2 API](#oauth2-api)
   - [Users API](#users-api)
-  - [Worksheet Items API](#worksheet-items-api)
+  - [Chatbox API](#chatbox-api)
+  - [Help API](#help-api)
   - [Workers API](#workers-api)
   - [Bundles API](#bundles-api)
+  - [Interpret API](#interpret-api)
```
percyliang: Can we call this `worksheet-interpretation-api` or something (to say it's for worksheets)?

codalab/lib/worksheet_util.py
```diff
@@ -379,100 +375,6 @@ def friendly_render_dep(dep):
     return None
 
 
-def interpret_file_genpaths(client, requests):
```
percyliang, line 382: Do we have a way to think about which stuff goes in `worksheet_util.py` and what goes in `interpet,py`?  Maybe would be good to document this.

codalab/rest/cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+    The CLI uses JsonApiClient to communicate back to the REST API.
+    This is admittedly not ideal since now the REST API is essentially
+    making HTTP requests back to itself. Future potential solutions might
+    include creating a subclass of JsonApiClient that can reroute HTTP
+    requests directly to the appropriate Bottle view functions.
+    """
+    output_buffer = StringIO()
+    rest_client = JsonApiClient(rest_url(), get_user_token)
+    manager = CodaLabManager(
```
percyliang, line 108: This always makes me really nervous.  Can we double check that we didn't do anything that opens up new security holes?

percyliang (commented):
Nice!

codalab/lib/worksheet_util.py
kashizui: Good idea, I'll document it. The distinction is that worksheet_util doesn't make any calls to the model, they're kind of just like static helper functions.

codalab/rest/cli.py
kashizui: This needs to be merged first, then I'll add a PR on worksheets to use the new route, then I can create a new PR to delete all the `DEPRECATED ROUTE`s. I'll add a comment somewhere or create an issue.

codalab/rest/cli.py
kashizui: I actually don't know where this is used, I'll look into it and make renames/deletions when they make sense.

codalab/rest/cli.py
kashizui: Sure thing I'll take a look.
@fabeschan would be great to get a second set of eyes on this as well :)

docs/rest.md
kashizui: Just in the docs or in the API routes as well?

percyliang: Definitely in the docs.  API routes - I could go either way.  Probably fine to leave to keep it short.  (Alternatively, we could do winterpret.)

codalab/rest/bundles.py
```diff
@@ -343,22 +344,36 @@ def _fetch_bundle_contents_info(uuid, path=''):
           "link": "<string representing target if file is a symbolic link>",
           "type": "<file|directory|link>",
           "size": <size of file in bytes>,
```
fabeschan, line 355: I wonder if you also want to update this part to reflect the new parameter

codalab/rest/cli.py
fabeschan: okay

kashizui:
Since I removed all uses of legacy or deprecated routes from codalab/codalab-worksheets#309, I was able to completely remove the legacy module now.

Also made some modifications to the API necessary for the frontend migration and improved bundle metadata validation.

Ready for final reviews @percyliang @fabeschan @maxwang7

codalab/client/json_api_client.py
```diff
@@ -603,17 +603,49 @@ def upload_contents_blob(self, bundle_id, fileobj=None, params=None,
...
                 progress_callback=progress_callback)
 
-    @wrap_exception('Unable to make RPC call \'{1}\'')
-    def rpc(self, method, *args, **kwargs):
-        request = {
-            'method': method,
-            'args': args,
-            'kwargs': kwargs,
-        }
+    @wrap_exception('Unable to interpret file genpaths')
```
fabeschan, line 606: Perhaps dumb question: what's a genpath?

fabeschan: (got a thorough explanation in person)

fabeschan (approved):
LGTM great work

codalab/rest/cli.py
fabeschan: I think the way this is done makes sense? Unless I missed something

codalab/lib/worksheet_util.py
kashizui: done

codalab/rest/bundles.py
kashizui: parameter got removed

PR #748, opened 2017-05-03
Support human readable fields in contents API

kashizui:
Necessary to support UI file browser while maintaining a unified
formatting policy.

Paired with codalab/codalab-worksheets#308

please review

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -363,6 +363,8 @@ def _fetch_bundle_contents_info(uuid, path=''):
 
     check_bundles_have_read_permission(local.model, request.user, [uuid])
     info = local.download_manager.get_target_info(uuid, path, depth)
+    if info is None:
+        abort(httplib.NOT_FOUND, 'Bundle not found')
 
     def render_human_readable(target_info):
         if 'size' in target_info:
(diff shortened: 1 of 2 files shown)


codalab/rest/bundles.py
```diff
@@ -343,22 +345,36 @@ def _fetch_bundle_contents_info(uuid, path=''):
...
+    human_readable = query_get_bool('human_readable', default=False)
     if depth < 0:
         abort(httplib.BAD_REQUEST, "Depth must be at least 0")
 
     check_bundles_have_read_permission(local.model, request.user, [uuid])
+    info = local.download_manager.get_target_info(uuid, path, depth)
+
+    def render_human_readable(target_info):
+        if 'size' in target_info:
+            target_info['size'] = formatting.size_str(target_info['size'])
```
percyliang: Not sure I like multiplexing between these two values...can we always just return two fields, `size` and `size_str` and not even have the flag?

percyliang (commented):
Also, what is the standard way to handle these settings?  Seems like options are to do formatting on server side or client side (which results in duplicate code and inconsistencies (unless we use node)...

codalab/rest/bundles.py
kashizui: Hm yeah we could do that directly in DownloadManager.

kashizui:
I'm not sure what is standard...

kashizui:
Couldn't do it in download_util where target_info is generated since it doesn't have access to the formatting.

fabeschan (approved):
lgtm

PR #749, opened 2017-05-07
Mount bundles as a filesystem

fabeschan:
Mount bundles locally as a read-only filesystem

If the bundle is a single file, the mountpoint will look like a directory that contains that single file.

- [x] Works on public bundles

- [x] Works on private bundles (with correct account permissions)

- [x] Works on running bundles located on a public worker

- [x] Works on running bundles located on a private worker (with correct account permissions)


codalab/lib/bundle_cli.py
```diff
@@ -1730,6 +1732,29 @@ def wrap(string):
                 self.print_target_info(client, (bundle_uuid, item['name']), decorate=True)
 
     @Commands.command(
+        'mount',
+        help=[
+            'Mount the contents of a bundle at a read-only mountpoint.',
+        ],
+        arguments=(
+            Commands.Argument('target_spec', help=TARGET_SPEC_FORMAT, completer=TargetsCompleter),
+            Commands.Argument('--mountpoint', help='Empty directory path to set up as the mountpoint for FUSE.'),
```
percyliang, line 1745: Can we just make this an optional argument so you can do:

    cl mount 0x4252 data
    cl mount data

codalab/lib/bundle_cli.py
```diff
@@ -1730,6 +1732,29 @@ def wrap(string):
...
+        self._fail_if_headless(args)  # Disable on headless systems
+
+        client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
+        target = self.parse_target(client, worksheet_uuid, args.target_spec)
+        uuid, path = target
+
+        mountpoint = path_util.normalize(args.mountpoint)
+        path_util.check_isvalid(mountpoint, 'mount')
+        print >>self.stdout, 'BundleFUSE will run and maintain the mounted filesystem in the foreground. CTRL-C to cancel.'
+        bundle_fuse.bundle_mount(client, mountpoint, target)
```
percyliang: When we press ctrl+c, can we print out a message saying shutting down?

codalab/lib/bundle_cli.py
```diff
@@ -1730,6 +1732,29 @@ def wrap(string):
...
+    def do_mount_command(self, args):
+        self._fail_if_headless(args)  # Disable on headless systems
+
+        client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
+        target = self.parse_target(client, worksheet_uuid, args.target_spec)
+        uuid, path = target
+
+        mountpoint = path_util.normalize(args.mountpoint)
+        path_util.check_isvalid(mountpoint, 'mount')
+        print >>self.stdout, 'BundleFUSE will run and maintain the mounted filesystem in the foreground. CTRL-C to cancel.'
```
percyliang: print out more details: Mounting `<bundle name>(<uuid>) on <mountpoint>`

codalab/lib/bundle_fuse.py
```diff
@@ -0,0 +1,162 @@
...
+            mode = stat.S_IFREG | info['perm']
+            nlink = 1
+        elif info['type'] == 'directory':
+            mode = stat.S_IFDIR | info['perm']
+            nlink = 2
+        elif info['type'] == 'link':
+            mode = stat.S_IFLNK | info['perm']
+            nlink = 1
+
+        return { # time attributes are set to 0 for lack of a better idea
```
percyliang: Can we set this to the created time of the bundle?

codalab/lib/bundle_fuse.py
```diff
@@ -0,0 +1,162 @@
...
+    def yank_path(self, path):
+        """Clear cache of results from a specific path"""
+        for func in self._caches:
+            cache = {}
+            for key in self._caches[func].keys():
+                if path in key[0]:
+                    del self._caches[func][key]
+
+class BundleFuse(Operations):
+    def __init__(self, client, target):
```
percyliang: Class-level comment about what this class does.

codalab/lib/fuse.py
```diff
@@ -0,0 +1,978 @@
+# Copyright (c) 2012 Terence Honles <terence@honles.com> (maintainer)
```
percyliang: Can we acknowledge https://github.com/terencehonles/fusepy more directly or depend on it rather than just copying it?  If it's not too much of a pain.

percyliang (changes requested):
This is great!  Two remarks:
- Do you have a sense of the performance characteristics?  If you download a 100MB file using mount+cp versus download, what's the difference?
- We should add documentation to the Wiki on how to use `cl mount`.

codalab/lib/bundle_cli.py
fabeschan: Not sure what you mean there, if one of the arguments isn't provided.. 
What would your 2nd example `cl mount data` do?

codalab/lib/bundle_cli.py
fabeschan: Will do!

codalab/lib/bundle_cli.py
fabeschan: Will do!

codalab/lib/bundle_fuse.py
fabeschan: Will do!

codalab/lib/bundle_fuse.py
fabeschan: Yup, will look into it

codalab/lib/fuse.py
fabeschan: Hmm.. I copied the file simply because the readme on the github page seems to imply I should just download the file and include it there. Let me see if there's a package published somewhere

kashizui: You can just add the following line to `requirements.txt`
```
git+https://github.com/terencehonles/fusepy.git#v2.0.4
```

codalab/lib/bundle_cli.py
kashizui: `cl mount data` should get the UUID of the last bundle on the worksheet with the name "data" and mount it as a file/directory with the same name.

codalab/lib/bundle_fuse.py
```diff
@@ -0,0 +1,162 @@
...
+        self.bundle_uuid = target[0]
+        self.fd = 0 # file descriptor
+
+    # Helpers
+    # =======
+
+    @Memoize(timeout=5)
+    def _get_info(self, path):
+        ''' Set a request through the json api client to get info about the bundle '''
+        info = self.client.fetch_contents_info(self.bundle_uuid, path, 1)
```
kashizui, line 97: When a file doesn't exist, `info` will be `None`.

(In the future, a `NotFoundError` will be thrown instead, but I can handle that change.)

codalab/lib/bundle_fuse.py
```diff
@@ -0,0 +1,162 @@
...
+    # Filesystem methods
+    # ==================
+
+    def getattr(self, path, fh=None):
+        '''
+        Fetch standard filesystem attributes
+        (fabianc: Or just outright make some of them up if they don't really matter)
+        '''
+
+        info = self._get_info(path)
```
fabeschan: TODO: Return ENOENT on file not found

codalab/lib/fuse.py
fabeschan: 👍

codalab/lib/bundle_cli.py
fabeschan: I am hesitant to create the mountpoint for the user, because the mountpoint itself will need to have escalated permissions as a FUSE call is a system call. It will be risky to try to hide these intricacies from the user even if we are fine with adding the necessary bloat to the code to add a bunch of checks

codalab/lib/bundle_fuse.py
fabeschan: done

codalab/lib/bundle_fuse.py
fabeschan: done

codalab/lib/bundle_cli.py
kashizui: I'm a bit confused what the concern is here.. I thought the idea was just that
```
cl mount data
```
would just be equivalent to
```
cl mount data data
```

percyliang: That's what I had in mind.  I'm also fine with requiring the explicit mountpoint (e.g., `cl mount data data`).

percyliang (commented):
I tried it out...sometimes it seems like it's really slow.  I'm hooked up to worksheets.codalab.org and tried this:
cl mount 0xe0d5f6 --mountpoint foo
Even an ls takes something like 30 seconds.  Any ideas on what's going on?  Perhaps we can also have a verbose mode that allows us to see what calls are API calls are being made.

fabeschan:
`cl mount 0xe0d5f6 --mountpoint foo`
`ls -lah` was almost instant for me. I wonder if the libraries we are using are different in someway?

fabeschan:
`cp foo/mnist2500_X.txt .` took 109 seconds, making roughly 240 calls to `BundleFuse.read`

percyliang:
cl down 0xe0d5f6 takes 3.5 seconds, so I wonder what the gap is...buffering?

fabeschan:
`cl down` makes one request to the REST server, while `cp` from the mountpoint makes many hundreds of request, each with non-trivial overhead.

percyliang:
Right...was wondering if there is a way to buffer it so that when you're
doing things like `cp`, we also make few requests.  Also, is the overhead
of making a REST call in establishing the connection?  Could we keep the
connection alive?

Anyway, I think we should probably just get this PR checked in since it
works and then add another issue for improving performance.

fabeschan:
Yeah most of the overhead is in setting up and tearing down each connection. Not sure how to do buffering or keep-alive off my head atm. We'll see.

fabeschan:
> Anyway, I think we should probably just get this PR checked in since it
> works and then add another issue for improving performance.

yup agreed

PR #752, opened 2017-05-16
Support unicode on worksheets

kashizui:
Accidentally broke support for Unicode in worksheets when migrating away from legacy code.

Here's a fix along with an additional CLI test to go with it so it doesn't happen again.


codalab/model/bundle_model.py
```diff
@@ -97,6 +97,12 @@ def __init__(self, engine, default_user_info):
         self.public_group_uuid = ''
         self.create_tables()
 
+    def encode_str(self, value):
+        raise NotImplementedError
+
+    def decode_str(self, value):
+        raise NotImplementedError
+
```
kashizui, line 109: To make it clear that the concrete classes (i.e. MySQLModel) needs to implement these methods.

kashizui:
nevermind this broke some tests, going to have to think of a better way to do this

kashizui:
Finally passed tests, @percyliang please review!

codalab/lib/formatting.py
```diff
@@ -22,6 +22,8 @@ def contents_str(input_string):
     if input_string is None:
         return ''
 
+    if isinstance(input_string, unicode):
+        return input_string
     try:
         input_string.decode('utf-8')
```
percyliang: Is this necessary anymore?  When will we get non-unicode strings?

codalab/model/tables.py
```diff
@@ -78,7 +79,7 @@
   Column('uuid', String(63), nullable=False),
   Column('name', String(255), nullable=False),
   Column('owner_id', String(255), nullable=True),
-  Column('title', String(255), nullable=True),  # Short human-readable description of the worksheet
+  Column('title', Unicode(255), nullable=True),  # Short human-readable description of the worksheet
```
percyliang: Just out of curiosity, is `Text` unicode by default (what the metadata field values and worksheet items are)?

worker/codalabworker/rest_client.py
```diff
@@ -51,8 +51,13 @@ def _make_request(self, method, path, query_params=None, headers=None,
...
         if query_params is not None:
             path = path + '?' + urllib.urlencode(query_params)
-        request = urllib2.Request(self._base_url + path, data=data,
-                                  headers=headers)
+        # Everything needs to be utf-8 encoded or urllib2 will complain
+        if 'Content-Type' in headers:
+            headers['Content-Type'] += '; charset=utf-8'
+        if data and isinstance(data, unicode):
+            data = data.encode('utf-8')
+        request_url = (self._base_url + path).encode('utf-8')
```
percyliang, line 60: Even the url needs to be utf8 encoded?

codalab/model/mysql_model.py
```diff
@@ -44,7 +44,7 @@ class MySQLModel(BundleModel):
     def __init__(self, engine_url, default_user_info):
         if not engine_url.startswith('mysql://'):
             raise UsageError('Engine URL should start with mysql://')
-        engine = create_engine(engine_url, strategy='threadlocal', pool_size=20, max_overflow=100, pool_recycle=3600)
+        engine = create_engine(engine_url, strategy='threadlocal', pool_size=20, max_overflow=100, pool_recycle=3600, encoding='utf-8')
```
percyliang, line 49: Does this mean that we don't need to call `encode_str` and `decode_str` and can pass unicode strings directly into DB operations?

percyliang (commented):
Cool, just some questions.

Can you summarize what the status of unicode support is?  What still needs to be done to have completely proper unicode handling (decode/encode at boundaries and all unicode inside)?  Document this in the code for posterity (I don't know where is appropriate).

worker/codalabworker/rest_client.py
kashizui: Yes for some reason, mismatching types between the url and the request body results in Unicode decode errors..

codalab/model/mysql_model.py
kashizui: Only if `String(N, convert_unicode=True)` or `Unicode(N)` types are used.

codalab/model/tables.py
kashizui: `Text` is not unicode by default, `UnicodeText` is. Which is why I think there is some manual conversion of types. I would try to change more of the columns to `Unicode*` but I'm not prepared to test all those changes..

codalab/lib/formatting.py
kashizui: This does get used still in some places to mask strings that contain binary (non-unicode) content, stored in plain strings.

codalab/model/tables.py
percyliang: Ok, how about add a TODO for changing things to `UnicodeText` if that's the right thing to do?

codalab/lib/formatting.py
percyliang: Can we add a TODO to remove once everything is properly handled with unicode?

worker/codalabworker/rest_client.py
percyliang: If we believe that we eventually don't need this, make a comment to that effect.

percyliang (approved):

codalab/lib/formatting.py
kashizui: Actually I believe this is correct, since arbitrary streams of bytes should be stored as strings in Python2 (there is a new `bytes` type in Python3 to disambiguate). `unicode` type is only for Unicode data.

I can clean up the logic here and add comments to clarify.

PR #753, opened 2017-05-16
Follow stdout/stderr from the start when tailing runs

kashizui:
When using `cl run` or `cl mimic`, we want to to follow stdout and
stderr from the very beginning. Previously, using `cl run --tail` would
miss the first portion of the output if a lot of output was generated
very quickly.


percyliang (approved):
Ah, that's what was going on!

fabeschan (commented):
lgtm

PR #756, opened 2017-05-18
Resumable Workers

fabeschan:


fabeschan:
Yeah. there's probably a security vulnerability where a worker can spoof a dead one.

Perhaps validate `remote` field and make sure it matches? Although that can be spoofed too

fabeschan:
Given the fact that a `resume_bundle` request still checks whether the particular worker is authorized to handle the bundle's owner's bundles, the possibility of malicious spoofing is limited to workers of the same owner. So I don't think this is a big issue

kashizui:
Yeah, owners of workers have arbitrary control for what they do to actually execute the bundle and return the contents, so there's no point in adding extra security for that.

As long as no one can spoof the dead worker of another user (or more precisely, spoof the worker for a bundle that they have no write access to) we should already meet our guarantees

fabeschan:
@kashizui @percyliang review pls

codalab/lib/download_manager.py
```diff
@@ -77,8 +80,10 @@ def get_target_info(self, uuid, path, depth):
                 self._send_read_message(worker, response_socket_id, uuid, path, read_args)
                 with closing(self._worker_model.start_listening(response_socket_id)) as sock:
                     result = self._worker_model.get_json_message(sock, 60)
-                precondition(result is not None, 'Unable to reach worker')
-                if 'error_code' in result:
+                if result is None: # dead workers are a fact of life now
```
kashizui, line 83: 😍

codalab/rest/workers.py
```diff
@@ -120,6 +120,24 @@ def start_bundle(worker_id, uuid):
...
+    #given ID. If so, reports that it's starting to run and returns True.
+    Otherwise, returns False, meaning the worker shouldn't run the bundle.
+    """
+    bundle = local.model.get_bundle(uuid)
+    check_run_permission(bundle)
+    response.content_type = 'application/json'
+    if local.model.resume_bundle(bundle, request.user.user_id, worker_id,
+                                request.json['hostname'],
+                                request.json['start_time']):
+        print 'Resume bundle %s' % uuid
```
kashizui, line 137: clean up print statements

codalab/common.py
```diff
@@ -101,10 +101,12 @@ class State(object):
     RUNNING = 'running'   # Actually running
     READY = 'ready'       # Done running and succeeded
     FAILED = 'failed'     # Done running and failed
+    KILLED = 'killed'     # Killed by user
```
kashizui, line 104: What's the rationale for splitting a KILLED state out from FAILED?

worker/run.py
```diff
@@ -88,6 +109,27 @@ def run(self):
...
+        # Report that the bundle is running. We note the start time here for
+        # accurate accounting of time used, since the clock on the bundle
+        # service and on the worker could be different.
+        start_message = {
+            'hostname': socket.gethostname(),
+            'start_time': int(self._start_time),
+        }
+        if not self._bundle_service.resume_bundle(self._worker.id, self._uuid,
+                                                 start_message):
+            print("false!!")
```
kashizui, line 122: ?

worker/run.py
```diff
@@ -88,6 +109,27 @@ def run(self):
...
+            return False
+
+        # Start a thread for this run.
+        def resume_run(self):
+            Run._safe_update_run_status(self, 'Running')
+            Run._monitor(self)
+
+        print("resuming bundle {}, container {}".format(self._uuid, self._container_id))
```
kashizui, line 130: try to use `logger.info` or `logger.debug` as appropriate instead of print statements

codalab/common.py
```diff
@@ -101,10 +101,12 @@ class State(object):
...
+    WORKER_OFFLINE = 'worker_offline'
```
kashizui, line 105: Maybe we should have a state diagram showing the possible transitions between all these states..

kashizui (changes requested):
Generally looks good to me! Excited for robust workers.

Probably a big ask but is there any way to test all of this?

fabeschan:
@kashizui It's a perfectly reasonable ask to have tests covering this... I've been thinking about how to test this properly too, just don't have a good idea on how to go about doing it yet

Suggestions/brainstorm welcome!

kashizui:
Small tests:

- simple units tests for saving and loading runs

Medium tests:

- ?

kashizui:
it's tempting to try to come up with big integration tests to test how we think the worker should behave but those tests are so hard to maintain..

codalab/model/bundle_model.py
```diff
@@ -670,14 +687,55 @@ def start_bundle(self, bundle, user_id, worker_id, hostname, start_time):
 
         return True
 
+    def resume_bundle(self, bundle, user_id, worker_id, hostname, start_time):
+        '''
+        Marks the bundle as running but only if it is still scheduled to run
+        #on the given worker (done by checking the worker_run table). Returns
```
percyliang, line 693: extra `#`

codalab/model/worker_model.py
```diff
@@ -325,7 +325,7 @@ def send_json_message(self, socket_id, message, timeout_secs, autoretry=True):
 
                 if not success:
                     # Shouldn't be too expensive just to keep retrying.
-                    time.sleep(0.003)
+                    time.sleep(0.3) # changed from 0.003 to keep from rate-limiting due to dead workers
```
percyliang, line 328: If we actually expect this to fail, shouldn't we have exponential backoff?

percyliang (commented):
This is great!

Have you tested this code with shared_file_system?

codalab/common.py
fabeschan: To me, it makes sense to differentiate between a user aborted run and a run that failed and dies on its own. Too many times I've panicked before realizing a run shown as 'failed' was actually something I killed a while ago.

Perhaps the bigger reason is any signal that can give clues as to whether a failure is the user's fault or potentially our platform's fault is valuable in a practical sense.

codalab/model/worker_model.py
fabeschan: In principle, yes, expo backoff makes sense. Perhaps it will be more appropriate to switch all requests of this type to retry with expo back off in a separate PR tho.

fabeschan:
@percyliang I have not tried with shared_file_system. How does one usually set up a shared file system and test it?

codalab/model/worker_model.py
percyliang: Let's put a comment for now then.

codalab/common.py
percyliang: I think having `killed` is fine, but the thing is that I think we've implicitly assumed that we had a closed set of states, and some parts of the code assume that (also the public stats worksheet).  So we should check those places, and generally have a plan for thinking about states - are we going to keep on increasing the number of states, or is this it?

percyliang:
The NLP machines use `shared_file_system`.  It's just a matter of putting a flag in the `config.json` file and passing the `--shared-file-system` flag to workers.

kashizui:
Yeah for local testing just run both server and workers with the shared file system flag on the same machine, and it should just work!

fabeschan:
Add note: will be unblocked once a fix for #778 is approved and merged

fabeschan:
Closed for reboot: https://github.com/codalab/codalab-cli/pull/805

PR #762, opened 2017-05-25
Speed up BundleFuse

### LINKED ISSUE #757: Make cl mount faster
Currently the REST calls dominate the processing times for every read request. Try to find a way to buffer.

fabeschan:
Prefetch byte ranges.

Fixes #757

### DIFF
--- a/codalab/lib/bundle_fuse.py
+++ b/codalab/lib/bundle_fuse.py
@@ -7,6 +7,7 @@ import sys
 import errno
 import stat
 import time
+from collections import OrderedDict
 
 from contextlib import closing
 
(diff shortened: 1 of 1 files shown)


fabeschan:
I'll do a clean up once I get an approval for the overall approach

codalab/lib/bundle_fuse.py
```diff
@@ -64,6 +65,65 @@ def yank_path(self, path):
                 if path in key[0]:
                     del self._caches[func][key]
 
+class ByteRangeReader(object):
+
```
percyliang: Add high-level description.

codalab/lib/bundle_fuse.py
```diff
@@ -64,6 +65,65 @@ def yank_path(self, path):
...
+                print 'expired'
+                self.cache.pop(key)
+
+        # grab from client
+        byte_range = (chunk_id * self.chunk_size, chunk_id * self.chunk_size + self.chunk_size - 1)
+        with closing(self.client.fetch_contents_blob(self.bundle_uuid, path, byte_range)) as contents:
+            arr = contents.read()
+
+        if len(arr) == self.chunk_size: # only cache if fetched full chunk
+            print 'put to cache'
```
percyliang: Probably don't want these print statements, or guard them by some verbose flag.

percyliang (approved):
Looks generally reasonable.  In general, more documentation, especially about the performance characteristics about how the code is exercised in practice, would be nice.

Question: if you do a cp of a 30 MB file, what is `length` usually?  If this is larger than 1 MB, then we could have one call that grabs multiple chunks.  It seems like the REST API is still a bottleneck given the difference between using FUSE and `cl down`.  Probably worth quantifying this.

fabeschan:
Thanks @percyliang , I'll do a cleanup and add documentation. 

`length` is usually around ~13000 bytes

codalab/lib/bundle_fuse.py
```diff
@@ -64,6 +65,65 @@ def yank_path(self, path):
...
+        arr = ''
+        for chunk_id in range(start_chunk, end_chunk + 1):
+            arr += self._fetch_chunk(path, chunk_id)
+        return arr[offset - start_chunk * self.chunk_size:offset - start_chunk * self.chunk_size + length]
+
+    def _fetch_chunk(self, path, chunk_id):
+        '''
+        Fetch a chunk from the cache, or with the client as necessary.
+        Refreshes chunks that are older than self.timeout
+        Only save full chunks to the cache. Partial chunks are not cached.
```
percyliang: Say that partial chunks are only at the end of the file because they could grow...

Make the assumptions clear.

fabeschan: done

codalab/lib/bundle_fuse.py
fabeschan: done

maxwang7:
LGTM! I didn't do any perf testing, but functionally it works for me on Ubuntu 14.04.

PR #773, opened 2017-06-02
CPU-only run bundles prioritize workers without GPUs

maxwang7:
Tested on dev with a GPU worker. Ran `cl run date`, which ran on a CPU-only worker. Then ran `cl run date --request-gpus 1`, which ran on a worker with GPUs.

Code for excluding workers by tag was not included in this PR because of ongoing discussions.


fabeschan (approved):
lgtm

PR #779, opened 2017-06-09
Rename 'cl edit-image' to 'cl docker', remove --local from 'cl run'

### LINKED ISSUE #765: Local dev run mounts local dependencies and not uploaded bundles
This is confusing to users. An idea to make this more clear is to change the name of local-run to something else like `cl docker`.

fabeschan:
Fixes #765 and #776

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1276,12 +1276,13 @@ class BundleCLI(object):
             os.symlink(docker_dependency_path, child_path)
 
         dc = DockerClient()
-        container_id = dc.start_container(bundle_path, uuid, args.command, docker_image, request_network, dependencies)
+        container_id = dc.create_container(bundle_path, uuid, args.command, docker_image, request_network, dependencies, ['-it'])
         print >>self.stdout, '===='
-        print >>self.stdout, 'ContainerID: ', container_id
-        print >>self.stdout, 'Local Bundle ID: ', uuid
+        print >>self.stdout, 'Container ID: ', container_id[:12]
+        print >>self.stdout, 'Local Bundle UUID: ', uuid
         print >>self.stdout, 'You can find local bundle contents in: ', bundle_path
         print >>self.stdout, '===='
+        os.system('docker start -ai {}'.format(container_id))
 
     @Commands.command(
         'edit',
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -104,6 +104,7 @@ BUNDLE_COMMANDS = (
     'upload',
     'make',
     'run',
+    'docker',
     'edit',
     'detach',
     'rm',
@@ -118,12 +119,6 @@ BUNDLE_COMMANDS = (
     'kill',
     'write',
     'mount',
-)
-
-DOCKER_IMAGE_COMMANDS = (
-    'edit-image',
-    'commit-image',
-    'push-image',
 )
 
 WORKSHEET_COMMANDS = (
@@ -298,7 +293,6 @@ class Commands(object):
         max_length = max(
           len(command_name(command)) for command in itertools.chain(
               BUNDLE_COMMANDS,
-              DOCKER_IMAGE_COMMANDS,
               WORKSHEET_COMMANDS,
               GROUP_AND_PERMISSION_COMMANDS,
               USER_COMMANDS,
@@ -356,9 +350,6 @@ class Commands(object):
         Commands for users:
         {user_commands}
 
-        Commands for building Docker images:
-        {docker_image_commands}
-
         Commands for managing server:
         {server_commands}
 
@@ -366,7 +357,6 @@ class Commands(object):
         {other_commands}
         """).format(
             bundle_commands=command_group_help_text(BUNDLE_COMMANDS),
-            docker_image_commands=command_group_help_text(DOCKER_IMAGE_COMMANDS),
             worksheet_commands=command_group_help_text(WORKSHEET_COMMANDS),
             group_and_permission_commands=command_group_help_text(GROUP_AND_PERMISSION_COMMANDS),
             user_commands=command_group_help_text(USER_COMMANDS),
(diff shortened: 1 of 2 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -123,7 +124,6 @@
 )
 
 DOCKER_IMAGE_COMMANDS = (
-    'edit-image',
     'commit-image',
```
percyliang: Should remove these too, right?

codalab/lib/bundle_cli.py
```diff
@@ -1297,7 +1266,7 @@ def do_edit_image_command(self, args):
 
         print >>self.stdout, '===='
         print >>self.stdout, 'Entering container {}'.format(container_id[:8])
-        print >>self.stdout, 'Once you are happy with the changes, please exit the container (ctrl-D)'
+        print >>self.stdout, 'If you would like to commit the changes, please exit the container (ctrl-D)'
         print >>self.stdout, 'and commit your changes to a new image by running:'
         print >>self.stdout, ''
         print >>self.stdout, '\tcl commit-image {} [image-tag]'.format(container_id[:8])
```
percyliang: Discourage this behavior.  Instead ask them to create a Dockerfile (point to the Wiki).

codalab/lib/bundle_cli.py
```diff
@@ -1306,7 +1275,7 @@ def do_edit_image_command(self, args):
         os.system('docker start -ai {}'.format(container_id))
         print >>self.stdout, '===='
         print >>self.stdout, 'Exited from container {}'.format(container_id[:8])
-        print >>self.stdout, 'If you are happy with the changes, please commit your changes to a new'
+        print >>self.stdout, 'If you would like to commit the changes, please commit your changes to a new'
         print >>self.stdout, 'image by running:'
         print >>self.stdout, ''
         print >>self.stdout, '\tcl commit-image {} [image-tag]'.format(container_id[:8])
```
percyliang: Can you refactor to not have this duplicated?

codalab/lib/bundle_cli.py
percyliang: Actually, I think it would make sense to be able to specify a command here (move the logic from `run --local` basically).

The flow would be something like:
- Create Dockerfile
- docker build ...
- repeat until happy:
 * cl docker ... 'python main.py ...'
 * cl docker ... (inside it, run 'pip install')
- Edit Dockerfile to include all the pip install commands
- docker build ...
- docker push ...
- cl run --request-docker-image ...

If we wanted to be slicker, we could have `cl docker` take a `Dockerfile` directly, and have an option for appending commands to the `Dockerfile`.  Thoughts?

fabeschan:
@percyliang addressed PR comments

fabeschan:
ready for another review

codalab/lib/bundle_cli.py
```diff
@@ -1293,56 +1258,13 @@ def do_edit_image_command(self, args):
...
-    def do_push_image_command(self, args):
-        self._fail_if_headless(args)  # Disable on headless systems
+        container_id = dc.start_container(bundle_path, uuid, args.command, docker_image, request_network, dependencies)
         print >>self.stdout, '===='
-        print >>self.stdout, 'cl push-image has been deprecated and disabled. Please use docker push instead:'
-        print >>self.stdout, ''
-        print >>self.stdout, '\tdocker push [image-tag]'
-        print >>self.stdout, ''
+        print >>self.stdout, 'ContainerID: ', container_id
+        print >>self.stdout, 'Local Bundle ID: ', uuid
```
percyliang: ID ==> UUID

codalab/lib/bundle_cli.py
```diff
@@ -1293,56 +1258,13 @@ def do_edit_image_command(self, args):
...
-    )
-    def do_push_image_command(self, args):
-        self._fail_if_headless(args)  # Disable on headless systems
+        container_id = dc.start_container(bundle_path, uuid, args.command, docker_image, request_network, dependencies)
         print >>self.stdout, '===='
-        print >>self.stdout, 'cl push-image has been deprecated and disabled. Please use docker push instead:'
-        print >>self.stdout, ''
-        print >>self.stdout, '\tdocker push [image-tag]'
-        print >>self.stdout, ''
+        print >>self.stdout, 'ContainerID: ', container_id
```
percyliang: add space before ID

percyliang (changes requested):
This command should behave more like `docker`, which means that:
- stdout/stderr should go to the console by default
- should be able to specify 'bash' to drop into a shell
This shouldn't create a `local_bundles` directory.

The use case I imagine is that people just run `cl docker` multiple times interleaved between editing code, so this should feel very much like running a command.

Also, we need to print out instructions on what to do with missing packages and mention `Dockerfile`.

fabeschan:
Ya that's good idea

fabeschan:
> This shouldn't create a local_bundles directory.

For people who want to debug runs that create other files, this might be very useful. I suggest we keep this directory.

> Also, we need to print out instructions on what to do with missing packages and mention Dockerfile.

This shouldn't belong to the cli help printout, but rather on external documentation like the wiki/FAQ/blogposts

fabeschan:
@percyliang thoughts?

codalab/lib/bundle_cli.py
```diff
@@ -1293,55 +1258,13 @@ def do_edit_image_command(self, args):
...
-
+        container_id = dc.create_container(bundle_path, uuid, args.command, docker_image, request_network, dependencies, ['-it'])
         print >>self.stdout, '===='
-        print >>self.stdout, 'Entering container {}'.format(container_id[:8])
-        print >>self.stdout, 'Once you are happy with the changes, please exit the container (ctrl-D)'
-        print >>self.stdout, 'and commit your changes to a new image by running:'
-        print >>self.stdout, ''
-        print >>self.stdout, '\tcl commit-image {} [image-tag]'.format(container_id[:8])
-        print >>self.stdout, ''
+        print >>self.stdout, 'Container ID: ', container_id[:12]
```
percyliang, line 1281: extra space after ":", same for next lines

percyliang (commented):
I tried `cl docker --request-docker-image codalab/ubuntu:1.9 date` and it put me into a shell rather than running `date`...

fabeschan:
@percyliang `cl docker --request-docker-image codalab/ubuntu:1.9 date` did not drop me into a shell though..

PR #780, opened 2017-06-09
server uses local codalabworker code rather than pypi version in setup

maxwang7:
server uses local codalabworker code rather than pypi version in setup.py; workers will pull that code on check in

### DIFF
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -8,7 +8,7 @@ import httplib
 
 # Increment this on the develop branch when develop is merged into master.
 # http://semver.org/
-CODALAB_VERSION = '0.2.13'
+CODALAB_VERSION = '0.2.12'
 
 class IntegrityError(ValueError):
     """
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -8,7 +8,7 @@ import httplib
 
 # Increment this on the develop branch when develop is merged into master.
 # http://semver.org/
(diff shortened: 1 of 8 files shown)


percyliang (commented):
I think we should either pip install both codalab and codalabworker into the virtualenv (which I prefer) or neither.

fabeschan:
How would someone use this? Do I go into the repo and run something like `python setup.py server` or `python setup.py dev`?

percyliang:
If you're developing, you git clone and go into the repo and type
`./setup.sh server`.  Why do we have both dev and server?  Can we just
merge them?

At this point, `setup.sh` isn't doing that much, but only there to be a
friendly wrapper around the pypi setup, so it can tell you that virtualenv
is missing, etc.

maxwang7:
The reason I believe this change works (though tests are failing, will need to investigate):
- The server setup code checks out the code from Github (it doesn't do pip install right now)
- Then we pip install codalabworker locally
- When the server is started, it uses `codalab-cli/codalab/bin/cl`.

I could also change the code to do `$env/bin/pip install -e .`, which would install the `codalab` module locally, and then the Docker worker would start the server using `cl` rather than `codalab-cli/codalab/bin/cl`.

Investigating the CircleCI failures now...

maxwang7:
Cool, tests passed, will do a test deployment now.

PR #785, opened 2017-06-25
Last few changes

kashizui:
See commit messages.

@percyliang please review!

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -384,6 +384,7 @@ class Commands(object):
         """
         parser = CodaLabArgumentParser(prog='cl', cli=cli, add_help=False, formatter_class=argparse.RawTextHelpFormatter)
         parser.register('action', 'parsers', AliasedSubParsersAction)
+        parser.add_argument('-v', '--version', dest='print_version', action='store_true')
         subparsers = parser.add_subparsers(dest='command', metavar='command')
 
         # Build subparser for each subcommand
@@ -730,6 +731,9 @@ class BundleCLI(object):
 
         # Parse arguments
         argv = self.collapse_bare_command(argv)
+        if argv[0] == '-v' or argv[0] == '--version':
+            self.print_version()
+            return
         args = parser.parse_args(argv)
 
         # Bind self (BundleCLI instance) and args to command function
@@ -762,6 +766,9 @@ class BundleCLI(object):
                 self.exit('%s: %s' % (e.__class__.__name__, e))
         return structured_result
 
+    def print_version(self):
+        print >>self.stdout, 'CodaLab CLI version %s' % CODALAB_VERSION
+
     @Commands.command(
         'help',
         help=[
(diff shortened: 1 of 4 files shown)


codalab/bundles/named_bundle.py
```diff
@@ -16,7 +16,7 @@ class NamedBundle(Bundle):
 
     METADATA_SPECS = (
       MetadataSpec('name', basestring, 'Short variable name (not necessarily unique); must conform to %s.' % spec_util.NAME_REGEX.pattern, short_key='n'),
-      MetadataSpec('description', basestring, 'Full description of the bundle.', short_key='d'),
+      MetadataSpec('description', basestring, 'Full description of the bundle.', short_key='m', metavar='MESSAGE'),
```
kashizui: The `-d` short key was conflicting with `-d` for depth in mimic.

codalab/lib/bundle_cli.py
```diff
@@ -670,7 +684,6 @@ def get_missing_metadata(bundle_subclass, args, initial_metadata=None):
     )
 
     MIMIC_ARGUMENTS = (
-        Commands.Argument('-n', '--name', help='Name of the output bundle.'),
```
kashizui, line 673: `metadata_arguments([MakeBundle, RunBundle])` already includes name

codalab/bundles/named_bundle.py
percyliang: Hmm...isn't this going to break backward compatibility?  I'd rather just disable `-d` for depth because I don't think people are likely to use that, and if they are, then `--depth` is fine.

codalab/lib/bundle_cli.py
```diff
@@ -775,17 +788,20 @@ def do_help_command(self, args):
...
     )
     def do_status_command(self, args):
+        client, worksheet_uuid = self.manager.get_current_worksheet_uuid()
+        worksheet_info = client.fetch('worksheets', worksheet_uuid)
+
         if not self.headless:
             print >>self.stdout, "codalab_home: %s" % self.manager.codalab_home
             print >>self.stdout, "session: %s" % self.manager.session_name()
             address = self.manager.session()['address']
+            print >>self.stdout, "client_version: %s" % CODALAB_VERSION
```
percyliang, line 805: Still would prefer to have `cl -v` and `cl --version` work instead of having it here, since that's what people are used to typing.

percyliang (changes requested):
Cool, remember to update the Wiki too (wish there were a way to update it as part of this PR)...

kashizui:
@percyliang Made requested changes, will merge and update wiki once approved

percyliang (approved):
Huh, I can't check out the `hail-mary` branch anymore, but I'll trust that everything works as intended.

kashizui:
I might have force-pushed at some point, so you could try
```
git checkout kashizui/hail-mary
git fetch
git reset --hard origin/kashizui/hail-mary
```

PR #789, opened 2017-07-21
Don't strip trailing spaces in markdown

robinjia:
Trailing spaces are actually useful in markdown, as double spaces indicate line breaks without creating `<p>` elements, which add additional space between lines.  So, we shouldn't strip trailing spaces, only newlines.

Dunno what your policies are w.r.t. pull requests, but I already made the change locally so I figured I'd just send this.


percyliang (approved):
Cool.

PR #794, opened 2017-09-27
If --request-docker-image does not come with a tag, default to "latest"

### LINKED ISSUE #793: --request-docker-image grabs all versions
Migrated from: https://github.com/codalab/codalab-worksheets/issues/351

Current behavior: `--request-docker-image tensorflow/tensorflow` grabs all versions of tensorflow, which fills up disk space.

Desired behavior: grab only the `:latest` branch and make sure that we pull every time in case the latest tag changes.

fabeschan:
Rather than downloading every version/tag of the image
Also update worker version to 13

Fixes #793

### DIFF
--- a/worker/codalabworker/docker_client.py
+++ b/worker/codalabworker/docker_client.py
@@ -253,7 +253,7 @@ nvidia-docker-plugin not available, no GPU support on this worker.
     def download_image(self, docker_image, loop_callback):
         if len(docker_image.split(":")) < 2:
             logger.debug('Missing tag/digest on request docker image "%s", defaulting to latest', docker_image)
-            docker_image = ":".join([docker_image, "latest"])
+            docker_image = ':'.join([docker_image, 'latest'])
 
         logger.debug('Downloading Docker image %s', docker_image)
         with closing(self._create_connection()) as conn:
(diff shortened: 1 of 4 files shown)


worker/codalabworker/docker_client.py
```diff
@@ -248,8 +248,13 @@ def remove_image(self, repo_digest):
                 for action, target in line.items():
                     logger.debug('docker image %s: %s', action.lower(), target)
 
+    ''' Download the specified docker image with tag/digest. If no tag is specified, downloads the latest '''
     @wrap_exception('Unable to download Docker image')
     def download_image(self, docker_image, loop_callback):
+        if len(docker_image.split(":")) < 2:
+            logger.debug('Missing tag/digest on request docker image "%s", defaulting to latest', docker_image)
+            docker_image = ":".join([docker_image, "latest"])
```
percyliang: nit: I think we are using single quotes everywhere

percyliang (approved):
Another option would be to just throw an error and force a tag.  I think defaulting to 'latest' is better, but just to bring this up so we think it through.

Separately, we should really implement the other feature where you repull an image so you catch changes.

fabeschan:
Defaulting to latest is probably the most sensible, because doing a `docker pull imageref` without specifying a tag also results in the same behavior

worker/codalabworker/docker_client.py
fabeschan: fix'd

Issue #584, opened 2016-09-16
when scheduling jobs, pay attention to request-memory


percyliang:
Currently, worker reports free memory as what's not currently being used.  Instead, it should subtract away the memory that has been requested too via `--request-memory`.

Relevant line:
https://github.com/codalab/codalab-cli/blob/develop/codalab/worker/default_bundle_manager.py#L53

kashizui:
+1 this can be annoying

kashizui:
Also workers should actually report free memory

kashizui:
I just wanted to confirm what the desired behavior should be.. So there are two types of jobs:

- Jobs that don't request memory (let's call them "free-for-all" jobs) have no guarantee for how much memory they will or are allowed to use
- Jobs that do `--request-memory` (let's call them "explicit jobs") should be able to use that much memory on the worker without problems

Should free-for-all jobs and explicit jobs be allowed to coexist on the same worker (and how would that work)? or should a worker that currently has free-for-all jobs only be able to take more free-for-all jobs, while a worker that has explicit jobs can only take more explicit jobs that fit? 


I remember we came up with some trick where each job has an implicit "min" and a "max" requested memory, and using those we could implement a simple allocation algorithm but I don't think I've been able to replicate it exactly.. Something like, free-for-all jobs have `min=0` and `max=infty`, and explicit jobs have `min=max=request_memory`.

kashizui:
I've spent the afternoon thinking about this and I'm not sure what policy exists that could make free-for-all jobs play nice with explicit jobs. Since a free-for-all job is not checked (its memory usage is allowed to increase unbounded until machine goes OOM), scheduling a free-for-all job on the same worker as an explicit job nullifies the memory allocation guarantee for the latter. On the other hand, forcing free-for-all jobs and explicit jobs to be segregated might create resource fragmentation issues: once a machine has been claimed by free-for-all jobs, that machine cannot be used for explicit jobs until it is completely free.

kashizui:
Maybe the best solution is to forget free-for-all jobs completely, and force all unspecified jobs to take on some default request_memory value...

fabeschan:
Hmm, just wondering: what is docker's policy on resource allocation for containers with specified and unspecified resource requirements?

percyliang:
I think it's nice to be able to support free-for-all jobs and explicit
jobs.  I think the point of using the (min, max) formulation is that you
don't even have to think about the free-for-all jobs and explicit jobs
[...]

PR #800, opened 2017-10-15
Report unallocated memory only

kashizui:
First pass solution to #584 
Have workers report only allocated memory, while defining the potential memory usage of runs that don't have request_memory specified to be infinite.

Essentially implements:

jobs without request_memory:
* min: 0
* max: infty

jobs with request_memory:
* min: request_memory
* max: request_memory

```
for each job:
    for each worker:
        if worker.memory >= job.min_memory:
            worker.memory = max(0, worker.memory - job.max_memory)
            schedule(job, worker)
            break
``` 

One remaining problem here is that "free-for-all" jobs will still get scheduled on workers that already have "explicit" jobs (#584). To keep it simple, can fix this with appropriate if-statements in the scheduler code.

### DIFF
--- a/codalab/worker/default_bundle_manager.py
+++ b/codalab/worker/default_bundle_manager.py
@@ -50,12 +50,6 @@ class DefaultBundleManager(BundleManager):
                 if request_cpus > max_cpus:
                     failure_message = 'No workers with enough CPUs'
 
-            request_memory = self._compute_request_memory(bundle)
-            if request_memory:
-                max_memory = max(map(lambda worker: worker['memory_bytes'], workers_list))
-                if request_memory > max_memory:
-                    failure_message = 'No workers with enough memory'
-
             if failure_message is not None:
                 logger.info('Failing %s: %s', bundle.uuid, failure_message)
                 self._model.update_bundle(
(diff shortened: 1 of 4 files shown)


fabeschan:
Hmm lemme think if there is a better way to handle this "free-for-all" thing

fabeschan:
Hmm if I understand your pseudo code correctly, a single free-for-all job will reduce the worker's memory to 0?

percyliang:
But a free-for-all job can still run on that worker.

kashizui:
That’s right, once a free-for-all job gets assigned to a worker, its unallocated memory goes to 0, and then only free-for-all jobs can be further assigned to that worker.

fabeschan:
Ah, I see

percyliang:
What's the status of this?

fabeschan (approved):

kashizui:
This change is strictly better than what we had before (i.e. paying no attention to memory allocation), but still doesn't fully solve the problem of "guaranteeing X bytes of memory when a user requests X bytes of memory".

I just figured out a way for me to continue contributing to this patch through my employer's open-source policies. So I'm open to just merging this in or letting things simmer until I can come up with something better.

percyliang:
What about checking this in right now as a temporary solution and
submitting another PR if you happen to think of something?

kashizui:
I see that master's build is also failing.. is this still alright to merge?

percyliang:
Yeah...go ahead and merge...  we'll fix master.

PR #807, opened 2017-10-23
Authless SMTP support

krivard:
Some SMTP servers don't require u/p authentication; these flags allow such a server to be used with CodaLab.

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -366,10 +366,11 @@ class CodaLabManager(object):
     @cached
     def emailer(self):
         if 'email' in self.config:
+            # Default to authless SMTP (supported by some servers) if user/password is unspecified.
             return SMTPEmailer(
                 host=self.config['email']['host'],
-                user=self.config['email'].get('user','noreply@codalab.org'),#['user'],
-                password=self.config['email'].get('password',None),#['password'],
+                user=self.config['email'].get('user','noreply@codalab.org'), 
+                password=self.config['email'].get('password',None),
                 use_tls=True,
                 default_sender='CodaLab <noreply@codalab.org>',
                 server_email='noreply@codalab.org',
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -368,8 +368,8 @@ class CodaLabManager(object):
         if 'email' in self.config:
             return SMTPEmailer(
                 host=self.config['email']['host'],
-                user=self.config['email']['user'],
-                password=self.config['email']['password'],
+                user=self.config['email'].get('user','noreply@codalab.org'),#['user'],
+                password=self.config['email'].get('password',None),#['password'],
                 use_tls=True,
                 default_sender='CodaLab <noreply@codalab.org>',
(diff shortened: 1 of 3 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -368,8 +368,8 @@ def emailer(self):
         if 'email' in self.config:
             return SMTPEmailer(
                 host=self.config['email']['host'],
-                user=self.config['email']['user'],
-                password=self.config['email']['password'],
+                user=self.config['email'].get('user','noreply@codalab.org'),#['user'],
+                password=self.config['email'].get('password',None),#['password'],
```
maxwang7: Could you remove everything that's commented out? And also add a comment describing why you made those changes (the description you wrote in the conversation is fine, i.e. "Some SMTP servers don't require u/p authentication; these flags allow such a server to be used with CodaLab.")

codalab/lib/emailer.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, host, user, password, default_sender, server_email, port=587,
         self.server_email = server_email
         self.port = port
         self.use_tls = use_tls
+        self.do_login = self.password != None
```
maxwang7: Add parentheses around the `self.password != None` part of the expression for clarity.

monitor.py
```diff
@@ -73,8 +73,9 @@ def send_email(subject, message):
         return
 
     sender_host = sender_info['host']
-    sender_user = sender_info['user']
-    sender_password = sender_info['password']
+    sender_user = sender_info['user'] if 'user' in sender_info else 'noreply@codalab.org'
+    sender_password = sender_info['password'] if 'password' in sender_info else None
```
maxwang7: Could you use the `dictionary.get()` function instead?

maxwang7 (commented):
Thank you again for sending these in @krivard! I added some line level comments.

codalab/lib/emailer.py
```diff
@@ -59,7 +60,7 @@ def send_email(self, subject, body, recipient, sender=None,
                 mail_server.starttls()
                 mail_server.ehlo()
 
-            mail_server.login(self.user, self.password)
+            if self.do_login: mail_server.login(self.user, self.password)
```
fabeschan: nit: separate conditionals into two lines for clarity

monitor.py
```diff
@@ -85,7 +86,7 @@ def send_email(subject, message):
     msg['Subject'] = 'CodaLab on %s: %s' % (hostname, subject)
     msg['To'] = recipient
     msg['From'] = 'noreply@codalab.org'
-    s.login(sender_user, sender_password)
+    if do_login: s.login(sender_user, sender_password)
```
fabeschan: ditto

fabeschan (changes requested):
Just some nits on style

krivard:
Cool, I think I fixed everything so far, though check that the comment/s are what you had in mind -- I  clarified the variation in monitor.py with a link but it might be overkill.

maxwang7:
@krivard thank you for your contribution!

PR #810, opened 2017-11-02
Use standard python docker bindings for pulling images

jder:
The docker repository we're using requires authentication, which happens client-side. The standard docker python package includes support for this functionality (in particular, using docker credential helpers), so we started using it for pulling images. 

@percyliang asked that we put this up for PR to start discussion about whether we should start moving everything over to using this package rather than using the docker HTTP API directly.

### DIFF
--- a/worker/codalabworker/docker_client.py
+++ b/worker/codalabworker/docker_client.py
@@ -8,6 +8,7 @@ import ssl
 import struct
 import sys
 import subprocess
+import docker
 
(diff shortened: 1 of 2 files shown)


percyliang (approved):
Great!

Historically, a desideratum was to keep the codalabworker as self-contained as possible so that people could run it easily, but moving forward, switching to the `docker` library is the way to go.

PR #813, opened 2017-11-09
Run thread periodically resumes a running bundle

fabeschan:
If a server goes offline for more than 30 seconds for any reason (most typically during re-deployments), then running bundles will be treated as offline. To avoid this from happening, every run thread periodically attempts to resume a running bundle. So now this acts like a heartbeat.


percyliang (approved):
Could you document the behavior as a comment block in the code - in particular, what the states represent, what is the life cycle, and what the behavior is when various things happen (worker goes offline for X seconds and then comes back online)?

fabeschan:
Done

PR #817, opened 2017-11-13
Revival of Stephen's Memory Allocation PR

fabeschan:
Now that we've reverted master, let's see if Stephen's patches pass CLI tests.

Original PR: #800


fabeschan:
I haven't had much time to look into it but I think the local instance of the rest server crashes so CLI tests fail with that stack trace

kashizui:
Hm. I tested this locally at the time and it was working -- I'll look into this when I get the chance.

kashizui:
Lol it was a silly typo!

percyliang:
Thank goodness we have tests.

PR #822, opened 2017-12-05
Implements user search

maxwang7:
Associated worksheets PR: codalab-worksheets:#379: https://github.com/codalab/codalab-worksheets/pull/379

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1742,17 +1742,9 @@ class BundleModel(object):
         cols_to_select = [
             cl_user.c.user_id,
             cl_user.c.user_name,
-            cl_user.c.email,
-            cl_user.c.last_login,
-            cl_user.c.is_active,
             cl_user.c.first_name,
             cl_user.c.last_name,
             cl_user.c.affiliation,
-            cl_user.c.time_quota,
-            cl_user.c.time_used,
-            cl_user.c.disk_quota,
-            cl_user.c.disk_used,
-            cl_user.c.disk_quota,
         ]
         query = select(cols_to_select).distinct().where(clause).limit(limit)
         with self.engine.begin() as connection:
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1754,10 +1754,8 @@ class BundleModel(object):
             cl_user.c.disk_used,
             cl_user.c.disk_quota,
         ]
-        print clause
         query = select(cols_to_select).distinct().where(clause).limit(limit)
         with self.engine.begin() as connection:
-            print 'connection: {}'.format(connection)
             rows = connection.execute(query).fetchall()
             if not rows:
                 return []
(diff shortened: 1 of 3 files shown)


codalab/model/bundle_model.py
```diff
@@ -1726,6 +1726,41 @@ def get_user(self, user_id=None, username=None, check_active=True):
         if result:
             return result[0]
         return None
+    
+    def search_users(self, user_id, keywords):
+        clauses = []
+        limit = 5
+        for keyword in keywords:
+            if '.limit' not in keyword:
```
maxwang7, line 1734: I'm not sure the best way to filter out filters (e.g. `.limit` or `.sort=reverse`). Thoughts?

codalab/model/bundle_model.py
```diff
@@ -1726,6 +1726,41 @@ def get_user(self, user_id=None, username=None, check_active=True):
...
+            cl_user.c.is_active,
+            cl_user.c.first_name,
+            cl_user.c.last_name,
+            cl_user.c.affiliation,
+            cl_user.c.time_quota,
+            cl_user.c.time_used,
+            cl_user.c.disk_quota,
+            cl_user.c.disk_used,
+            cl_user.c.disk_quota,
+        ]
```
maxwang7, line 1748: Are there fields we would want to include/exclude?

percyliang: Don't include sensitive information (e.g., email)!  I'd only include information that's going to show up on the webpage.  Err on the side of including less.

codalab/rest/users.py
```diff
@@ -76,20 +79,27 @@ def fetch_user(user_spec):
...
-    # Combine username and email filters
-    usernames = set(request.query.get('filter[user_name]', '').split(','))
-    usernames |= set(request.query.get('filter[email]', '').split(','))
-    usernames.discard('')  # str.split(',') will return '' on empty strings
-    users = local.model.get_users(usernames=(usernames or None))
-    return allowed_user_schema()(many=True).dump(users).data
+    keywords = query_get_list('keywords')
+    if keywords:
+        users = local.model.search_users(request.user.user_id, keywords)
+        return allowed_user_schema()(many=True).dump(users).data
```
percyliang, line 95: Take this line out of the if/else to avoid duplication

codalab/model/bundle_model.py
```diff
@@ -1726,6 +1726,41 @@ def get_user(self, user_id=None, username=None, check_active=True):
...
+        query = select(cols_to_select).distinct().where(clause).limit(limit)
+        with self.engine.begin() as connection:
+            rows = connection.execute(query).fetchall()
+            if not rows:
+                return []
```
percyliang, line 1753: Does this return `None`?  Can't return rows directly?

codalab/model/bundle_model.py
percyliang: A hack is to just ignore anything that starts with a period.  Definitely should make this more generic.  If we're not going to respect `limit`, then we should document that.

percyliang (changes requested):
Mostly looks good, just a few localized changes to make.

PR #828, opened 2018-01-07
Initialize worker docker client ssl context to None by default

bkgoksel:
Without this, the worker fails when run in a system where Docker client connects to the Docker daemon through the DOCKER_HOST environment variable, but without TLS. The code was otherwise written to handle this case, but the None check for _ssl_context fails without this initialization.


percyliang (approved):

fabeschan (approved):
nice catch!

fabeschan:
@bkgoksel feel free to merge if unit tests pass locally

fabeschan:
@bkgoksel can you rebase your commits with the (now rewritten) master branch and see if tests pass?

PR #833, opened 2018-01-20
--allow-failed-dependencies also allows killed dependencies

### LINKED ISSUE #823: Depending on killed runs
Hi CodaLab team!

When running neural network jobs, we typically set a high iteration limit and kill the job once we see that the model has converged. However, once we've killed it, new jobs we create that depend on that run remain in the 'created' state, and never transition to 'staged' or 'running'.

We've tried using the command-line parameter `--allow-failed-dependencies` when launching our new jobs, but the new jobs still refuse to run. Perhaps a new parameter `--allow-killed-dependencies` is necessary? Otherwise, the behavior of `--allow-failed-dependencies` could be updated to allow killed dependencies also.

Let me know if you have any questions!
Thanks!
Connor

fabeschan:
Fixes #823

### DIFF
--- a/codalab/bundles/derived_bundle.py
+++ b/codalab/bundles/derived_bundle.py
@@ -10,7 +10,7 @@ from codalab.objects.metadata_spec import MetadataSpec
 
 class DerivedBundle(NamedBundle):
     METADATA_SPECS = list(NamedBundle.METADATA_SPECS)
-    METADATA_SPECS.append(MetadataSpec('allow_failed_dependencies', bool, 'Whether to allow this bundle to have failed dependencies.', default=False))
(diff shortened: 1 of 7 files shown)


fabeschan:
TODO: test this

percyliang (approved):

percyliang:
Actually, it might be better if we define a variable that includes all 'done' states.  I'm sure this logic exists elsewhere, and just want to make this consistent in case we need to add more states in the future.

fabeschan:
Wow I pulled in a lot of other unrelated commits

fabeschan:
> Actually, it might be better if we define a variable that includes all 'done' states. I'm sure this logic exists elsewhere, and just want to make this consistent in case we need to add more states in the future.

OK, let me revisit this at a later time. I will elect to close this for now.

PR #838, opened 2018-01-26
Implement GPU bandaid

fabeschan:
Tested!

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -292,14 +292,33 @@ class BundleManager(object):
                 else:
                     continue  # Try the next worker.
 
+    def _deduct_worker_resources(self, workers_list):
+        """
+        From each worker, subtract resources used by running bundles. Modifies the list.
+        """
+        for worker in workers_list:
+            for uuid in worker['run_uuids']:
+                bundle = self._model.get_bundle(uuid)
+                worker['cpus'] -= self._compute_request_cpus(bundle) or 1
+                worker['gpus'] -= self._compute_request_gpus(bundle) or 0
+                #TODO: memory_bytes
+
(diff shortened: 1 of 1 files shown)


codalab/worker/bundle_manager.py
```diff
@@ -334,7 +353,7 @@ def _filter_and_sort_workers(self, workers_list, bundle):
         # Sort workers list according to these keys in the following succession:
         #  - if the bundle doesn't request GPUs (only requests CPUs), prioritize workers that don't have GPUs
```
percyliang: Can we rewrite this comment in the "X, descending" style?

codalab/worker/bundle_manager.py
```diff
@@ -347,11 +366,12 @@ def _filter_and_sort_workers(self, workers_list, bundle):
                               bundle.dependencies))
         def get_sort_key(worker):
             deps = set(worker['dependencies'])
+            worker_id = worker['worker_id']
 
             # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
             if not self._compute_request_gpus(bundle):
-                return (-worker['gpus'], len(needed_deps & deps), worker['slots'] - len(worker['run_uuids']), random.random())
-            return (len(needed_deps & deps), worker['slots'] - len(worker['run_uuids']), random.random())
+                return (not has_gpu[worker_id], len(needed_deps & deps), worker['cpus'], random.random())
```
percyliang: Can we avoid having duplicate code here for the two cases?  Seems like defining a variable to be `not self._compute_request_gpus(bundle) and not has_gpu[worker_id]` (check my logic) and using that as the first variable would do the trick.

percyliang (approved):
LGTM modulo tidying up code

Issue #381, opened 2016-03-08
Add indices on access_token and refresh_token in database


kashizui:

kashizui:
already exists.

Issue #842, opened 2018-01-31
Make upload ignore macosx files


bkgoksel:
Duplicate of https://github.com/codalab/codalab-worksheets/issues/381
Moving here as it is a CLI issue

PR #843, opened 2018-01-31
Ignore default ignored files like DS_Store everywhere

bkgoksel:
We already had a list of filename patterns that we excluded when they were located in archives that were uploaded, Generalized that logic to basically always exclude those patterns by default, whether they're uploaded from within archives or uploaded from within directories.

Didn't add any new patterns for now. Let's see if anything else comes up.

Fixes: #381 #842

### DIFF
--- a/tests/lib/upload_manager_test.py
+++ b/tests/lib/upload_manager_test.py
@@ -32,7 +32,7 @@ class UploadManagerTest(unittest.TestCase):
             FakeBundle(), sources,
             follow_symlinks, exclude_patterns, remove_sources,
             git, unpack, simplify_archives)
-    
+
     def test_single_local_path(self):
         source = os.path.join(self.temp_dir, 'filename')
         self.write_string_to_file('testing', source)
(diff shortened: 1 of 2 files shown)


tests/lib/upload_manager_test.py
```diff
@@ -40,6 +40,16 @@ def test_single_local_path(self):
...
         self.check_file_contains_string(self.bundle_location, 'testing')
 
+    def test_ignored_files(self):
+        dsstore_file = os.path.join(self.temp_dir, '.DS_Store')
+        self.write_string_to_file('testing', dsstore_file)
+        source = os.path.join(self.temp_dir, 'filename')
+        self.write_string_to_file('testing', source)
+        self.do_upload([self.temp_dir])
+        self.assertTrue(os.path.exists(os.path.join(self.bundle_location, 'filename')))
+        self.assertFalse(os.path.exists(os.path.join(self.bundle_location, '.DS_Store')))
```
percyliang, line 51: Add tests for the other patterns too

percyliang (approved):
LGTM modulo comment

tests/lib/upload_manager_test.py
bkgoksel: Adding one for __MACOSX too, but I have no idea what the third pattern is for, and I've never had files besides __MACOSX and .DS_Store before, so it's not clear to me what to test that pattern with.

fabeschan (approved):
Good job! Looks good

fabeschan:
Make sure to fix tests though

PR #844, opened 2018-01-31
Removed newline-adding logic from mimic

### LINKED ISSUE #788: Stop adding newlines between cl mimic bundles
This is totally a nit, but I find it annoying that when you cl mimic, it adds a newline in the worksheet markdown between every bundle.  I usually try to put things in tables to control what columns show up on the web interface, but the newlines mess this up and I have to go in and delete them every time.

bkgoksel:
Fixes #788 
Removing the newline adding logic completely. 

If we find that this breaks some workflows, this commit can be reverted as it is an isolated change.

### DIFF
--- a/codalab/lib/bundle_util.py
+++ b/codalab/lib/bundle_util.py
@@ -242,7 +242,6 @@ def mimic_bundles(client,
         new_bundle_uuids_added = set()
 
         # Whether there were items that we didn't include in the prelude (in which case we want to put '')
-        skipped = True
 
         if len(host_worksheet_uuids) > 0:
             # Choose a single worksheet.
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_util.py
```diff
@@ -242,7 +242,6 @@ def newline():
         new_bundle_uuids_added = set()
 
         # Whether there were items that we didn't include in the prelude (in which case we want to put '')
```
percyliang: Remove comment too

percyliang: Also, can we safely delete `newline()`?

percyliang (approved):
LGTM modulo cleanup

fabeschan:
(You can ignore the failing travis CI runs)

PR #848, opened 2018-02-09
Running bundles now continuously heartbeat the server during dependency/image downlaods

fabeschan:


worker/codalabworker/run.py
```diff
@@ -127,14 +127,7 @@ def resume(self):
         if not self._bundle_service.resume_bundle(self._worker.id, self._uuid,
                                                  start_message):
             return False
-
-        # Start a thread for this run.
-        def resume_run(self):
```
percyliang, line 132: How come we don't need a thread anymore?  That seems orthogonal sending a heartbeat?

fabeschan: Ya so the reason for that is this thread that runs `resume` only starts after dependencies and images are downloaded, when `resume` should be repeatedly called during those phases. So we remove this whole thread set up and put `resume` in `_throttled_updater` which already runs periodically on a thread.

percyliang (approved):

worker/codalabworker/run.py
fabeschan: Turns out we do need a thread.. whoops

PR #850, opened 2018-02-09
 Limit the maximum serialized json length of worker dependency list

fabeschan:
Todo: test this thoroughly


percyliang (commented):
What do you think about bounding the number of dependencies rather than the serialized length?  It seems like a more intuitive quantity to reason about (though admittedly less directly addressing the problem).  Just wanted to talk through the pros/cons.

fabeschan:
Both are fine, but direct is better imo as it deals with the number of characters (and therefore bytes)

percyliang (approved):

PR #853, opened 2018-02-14
Containers are now memory-limited by --request-memory

fabeschan:
Fix bug in _filter_and_sort_workers()

Add tests for BundleManager


codalab/worker/bundle_manager.py
```diff
@@ -371,7 +371,7 @@ def get_sort_key(worker):
 
             # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
             t = not self._compute_request_gpus(bundle) and has_gpu[worker_id]
```
percyliang: It would be a good idea to give `t` a more meaningful name, so it's easier to reason about this code.

percyliang (approved):
Ah!

PR #856, opened 2018-02-15
Remove "slots" column from workers

fabeschan:
Test this thoroughly.

Also make sure to make the appropriate changes in `codalab-deployment`


worker/codalabworker/main.py
```diff
@@ -46,9 +46,6 @@ def main():
                              'the least recently used images are removed first. '
                              'Worker will not remove any images if this option '
                              'is not specified.')
-    parser.add_argument('--slots', type=int, default=1,
```
percyliang, line 49: We need some way of capping the number of CPUs (and GPUs, memory) available to the worker, no?

fabeschan: Yup, we do

PR #857, opened 2018-02-16
Enforce resource limits properly

bkgoksel:
Fixes broken disk quota enforcement by preventing users from going above their disk quota at all by:
* failing upload bundles that would go over users' disk quotas
* treating the remaining disk quota of a user as a hard limit for `request_disk`

Previously we were only preventing new bundles from being created after users had already filled their disk quota, allowing arbitrarily sized bundles to be created before preventing new bundles from being created.

After merging this merge [the deployments repo PR](https://bitbucket.org/percyliang/codalab-deployment/pull-requests/4/global-max-request-disk-updated-to-100gb/diff) to increase the global request disk limit to let people have larger run bundles if their disk quotas allow it

It also unifies the logic for enforcing resource limits in one  place, adding correct enforcement for time quota as well. Also adds helpful error messages when refusing to run a bundle due to resource limits.


codalab/lib/upload_manager.py
```diff
@@ -148,7 +148,7 @@ def _simplify_directory(self, path, child_path=None):
         path_util.rename(child_path, path)
         path_util.remove(temp_path)
 
-    def update_metadata_and_save(self, bundle, new_bundle):
+    def update_metadata_and_save(self, bundle, new_bundle, raise_disk_quota_full=False):
```
percyliang: When would we want `raise_disk_quota_full` to be False?

bkgoksel: I don't know if this is a case in our codebase atm, but theoretically we could call this function to update the metadata of an existing bundle (not newly uploaded)
If for some reason the user went over their disk quota in the meantime for some other reason, we don't want to throw an error for an unrelated update metadata call. 

So basically if we're calling this method from somewhere that's not uploading new contents, we might want this to be false.

This was my reasoning for adding an optional argument, but I'm open to changing my mind on this.

bkgoksel: Also realized another potential problem with this. If the bundle is not just created, its data size will already be subtracted from the user's available disk quota, so this check should not be made in that case.

It really seems like this should fall under the `new_bundle` case, but that's not how we make the call (it's set to false when we call this after initial upload), and in fact in our entire codebase we never call this function with `new_bundle=True`, so I'm not sure what purpose that flag serves.

worker/codalabworker/formatting.py
```diff
@@ -53,16 +53,16 @@ def parse_size(s):
     """
     try:
         if s[-1].isdigit():
-            return float(s)
+            return int(s)
```
percyliang, line 56: I'm trying to remember why the `float` was there - must have been a reason for it, since it's a thing that you kind of go out of your way to do.  Maybe it was to avoid overflow (though Python uses big decimals automatically anyway)?  Anyway, can't think of it.

percyliang (approved):
LGTM

codalab/lib/upload_manager.py
```diff
@@ -166,20 +166,18 @@ def update_metadata_and_save(self, bundle, new_bundle):
 
         data_hash = '0x%s' % (path_util.hash_directory(bundle_path, dirs_and_files))
         data_size = path_util.get_size(bundle_path, dirs_and_files)
+        if enforce_disk_quota:
+            disk_left = self._bundle_model.get_user_disk_quota_left(bundle.owner_id)
+            if data_size > disk_left:
+                raise UsageError("Can't save bundle, bundle size %s greater than user's disk quota left: %s" % (data_size, disk_left))
```
fabeschan, line 172: does a bundle that gets expanded beyond the disk quota actually fall into a fail state after reaching this point?

codalab/worker/bundle_manager.py
```diff
@@ -421,6 +420,22 @@ def _compute_request_memory(self, bundle):
             return formatting.parse_size(bundle.metadata.request_memory)
         return self._default_request_memory
 
+    def _compute_request_disk(self, bundle):
+        """
+        Compute the disk limit used for scheduling the run.
+        """
+        if bundle.metadata.request_disk:
+            return formatting.parse_size(bundle.metadata.request_disk)
+        return self._default_request_disk
```
fabeschan: Double check that if this returns None (when self._default_request_disk is returned) then whatever is using the result of this function call doesn't crash

codalab/worker/bundle_manager.py
```diff
@@ -421,6 +420,22 @@ def _compute_request_memory(self, bundle):
...
+
+    def _compute_request_time(self, bundle):
+        """
+        Compute the time limit used for scheduling the run.
+        """
+        if bundle.metadata.request_time:
+            return formatting.parse_duration(bundle.metadata.request_time)
+        return self._default_request_time
```
fabeschan: ditto

codalab/worker/default_bundle_manager.py
```diff
@@ -34,6 +34,14 @@ def _schedule_run_bundles(self):
         self._schedule_run_bundles_on_workers(workers, user_owned=True)
         self._schedule_run_bundles_on_workers(workers, user_owned=False)
 
+    def _fail_on_excess_resource(self, value, user_fail_string=None, global_fail_string=None, user_max=None, global_max=None):
```
fabeschan: can you choose a different name for this fn? Sounds almost the same as the one below

fabeschan (changes requested):
some nits

codalab/lib/upload_manager.py
bkgoksel: We have that logic builtin for run bundles. Run bundles fail as soon as their data size hits `request_disk`. This change prevents upload bundles from overcoming the limit.

I'd say this is not the best way of doing it for upload bundles as this function only gets called after the upload is done. However trying to enforce the quota at every possible upload code path would be very messy (especially in places where we just call command line programs like git clone). So this was the second best way I could think of as at least it cleans up the disk afterwards.

codalab/worker/bundle_manager.py
bkgoksel: This should not return `None`. We should always have a default request disk to limit how much disk can be used by any run. If we do not want to impose an arbitrary default request disk, I can make changes so it defaults to user disk quota left.  Thoughts? @percyliang @fabeschan

codalab/worker/bundle_manager.py
bkgoksel: ditto my reply above

codalab/worker/default_bundle_manager.py
bkgoksel: will do 👍

fabeschan (approved):

fabeschan:
So I believe `BundleManager. _default_request_XYZ` do not actually modify the bundle's corresponding metadata, and for things like memory, cpu and gpus, this behavior is not transparent to the user. My recommendation is to get rid of these _default values from the BundleManager and incorporate them directly into the bundle metadata (in bundle_cli.py perhaps)

Thoughts? @bkgoksel @percyliang

codalab/worker/default_bundle_manager.py
```diff
@@ -34,6 +34,14 @@ def _schedule_run_bundles(self):
         self._schedule_run_bundles_on_workers(workers, user_owned=True)
         self._schedule_run_bundles_on_workers(workers, user_owned=False)
 
+    def _check_resource_failure(self, value, user_fail_string=None, global_fail_string=None, user_max=None, global_max=None):
```
fabeschan, line 37: Should this function check against the bundle's request_queue?

percyliang:
Transparency sounds good to me.

bkgoksel:
That makes sense to me, but the question is do we still want to make these values configurable? 

Another point is that it's a bit more resilient to have server side defaults so that malformed requests don't break things (maybe from old client versions)

fabeschan:
Once it's done, we can always make these values configurable later on.

and re: server side verification - this is something we need anyway, since we are hosting a rest endpoint

PR #860, opened 2018-02-18
Use standard docker library for pulls

jder:
Recreated version of [this PR](https://github.com/codalab/codalab-cli/pull/810) which passes the new travis tests: 

> The docker repository we're using requires authentication, which happens client-side. The standard docker python package includes support for this functionality (in particular, using docker credential helpers), so we started using it for pulling images.

The first commit is the actual change, the second fixes CI. (The worker.sh script needs to run inside the virtualenv so it can use the python deps installed for the worker; I think it was just luck this wasn't a problem before. If you like we could just activate the virtualenv at the top and drop all the ./venv/bin paths everywhere.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
   - printf "$CODALAB_USERNAME\n$CODALAB_PASSWORD\n" > /home/travis/.codalab/root.password
   - chmod 600 /home/travis/.codalab/root.password
   - sleep 15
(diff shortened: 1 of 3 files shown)


percyliang (approved):

fabeschan:
I don't know the proper way to trigger travis on commits from external forks, so I'm going to merge this into `docker-pr`. If travis tests pass there then feel free to merge it into master as well.

fabeschan:
Since tests passed, this PR has been (manually) merged into master.

PR #862, opened 2018-02-21
Adds support for optional named macro inputs

bkgoksel:
Old-style macros still work, and numbered and named arguments can be used together as well.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1922,7 +1922,7 @@ class BundleCLI(object):
         'macro',
         help=[
             'Use mimicry to simulate macros.',
-            '  macro M A B named-input:C <=> mimic M-in1 M-in2 M-in-named-input M-out A B C'
+            '  macro M A B <name1>:C <name2>:D <=> mimic M-in1 M-in2 M-in-name1 M-in-name2 M-out A B C D'
         ],
         arguments=(
             Commands.Argument('macro_name', help='Name of the macro (look for <macro_name>-in1, <macro_name>-in-<name>, ..., and <macro_name>-out bundles).'),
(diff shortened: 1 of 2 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1922,11 +1922,11 @@ def do_mimic_command(self, args):
         'macro',
         help=[
             'Use mimicry to simulate macros.',
-            '  macro M A B   <=>   mimic M-in1 M-in2 M-out A B'
+            '  macro M A B named-input:C <=> mimic M-in1 M-in2 M-in-named-input M-out A B C'
```
percyliang: can you make it clearer what the generalization is?
`<name1>:C <name2>:D` ...

codalab/lib/bundle_cli.py
```diff
@@ -1939,8 +1939,21 @@ def do_macro_command(self, args):
         if not getattr(args, metadata_util.metadata_key_to_argument('name')):
             setattr(args, metadata_util.metadata_key_to_argument('name'), 'new')
         # Reduce to the mimic case
-        args.bundles = [args.macro_name + '-in' + str(i+1) for i in range(len(args.bundles))] + \
-                       [args.macro_name + '-out'] + args.bundles
+        named_inputs = [bundle for bundle in args.bundles if ':' in bundle]
+        named_input_names = [bundle.split(':',1)[0] for bundle in named_inputs]
+
+        named_user_inputs = [bundle.split(':',1)[1] for bundle in named_inputs]
```
percyliang: Might be better to combine this with previous line and use `named_input_names, named_user_inputs = zip(*[...])`

codalab/lib/bundle_cli.py
```diff
@@ -1939,8 +1939,21 @@ def do_macro_command(self, args):
...
+        named_macro_inputs = [args.macro_name + '-in-' + named_in for named_in in named_input_names]
+
+        numbered_user_inputs = [bundle for bundle in args.bundles if ':' not in bundle]
```
percyliang: Would be better to have the logic for checking `:` in bundle deduped / more localized - first partition the args (`itertools.partition`), and then process each one.

test-cli.py
```diff
@@ -924,21 +924,39 @@ def test(ctx):
...
-    uuid4 = run_command([cl, 'mimic', uuid1, uuid3, '-n', 'new'])
-    check_equals(data_hash(uuid2), data_hash(uuid4))
-    uuid5 = run_command([cl, 'mimic', uuid1, uuid2, uuid3, '-n', 'new'])
-    check_equals(data_hash(uuid2), data_hash(uuid5))
-    uuid6 = run_command([cl, 'macro', name, uuid3, '-n', 'new'])
-    check_equals(data_hash(uuid2), data_hash(uuid6))
+    input_mimic_uuid = run_command([cl, 'mimic', input_uuid, new_input_uuid, '-n', 'new'])
+    check_equals(data_hash(simple_out_uuid), data_hash(input_mimic_uuid))
+
+    full_mimic_uuid = run_command([cl, 'mimic',input_uuid, simple_out_uuid, new_input_uuid, '-n', 'new'])
```
percyliang: nit: put space after `,` (also fix other places too)

percyliang (changes requested):
Functionality looks good.  Just some low-level comments.

codalab/lib/bundle_cli.py
bkgoksel: Done

codalab/lib/bundle_cli.py
bkgoksel: The zip approach doesn't work if there are no named inputs unless wrapped in another if block. Instead I just looped over all the input bundles and did all the processing bundle-by-bundle

codalab/lib/bundle_cli.py
bkgoksel: `itertools.partition` isn't a python builtin but a recipe they recommend using. I didn't want to define an inline partition function just for this, so I also just encompassed this logic in the bundle-by-bundle processing for loop.

test-cli.py
bkgoksel: Done

percyliang (approved):

PR #865, opened 2018-02-22
Resource Isolation

### LINKED ISSUE #841: CPU and GPU Isolation
Keep track of CPU and GPU devices on workers, and assign them to running bundles during scheduling. Implement resource isolation for better behavior.

There's no way to limit/reserve a specific amount of GPU memory for a process, but we should be able to at least isolate GPUs.

fabeschan:
Fixes #841 

- [x]  Remove worker "slots"
- [x]  Add cpuset and gpuset to worker + worker_run tables
- [x]  Run bundle containers now use cpuset assignment
- [x]  Nvidia docker arguments now use gpuset assignment
- [x]  Fix bug in fill_missing_metadata() not using provided default values
- [x]  Bundle manager now fails run bundles that request more gpus than available on any worker
- [x]  Resume bundle now pushes cpuset and gpuset info

EDIT:

- [x]  Worker is responsible for assigning specific CPU/GPUs to bundles, not the server.


alembic/versions/2018022007_add_columns_cpuset_and_gpuset__19b1e7f3608b.py
```diff
@@ -0,0 +1,24 @@
+"""Add columns cpuset and gpuset to worker_run table
```
percyliang: Why do we have both this migration that adds cpuset/gpuset and the one that replaces cpus/gpus?

codalab/model/bundle_model.py
```diff
@@ -588,17 +588,36 @@ def set_waiting_for_worker_startup_bundle(self, bundle, job_handle):
...
         """
         Sets the bundle to STARTING, updating the last_updated metadata. Adds
         a worker_run row that tracks which worker will run the bundle.
+
+        Return a dict consisting of the updated worker_run_row if success or None otherwise
         """
+        user_id = worker['user_id']
+        worker_id = worker['worker_id']
+
+        unused_cpuset = set(worker['cpuset'])
```
percyliang: One thing that we should make sure we support is that there's going to be a PR that has a worker backed by AWS batch which effectively has an infinite number of CPUs/GPUs, in which case cpuset/gpuset will be something like None.

codalab/worker/bundle_manager.py
```diff
@@ -62,8 +63,8 @@ def parse(to_value, field):
         self._max_request_memory = parse(formatting.parse_size, 'max_request_memory')
         self._max_request_disk = parse(formatting.parse_size, 'max_request_disk')
 
-        self._default_request_cpus = config.get('default_request_cpus')
-        self._default_request_gpus = config.get('default_request_gpus')
+        self._default_request_cpus = config.get('default_request_cpus', 1)
```
percyliang, line 65: Why have defaults here for some fields but not others?

codalab/worker/bundle_manager.py
```diff
@@ -284,56 +285,56 @@ def _schedule_run_bundles_on_workers(self, workers, user_owned):
             else:
                 workers_list = workers.user_owned_workers(self._model.root_user_id)
 
-            workers_list = self._filter_and_sort_workers(workers_list, bundle)
+            # save _compute_request_X values because somehow they get mutated somewhere down the line (wtf!)
+            # i.e. multiple calls may return different values; which screws up scheduling really badly
+            # TODO: hunt down this bug
```
percyliang: A multithreading issue?  Want to fix before merging?

codalab/worker/bundle_manager.py
```diff
@@ -284,56 +285,56 @@ def _schedule_run_bundles_on_workers(self, workers, user_owned):
...
-        has_gpu = {}
+        worker_has_gpu = {} # keep track of which workers have GPUs
+        worker_free_cpus = {}
+        worker_free_gpus = {}
+
+        # From each worker, initialize with the cpuset/gpuset lengths
         for worker in workers_list:
             worker_id = worker['worker_id']
-            has_gpu[worker_id] = True if worker['gpus'] > 0 else False
+            worker_has_gpu[worker_id] = len(worker['gpuset']) > 0
```
percyliang: why do we need this this? you can just compute from worker_free_gpus locally to avoid having as much state.

worker/codalabworker/docker_client.py
```diff
@@ -436,7 +438,9 @@ def start_container(self, bundle_path, uuid, command, docker_image,
             'Entrypoint': [""], # unset entry point regardless of image
             'HostConfig': {
                 'Binds': volume_bindings,
+                'NetworkMode': network_name,
```
percyliang: Why is this part of this PR?

worker/codalabworker/docker_client.py
```diff
@@ -445,9 +449,7 @@ def start_container(self, bundle_path, uuid, command, docker_image,
 
         # TODO: Allocate the requested number of GPUs and isolate
```
percyliang: Is this still a TODO?

percyliang (commented):
Nice!  The logic is much more sane now.

percyliang:
Random thought: It feels a bit strange that the master is managing which CPUs and GPUs the worker should use if they are interchangeable...

alembic/versions/2018022007_add_columns_cpuset_and_gpuset__19b1e7f3608b.py
fabeschan: This one keeps track of cpu/gpusets occupied per bundle ; the other one is for storing cpu/gpusets offered by the worker.

codalab/worker/bundle_manager.py
fabeschan: actually i don't think these are used at all; none of our workers have these _default_* configs defined. I think I'll just revert it to whatever it was before.

codalab/worker/bundle_manager.py
fabeschan: I've been looking into this for 5 hours, narrowing the cause for some wacky scheduling behavior down to this part, and this is where I could no longer make progress. So this is why I decided to implement a work around instead of spending more (unbounded) time to fix this bug.

codalab/worker/bundle_manager.py
fabeschan: We differentiate between workers that have 0 gpus and workers that have 0 _free_ gpus; this is so that we can schedule CPU-only bundles to CPU-only workers (aka 0 gpu workers) preferentially.

worker/codalabworker/docker_client.py
fabeschan: nope, I should remove this

worker/codalabworker/docker_client.py
fabeschan: Just a small re-arrangement, moving the same functionality from line 450 to here.

fabeschan:
Added todo: Worker is responsible for assigning specific CPU/GPUs to bundles, not the server.

codalab/worker/bundle_manager.py
fabeschan: nevermind. These fields are used by _compute_request_*pus and they must be integers rather than NoneType otherwise they crash something else down the line. This is why the change are made

fabeschan:
@percyliang ready for another review

worker/codalabworker/worker.py
```diff
@@ -45,16 +45,21 @@ class Worker(object):
...
         self.shared_file_system = shared_file_system
         self._bundle_service = bundle_service
         self._docker = docker
         self._docker_network_prefix = docker_network_prefix
-        self._slots = slots
+
+        self._resource_lock = threading.Lock() # lock for cpuset and gpuset
+        self._cpuset = cpuset
+        self._gpuset = gpuset
+        self._cpuset_free = set(cpuset)
```
percyliang: It's weird if `cpuset` is not a set...make it a set or call it `cpus`?

worker/codalabworker/worker.py
```diff
@@ -92,13 +97,30 @@ def run(self):
         if not self.shared_file_system:
             self._dependency_manager.start_cleanup_thread()
 
+        self._worker_state_manager.resume_previous_runs(
+                lambda run_info: Run.deserialize(
+                    self._bundle_service, self._docker, self._image_manager, self, run_info)
+        )
+        with self._resource_lock:
+            run_sets = self._worker_state_manager.map_runs(lambda run: (run._cpuset, run._gpuset))
+            for cpuset, gpuset in run_sets:
```
percyliang: Please document what the high-level logic here is

worker/codalabworker/worker.py
```diff
@@ -188,15 +206,53 @@ def _checkin(self):
                     self._exiting = True
                 self._should_upgrade = True
 
+    def _allocate_cpu_and_gpu_sets(self, request_cpus, request_gpus):
+        """
+        Allocate a cpuset and gpuset to assign to a bundle based on given requested resources.
```
percyliang, line 214: Write that `request_cpus` is a number.

worker/codalabworker/main.py
```diff
@@ -115,5 +127,50 @@ def main():
...
+            raise ValueError("CPUSET_STR invalid format: must be a string of comma-separated integers")
+
+        if not len(cpuset) == len(set(cpuset)):
+            raise ValueError("CPUSET_STR invalid: CPUs not distinct values")
+        if not all(cpu in range(cpu_count) for cpu in cpuset):
+            raise ValueError("CPUSET_STR invalid: CPUs out of range")
+    return set(cpuset)
+
+def parse_gpuset_args(docker_client, arg):
+    if arg == '':
```
percyliang: Document what the format of `arg` is

worker/codalabworker/main.py
```diff
@@ -115,5 +127,50 @@ def main():
...
+
+    info = docker_client.get_nvidia_devices_info()
+    all_gpus = []
+    if info is not None:
+        for d in info['Devices']:
+            m = re.search('^/dev/nvidia(\d+)$', d['Path'])
+            all_gpus.append(int(m.group(1)))
+
+    if arg == 'ALL':
+        if info is None:
```
percyliang: This is unnecessary

worker/codalabworker/main.py
```diff
@@ -115,5 +127,50 @@ def main():
 
     worker.run()
 
+def parse_cpuset_args(arg):
+    cpu_count = multiprocessing.cpu_count()
```
percyliang: Document what the format of `arg` is

codalab/worker/bundle_manager.py
```diff
@@ -299,23 +299,20 @@ def _deduct_worker_resources(self, workers_list):
...
-
-        Note: the worker slots field is no longer in use so it is effectively ignored
         """
 
         # keep track of which workers have GPUs
         has_gpu = {}
         for worker in workers_list:
             worker_id = worker['worker_id']
-            has_gpu[worker_id] = True if worker['gpus'] > 0 else False
+            has_gpu[worker_id] = worker['gpus'] > 0
```
percyliang, line 315: Still don't understand why `has_gpu` is needed...can't you always just write `worker['gpus'] > 0` instead?

codalab/bundles/run_bundle.py
```diff
@@ -27,8 +27,8 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_time', basestring, 'Amount of time (e.g., 3, 3m, 3h, 3d) allowed for this run.', formatting='duration'))
     METADATA_SPECS.append(MetadataSpec('request_memory', basestring, 'Amount of memory (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run.', formatting='size'))
     METADATA_SPECS.append(MetadataSpec('request_disk', basestring, 'Amount of disk space (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run.', formatting='size'))
-    METADATA_SPECS.append(MetadataSpec('request_cpus', int, 'Number of CPUs allowed for this run.'))
-    METADATA_SPECS.append(MetadataSpec('request_gpus', int, 'Number of GPUs allowed for this run.'))
+    METADATA_SPECS.append(MetadataSpec('request_cpus', int, 'Number of CPUs allowed for this run.', default=1))
```
percyliang, line 30: Is this the place for defaults?  It seems strange that defaults would only be available for the cpu and gpu flags when some sort of defaults are of course in place for the other fields...

percyliang (approved):
Mostly looks good...just a few comments/questions.

Also, did we double check that there aren't any threading issues on the worker side in manipulating cpuset/gpuset?

codalab/bundles/run_bundle.py
fabeschan: All MetadataSpecs have a default default which is the default constructor for the spec's data type, unless overridden by a custom default value, which we need for request_cpus somewhere

codalab/worker/bundle_manager.py
fabeschan: because later on self._deduct_worker_resources(workers_list) will reduce a particular worker's worker['gpus'] to 0 even though that worker does have a gpu (just not free)

worker/codalabworker/main.py
fabeschan: oops you're right

worker/codalabworker/worker.py
fabeschan: _cpuset is in fact a set, and _cpuset_free is initialized as a copy of _cpuset, hence `set(cpuset)`. I will add a comment to clarify this

worker/codalabworker/worker.py
```diff
@@ -188,15 +206,53 @@ def _checkin(self):
...
+        Side effects: updated self._cpuset_free and self._gpuset_free
```
fabeschan: updates*

worker/codalabworker/worker.py
fabeschan: ok

worker/codalabworker/worker.py
fabeschan: ok

worker/codalabworker/main.py
fabeschan: ok

worker/codalabworker/main.py
fabeschan: ok

fabeschan:
Re: threading issues

I want a second or third pair of eyes to go through the threaded parts because I'm not too confident about my multithreading abilities

@bkgoksel could you take a look at this PR as well?

worker/codalabworker/worker.py
```diff
@@ -188,15 +209,62 @@ def _checkin(self):
...
             bundle_path = bundle['location']
         else:
             bundle_path = self._dependency_manager.get_run_path(bundle['uuid'])
+
+        success, cpuset, gpuset = self._allocate_cpu_and_gpu_sets(
+                resources['request_cpus'], resources['request_gpus'])
+
+        if not success: # revert self._cpuset_free and self._gpuset_free in-place
+            logger.debug('Unsuccessful allocation of cpu and gpu sets for bundle %s', bundle['uuid'])
+            self._deallocate_cpu_and_sets(cpuset, gpuset)
```
bkgoksel: Is it ever possible to get a case where `success` is `False` but `cpuset` and `gpuset` are non-empty?
It looks like that should not be the case but then why do we need this function?

The threading here LGTM otherwise.

bkgoksel (commented):
small question about deallocating Xpusets

worker/codalabworker/worker.py
fabeschan: Ya on second thoughts that `success` may not be necessary. It's probably a remnant of past iterations by now

worker/codalabworker/worker.py
```diff
@@ -188,15 +209,60 @@ def _checkin(self):
...
         if self.shared_file_system:
             bundle_path = bundle['location']
         else:
             bundle_path = self._dependency_manager.get_run_path(bundle['uuid'])
+
+        cpuset, gpuset = self._allocate_cpu_and_gpu_sets(
+                resources['request_cpus'], resources['request_gpus'])
+
+        if len(cpuset) == 0 and len(gpuset) == 0: # revert self._cpuset_free and self._gpuset_free in-place
+            logger.debug('Unsuccessful allocation of cpu and gpu sets for bundle %s', bundle['uuid'])
```
fabeschan, line 257: Likely unhandled case: what is/should be the bundle status at this point? Does it get failed automagically? Probably not.

Add a call to the server to fail the bundle here? <- is this good design?
Add a call to the server to restage the bundle here? <- is this good design?

Thoughts? @percyliang @bkgoksel

percyliang: Why would this happen?  Seems like we should work hard to make sure this doesn't happen, in which case, it doesn't matter what we do here, and the simplest thing is to just fail the bundle.

PR #870, opened 2018-03-02
Add support for referring to bundles from other worksheets using `worksheet//bundle` syntax

bkgoksel:


codalab/lib/bundle_cli.py
```diff
@@ -88,13 +88,15 @@
...
-GLOBAL_SPEC_FORMAT = "[<alias>::|<address>::](<uuid>|<name>)"
 ADDRESS_SPEC_FORMAT = "(<alias>|<address>)"
-TARGET_SPEC_FORMAT = '(<uuid>|<name>)[%s<subpath within bundle>]' % (os.sep,)
-ALIASED_TARGET_SPEC_FORMAT = '[<key>:]' + TARGET_SPEC_FORMAT
+BASIC_SPEC_FORMAT = '(<uuid>|<name>)'
 BUNDLE_SPEC_FORMAT = '(<uuid>|<name>|^<index>)'
-GLOBAL_BUNDLE_SPEC_FORMAT = '((<uuid>|<name>|^<index>)|(<alias>|<address>)::(<uuid>|<name>))'
+GLOBAL_SPEC_FORMAT = "[%s::]%s" % (ADDRESS_SPEC_FORMAT, BASIC_SPEC_FORMAT)
 WORKSHEET_SPEC_FORMAT = GLOBAL_SPEC_FORMAT
+REMOTE_BUNDLE_SPEC_FORMAT = '(%s//%s)' % (WORKSHEET_SPEC_FORMAT, BUNDLE_SPEC_FORMAT)
```
percyliang: Don't use the word "remote" - that suggests a different CodaLab instance - e.g., "main::foo".  I might even swap 'remote' and 'global', but probably there's something better.

codalab/lib/bundle_cli.py
```diff
@@ -497,26 +499,39 @@ def simple_group_str(info):
...
         # Resolve the bundle_spec to a particular bundle_uuid.
         bundle_uuid = BundleCLI.resolve_bundle_uuid(client, worksheet_uuid, bundle_spec)
         return (bundle_uuid, subpath)
 
     def parse_target_specs(self, items):
         targets = []
         for item in items:
-            if ':' in item:
-                (key, target) = item.split(':', 1)
+            if '::' in item:
```
percyliang: Add documentation to `parse_target_specs` about what `items` is.

percyliang: In particular, it's confusing that it's quite different than `parse_target` (it was confusing before you made your change, but now that the logic is getting redone, might as well make it clear what it is).

codalab/lib/cli_util.py
```diff
@@ -46,8 +48,13 @@ def desugar_command(orig_target_spec, command):
     val2key = {}  # e.g., a.txt => b1 (use first key)
 
     def get(dep):  # Return the key
-        if ':' in dep:  # :<bundle_spec> or <key>:<bundle_spec>
-            key, val = dep.split(':', 1)
+        if '::' in dep:
```
percyliang: Can we consolidate the spec parsing logic and factor this out with the `parse_target` and `parse_target_specs` methods?

percyliang (changes requested):
Functionality looks great.  Just comments about refactoring and making it clear.  For this type of thing, adding unit tests might be easier.

codalab/lib/bundle_cli.py
```diff
@@ -732,6 +732,7 @@ def do_command(self, argv, stdout=None, stderr=None):
             structured_result = command_fn()
         else:
             try:
+                """
```
percyliang: Why comment this out?

bkgoksel: Should we just remove it? Seems like this is the wrong way to have an optional profiler anyways.

codalab/lib/bundle_cli.py
```diff
@@ -495,46 +502,39 @@ def simple_group_str(info):
         return '%s(%s)' % (contents_str(info.get('name')), info['id'])
 
     @staticmethod
-    def parse_target(client, worksheet_uuid, target_spec):
+    def parse_target_bundle(client, worksheet_uuid, target_bundle_spec):
```
percyliang: Can we leave the name "target" instead of "target_bundle".  "target" is the technical term for bundle/path...

codalab/lib/bundle_cli.py
```diff
@@ -495,46 +502,39 @@ def simple_group_str(info):
...
         """
-        Helper: A target_spec is a bundle_spec[/subpath].
+        Helper: A target_spec is a [worksheet_spec//]bundle_spec[/subpath].
```
percyliang: Document what is returned.

codalab/lib/cli_util.py
```diff
@@ -24,19 +24,36 @@ def nested_dict_get(o, *args, **kwargs):
...
         for arg in args:
-            o = o[arg]
-        return o
+            obj = obj[arg]
+        return obj
     except (KeyError, TypeError):
         return default
 
+def parse_target_spec(spec):
+    key = ''
```
percyliang: Document what this function does (input / output example)

percyliang (approved):
Looks good - just add some more comments.

PR #875, opened 2018-03-07
Fix sleep dependent tests by infinite looping instead of sleeping once

bkgoksel:
These tests nondeterministically fail as the tests depend on getting to interact with bundles before they finish running, but an arbitrary sleep amount doesn't guarantee that. Changing them all to `while true; do sleep X; done` instead.

### DIFF
--- a/test-cli.py
+++ b/test-cli.py
@@ -855,7 +855,7 @@ def test(ctx):
     uuid = run_command([cl, 'run',
                         'dir:' + dep_uuid,
                         'file:' + dep_uuid + '/a.txt',
-                        'ls dir; cat file; seq 1 10; touch done; sleep 60'])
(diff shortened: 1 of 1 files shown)


fabeschan:
What is this test supposed to test? Why does making a finite sleep into an infinite sleep make sense?

bkgoksel:
I think they test various functionality around read and kill commands. The read test depends on the job still being running a little bit after starting it to check some things. Without the infinite loop sometimes the run successfully finishes running before then and the test fails. At other times the test gets to the check before the run finishes and it passes. 

The infinite loop ensures then job will run till we kill it and thus the test reliably reaches the job while it's running. There's no reason to allow our test to fail because the sleep ended before the test got there, so an infinite loop makes our test deterministic. 

I don't really know what these tests do, though. Just didn't want to remove them but still wanted to make our tests less flaky.

fabeschan (approved):

PR #877, opened 2018-03-08
Bugfix: Offline bundles should be removed from worker_run

### LINKED ISSUE #876: Offline bundles should be removed from worker_run

fabeschan:
Fixes #876

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -620,6 +620,7 @@ class BundleModel(object):
     def set_offline_bundle(self, bundle):
         """
         Sets the bundle to WORKER_OFFLINE, updating the last_updated metadata.
+        Remove the corresponding row from worker_run if it exists.
         """
         with self.engine.begin() as connection:
             # Check that it still exists.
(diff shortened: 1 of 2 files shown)


percyliang (commented):
Can you give more context about what problem this fixes?

fabeschan:
Previously, the only way an offline row in worker_run can be removed is when the worker that is assigned to it is removed (as a dead worker).

If the worker never gets detected as 'dead' by the server for whatever reason then the offline row stays in worker_run forever.

Also there was a bug in resume_bundle() where the last_updated timestamp is not updated if the bundle is in RUNNING state. This field should be regularly updated so that last_updated doesn't fall too far behind and get moved to WORKER_OFFLINE

percyliang (approved):

codalab/model/bundle_model.py
```diff
@@ -628,6 +629,10 @@ def set_offline_bundle(self, bundle):
                 # The user deleted the bundle.
                 return False
 
+            # Delete row in worker_run
```
percyliang, line 632: Is it documented somewhere what the semantics of something being in worker_run is?  This seems like tricky logic, and would be good to document how one should reason about the flow.

fabeschan: I don't recall seeing anything detailing worker_run

and to be honest I'm only making (hopefully educated) guesses to guide my changes

PR #878, opened 2018-03-09
Refactor key, target parsing and resolving logic

bkgoksel:
* Write a unified, regex-based, robust `key:target` parsing function in `cli_util`
* Write a unified, regex-based, robust `instance::worksheet//bundle/sub/path` parsing function in `cli_util`
* Refactor all target parsing cli methods in `bundle_cli` to use these methods when determining targets (including `cl add bundle`, enabling // syntax support for that method as well)
* Update `desugar_command` to work with these methods

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -544,7 +544,7 @@ class BundleCLI(object):
     def resolve_key_targets(self, client, worksheet_uuid, target_specs):
         """
         Helper: target_specs is a list of strings which are [<key>]:<target>
-        Returns: List<Tuple<key, Tuple<bundle_uuid, subpath>>>
+        Returns: [(key, (bundle_uuid, subpath)), ...]
         """
         targets = []
         target_keys_values = [parse_key_target(spec) for spec in target_specs]
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -59,8 +59,11 @@ from codalab.lib import (
 )
 from codalab.lib.cli_util import (
     nested_dict_get,
+    parse_key_target,
     parse_target_spec,
-    desugar_command
+    desugar_command,
+    INSTANCE_SEPARATOR,
+    WORKSHEET_SEPARATOR
 )
 from codalab.objects.permission import (
     group_permissions_str,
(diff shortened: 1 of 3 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -2091,11 +2091,16 @@ def do_add_command(self, args):
 
         elif args.item_type == 'bundle':
             for bundle_spec in args.item_spec:
-                source_client, source_spec = self.parse_spec(bundle_spec)
+                if '//' in bundle_spec:
+                    source_worksheet_spec, source_bundle_spec = tuple(bundle_spec.split('//', 1))
```
percyliang: `tuple` is not needed, right?

codalab/lib/bundle_cli.py
```diff
@@ -2091,11 +2091,16 @@ def do_add_command(self, args):
 
         elif args.item_type == 'bundle':
             for bundle_spec in args.item_spec:
-                source_client, source_spec = self.parse_spec(bundle_spec)
+                if '//' in bundle_spec:
```
percyliang: Can we consolidate this logic?  Lack of consolidation was the reason we didn't catch this the first time around...

codalab/lib/bundle_cli.py
bkgoksel: Yep, confused this with some other time where I needed a tuple around a split call. Removing

codalab/lib/bundle_cli.py
bkgoksel: Good point. Will refactor the general client/worksheet resolution and target parsing logic.

codalab/lib/bundle_cli.py
```diff
@@ -501,43 +504,58 @@ def simple_user_str(user):
...
         bundle_uuid = BundleCLI.resolve_bundle_uuid(client, worksheet_uuid, bundle_spec)
-        return (bundle_uuid, subpath)
 
-    def parse_key_targets(self, client, worksheet_uuid, target_specs):
+        return (client, worksheet_uuid, bundle_uuid, subpath)
+
+    def resolve_key_targets(self, client, worksheet_uuid, target_specs):
         """
         Helper: target_specs is a list of strings which are [<key>]:<target>
+        Returns: List<Tuple<key, Tuple<bundle_uuid, subpath>>>
```
percyliang: I'd just write `[(key, (bundle_uuid, subpath)), ...]` to make it easier to read.

percyliang (approved):
This is much nicer, thanks!

PR #884, opened 2018-03-11
Hotfix for new defaults system backwards compatibility

bkgoksel:
The new default resource requests system means that run bundles created under a version older than 0.2.24 without specified requested resources don't run under  0.2.24 anymore. 

This hotfix brings back bundle manager-side defaults so that we can deprecate the old style bundles at a later time.

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -388,30 +388,45 @@ class BundleManager(object):
         """
         Compute the CPU limit used for scheduling the run.
         """
+        #TODO: Remove this once we want to deprecate old versions
+        if bundle.metadata.request_cpus is None:
+            return 1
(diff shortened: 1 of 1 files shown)


percyliang (approved):
Can we print out a deprecated message and ask people to upgrade?

bkgoksel:
This is more of an internal thing (as Fabian and I discussed on slack), so there's no action needed on behalf of users to prevent this from happening again (i.e. even if they don't upgrade their CLIs or keep running jobs however they did in the past, the runs will be fine). Once the existing staged runs are complete, this could not be an issue again. So I don't think we need to print a deprecated message in this case.

bkgoksel:
Merging this for now so things start working again. If we want to make any other changes can make another hotfix.

PR #889, opened 2018-03-14
Avoid keyless dependencies

bkgoksel:


fabeschan (approved):

codalab/lib/bundle_cli.py
```diff
@@ -558,6 +558,9 @@ def resolve_key_targets(self, client, worksheet_uuid, target_specs):
                 else:
                     raise UsageError('Must specify keys when packaging multiple targets!')
             _, worksheet_uuid, bundle_uuid, subpath = self.resolve_target(client, worksheet_uuid, target_spec, allow_remote=False)
+            if not key:
+                instance, worksheet_spec, bundle_spec, subpath = parse_target_spec(target_spec)
```
percyliang: Seems like it's non-ideal to have this call again, and that `resolve_target` should really return both the `bundle_uuid` and the `bundle_spec` (in a dict).

percyliang (approved):
Ok to merge if urgent.

codalab/lib/bundle_cli.py
bkgoksel: No need for this (both the extra call and having both uuid and spec) with the new change below

PR #894, opened 2018-03-28
Remove worker checkout

fabeschan:
Since workers now get automagically cleaned up after not checking in with the server for 30 seconds, code paths for worker checkout are no longer necessary.


percyliang (approved):
Could you add a comment to the PR why this is an okay thing to do now?

PR #896, opened 2018-04-02
Worker 2.0

fabeschan:
This is a realization of the worker system revamp. This PR is currently deployed onto the test instance.

- [x] Make checkins gr8 again

- [x] (Json)StateCommitter class for committing state

- [ ] Rewrite/fix unit tests

- [x] DependencyManager and DockerImageManager

- [x] Remove janky Run-to-DependencyManager reach arounds

- [x] Worker "runs" are now defined in terms of stages (a 'stage' machine)

- [x] Remove REST endpoint for resume_bundle

- [x] Decrease worker wait time to 3.0s

- [ ] Clean up threads for DependencyManager and DockerImageManager 

- [ ] Add kill bundle support

- [ ] Be able to remove dependencies

- [ ] Implement _netcat, _write, _upgrade

- [x] Workers report full memory instead of memory after deductions from running bundles


codalab/rest/workers.py
```diff
@@ -23,7 +23,7 @@ def checkin(worker_id):
     Waits for a message for the worker for WAIT_TIME_SECS seconds. Returns the
     message or None if there isn't one.
     """
-    WAIT_TIME_SECS = 5.0
+    WAIT_TIME_SECS = 3.0
```
bkgoksel, line 26: What's the reasoning for this change?

codalab/model/bundle_model.py
```diff
@@ -731,13 +730,14 @@ def resume_bundle(self, bundle, user_id, worker_id, hostname, start_time):
...
                 connection.execute(cl_worker_run.insert().values(worker_run_row))
 
-            bundle_update = {
-                'state': State.RUNNING,
-                'metadata': {
-                    'last_updated': int(time.time()),
-                },
-            }
-            self.update_bundle(bundle, bundle_update, connection)
+            if row.state == State.RUNNING or row.state == State.WORKER_OFFLINE:
```
bkgoksel, line 733: What was the problem when this if condition wasn't here, restarting FAILED runs?

codalab/rest/workers.py
```diff
@@ -36,6 +36,24 @@ def checkin(worker_id):
...
         request.json['dependencies'])
+
+    for uuid, run in request.json['runs'].items():
+        bundle = local.model.get_bundle(uuid)
+        local.model.resume_bundle(bundle, request.user.user_id, worker_id,
+                                request.json['hostname'], run['start_time'])
+
+        # this is a hack (or maybe not?)
+        # TODO: unhack this
+        metadata_update = {
```
bkgoksel, line 47: What's the hack here? Updating all the runs' metadata with each checkin?

worker/codalabworker/dependency_manager.py
```diff
@@ -1,298 +1,206 @@
...
+
+    def _process_dependencies(self):
+        with synchronized(self):
+            for entry in self._dependencies.keys():
+                dependency_state = self._dependencies[entry]
+                self._dependencies[entry] = self._transition_dependency_state(dependency_state)
+
+    def _transition_dependency_state(self, dependency_state):
+        stage = dependency_state.stage.upper()
+        return getattr(self, '_transition_dependency_state_from_' + stage)(dependency_state)
```
bkgoksel: This kind of string manipulation to make the correct call seems a bit hacky to me. Maybe we can find a more transparent/explicit way of doing this? Like a dict from state names to functions?

bkgoksel: I think this also needs more documentation as to what `transition_dependency_state_from_X` means and what this function itself does.

worker/codalabworker/docker_image_manager.py
```diff
@@ -6,121 +7,128 @@
...
+
+    def _process_images(self):
+        with synchronized(self):
+            for entry in self._images.keys():
+                image_state = self._images[entry]
+                self._images[entry] = self._transition_image_state(image_state)
+
+    def _transition_image_state(self, image_state):
+        stage = image_state.stage.upper()
+        return getattr(self, '_transition_image_state_from_' + stage)(image_state)
```
bkgoksel: Same comments as in dependency manager. Also we have some code duplication here. I wonder if we could deduplicate the manager logic somehow between docker images and dependencies

worker/codalabworker/fsm.py
```diff
@@ -0,0 +1,75 @@
+import os
```
bkgoksel, line 1: The name of this file is non-descriptive, is it file state manager or something like that? It should have a full name.

worker/codalabworker/main.py
```diff
@@ -110,11 +113,21 @@ def main():
...
+    dependency_manager = LocalFileSystemDependencyManager(
+            JsonStateCommitter(os.path.join(args.work_dir, 'dependencies-state.json')),
+            bundle_service, args.work_dir, max_work_dir_size_bytes, args.max_dependencies_serialized_length)
+
+    image_manager = DockerImageManager(
+            docker_client, JsonStateCommitter(os.path.join(args.work_dir, 'images-state.json')),
+            max_images_bytes
+    )
+
+    state_committer = JsonStateCommitter(os.path.join(args.work_dir, 'worker-state.json'))
```
bkgoksel: Small nit, we might want to be uniform in where we initialize the state managers (within constructor or outside) for all 3 components.

worker/codalabworker/pyjson.py
```diff
@@ -0,0 +1,61 @@
+import json
```
bkgoksel, line 1: I wonder if we could release this as a standalone pypi package instead of bundling it in codalab since it looks really standalone. Also wonder if there are other solutions that enable JSON serialization of sets/namedtuples.

worker/codalabworker/synchronized.py
```diff
@@ -0,0 +1,100 @@
+"""
```
bkgoksel: Similar to PyJSON, this seems pretty standalone. I like the layer of abstraction we get from this, but this class needs more documentation as it also makes our concurrency a bit more opaque. (And maybe even comments in places where we use it?)

worker/codalabworker/worker.py
```diff
@@ -30,111 +49,73 @@
...
 
-                # resume previous runs once in the beginning, but after checkin
-                # this is not an ideal design because initial checkin can assign worker with new runs
-                # but resources from previous runs are not re-allocated to the worker yet; this can cause
-                # performance problems
-                if not resumed_prev_runs:
-                    resume_previous_runs()
-                    resumed_prev_runs = True
-                self._worker_state_manager.save_state()
                 if not self._last_checkin_successful:
```
bkgoksel, line 124: Maybe let's change the name of this variable to `first_checkin` or something? As it stands this code block makes it seem like we declare successful check in when the check-in failed.

worker/codalabworker/worker.py
```diff
@@ -144,29 +125,20 @@ def resume_previous_runs():
...
-            return self._exiting
-
-    def _should_run(self):
-        if not self._is_exiting():
-            return True
-        return self._worker_state_manager.has_runs()
+        self._image_manager.stop()
+        self._dependency_manager.stop()
+
+    def _get_runs_for_checkin(self):
```
bkgoksel, line 136: So we are going with checkin all runs for now?

worker/codalabworker/worker.py
```diff
@@ -175,26 +147,30 @@ def _get_installed_memory_bytes(self):
             # Fallback to sysctl when os.sysconf('SC_PHYS_PAGES') fails on OS X
             return int(check_output(['sysctl', '-n', 'hw.memsize']).strip())
 
-    def _get_allocated_memory_bytes(self):
-        return sum(self._worker_state_manager.map_runs(lambda run: run.requested_memory_bytes))
-
     def _get_memory_bytes(self):
-        return max(0, self._get_installed_memory_bytes() - self._get_allocated_memory_bytes())
+        return self._get_installed_memory_bytes()
```
bkgoksel, line 156: What is the reason for this change?

worker/codalabworker/worker.py
```diff
@@ -218,158 +194,187 @@ def _checkin(self):
                     self._exiting = True
                 self._should_upgrade = True
 
-    def _allocate_cpu_and_gpu_sets(self, request_cpus, request_gpus):
+    def _execute_bundle_service_command_with_retry(self, f):
+        # Retry for 6 hours before giving up.
+        retries_left = 2 * 60 * 6
```
bkgoksel, line 204: Let's make this a constant instead of a magic number.

worker/codalabworker/worker.py
```diff
@@ -218,158 +194,187 @@ def _checkin(self):
...
-            for j in range(request_gpus):
-                gpuset.add(self._gpuset_free.pop())
-            return cpuset, gpuset
+        if len(cpuset) < request_cpus or len(gpuset) < request_gpus:
+            raise Exception("Not enough cpus or gpus to assign!")
 
-    def _deallocate_cpu_and_sets(self, cpuset, gpuset):
-        """
-        Release held up cpus and gpus
+        def propose_set(resource_set, request_count):
```
bkgoksel, line 243: nit: maybe add a small comment to say this returns a subset of the first n elements of the set

worker/codalabworker/worker.py
```diff
@@ -218,158 +194,187 @@ def _checkin(self):
...
+                    cpuset=None, gpuset=None, info={},
+            )
 
-        if len(cpuset) == 0 and len(gpuset) == 0: # revert self._cpuset_free and self._gpuset_free in-place
-            logger.debug('Unsuccessful allocation of cpu and gpu sets for bundle %s', bundle['uuid'])
-            return
+            with synchronized(self):
+                self._runs[bundle_uuid] = run_state
+        else:
+            print >>sys.stdout, 'Bundle {} no longer assigned to this worker'.format(bundle_uuid)
```
bkgoksel: Just out of curiosity, what would be a case where a run's assigned worker changes? Does this also include the case where the run is killed server-side?

worker/codalabworker/worker.py
```diff
@@ -218,158 +194,187 @@ def _checkin(self):
...
+                    'error_code': code,
+                    'error_message': message,
+                }
+                self._bundle_service.reply(self.id, socket_id, message)
+
+            dep_paths = set([dep['child_path'] for dep in run_state.bundle['dependencies']])
+            bundle_uuid = run_state.bundle['uuid']
+            try:
+                read_type = read_args['type']
+                if read_type == 'get_target_info':
```
bkgoksel, line 300: Could we decompose this a little bit, maybe with sub-functions for different read types?

worker/codalabworker/worker.py
```diff
@@ -385,3 +390,226 @@ def _upgrade(self):
...
+            # transition all runs
+            for bundle_uuid in self._runs.keys():
+                run_state = self._runs[bundle_uuid]
+                self._runs[bundle_uuid] = self._transition_run_state(run_state)
+
+            # filter out finished runs
+            self._runs = {k: v for k, v in self._runs.items() if v.stage != RunStage.FINISHED}
+
+    def _transition_run_state(self, run_state):
+        stage = run_state.stage.upper()
```
bkgoksel: Ditto earlier comment about maybe using a dict here instead of string manipulation

worker/codalabworker/worker.py
```diff
@@ -385,3 +390,226 @@ def _upgrade(self):
...
+                status_message = 'Downloading dependency %s: %s done (archived size)' % (
+                            dep['child_path'], size_str(dependency_state.size_bytes))
+                return run_state._replace(run_status=status_message)
+            elif dependency_state.stage == DependencyStage.FAILED:
+                # Failed to download dependency; -> FINALIZING
+                run_state.info['failure_message'] = 'Failed to download dependency %s: %s' % (
+                        dep['child_path'], '') #TODO: get more specific message
+                return run_state._replace(stage=RunStage.FINALIZING, info=run_state.info)
+
+        docker_image = run_state.resources['docker_image']
```
bkgoksel, line 439: Can we start the downloads of dependencies and docker images in parallel?

worker/codalabworker/worker.py
```diff
@@ -385,3 +390,226 @@ def _upgrade(self):
...
+
+        # 4) Start container
+        container_id = self._docker.start_container(
+            run_state.bundle_path, bundle_uuid, run_state.bundle['command'],
+            run_state.resources['docker_image'], docker_network, dependencies,
+            cpuset, gpuset, run_state.resources['request_memory']
+        )
+
+        digest = self._docker.get_image_repo_digest(run_state.resources['docker_image'])
+        #self._image_manager.touch_image(digest)
```
bkgoksel: Why is this commented out?

worker/codalabworker/worker.py
```diff
@@ -385,3 +390,226 @@ def _upgrade(self):
...
+                if run_state.container_id is not None:
+                    while True:
+                        try:
+                            finished, _, _ = self._docker.check_finished(run_state.container_id)
+                            if finished:
+                                self._docker.delete_container(run_state.container_id)
+                                break
+                        except DockerException:
+                            traceback.print_exc()
+                            time.sleep(1)
```
bkgoksel, line 526: Do we have a risk of infinite loops here? Are exceptions guaranteed to go away eventually?

worker/codalabworker/worker.py
```diff
@@ -385,3 +390,226 @@ def _upgrade(self):
...
+
+                # Clean-up dependencies.
+                for dep in run_state.bundle['dependencies']:
+                    #self._dependency_manager.remove_dependency(
```
bkgoksel, line 530: Why are these commented out?

bkgoksel (changes requested):
Some nits and questions here and there, but overall it reflects our design well I think. Maybe we should loop Josh from Semantic Machines in. Also would be good to add easily runnable tests to test this thoroughly.

codalab/model/bundle_model.py
fabeschan: issue was that it could restart “Done” runs after they get finalized

codalab/rest/workers.py
fabeschan: faster iterations. Given each run goes through 4+ stages, each stage requires a check in so the minimum turn around would be 4 * waittime = 20s if waittime=5 or 12s if waittime=3

codalab/rest/workers.py
fabeschan: just feels uneasy calling resume bundle and metadata updates for each bundle. might not be as terrible as i thought though, possibly just requires some renaming of functions

worker/codalabworker/dependency_manager.py
fabeschan: documentation: see DependencyStage

worker/codalabworker/docker_image_manager.py
fabeschan: ya would be good

worker/codalabworker/fsm.py
fabeschan: initially meant to be finite state/stage system. might not be to fitting anymore

worker/codalabworker/synchronized.py
fabeschan: on second thoughts, we haven’t been using this module anywhere near its potential. in fact we can just replace every synchronized call with a reentrant lock.

worker/codalabworker/worker.py
fabeschan: i am, not sure if we are. that’s up to you guys

codalab/rest/workers.py
bkgoksel: Makes sense.

worker/codalabworker/worker.py
fabeschan: we dont report the leftover cpuset and gpusets of a worker; but we report the leftover memory; this change is made so that the worker reports its total memory, like cpuset and gpusets

worker/codalabworker/dependency_manager.py
bkgoksel: Yeah found that later while reading the PR. What do you think about making a state->fn dict instead of `getattr`?

worker/codalabworker/worker.py
fabeschan: not sure, but this is in the original code. I am simply making it more prominent when it occurs

worker/codalabworker/synchronized.py
bkgoksel: I think that might be better for future maintainability.

worker/codalabworker/worker.py
fabeschan: ya prob a good idea

worker/codalabworker/worker.py
fabeschan: ya I had them download in parallel before but unknowingly threw that away when i decided to return stuff early to make code cleaner. should revert this function back to what it was before

worker/codalabworker/worker.py
fabeschan: because this function has not yet been (properly) implemented

worker/codalabworker/worker.py
fabeschan: not sure; i copied this from current code base

worker/codalabworker/worker.py
fabeschan: because this method has not yet been properly implemented

worker/codalabworker/dependency_manager.py
fabeschan: sure

worker/codalabworker/fsm.py
fabeschan: s/to/so

worker/codalabworker/synchronized.py
fabeschan: sounds good

worker/codalabworker/worker.py
fabeschan: first_checkin doesn’t fit because we want to print out that message whenever there was a change from checkin failure to checkin success

worker/codalabworker/worker.py
fabeschan: fixed

bkgoksel:
Finished with the merge of #909

Issue #866, opened 2018-02-23
Re-adding a group owner/admin to their group strips their admin status


bkgoksel:
Currently if `uadd <user> <group>` is called when `<user>` is an admin or owner of `<group>`, `<user>` loses all admin privileges on the group. 

This can cause the following situations:
* Group owner unknowingly calls `uadd <themselves> <group>`, stripping themselves of their admin rights, leaving the group adminless (cannot be modified by any user except root).
* Group admin calls `uadd <owner> <group>` to strip the owner (or any other admin) of admin privileges, gaining control of the entire group.

The first scenario is exacerbated by the fact that `ginfo` still lists the owner as `owner`, but they silently lose their admin rights. (Already happened in multiple cases)

bkgoksel:
Fixed by #902

PR #902, opened 2018-04-11
Don't allow the relegation of a group owner to non-admin status

bkgoksel:
Currently this is allowed (by any admin calling `uadd` with the group owner's user spec, detailed in #866 ), this should disable that, even by the group owner itself.

Also adds a test for this (and adds groups as a cleanable resource to the test harness)

This is the most egregious group permissioning  issue we have at the moment. In the future we might want to think about separating the `uadd` semantics so it only adds new members and move the admin status change to a different method.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2747,6 +2747,7 @@ class BundleCLI(object):
         ],
         arguments=(
             Commands.Argument('user_spec', nargs='?', help='Username or id of user to show [default: the authenticated user]'),
+            Commands.Argument('-f', '--field', help='Print out these comma-separated fields.'),
         ),
     )
     def do_uinfo_command(self, args):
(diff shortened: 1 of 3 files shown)


test-cli.py
```diff
@@ -75,6 +75,26 @@ def current_worksheet():
     assert m is not None
     return m.group(1)
 
+def current_user():
+    """
+    Return the uuid and username of the current username in a tuple
```
percyliang: nit: current user?

test-cli.py
```diff
@@ -75,6 +75,26 @@ def current_worksheet():
...
+    Does so by parsing the output of `cl uinfo` which by default returns the info
+    of the current user
+    """
+    my_name = run_command([cl, 'uinfo'])
```
percyliang: This is kind of gross...can we add flags to `cl uinfo` to make it return userid or username (kind of how `cl info -f` works) isntead?  In general, it'd be nice to have `cl` commands that have a mode that outputs JSON instead of text, so actually it might be easier to do that.

test-cli.py
```diff
@@ -1125,6 +1154,30 @@ def test(ctx):
...
+    # Check that you are added to your own group
+    group_info = run_command([cl, 'ginfo', group_name])
+    check_contains(user_name, group_info)
+    my_groups = run_command([cl, 'gls'])
+    check_contains(group_name, my_groups)
+
+    # Try to relegate yourself to non-admin status
+    run_command([cl, 'uadd', user_name, group_name], expected_exit_code=1)
+
+    # TODO: Test other group membeship semantics:
```
percyliang: typo

test-cli.py
bkgoksel: I guess this goes back to whether we want to make something like a Python API, or maybe open up the REST API more directly to the users. For now I'd prefer to add the `-f` field to `uinfo` until we decide if/how we want to make programmatic codalab usage easier

percyliang: Sounds good.

codalab/lib/bundle_cli.py
```diff
@@ -2758,34 +2759,42 @@ def do_uinfo_command(self, args):
...
-                    'affiliation', 'url', 'date_joined'):
-            print_attribute(key, user.get(key, None))
+        default_fields = ('id', 'user_name', 'first_name', 'last_name',
+                    'affiliation', 'url', 'date_joined', 'last_login',
+                    'email', 'time', 'disk')
+
+        fields = fields.split(',') if fields else default_fields
+
+        for field in fields:
+            print_attribute(field, user)
```
percyliang: There should be two modes: one is pretty printing all the attributes, and the other is printing out individual fields (which should not have the field names).  Just match the behavior of `cl info -f`.

percyliang (approved):

PR #903, opened 2018-04-11
Better test coverage for target parsing and runs

bkgoksel:
Unit tests for key target spec parsing and couple more run integration tests.
To prevent cases where we break basic run syntax in the future.


percyliang (approved):
Thanks!

PR #906, opened 2018-04-26
Delete user

w0nche0l:
https://github.com/codalab/codalab-cli/issues/714

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1833,6 +1833,66 @@ class BundleModel(object):
 
     def delete_user(self, user_id=None):
         with self.engine.begin() as connection:
+
+            # ==User verification==
+
+            connection.execute(cl_user_verification.delete().where(
+                cl_user_verification.c.user_id == user_id
+            ))
+
+            connection.execute(cl_user_reset_code.delete().where(
+                cl_user_reset_code.c.user_id == user_id
+            ))
+
+            # ==OAuth2==
+
+            connection.execute(oauth2_auth_code.delete().where(
+                oauth2_auth_code.c.user_id == user_id
+            ))
+
+            connection.execute(oauth2_token.delete().where(
+                oauth2_token.c.user_id == user_id
+            ))
+
+            connection.execute(oauth2_client.delete().where(
+                oauth2_client.c.user_id == user_id
+            ))
+
+            # ==Workers==
+
+            connection.execute(cl_worker_run.delete().where(
+                cl_worker_run.c.user_id == user_id
+            ))
+
+            # ==User Groups==
+
+            connection.execute(cl_user_group.delete().where(
+                cl_user_group.c.user_id == user_id,
+            ))
+
+            # ==Bundle==
+
+            # TODO: Get bundles owned by user and call delete_bundles on those
+
+            # ==Worksheet==
+
+            # TODO: Get worksheets owned by user and call delete_bundles on those
+
+            # ==Group==
+
+            # TODO: Get groups owned by user and call delete_bundles on those
+
+            # ==Event==
+            connection.execute(cl_event.delete().where(
+                cl_event.c.user_id == user_id,
+            ))
+
+            # ==Chat==
+            connection.execute(cl_chat.delete().where(
+                cl_chat.c.sender_user_id == user_id or
+                cl_chat.c.recipient_user_id == user_id
+            ))
+
             connection.execute(cl_user.delete().where(
                 cl_user.c.user_id == user_id,
             ))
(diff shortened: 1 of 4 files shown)


codalab/rest/users.py
```diff
@@ -73,6 +76,56 @@ def fetch_user(user_spec):
...
+            error_message += "User %s owns groups, can't delete. UUIDs: %s \n" % \
+                (user_id, ",".join(group_uuids))
+
+        if error_message is not "":
+            abort(httplib.NOT_FOUND, error_message)
+
+        local.model.delete_user(user_id=user.user_id)
+
+    # Return list of deleted id as meta
+    abort(httplib.NO_CONTENT)
```
percyliang: Huh?  I'm confused how the comment and code connect.

codalab/rest/users.py
```diff
@@ -73,6 +76,56 @@ def fetch_user(user_spec):
...
+            abort(httplib.UNAUTHORIZED, "Cannot delete root user.")
+        user = local.model.get_user(user_id=user_id)
+        if user is None:
+            abort(httplib.NOT_FOUND, "User %s not found" % user_id)
+
+        """
+        Check for owned bundles, worksheets, and groups.
+        If any are found, then do not allow user to be deleted.
+        """
+        error_message = ""
```
percyliang: nit: the convention that we try to use in the codebase is single quotes, so please use that everywhere.

codalab/rest/users.py
```diff
@@ -73,6 +76,56 @@ def fetch_user(user_spec):
...
+            error_message += "User %s owns worksheets, can't delete. UUIDs: %s \n" % \
+                (user_id, ",".join(worksheet_uuids))
+
+        groups = local.model.batch_get_groups(owner_id=user_id)
+        if groups is not None and len(groups) > 0:
+            group_uuids = [group.uuid for group in groups]
+            error_message += "User %s owns groups, can't delete. UUIDs: %s \n" % \
+                (user_id, ",".join(group_uuids))
+
+        if error_message is not "":
```
percyliang: `!= ''`

codalab/rest/users.py
```diff
@@ -73,6 +76,56 @@ def fetch_user(user_spec):
...
+        """
+        Check for owned bundles, worksheets, and groups.
+        If any are found, then do not allow user to be deleted.
+        """
+        error_message = ""
+
+        bundles = local.model.batch_get_bundles(owner_id=user_id)
+        if bundles is not None and len(bundles) > 0:
+            bundle_uuids = [bundle.uuid for bundle in bundles]
+            error_message += "User %s owns bundles, can't delete. UUIDs: %s \n" % \
```
percyliang: clarify: "can't delete user"

Remove extra space before newline.

I'd also build up a list of error messages and then join with '\n' at the end - which is semantically better.

codalab/model/bundle_model.py
```diff
@@ -1831,6 +1831,60 @@ def add_user(self, username, email, first_name, last_name, password,
 
         return user_id, verification_key
 
+    def delete_user(self, user_id=None):
+        with self.engine.begin() as connection:
+
+            # ==User verification==
```
percyliang: My preference would be to make each `execute` command one line and remove all the newlines between statements in the same logical grouping...  In general, putting everything on one line is a bad idea and hard to read, but in this case, I think it makes it easier to see the overall structure, which is harder to get right than the individual details.  (I'm really worried about adding a table somewhere and forgetting to update this.)

codalab/rest/users.py
```diff
@@ -73,6 +76,56 @@ def fetch_user(user_spec):
...
+                (user_id, ",".join(bundle_uuids))
+
+        worksheets = local.model.batch_get_worksheets(fetch_items=False, owner_id=user_id)
+        if worksheets is not None and len(worksheets) > 0:
+            worksheet_uuids = [worksheet.uuid for worksheet in worksheets]
+            error_message += "User %s owns worksheets, can't delete. UUIDs: %s \n" % \
+                (user_id, ",".join(worksheet_uuids))
+
+        groups = local.model.batch_get_groups(owner_id=user_id)
```
bkgoksel: What is our plan for the groups that the user is a member of? We should make sure to test that case and not break those groups.

codalab/rest/users.py
bkgoksel: We should probably return the list of deleted users' ids as the comment says just to do the right thing.

codalab/rest/users.py
w0nche0l: So if the user is just a member of a group, then the deletion code in bundle_model.py will remove them. If the user is an owner of a group, then they will not be deleted because of this check.

bkgoksel:
LGTM.

percyliang (approved):

PR #908, opened 2018-04-30
Fix autocomplete

w0nche0l:
Before: Commands that used the TargetsCompleter class for autocompletion (run, docker, mount, cat), did not have functioning autocompletion. 
So if you had a bundle in your worksheet called "bundle-name", and typed in "cl run bundl" and hit tab, nothing would happen. 
After: Autocompletion works as expected for the aforementioned commands. So hitting tab for "cl run bundl" will complete it to "cl run bundle-name"

### DIFF
--- a/codalab/lib/completers.py
+++ b/codalab/lib/completers.py
@@ -116,7 +116,7 @@ def UnionCompleter(*completers):
 # TODO: https://github.com/codalab/codalab-cli/issues/223
 class TargetsCompleter(CodaLabCompleter):
     def __call__(self, prefix, action=None, parsed_args=None):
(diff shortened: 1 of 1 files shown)


bkgoksel:
What behavior does this fix? Did autocomplete not work at all before this and now it works as much as it did before?

w0nche0l:
The latter; because the argument allow_incomplete is not defined for parse_key_target, any autocomplete using TargetsCompleter didn't work.

percyliang (commented):
In the PR description, could you describe the "Before" and "After" behavior (with a concrete example)?

Issue #893, opened 2018-03-26
jobs take a long time to go from created to staged to running


percyliang:
Before, this was just instantaneous.  Now, it takes up to 30 seconds for even simple things like `cl run date`.

bkgoksel:
For bookkeeping, this seems to be a prod-only issue. 
Since prod and dev are running the same version of codalab, this might be slightly annoying to debug, something in our bundle manager iteration is taking considerably longer on prod than dev. My main hunch right now is the number of workers since that's the main difference between prod and dev. Still investigating

bkgoksel:
Looks like this is fixed with the new worker.

Issue #882, opened 2018-03-09
Unable to reserve worker resources for resumed runs


fabeschan:
When a worker starts up, it tries to do two things: checkin with the server, and resume previous runs.

There exists a conflict between the two actions:
If the checkin happens first, then the server bundle manager might assign new run bundles to the worker without any regard for the resources taken up by those previous runs, because the rows for them have not yet been inserted into `worker_run`.

If resume previous runs happens first, then the server throws an exception because resuming previous runs involves inserting rows into `worker_run`; but these rows reference a worker that hasn't checked in. This is a foreign reference violation.

bkgoksel:
Hopefully will be fixed with #896

bkgoksel:
Solved in the PR as of now

Issue #881, opened 2018-03-09
Worker Run() objects are not resumable


fabeschan:
Runs can be resumed if the worker dies while the container is running, but if the Run object is interrupted at any other point, the worker will be unable to recover it. This is the reason why a worker that goes offline while one of its bundles are "uploading results" is unrecoverable.

bkgoksel:
Hopefully will be fixed with #896

bkgoksel:
Solved in PR for #909

Issue #880, opened 2018-03-09
Worker disk space getting full


fabeschan:
DependencyManager and/or ImageManager isn't cleaning up disk space as well as it should. `/mnt` is showing 100% used on `prod-worker0`

bkgoksel:
Closing for now as new worked update is merged. I manually cleaned all old images on 8/3/2018.
If we see that we're still not cleaning up properly can reopen.
One thing to look at in the future might be to limit docker log size so old logs are discarded.

PR #909, opened 2018-05-07
The Great Worker Reconciliation

bkgoksel:
Tries to unify [SemanticMachines' AWSBatch Worker changes](https://github.com/semanticmachines/codalab-cli/tree/aws_batch) and [Fabian's Worker 2.0 PR](https://github.com/codalab/codalab-cli/tree/workers/)

Consult the [Google Doc](https://docs.google.com/document/d/1Fff22YIrzBe1KCNUzrraC9WA93AFITTDNdTQLtAzNWc/edit?usp=sharing) for the new abstractions.

Still a lot to be done:

- [x] Write the base classes (RunManager, RunState, Reader)
- [x] Update `worker.py` code to use the new interface
- [x] Update worker's `main.py` to initialize the new classes correctly
- [x] Write the `LocalRunManager` class
    - [x] Implement read
    - [x] Implement write
    - [x] Implement netcat
    - [x] Implement resource allocation
- [x] Testing for `LocalRunManager`
    - [x] Local testing to get runs
        - [x] Fix the `kill` issue
    - [x] Get docker setup working (currently broken)
    - [x] Deeper functionality tests
        - [x] Killing (worker|run) during (dependency|image) downloads, check what happens to dependencies
        - [x] Test successful resumes when worker stopped during all the run phases (#881)
        - [x] Stop server/worker mid run, restart
        - [x] Large job result uploads
        - [x] Stop worker, have run go worker_offline, delete run server-side, bring worker back up
        - [x] Multiple runs started around the same time having the same dependency
        - [x] Assign max CPU runs to worker,  stop it after runs start, stage more runs server side, reconnect worker, ensure old runs are resumed correctly (#882).
    - [ ] Unit tests
- [ ] Fix all the worker related issues as part of the epic
    - [x] #880 
        - Get dependency/image cleanups working
    - [ ] #893 
        - Investigate the potential reason/experiment with new worker code
    - [x] #881 
    - [x] #882
    - [ ] [#398](https://github.com/codalab/codalab-worksheets/issues/398)
        - Timeouts for each stage
- [ ] ~~The meaning of a run state is ambiguous. There seems to be states that are internal to the execution environment and then the user-facing server side states. How these map to each other is very unclear in the current implementation. How do we fix this?~~
- [x] ~~Think about AWS Batch integration into this API~~

Small issues from testing:
- [x] Documentation
- [ ] ~~Figure out a better abstraction for the threaded state transitions~~
- [x] Fix the scratch folder routing
- [x] Fix underscore run_manager naming
- [x] Go back to `namedtuple` `RunState`s and erase the unused class
- [x] Figure out the `socket_id` issue for bundle actions like `kill`
- [x] Move run cleanup to a more global state transition from uploading

Some questions:
- [ ] Do we need to recheck run_state in the individual action fn's, after the worker acquired it (i.e. does it always need to be in a lock? If that's the case what's the best way of abstracting this)


worker/codalabworker/fsm.py
```diff
@@ -0,0 +1,95 @@
...
+    # if thread still alive -> DOWNLOADING else -> READY
+    DOWNLOADING = 'DOWNLOADING'
+
+    # -> READY
+    READY = 'READY'
+
+    # -> FAILED
+    FAILED = 'FAILED'
+
+class StateTransitioner(object):
```
fabeschan: add comments to explain what a transition function is and what a check funciton is

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -0,0 +1,239 @@
...
+
+class LocalRunManager(BaseRunManager):
+    """
+    LocalRunManager executes the runs locally, each one in its own Docker
+    container. It manages its cache of local Docker images and its own local
+    Docker network.
+    """
+    def __init__(self, worker, docker, image_manager, dependency_manager,
+            state_committer, bundles_dir, cpuset, gpuset, docker_network_prefix='codalab_worker_network'):
+        self._worker = worker
```
fabeschan: why are some members prepended with `_` while others aren't?

worker/codalabworker/run_manager.py
```diff
@@ -0,0 +1,173 @@
...
+        raise NotImplementedError
+
+    def summarize_file(self, run_state, path, dep_paths, args, reply_fn):
+        """
+        Calls reply_fn(err, msg, data) with the file summary of the path
+        in the data field, or with err if there is an error
+        """
+        raise NotImplementedError
+
+class RunState(object):
```
fabeschan: does this get serialized to json properly?

fabeschan:
local dependency manager also needs to handle the concept of child dependencies

fabeschan:
think about ways to unit test these new additions

worker/codalabworker/fsm.py
bkgoksel: Good point.

worker/codalabworker/local_run/local_run_manager.py
bkgoksel: Artifact of something I was trying earlier dealing with access to these from a member class, will change back to consistent.

worker/codalabworker/run_manager.py
bkgoksel: Good point. The reason I did this was thinking that different architectures might want to store different things in RunStates, but we might want all RunStates to have a base minimum set of fields, I was hoping the Class would make it more extensible. I have to look into serialization. Maybe having it inherit from Dict instead of object could help.

worker/codalabworker/fsm.py
```diff
@@ -0,0 +1,96 @@
...
+class BaseStateCommitter(object):
+    def load(self):
+        """ Load and return the state """
+        raise NotImplementedError
+
+    def commit(self, state):
+        """ Commit the state """
+        raise NotImplementedError
+
+class JsonStateCommitter(BaseStateCommitter):
```
fabeschan: what if the state file format needs to change? how will you handle upgrades aka schema migrations

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,216 @@
...
+        self._dependencies = {}
+        self._downloading = {}
+        self._load_state()
+
+        self._stop = False
+        self._cleanup_sleep_secs = 10
+        self._main_thread = None
+
+    def _save_state(self):
+        with self._lock:
```
fabeschan: add comment to explain why this mangling is necessary

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,216 @@
...
+            dependencies = {'{}+{}'.format(*k): v for k, v in self._dependencies.items()}
+            self._state_committer.commit(dependencies)
+
+    def _load_state(self):
+        with self._lock:
+            dependencies = self._state_committer.load()
+            self._dependencies = {tuple(k.split('+')): v for k, v in dependencies.items()}
```
fabeschan: ditto

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,256 @@
...
+                        self._run_manager.uploading[bundle_uuid]['run_status'] = run_status
+
+                self._run_manager.upload_bundle_contents(bundle_uuid, run_state.bundle_path, update_status)
+            except Exception:
+                traceback.print_exc()
+
+        bundle_uuid = run_state.bundle['uuid']
+        if bundle_uuid not in self._run_manager.uploading:
+            self._run_manager.uploading[bundle_uuid] = {
+                'thread': threading.Thread(target=upload_results, args=[]),
```
fabeschan: maybe pull out the common "thread+return/err message" functionality and build an abstraction for it or something

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,256 @@
...
+                    'failure_message': failure_message,
+                }
+                self._run_manager.finalize_bundle(bundle_uuid, finalize_message)
+            except Exception:
+                traceback.print_exc()
+
+        bundle_uuid = run_state.bundle['uuid']
+        if bundle_uuid not in self._run_manager.finalizing:
+            self._run_manager.finalizing[bundle_uuid] = {
+                'thread': threading.Thread(target=finalize, args=[]),
```
fabeschan: ditto about the "thread+return/err message" thingy

worker/codalabworker/fsm.py
bkgoksel: I'm not sure if handling that is worth the design/dev time. These state files are only for the recovery of the running jobs in case the worker dies mid job. In a controlled update setting we can either manually update/deal with the schema change at the update that changes it.

worker/codalabworker/docker_client.py
```diff
@@ -302,7 +302,7 @@ def remove_image(self, repo_digest):
 
     ''' Download the specified docker image with tag/digest. If no tag is specified, downloads the latest '''
     @wrap_exception('Unable to download Docker image')
-    def download_image(self, docker_image, loop_callback):
+    def download_image(self, docker_image, progress_callback):
```
fabeschan, line 302: document purpose and expected behavior of `progress_callback`

worker/codalabworker/fsm.py
fabeschan: okay

codalab/model/bundle_model.py
```diff
@@ -731,13 +730,14 @@ def resume_bundle(self, bundle, user_id, worker_id, hostname, start_time):
...
                 connection.execute(cl_worker_run.insert().values(worker_run_row))
 
-            bundle_update = {
-                'state': State.RUNNING,
-                'metadata': {
-                    'last_updated': int(time.time()),
-                },
-            }
-            self.update_bundle(bundle, bundle_update, connection)
+            if row.state == State.RUNNING or row.state == State.WORKER_OFFLINE:
```
percyliang: What about the else case - is that an unexpected error?  Should probably comment / print a warning message or something...

codalab/rest/workers.py
```diff
@@ -23,19 +23,33 @@ def checkin(worker_id):
     Waits for a message for the worker for WAIT_TIME_SECS seconds. Returns the
     message or None if there isn't one.
     """
-    WAIT_TIME_SECS = 5.0
+    WAIT_TIME_SECS = 3.0
 
     torque_worker = ('torque' in local.config['workers'] and
```
percyliang: remove since not used

codalab/rest/workers.py
```diff
@@ -23,19 +23,33 @@ def checkin(worker_id):
...
 
     socket_id = local.worker_model.worker_checkin(
         request.user.user_id, worker_id, request.json['tag'],
         request.json['cpus'], request.json['gpus'], request.json['memory_bytes'],
         request.json['dependencies'])
+
+    for uuid, run in request.json['runs'].items():
+        bundle = local.model.get_bundle(uuid)
+        local.model.resume_bundle(bundle, request.user.user_id, worker_id,
+                                request.json['hostname'], run['start_time'])
```
percyliang: fix alignment

codalab/worker/bundle_manager.py
```diff
@@ -388,7 +395,6 @@ def _compute_request_cpus(self, bundle):
         """
         Compute the CPU limit used for scheduling the run.
         """
-        #TODO: Remove this once we want to deprecate old versions
```
percyliang: Probably should put a comment here that this is really only here for backward compatibility with old DBs.

codalab/worker/bundle_manager.py
```diff
@@ -397,7 +403,6 @@ def _compute_request_gpus(self, bundle):
         """
         Compute the GPU limit used for scheduling the run.
         """
-        #TODO: Remove this once we want to deprecate old versions
```
percyliang: ditto, and same for the instances below (no need to replicate the entire comment)

worker/codalabworker/fsm.py
```diff
@@ -0,0 +1,91 @@
...
+    def load(self, default=None):
+        if not os.path.exists(self._state_file):
+            return dict() if default is None else default
+        with open(self._state_file) as json_data:
+            return pyjson.load(json_data)
+
+    def commit(self, state):
+        """ Write out the state in JSON format to a temporary file and rename it into place """
+
+        dirname, basename = os.path.split(self._state_file)
```
percyliang: not used

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+            for digest, state in self._images.items():
+                if state.stage == DependencyStage.FAILED:
+                    if now - state.last_used > self._max_age_failed_seconds:
+                        del self._images[digest]
+
+            for digest, state in self._images.items():
+                self._images[digest] = self.transition(state)
+
+    def _cleanup(self):
+        while True:
```
percyliang: add docstring

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+                              size_str(self._max_images_bytes))
+                with self._lock:
+                    failed_images = {digest: image for digest, image in self._images.items() if image.stage == DependencyStage.FAILED}
+                    ready_images = {digest: image for digest, image in self._images.items() if image.stage == DependencyStage.READY}
+                    if failed_images:
+                        digest_to_remove = min(failed_images, key=lambda i: failed_images[i].last_used)
+                    elif ready_images:
+                        digest_to_remove = min(ready_images, key=lambda i: ready_images[i].last_used)
+                    else:
+                        # TODO: What do we do if there are only downloading images but together they are bigger than the quota
```
percyliang: print warning at least

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+                        break
+                    try:
+                        self._docker.remove_image(digest_to_remove)
```
percyliang: will crash if follow `else`

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+            return (digest in self._images)
+
+    def get(self, digest):
+        now = time.time()
+        with self._lock:
+            if not self.has(digest):
+                self._images[digest] = DockerImageState(stage=DependencyStage.DOWNLOADING, digest=digest, killed=False, last_used=now, message="")
+
+            # update last_used as long as it isn't in FAILED
+            if self._images[digest].stage != DependencyStage.FAILED:
```
percyliang, line 141: why need to special case this?

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, docker, state_committer, max_images_bytes, max_age_failed_seconds=60):
+        super(DockerImageManager, self).__init__()
+        self.add_transition(DependencyStage.DOWNLOADING, self._transition_from_DOWNLOADING)
+        self.add_transition(DependencyStage.READY, self._transition_from_READY)
+        self.add_transition(DependencyStage.FAILED, self._transition_from_FAILED)
+
+        self._state_committer = state_committer
+        self._docker = docker
+        self._images = {} # digest -> DockerImageState
+        self._downloading = {}
```
percyliang: comment

percyliang: This isn't serialized - is there a problem there?

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+logger = logging.getLogger(__name__)
+
+DockerImageState = namedtuple('DockerImageState', 'stage digest killed last_used message')
+
+class DockerImageManager(StateTransitioner, BaseDependencyManager):
+
+    def __init__(self, docker, state_committer, max_images_bytes, max_age_failed_seconds=60):
+        super(DockerImageManager, self).__init__()
+        self.add_transition(DependencyStage.DOWNLOADING, self._transition_from_DOWNLOADING)
+        self.add_transition(DependencyStage.READY, self._transition_from_READY)
```
percyliang: seems a bit silly to have to explicitly add these for terminal states...  Should have `add_terminal(DependencyStage.READY)`

worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -0,0 +1,183 @@
...
+
+logger = logging.getLogger(__name__)
+
+DockerImageState = namedtuple('DockerImageState', 'stage digest killed last_used message')
+
+class DockerImageManager(StateTransitioner, BaseDependencyManager):
+
+    def __init__(self, docker, state_committer, max_images_bytes, max_age_failed_seconds=60):
+        super(DockerImageManager, self).__init__()
+        self.add_transition(DependencyStage.DOWNLOADING, self._transition_from_DOWNLOADING)
```
percyliang: I find it a bit confusing that the `StateTransitioner` is a parent class, since this pertains to a single image state, whereas this class is about the manager that deals with all of them.  Maybe we discussed keeping it in one class to avoid too much indirection?  Anyway, a comment would be nice.

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+
+class DownloadAbortedException(Exception):
+    """
+    Exception raised by the download if a download is killed before it is complete
+    """
+    def __init__(self, message):
+        super(DownloadAbortedException, self).__init__(message)
+
+class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager):
+    DEPENDENCIES_DIR_NAME = 'dependencies'
```
percyliang: docstring

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+    def __init__(self, state_committer, bundle_service,
+            worker_dir, max_cache_size_bytes, max_serialized_length):
+
+        super(LocalFileSystemDependencyManager, self).__init__()
+        self.add_transition(DependencyStage.DOWNLOADING, self._transition_from_DOWNLOADING)
+        self.add_transition(DependencyStage.READY, self._transition_from_READY)
```
percyliang: same comment as for `DockerImageManager`

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+        self._max_serialized_length = max_serialized_length or float('inf')
+        self.dependencies_dir = os.path.join(worker_dir, LocalFileSystemDependencyManager.DEPENDENCIES_DIR_NAME)
+        if not os.path.exists(self.dependencies_dir):
+            logger.info('{} doesn\'t exist, creating.'.format(self.dependencies_dir))
+            os.makedirs(self.dependencies_dir, 0770)
+
+        self._lock = threading.RLock()
+
+        self._paths = set()
+        self._dependencies = dict()
```
percyliang: `{}`, comment on what the key/value are

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+        self._downloading = dict()
```
percyliang: `{}`, comment on what the key/value are

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+    def _process_dependencies(self):
+        with self._lock:
+            for entry, state in self._dependencies.items():
+                self._dependencies[entry] = self.transition(state)
+
+    def _cleanup(self):
+        while True:
+            with self._lock:
+                bytes_used = sum(dep.size_bytes for dep in self._dependencies.values())
+                serialized_dependencies = {'{}+{}'.format(*k): v for k, v in self._dependencies.items()}
```
percyliang: why don't we serialize dependencies directly instead of having `format` as a string?

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+                                  size_str(serialized_length),
+                                  size_str(self._max_serialized_length))
+                    failed_deps = {dep: state for dep, state in self._dependencies.items() if dep.stage == DependencyStage.FAILED}
+                    ready_deps = {dep: state for dep, state in self._dependencies.items() if dep.stage == DependencyStage.READY and not state.dependents}
+                    if failed_deps:
+                        dep_to_remove = min(failed_deps, key=lambda i: failed_deps[i].last_used)
+                    elif ready_deps:
+                        dep_to_remove = min(ready_deps, key=lambda i: ready_deps[i].last_used)
+                    else:
+                        # TODO: What do we do if there are only downloading deps but together they are bigger than the quota
```
percyliang: print error

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+                        break
+                    try:
+                        self._paths.remove(self._dependencies[dep_to_remove].path)
```
percyliang: `dep_to_remove` might be None

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+            if self.has(dependency):
+                dep_state = self._dependencies[dependency]
+                if uuid in dep_state.dependents:
+                    dep_state.dependents.remove(uuid)
+                if not dep_state.dependents:
+                    dep_state = dep_state._replace(killed=True)
+                    self._dependencies[dependency] = dep_state
+
+    def _assign_path(self, dependency):
+        parent_uuid, parent_path = dependency
```
percyliang: docstring

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+        # You could have a conflict between, for example a/b_c and
+        # a_b/c. We have to avoid those.
+        with self._lock:
+            while path in self._paths:
+                path = path + '_'
+            self._paths.add(path)
+        return path
+
+    def _store_dependency(self, dependency_path, fileobj, target_type):
+        try:
```
percyliang: docstring

worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -0,0 +1,277 @@
...
+            return dependency_state._replace(stage=DependencyStage.FAILED, message=failure_message)
+
+    def _transition_from_READY(self, dependency_state):
+        return dependency_state
+
+    def _transition_from_FAILED(self, dependency_state):
+        return dependency_state
+
+
+class SharedFileSystemDependencyManager(BaseDependencyManager):
```
percyliang: Huh?  This class doesn't seem to be used.  Why is everything not implemented?

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -0,0 +1,248 @@
...
+        self._docker_network_prefix = docker_network_prefix
+
+        # These members are public as the run state manager needs access to them
+        self.docker = docker
+        self.image_manager = image_manager
+        self.dependency_manager = dependency_manager
+        self.cpuset = cpuset
+        self.gpuset = gpuset
+
+        self.runs = {}
```
percyliang: document for these what the key/value types are....

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -0,0 +1,248 @@
...
+                stage=LocalRunStage.PREPARING, run_status='', bundle=bundle,
+                bundle_path=os.path.realpath(bundle_path), resources=resources,
+                start_time=now, container_id=None, docker_image=None, is_killed=False,
+                has_contents=False, cpuset=None, gpuset=None, info={},
+        )
+        with self.lock:
+            self.runs[bundle_uuid] = run_state
+
+    def get_kill_message(self):
+        # TODO: There can be other kill messages
```
percyliang: ?

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -0,0 +1,248 @@
...
+            for run_state in self.runs.values():
+                if run_state.stage == LocalRunStage.RUNNING:
+                    cpuset -= run_state.cpuset
+                    gpuset -= run_state.gpuset
+
+        if len(cpuset) < request_cpus or len(gpuset) < request_gpus:
+            raise Exception("Not enough cpus or gpus to assign!")
+
+        def propose_set(resource_set, request_count):
+            return set(list(resource_set)[:request_count])
```
percyliang, line 201: maybe we should sort to make this deterministic

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -0,0 +1,248 @@
...
+        container_ip = self.docker.get_container_ip(
+                self.docker_network_external_name, run_state.container_id)
+        if not container_ip:
+            container_ip = self.docker.get_container_ip(
+                    self.docker_network_internal_name, run_state.container_id)
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect((container_ip, port))
+        s.sendall(message)
+
+        total_data=[]
```
percyliang: add space

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -0,0 +1,248 @@
...
+        while True:
+            data = s.recv(1024)
```
percyliang: isn't 1024 kind of small for a buffer size?

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+    FINALIZING = 'LOCAL_RUN.FINALIZING'
+
+    """
+    Finished means the worker is done with this run
+    """
+    FINISHED = 'LOCAL_RUN.FINISHED'
+
+LocalRunState = namedtuple('RunState',
+    ['stage', 'run_status', 'bundle', 'bundle_path', 'resources', 'start_time',
+'container_id', 'docker_image', 'is_killed', 'has_contents', 'cpuset', 'gpuset', 'info'])
```
percyliang: fix indent

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    Manages the state machine of the runs running on the local machine
+    """
+
+    def __init__(self, run_manager):
+        super(LocalRunStateMachine, self).__init__()
+        self._run_manager = run_manager
+        self.add_transition(LocalRunStage.PREPARING, self._transition_from_PREPARING)
+        self.add_transition(LocalRunStage.RUNNING, self._transition_from_RUNNING)
+        self.add_transition(LocalRunStage.CLEANING, self._transition_from_CLEANING)
```
percyliang: More accurate: CLEANING_UP

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+        image_state = self._run_manager.image_manager.get(docker_image)
+
+        # then inspect the state of every dependency/image to see whether all of them are ready
+        for dep in run_state.bundle['dependencies']:
+            dependency = (dep['parent_uuid'], dep['parent_path'])
+            dependency_state = self._run_manager.dependency_manager.get(run_state.bundle['uuid'], dependency)
+            if dependency_state.stage == DependencyStage.DOWNLOADING:
+                status_message = 'Downloading dependency %s: %s done (archived size)' % (
+                            dep['child_path'], size_str(dependency_state.size_bytes))
+                return run_state._replace(run_status=status_message)
```
percyliang: This is not quite accurate - should concatenate the status messages from all the dependencies and docker image...

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+        # 4) Start container
+        container_id = self._run_manager.docker.start_container(
+            run_state.bundle_path, bundle_uuid, run_state.bundle['command'],
+            run_state.resources['docker_image'], docker_network, dependencies,
+            cpuset, gpuset, run_state.resources['request_memory']
+        )
+
+        digest = self._run_manager.docker.get_image_repo_digest(run_state.resources['docker_image'])
+
+        return run_state._replace(stage=LocalRunStage.RUNNING, run_status='Running',
```
percyliang: make run_status a bit longer so it looks more like a status message than a stage.

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+        new_info = check_and_report_finished(run_state)
+        run_state.info.update(new_info)
+        run_state = run_state._replace(info=run_state.info)
+
+        if run_state.is_killed and run_state.container_id is not None:
+            try:
+                self._run_manager.docker.kill_container(run_state.container_id)
+            except DockerException:
+                traceback.print_exc()
+            return run_state._replace(stage=LocalRunStage.CLEANING, container_id=None)
```
percyliang: `is_killed` really means kill request is there, not that it is killed? maybe variable name should reflect this better

Should we transition into CLEANING?  What happens if the kill request failed?

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+        if run_state.info['finished']:
+            logger.debug('Finished run with UUID %s, exitcode %s, failure_message %s',
+                 bundle_uuid, run_state.info['exitcode'], run_state.info['failure_message'])
+            return run_state._replace(stage=LocalRunStage.CLEANING, run_status='Uploading results')
```
percyliang: this run status message is inaccurate

percyliang: It's a bit weird since the run status really aligns better with transition function for the the target stage (`_transition_from_CLEANING`) rather than this function.

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+            child_path = os.path.join(run_state.bundle_path, dep['child_path'])
+            try:
+                remove_path(child_path)
+            except Exception:
+                traceback.print_exc()
+
+        if run_state.has_contents:
+            return run_state._replace(stage=LocalRunStage.UPLOADING_RESULTS, run_status='Uploading results')
+        else:
+            return run_state._replace(stage=LocalRunStage.FINALIZING)
```
percyliang: don't need to set `run_status`?

worker/codalabworker/local_run/local_run_state.py
```diff
@@ -0,0 +1,277 @@
...
+
+                self._run_manager.upload_bundle_contents(bundle_uuid, run_state.bundle_path, progress_callback)
+            except Exception:
+                traceback.print_exc()
+
+        bundle_uuid = run_state.bundle['uuid']
+        if bundle_uuid not in self._run_manager.uploading:
+            self._run_manager.uploading[bundle_uuid] = {
+                'thread': threading.Thread(target=upload_results, args=[]),
+                'run_status': ''
```
percyliang: why empty?

worker/codalabworker/local_run/local_run_state.py
percyliang: +1

percyliang (changes requested):
Overall looks good and fairly understandable!  Comments are mostly lower-level.

worker/codalabworker/local_run/local_run_state.py
bkgoksel: Ideally yes, but that would make a very long status message. If we had a way to show this status better on the UI instead of displaying a single string, I would prefer to do it this way.

worker/codalabworker/local_run/local_run_state.py
bkgoksel: `is_killed` in this context is taken to mean the run was killed by the user server-side. So in a sense the run is already killed and now the run manager has to deal with it (this is the semantics we want to enforce) 

I don't know under what condition the kill_container call might fail and so I don't know how to handle that case. For all it's worth all this exception capturing was already there in the first place and I was reluctant to remove it, maybe we should? Transitioning to cleaning up is still the best option I think.

worker/codalabworker/local_run/local_run_state.py
bkgoksel: These transition functions prepare the run state for the next stage, so in this case the return RunState object will be FINISHED.

worker/codalabworker/local_run/docker_image_manager.py
bkgoksel: this dict is for keeping track of the download threads so no need to serialize.

worker/codalabworker/local_run/local_run_state.py
percyliang: Could we append a short summary (e.g., "N others")

PR #910, opened 2018-05-11
Dennis error message

w0nche0l:
Makes the error message clearer that the reason we are unable to to execute this command is because it is being run from the web CLI.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2952,7 +2952,7 @@ class BundleCLI(object):
 
     def _fail_if_headless(self, args):
         if self.headless:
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -2952,7 +2952,7 @@ def do_bs_health_check(self, args):
 
     def _fail_if_headless(self, args):
         if self.headless:
-            raise UsageError('Cannot execute CLI command: %s' % args.command)
+            raise UsageError('Cannot execute CLI command "%s" because you are using a headless CLI.' % args.command)
```
percyliang: This will make no sense from the web browser.  Should say: You are only allowed to execute command "%s" from the CLI.

percyliang (approved):

PR #911, opened 2018-05-12
Web CLI cat support

w0nche0l:
Before: cat command would not work in the web CLI.
After: cat command displays what is printed in the side panel under contents.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1788,7 +1788,6 @@ class BundleCLI(object):
         ),
     )
     def do_cat_command(self, args):
-        self._fail_if_headless(args)  # Files might be too big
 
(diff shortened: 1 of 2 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1808,10 +1807,20 @@ def print_target_info(self, client, bundle_uuid, subpath, head=None, tail=None):
...
+                kwargs['head'] = 50
+                kwargs['tail'] = 50
+                kwargs['truncation_text'] = '\n... truncated ...\n\n'
+
             contents = client.fetch_contents_blob(bundle_uuid, subpath, **kwargs)
             with closing(contents):
                 shutil.copyfileobj(contents, self.stdout)
 
+            if self.headless:
+                print >>self.stdout, '--Web CLI detected, truncated output to first 50 and last 50 lines.--'
```
percyliang, line 1822: Don't need this since "... truncated ..." will be shown, right?

Issue #427, opened 2016-05-07
CLI command to list users


kashizui:
`cl uls`

List users using the users API, and in particular perhaps an `--online` or `--active` flag to search to users who have recently been active. (This maybe can be restricted to admin users only.)
The latter would be equivalent of one or both of the following (written by Konstantin):

```
SELECT DISTINCT user_name FROM oauth2_token JOIN user ON oauth2_token.user_id = user.user_id WHERE client_id = "codalab_cli_client" AND expires > utc_timestamp();
SELECT user_name, last_login FROM user WHERE last_login > utc_timestamp() - 30 * 24 * 60 * 60 ORDER BY last_login DESC;
```

kashizui:
Simply need to call the Users API

`JsonApiClient.fetch('users')`

Can add more query parameters to `rest/users.py` if necessary to implement more fancy features.

percyliang:
Want to also be able to search by username, email, etc.

Levilian:
Lists users on CodaLab (returns 10 results by default).
    uls <keyword> ... <keyword>         : Username or id contains each <keyword>.
    uls user_name=<value>               : Name is <value>, where `user_name` can be any metadata field (e.g., email).

    uls .limit=<limit>                  : Limit the number of results to the top <limit> (e.g., 50).
    uls .offset=<offset>                : Return results starting at <offset>.

    uls .joined_before=<datetime>       : Returns users joined before (inclusive) given ISO 8601 timestamp (e.g., .before=2042-03-14).
    uls .joined_after=<datetime>        : Returns users joined after (inclusive) given ISO 8601 timestamp (e.g., .after=2120-10-15T00:00:00-08).
    uls .active_before=<datetime>       : (Root user only) Returns users last logged in before (inclusive) given ISO 8601
    timestamp (e.g., .before=2042-03-14).
    uls .active_after=<datetime>        : (Root user only) Returns users last logged in after (inclusive) given ISO 8601 timestamp (e.g., .after=2120-10-15T00:00:00-08).

    uls .disk_used_less_than=<percentage> or <float>      : (Root user only) Returns users whose disk usage less than (inclusive) given value (e.g., .disk_used_less_than=70% or 0.3).
    uls .disk_used_more_than=<percentage> or <float>      : (Root user only) Returns users whose disk usage more than (inclusive) given value (e.g., .disk_used_more_than=70% or 0.3).
[...]

PR #915, opened 2018-05-27
Interpret overhaul, adding MarkupSchema

w0nche0l:
Converted currently existing code to use MarkupBlockSchema instead of a raw dictionary. 
Created MarkupBlockSchema. 
Created a new block_schemas.py to avoid circular dependency errors; added BlockModes enum class to keep track of mode strings in centralized location. 
See https://github.com/codalab/codalab-worksheets/pull/427 for front-end changes.

WIP: Other blocks. 

Final goal of this PR: to use BlockSchemas for all interpreted items.


codalab/lib/worksheet_util.py
```diff
@@ -617,11 +618,11 @@ def raise_genpath_usage_error():
...
                 if is_missing(bundle_info):
-                    interpreted_items.append({
-                        'mode': TYPE_MARKUP,
-                        'interpreted': 'ERROR: cannot access bundle',
-                        'properties': {},
-                    })
+                    interpreted_items.append(
+                        MarkupBlockSchema().load({
+                            'id': len(interpreted_items),
+                            "text": "ERROR: cannot access bundle",
```
percyliang: be consistent: use single quotes

codalab/lib/worksheet_util.py
```diff
@@ -617,11 +618,11 @@ def raise_genpath_usage_error():
 
             for item_index, bundle_info in bundle_infos:
                 if is_missing(bundle_info):
-                    interpreted_items.append({
-                        'mode': TYPE_MARKUP,
-                        'interpreted': 'ERROR: cannot access bundle',
-                        'properties': {},
-                    })
+                    interpreted_items.append(
```
percyliang: Should call these `blocks` instead of `interpreted_items` to avoid confusion...

codalab/rest/block_schemas.py
```diff
@@ -0,0 +1,96 @@
+"""
+Marshmallow schemas for serializing resource dicts into JSON API documents, and vice-versa.
```
percyliang, line 2: Should say something about which schemas (worksheet blocks).

Naming: at the top-level, should be specific and use `worksheet_block` (which goes for the file name).

codalab/rest/block_schemas.py
```diff
@@ -0,0 +1,96 @@
...
+    ValidationError,
+    validate,
+    validates_schema,
+)
+from marshmallow_jsonapi import Schema, fields
+
+
+# Enum that represents different modes for a block.
+class BlockModes:
+    MARKUP_MODE = "markup_block"
```
percyliang, line 18: Can we just make the Python variable match the actual string?

codalab/rest/block_schemas.py
```diff
@@ -0,0 +1,96 @@
...
+
+    values = (MARKUP_MODE, RECORD_MODE, TABLE_MODE, BUNDLE_CONTENTS_MODE,
+              BUNDLE_HTML_MODE, BUNDLE_IMAGE_MODE, GRAPH_MODE, WSEARCH_MODE,
+              SEARCH_MODE)
+
+
+STATUS_STRINGS = ("unknown", "ready", "not_found", "no_permission")
+
+
+class FetchStatusSchema(fields.Field):
```
percyliang, line 33: Document

codalab/rest/block_schemas.py
```diff
@@ -0,0 +1,96 @@
...
+    GRAPH_MODE = "graph_block"
+    WSEARCH_MODE = "wsearch_block"
+    SEARCH_MODE = "search_block"
+
+    values = (MARKUP_MODE, RECORD_MODE, TABLE_MODE, BUNDLE_CONTENTS_MODE,
+              BUNDLE_HTML_MODE, BUNDLE_IMAGE_MODE, GRAPH_MODE, WSEARCH_MODE,
+              SEARCH_MODE)
+
+
+STATUS_STRINGS = ("unknown", "ready", "not_found", "no_permission")
```
percyliang: Need to think more carefully about whether this is the right set of status strings...

For example, if I don't have permission, should I get a 'not_found' or 'no_permission' (which reveals the existence of certain resources even if I don't have access)...

codalab/rest/interpret.py
```diff
@@ -316,14 +317,15 @@ def error_data(mode, message):
         if item is None:
             continue
         mode = item['mode']
-        data = item['interpreted']
-        properties = item['properties']
+        # TODO: remove these fields entirely, using fields unique to each mode instead.
```
percyliang: What's the migration plan?  Why keep these here if we plan to use the block-schema-specific fields...

codalab/lib/worksheet_util.py
w0nche0l: Oops @ being consistent; I thought we supposed to be using double quotes throughout the codebase?

codalab/rest/interpret.py
w0nche0l: I'm going to update this PR to remove these entirely, they're in there currently to keep every other block type working while I test.

codalab/lib/worksheet_util.py
w0nche0l: sounds good!

codalab/rest/block_schemas.py
w0nche0l: I think it is okay to have a distinction between not_found and no_permission, since we already do that for resources in our current front-end.

w0nche0l: What is unknown? Just like a generic catch-all?

w0nche0l: Should there be a "fetching" status that represents something that is currently trying to be fetched?

percyliang: I think `unknown` is supposed to be that.  I don't think it should be called `fetching` because it should apply when no fetching request has been made.  Maybe `pending`?

codalab/rest/interpret.py
```diff
@@ -296,7 +296,6 @@ def head_target(target, max_num_lines, replace_non_unicode=False):
 
 
 # Default number of lines to pull for each display mode.
-DEFAULT_CONTENTS_MAX_LINES = 10
```
w0nche0l, line 301: Moved this constant to worksheet_util.py

codalab/rest/interpret.py
```diff
@@ -323,40 +322,55 @@ def error_data(mode, message):
...
+                pass
             elif mode == 'record' or mode == 'table':
                 # header_name_posts is a list of (name, post-processing) pairs.
                 header, contents = data
                 # Request information
                 contents = interpret_genpath_table_contents(contents)
                 data = (header, contents)
-            elif mode == 'contents':
+
+                # TODO: remove this and use new schemas
```
w0nche0l: All of these are marking things that I haven't migrated.

codalab/rest/worksheet_block_schemas.py
```diff
@@ -0,0 +1,105 @@
...
+    mode = fields.Constant(BlockModes.image_block)
+
+    image_data = fields.String()
+    height = fields.Integer()
+    width = fields.Integer()
+
+
+class BundleHTMLBlockSchema(BundleBlockSchema):
+    mode = fields.Constant(BlockModes.html_block)
+    max_lines = fields.Constant(sys.maxint)
```
w0nche0l: Currently, the html display function is broken on production. This is a hack to make sure the entire HTML file gets displayed, but I'm not a huge fan. However, file_util:summarize_file explicitly asserts to make sure that there is some limit on the number of lines. Help!

codalab/rest/worksheet_block_schemas.py
```diff
@@ -0,0 +1,105 @@
...
+    ValidationError,
+    validate,
+    validates_schema,
+)
+from marshmallow_jsonapi import Schema, fields
+import sys
+
+# Enum that represents different modes for a block.
+class BlockModes:
+    markup_block = 'markup_block'
```
w0nche0l: Renamed to match strings.

codalab/rest/worksheet_block_schemas.py
```diff
@@ -0,0 +1,105 @@
...
+    graph_block = 'graph_block'
+    wsearch_block = 'wsearch_block'
+    search_block = 'search_block'
+
+    values = (markup_block, record_block, table_block, contents_block,
+              html_block, image_block, graph_block, wsearch_block,
+              search_block)
+
+
+STATUS_STRINGS = ('unknown', 'pending', 'ready', 'not_found', 'no_permission')
```
w0nche0l: Added pending status (not used yet), to be used when the front-end takes a block at unknown status and sends it off to be refined.

codalab/lib/worksheet_util.py
```diff
@@ -615,28 +624,52 @@ def raise_genpath_usage_error():
...
+                            'error_message': '',
+                        },
+                        'width': properties.get('width', None),
+                        'height': properties.get('height', None),
+                    }).data)
+                elif mode == 'html':
+                    interpreted_items.append(BundleHTMLBlockSchema().load({
+                        'target_genpath': target_genpath,
+                        'bundle_info': copy.deepcopy(bundle_info),
+                        'status': {
```
w0nche0l: Maybe I should define a constant for initialization of statuses?

percyliang: Yes, I think it's worth defining a bunch of helpers to reduce code duplication as long as it doesn't introduce layers of indirection that make it hard to understand.

codalab/rest/worksheet_block_schemas.py
```diff
@@ -0,0 +1,105 @@
...
+class BlockModes:
+    markup_block = 'markup_block'
+    record_block = 'record_block'
+    table_block = 'table_block'
+    contents_block = 'contents_block'
+    html_block = 'html_block'
+    image_block = 'image_block'
+    graph_block = 'graph_block'
+    wsearch_block = 'wsearch_block'
+    search_block = 'search_block'
```
percyliang: I think we decided that `search` was not going to be a separate type of block, but a mechanism for retrieving a list of bundles; specifying a list of bundles verbatim is another mechanism.

codalab/rest/worksheet_block_schemas.py
percyliang: What about deleting the html display functionality?  Does it actually get used?  For rich enough HTML, one would want to display it in another window anyway because we can't render arbitrary HTML with Javascript anyway in an embedded way...  Haven't thought this through but just bringing it up.

codalab/rest/worksheet_block_schemas.py
```diff
@@ -0,0 +1,175 @@
...
+
+class FetchStatusSchema(PlainSchema):
+    """
+    Schema that represents the status of fetching a resource.
+    """
+    code = fields.String(validate=validate.OneOf(set(STATUS_STRINGS)))
+    error_message = fields.String()
+
+    @staticmethod
+    def get_initial_status():
```
percyliang: rename to `get_unknown_status`

codalab/lib/bundle_cli.py
```diff
@@ -2339,37 +2339,31 @@ def do_print_command(self, args):
...
         for item in interpreted['items']:
             mode = item['mode']
-            data = item['interpreted']
-            properties = item['properties']
             print >>self.stdout, ''  # Separate interpreted items
-            if mode == 'markup' or mode == 'contents':
-                if mode == 'contents':
-                    maxlines = properties.get('maxlines')
+            if mode == 'markup_block' or mode == 'contents_block':
+                bundle_info = item['bundles_spec']['bundle_infos'][0]
```
percyliang: If markup, this shouldn't exist, right?

percyliang: Probably these two modes should just be separate branches of the if.

codalab/lib/worksheet_util.py
```diff
@@ -36,25 +36,43 @@
...
+    SubworksheetsBlock,
+    BundleUUIDSpecSchema,
+)
 
 
 # Note: this is part of the client's session, not server side.
 CURRENT_WORKSHEET = '.'
 
-# Types of worksheet items
+# Types of raw worksheet items
```
percyliang: say '(raw)', since by default worksheet items are raw now...

codalab/lib/worksheet_util.py
percyliang: Still need to do this, right?

codalab/lib/worksheet_util.py
```diff
@@ -679,45 +712,59 @@ def raise_genpath_usage_error():
...
-                #'bundle_info': [copy.deepcopy(bundle_info) for item_index, bundle_info in bundle_infos]
-            })
+            try:
+                max_lines = int(properties.get('maxlines', DEFAULT_CONTENTS_MAX_LINES))
+            except ValueError:
+                raise UsageError("maxlines must be integer")
+
+            interpreted_items.append(GraphBlockSchema().load({
+                'trajectories': trajectories,
+                'bundles_spec': BundleUUIDSpecSchema().load(BundleUUIDSpecSchema.create_json([bundle_infos[0][1]])).data,
```
percyliang: Keep the comment that we're taking only the first bundle

codalab/rest/interpret.py
```diff
@@ -183,6 +186,12 @@ def fetch_interpreted_worksheet(uuid):
     # Fetch items.
     worksheet_info['raw'] = get_worksheet_lines(worksheet_info)
 
+    # Replace searches with raw items.
```
percyliang, line 196: Do you think you'll be able to unify this with `fetch_worksheet` from `worksheets.py`?

codalab/rest/interpret.py
```diff
@@ -584,3 +544,39 @@ def resolve_items_into_infos(items):
         value_obj = formatting.string_to_tokens(i['value']) if i['type'] == TYPE_DIRECTIVE else i['value']
         new_items.append((bundle_info, subworksheet_info, value_obj, i['type']))
     return new_items
+
+
+def expand_raw_item(raw_item):
+    """
+    Some raw items must be expanded into more raw items.
```
percyliang: Be more specific about how it's just search items...

codalab/rest/interpret.py
```diff
@@ -183,6 +186,12 @@ def fetch_interpreted_worksheet(uuid):
     # Fetch items.
     worksheet_info['raw'] = get_worksheet_lines(worksheet_info)
 
+    # Replace searches with raw items.
+    # This needs to be done before get_worksheet_lines because this replaces
+    # user-written raw items.
+    worksheet_info['items'] = list(chain.from_iterable(
+        [expand_raw_item(raw_item) for raw_item in worksheet_info['items']]))
```
percyliang: This really worries me...`items` is now a list of a list of items?  It'd be nice if `items` really just corresponded to the raw items.  I thought the search directives would only influence the blocks.

codalab/rest/interpret.py
```diff
@@ -306,134 +314,86 @@ def resolve_interpreted_items(interpreted_items):
...
-                    data = None
-            elif mode == 'html':
-                target_info = rest_util.get_target_info(data, 0)
-                if target_info is not None and target_info['type'] == 'file':
-                    data = head_target(data, None)
+                    item['status']['code'] = 'ready'
+                    if mode == BlockModes.contents_block:
+                        item['lines'] = head_target((item['bundles_spec']['bundle_infos'][0]['uuid'], item['target_genpath']), item['max_lines'], replace_non_unicode=True)
+                    elif mode == BlockModes.html_block:
+                        item['html_lines'] = head_target((item['bundles_spec']['bundle_infos'][0]['uuid'], item['target_genpath']), item['max_lines'])
```
percyliang: html_lines doesn't make much sense since truncated HTML isn't valid.  Probably better to just have `html` and slurp up the entire file (like how we treat images).

I thought we were going to deprecate HTML anyway?

codalab/rest/interpret.py
```diff
@@ -306,134 +314,86 @@ def resolve_interpreted_items(interpreted_items):
...
-                    data = base64.b64encode(cat_target(data))
-                else:
-                    data = None
-            elif mode == 'graph':
-                try:
-                    max_lines = int(properties.get('maxlines', DEFAULT_CONTENTS_MAX_LINES))
-                except ValueError:
-                    raise UsageError("maxlines must be integer")
-
+                    item['status']['code'] = 'not_found'
```
percyliang: probably should have constants defined for the various codes...

codalab/rest/interpret.py
```diff
@@ -306,134 +314,86 @@ def resolve_interpreted_items(interpreted_items):
     appropriate information, replacing the 'interpreted' field in each item.
     The result can be serialized via JSON.
     """
-    def error_data(mode, message):
-        if mode == 'record' or mode == 'table':
-            return (('ERROR',), [{'ERROR': message}])
-        else:
-            return [message]
+    def error_data(item_index, message):
```
percyliang: Why make this mutable?  If you really need it to mutate, then rename to `set_error_data`.

percyliang (changes requested):
Main comment is regarding how expanding raw items is handled.

codalab/rest/interpret.py
w0nche0l: No, the chain.from_iterable flattens the list of list of items from the list comprehension into a list of items. Before I changed anything worksheet_info['items'] was populated with raw items, and then was replaced with the interpreted items. I can make it so that this is completely separated into worksheet_info['raw_items'], worksheet_info['replaced_raw_items'], and worksheet_info['interpreted_blocks'], if that makes more sense. 

I think this actually fits really nicely into how our parser/interpreter works currently, because a search is an alias for a list of bundles (or worksheets, in the case of wsearch), and this means that we don't have to include extra logic for every single block on how to handle searches.

codalab/rest/interpret.py
w0nche0l: Yeah we were going to deprecate HTML, should I do it in this change? was just replicating functionality for this change and then take it out in a separate one.

codalab/rest/interpret.py
w0nche0l: I got the impression that this entire function is mutating the contents of interpreted_items anyway, I can make a non-mutating version that creates a copy first and fills it in.

codalab/rest/interpret.py
w0nche0l: I'm not sure I understand this question, fetch_worksheet only cares about the raw items in the worksheet, whereas this is replacing any raw items that contain search directives with the result of the search? For example, if my worksheet just contained 
"% search .limit=2"
I would not want the worksheet to seem like it actually contained
[bundle1]{0xw50u5oijwed} 
[bundle2]{0xoih0w3523n5}

codalab/rest/interpret.py
w0nche0l: yeah, sounds good.

codalab/lib/worksheet_util.py
w0nche0l: yeah, wanted to simplify the diff view by doing this last

codalab/lib/bundle_cli.py
w0nche0l: yeah, will definitely change this.

codalab/rest/interpret.py
percyliang: I just meant that there would be one REST endpoint that would return both the (raw) items as well as the blocks.  Otherwise there's some code duplication in two places, and there's inconsistent handling of owner, for example.

codalab/rest/interpret.py
percyliang: Well, one question is whether we want the conversion from items to blocks to be oblivious of search directives.  Certainly doing this replacement simplifies the blocks logic, but it does introduce 3 objects to reason about (raw items, replaced items, and blocks) rather than 2.  Also, it precludes us from later trying to lazily resolve the search directives during interpretation time rather than all at once.

w0nche0l: I don't think it gives us 3 objects to reason about, since replaced items are identical to raw items from the perspective of the raw -> block interpreter. 

Also, as we discussed before, search directives are important to the codalab architecture, but also used very sparsely, mostly in the homepage and to perform debugging. Therefore, I think it would be much more efficient to add logic for lazily resolving other directives, such as image and table, since that functionality would apply to all worksheets. Also, those gains would be realized for searches, since searches resolve into table blocks anyway.

codalab/rest/interpret.py
w0nche0l: Ah okay I see what you mean, I'll take a look.

codalab/rest/interpret.py
percyliang: As long as only raw items and blocks are exposed to the frontend, then I'm happy.  If the replaced items can be encapsulated in the raw items => blocks conversion (even though we do break it up further, which is great)...

I don't understand your point about searches - can you please rephrase?

Are we committing to resolving all the search directives whenever we fetch a worksheet?  This seems to make certain things off limits - like being able to lazily fetch a very long list of bundles or worksheets dynamically...  Maybe it's fine to do that, but it seems worth thinking through the consequences.

w0nche0l: Yes, currently the frontend is completely unaware of the raw items. It renders purely based on the blocks. 

Sorry for being unclear; I think what I was trying to say is that the plan is already to lazily resolve the contents of display directives that require more than just the bundle metadata, such as images, contents, and genpaths in tables. These all require disk reads, and are what contribute most to the slow load times of large and complex worksheets. For example, if a table of 10 items wants the first line of the file results.txt, that will hit the file system 10 separate times. I was under the impression that this refactor is laying the grounds for lazily loading these disk-intensive tasks. 

Meanwhile, search is definitely not a common tool for most worksheets, it is mostly used as a diagnostic tool for us to build dashboards such as the current homepage. Additionally, it will gain the same performance gains that I described above. 

We are committing to resolving all of the search directives whenever we fetch a worksheet, but that was already being done previously, so it wouldn't make anything slower than it already is now. I'm having a hard time thinking of a use case for lazily fetching a very long (>1000) list of bundles dynamically.

codalab/rest/interpret.py
w0nche0l: There's not a very clean way to unify this, I think this should be separated into its own diff after this is merged.

codalab/lib/worksheet_util.py
```diff
@@ -679,45 +709,61 @@ def raise_genpath_usage_error():
...
-            })
+            try:
+                max_lines = int(properties.get('maxlines', DEFAULT_CONTENTS_MAX_LINES))
+            except ValueError:
+                raise UsageError("maxlines must be integer")
+
+            blocks.append(GraphBlockSchema().load({
+                'trajectories': trajectories,
+                'bundles_spec': BundleUUIDSpecSchema().load(BundleUUIDSpecSchema.create_json([bundle_infos[0][1]])).data, # Only show the first one for now
+                # 'bundles_spec': BundleUUIDSpecSchema().load(BundleUUIDSpecSchema.create_json(
```
bkgoksel, line 748: Do we know what this is or is it just legacy commented out code?

bkgoksel (approved):
I'm not super familiar with the details of how everything works but the code looks good to me.

PR #916, opened 2018-05-27
Adding referer url to the help emails

w0nche0l:
Before: help emails only contained the identity of the requester, as well as the message they sent.  
After: help emails also include a direct link to the url that the help message was sent from. This will help us resolve help emails without as much back and forth.


percyliang (approved):
Great!

PR #920, opened 2018-06-25
Better exception handling in get_target_info

bkgoksel:
Previously the get_target_info function would handle exceptions internally and return an uninformative None in case of exception. Now it raises these exceptions so its callers can propagate better error messages when get_target_info is called with invalid bundle uuids or paths.

Added to help fix the SQuAD 2.0 evaluation bug where the existing system means we get useless error messages.

Also fixes linter errors in the touched files.

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -568,7 +568,9 @@ class JsonApiClient(RestClient):
                        (bundle_id, urllib.quote(target_path))
         response = self._make_request('GET', request_path,
                                       query_params={'depth': depth})
-
+        # TODO Kerem: Remove this after debug done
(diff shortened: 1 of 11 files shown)


percyliang (approved):
Looks fine to me, but I think we need to make frontend changes to preserve behavior.  In general, need to test that this preserves the functionality.

bkgoksel:
Can't figure out why the tests are failing, abandoning this for now as it has become a time sink

bkgoksel:
Closing for now. Maybe will get back to it after the new worker work is done.

bkgoksel:
Issue also surfaced when working on new worker, resurfacing

PR #914, opened 2018-05-14
Added ability to search for worksheets shared with you through groups

w0nche0l:
before: feature didn't exist
after: can search within a single group that you are a part of, as well as all groups that you are in.

PR #922, opened 2018-07-12
Add group=<group_spec> and .shared keywords to search command (for bundle search)

bkgoksel:
Basically the same changes as in #914 but for the bundles search command


percyliang (approved):
Cool!

PR #924, opened 2018-07-28
Fixing bug 

w0nche0l:
Fixes bug where the server does not know how to deal with non-list return values from search.
`search_bundle_uuids` will return a single (non-list) number when looking for either a count or a sum; this returns that number as a markdown item, instead of crashing.

### DIFF
--- a/codalab/rest/interpret.py
+++ b/codalab/rest/interpret.py
@@ -584,9 +584,12 @@ def expand_raw_item(raw_item):
         if is_search:
             keywords = rest_util.resolve_owner_in_keywords(keywords)
             bundle_uuids = local.model.search_bundle_uuids(request.user.user_id, keywords)
-            bundle_infos = rest_util.get_bundle_infos(bundle_uuids)
-            for bundle_uuid in bundle_uuids:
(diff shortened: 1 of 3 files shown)


codalab/rest/interpret.py
```diff
@@ -584,9 +584,12 @@ def expand_raw_item(raw_item):
         if is_search:
             keywords = rest_util.resolve_owner_in_keywords(keywords)
             bundle_uuids = local.model.search_bundle_uuids(request.user.user_id, keywords)
```
percyliang: Yikes this function is pretty confusing...perhaps call it `result` instead `bundle_uuids`.  In the future, we probably want to change this to return a dict with either `bundle_uuids` or `value`...

percyliang (approved):

bkgoksel:
Just changed `search_bundle_uuids` to `search_bundles` and had it return a dict with an `is_aggregate` bool field that is `True` if the result is a scalar and `False` if it's a list of bundles, and with the result stored in a `result` field. Made corresponding changes at its callers.

w0nche0l:
Thanks for the thorough cleanup @bkgoksel.

PR #925, opened 2018-07-30
Create new Codalab Default docker images, auto-choose CUDA image if gpus requested

bkgoksel:
This PR:
1. Adds two new Codalab docker images `codalab/default-cpu` and `codalab/default-gpu` with Dockerfiles version controlled in the main repo. The two images are both based off Ubuntu 16.04, have the same packages and libraries installed, but the GPU version also comes with CUDA Toolkit 9.2 and prefers GPU versions of libraries when possible (as in tensorflow and pytorch)
2. Changes the default docker image of runs to be `codalab/default-cpu:latest`
3. Adds further logic to bundle manager such that if a run doesn't specify its requested docker image but it has requested GPUs, `codalab/default-gpu:latest` is picked, if no GPUs are requested `codalab/default-cpu:latest` is picked.

Need to do:
- [x] Add basic tests for docker images
- [x] Clean up unused references to default_docker_image across the codebase
- [x] Think about documentation changes that might be necessary with this.


codalab/worker/bundle_manager.py
```diff
@@ -458,11 +458,14 @@ def _compute_request_time(self, bundle):
...
         """
         Set docker image to be the default if not specified
-        The default is for backwards compatibilty for
-        runs from before when we added client-side defaults
+        Unlike other metadata fields this can actually be None
+        from client
         """
         if not bundle.metadata.request_docker_image:
-            return 'codalab/ubuntu:1.9'
+            if bundle.metadata.request_gpus:
```
percyliang, line 488: This is nice!

docker/default_image/Dockerfile.cpu
```diff
@@ -0,0 +1,123 @@
+FROM ubuntu:xenial
+MAINTAINER CodaLab Team "codalab.worksheets@gmail.com"
+
+# Installed packages:
```
percyliang: Seems bad to have this list duplicated with the install commands below... can we unify?

docker/default_image/Dockerfile.cpu
```diff
@@ -0,0 +1,123 @@
...
+# dlib
+# OpenAI gym[all]
+# NLTK (no data)
+# tensorflow CPU
+# keras
+# tensorboard
+# pytorch CPU
+# torchvision
+
+RUN apt-get -y update; apt-get -y upgrade;
```
percyliang: Don't need `;` at end of lines...

docker/default_image/Dockerfile.gpu
```diff
@@ -0,0 +1,124 @@
+FROM nvidia/cuda:9.2-runtime
```
percyliang: Is there a way to share commands between GPU and CPU images?

docker/default_image/Dockerfile.cpu
```diff
@@ -0,0 +1,123 @@
+FROM ubuntu:xenial
+MAINTAINER CodaLab Team "codalab.worksheets@gmail.com"
+
+# Installed packages:
+
+# default-jdk
```
percyliang: Is this going to give you Oracle (would be ideal)?

percyliang (commented):
This is great!

docker/default_image/Dockerfile.gpu
bkgoksel: I searched around the internet for a bit to figure this out but there doesn't seem to be a clean way because they are derived from different base images. 

The only way to do this would be to do some command-line shenanigans ourselves to dynamically insert the shared commands into a temporary Dockerfile when building the images, but I don't know if it is worth it.

docker/default_image/Dockerfile.cpu
bkgoksel: Not sure, probably a better idea to specify Oracle directly.

docker/default_image/Dockerfile.cpu
bkgoksel: Can just get rid of the comments and make proper documentation elsewhere with a list of packages. It's probably a good idea to specify exact versions in the file and documentation so we can keep track of changes over time.

docker/default_image/Dockerfile.gpu
percyliang: We can have both docker files have a block that is meant to be identical and put a comment to that effect.

bkgoksel: That was my intention with the `# CPU specific commands below` `# GPU specific commands below` comments but can make that more explicit, especially after getting rid of the huge comment block on top.

PR #926, opened 2018-08-01
Added support for old-style mimic/macro bundle parsing

bkgoksel:
Now for mimic and macro commands, both `worksheet/bundle` and `worksheet//bundle` bundle specs work. However if there is ambiguity where the user uses `cl mimic a/b` and a bundle named `a` exists, `a` will be interpreted as a bundle.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2027,7 +2027,16 @@ class BundleCLI(object):
         Use args.bundles to generate a call to bundle_util.mimic_bundles()
         """
         client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -2027,7 +2027,16 @@ def mimic(self, args):
         Use args.bundles to generate a call to bundle_util.mimic_bundles()
         """
         client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
-        bundle_uuids = self.resolve_bundle_specs(client, worksheet_uuid, args.bundles)
+        try:
+            bundle_uuids = self.resolve_bundle_specs(client, worksheet_uuid, args.bundles)
```
percyliang: `resolve_bundle_specs` is kind of a confusing name...it calls `resolve_target` which calls `resolve_bundle_uuids`...can we rename / document this more clearly?

PR #930, opened 2018-08-06
Fail bundles with killed dependencies if --allow-failed-dependencies not specified

bkgoksel:
Right now these bundles are left in `created` stage forever.

This PR has a large diff because of auto code formatting.

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -14,6 +14,7 @@ from codalab.lib import bundle_util, formatting, path_util
 from codalabworker.file_util import remove_path
 from codalabworker.bundle_state import State
 
+
 logger = logging.getLogger(__name__)
 
 WORKER_TIMEOUT_SECONDS = 60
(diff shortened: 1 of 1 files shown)


codalab/worker/bundle_manager.py
```diff
@@ -472,11 +526,16 @@ def _construct_run_message(self, worker, bundle):
...
+        message['bundle'] = bundle_util.bundle_to_bundle_info(
+            self._model, bundle)
         if self._worker_model.shared_file_system and worker['user_id'] == self._model.root_user_id:
-            message['bundle']['location'] = self._bundle_store.get_bundle_location(bundle.uuid)
+            message['bundle'][
+                'location'] = self._bundle_store.get_bundle_location(
+                    bundle.uuid)
             for dependency in message['bundle']['dependencies']:
-                dependency['location'] = self._bundle_store.get_bundle_location(dependency['parent_uuid'])
+                dependency[
```
percyliang: This looks really ugly...can we not do this?

percyliang (commented):
In general, can we separate the code formatting PRs from the functional changes?

bkgoksel:
My bad, I installed an auto code formatter (https://github.com/google/yapf) to use for my other projects but forgot to turn it off while working on codalab. I can undo these changes and re-commit with just the semantic changes.

In general how do you feel about configuring an auto code formatter and generally enforcing it on all the commits? YAPF looks pretty customizable and we can try to configure it to avoid ugly changes like the earlier. It might help with long term consistency of code style. I think we can even add a CI check with it for PRs and everything. It also allows us to reformat the existing codebase while guaranteeing no functional changes will be made.

percyliang:
I think it's a reasonable thing to do...we can just make a separate PR to
reformat all the code.  Maybe send one out so at least we can see what it
will look like.

PR #932, opened 2018-08-07
Switch to auto-formatted code with Black

bkgoksel:
Black(https://github.com/ambv/black) is a more opinionated code formatter than YAPF. It's less customizable than YAPF but seems to produce better results by default. It also checks against functionality regressions by constructing ASTs for code before and after formatting and asserting nothing has changed.

It requires Python 3.6 to run and expects code compliant with Python3 syntax, so this PR also includes changes that make our codebase syntax-wise compliant with Python 3 (mainly print calls with parentheses, `except Exception as e` instead of `except Exception, e` and converting octets from `0770` to `0o770`)


alembic/versions/2014080613_add_owner_id_to_bundle_341ee10697f1.py
```diff
@@ -7,7 +7,7 @@
 """
 
 # revision identifiers, used by Alembic.
-revision = '341ee10697f1'
+revision = "341ee10697f1"
```
percyliang: Can we just stick with single quotes throughout?

alembic/versions/2014090815_switching_owner_id_to_str_143201389156.py
```diff
@@ -7,24 +7,60 @@
...
 from alembic import op
 import sqlalchemy as sa
 
 
 def upgrade():
     ### commands auto generated by Alembic - please adjust! ###
-    op.alter_column("bundle", 'owner_id', type_=sa.String(255), existing_type=sa.Integer(), nullable=True)
-    op.alter_column("worksheet", 'owner_id', type_=sa.String(255), existing_type=sa.Integer(), nullable=True)
-    op.alter_column("group", 'owner_id', type_=sa.String(255), existing_type=sa.Integer(), nullable=True)
+    op.alter_column(
```
percyliang: Make it not expand?

bkgoksel:
Started from scratch, switched off string quotation change.

My opinion is that having auto formatting by default and manually changing formatting around the code base as we run into non-optimal things is a better approach than the opposite.

percyliang:
I agree with that generally.

How do we make sure the exceptional changes don't get clobbered by the
automatic formatting?

We should tweak things so that the automatic changes are as good as
possible to minimize the gap.

bkgoksel:
Any code wrapped in a pair of `# fmt: off` `# fmt: on` comments doesn't get modified or flagged as not compliant with formatting. This is mainly useful for things like argument lists where we would tolerate going over the line length limit for compactness/uniformity.

I was talking to a friend who is trying to set up auto formatting for a large python code base where he works and he also claimed that Black was the best one they've tried, but unfortunately it doesn't offer any customization beyond string quote standardization (on or off, off for us) and line length limit (default 88, 100 for us)

codalab/bundles/derived_bundle.py
```diff
@@ -10,31 +10,41 @@
 
 class DerivedBundle(NamedBundle):
     METADATA_SPECS = list(NamedBundle.METADATA_SPECS)
-    METADATA_SPECS.append(MetadataSpec('allow_failed_dependencies', bool, 'Whether to allow this bundle to have failed or killed dependencies.', default=False))
+    # Don't format specs
+    # fmt: off
+    METADATA_SPECS.append(
```
percyliang, line 15: Thought turning formatting off would leave it one line.

codalab/client/json_api_client.py
```diff
@@ -1,14 +1,11 @@
 import httplib
 import socket
 import sys
+import six
```
percyliang, line 4: This seems to be semantics changing...

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -20,15 +20,26 @@ class LocalRunManager(BaseRunManager):
...
     # Network buffer size to use while proxying with netcat
     NETCAT_BUFFER_SIZE = 4096
     # Number of seconds to wait for bundle kills to propagate before forcing kill
     KILL_TIMEOUT = 100
     # Directory name to store running bundles in worker filesystem
     BUNDLES_DIR_NAME = 'runs'
 
-    def __init__(self, worker, docker, image_manager, dependency_manager,
-                 commit_file, cpuset, gpuset, work_dir, docker_network_prefix='codalab_worker_network'):
+    def __init__(
```
percyliang, line 31: Is there a way to prevent this from happening?

percyliang (approved):
Generally looks good.  Don't like everything, but I think it's net positive.

codalab/bundles/derived_bundle.py
bkgoksel: I just hand modified those back, forgot to put the append and the actual spec on the same line, can do that.

codalab/client/json_api_client.py
bkgoksel: This was a manual change I made to make our code base syntax compatible with both Python 2 and 3 as per https://pythonhosted.org/six/index.html#six.reraise
 Python 3 syntax compatibility is necessary for this formatter as it builds ASTs and compares them to ensure functionality stays the same but does so in Python 3.6

worker/codalabworker/local_run/local_run_manager.py
bkgoksel: This library has a strong tendency to start argument lists on new lines and separating each argument into its own line. I read that their rationale behind this style is to have minimal/clearer diffs when arguments change.
I'm personally neutral about this approach but if it's not something we want to adopt then we should probably reconsider using Black as it is really unavoidable and will probably happen anywhere many arguments are passed.

percyliang: I think that's logical.

PR #937, opened 2018-08-13
Add a kill when idle flag to the worker

bkgoksel:
When the flag is specified, the worker checks whether it has any assigned/unfinished runs before each check-in (except for the first check-in).

If there are no more runs, it quits.

The Slurm worker invokes one-off worker scripts for the user whenever there are staged jobs for that user. There are three separate kinds of workers: CPU(john), low-priority GPU (jag-low) and high-priority GPU (jag-hi).

### DIFF
--- a/worker/codalabworker/worker.py
+++ b/worker/codalabworker/worker.py
@@ -31,7 +31,7 @@ class Worker(object):
         worker_id,  # type: str
         tag,  # type: str
         work_dir,  # type: str
-        kill_when_idle,  # type: str
+        exit_when_idle,  # type: str
         bundle_service,  # type: BundleServiceClient
     ):
         self.id = worker_id
@@ -39,7 +39,7 @@ class Worker(object):
         self._tag = tag
         self._work_dir = work_dir
         self._bundle_service = bundle_service
-        self._kill_when_idle = kill_when_idle
+        self._exit_when_idle = exit_when_idle
         self._stop = False
         self._last_checkin_successful = False
         self._run_manager = create_run_manager(self)
@@ -51,7 +51,7 @@ class Worker(object):
                 self._run_manager.process_runs()
                 self._run_manager.save_state()
                 if (
-                    self._kill_when_idle
+                    self._exit_when_idle
                     and self._last_checkin_successful
                     and len(self._run_manager.all_runs) == 0
                 ):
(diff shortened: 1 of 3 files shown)


scripts/slurm-worker.py
```diff
@@ -0,0 +1,94 @@
...
+
+def main():
+    # Log in with the server
+    subprocess.check_call('cl work {}::'.format(SERVER_INSTANCE), shell=True)
+    print('Logged in to {}'.format(SERVER_INSTANCE))
+
+    # Infinite loop until user kills us
+    num_runs = 0
+    while True:
+        run_lines = subprocess.check_output('cl search .mine state=staged', shell=True)
```
percyliang: use `-u`

scripts/slurm-worker.py
```diff
@@ -0,0 +1,94 @@
...
+                worker_home = '~/worker-{}'.format(num_runs)
+                num_runs += 1
+                worker_invocation = '{} --work-dir {} --tag jag-hi'.format(BASE_WORKER_INVOCATION, worker_home)
+                prepare_command = 'mkdir {}'.format(worker_home)
+                cleanup_command = 'rm -rf {}'.format(worker_home)
+                final_command = '{} \'{}; {}; {}\''.format(run_command, prepare_command, worker_invocation, cleanup_command)
+                print(final_command)
+                subprocess.check_call(final_command, shell=True)
+                print('Started worker for run {uuid}'.format(**run))
+            for run in jag_lo_runs:
```
percyliang: refactor to avoid code duplication

worker/codalabworker/worker.py
```diff
@@ -24,12 +24,22 @@
...
 class Worker(object):
-    def __init__(self, create_run_manager, commit_file, worker_id, tag, work_dir, bundle_service):
+    def __init__(
+        self,
+        create_run_manager,  # type: Callable[[Worker], BaseRunManager]
+        commit_file,  # type: str
+        worker_id,  # type: str
+        tag,  # type: str
+        work_dir,  # type: str
+        kill_when_idle,  # type: str
```
percyliang: `exit_when_idle`

bkgoksel:
Separated the `exit_when_idle` addition to the worker and the Slurm worker as I'd like to deploy a new  version soon while bringing the workers down to clean up all old docker images.

PR #939, opened 2018-08-14
Slurm worker

bkgoksel:
Script that loops waiting for a user's bundles and fires one-use worker on slurm every time there's a run. Also handles translating codalab flags to slurm flags and Stanford NLP cluster specific queues.

Maybe we don't want this script as part of the main repo but opening the PR to track its status.

### DIFF
--- a/scripts/slurm-worker.py
+++ b/scripts/slurm-worker.py
@@ -529,6 +529,20 @@ class SlurmWorkerDaemon(Daemon):
         else:
             return val
 
+    def prune_logs(self):
+        try:
+            pf = open(self.pidfile, 'r')
+            pid = int(pf.read().strip())
+            pf.close()
+        except IOError:
+            pid = None
+        if pid:
+            message = "Daemon running, please stop before pruning.\n"
+            sys.stderr.write(message)
+            sys.exit(1)
+        else:
+            shutil.rmtree(self.log_dir)
+
     @staticmethod
     def list_instances(root_dir):
         print('Slurm worker daemons:')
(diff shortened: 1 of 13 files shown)


scripts/slurm-worker.py
```diff
@@ -0,0 +1,105 @@
...
+        nlprun_binary,
+        '-q {}'.format(queue),
+        '-c {}'.format(run_fields['num_cpus']),
+        '-r {}'.format(run_fields['request_memory']),
+        '-g {}'.format(run_fields['request_gpus']),
+        '-p {}'.format(priority),
+    ]
+
+    run_command = ' '.join(nlprun_flags)
+    worker_home = '~/worker-{}'.format(run_number)
```
percyliang: make this prefix a defined constant

scripts/slurm-worker.py
```diff
@@ -0,0 +1,105 @@
+import math
+import subprocess
+import time
+
+from codalab.lib.formatting import parse_duration
+
+SERVER_INSTANCE = 'https://worksheets-dev.codalab.org'
+NLPRUN_BINARY = '/u/nlp/bin/nlprun'
+CL_WORKER_BINARY = 'u/nlp/bin/cl-worker'
```
percyliang: missing `/`?

scripts/slurm-worker.py
```diff
@@ -0,0 +1,105 @@
+import math
+import subprocess
+import time
+
+from codalab.lib.formatting import parse_duration
```
percyliang: Document what this file does.  Perhaps rename this file to `nlp-slurm-worker.py`?  Or we can call `srun` directly instead of going through `nlprun` to make it more generic...

PR #942, opened 2018-08-20
Dependency manager hotfixes

bkgoksel:
Fix dict sorting in local dependency manager to get rid of key error, drastically increase serialized length limit of dependency cache as it fires too often, grinding the system to a halt as it is currently. 

Redo concurrency in dependency manager to reduce lock contentions.

### DIFF
--- a/worker/codalabworker/main.py
+++ b/worker/codalabworker/main.py
@@ -70,9 +70,8 @@ def main():
     parser.add_argument(
         '--max-dependencies-serialized-length',
         type=int,
-        default=60000,
-        help='Maximum length of serialized json of dependency list of worker '
-        '(e.g., 50, 30000, 60000).',
+        default=10000000,
+        help='Maximum length of serialized json of dependency list of worker in bytes',
     )
     parser.add_argument(
         '--max-image-cache-size',
--- a/worker/codalabworker/local_run/docker_image_manager.py
+++ b/worker/codalabworker/local_run/docker_image_manager.py
@@ -117,12 +117,12 @@ class DockerImageManager(StateTransitioner, BaseDependencyManager):
                     }
                     if failed_images:
                         digest_to_remove = min(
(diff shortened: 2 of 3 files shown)


worker/codalabworker/main.py
```diff
@@ -70,9 +70,8 @@ def main():
     parser.add_argument(
         '--max-dependencies-serialized-length',
         type=int,
-        default=60000,
-        help='Maximum length of serialized json of dependency list of worker '
-        '(e.g., 50, 30000, 60000).',
+        default=10000000,
```
percyliang, line 73: How come this wasn't an issue before?

percyliang (approved):

worker/codalabworker/main.py
bkgoksel: potential reasons:
- old worker stored less state about dependencies, allowing more dependencies to be serialized in this size
- it did fire up a lot but because the system was more threaded constantly cleaning up dependencies didn't trip it up.

I don't really know at this point, and I'm not even convinced this is the reason we're having issues but looking at logs I saw this getting triggered too frequently and there's no reason to limit this to something like 58K if the actual dependencies themselves aren't taking too much space. I'm not even convinced we need this at all but didn't want to risk removing it.

fabeschan:
@bkgoksel @percyliang iirc when I wrote it I set it to 60k because approximately the number of bytes available for a mysql entry of that datatype. If the string is longer than the limit it would cause the worker to crash when trying to deserialize it

bkgoksel:
@fabeschan But that doesn't apply anymore because we serialize to JSON instead of Mysql now right? Or are worker dependencies also stored in a DB somewhere?

fabeschan:
@bkgoksel at one point we were serializing to JSON as a mysql entry and .json file; whether or not this is still the case I'm not sure.

PR #945, opened 2018-08-28
Reduce max-serialized-length and turn from argument into constant

bkgoksel:
Remove the max_serialized_length worker argument and turn it into a constant (as this is a server-side hard limit right now and we don't want people changing it). Also add a TODO to find a better way around this

### DIFF
--- a/worker/codalabworker/local_run/local_dependency_manager.py
+++ b/worker/codalabworker/local_run/local_dependency_manager.py
@@ -41,7 +41,7 @@ class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager)
     """
 
     DEPENDENCIES_DIR_NAME = 'dependencies'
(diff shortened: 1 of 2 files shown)


worker/codalabworker/local_run/local_dependency_manager.py
```diff
@@ -41,10 +41,14 @@ class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager)
     """
 
     DEPENDENCIES_DIR_NAME = 'dependencies'
-
-    def __init__(
-        self, commit_file, bundle_service, worker_dir, max_cache_size_bytes, max_serialized_length
-    ):
+    # TODO:bkgoksel: The server writes these to the worker_dependencies table, which stores the dependencies
```
percyliang: nit: `TODO(bkgoksel)`

percyliang (approved):

PR #946, opened 2018-08-29
Switch to Docker Python SDK

bkgoksel:


worker/codalabworker/docker_client.py
```diff
@@ -434,74 +311,36 @@ def start_container(
...
-            )
-            create_response = create_conn.getresponse()
-            if create_response.status != 201:
-                raise DockerException(create_response.read())
-            container_id = json.loads(create_response.read())['Id']
-
-        # Start the container.
-        logger.debug(
-            'Starting Docker container for UUID %s with command %s, container ID %s,\n \t\t volume bindings: %s',
+        container_id = self.create_bundle_container(
```
percyliang, line 351: Can we used named arguments to avoid bugs?

worker/codalabworker/docker_client.py
```diff
@@ -590,35 +413,16 @@ def get_logs(self, container_id):
...
-
-    def connect(self):
-        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        self.sock.settimeout(300)
-        self.sock.connect('//var/run/docker.sock')
+        contents = self._client.logs(stdout=True, stderr=True)
+        stderr = []
+        stdout = []
+        start = 0
+        while start < len(contents):
```
percyliang: I'm surprised we still need low-level stuff like this if we're using the docker API.  Add a comment?

percyliang (approved):
Looks a lot cleaner.  LGTM assuming everything works as intended (I feel like there are always subtle differences when such sweeping changes are made).

worker/codalabworker/docker_client.py
bkgoksel: Good point, will do.

worker/codalabworker/docker_client.py
bkgoksel: Yeah their documentation is pretty flaky and it just says the API returns a string but it's unclear how the stderr and stdout are separated. I was planning to actually call their logs command against a live container to double check that the python sdk format is the same as the http API format.

This function also isn't used anywhere anymore so I might just remove it.

bkgoksel: Removed the function, the API itself seems to return well formed output.

PR #953, opened 2018-09-13
Overwrite existing path instead of failing when downloading dependencies

### LINKED ISSUE #951: Dependency manager fails dependency downloads if untracked files by the same name exist
Currently the dependency manager fails the dependency download if a folder by the same name as the dependency exists in local disk but the dependency itself is not found in the state dict. What we should do instead is to overwrite the un-tracked files and re-download the dependency files (to err on the side of caution)

bkgoksel:
Fixes #951

### DIFF
--- a/worker/codalabworker/local_run/local_dependency_manager.py
+++ b/worker/codalabworker/local_run/local_dependency_manager.py
@@ -310,6 +310,9 @@ class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager)
         """
         Copy the dependency fileobj to its path in the local filesystem
         Overwrite existing files by the same name if found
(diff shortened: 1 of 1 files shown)


percyliang (approved):
Why would the path already exist?  Some inconsistency.  Please explain in a comment.

bkgoksel:
The only way this can happen is if the state file gets modified/deleted and the worker is restarted. We don't have an automated workflow to completely clear these directories on updates so while I try to make sure no inconsistencies are left after updates I guess it happened j

bkgoksel:
Another solution to this issue would be to delete any files not in the committed state on start

percyliang:
I think it's good to be robust to arbitrary changes on the file system.

PR #954, opened 2018-09-13
Discard failed dependencies after 10 seconds

### LINKED ISSUE #952: Failed dependencies stay failed for a long time
If a requested dependency fails, any future run requesting the same dependency and assigned to the same worker will fail immediately potentially infinitely (unless the dependency cache is cleared and the failed entry for the dependency is removed). Instead we want to clear out failed dependencies after a while and retry if they are ever requested again.

bkgoksel:
Fixes #952


percyliang (approved):
Can you add comments on the motivation for needing this?

PR #958, opened 2018-09-14
Don't delete docker images if needed by a current run.

### LINKED ISSUE #956: Docker image manager cleanup can hang run bundles indefinitely
At least due to #943 and #955 sometimes the docker image manager gets to a state where disk use > size limit but with no stale cached docker images to cleanup. In these cases, the next time a run is assigned, the docker image manager's clenaup process deletes the docker image required for that run as soon as it is downloaded (since it is set to delete the earliest image it can find, which happens to be the last one). The run manager then checks whether the image exists, and finding that it doesn't, starts its download once more, restarting this cycle.

We should at least implement a mechanism in the docker image manager similar to the one in the dependency manager that allows the run manager to register deregister runs to individual images so that images that are currently required by some run are not deleted.

bkgoksel:
Make get/release a core part of dependency managers, implement it in docker image manager as well.
Fixes #956

### DIFF
--- a/worker/codalabworker/fsm.py
+++ b/worker/codalabworker/fsm.py
@@ -50,10 +50,18 @@ class BaseDependencyManager(object):
         """ Return whether or not the corresponding DependencyState exists in the manager """
         raise NotImplementedError
 
-    def get(self, dependency):
+    def get(self, uuid, dependency):
         """
         Start downloading the corresponding dependency if not already in progress.
+        Register that the given uuid is a dependent of this dependency.
         Return the corresponding DependencyState.
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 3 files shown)


percyliang (approved):
Are there cases where something crashes and we're left with an image that we can't delete?

bkgoksel:
It is possible but such a situation would require manual intervention anyways (would be a larger crash with a run stuck in a non terminal state), and as long as the run is resumed the image would be deleted (alternatively if the worker cannot be resumed/needs to be hard reset, we'd want to reset the images and dependencies as well anyways) 
Furthermore the same risk exists for dependencies right now anyways and hasn't been an issue afaik.

PR #962, opened 2018-09-26
Encode filenames if non-ASCII found

bkgoksel:
Adds the fix mentioned in https://github.com/fabric/fabric/issues/1292#issuecomment-82621653


bkgoksel:
Merging for now but as a note, I didn't find a minimal sample to reproduce this and on basic tests os.path.join (the normally failing call) deals with non-ASCII characters in filenames just fine, so clearly there's something more going on here. Don't have time to investigate further and this is a rare issue so merging this and closing the issue for now.

PR #967, opened 2018-09-30
Gracefully skip when key not found in print attribute

### LINKED ISSUE #963: cl uinfo crashes
When trying to do `cl uinfo` on any user that's not yourself crashes with ```  File "/home/pliang/.local/lib/python2.7/site-packages/codalab/lib/bundle_cli.py", line 3528, in print_attribute
    print >>self.stdout, u'{:<15}: {}'.format(key, value).encode('utf-8')
UnboundLocalError: local variable 'value' referenced before assignment```

bkgoksel:
Fixes #963

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -3520,9 +3520,9 @@ class BundleCLI(object):
                     else:
                         value = user[key]
                 except KeyError:
-                    pass
+                    return
             else:
-                value = user.get(key, None)
+                value = user[key]
 
             if should_pretty_print:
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -3520,9 +3520,9 @@ def print_attribute(key, user, should_pretty_print):
                     else:
                         value = user[key]
                 except KeyError:
-                    pass
+                    return
```
percyliang, line 3523: nit: `return None` to make it explicit

percyliang (approved):

PR #968, opened 2018-09-30
Make files open in browser instead of download.

w0nche0l:
Changing Content Disposition metadata when we serve so that they can be opened in the browser.

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -51,7 +51,6 @@ def _fetch_bundle(uuid):
        for displaying the bundle info, `0` to omit them. Default is `0`.
      - `include`: comma-separated list of related resources to include, such as "owner"
     """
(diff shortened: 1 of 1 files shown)


codalab/rest/bundles.py
```diff
@@ -51,6 +51,7 @@ def _fetch_bundle(uuid):
        for displaying the bundle info, `0` to omit them. Default is `0`.
      - `include`: comma-separated list of related resources to include, such as "owner"
     """
+    print 'bundles?'
```
percyliang: remove

percyliang (approved):
What happens with binary files (e.g., images)?

w0nche0l:
@percyliang they get opened in the browser as well, if the files are openable by the browser. If they're something like a .zip file, they'll still get downloaded.

PR #969, opened 2018-10-01
pip install worker in venv on server-side setup and get rid of worker.sh

### LINKED ISSUE #964: worker.sh doesn't work out of the box
Currently worker.sh doesn't work out of the box unless `codalabworker` is pip installed.
We should probably install the worker folder with pip in the `setup.sh` script. We should also probably get rid of `worker.sh` and promote the use of the `cl-worker` shortcut to launch workers.

bkgoksel:
Fixes #964 (and gets rid of worker.sh, so after deploying this we should update all our documentation to indicate people should use the `cl-worker` command to launch the worker)


setup.sh
```diff
@@ -47,6 +48,7 @@ $env/bin/pip install -U setuptools
 echo -e "${info}[*] Installing Python packages into $env...${reset}"
 if [ "$1" == "server" ]; then
   $env/bin/pip install -r $codalabdir/requirements-server.txt
+  $env/bin/pip install -e $workerdir
```
percyliang: Nit: Why define `workerdir` if it's going to be used once?  If put `$codalabdir/worker` here directly, it's clearer that they're both in `$codalabdir`...

percyliang (approved):

PR #970, opened 2018-10-01
Force worker to exit when authentication error occurs

w0nche0l:
If worker has an authentication error, then throw an exception that also marks the worker as stopped.  
I'm a little iffy about the implementation on this one, @percyliang @bkgoksel, any suggestions? Is there a cleaner way to achieve this that doesn't use exceptions as a de-facto message passing mechanism?

### DIFF
--- a/worker/codalabworker/bundle_service_client.py
+++ b/worker/codalabworker/bundle_service_client.py
@@ -39,10 +39,12 @@ def wrap_exception(message):
 
     return decorator
 
+
 class BundleAuthException(RestClientException):
     """
     Exception raised by the BundleServiceClient methods if auth error occurs.
     """
+
 
 class BundleServiceException(RestClientException):
     """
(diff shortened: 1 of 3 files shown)


bkgoksel (changes requested):
I think it would be better if the `BundleServiceClient` raised an error immediately if initialized with invalid username-password arguments. 

I'm thinking calling `self._get_access_token` as the last step of the `init` method and `try ... catch` the authorization error there, if it fails, just raise the error from within `init`. That way we guarantee that any `BundleServiceClient` object will have valid credentials.

That way instead of allowing the worker to boot and start its loop before signaling out, we can catch the authorization error in `worker/codalabworker/main.py#144` and quit immediately (or ask the user for a username/password combo once more if we prefer that)

bkgoksel:
Although thinking back I'm not sure if asking for a password again is the best option here.  We also allow the specification of a password file (and even specifying credentials as environment variables) and if the credentials there are invalid I don't think there's anything we can do except for quitting to allow the user to update that file (or their environment variables)

bkgoksel:
Pushed the changes I suggested earlier.

bkgoksel (approved):

PR #975, opened 2018-10-04
Make Reader use a dict of handlers

fabeschan:


worker/codalabworker/run_manager.py
```diff
@@ -120,19 +120,22 @@ def memory_bytes(self):
 
 
 class Reader(object):
+
+    read_handlers = {
```
bkgoksel: `self` is not defined in this context as handler is a class variable. As a result this doesn't work in its current form. I guess one way to make it work would be to build this handle dict in `__init__` and make it an instance variable.

bkgoksel (changes requested):
Good refactor but doesn't work as it is.

worker/codalabworker/run_manager.py
fabeschan: good idea

fabeschan: done

bkgoksel:
Looks good! Thanks for this PR!

bkgoksel (approved):

PR #976, opened 2018-10-06
Make sure run_status is updated when finalizing/finishing bundle

fabeschan:
Problem:
When a bundle is being finalized, its metadata.run_status is not
updated to reflect that it is "Finalizing Bundle", in fact leaving
the run_status to be whatever it was the last time resume_bundle
was called.

This forever leaves run_status as typically one of "Upload started"
or "Uploading results: yyy done (archived size)", neither of which
look good and worse, may confuse an inexperienced user.
    
Solution:
Update the bundle metadata using resume_bundle when
finalizing it, and update metadata again during finish_bundle.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -786,7 +786,12 @@ class BundleModel(object):
             ).fetchone()
             if not row:
                 return False
+
             if state == State.FINALIZING:
+                # update bundle metadata using resume_bundle one last time before finalizing it
+                self.resume_bundle(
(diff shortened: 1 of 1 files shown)


fabeschan:
Example from https://worksheets.codalab.org/bundles/0xa8480f2ea31e4cea87e7bb3d25a34fc6/

screenshot 2018-10-05 23 05 11

bkgoksel (approved):
Been meaning to investigate/fix this for a while, thanks for the PR. LGTM.

bkgoksel (approved):

fabeschan:
np

PR #979, opened 2018-10-11
Enforce style in Travis CI

bkgoksel:
Our style checker (ambv/black) is an automated formatter as well, so to take full advantage we should ensure our master always conforms to it. This PR adds its style check to our Travis builds.


bkgoksel:
Bad name, bad base branch.

PR #983, opened 2018-10-15
Fix/940 bundle run time

nikhilxb:
- Make calculation of bundle run time consistent
- Refactor bundle state transition methods


codalab/model/bundle_model.py
```diff
@@ -680,7 +664,7 @@ def set_starting_bundle(self, bundle, user_id, worker_id):
 
             bundle_update = {
                 'state': State.STARTING,
-                'metadata': {'last_updated': int(time.time())},
+                'metadata': {},
```
bkgoksel: No need to have an empty dict here, just remove.

codalab/model/bundle_model.py
```diff
@@ -713,7 +696,7 @@ def set_offline_bundle(self, bundle):
 
             bundle_update = {
                 'state': State.WORKER_OFFLINE,
-                'metadata': {'last_updated': int(time.time())},
+                'metadata': {},
```
bkgoksel: ditto

bkgoksel (changes requested):
Generally LGTM. Will do a second pass after the refactoring we talked about.

bkgoksel:
Oh 2 more things:
1. Install https://github.com/ambv/black and make sure the code is properly formatted (make sure black uses the pyproject.toml file for configuration)
2. Resolve the conflicts with master so it can be auto merged

codalab/model/bundle_model.py
```diff
@@ -689,10 +705,9 @@ def set_starting_bundle(self, bundle, user_id, worker_id):
 
             return True
 
-    def set_offline_bundle(self, bundle):
+    def transition_bundle_offline(self, bundle):
```
bkgoksel: To be completely proper let's call this `transition_bundle_worker_offline`

codalab/model/bundle_model.py
```diff
@@ -821,39 +812,40 @@ def resume_bundle(self, bundle, bundle_update, row, user_id, worker_id, hostname
...
 
         self.update_bundle(
-            bundle, {'state': bundle_update['state'], 'metadata': metadata_update}, connection
+            bundle, bundle_update, connection
         )
 
         self.update_events_log(
             user_id=bundle.owner_id,
             user_name=None,  # Don't know
-            command='resume_bundle',
```
bkgoksel, line 845: I don't know how events_log works but let's make sure we don't break backwards compatibility here. @percyliang thoughts?

bkgoksel (dismissed):
LGTM - the nits, formatting, rebasing and the backward compatibility question regarding the events log

codalab/model/bundle_model.py
bkgoksel: This seems OK!

bkgoksel:
Should move the applicable metadata fields from RunBundle to DerivedBundle to fix the tests

bkgoksel (approved):
LGTM as long as the builds pass! 

Looks like there are some unrelated files that were changed by the autoformatter black when you ran it. Ideally we wouldn't have these but looks like black recently updated some small rules and that got some stuff changed. I'm planning to freeze our version of black at what it is currently to avoid this in the future but I think it's fine if this one PR touches a couple extra files.

(BTW this is the exact reason I want to enforce black at travis-build level so the same version of black etc are guaranteed to be ran at the end of the day before any commit is merged, and we should never get these kinds of unrelated changes in PRs.)

bkgoksel:
Broken up into up-to-date PRs, unneeded now.

PR #995, opened 2018-11-08
Fix/994

bkgoksel:
Automatically sets the `source_url` metadata field of upload bundles to be the source url (or string representation of list of source urls if multiple) if uploaded as a URL.


percyliang (approved):
I don't think no one knows about `source_url`, but I guess it doesn't hurt to add it.

bkgoksel:
We can advertise it in the future! I think someone asked at NLP lunch how could they know what version/in what state a certain dataset bundle they've found is. This would be a good way for people to know the provenance of a dataset bundle in a bundle-reuse heavy future.

PR #1001, opened 2018-11-26
CLI fixes

### LINKED ISSUE #990: Update `cl add` to not require current worksheet
Should make use of the current worksheet context by default.

### LINKED ISSUE #993: Current worksheet printing can be mistaken for URL
currently cl work prints `Currently on https://worksheets.codalab.org::home-user` which looks like a URL but is actually an `instance::worksheet` spec. Let's make it clear that people shouldn't try to go to that URL.

nikhilxb:
- Closes #456: web cl alias Internal Server Error
- Closes #990: cl add not require current worksheet
- Closes #993: cl work now prints proper URL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ codalab/tests
 codalab/tags
 notes.otl
 codalab-worker-scratch/
+frontend/node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -57,7 +57,7 @@ dist/
 build/
 codalab_env/
 *.egg-info
-venv/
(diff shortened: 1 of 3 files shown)


bkgoksel:
Seems like there's only 1 failing test left here and we can merge this after that is fixed:
`cl run source:http://localhost:2900::0xb5cb467a56d74c6a9b54b6b2fe1aa8d2 //temp-test-cli-639396 cat source/stdout -n temp-test-cli-780945 (exit code 1, expected 0)`

Issue #829, opened 2018-01-12
UnicodeDecodeError: 'ascii' codec


fabeschan:
`2018-01-12 02:52:33,223 Received read message: {u'uuid': u'0x754c7cf74e5c416d9af1330692617d58', u'socket_id': 3103506, u'meta': {u'version': u'0.2.21'}, u'path': u'stderr', u'read_args': {u'truncation_text': u'\n... [truncated] ...\n\n', u'num_tail_lines': 50, u'type': u'summarize_file', u'max_line_length': 128, u'num_head_lines': 50}, u'type': u'read'}
Traceback (most recent call last):
  File "/opt/worker/codalabworker/run.py", line 393, in read
    read_args['truncation_text']))
  File "/opt/worker/codalabworker/file_util.py", line 210, in summarize_file
    return ''.join(lines)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 24: ordinal not in range(128)`

bkgoksel:
Seems related to #256

bkgoksel:
Seems related to #502

Issue #961, opened 2018-09-26
Non-ASCII unicode characters break everything


bkgoksel:
We regularly have UnicodeDecodeError issues where users manage to get non-ASCII characters into bundles, commands etc. These issues spring all around our codebase and are usually catastrophic in nature. (all workers down, degenerate worksheets that can't be ever opened again etc.) For some of these we have localized fixes but it would be good to handle these consistently everywhere.

Issue #502, opened 2016-06-15
Correctly and consistently handle Unicode characters


kashizui:
There are a lot of places in the codebase that crash due to unicode characters. Fixing them one-by-one is probably a bad idea and will lead to even flakier code. We need to try to think of a general way to handle Unicode characters properly everywhere.

- [ ] Encode all strings as UTF-8 before printing in BundleCLI? (Non-encoded Unicode characters cause an exception to be raised by the StringIO instance that buffers output when using the web terminal.)
- [ ] Unicode in web terminal commands cause error (see below)
- [ ] Worksheet markdown should be unicode safe
- [x] Unicode characters in bundle commands currently crash the worksheet when generating the raw lines (codalab/codalab-worksheets#300)
- [ ] Change `String` columns to `Unicode` as necessary for automatic conversion
- [ ] and others
- [ ] Add tests a bunch of tests for adding and displaying Unicode content so that bugs don't happen again

kashizui:
Adding another:
```
POST /api/worksheets/command/

Form params:

{"worksheet_uuid":"0x356dfea3aa93477ebc5c04d33c661ec6","raw_command":{"k":"title","v":"run â€˜dateâ€™","action":"worksheet-edit"}}=''

Local variables:
action=u'worksheet-edit'
raw_command_map={u'action': u'worksheet-edit', u'k': u'title', u'v': u'run \u2018date\u2019'}
key=u'title'
value=u'run \u2018date\u2019'

Traceback (most recent call last):
  File "/home/azureuser/codalab-cli/codalab/server/rest_server.py", line 146, in wrapper
    return callback(*args, **kwargs)
  File "/home/azureuser/codalab-cli/venv/local/lib/python2.7/site-packages/bottle.py", line 1732, in wrapper
    rv = callback(*a, **ka)
  File "/home/azureuser/codalab-cli/codalab/server/json_api_plugin.py", line 17, in wrapper
    result = callback(*args, **kwargs)
  File "/home/azureuser/codalab-cli/codalab/rest/legacy.py", line 712, in post_worksheets_command
    data['command'] = service.get_command(data['raw_command'])
  File "/home/azureuser/codalab-cli/codalab/rest/legacy.py", line 540, in get_command
    return worksheet_util.get_worksheet_info_edit_command(raw_command_map)
  File "/home/azureuser/codalab-cli/codalab/lib/worksheet_util.py", line 101, in get_worksheet_info_edit_command
[...]

PR #1003, opened 2018-11-29
Unicode tests

nikhilxb:


setup.py
```diff
@@ -12,7 +12,7 @@
 
 
 def get_requirements(*requirements_file_paths):
-    requirements = ['codalabworker>=0.2.32']
+    requirements = ['codalabworker>=0.2.36']
```
bkgoksel: We may as well make this 0,2,38 since that's the latest version

tests/files/你好世界😊.txt
```diff
@@ -0,0 +1 @@
+This file has unicode in its filename!
```
percyliang: I don't think we should have weird characters in the filename since I think it screws up Windows installations I believe.  We can put it in a zip file though.

.travis.yml
```diff
@@ -39,3 +39,20 @@ before_script:
 script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then ./venv/bin/black . --check; fi
   - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then ./venv/bin/python test-cli.py default; fi
+  - mysql -e "CREATE DATABASE codalab_bundles;"
```
bkgoksel: The steps that are not directly the test but that are part of the setup are moved to the `before_script` section of this file. As you can see above we already have all the commands like `CREATE DATABASE` there. Please don't re-do them here. 

As far as I can tell we just need to add another line to run your worker tests like the following:
`if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then ./test-worker.sh; fi`

The reason to use if statements is to only run tests on Python 2 (and only run the format checker on Python3)

test-cli.py
```diff
@@ -97,20 +97,20 @@ def get_uuid(line):
     return m.group(1)
 
 
-def sanitize(string):
+def sanitize(string, max_chars=256):
     try:
         string = string.decode('utf-8')
-        n = 256
```
bkgoksel, line 103: Nice cleanup!

test-cli.py
```diff
@@ -97,20 +97,20 @@ def get_uuid(line):
...
+            string = string[:max_chars] + ' (...more...)'
         return string
     except UnicodeDecodeError:
         return '<binary>\n'
 
 
-def run_command(args, expected_exit_code=0):
-    # Decode args into ASCII as travis can't handle them otherwise
-    args = [arg.encode('ascii', 'ignore') for arg in args]
+def run_command(args, expected_exit_code=0, max_output_chars=256):
```
bkgoksel, line 112: What happened here? If the commented out line is no longer relevant (and the comment statement is no longer correct), please remove them.

test-worker.sh
```diff
@@ -0,0 +1,8 @@
+# Test working for docker < max & dependency < max
+./venv/bin/python worker/codalabworker/main.py \
+ --server https://worksheets.codalab.org \
```
bkgoksel: For the sake of travis testing, we should run this worker against a local server, or maybe make this bash script take the server as an argument so we can run this worker against local dev in travis.

bkgoksel (changes requested):
The actual unicode handling parts and the logic all around looks good.
Couple nits:

- The `copy` test module is failing.
- A bunch of commits from a different PR (docker python sdk PR) look to be mixed in with these, please rebase with up-to-date master before merging
- Formatting tests fail for the following files, please format before merging:
```
codalab/lib/metadata_util.py
codalab/lib/unicode_util.py
codalab/lib/bundle_cli.py
codalab/rest/schemas.py
```

codalab/objects/metadata_spec.py
```diff
@@ -22,7 +22,7 @@ def bool_constructor(x=False):
 def unicode_constructor(s=""):
     # Need this method so that any uninterpretable characters in the database will be
     # replaced with '?'.
-    return unicode(s, errors='replace')
+    return unicode(s, encoding'ascii', errors='replace')
```
percyliang: missing =

setup.py
bkgoksel: This comment is outdated now as the latest version is 0.2.41 but I don't think we need to change it in this PR anyways.

tests/files/你好世界😊.txt
bkgoksel: I'll try this on my windows machine and report if it breaks it. I believe it's spaces in filenames that really break windows, chinese and emoji should be fine.

nikhilxb:
@bkgoksel @percyliang All tests are passing :) Will pull into master?

codalab/objects/metadata_spec.py
```diff
@@ -19,6 +19,15 @@ def bool_constructor(x=False):
     return x == 'True'
 
 
+def unicode_constructor(s=""):
```
percyliang, line 22: Makes me kind of sad that we have to do this.  Can we put a big TODO that gives context - like we basically blow away unicode, but we need to support unicode in the future.

percyliang (approved):

codalab/objects/metadata_spec.py
bkgoksel: Looks fixed in latest commit

bkgoksel (approved):

codalab/objects/metadata_spec.py
bkgoksel: Yeah, on top of that I say we leave #502 open since this is more of a bandaid fix. We can definitely close #829 and #961 for now after this is merged.

nikhilxb: When we migrate to Python 3, good unicode handling will come for free. In addition, I suggest we adopt types, as it was really annoying how half of our strings are `str` and the other half is `unicode` but you have no idea when (even for the same function sometimes we get different typed inputs).

PR #1004, opened 2018-12-04
Upgrade marshmallow to a more recent version

bkgoksel:
To fix https://github.com/codalab/codalab-worksheets/issues/450 and also to let us use marshmallow elsewhere


bkgoksel:
Upgrading marshmallow-jsonapi to 0.15.1 works but upgrading it to 0.20.5 fails
Upgrading marshmallow to 2.16.3 from 2.6.1 also fails.

In the end only upgrading marshmallow-jsonapi to 0.15.1 which should fix the security vulnerability.

PR #1011, opened 2019-01-07
Docker image removal revamp

bkgoksel:
Force deletes docker images, uses a register/release dependent
mechanism to avoid deleting images of running containers.

Needed to handle the following edge case:

1. Someone requests a docker image (tensorflow:gpu-latest)
2. Someone requests the same docker image under a different tag (tensorflow:1.12.0-gpu)
3. The time comes to clear this image from the cache

The cache stores the image using the digest (tensorflow@sha256...) but we cannot delete images by digest (due to https://github.com/moby/moby/issues/24688). As a result, we delete images by id (can't get repo/tag since repo/tag -> id is a many-to-one relationship).

However, if docker image pull was called more than once for a given image under different tags (ie tensorflow:gpu-latest and tensorflow:1.12.0-gpu), then docker does not allow removing this image by id because it is referred to by more than one repository/tag, UNLESS the --force option is specified.

However, adding the --force option means we risk deleting images of running containers, so now the docker image manager needs to keep track of which images are currently being used by some run. 

To do so, this PR adds a mechanism by which the docker image manager also saves which runs are using a given image and avoids calling remove on those images until the run manager explicitly calls release when the container is stopped.

### DIFF
--- a/worker/codalabworker/docker_utils.py
+++ b/worker/codalabworker/docker_utils.py
@@ -16,10 +16,11 @@ from formatting import parse_size
 MIN_API_VERSION = '1.17'
 NVIDIA_RUNTIME = 'nvidia'
 DEFAULT_RUNTIME = 'runc'
+DEFAULT_TIMEOUT = 720
 
 
 logger = logging.getLogger(__name__)
-client = docker.from_env()
+client = docker.from_env(timeout=DEFAULT_TIMEOUT)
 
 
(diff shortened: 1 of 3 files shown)


bkgoksel:
Use the `tags` attribute of the image object to get a list of tags for the image and try to delete each tag, if any fails just skip deleting it like we did beforehands

PR #1021, opened 2019-02-04
Slurm Run Manager

bkgoksel:


percyliang:
@bkgoksel should we close this?

bkgoksel:
Closed, but let's keep the branch around in case in the future we need a blueprint for this kind of Run Manager.

PR #1029, opened 2019-02-19
Worksheets migration

w0nche0l:
Feature parity migration of the Django+React hybrid frontend located in https://github.com/codalab/codalab-worksheets to pure React+CreateReactApp frontend.


frontend/public/index.html
```diff
@@ -0,0 +1,65 @@
...
+  <head>
+    <meta charset="utf-8">
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <meta name="theme-color" content="#000000">
+    <!--
+      manifest.json provides metadata used when your web app is added to the
+      homescreen on Android. See https://developers.google.com/web/fundamentals/web-app-manifest/
+    -->
+    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
```
percyliang, line 17: Is PUBLIC_URL necessary?  Why can't it be empty?

frontend/README.md
```diff
@@ -0,0 +1 @@
+## CodaLab Worksheets
```
percyliang, line 1: Add documentation

frontend/src/CodalabApp.js
```diff
@@ -0,0 +1,160 @@
...
+} from './components/PasswordReset';
+
+// Routes
+import BundleRoute from './routes/BundleRoute';
+
+import history from './history';
+import Cookies from 'universal-cookie';
+
+////////////////////////////////////////////////////////////
+// 1. Click the public page
```
percyliang: What's this?

frontend/src/components/_Example/Example.js
```diff
@@ -0,0 +1,50 @@
+// @flow
```
percyliang: All this stuff is not used and can be deleted, right?

frontend/src/components/worksheets/WorksheetChatBox.js
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
```
percyliang: You can just delete this since we never use it.

frontend/src/routes/BundleRoute.js
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
```
percyliang, line 1: Why do we have one of these only for bundles?

percyliang (commented):
Really nice to see it all come together!

- Can you put a TODO next to the things that we should do in a second pass (i.e., removing jquery and making things more pure React).
- Shall we use tabWidth 2 instead of 4?

frontend/public/index.html
w0nche0l: public url is what lets it link to files that are being statically hosted in the public folder, rather than the built javascript files in the src folder

frontend/src/CodalabApp.js
w0nche0l: random stuff left over from tutorials :)

frontend/src/components/_Example/Example.js
w0nche0l: yes

PR #1097, opened 2019-03-15
One-command deployment via docker-compose

bkgoksel:
Adds a convenience script that brings up a whole CodaLab server installation (ie database, backend, frontend, nginx but no worker) using `latest` Docker images and docker-compose for orchestration. Optionally builds the Docker images locally with the `local-dev` tag and uses those to bring up the installation.

Tasks done:
- [x] Bring up an entire cluster in docker-compose
- [x] Integrate various options (first configuration of db etc)
- [x] Add customizability via environment variables
- [x] Make the Travis changes to have tests just build images and run tests on them

Tasks left:
- [x] Make tests pass:
  - [x] Upgrade docker-compose on Travis
  - [x] Move the `temp_instance` method in tests to use docker to bring up a Codalab Instance from within python
  - [x] Fix issue where copy tests fail: #1112
  - [x] **Fix netcat test failure**
    *Issue here is that our worker really assumes to be on `host` docker network mode with the ability to connect to any of its spawned docker containers by their IPs.*:
    - [x] **Use another docker network that includes the worker container and any run container to enable communications between the worker and runs without resorting to host mode.**
  - [x] Fix `docker` test failure
  - [x] Fix `competitions` test failure
  - [x] Fix or disable `events` command tests
- [ ] Make secret key customizable via environment variable as well
- [x] Better document (help for various customizable env vars)
- [x] Simplify the service starting and image building processes to avoid long repeated builds

May need to skip:
- [ ] Add tests to the `docker-compose` orchestration to streamline everything:
*Tried this but looks like this leads to docker hanging  without any output with no clear path to solution for now*


docker/service/start_service.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+# start_service.sh
+# Start a full Codalab Worksheets Server service
```
percyliang: "Server service" is redundant - just service?

docker/service/start_service.sh
```diff
@@ -0,0 +1,71 @@
...
+                        ;;
+    -h | --help )       usage
+                        exit
+  esac
+done
+
+if [ "$BUILD" = "1" ]; then
+  CODALAB_VERSION=local-dev
+  echo "==> Building 'local-dev' Docker images"
+  ./docker/build_images.sh local-dev
```
percyliang: For all of these commands, do we want to put a `|| exit 1` after each one to not continue if something fails?

docker/service/start_service.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+# start_service.sh
+# Start a full Codalab Worksheets Server service
+
+set -a
+
+usage()
+{
+  echo "Starts a full Codalab Worksheets Server service. Optionally builds docker images for it. If not building local images, 'latest' tags used, otherwise images are built with the 'local-dev' tag and these images are used. [[-b --build: Build docker images first] [-h --help: get usage help]]"
```
percyliang: Can we wrap this?
Same comment about "Server service".
Improve the description of this usage.

docker/service/files/nginx.conf
```diff
@@ -0,0 +1,149 @@
...
+	# gzip_proxied any;
+	# gzip_comp_level 6;
+	# gzip_buffers 16 8k;
+	# gzip_http_version 1.1;
+	# gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
+
+	##
+	# Virtual Host Configs
+	##
+
```
percyliang, line 62: weird indentation?

docker/service/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+      - rest-server
+
+  frontend:
+    image: codalab/frontend:${CODALAB_VERSION}
+    ports:
+      - 2700:2700
+    depends_on:
+      - rest-server
+
+  nginx:
```
percyliang, line 91: Should we have an option to bring up a worker too while we're at it?

docker/service/start_service.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+# start_service.sh
```
percyliang: Unfortunately we aren't consistent with using `_` or `-`...I'd prefer `-`, but consistency is more important...

percyliang (approved):
This is great!

PR #1136, opened 2019-04-16
Frontend dev mode for docker service

bkgoksel:
With this PR, appending the `-d` option to the `start-service.sh` script brings up a dev mode frontend docker container that just mounts the local directory


Matt-F-Wu:
Not sure if I am doing this incorrectly, but running `./docker/service/start-service.sh -d` doesn't seem to start anything at localhost. Also, when I did 'npm start run' within `/frontend`, the frontend seems broken (can't talk to the backend, login not working).
Am I doing this wrong? Thank you!

PR #1145, opened 2019-05-13
Single click deployment/service wrapper in Python

bkgoksel:
Adds a python script that handles the deployment of an entire Codalab server stack locally. Configuration options allow quick local development as well as highly customizable production deployments (with support for things like SSL etc). Also adapts the CI configuration for tools like Travis.


.travis.yml
```diff
@@ -15,10 +15,12 @@ jobs:
...
       install:
         - pip install --upgrade pip MySQL-python
-        - ./docker/build-images.sh ${TRAVIS_BRANCH} -p
+        - ./codalab-service.py build bundleserver -v ${TRAVIS_BRANCH} --push
+        - ./codalab-service.py build worker -v ${TRAVIS_BRANCH} --push
+        - ./codalab-service.py build frontend -v ${TRAVIS_BRANCH} --push
         - ./setup.sh server
       script:
-        - ./docker/codalab-service/start-service.sh -i -t -w
+        - ./codalab-service.py start -i -w -t --root-user codalab --root-pwd testpassword --worker-dir /home/travis/codalab/worker-dir --rest-port 2900 --mysql-port 3306 --version ${TRAVIS_BRANCH}
```
percyliang: can we use the long form `--root-password`?

.travis.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       services:
         - docker
       env:
-        - CODALAB_USERNAME=codalab CODALAB_PASSWORD=testpassword CODALAB_ROOT_PASSWORD=testpassword CI=false CODALAB_SERVICE_HOME=/home/travis/codalab/home/ CODALAB_BUNDLE_STORE=/home/travis/codalab/bundles/ CODALAB_MYSQL_MOUNT=/home/travis/codalab/mysql/ CODALAB_WORKER_DIR=/home/travis/codalab/worker-dir/ CODALAB_MYSQL_PORT=33306 CODALAB_VERSION=${TRAVIS_BRANCH}
+        - CODALAB_USERNAME=codalab CODALAB_PASSWORD=testpassword CI=false
```
percyliang: can we have a more consistent scheme for default passwords (mysql, etc.)?  either all `codalab` or something?

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        'docker_pwd': None,
+        'build_locally': False,
+        'image': 'all',
+        'test_build': False,
+        'user_compose_file': None,
+        'start_worker': False,
+        'initial_config': False,
+        'root_dir': None,
+        'instance_name': 'codalab',
+        'mysql_root_pwd': 'mysql_root_pwd',
```
percyliang: see comments above about using `password` and consistency on passwords

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        'mysql_port': None,
+        'use_ssl': False,
+        'ssl_cert_file': None,
+        'ssl_key_file': None,
+        'follow': False,
+        'tail': None,
+        'tests': ['default'],
+    }
+
+    ARG_TO_ENV_VAR = {
```
percyliang: Could we combine this and the block above, so we have one place where all the options are defined?

percyliang: Actually better yet, can we just merge that with the options parsing block below, so that everything is in one place?  Especially having the documentation right would be really helpful.

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    @staticmethod
+    def _get_parser():
+        parser = argparse.ArgumentParser(description="Manages your local CodaLab worksheets back-end deployment using docker-compose")
+        subparsers = parser.add_subparsers(dest='command', description='Command to run')
+
+        # SUBCOMMANDS
+
+        start_cmd = subparsers.add_parser('start', help='Start a CodaLab backend instance')
```
percyliang: frontend too, right?

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        logs_cmd = subparsers.add_parser('logs', help='View logs for existing CodaLab instance')
+        test_cmd = subparsers.add_parser('test', help='Run tests against an existing CodaLab instance')
+        build_cmd = subparsers.add_parser('build', help='Build CodaLab docker images using the local codebase')
+        run_cmd = subparsers.add_parser('run', help='Run a command inside a service container')
+
+        stop_cmd = subparsers.add_parser('stop', help='Stop any existing CodaLab backend instance')
```
percyliang: any instance => any instances (same below)

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        down_cmd = subparsers.add_parser('down', help='Bring down any existing CodaLab backend instance')
+        restart_cmd = subparsers.add_parser('restart', help='Restart any existing CodaLab backend instance')
+
+        #  BUILD SETTINGS
+
+        for cmd in [build_cmd, start_cmd]:
+            cmd.add_argument('--version', '-v', type=str, help='CodaLab version to use for building and deployment', default=argparse.SUPPRESS)
```
percyliang: why do you want to suppress these?

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+            cmd.add_argument('--docker-user', type=str, help='DockerHub username to push images from', default=argparse.SUPPRESS)
+            cmd.add_argument('--docker-pwd', type=str, help='DockerHub password to push images from', default=argparse.SUPPRESS)
+
+        build_cmd.add_argument('image', default=argparse.SUPPRESS, help='Images to build', choices=['bundleserver', 'frontend', 'worker', 'default-cpu', 'default-gpu', 'all'], nargs='?')
+
+        #  DEPLOYMENT SETTINGS
+        for cmd in [start_cmd, stop_cmd, restart_cmd, down_cmd, logs_cmd]:
+            cmd.add_argument('--instance-name', type=str, help='Docker compose name to use for instance', default=argparse.SUPPRESS)
+
+        start_cmd.add_argument('--build-locally', '-b', action='store_true', help='If specified build VERSION using local code.', default=argparse.SUPPRESS)
```
percyliang: Add comma: 'If specified, build...' (same for lines below)

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        start_cmd.add_argument('--test-build', '-t', action='store_true', help='If specified run tests on the build.', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--user-compose-file', type=str, help='If specified path to a user-defined Docker compose file that overwrites the defaults', default=argparse.SUPPRESS)
```
percyliang: why not `--docker-compose-file`?

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        start_cmd.add_argument('--start-worker', '-w', action='store_true', help='If specified start a CodaLab worker on this machine.', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--initial-config', '-i', action='store_true', help='If specified, save the initial configuration of the instance (defaults to true if the service home or the database mounts are ephemeral)', default=argparse.SUPPRESS)
+
+        #  USER CREDENTIALS
+
+        start_cmd.add_argument('--mysql-root-pwd', type=str, help='Root password for the database', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--mysql-user', type=str, help='MYSQL username for the Codalab MYSQL client', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--mysql-pwd', type=str, help='MYSQL password for the Codalab MYSQL client', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--root-user', type=str, help='Codalab username for the Codalab admin user', default=argparse.SUPPRESS)
```
percyliang: maybe this should be `--codalab-user` (root is vague in the context of these other options)

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        start_cmd.add_argument('--root-pwd', type=str, help='Codalab password for the Codalab admin user', default=argparse.SUPPRESS)
+
+        #  HOST FILESYSTEM MOUNTS
+
+        start_cmd.add_argument('--uid', type=str, help='Linux UID that owns the files created by Codalab. default=(ID of the user running this script)', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--service-home', type=str, help='Path on the host machine to store home directory of the Codalab server (by default nothing is stored', default=argparse.SUPPRESS)
```
percyliang: `--codalab-home`?  that's what we've called it in the past

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        start_cmd.add_argument('--mysql-mount', type=str, help='Path on the host machine to store mysql data files, by default the database is ephemeral', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--worker-dir', type=str, help='Path on the host machine to store worker data files, by default these are ephemeral', default=argparse.SUPPRESS)
+        start_cmd.add_argument('--bundle-store', type=str, help='Path on the host machine to store bundle data files, by default these are ephemeral', default=[], dest='bundle_stores', action='append')
+
+        start_cmd.add_argument('--worker-docker-network-name', type=str, help='Name of the docker network that includes the worker and runs', default=argparse.SUPPRESS)
+
+        #  HOST PORT MOUNTS
+
+        start_cmd.add_argument('--http-port', type=str, help='HTTP port for the server to listen on', default=argparse.SUPPRESS)
```
percyliang: if ssl is setup, then this is https? (document)

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+            if var in env:
+                setattr(self, arg, env[var])
+        self._apply_defaults()
+
+
+class CodalabServiceManager(object):
+
+    @staticmethod
+    def resolve_env_vars(args):
+        environment = {
```
percyliang: Document it that environment variables are needed for docker-compose?

codalab-service.py
```diff
@@ -0,0 +1,387 @@
...
+        self.bring_up_service('frontend')
+        print("[CODALAB] ==> Starting nginx")
+        self.bring_up_service('nginx')
+        if self.args.start_worker:
+            print("[CODALAB] ==> Starting worker")
+            self.bring_up_service('worker')
+
+    def build(self):
+        print("[CODALAB] => Building Docker images")
+        IMAGE_TO_SUFFIX = {
```
percyliang: can this just be the identity map?

docker/dockerfiles/Dockerfile.server
```diff
@@ -18,6 +18,7 @@ RUN apt-get update; apt-get install -y \
     zip;
 
 # Install dependencies
+COPY docker/compose_files/files/wait-for-it.sh /data/bin/wait-for-it.sh
```
percyliang: why `/data`?  seems like a semantically odd choice

percyliang (changes requested):
It will be really nice to have this single script!  Main comments are nitpicking naming and wondering if we can consolidate the massive set of options a bit more.  Also would be great to have a block of documentation at the top giving the high-level picture of what's going on here.

.travis.yml
bkgoksel: Good idea, done

.travis.yml
bkgoksel: Set them all to codalab for now.

codalab-service.py
bkgoksel: See my comment below about env vars. If we don't allow env var-based configuration this could be a simple argparse thing.

codalab-service.py
bkgoksel: Yep will fix

codalab-service.py
bkgoksel: done

codalab-service.py
bkgoksel: Done

codalab-service.py
bkgoksel: There are already quite a few docker compose files that configure the service, this argument is for situations where you have advanced customizations that go beyond what this script can do. If you specify a user docker compose file, any field in it overrides what's in the service setup.

In that sense it's a docker compose file that allows the user to override the provided docker-compose files. Calling it just `docker-compose-file` didn't seem to convey that but maybe we can change it to `docker-compose-override-file` or something?

codalab-service.py
bkgoksel: This is a trick to allow setting variables from 3 places with sensible precedence (CLI > env > defaults). See my comment below on whether we need this or not.

codalab-service.py
bkgoksel: Changed

codalab-service.py
bkgoksel: As things stand, HTTPS always listens on port 443 and this actually only changes the HTTP port regardless of whether you're using SSL or not. 

After your comment, I think it makes more sense to have it modify HTTPS if SSL enabled, HTTP otherwise.

codalab-service.py
bkgoksel: Yep, I guess there's no harm to renaming our Dockerfiles, which I was trying to avoid.

docker/dockerfiles/Dockerfile.server
bkgoksel: This is where it was placed in the old deploy script I started from months ago but I agree, I'll probably put everything in `/opt`

codalab-service.py
percyliang: I think as long as we can read it from a file, I don't think we need to support var-based config.  One other thing I thought of: one thing that would be helpful is to have it print out the environment variables that it's setting so if people want to do manual override, it's easy for them.  Similarly, printing out the docker commands that are run (maybe in a dry-run way) would be useful, so we have some flexibility.

codalab-service.py
bkgoksel: The users don't need to set env vars anymore, and we don't need to parse env vars actually (in this setup).

Here's some context: 
- The only way to modify  the behavior of docker-compose files is to use environment variables.
- In the earlier version of the script (written in bash), instead of trying to parse many command line arguments and then issue the docker-compose commands with env variables initialized from those, I just passed on the user's current env vars to the command. As a result, the user had to set every important env variable.
- In the python script, the env vars that are applied to the compose commands and the env vars of the user's computer are completely separate. Thanks to the subprocess library, I construct the set of necessary variables for each compose command.
- As a result, it doesn't really matter what env variables are set on the host machine, we can reconstruct all the necessary env variables from the command line arguments of the script before calling the compose command.
- For backwards compatibility, I initially tried to write the python script such that its arguments can be set from environment variables AND command line arguments.
- The precedence there is CLI argument > env var > default.
- What that means is that all arguments have some default which is used if that argument is not set on the command line or the environment. If there's an env var but no CLI argument, the env var is used, if there's an env var and also a command line argument, the command line argument overwrites the value from the env var.
- To do this 3-way merging of arguments (between defaults, env vars and CLI args) I had to do all the custom stuff with mappings (from env var name to arg name) and suppressing the default `default` behavior of argparse (which doesn't play nicely with using a custom Namespace that can have values set at other points).

Our options:
1. Get rid of env var parsing for arguments: We can tell users to just pass in command line arguments for any configuration and get rid of all that messy logic and mappings. I've already rewrote the travis script to use CLI args for testing, we can just move everything to use the CLI args and tell people that's how you configure the codalab instance.

Pros:
    - Get rid of messy argument parsing logic
Cons:
    - If you have some local defaults you want to persist,  there's no way to do that and you need to use a long command with many arguments every time you do something

2. Instead of env vars, use some config file parsing (like JSON).

Pros:
    - Reduce messy merging logic (no more need for mapping env var names to arg names etc)
    - Allow people to save local defaults for easily redeploying etc
Cons:
    - Most of the messy logic still there (since we still need to 3-way merge CLI, JSON and default values)

3. Keep it like this with env vars being the way to persist your defaults

Pros:
    - Local defaults possible
    - No rewriting code portions
Cons:
    - No reduction in mess
    - Env vars are more opaque and may change since last time service started etc. So not great for rerunning etc.


Looking at this, option (2) seems to be the best option, but we can also go with (1) and tell people to save the long commands they use to start the service. 

What do you think @percyliang ?

codalab-service.py
bkgoksel: Done

codalab-service.py
bkgoksel: done (changed to `codalab`)

codalab-service.py
percyliang: Even if we decide on doing the 3-way merge, I think there should be a way to writing the code so that things are more declarative and defined in one place, so the code looks more like enumerating over arguments/fields/vars rather than having it be flattened.

I think we should definitely support loading from file/environment, not just from command-line args.  On a shared system, the password would show up when any user does a `ps`, which is bad.  I think loading from a config file which is isomorphic to the arg parse arguments would be ideal (you write a simple converter than converts dicts to args).

codalab-service.py
bkgoksel: The problem is, the end commands end up having two parts to them (a set of env vars and the actual command), so even if we printed the commands being executed, just running those wouldn't be enough. We could print the env vars AND the commands (in the copyable `VAR=val VAR2=val2 cmd --args` format) but these would be huge strings given the complexity of env vars and commands being run.
Maybe we can add a verbosity flag to enable printing these?

codalab-service.py
bkgoksel: One option here is to use a 3rd party CLI building framework like [click](https://click.palletsprojects.com/en/7.x/) 

The advantage is things like env var based configuration are handled natively and it has builtin handlers for things like passwords, confirmations etc.

The main downside is it is just another library to learn, and in particular click has a decorator-based approach to generating the CLI rather than the more explicit argparse, which would look unfamiliar to any developer used to the argparse style.

PR #1149, opened 2019-05-20
[HOTFIX] Fix build issue: mysql needs to be root

bkgoksel:


bkgoksel:
Irrelevant after #1145

PR #1151, opened 2019-05-24
Get rid of time quota and implement parallel running jobs quota

bkgoksel:
This PR removes the concept of `time_quota` from CodaLab (while keeping the `time_used` field for statistics purposes) and instead implements a `parallel_job_quota` which limits the number of jobs a user may have running on public CodaLab workers (while still allowing them to run runs on their own workers)

This field is also editable via the CLI on a per user basis.

Runs that are staged once the user has the maximum number of parallel running jobs are kept at staged state.


codalab/lib/bundle_cli.py
```diff
@@ -3467,7 +3467,9 @@ def do_chown_command(self, args):
             Commands.Argument('--affiliation', help='Affiliation'),
             Commands.Argument('--url', help='Website URL'),
             Commands.Argument(
-                '-t', '--time-quota', help='Total amount of time allowed (e.g., 3, 3m, 3h, 3d)'
+                '-p',
+                '--parallel-job-quota',
+                help='Total amount of jobs the user may running have at a time',
```
percyliang: add "on public workers"

codalab/lib/bundle_cli.py
```diff
@@ -3467,7 +3467,9 @@ def do_chown_command(self, args):
             Commands.Argument('--affiliation', help='Affiliation'),
             Commands.Argument('--url', help='Website URL'),
             Commands.Argument(
-                '-t', '--time-quota', help='Total amount of time allowed (e.g., 3, 3m, 3h, 3d)'
+                '-p',
```
percyliang, line 3470: add `type=int`

codalab/lib/codalab_manager.py
```diff
@@ -318,11 +318,11 @@ def session(self):
     @cached
     def default_user_info(self):
         info = self.config['server'].get(
-            'default_user_info', {'time_quota': '1y', 'disk_quota': '1t'}
+            'default_user_info', {'disk_quota': '1t', 'parallel_job_quota': '10'}
```
percyliang: make `10` an integer

codalab/model/bundle_model.py
```diff
@@ -2322,11 +2322,19 @@ def increment_user_time_used(self, user_id, amount):
...
         user_info = self.get_user_info(user_id)
-        time_quota = user_info['time_quota']
-        time_used = user_info['time_used']
-        return time_quota - time_used
+        parallel_job_quota = user_info['parallel_job_quota']
+        with self.engine.begin() as connection:
+            run_rows = connection.execute(
+                select([cl_worker_run.c.run_uuid]).where(
+                    cl_worker_run.c.run_uuid.in_(
+                        select([cl_bundle.c.uuid]).where(cl_bundle.c.owner_id == user_id)
```
percyliang: exclude private workers

codalab/model/bundle_model.py
```diff
@@ -2322,11 +2322,19 @@ def increment_user_time_used(self, user_id, amount):
...
+                    )
+                )
+            ).fetchall()
+        jobs_active = len(run_rows)
+        return parallel_job_quota > jobs_active
```
percyliang: Should return the difference, not a boolean.

codalab/model/tables.py
```diff
@@ -242,7 +242,7 @@
     Column('affiliation', String(255, convert_unicode=True), nullable=True),
     Column('url', String(255, convert_unicode=True), nullable=True),
     # Quotas
-    Column('time_quota', Float, nullable=False),  # Number of seconds allowed
+    Column('parallel_job_quota', Integer, nullable=False),  # Number of parallel jobs allowed
```
percyliang: Wondering if we should name this `parallel_run_quota` to avoid introducing a new term `job`

codalab/worker/bundle_manager.py
```diff
@@ -311,6 +311,12 @@ def _schedule_run_bundles_on_workers(self, workers, user_owned):
             if user_owned:
                 workers_list = workers.user_owned_workers(bundle.owner_id)
             else:
+                if not self._model.get_user_parallel_job_quota_left(bundle.owner_id):
+                    logger.info(
+                        "User %s has no parallel job quota left, skipping job for now",
```
percyliang: It would be good if the function returned the actual quota and number of running jobs so we can report both numbers and the difference to give more information.

bkgoksel:
I'm planning to add tests as well

codalab/lib/codalab_manager.py
percyliang: Also, 10 sounds like a lot given that we're likely to not have more than 10 runs at once.  Maybe 3?

codalab/worker/bundle_manager.py
```diff
@@ -311,6 +311,12 @@ def _schedule_run_bundles_on_workers(self, workers, user_owned):
             if user_owned:
```
percyliang, line 311: Does this loop go through the runs in the order in which they are created?  I assume yes, but probably should be make explicit.

percyliang (changes requested):
Great!  Main comment is about naming and excluding private workers.

I was thinking about the logic a bit more, and it doesn't resolve the following resource hogging scenario.  Suppose there are 4 workers, and each person's quota is 2.  Then if the queue looks like this:
A A A A A A A A A A A A B B B B B B B B C
Then A and B's jobs will all be run first before C gets a chance.
Maybe this is still better than nothing?

codalab/worker/bundle_manager.py
bkgoksel: Looking into the `batch_get_bundles` function, the bundles are sorted by `id`, which I'm guessing is the same as creation order.

I think ideally there would be a separate scheduler component that would take in this list and sort  it with a more involved logic, but I'm not sure this PR is the right place to do this, I think just to keep the PR simple, I shouldn't change this part of the code.

codalab/worker/bundle_manager.py
bkgoksel: The only case where this log  would print is if their number of runs is equal to the quota, so I don't see the value of printing both numbers  here, unless we want to print it for every run being started.

bkgoksel:
> 
> 
> Great! Main comment is about naming and excluding private workers.
> 
> I was thinking about the logic a bit more, and it doesn't resolve the following resource hogging scenario. Suppose there are 4 workers, and each person's quota is 2. Then if the queue looks like this:
> A A A A A A A A A A A A B B B B B B B B C
> Then A and B's jobs will all be run first before C gets a chance.
> Maybe this is still better than nothing?

Yeah this is a very valid concern, and what we need ultimately is a weighted scheduler that would prevent the scenario you outlined and much more. A point based scheduler would also mean that if all workers are free, we wouldn't be artificially limiting the number of runs for a user (in this setup, even if there are free workers and no other staged runs, you can't run more than your quota).

However this took me a day to get done, whereas a full on scheduler would be a larger undertaking, so I think this is a good first step.

codalab/worker/bundle_manager.py
percyliang: Yeah, let's leave the code as is, but put a comment documenting the limitations and the need for a real scheduler.

codalab/worker/bundle_manager.py
percyliang: Can you print out the one number then...User %s already has {} runs running...'
s/job/run

codalab/model/bundle_model.py
```diff
@@ -2322,11 +2322,24 @@ def increment_user_time_used(self, user_id, amount):
...
+            # Root user has no parallel run quota
+            return parallel_run_quota
+        with self.engine.begin() as connection:
+            active_runs = connection.execute(
+                select([cl_worker_run.c.run_uuid]).where(
+                    and_(
+                        cl_worker_run.c.run_uuid.in_(
+                            select([cl_bundle.c.uuid]).where(cl_bundle.c.owner_id == user_id)
+                        ),
+                        cl_worker_run.c.user_id != user_id,
```
percyliang, line 2340: comment on the logic

percyliang (approved):

PR #1155, opened 2019-06-06
Worker robustness improvements

### LINKED ISSUE #1006: Outdated method docstring in docker image manager
https://github.com/codalab/codalab-cli/blob/e76863278421a6b91af86f7ac968e36f727ba4b7/worker/codalabworker/local_run/docker_image_manager.py#L123

### LINKED ISSUE #1154: Dependency symlink [File exists] issue in worker
Many workers went offline on the same day with the following error message:
```
2019-06-03 19:53:31,621 http://localhost:None "GET /v1.35/images/theerit/quac_flowqa_submission:latest/json HTTP/1.1" 200 None
  File "/opt/worker/codalabworker/worker.py", line 51, in start
  File "/opt/worker/codalabworker/local_run/local_run_manager.py", line 171, in process_runs
  File "/opt/worker/codalabworker/fsm.py", line 35, in transition
  File "/opt/worker/codalabworker/local_run/local_run_state.py", line 221, in _transition_from_PREPARING
    os.symlink(docker_dependency_path, child_path)
OSError: [Errno 17] File exists

```

Unclear as of yet why this could be happening.

bkgoksel:
1. Make sure dependency child paths are nonempty 08778cb : should fix #1154 
2. Fix docker image manager docstring method 6b77c79 : should fix #1006 
3. More robust exception handling in worker (catch any unhandled exception for each run and only fail the run instead of crashing) 2959473
4. Connect to the correct local instance when launching tests from the service script ad43130 & 11a53da

### DIFF
--- a/worker/codalabworker/local_run/local_run_state.py
+++ b/worker/codalabworker/local_run/local_run_state.py
@@ -106,13 +106,31 @@ class LocalRunStateMachine(StateTransitioner):
         assign_cpu_and_gpu_sets_fn,
     ):
         super(LocalRunStateMachine, self).__init__()
-        self.add_transition(LocalRunStage.PREPARING, self._transition_from_PREPARING)
-        self.add_transition(LocalRunStage.RUNNING, self._transition_from_RUNNING)
-        self.add_transition(LocalRunStage.CLEANING_UP, self._transition_from_CLEANING_UP)
         self.add_transition(
-            LocalRunStage.UPLOADING_RESULTS, self._transition_from_UPLOADING_RESULTS
+            LocalRunStage.PREPARING,
+            self._transition_from_PREPARING,
+            exception_handler_function=self._handle_exception,
         )
-        self.add_transition(LocalRunStage.FINALIZING, self._transition_from_FINALIZING)
+        self.add_transition(
+            LocalRunStage.RUNNING,
+            self._transition_from_RUNNING,
+            exception_handler_function=self._handle_exception,
+        )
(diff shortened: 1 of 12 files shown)


worker/codalabworker/local_run/local_run_state.py
```diff
@@ -106,13 +106,31 @@ def __init__(
...
+        )
+        self.add_transition(
+            LocalRunStage.UPLOADING_RESULTS,
+            self._transition_from_UPLOADING_RESULTS,
+            exception_handler_function=self._handle_exception,
+        )
+        self.add_transition(
+            LocalRunStage.FINALIZING,
+            self._transition_from_FINALIZING,
+            exception_handler_function=self._handle_exception,
```
percyliang, line 132: If every stage has the same handler, why is it an argument?

percyliang (approved):

worker/codalabworker/local_run/local_run_state.py
bkgoksel: Yeah I was undecided between making it one handler for all exceptions versus specifying individual, and initially I thought I might need different behavior for each (which I kinda do now as well, if you look at the if statement inside the handler function). 

I should probably make it one default for all states, just to keep things simple.

bkgoksel:
Enforcing dependency child paths to be nonempty breaks make bundles, which allow this. I need to update the logic there...

bkgoksel:
Closing as split into PRs #1156 #1157 #1159

PR #1014, opened 2019-01-29
Migrate everything from Python 2.7 to Python 3.6

bkgoksel:
As pip has recently told us in all caps:

`DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7`

This is the beginning of a long, uncertain journey that we nevertheless must take.
The first step is to run `2to3` on our entire codebase and hope that the world doesn't burn around us.

Godspeed,

PR #1158, opened 2019-06-06
Upgrade to python 3

epicfaace:
I just ran 2to3 on it, so probably need more testing before actually merging this.


bkgoksel:
#1014 also exists, where I enumerated some further changes that we'd need.
Also I think 3.7 is a better target than 3.6 since it's a bit further without backwards incompatibilities (AFAIK)

epicfaace:
Oh, whoops

PR #1159, opened 2019-06-06
Better dealing with nonexistent containers and container removal

bkgoksel:
Better dealing with container removal and non-existing containers in the worker state machine.


worker/codalabworker/local_run/local_run_state.py
```diff
@@ -392,12 +412,18 @@ def _transition_from_CLEANING_UP(self, run_state):
         """
         bundle_uuid = run_state.bundle['uuid']
         if run_state.container_id is not None:
-            while True:
+            while docker_utils.container_exists(run_state.container):
                 try:
                     finished, _, _ = docker_utils.check_finished(run_state.container)
                     if finished:
                         run_state.container.remove(force=True)
+                        run_state = run_state._replace(container=None, container_id=None)
```
epicfaace: @bkgoksel why was this line of code needed?

bkgoksel: It's been a while since I opened this PR, but basically we use None-checks on these fields to decide whether this run still has a container attached to it, so the right time to set these back to None is immediately after we're sure we managed to remove the container without running into issues, which is here.

bkgoksel:
Revisited this. Should be uncontroversial and mergeable in this shape. @percyliang @epicfaace

codalab/worker/local_run/local_run_state.py
```diff
@@ -407,12 +408,18 @@ def _transition_from_CLEANING_UP(self, run_state):
...
                     finished, _, _ = docker_utils.check_finished(run_state.container)
                     if finished:
                         run_state.container.remove(force=True)
+                        run_state = run_state._replace(container=None, container_id=None)
                         break
+                    else:
+                        try:
+                            run_state.container.kill()
+                        except docker.errors.APIError:
+                            logger.error(traceback.format_exc())
```
percyliang, line 420: Why does this not have a sleep like the following `except`?

codalab/worker/local_run/local_run_state.py
```diff
@@ -370,14 +370,15 @@ def check_disk_utilization():
...
-                    # Otherwise all well
-                    logger.error(traceback.format_exc())
+            if docker_utils.container_exists(run_state.container):
+                try:
+                    run_state.container.kill()
+                except docker.errors.APIError:
+                    finished, _, _ = docker_utils.check_finished(run_state.container)
+                    if not finished:
+                        # If we can't kill a Running container, something is wrong
+                        # Otherwise all well
```
percyliang: Delete this line comment

percyliang (approved):

codalab/worker/local_run/local_run_state.py
bkgoksel: Good point, added.

codalab/worker/local_run/local_run_state.py
bkgoksel: Done

PR #1184, opened 2019-06-07
Fix frontend dockerfile location

epicfaace:
The dockerfile was pointing to the wrong path, so when running frontend in dev mode, it wasn't working.


bkgoksel:
Ha, this was silly. The build is failing because Travis tags the docker images with the branch name which in this case includes a slash character that's invalid. We can try to strip these characters in the Travis config file probably.

PR #1185, opened 2019-06-07
Show --diff on black --check

epicfaace:
This way, when a formatting error happens, you can tell what's wrong. For example, see https://travis-ci.org/codalab/codalab-worksheets/jobs/542855976

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ jobs:
       language: python
       python: 3.6
       install: pip install --upgrade black==18.9b0
-      script: black . --check
+      script: black . --check --diff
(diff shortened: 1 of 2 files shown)


bkgoksel (commented):
Good idea! Let's also fix the formatting error displayed here so the build passes.

bkgoksel:
I just made the formatting change to get builds working again. Thanks for the addition!

bkgoksel:
Hmm make sure you're using the same version of black as we are (18.9b0)
Annoyingly they changed a few rules since then and we'd have to reformat the whole codebase to make it compliant with the never black versions.

PR #1217, opened 2019-06-13
Remove spacing when bundle row is expanded

epicfaace:
So now it looks like this (without any extra empty space between rows):

image

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -264,7 +264,7 @@ class BundleRow extends Component {
                         </TableCell>
                     </TableRow>
                 )}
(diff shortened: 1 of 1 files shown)


percyliang:
I think we want to add more spacing before and after the bundle row that is expanded to make it visually pop out more...

epicfaace:
@percyliang but then when you add more spacing in between rows, it ends up making the row shift to a different position, which is not great for UX and what we discussed on Monday. A better way to make it visually pop out is to add highlight / shadow.

percyliang:
I actually can't tell - what has changed visually in this PR?

percyliang:
Oh, I think I see now - it's the vertical space above the row.

percyliang (approved):

percyliang:
Already fixed in `frontend` now.

PR #1218, opened 2019-06-13
Print out commands that will be run for transparency and enable dry-run

percyliang:


bkgoksel (approved):
Needed addition, thanks!

PR #1219, opened 2019-06-13
Frontend fix new run command

percyliang:


percyliang:
Oops, didn't mean master.

Issue #1225, opened 2019-06-15
cl-worker at 100% utilization even when no jobs running


percyliang:
Happens when I run a worker locally, and after a job completes.

percyliang:
Fixed by #1261

PR #1261, opened 2019-06-28
Streamline services

percyliang:
- Changed `codalab_service.py` to allow for finer-grained, incremental control of what to run
- Got rid of dev docker images (just use the regular ones)
- Fix the 100% CPU worker problem (#1225)
- Fix tests (disabled the temp instance thing for now)


epicfaace:
@percyliang can you get the build passing & merge when you get a chance -- that way I can merge these changes into #1014 (and update that PR accordingly)?

PR #1263, opened 2019-06-30
Delete button for worksheets

### LINKED ISSUE #1169: Add button to delete a worksheet
Should show up near the worksheet details in the upper-right next to the worksheet info.

epicfaace:
Fixes #1169.

image

image

### DIFF
--- a/frontend/src/components/worksheets/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet.js
@@ -78,6 +78,26 @@ var WorksheetContent = (function() {
         });
     };
 
+    WorksheetContent.prototype.deleteWorksheet = function(props) {
+        if (this.info === undefined) return;
+        $('#update_progress').show();
+        $('#save_error').hide();
+        $.ajax({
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/Worksheet.js
```diff
@@ -687,6 +707,29 @@ class Worksheet extends React.Component {
         });
     }
 
+    delete() {
+        if (!window.confirm("Are you sure you want to delete this worksheet?")) {
```
percyliang: Add elaboration: "(Note that this does not delete the bundles, but just detaches them from the worksheet.)"

percyliang (approved):

frontend/src/components/worksheets/Worksheet.js
epicfaace:
```suggestion
        if (!window.confirm("Are you sure you want to delete this worksheet? (Note that this does not delete the bundles, but just detaches them from the worksheet.)")) {
```

PR #1264, opened 2019-06-30
Dashboard redirect

### LINKED ISSUE #1230: when logged in, landing page should be dashboard

epicfaace:
Fixes #1230.
- When the logged-in user goes to the home page at first, it redirects them to http://localhost/worksheets?name=dashboard. This page is represented by the `WorksheetNameSearch` component, which does a REST API call to get the right worksheet ID and then redirects the user to the dashboard.
- Redirects are all done JS-side with a loading animation to make the user experience fast and smooth. I've also changed the log in / sign up / home page links on the home page and navbar to use react-router's `Link` rather than an `<a>` tag -- this also speeds up navigation.
- When the user clicks on the "Codalab" logo on the navbar, it sends them to http://localhost/home which will show the home page regardless of whether the user is logged in or not.
- With these changes, we may no longer need the `/rest/worksheets/?name=dashboard` endpoint, and it was unusual anyway to have the user be directly sent to REST endpoint in the browser -- are you fine with removing it @percyliang ?

### DIFF
--- a/frontend/src/components/worksheets/WorksheetNameSearch.js
+++ b/frontend/src/components/worksheets/WorksheetNameSearch.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import queryString from 'query-string';
+
+export default class extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            loading: true,
+            error: false
+        }
+    }
+    async componentDidMount() {
+        const { name } = queryString.parse(this.props.location.search);
+        try {
+            const response = await fetch(`/rest/worksheets?specs=${name}`).then(e => e.json());
+            const uuid = response.data[0].id;
+            this.props.history.push(`/worksheets/${uuid}/`);
+        }
+        catch (e) {
+            console.error(e);
+            this.setState({ error: true, loading: false });
+        }
+    }
+    render() {
+        const styles = {
+            position: "absolute",
+            top: "50%",
+            left: "50%",
+        }
+        return <div>
+            {this.state.loading && <div style={styles}><img src={`${process.env.PUBLIC_URL}/img/Preloader_Small.gif`} /></div>}
+            {this.state.error && <div>Error.</div>}
+        </div>
+    }
+}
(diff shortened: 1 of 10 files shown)


frontend/src/components/worksheets/WorksheetNameSearch.js
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+    async componentDidMount() {
+        const { name } = queryString.parse(this.props.location.search);
+        try {
+            const response = await fetch(`/rest/worksheets?specs=${name}`).then(e => e.json());
+            const uuid = response.data[0].id;
+            this.props.history.push(`/worksheets/${uuid}/`);
+        }
+        catch (e) {
```
percyliang: style nit: put on previous line

frontend/src/components/worksheets/WorksheetNameSearch.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import queryString from 'query-string';
+
+export default class extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            loading: true,
+            error: false
+        }
```
percyliang: style nit: add semicolon

frontend/src/components/worksheets/WorksheetNameSearch.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+    async componentDidMount() {
+        const { name } = queryString.parse(this.props.location.search);
+        try {
+            const response = await fetch(`/rest/worksheets?specs=${name}`).then(e => e.json());
```
percyliang, line 15: why `specs` instead of `name`, as it was before?

percyliang (approved):

frontend/src/components/worksheets/WorksheetNameSearch.js
epicfaace: See this:
https://github.com/codalab/codalab-worksheets/blob/cfb1d8d5ffef5b03c192c75a4cc9a7796867d693/codalab/rest/worksheets.py#L119
Specifying `specs` will return just one result, and also create a worksheet if it hasn't already been created -- this is the same behavior as that of the `/rest/worksheets/?name=dashboard` endpoint.

frontend/src/components/worksheets/WorksheetNameSearch.js
epicfaace:
```suggestion
        };
```

Issue #1094, opened 2019-03-11
Fix frontend npm build warnings


bkgoksel:
Currently we set `CI=false` in `.travis.yml` to prevent the default npm behavior of treating warnings as errors on prod builds since we don't have time to fix all the warnings before merging into master. Ideally we should NOT have any warnings and turn on the treat warnings as errors behavior to avoid introducing more code smells on the frontend. This issue should track the fixing of those warnings and the removal of the `CI=false` line in the `.travis.yml` file.
The warnings include:
- `===` instead of `==`
- unused vars
- images without alt text (bad for accessibility)
- fake links that should really just be buttons
- unused imports ?

percyliang:
Any errors should show up when you run `pre-commit.sh`.

PR #1265, opened 2019-06-30
Fix JS warnings

epicfaace:
For #1094:
- Fix one of the warnings: Replace == and != with === and !==
- Set CI=true and silence the remaining warnings in `frontend/.eslintrc.json` -- this will make sure that no additional errors come up, and then we can remove these lines from the json file step-by-step as we fix these lint issues.
- Add a pre-commit hook to run `eslint` on the changed files, so that future eslint issues (other than the exceptions above) don't arise. (In a future PR, we can also make it run `prettier` as well in order to maintain consistent format.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ jobs:
       services:
         - docker
       env:
-        - CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
+        - CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=true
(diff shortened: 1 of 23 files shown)


percyliang (approved):
Great.  Why suppress the warnings though if we're going to fix them anyway?

epicfaace:
That way, we can make sure other warnings (other than the suppressed ones) don't pop up. It's better to start enforcing some checks rather than waiting to fix everything, which may take a while.

PR #1276, opened 2019-07-08
better error message for the case where no digest is available

percyliang:


worker/codalabworker/local_run/docker_image_manager.py
```diff
@@ -144,7 +144,12 @@ def get(self, image_spec):
             image_spec += ':latest'
         try:
             image = self._docker.images.get(image_spec)
-            digest = image.attrs.get('RepoDigests', [image_spec])[0]
+            digests = image.attrs.get('RepoDigests', [image_spec])
+            if len(digests) == 0:
+                return ImageAvailabilityState(
+                    digest=None, stage=DependencyStage.FAILED, message='No digest available for {}, probably because it was built locally; delete the docker image on the worker and try again'.format(image_spec),
```
epicfaace:
```suggestion
                    digest=None, stage=DependencyStage.FAILED, message='No digest available for {}, probably because it was built locally; delete the Docker image on the worker and try again'.format(image_spec),
```

epicfaace (approved):
Looks good, do you think we should have tests for docker_image_manager?

percyliang:
Yes, we should have a larger discussion about how to robustify our tests.

PR #1277, opened 2019-07-08
tweak names (delete is more descriptive)

percyliang:


epicfaace:
Looks good, but do we need to worry about backwards compatibility (yet)?

percyliang:
`codalab_service.py` is so new, that this shouldn't be a worry, but definitely good to keep in mind for the future.

PR #1284, opened 2019-07-11
simplify the installation

percyliang:
Removed setup.sh, now that everything should either be run through Docker or a pip package.

Removed virtualenv in the docker files since it is unnecessary.

When building the docker image, add files selectively so it doesn't accidentally add a bunch of junk files.

Remove execute permissions on a bunch of scripts so the Python invocation can be explicit.


.travis.yml
```diff
@@ -15,11 +15,9 @@ jobs:
         - sudo mv docker-compose /usr/local/bin
         - sudo service mysql stop
       install:
-        - pip install --upgrade pip MySQL-python
         - ./codalab_service.py build services -v ${TRAVIS_BRANCH} --push
-        - ./setup.sh server
       script:
-        - ./codalab_service.py start -t --mysql-port 3306 --version ${TRAVIS_BRANCH}
+        - ./codalab_service.py start -s default test --mysql-port 3306 --version ${TRAVIS_BRANCH}
```
epicfaace, line 20: What's the reason for removing the -t command?

percyliang: I made all the tests run in docker for consistency (since `venv` doesn't exist anymore), and it was simpler to make `test` another "service".

Issue #1063, opened 2019-03-11
generated summary inserts bundle uuid references too aggressively


w0nche0l:
_From @fabeschan on June 3, 2017 8:15_

I have a bundle called `train` that has some uuid

Also a run bundle with command that includes the word `train` multiple times, none of them referring to the bundle named the same thing, yet the generated summary inserts the bundle uuid reference immediately after every `train` found.

Here is a screenshot (look at bundles named `run-answer`)

![screenshot 2017-06-03 01 09 35](https://cloud.githubusercontent.com/assets/4369332/26751962/0ebff516-47fa-11e7-905a-eab657c696d9.png)

_Copied from original issue: codalab/codalab-worksheets#333_

percyliang:
We should probably just get rid of these {0x...} annotations since it's confusing.
Let's find another way to make dependencies more clear.

candicegjing:
@percyliang I want to confirm with you that we want to remove those dependencies annotation for now and not change anything else right?

percyliang:
Yes.

PR #1292, opened 2019-07-16
Remove depedency annotation for running bundles #1063

candicegjing:
This PR is to fix issue in #1063 . I removed call to friendly_render_dependency in running bundle.

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -419,15 +419,7 @@ def interpret_genpath(bundle_info, genpath):
                 args.append(friendly_render_dep(dep)[1])
             return '= ' + ' '.join(args)
         elif bundle_type == 'run':
-            command = bundle_info['command']
-            for dep in deps:
-                key, value = friendly_render_dep(dep)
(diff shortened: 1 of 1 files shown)


codalab/lib/worksheet_util.py
```diff
@@ -420,13 +420,6 @@ def friendly_render_dep(dep):
...
         elif bundle_type == 'run':
             command = bundle_info['command']
-            for dep in deps:
-                key, value = friendly_render_dep(dep)
-                # Replace full-word occurrences of key in the command with an indicator of the dependency.
-                # Of course, a string match in the command isn't necessary a semantic reference to the dependency,
-                # and there are some dependencies which are not explicit in the command.
-                # But this can be seen as a best-effort attempt.
-                command = re.sub(r'\b%s\b' % key, value, command)
             return '! ' + command
```
percyliang: Can combine the previous two lines

percyliang (approved):

codalab/lib/worksheet_util.py
candicegjing: Updated.

PR #1294, opened 2019-07-19
initialize alembic

percyliang:
Will make these instructions work:
https://github.com/codalab/codalab-worksheets/wiki/Server-Setup#database-migrations


candicegjing:
Looks good to me.

PR #1295, opened 2019-07-19
Add timeout function for bundles in uploading, staged and running state

candicegjing:


codalab/worker/bundle_manager.py
```diff
@@ -17,7 +17,10 @@
 
 logger = logging.getLogger(__name__)
 
+
 WORKER_TIMEOUT_SECONDS = 60
+SECONDS_PER_DAY = 86400
```
percyliang: Write 60 * 60 * 24 out

codalab/worker/bundle_manager.py
```diff
@@ -17,7 +17,10 @@
...
+BUNDLE_TIMEOUT_SECONDS = SECONDS_PER_DAY * 60
```
percyliang: Write comment about what this is

codalab/worker/bundle_manager.py
```diff
@@ -86,6 +89,7 @@ def _run_iteration(self):
         self._stage_bundles()
         self._make_bundles()
         self._schedule_run_bundles()
+        self._bring_zombie_bundles_to_fail()
```
percyliang: rename to `_fail_unresponsive_bundles`

codalab/worker/bundle_manager.py
```diff
@@ -529,3 +533,20 @@ def _construct_run_message(self, worker, bundle):
         resources['request_network'] = bundle.metadata.request_network
 
         return message
+
+    def _bring_zombie_bundles_to_fail(self):
+        failure_message = "Bundle exceeds maximum time limit {}.".format(BUNDLE_TIMEOUT_SECONDS)
```
percyliang: This is confusing since the bundle time limit is another thing (`--request-time`).  Instead: Bundle has not responded in {} days.  I'd make it clear what the units here are.

codalab/worker/bundle_manager.py
```diff
@@ -529,3 +533,20 @@ def _construct_run_message(self, worker, bundle):
...
+        bundles_to_fail = (
+            self._model.batch_get_bundles(state=State.UPLOADING)
+            + self._model.batch_get_bundles(state=State.STAGED)
+            + self._model.batch_get_bundles(state=State.RUNNING)
+        )
+        now = time.time()
+
+        for bundle in bundles_to_fail:
+            if now - bundle.metadata.created > BUNDLE_TIMEOUT_SECONDS:
```
percyliang: This is not ideal because a bundle could have waited a while until it got staged.  Is there a last updated field for runs?  If not, then just put a comment here and leave it as is.

codalab/worker/bundle_manager.py
```diff
@@ -529,3 +533,20 @@ def _construct_run_message(self, worker, bundle):
         resources['request_network'] = bundle.metadata.request_network
 
         return message
+
+    def _bring_zombie_bundles_to_fail(self):
+        failure_message = "Bundle exceeds maximum time limit {}.".format(BUNDLE_TIMEOUT_SECONDS)
+        bundles_to_fail = (
+            self._model.batch_get_bundles(state=State.UPLOADING)
```
percyliang, line 545: Is there an easy way to make one database call here?  If not, okay to leave.

codalab/worker/bundle_manager.py
percyliang: Make a `BUNDLE_TIMEOUT_DAYS` field for readability

percyliang (changes requested):

codalab/worker/bundle_manager.py
candicegjing: I only found `last_updated` in bundles with `bundle_type == run`. I didn't find this field in other types of bundles. So maybe a better way would be checking on different field based on bundle_type?

codalab/worker/bundle_manager.py
candicegjing: I tried to put all three conditions in same call, but it complains about repeated keyword in this function: https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L126. It takes a dictionary.

codalab/worker/bundle_manager.py
```diff
@@ -17,7 +17,11 @@
 
 logger = logging.getLogger(__name__)
 
+
 WORKER_TIMEOUT_SECONDS = 60
+SECONDS_PER_DAY = 60 * 60 * 24
+# Maximum days to fail unresponsive bundles in uploading, staged and running state
+BUNDLE_TIMEOUT_DAYS = SECONDS_PER_DAY * 60
```
percyliang, line 24: `BUNDLE_TIMEOUT_DAYS` should be 60

codalab/worker/bundle_manager.py
percyliang: Ok, maybe okay to leave it as is.

codalab/worker/bundle_manager.py
```diff
@@ -529,3 +534,24 @@ def _construct_run_message(self, worker, bundle):
...
+        failure_message = "Bundle hasn't respond in {} days.".format(BUNDLE_TIMEOUT_DAYS)
+        bundles_to_fail = (
+            self._model.batch_get_bundles(state=State.UPLOADING)
+            + self._model.batch_get_bundles(state=State.STAGED)
+            + self._model.batch_get_bundles(state=State.RUNNING)
+        )
+        now = time.time()
+
+        for bundle in bundles_to_fail:
+            if now - bundle.metadata.created > BUNDLE_TIMEOUT_DAYS:
```
percyliang, line 552: multiply by SECONDS_PER_DAY here

codalab/worker/bundle_manager.py
```diff
@@ -529,3 +534,24 @@ def _construct_run_message(self, worker, bundle):
         resources['request_network'] = bundle.metadata.request_network
 
         return message
+
+    def _fail_unresponsive_bundles(self):
+        """
+        Fail bundles in uploading, staged and running state if we haven't heard from them for more than
+        BUNDLE_TIMEOUT_DAYS days.
+        """
+        failure_message = "Bundle hasn't respond in {} days.".format(BUNDLE_TIMEOUT_DAYS)
```
percyliang, line 543: Make message more specific: Bundle has been stuck in {state} for {} days, failing.

percyliang (changes requested):

Issue #1252, opened 2019-06-24
cl workers should report free disk space and what job they're running


percyliang:

percyliang:
`cl workers` should display have two more columns.

See `_checkin` in `worker.py` to see where the worker is generating other fields such as memory.

Probably want to add a function in `local_run_manager` to get the disk free by calling `df <worker_scratch_directory>` (should wrap it in a utility function).

candicegjing:
@percyliang Want to clarify with you on following questions:
1. "Free disk space" meaning free disk space per user or on whole system?
2. 2rd column is going to display all of the running jobs concatenated together?

percyliang:
1. On the whole system (`df` on the worker scratch directory).
2. This is a good question...this could get a bit long...ideally, we'd return this data to the CLI, and there would be command-line options to show or hide this information.  By default, we can just concatenate.  Later, we should build a frontend for this, in which case we can do fancier things in Javascript.

percyliang:
Fixed via #1296

PR #1296, opened 2019-07-19
Add available disk space and running jobs cl workers

candicegjing:
This ticket is to implement function mentioned #1252 .
1. Added a new column to `worker` table: `free_disk_bytes`
2. Added a new column to `cl worker`: `runs`

### DIFF
--- a/alembic/versions/2019072522_add_free_disk_bytes_0119bb78b368.py
+++ b/alembic/versions/2019072522_add_free_disk_bytes_0119bb78b368.py
@@ -1,0 +1,22 @@
+"""add free disk bytes
+
+Revision ID: 0119bb78b368
+Revises: d24522e04593
+Create Date: 2019-07-25 22:57:37.611894
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '0119bb78b368'
+down_revision = 'd24522e04593'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('worker', sa.Column('free_disk_bytes', sa.BigInteger(), nullable=True))
+
+
+def downgrade():
+    op.drop_column('worker', 'free_disk_bytes')
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1021,8 +1021,18 @@ class BundleCLI(object):
     def do_workers_command(self, args):
         client = self.manager.current_client()
         raw_info = client.get_workers_info()
+        raw_info.sort(key=lambda r: r['worker_id'])
 
-        columns = ['worker_id', 'cpus', 'gpus', 'memory_bytes', 'last_checkin', 'tag']
+        columns = [
+            'worker_id',
+            'cpus',
+            'gpus',
+            'memory',
+            'free_disk',
+            'last_checkin',
+            'tag',
+            'runs',
+        ]
 
         data = []
 
@@ -1032,15 +1042,16 @@ class BundleCLI(object):
                     'worker_id': worker['worker_id'],
                     'cpus': '{}/{}'.format(worker['cpus_in_use'], worker['cpus']),
                     'gpus': '{}/{}'.format(worker['gpus_in_use'], worker['gpus']),
-                    'memory_bytes': formatting.size_str(worker['memory_bytes']),
+                    'memory': formatting.size_str(worker['memory_bytes']),
+                    'free_disk': formatting.size_str(worker['free_disk_bytes']),
                     'last_checkin': '{} ago'.format(
                         formatting.duration_str(int(time.time()) - worker['checkin_time'])
                     ),
                     'tag': worker['tag'],
+                    'runs': ",".join([uuid[0:8] for uuid in worker['run_uuids']]),
                 }
             )
 
-        print >>self.stdout, 'Workers Info:'
         self.print_table(columns, data)
 
     @Commands.command(
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -352,6 +352,7 @@ worker = Table(
     Column('cpus', Integer, nullable=False),  # Number of CPUs on worker.
     Column('gpus', Integer, nullable=False),  # Number of GPUs on worker.
     Column('memory_bytes', BigInteger, nullable=False),  # Total memory of worker.
+    Column('free_disk_bytes', BigInteger, nullable=True),  # Available disk space on worker.
     Column(
         'checkin_time', DateTime, nullable=False
     ),  # When the worker last checked in with the bundle service.
--- a/codalab/model/worker_model.py
+++ b/codalab/model/worker_model.py
@@ -37,7 +37,9 @@ class WorkerModel(object):
         self._socket_dir = socket_dir
         self.shared_file_system = shared_file_system
 
-    def worker_checkin(self, user_id, worker_id, tag, cpus, gpus, memory_bytes, dependencies):
+    def worker_checkin(
+        self, user_id, worker_id, tag, cpus, gpus, memory_bytes, free_disk_bytes, dependencies
+    ):
         """
         Adds the worker to the database, if not yet there. Returns the socket ID
         that the worker should listen for messages on.
@@ -48,6 +50,7 @@ class WorkerModel(object):
                 'cpus': cpus,
                 'gpus': gpus,
                 'memory_bytes': memory_bytes,
+                'free_disk_bytes': free_disk_bytes,
                 'checkin_time': datetime.datetime.now(),
             }
             existing_row = conn.execute(
@@ -162,8 +165,10 @@ class WorkerModel(object):
                 'cpus': row.cpus,
                 'gpus': row.gpus,
                 'memory_bytes': row.memory_bytes,
+                'free_disk_bytes': row.free_disk_bytes,
                 'checkin_time': row.checkin_time,
                 'socket_id': row.socket_id,
+                # run_uuids will be set later
                 'run_uuids': [],
                 'dependencies': row.dependencies
                 and self._deserialize_dependencies(row.dependencies),
--- a/test_cli.py
+++ b/test_cli.py
@@ -25,7 +25,6 @@ import os
 import random
 import re
 import shutil
-import stat
 import subprocess
 import sys
 import time
@@ -1617,6 +1616,29 @@ def test(ctx):
     # edits that introduce unicode.
     # uuid = run_command([cl, 'upload', test_path('a.txt')])
     # run_command([cl, 'edit', uuid], 1)
+
+
+@TestModule.register('workers')
+def test(ctx):
+    # Run workers command
+    result = run_command([cl, 'workers'])
+    lines = result.split("\n")
+
+    # Output should contain at least 3 lines as following:
+    # worker_id        cpus  gpus  memory  free_disk  last_checkin  tag  runs
+    # -----------------------------------------------------------------------
+    # 7a343e1015c7(1)  0/2   0/0   2.0g    32.9g      2.0s ago
+    check_equals(True, len(lines) >= 3)
+
+    # Check header which includes 8 columns in total from output.
+    header = lines[0]
+    check_contains(
+        ['worker_id', 'cpus', 'gpus', 'memory', 'free_disk', 'last_checkin', 'tag', 'runs'], header
+    )
+
+    # Check number of not null values. First 6 columns should be not null. Column "tag" and "runs" could be empty.
+    worker_info = lines[2].split()
+    check_equals(True, len(worker_info) >= 6)
 
 
 if __name__ == '__main__':
--- a/worker/codalabworker/local_run/local_run_manager.py
+++ b/worker/codalabworker/local_run/local_run_manager.py
@@ -1,7 +1,6 @@
-import copy
 import logging
 import os
-from subprocess import check_output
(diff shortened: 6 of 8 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1037,6 +1046,8 @@ def do_workers_command(self, args):
                         formatting.duration_str(int(time.time()) - worker['checkin_time'])
                     ),
                     'tag': worker['tag'],
+                    'available_disk_space': worker['available_disk_space'] / 1024 ** 3,
```
percyliang: Call `formatting.size_str`

codalab/lib/bundle_cli.py
```diff
@@ -1022,7 +1022,16 @@ def do_workers_command(self, args):
...
 
-        columns = ['worker_id', 'cpus', 'gpus', 'memory_bytes', 'last_checkin', 'tag']
+        columns = [
+            'worker_id',
+            'cpus',
+            'gpus',
+            'memory_bytes',
+            'last_checkin',
+            'tag',
+            'available_disk_space',
```
percyliang: For consistency, append `_bytes`.  I think `free_disk_bytes` is a bit clearer / shorter.

codalab/model/worker_model.py
```diff
@@ -49,6 +51,7 @@ def worker_checkin(self, user_id, worker_id, tag, cpus, gpus, memory_bytes, depe
                 'gpus': gpus,
                 'memory_bytes': memory_bytes,
                 'checkin_time': datetime.datetime.now(),
+                'available_disk_space': available_disk_space,
```
percyliang: Move this line to right after `memory_bytes` everywhere to try to keep the parallel structure

codalab/model/worker_model.py
```diff
@@ -167,6 +170,7 @@ def get_workers(self):
                 'run_uuids': [],
                 'dependencies': row.dependencies
                 and self._deserialize_dependencies(row.dependencies),
+                'available_disk_space': row.available_disk_space,
```
percyliang: Same here

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +359,22 @@ def memory_bytes(self):
...
+        # Return -1 when there is an error.
+        if error:
+            logger.info(error.strip())
+            return -1
+
+        if output:
+            lines = output.decode().split("\n")
+            index = lines[0].split().index("Available")
+            # df reports the disk space in unit of 1K blocks by default.
+            # Original df output times 1024 will transfer the result into bytes.
```
percyliang: Simply write: '...which we convert into bytes'

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +359,22 @@ def memory_bytes(self):
         except ValueError:
             # Fallback to sysctl when os.sysconf('SC_PHYS_PAGES') fails on OS X
             return int(check_output(['sysctl', '-n', 'hw.memsize']).strip())
+
+    @property
+    def available_disk_space(self):
+        """
+        Available disk space by bytes of this RunManager.
+        """
+        p = Popen(["df", "./"], stdout=PIPE)
```
percyliang: Use `work_dir` instead of `.`.

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +359,22 @@ def memory_bytes(self):
...
+        output, error = p.communicate()
+        # Return -1 when there is an error.
+        if error:
+            logger.info(error.strip())
+            return -1
+
+        if output:
+            lines = output.decode().split("\n")
+            index = lines[0].split().index("Available")
```
percyliang: What happens if this code throws an exception for whatever reason?  Is it caught somewhere?  Just make sure this doesn't crash the worker.

percyliang (changes requested):

alembic/versions/2019072522_add_free_disk_bytes_0119bb78b368.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '0119bb78b368'
+down_revision = 'd24522e04593'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
```
percyliang: this should have the command for adding the column and shouldn't be empty

codalab/model/tables.py
```diff
@@ -352,6 +352,7 @@
     Column('cpus', Integer, nullable=False),  # Number of CPUs on worker.
     Column('gpus', Integer, nullable=False),  # Number of GPUs on worker.
     Column('memory_bytes', BigInteger, nullable=False),  # Total memory of worker.
+    Column('free_disk_bytes', BigInteger, nullable=False),  # Available disk space on worker
```
percyliang: nit: for consistency, end with period

codalab/lib/bundle_cli.py
```diff
@@ -1033,10 +1042,12 @@ def do_workers_command(self, args):
                     'cpus': '{}/{}'.format(worker['cpus_in_use'], worker['cpus']),
                     'gpus': '{}/{}'.format(worker['gpus_in_use'], worker['gpus']),
                     'memory_bytes': formatting.size_str(worker['memory_bytes']),
+                    'free_disk_bytes': formatting.size_str(worker['free_disk_bytes']),
```
percyliang: can you change the column headers to `memory` and `free_disk` since the values are already formatted, so the `bytes` is inaccurate?  Also will save some horizontal space.

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +360,26 @@ def memory_bytes(self):
...
+                return -1
+
+            if output:
+                lines = output.decode().split("\n")
+                index = lines[0].split().index("Available")
+                # We convert the original result from df command in unit of 1Kbyte blocks into bytes.
+                return int(lines[1].split()[index]) * 1024
+
+        except OSError as e:
+            logger.info("Failed to run command '%s' : %s".format("df " + self._work_dir, str(e)))
```
percyliang: Should use logger.error for errors
Also, remove space before ':'

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +360,26 @@ def memory_bytes(self):
...
+        """
+        Available disk space by bytes of this RunManager.
+        """
+        try:
+            p = Popen(["df", self._work_dir], stdout=PIPE)
+            output, error = p.communicate()
+            # Return -1 when there is an error.
+            if error:
+                logger.info(error.strip())
+                return -1
```
percyliang: Hm, returning `-1` doesn't seem great because the semantics of an integer return value become more complex.  Maybe `None` is a bit better?

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +360,26 @@ def memory_bytes(self):
...
+
+            if output:
+                lines = output.decode().split("\n")
+                index = lines[0].split().index("Available")
+                # We convert the original result from df command in unit of 1Kbyte blocks into bytes.
+                return int(lines[1].split()[index]) * 1024
+
+        except OSError as e:
```
percyliang: This isn't the only error that can happen...  if the format of `df` changes for some reason, then the parsing logic can fail too

percyliang (changes requested):

worker/codalabworker/local_run/local_run_manager.py
candicegjing: I also thought about the problems of parsing `df` result. `shutil` in Python3 provide `disk_usage` function which will return same result as `df`, but not in Python2. `psutil` in Python2 provides the same function. Not sure if this would be a better option?

worker/codalabworker/local_run/local_run_manager.py
candicegjing: I also updated the value of `nullable` in column definition and alembic revision file to be `True`.

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +360,32 @@ def memory_bytes(self):
...
+            output, error = p.communicate()
+            # Return None when there is an error.
+            if error:
+                logger.error(error.strip())
+                return None
+
+            if output:
+                lines = output.decode().split("\n")
+                header = lines[0]
+                if header.find("Available") > -1 and len(lines) > 1:
```
percyliang: Write `!= -1` to be more precise with the semantics and I think writing `>= 2` instead of `> 1` makes it easier to interpret.

codalab/lib/bundle_cli.py
```diff
@@ -1032,11 +1041,13 @@ def do_workers_command(self, args):
...
                     'cpus': '{}/{}'.format(worker['cpus_in_use'], worker['cpus']),
                     'gpus': '{}/{}'.format(worker['gpus_in_use'], worker['gpus']),
-                    'memory_bytes': formatting.size_str(worker['memory_bytes']),
+                    'memory': formatting.size_str(worker['memory_bytes']),
+                    'free_disk': formatting.size_str(worker['free_disk_bytes']),
                     'last_checkin': '{} ago'.format(
                         formatting.duration_str(int(time.time()) - worker['checkin_time'])
                     ),
                     'tag': worker['tag'],
+                    'running_jobs': ",".join(worker['run_uuids'][0:8]) or "NA",
```
percyliang: - Isn't `worker['run_uuids']` supposed to be a list of uuids?
- I'd remove `NA` and just return `''` if there are no runs - n/a means that it doesn't make sense, but it does here - it just happens to be empty.
- I'd replace `running_jobs` with just `runs` for simplicity.

codalab/model/worker_model.py
```diff
@@ -162,6 +165,7 @@ def get_workers(self):
                 'cpus': row.cpus,
                 'gpus': row.gpus,
                 'memory_bytes': row.memory_bytes,
+                'free_disk_bytes': row.free_disk_bytes,
                 'checkin_time': row.checkin_time,
                 'socket_id': row.socket_id,
                 'run_uuids': [],
```
percyliang, line 172: Put a comment that this is set later

worker/codalabworker/local_run/local_run_manager.py
percyliang: Actually, reflecting on this, it's probably better not to have these conditions at all and just let any errors fall to the `except` statement to keep this code cleaner.  I separate errors into two categories: (i) normal errors that happen - these should be handled explicitly with checks and appropriate error messages, and (ii) abnormal errors that suggest we have a bug or something is deeply wrong - if this happens, someone needs to go investigate - for that, it's fine to just have the blanket `except Exception as e`.

percyliang (changes requested):

worker/codalabworker/local_run/local_run_manager.py
candicegjing: I think you are right. If str.index() isn't able to find the column "Available", it will throw an exception. If return value from df isn't a valid format, when we cast it to integer, it will also throw an exception. Both cases will be caught by the outer try-catch block.

codalab/model/worker_model.py
```diff
@@ -154,6 +157,7 @@ def get_workers(self):
             ).fetchall()
             worker_run_rows = conn.execute(cl_worker_run.select()).fetchall()
 
+        # free_disk_bytes will be set later.
```
percyliang: You mean `run_uuids`?  Make this comment on the same line as `'run_uuids': []`.

worker/codalabworker/local_run/local_run_manager.py
```diff
@@ -360,3 +360,27 @@ def memory_bytes(self):
...
+            output, error = p.communicate()
+            # Return None when there is an error.
+            if error:
+                logger.error(error.strip())
+                return None
+
+            if output:
+                lines = output.decode().split("\n")
+                index = lines[0].split().index("Available")
+                # We convert the original result from df command in unit of 1Kbyte blocks into bytes.
```
percyliang: nit: 1Kbyte => 1KB

percyliang (approved):
A few remaining comments...otherwise LGTM.

One thing: could you add a test to `test_cli.py` that just calls `cl workers` (this code is currently not exercised in tests).

test_cli.py
```diff
@@ -1619,6 +1618,23 @@ def test(ctx):
     # run_command([cl, 'edit', uuid], 1)
 
 
+@TestModule.register('workers')
+def test(ctx):
+    # Run workers command
+    result = run_command([cl, 'workers'])
+    # Output should contain 3 lines
```
percyliang: I don't you can assume there's exactly one worker (in any case, should document why 3).  Just assume there's at least one.

candicegjing: I see. I ran test locally and it always returns me 1 worker. So I just put 3 lines there.

codalab/model/worker_model.py
candicegjing: Hmm, actually, why this will be set later?

percyliang: All I mean is that currently, it looks like from `'run_uuids': []` that it's set to empty, but a few lines later, it gets populated.

candicegjing: I see.

PR #1298, opened 2019-07-21
Print subprocess output live

epicfaace:
Subprocess output was suppressed for docker commands with the new setup, which means that test outputs no longer show up. This change fixes that and shows output live, as well as shows the error when the command fails.


epicfaace:
Actually, this code isn't compatible with python 2 :(

codalab_service.py
```diff
@@ -606,9 +606,16 @@ def _run_compose_cmd(self, cmd):
...
         if not self.args.dry_run:
-            subprocess.check_call(
-                command_string, cwd=self.compose_cwd, env=self.compose_env, shell=True
-            )
+            try:
+                popen = subprocess.Popen(
+                    command_string, cwd=self.compose_cwd, env=self.compose_env, shell=True,
+                    stdout=subprocess.PIPE
+                )
+                for stdout_line in popen.stdout:
```
percyliang, line 592: Do we need to do this for stderr too?

epicfaace: Not sure...

epicfaace: Probably.

codalab_service.py
```diff
@@ -579,9 +580,20 @@ def _run_compose_cmd(self, cmd):
...
+                popen = subprocess.Popen(
+                    command_string,
+                    cwd=self.compose_cwd,
+                    env=self.compose_env,
+                    shell=True,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                )
+                for stdout_line in popen.stdout:
+                    print("process: " + stdout_line.decode(), end="")
```
```suggestion
                    print("process: " + stdout_line.decode('utf-8').encode('ascii', errors='replace'), end="")
```

codalab_service.py
```diff
@@ -579,9 +580,20 @@ def _run_compose_cmd(self, cmd):
...
+            except subprocess.CalledProcessError as e:
+                print("CalledProcessError: {}, {}".format(str(e), e.output))
```
```suggestion
                print("CalledProcessError: {}, {}".format(str(e), e.output.decode('utf-8').encode('ascii', errors='replace')))
```

codalab_service.py
```diff
@@ -579,9 +580,27 @@ def _run_compose_cmd(self, cmd):
...
+                popen = subprocess.Popen(
+                    command_string,
+                    cwd=self.compose_cwd,
+                    env=self.compose_env,
+                    shell=True,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                )
+                for stdout_line in popen.stdout:
+                    print(
```
percyliang, line 593: Yuck :-)  Can we put a comment that says this will be greatly simplified with Python 3?

percyliang (approved):

codalab_service.py
epicfaace: Actually, this won't be simplified with python 3. The encode/decode calls are there because Travis only prints ascii. Once this is python 3, we just need to remove the "from future" import at the beginning.

percyliang: Really?  I feel like I've seen unicode being printed out on Travis.  Is it a matter of setting some encoding environment variable?

PR #1300, opened 2019-07-24
Add timeout bundle function

candicegjing:
This PR is created to correct branch naming typo in #1295 and continue fixing the rest of the code review comments from #1295 .
Testing:
1. To test bundle in **uploading** state, I created a 2G data file, and set the timeout to be 30 seconds. (I measured the time of uploading the entire file first. 30 seconds is before uploading finished.)
2. To test bundle in **staged** state, I submitted a bundle without specifying `requested-memory`, so that it stucks in `staged` state in my local environment.
3. To test bundle in **running** state, I created a simple file which will sleep every second for total of 300 seconds and set the timeout to be 200 seconds.
All three test cases passed in my local environment.

### DIFF
--- a/codalab/worker/bundle_manager.py
+++ b/codalab/worker/bundle_manager.py
@@ -17,7 +17,11 @@ from codalabworker.bundle_state import State
 
 logger = logging.getLogger(__name__)
 
+
 WORKER_TIMEOUT_SECONDS = 60
+SECONDS_PER_DAY = 60 * 60 * 24
+# Fail unresponsive bundles in uploading, staged and running state after this many days.
+BUNDLE_TIMEOUT_DAYS = 60
 
 
 class BundleManager(object):
@@ -86,6 +90,7 @@ class BundleManager(object):
         self._stage_bundles()
         self._make_bundles()
         self._schedule_run_bundles()
+        self._fail_unresponsive_bundles()
 
     def _schedule_run_bundles(self):
         """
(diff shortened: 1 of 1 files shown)


codalab/worker/bundle_manager.py
```diff
@@ -17,7 +17,11 @@
 
 logger = logging.getLogger(__name__)
 
+
 WORKER_TIMEOUT_SECONDS = 60
+SECONDS_PER_DAY = 60 * 60 * 24
+# Maximum days to fail unresponsive bundles in uploading, staged and running state
```
percyliang: Revivse: "Fail unresponsive bundles ... after this many days"

codalab/worker/bundle_manager.py
```diff
@@ -529,3 +534,25 @@ def _construct_run_message(self, worker, bundle):
...
+        """
+        bundles_to_fail = (
+            self._model.batch_get_bundles(state=State.UPLOADING)
+            + self._model.batch_get_bundles(state=State.STAGED)
+            + self._model.batch_get_bundles(state=State.RUNNING)
+        )
+        now = time.time()
+
+        for bundle in bundles_to_fail:
+            if now - bundle.metadata.created > BUNDLE_TIMEOUT_DAYS * SECONDS_PER_DAY:
```
percyliang, line 553: Add a comment saying this should really be the last updated time logically, but we are using created for simplicity.

percyliang (approved):

Issue #1314, opened 2019-08-06
[USER ISSUE] How to setup email for own server


gumulka:
**Describe the bug**
I can't find a way to send the validation mails in my own codalab-Instance

**To Reproduce**
Steps to reproduce the behavior :
See https://github.com/codalab/codalab-worksheets/wiki/Server-Setup
1. Download codalab,
2. start server
3. Sign up as new user
4. wait forever for a mail

**Expected behavior**
See the configuration for Mail somewhere in server-setup

**Screenshots**
N/A

**Additional context**

PR #1315, opened 2019-08-07
Make all mail parameters configurable

gumulka:
It is now possible to change the port and sender configurations for all instances, with still having reasonable defaults.

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -395,9 +395,10 @@ class CodaLabManager(object):
                 host=self.config['email']['host'],
                 user=self.config['email'].get('user', 'noreply@codalab.org'),
                 password=self.config['email'].get('password', None),
-                use_tls=True,
-                default_sender='CodaLab <noreply@codalab.org>',
-                server_email='noreply@codalab.org',
+                use_tls=self.config['email'].get('use_tls',True),
+                port=self.config['email'].get('port', 587),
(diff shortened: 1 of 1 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -395,9 +395,10 @@ def emailer(self):
                 host=self.config['email']['host'],
                 user=self.config['email'].get('user', 'noreply@codalab.org'),
                 password=self.config['email'].get('password', None),
-                use_tls=True,
-                default_sender='CodaLab <noreply@codalab.org>',
-                server_email='noreply@codalab.org',
+                use_tls=self.config['email'].get('use_tls',True),
+                port=self.config['email'].get('port', 587),
+                default_sender=self.config['email'].get('default_sender','CodaLab <noreply@codalab.org>'),
```
percyliang: Could we change the config keys:
- `default_sender` -> `sender_from`
- `server_email` -> `sender_email`
to make it semantically more clear?

percyliang (approved):
Thanks!  Just a small request for naming change.

epicfaace:
Do we already have config parameters documented? If so, should probably update the documentation.

codalab/lib/codalab_manager.py
gumulka: Sure. and already done

gumulka:
> Do we already have config parameters documented? If so, should probably update the documentation.

No, you don't. See #1314

PR #1317, opened 2019-08-11
Update docs

percyliang:
- Took a pass over all the MD files and made sure they were up to date.
- Also, auto-generate the CLI reference just like we do for the REST API reference.


docs/Server-Setup.md
```diff
@@ -126,10 +126,25 @@ To fix any style issues for the Python code:
...
+    virtualenv -p python2.7 venv2.7
+    venv2.7/bin/pip install -r requirements-server.txt
+    venv2.7/bin/python scripts/gen-rest-docs.py
+    venv2.7/bin/python scripts/gen-cli-docs.py
+
+To generate the readthedocs documentation to preview locally (outputs to `site`):
+
+    virtualenv -p python2.7 venv2.7
+    venv2.7/bin/pip install -r requirements.docs.txt
+    venv2.7/bin/mkdocs build
```
epicfaace: We could also add something about `mkdocs serve` which does a live preview.

epicfaace (approved):

PR #1325, opened 2019-08-20
Improve things around Travis builds

percyliang:
- Pull Docker images before building Docker images to use as a cache
- Flush more often so outputs from tests show up in real-time
- Tie versions to git branches to simplify things

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -587,7 +587,7 @@ class CodalabServiceManager(object):
             try:
                 subprocess.check_call(command_string, shell=True, cwd=self.root_dir)
                 success = True
-            except OSError as e:
+            except subprocess.CalledProcessError as e:
                 success = False
                 if not allow_fail:
                     raise e
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -555,9 +555,10 @@ class CodalabServiceManager(object):
 
     def build_image(self, image):
         print_header('Building {} image'.format(image))
+        docker_image = 'codalab/{}:{}'.format(image, self.args.version)
         self._run_docker_cmd(
-            'build -t codalab/%s:%s -f docker/dockerfiles/Dockerfile.%s .'
-            % (image, self.args.version, image)
+            'build --cache-from %s -t %s -f docker/dockerfiles/Dockerfile.%s .'
+            % (docker_image, docker_image, image)
         )
 
     def push_image(self, image):
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -528,7 +528,7 @@ class CodalabServiceManager(object):
                     raise
         for bundle_store in self.args.bundle_stores:
             try:
-                os.makedirs(self.args.bundle_store)
+                os.makedirs(bundle_store)
             except OSError as e:
                 if e.errno != errno.EEXIST:
                     raise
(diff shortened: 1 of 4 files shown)


codalab_service.py
```diff
@@ -579,21 +616,31 @@ def _run_compose_cmd(self, cmd):
...
                     command_string,
                     cwd=self.compose_cwd,
                     env=self.compose_env,
                     shell=True,
                     stdout=subprocess.PIPE,
-                    stderr=subprocess.PIPE,
+                    stderr=subprocess.STDOUT,
                 )
-                for stdout_line in popen.stdout:
+                # Note: don't do `for stdout_line in popen.stdout` because that buffers.
```
epicfaace, line 631: Looks good. What exactly was the issue with using `stderr=subprocess.STDOUT` and `for stdout_line in popen.stdout`?

codalab_service.py
```diff
@@ -203,7 +209,7 @@ def _get_parser():
                 '--version',
                 '-v',
                 type=str,
-                help='CodaLab version to use for building and deployment',
+                help='CodaLab version to use for building and deployment (should be branch name)',
```
epicfaace: Since it's been removed from Server-Setup.md, can we add a note that it's only used for Travis?

Also, what's the default value for this when running locally?

test_cli.py
```diff
@@ -133,7 +134,9 @@ def run_command(args, expected_exit_code=0, max_output_chars=256, env=None, incl
     else:
         colorize = Colorizer.cyan
     print(colorize(" (exit code %s, expected %s)" % (exitcode, expected_exit_code)))
+    sys.stdout.flush()
     print(sanitize(output, max_output_chars))
+    sys.stdout.flush()
```
epicfaace, line 139: What did adding these flush calls fix?

codalab_service.py
```diff
@@ -213,6 +219,12 @@ def _get_parser():
                 help='If specified, mount local code for frontend so that changes are reflected right away',
                 default=argparse.SUPPRESS,
             )
+            cmd.add_argument(
+                '--pull',
+                action='store_true',
+                help='If specified, pull images to Docker Hub (for caching)',
```
```suggestion
                help='If specified, pull images from Docker Hub (for caching)',
```

codalab_service.py
percyliang: Done.  Defaults to the git branch name.

test_cli.py
percyliang: `docker exec codalab_rest-server_1 python test_cli.py` wasn't flushing...

codalab_service.py
percyliang: It would just buffer the output, so that the output wouldn't show up line by line.

codalab_service.py
```diff
@@ -203,7 +209,7 @@ def _get_parser():
                 '--version',
                 '-v',
                 type=str,
-                help='CodaLab version to use for building and deployment',
+                help='CodaLab version to use for building and deployment (defaults to branch name, set only for Travis CI)',
```
epicfaace, line 212:
```suggestion
                help='CodaLab version to use for building and deployment (defaults to branch name, needs to be set only for Travis CI)',
```

Otherwise it sounds like the default is set only on Travis CI (and not locally).

PR #1328, opened 2019-08-21
Make deployment easier

percyliang:
Two main changes:
- `./codalab-service.py pull` grabs the images.
- `./codalab-service.py start` does migrations for you so you don't have to do an `update` explicitly, which was error prone before (if you forget).

Also, some drive-by improvements.

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -38,7 +38,7 @@ def should_run_service(args, service):
 
 
 def need_image_for_service(args, image):
-    """Does `image` support a service we want to run."""
+    """Does `image` support a service we want to run?"""
     for service, service_image in SERVICE_TO_IMAGE.items():
         if should_run_service(args, service) and image == service_image:
             return True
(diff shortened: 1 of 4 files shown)


codalab_service.py
```diff
@@ -454,7 +468,7 @@ def resolve_env_vars(args):
             'CODALAB_HTTP_PORT': args.http_port,
             'CODALAB_VERSION': args.version,
             'CODALAB_WORKER_NETWORK_NAME': '%s-worker-network' % args.instance_name,
-            #'PATH': os.environ['PATH'],
+            'PATH': os.environ['PATH'],
```
epicfaace, line 471: Oh yeah, thanks for re-adding this (#1161). We need to eventually include macOS on Travis tests.

codalab_service.py
```diff
@@ -37,16 +37,21 @@ def should_run_service(args, service):
     )
 
 
+def need_image_for_service(args, image):
+    """Does `image` support a service we want to run."""
```
```suggestion
    """Does `image` support a service we want to run?"""
```

epicfaace (approved):

Issue #1280, opened 2019-07-08
add monitor.py to codalab_service.py


percyliang:

candicegjing:
Some questions about this ticket:
1. When do we want to start the monitoring script? Do we want to start it in the same time as rest-server service? Or do we want it to be run in the same way as test_cli.py?
2. Since monitor.py keeps running a while loop forever, do we want it to continuously send back monitoring status? When I tested calling monitor.py from codalab_service.py, somehow all the log messages were blocked to show up in the caller process until it got killed manually. So I am want to  confirm its right behavior here. 
3. Where do we want to place the monitoring log file? Since container is stateless, I guess it might not be a good idea to store it there. Then another option would be somewhere in the local machine?

percyliang:
1) It shouldn't be started by default (which is local dev).  But if you do `-s monitor` then it should start it.
2) It should continue to run continuously in a loop and emails once in a while.  So we should make sure the email part is configured properly.
3) By default, it can just be in the container, but you can add a flag (like how `codalab_home` and `bundle_stores` are handled).

candicegjing:
For 3, my understanding is that if we output logs to container, we won't be able to retrieve it, as container is stateless.

percyliang:
You can still retrieve it while the container is running (`docker logs`).  This is probably fine for now actually so don't even worry about the flag.  The main purpose of the monitor script is to send out the emails anyway.

PR #1329, opened 2019-08-21
Add monitor script to codalab service script

candicegjing:
This PR is to implement functions for #1280 
1. I tested sending email using my personal gmail account and it worked. But I haven't figured out how to make the right authless SMTP setting yet. I tried several ways but none of them are working. They all were asking for authentication.
2. Since monitor.py reads from config.json. Not sure if we want to add this part of email configuration to the file when it gets generated.
3. I removed `localhost::` when we run `cl work` command to avoid  `codalab.client.json_api_client.JsonApiException: Unable to fetch worksheets: <urlopen error [Errno 99] Cannot assign requested address>` exception.

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -10,9 +10,12 @@ import argparse
 import errno
 import os
 import subprocess
-import sys
 
-SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker']
+DEFAULT_SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker', 'init']
+
+ALL_SERVICES = DEFAULT_SERVICES + ['test', 'monitor']
+
+ALL_NO_SERVICES = ['no-' + service for service in ALL_SERVICES]
 
 SERVICE_TO_IMAGE = {
     'frontend': 'frontend',
@@ -28,13 +31,12 @@ def print_header(description):
 
 def should_run_service(args, service):
     # `default` is generally used to bring up everything for local dev or quick testing.
-    # `default-no-worker` is generally used for real deployment since we don't
-    # want a worker running on the same machine.
-    return (
-        service in args.services
-        or (service != 'test' and 'default' in args.services)
-        or (service != 'test' and service != 'worker' and 'default-no-worker' in args.services)
-    )
+    # `default no-worker` is generally used for real deployment since we don't want a worker running on the same machine.
+    services = [] if args.services is None else args.services
+    if 'default' in args.services:
+        services.extend(DEFAULT_SERVICES)
+
+    return (service in services) and ('no-' + service not in services)
 
 
 def need_image_for_service(args, image):
@@ -100,6 +102,10 @@ class CodalabArgs(argparse.Namespace):
         'ssl_key_file': None,
         'follow': False,
         'tail': None,
+        'admin_email': None,
+        'email_host': None,
+        'email_user': None,
+        'email_password': None,
     }
 
     ARG_TO_ENV_VAR = {
@@ -125,6 +131,10 @@ class CodalabArgs(argparse.Namespace):
         'mysql_port': 'CODALAB_MYSQL_PORT',
         'ssl_cert_file': 'CODALAB_SSL_CERT_FILE',
         'ssl_key_file': 'CODALAB_SSL_KEY_FILE',
+        'admin_email': 'CODALAB_ADMIN_EMAIL',
+        'email_host': 'CODALAB_EMAIL_HOST',
+        'email_user': 'CODALAB_EMAIL_USER',
+        'email_password': 'CODALAB_EMAIL_PASSWORD',
     }
 
     @staticmethod
@@ -301,7 +311,7 @@ class CodalabArgs(argparse.Namespace):
             '-s',
             nargs='*',
             help='List of services to run',
-            choices=SERVICES + ['default', 'default-no-worker', 'init', 'update', 'test'],
+            choices=ALL_SERVICES + ALL_NO_SERVICES + ['default'],
             default=argparse.SUPPRESS,
         )
 
@@ -311,6 +321,30 @@ class CodalabArgs(argparse.Namespace):
             '--mysql-root-password',
             type=str,
             help='Root password for the database',
+            default=argparse.SUPPRESS,
+        )
+
+        #  EMAIL CREDENTIALS
+
+        start_cmd.add_argument(
+            '--admin-email',
+            type=str,
+            help='Recipient email address for receiving email',
+            default=argparse.SUPPRESS,
+        )
+        start_cmd.add_argument(
+            '--email-host', type=str, help='Email host for sending email', default=argparse.SUPPRESS
+        )
+        start_cmd.add_argument(
+            '--email-user',
+            type=str,
+            help='Username of email account for sending email',
+            default=argparse.SUPPRESS,
+        )
+        start_cmd.add_argument(
+            '--email-password',
+            type=str,
+            help='Password of email account for sending email',
             default=argparse.SUPPRESS,
         )
 
@@ -392,7 +426,7 @@ class CodalabArgs(argparse.Namespace):
             nargs='*',
             default='default',
             help='Services to print logs for',
-            choices=SERVICES + ['default'],
+            choices=ALL_SERVICES + ['default'],
         )
         logs_cmd.add_argument(
             '--follow',
@@ -415,7 +449,7 @@ class CodalabArgs(argparse.Namespace):
             'service',
             metavar='SERVICE',
             type=str,
-            choices=SERVICES,
+            choices=ALL_SERVICES,
             help='Service container to run command on',
         )
         run_cmd.add_argument('cmd', metavar='CMD', type=str, help='Command to run')
@@ -495,6 +529,15 @@ class CodalabServiceManager(object):
             environment['CODALAB_SSL_KEY_FILE'] = args.ssl_key_file
         if 'DOCKER_HOST' in os.environ:
             environment['DOCKER_HOST'] = os.environ['DOCKER_HOST']
+        if args.admin_email:
+            environment['CODALAB_ADMIN_EMAIL'] = args.admin_email
+        if args.email_host:
+            environment['CODALAB_EMAIL_HOST'] = args.email_host
+        if args.email_user:
+            environment['CODALAB_EMAIL_USER'] = args.email_user
+        if args.email_password:
+            environment['CODALAB_EMAIL_PASSWORD'] = args.email_password
+
         return environment
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


codalab_service.py
```diff
@@ -28,12 +27,17 @@ def print_header(description):
...
 def should_run_service(args, service):
     # `default` is generally used to bring up everything for local dev or quick testing.
-    # `default-but-worker` is generally used for real deployment since we don't
+    # `default-no-worker` is generally used for real deployment since we don't
     # want a worker running on the same machine.
     return (
         service in args.services
-        or (service != 'test' and 'default' in args.services)
-        or (service != 'test' and service != 'worker' and 'default-no-worker' in args.services)
+        or (service != 'test' and service != 'monitor' and 'default' in args.services)
```
percyliang: This logic is getting a bit convoluted...can we simplify (and generalize) by just defining a new `DEFAULT_SERVICES` list, and then making a "no" version of every service (i.e. `['no-' + service for service in ALL_SERVICES]`), so that you can do `-s default no-worker` instead?

monitor.py
```diff
@@ -265,7 +265,7 @@ def check_disk_space(paths):
 
 
 # Make sure we can connect (might prompt for username/password)
-if subprocess.call(['cl', 'work', 'localhost::']) != 0:
+if subprocess.call(['cl', 'work']) != 0:
```
percyliang, line 268: Why didn't `localhost::` work?  `localhost` should be a default alias...

percyliang (changes requested):

percyliang:
For setting up email, we should add some flags/environment variables into `codalab_service.py` which are then copied into `config.json` during the `init` service with `cl config`...

For testing email, did you try using the sendgrid account?

monitor.py
candicegjing: This is what I got when `localhost::` specified:
```
process: Traceback (most recent call last):
process:   File "/usr/local/bin/cl", line 11, in <module>
process:     load_entry_point('codalab', 'console_scripts', 'cl')()
process:   File "/opt/codalab-worksheets/codalab/bin/cl.py", line 117, in main
process:     cli.do_command(sys.argv[1:])
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 843, in do_command
process:     structured_result = command_fn()
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 837, in <lambda>
process:     command_fn = lambda: args.function(self, args)
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 2779, in do_work_command
process:     client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 707, in parse_client_worksheet_uuid
process:     worksheet_uuid = self.resolve_worksheet_uuid(client, base_worksheet_uuid, spec)
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 635, in resolve_worksheet_uuid
process:     'worksheets', params={'base': base_worksheet_uuid, 'specs': [worksheet_spec]}
process:   File "/opt/codalab-worksheets/codalab/client/json_api_client.py", line 425, in fetch_one
process:     results = self.fetch(resource_type, resource_id=resource_id, params=params)
process:   File "/opt/codalab-worksheets/codalab/client/json_api_client.py", line 50, in wrapper
process:     sys.exc_info()[2],
process:   File "/opt/codalab-worksheets/codalab/client/json_api_client.py", line 16, in wrapper
process:     return f(*args, **kwargs)
process:   File "/opt/codalab-worksheets/codalab/client/json_api_client.py", line 415, in fetch
process:     query_params=self._pack_params(params),
process:   File "/opt/codalab-worksheets/worker/codalabworker/rest_client.py", line 91, in _make_request
process:     with closing(urllib2.urlopen(request)) as response:
process:   File "/usr/lib/python2.7/urllib2.py", line 154, in urlopen
process:     return opener.open(url, data, timeout)
process:   File "/usr/lib/python2.7/urllib2.py", line 429, in open
process:     response = self._open(req, data)
process:   File "/usr/lib/python2.7/urllib2.py", line 447, in _open
process:     '_open', req)
process:   File "/usr/lib/python2.7/urllib2.py", line 407, in _call_chain
process:     result = func(*args)
process:   File "/usr/lib/python2.7/urllib2.py", line 1228, in http_open
process:     return self.do_open(httplib.HTTPConnection, req)
process:   File "/usr/lib/python2.7/urllib2.py", line 1198, in do_open
process:     raise URLError(err)
process: codalab.client.json_api_client.JsonApiException: Unable to fetch worksheets: <urlopen error [Errno 99] Cannot assign requested address>
```

candicegjing: With this logic : https://github.com/codalab/codalab-worksheets/blob/master/monitor.py#L268-L269, monitor.py won't run once it throws an error.

percyliang: I was trying to figure out why the error was happening.  I think the problem is that `localhost::` points to http://localhost, but inside docker, you need to use `http://rest-server`.  The `localhost::` was probably there for a reason, but I think this case it's probably fine.

candicegjing: The thing is that if you run `cl work localhost::` within the docker directly or `docker exec -it codalab_rest-server_1 cl work localhost::`, both of them worked fine. But with the wait-for-it.sh wrapper, it failed. I tested `cl work`, it seems default to localhost...

codalab_service.py
```diff
@@ -28,12 +40,15 @@ def print_header(description):
...
 def should_run_service(args, service):
     # `default` is generally used to bring up everything for local dev or quick testing.
-    # `default-but-worker` is generally used for real deployment since we don't
-    # want a worker running on the same machine.
-    return (
-        service in args.services
-        or (service != 'test' and 'default' in args.services)
-        or (service != 'test' and service != 'worker' and 'default-no-worker' in args.services)
+    # `default no-worker` is generally used for real deployment since we don't want a worker running on the same machine.
+    # default_services includes 'mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker', 'init', 'update'
```
percyliang: don't list this here to avoid having this list in multiple places

codalab_service.py
```diff
@@ -28,12 +40,15 @@ def print_header(description):
...
+    default_services = DEFAULT_SERVICES[0:8] if 'default' in args.services else []
```
percyliang: Don't use absolute indexing (8, 3)...

codalab_service.py
```diff
@@ -10,9 +10,21 @@
...
-SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker']
+DEFAULT_SERVICES = [
+    'mysql',
+    'nginx',
+    'frontend',
+    'rest-server',
+    'bundle-manager',
+    'worker',
+    'init',
+    'update',
```
percyliang: You should have `DEFAULT_SERVICES` for the first 8 elements, and then `ALL_SERVICES = DEFAULT_SERVICES + ['test', 'monitor']`.

You need to merge master in (`update` is gone now).

codalab_service.py
```diff
@@ -288,6 +311,33 @@ def _get_parser():
             default=argparse.SUPPRESS,
         )
 
+        #  MONITOR EMAIL CREDENTIALS
+
+        start_cmd.add_argument(
+            '--admin_email',
```
percyliang: `--admin-email` (use dashes everywhere for flags)

codalab_service.py
```diff
@@ -288,6 +311,33 @@ def _get_parser():
...
+        start_cmd.add_argument(
+            '--email_host',
+            type=str,
+            help='Email host for sending monitoring email',
+            default=argparse.SUPPRESS,
+        )
+        start_cmd.add_argument(
+            '--email_user',
+            type=str,
+            help='Username of email account for sending monitoring email',
```
percyliang: These emails should not just be for monitoring, but all communictions.

codalab_service.py
```diff
@@ -366,7 +416,7 @@ def _get_parser():
             nargs='*',
             default='default',
             help='Services to print logs for',
-            choices=SERVICES + ['default'],
+            choices=DEFAULT_SERVICES[0:6] + ['default'],
```
percyliang: avoid absolute indexing

percyliang (changes requested):

codalab_service.py
```diff
@@ -10,9 +10,12 @@
 import errno
 import os
 import subprocess
-import sys
 
-SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker']
+DEFAULT_SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker', 'init']
+
+NO_DEFAULT_SERVICES = ['no-' + service for service in DEFAULT_SERVICES]
```
percyliang: Might as well support this for all services (define `ALL_NO_SERVICES` - also notice the naming change).

codalab_service.py
```diff
@@ -28,12 +31,15 @@ def print_header(description):
...
-    # `default-no-worker` is generally used for real deployment since we don't
-    # want a worker running on the same machine.
-    return (
-        service in args.services
-        or (service != 'test' and 'default' in args.services)
-        or (service != 'test' and service != 'worker' and 'default-no-worker' in args.services)
+    # `default no-worker` is generally used for real deployment since we don't want a worker running on the same machine.
+    default_services = DEFAULT_SERVICES if 'default' in args.services else []
+    # remove "no-" prefix to get the list of services that will be excluded from execution
+    no_default_services = [svc.replace("no-", "") for svc in args.services]
```
percyliang: This doesn't seem right...because you only want to do it for services that contain `no-`.

I think a better way to do this would be:
- Let `services` be `args.services` with `default` expanded into `DEFAULT_SERVICES`.
- Return whether `service in services` and not `negate_service(service) in services`.

codalab_service.py
```diff
@@ -415,7 +453,7 @@ def _get_parser():
             'service',
             metavar='SERVICE',
             type=str,
-            choices=SERVICES,
+            choices=DEFAULT_SERVICES,
```
percyliang: Shouldn't this be `ALL_SERVICES`?  `DEFAULT_SERVICES` might change over time.

codalab_service.py
```diff
@@ -738,6 +785,27 @@ def start_services(self):
                 root=(not self.args.codalab_home),
             )
 
+            print_header('Configuring monitoring service')
+            if all(
```
percyliang: Break this up into 4 if statements...if a subset are specified, just set a subset.

percyliang (changes requested):

codalab_service.py
```diff
@@ -28,13 +31,13 @@ def print_header(description):
...
-    # `default-no-worker` is generally used for real deployment since we don't
-    # want a worker running on the same machine.
-    return (
-        service in args.services
-        or (service != 'test' and 'default' in args.services)
-        or (service != 'test' and service != 'worker' and 'default-no-worker' in args.services)
-    )
+    # `default no-worker` is generally used for real deployment since we don't want a worker running on the same machine.
+    # remove "no-" prefix to get the list of services that will be excluded from execution
+    no_services = [svc.replace("no-", "") for svc in args.services if svc.startswith("no-")]
```
percyliang: This seems unnecessary, since you can just do `'no-' + service in services`, right?

codalab_service.py
```diff
@@ -738,6 +782,34 @@ def start_services(self):
                 root=(not self.args.codalab_home),
             )
 
+            if 'CODALAB_ADMIN_EMAIL' in self.compose_env:
```
percyliang: you could write this more compactly as `for property, env_var in [('server/admin_email', 'CODALAB_ADMIN_EMAIL'), ...]:`...

percyliang (changes requested):

codalab_service.py
candicegjing: But the service in services might not all start with "no-", and we want to find out the ones that start with "no-", right?

percyliang: That's fine, those won't match anything starting with `no-`, so they shouldn't affect anything.

percyliang (approved):
Great!

PR #1334, opened 2019-08-25
cleanup/refactor/simplify deployment

percyliang:
When trying to deploy, realized that there were many inconsistently named variables, many places where defaults were set, environment variables, arguments, different types of deployment, various overrides everywhere, making it hard to reason about.  This PR attempts to organize the configuration inside `codalab_service.py` as much as possible so that there is one source of truth.

A few things to note:
- Make monitor service run as a docker compose service rather than a command (so it runs in the background).
- Rename variables so that we use `*_USERNAME`, `*_PASSWORD` consistently, and try to make it so that there's only one name in the environment variable inside/outside docker.
- Before, the files in a deployment could be temporary in a docker volume or externalized.  Make the latter the only option to make local dev and production more similar, and makes it more transparent where the data is.
- Make services run in user mode so that the files created have the same owner as the user running `./codalab_service.py` as opposed to root (this is not yet fixed for the worker due to needing access to Docker).  This requires modifying the tests, etc. to write to `/tmp`.
- Some more general housecleaning.


codalab_service.py
```diff
@@ -158,269 +241,75 @@ def _get_parser():
             'delete',
             help='Bring down any existing CodaLab service instances (and delete all non-external data!)',
         )
-        restart_cmd = subparsers.add_parser(
```
epicfaace, line 161: Why are we removing the restart command?

codalab_service.py
```diff
@@ -158,269 +241,75 @@ def _get_parser():
...
-                default=argparse.SUPPRESS,
-            )
-            cmd.add_argument(
-                '--mysql-user',
-                type=str,
-                help='MySQL username for the CodaLab MySQL client',
-                default=argparse.SUPPRESS,
-            )
+            ## Arguments for `start`
+            # for cmd in [start_cmd]:
```
epicfaace: Delete this line

codalab_service.py
```diff
@@ -158,269 +241,75 @@ def _get_parser():
...
+            for arg in CODALAB_ARGUMENTS:
+                unnamed = ['--' + arg.name.replace('_', '-')]
```
epicfaace, line 228: Can we rename "unnamed" and "named" to "argument_name" and "argument_kwargs"? Otherwise it's not really clear what they refer to.

codalab_service.py
```diff
@@ -158,269 +241,75 @@ def _get_parser():
...
+                if arg.has_constant_default():
+                    named['default'] = arg.default
+                if arg.type == bool:
+                    named['action'] = 'store_true'
+                else:
+                    named['type'] = arg.type
+                cmd.add_argument(*unnamed, **named)
+
+            # Arguments for `start`
             cmd.add_argument(
```
epicfaace, line 242: Why are these arguments (build, pull, etc.) not defined in `CODALAB_ARGS` like the other args?

codalab_service.py
```diff
@@ -433,195 +322,116 @@ def _get_parser():
...
-            default=argparse.SUPPRESS,
+            default=True,
         )
         logs_cmd.add_argument(
             '--tail',
             '-t',
             type=int,
             help='If specified, tail TAIL lines from the ends of each log',
-            default=argparse.SUPPRESS,
+            default=100,
```
epicfaace, line 303: Why did we have `argparse.SUPPRESS` here in the first place?

codalab_service.py
```diff
@@ -433,195 +322,116 @@ def _get_parser():
...
+                setattr(args, arg.name, os.environ[arg.env_var])
+
+        # Set constant default values
+        for arg in CODALAB_ARGUMENTS:
+            if getattr(args, arg.name, None):  # Skip if set
+                continue
+            if arg.has_constant_default():
+                setattr(args, arg.name, arg.default)
+
+        # Set functional default values (needs to be after everything)
```
epicfaace, line 339: Why do functional default values have a lower priority than constant defaults?

codalab_service.py
```diff
@@ -724,76 +534,76 @@ def run_service_cmd(self, cmd, root=False, service='rest-server'):
         if root:
             uid = '0:0'
         else:
-            uid = self.compose_env['CODALAB_UID']
+            uid = self.args.uid
         self._run_compose_cmd(
-            ('run --no-deps --rm --entrypoint="" --user=%s ' % uid)
+            ('run --no-deps --rm --user=%s ' % uid)
```
epicfaace, line 510: Why did you remove --entrypoint="" ?

codalab_service.py
```diff
@@ -724,76 +534,76 @@ def run_service_cmd(self, cmd, root=False, service='rest-server'):
...
         if should_run_service(self.args, 'init'):
-            print_header('Configuring the service')
-            self.run_service_cmd(
-                "%scl config server/engine_url %s && cl config cli/default_address http://rest-server:2900 && cl config server/rest_host 0.0.0.0"
-                % (cmd_prefix, mysql_url),
-                root=(not self.args.codalab_home),
-            )
+            print_header('Populating config.json')
+            commands = []
+            for config_prop, value in [
```
epicfaace: This code might look more straightforward if you used a list comprehension to assign the value of `commands`.

codalab_service.py
```diff
@@ -724,76 +534,76 @@ def run_service_cmd(self, cmd, root=False, service='rest-server'):
...
             print_header('Creating home and dashboard worksheets')
             self.run_service_cmd(
-                "/opt/wait-for-it.sh rest-server:2900 -- cl logout && cl status && ((cl new home && cl new dashboard) || exit 0)",
-                root=(not self.args.codalab_home),
+                wait_rest_server(
+                    'cl logout && cl status && ((cl new home && cl new dashboard) || exit 0)'
+                )
             )
 
-            for property, env_var in [
```
epicfaace, line 784: Why was this code removed?

docker/compose_files/docker-compose.ssl.yml
```diff
@@ -3,9 +3,8 @@ version: '3.5'
 services:
   nginx:
     ports:
-      - 443:443
+     - ${CODALAB_HTTPS_PORT}:${CODALAB_HTTPS_PORT}
```
epicfaace, line 6: Where is this environment variable set?

docker/compose_files/docker-compose.yml
```diff
@@ -1,91 +1,111 @@
...
     command: cl server
-    environment:
-      *codalab-env
-    networks:
-      - service
-    ports:
-      - ${CODALAB_REST_PORT}:2900
-    logging:
-      *non-blocking-logging
+    <<: *codalab-base
```
epicfaace, line 50: What do the leading * and &'s mean in docker-compose files?

docker/dockerfiles/Dockerfile.server
```diff
@@ -47,5 +48,7 @@ COPY views views
 
 RUN pip install -e .
 
-ENV CODALAB_HOME=/home/codalab
```
epicfaace, line 50: Why did we remove this line?

epicfaace (commented):
Can you re-generate the CLI docs in this PR?

codalab_service.py
percyliang: It wasn't being used.  `start` already does a restart.

codalab_service.py
percyliang: Done.

codalab_service.py
percyliang: Want to avoid 'argument' since in this context, it means something else.  Added some comments to clarify.

codalab_service.py
percyliang: These are not properties of a deployment, but actions you take on the deployment.

codalab_service.py
percyliang: Because we were filling in defaults using a different mechanism, which has been updated now.

codalab_service.py
percyliang: Done.

docker/compose_files/docker-compose.ssl.yml
percyliang: in `codalab_service.py` as a default (search for 443).

codalab_service.py
percyliang: Just refactored.

docker/compose_files/docker-compose.yml
percyliang: It's a yaml thing to allow for modular definitions.

docker/dockerfiles/Dockerfile.server
percyliang: Want to have one place where things are set rather than having defaults in various places spread out.

codalab_service.py
epicfaace: Ok

codalab_service.py
```diff
@@ -802,16 +585,10 @@ def start_services(self):
...
+                wait_rest_server('python test_cli.py --instance {} default'.format(rest_url))
             )
 
-        if should_run_service(self.args, 'monitor'):
-            print_header('Running monitoring service')
-            self.run_service_cmd(
-                "/opt/wait-for-it.sh rest-server:2900 -- python monitor.py",
-                root=(not self.args.codalab_home),
-            )
+        self.bring_up_service('monitor')
```
epicfaace, line 590: It seems like `monitor` is being run by default?

codalab_service.py
```diff
@@ -3,28 +3,55 @@
...
+  variables and pass them through to the actual service (see Docker files for
+  exact logic).
 """
 from __future__ import print_function
 
 import argparse
 import errno
 import os
 import subprocess
+from collections import namedtuple
```
epicfaace: We don't need this line

codalab_service.py
percyliang: no, the check is done inside `bring_up_service`

codalab_service.py
percyliang: removed.

codalab_service.py
epicfaace: Got it, thanks

epicfaace (approved):

Issue #1272, opened 2019-06-30
button to expand bundle details


epicfaace:
- [x] Create sticky header with the title, worksheet metadata (#1337)

- [x] Add the upload/run/text buttons to the header, with their functionality - append to the selected item when clicked. Delete the buttons that currently appear.

- [x] Default - first item should be selected.

- [x] When click a worksheet item, don’t show the details, just select it. Have a separate ‘i’ button to show the details.

PR #1337, opened 2019-08-26
Frontend sticky header, css loading

epicfaace:
- Add sticky header, #1272 

- Also make imports.css and overrides.css be imported directly through the JS code, as opposed to being a static file. It took too long to rebuild the docker container each time I changed the css file, so that's why I ended up doing this in this PR.

Top of the page:

image

Scrolling: 

image


percyliang (approved):
Why do you need to rebuild the docker image when you change css?  With the `-d` option, changes should be hot-loaded live.

Are there any side effects of having Javascript import css (e.g., performance)?

We should revamp all of this sometime to load through JSS.

epicfaace:
With the `-d` option, it reloaded only the js, not the css, though. (Now that I think of it, maybe this was because I didn't clear my cache?)

Letting JavaScript import css will end up putting the styles in an inline `<style>` tag, which could potentially help performance as it's less network requests. A benefit of using it this way is that old css won't be cached. Create-react-app also autoprefixes css selectors for better compatibility. I could imagine applying further optimizations / preprocessing of the CSS (maybe use something like SASS) to make it faster, but that's only possible if the CSS isn't loaded as a static file as it is now. 

I don't know if JSS is the right way to go, as it would require a lot of CSS refactoring, and may not be much better (see https://medium.com/@gajus/stop-using-css-in-javascript-for-web-development-fa32fb873dcc) -- we could continue to import css like this and experiment with further optimizations though (i.e., loading the css in an inline style tag vs separate file) through the JS build configuration.

PR #1341, opened 2019-08-31
Tweaks for deployment

percyliang:
Made a bunch of changes while trying to deploy to `worksheets-dev.codalab.org`, including passing in more config options, fixing a bug with SSL/nginx config, cleanup the bundle store, bug with command-line parsing, and general cleanup.


epicfaace:
Can't review this PR because github can't show the diff (see https://github.com/codalab/codalab-worksheets/compare/master...tweaks-for-deployment -- it says "master and tweaks-for-deployment are entirely different commit histories.")

I tried changing the base branch as suggested [here](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch), but then it just automatically closed the PR (and I can't re-open it)! Maybe it's a github issue; I've asked a question to github support.

PR #1343, opened 2019-09-03
Tweaks for deployment

percyliang:
Trying again...

Made a bunch of changes while trying to deploy to worksheets-dev.codalab.org, including passing in more config options, fixing a bug with SSL/nginx config, cleanup the bundle store, bug with command-line parsing, and general cleanup.


docs/Server-Setup.md
```diff
@@ -133,29 +133,17 @@ You can also start an instance and run tests on it:
...
-    virtualenv -p python3.6 venv3.6
-    venv3.6/bin/pip install black
-    venv3.6/bin/black codalab worker scripts *.py --diff
-
 These must pass before you submit a PR.
 
-## Other
-
-To auto-generate the REST reference and CLI references:
+## Pre-commit
```
epicfaace, line 138: We can (later) integrate this with https://pre-commit.com/

codalab/rest/workers.py
```diff
@@ -27,11 +27,11 @@ def checkin(worker_id):
     socket_id = local.worker_model.worker_checkin(
         request.user.user_id,
         worker_id,
-        request.json["tag"],
-        request.json["cpus"],
-        request.json["gpus"],
-        request.json["memory_bytes"],
-        request.json["free_disk_bytes"],
+        request.json.get("tag"),
```
epicfaace, line 32: Why do we need this? Does request.json sometimes not have the right keys?

docker/compose_files/files/nginx.conf
```diff
@@ -99,19 +98,6 @@ http {
     #proxy_read_timeout 1200;
     #send_timeout 1200;
 
-    location /bundleservice {
```
epicfaace, line 102: Why did you decide to merge /bundleservice with the root url / ? What did we have at / before?

pre-commit.sh
```diff
@@ -0,0 +1,28 @@
...
+  venv2.7/bin/pip install -r requirements-server.txt || exit 1
+  venv2.7/bin/pip install -r requirements.docs.txt || exit 1
+  rm -rf worker/codalabworker.egg-info  # Need to clear because of different Python versions
+  venv2.7/bin/pip install -e worker  # Not sure why this is necessary
+  venv2.7/bin/pip install -e .
+fi
+
+if ! [ -e venv3.6 ]; then
+  virtualenv -p python3.6 venv3.6 || exit 1
+  venv3.6/bin/pip install black || exit 1
```
epicfaace:
```suggestion
  venv3.6/bin/pip install black==18.9b0 || exit 1
```

pre-commit.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# TODO: merge into one once we have Python 3.
+
+if ! [ -e venv2.7 ]; then
+  virtualenv -p python2.7 venv2.7 || exit 1
+  venv2.7/bin/pip install -r requirements-server.txt || exit 1
+  venv2.7/bin/pip install -r requirements.docs.txt || exit 1
+  rm -rf worker/codalabworker.egg-info  # Need to clear because of different Python versions
+  venv2.7/bin/pip install -e worker  # Not sure why this is necessary
+  venv2.7/bin/pip install -e .
```
epicfaace, line 14: Why are we installing codalab etc. before each commit? (it seems like the main purpose of ./pre-commit.sh might be just to fix style and generate docs?)

epicfaace (approved):

docker/compose_files/files/nginx.conf
percyliang: `/bundleservice` no longer exists - this is from the distant past.

codalab/rest/workers.py
percyliang: Old workers might not have all the fields, but we don't want to crash.  Added a comment.

pre-commit.sh
percyliang: Generating docs requires having a running version of CodaLab (since it needs to load all the libraries).

Issue #1333, opened 2019-08-25
Remove SQLite support


epicfaace:
We don't need SQLite, so we should just stick with MySQL support to simplify functionality.

candicegjing:
Closed via #1351 .

PR #1351, opened 2019-09-05
Remove SQLite support

candicegjing:
This PR is to implement features in #1333


percyliang (approved):
LGTM. @epicfaace : did you have anything else in mind?

epicfaace:
Looks good. @candicegjing can you say "fixes #1333" in the PR description next time? That way, it will automatically close the issue once the PR is merged to master.

epicfaace:
I'm not sure if we'd need to change this test too? https://github.com/codalab/codalab-worksheets/blob/e22e2cedaa25a8145c655bdca0720c12468db8e3/tests/model/bundle_model_test.py#L88

candicegjing:
@epicfaace I saw that one as well when I was testing the changes. To change to MySQL, we need a username and password. I am not sure how we want to handle that part in the testing code. Or maybe we can read it from a config file?

percyliang:
Hm, it's probably not worth doing this right now.  I think we can just
remove this test since it's not doing much anyway.

candicegjing:
I double-checked our source code and didn't find anything calling `BundleModelTest`. So I removed the entire class.

epicfaace (approved):

epicfaace:
> I double-checked our source code and didn't find anything calling `BundleModelTest`. So I removed the entire class.

Yeah, that's expected because it's just a test. Thanks!

Issue #1350, opened 2019-09-05
Add .limit to `cl wsearch` and have it default to 10, just like `cl search`


percyliang:

candicegjing:
Closed via #1353 .

PR #1353, opened 2019-09-05
Add .limit=10 to cl wsearch

candicegjing:
This PR implemented features in #1350 .

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -3060,11 +3060,12 @@ class BundleCLI(object):
         'wls',
         aliases=('wsearch', 'ws'),
         help=[
-            'List worksheets on the current instance matching the given keywords.',
+            'List worksheets on the current instance matching the given keywords (returns 10 results by default).',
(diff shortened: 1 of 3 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -3035,6 +3035,7 @@ def display_blocks(self, client, worksheet_info, interpreted_blocks):
             '  wls group=<group_spec>  : List worksheets shared with the group identfied by group_spec.',
             '  wls .mine               : List my worksheets.',
             '  wls .shared             : List worksheets that have been shared with any of the groups I am in.',
+            '  wls .limit=10           : Limit the number of results to the top 10.',
```
percyliang, line 3068: Make it clear in the help message that we're limiting to top 10 results (perhaps on the first line).

percyliang (changes requested):
I don't actually see the logic to do the .limit....

candicegjing:
I think `.limit` function is already implemented in the code. I updated its default value to be 10.

percyliang (approved):

PR #1370, opened 2019-09-07
Duplicate mount points in default config

jyh1:


percyliang:
Bundle mount should not point to the `bundles` directory - it points to the thing that needs to be mounted for bundles to be read.  For example, in our deployment, `$CODALAB_HOME/partitions` links to `/data`, which contains all the mountpoints for various drives.  Then $CODALAB_BUNDLE_MOUNT would be `/data`.

Setting the mount point to home is kind of like a no-op.  (This should probably be more clearly documented.)  Is the current setting causing problems for you?

jyh1:
> Bundle mount should not point to the `bundles` directory - it points to the thing that needs to be mounted for bundles to be read. For example, in our deployment, `$CODALAB_HOME/partitions` links to `/data`, which contains all the mountpoints for various drives. Then $CODALAB_BUNDLE_MOUNT would be `/data`.
> 
> Setting the mount point to home is kind of like a no-op. (This should probably be more clearly documented.) Is the current setting causing problems for you?

The problem is that the current default directory for [codalab_home](https://github.com/codalab/codalab-worksheets/blob/538085f39e302a242d62458f067471ea63343c4e/codalab_service.py#L168-L171) is the same with [bundle_mount](https://github.com/codalab/codalab-worksheets/blob/538085f39e302a242d62458f067471ea63343c4e/codalab_service.py#L176). This causes a duplicate mount error in [docker](https://github.com/codalab/codalab-worksheets/blob/538085f39e302a242d62458f067471ea63343c4e/docker/compose_files/docker-compose.yml#L44-L45).

percyliang:
Can you provide details about command you ran and what error you got?  I'm not getting a duplicate mount error.

jyh1:
I run into the error whenever the default config is used, like in `./codalab_service.py stop`.  The most relevant information seems to be:
>process: Duplicate mount points: [/home/user/codalab-worksheets/var/codalab/home:/home/user/codalab-worksheets/var/codalab/home:rw, /home/user/codalab-worksheets/var/codalab/home:/home/user/codalab-worksheets/var/codalab/home:rw]

Complete error information:

>$ docker --version
> Docker version 18.09.1, build 4c52b90
>$ docker-compose --version
> docker-compose version 1.24.1, build 4667896b

> $ ./codalab_service.py stop
> (cd /home/user/codalab-worksheets/docker/compose_files; CODALAB_VERSION=master CODALAB_INSTANCE_NAME=codalab CODALAB_WORKER_NETWORK_NAME=codalab-worker-network CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CODALAB_MYSQL_HOST=mysql CODALAB_MYSQL_PORT=3306 CODALAB_MYSQL_DATABASE=codalab_bundles CODALAB_MYSQL_USERNAME=codalab CODALAB_MYSQL_PASSWORD=codalab CODALAB_MYSQL_ROOT_PASSWORD=codalab CODALAB_UID=1012:1013 CODALAB_HOME=/home/user/codalab-worksheets/var/codalab/home CODALAB_BUNDLE_MOUNT=/home/user/codalab-worksheets/var/codalab/home CODALAB_MYSQL_MOUNT=/home/user/codalab-worksheets/var/codalab/mysql CODALAB_MONITOR_DIR=/home/user/codalab-worksheets/var/codalab/monitor CODALAB_WORKER_DIR=/home/user/codalab-worksheets/var/codalab/worker CODALAB_HTTP_PORT=80 CODALAB_HTTPS_PORT=443 CODALAB_FRONTEND_PORT=2700 CODALAB_REST_PORT=2900 CODALAB_REST_NUM_PROCESSES=1 CODALAB_USER_DISK_QUOTA=100g CODALAB_USER_TIME_QUOTA=100y CODALAB_USER_PARALLEL_RUN_QUOTA=100 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/user/codalab-worksheets/codalab/bin:/home/user/codalab-worksheets/codalab/bin HOSTNAME=bionlp-DGX-Station docker-compose -p codalab -f docker-compose.yml stop)
> process: The CODALAB_ADMIN_EMAIL variable is not set. Defaulting to a blank string.
> process: The CODALAB_EMAIL_HOST variable is not set. Defaulting to a blank string.
> process: The CODALAB_EMAIL_USERNAME variable is not set. Defaulting to a blank string.
> process: The CODALAB_EMAIL_PASSWORD variable is not set. Defaulting to a blank string.
> process: Duplicate mount points: [/home/user/codalab-worksheets/var/codalab/home:/home/user/codalab-worksheets/var/codalab/home:rw, /home/user/codalab-worksheets/var/codalab/home:/home/user/codalab-worksheets/var/codalab/home:rw]
> Traceback (most recent call last):
>   File "./codalab_service.py", line 644, in <module>
>     main()
>   File "./codalab_service.py", line 90, in main
>     service_manager.execute()
>   File "./codalab_service.py", line 425, in execute
>     self._run_compose_cmd('stop')
>   File "./codalab_service.py", line 510, in _run_compose_cmd
>     raise Exception('Command exited with code {}'.format(popen.returncode))
> Exception: Command exited with code 1

percyliang:
That's strange...I don't get that error (I'm on Docker 19.03.2-beta1).  Does something like `docker run -v /tmp:/tmp -v /tmp:/tmp -ti --rm codalab/default-cpu bash` work?

percyliang:
Hm, I see it happening on Docker 18...okay, maybe we should just change the default bundle mount to something harmless like `/tmp` (just want it to be something that is a no-op).  Unfortunately, I don't think you can do conditionals in docker compose files.

jyh1:
Yes, just changing it to `/tmp` might be better..

codalab_service.py
```diff
@@ -173,7 +173,7 @@ def has_callable_default(self):
     CodalabArg(
         name='bundle_mount',
         help='Path to bundle data (just for mounting into Docker)',
-        default=var_path('bundles'),
+        default=var_path('/tmp'),
```
percyliang: should just be `default='/tmp'`

jyh1: I was not exactly sure here because `var_path` function seems to return a lambda function. Is it OK to just set it to a string?  `var_path` calls `os.path.join` function which will correctly handle appending a absolute path. Maybe `default=lambda _: '/tmp'` for more clearance?

percyliang: `/tmp` is fine (lambda or constant works).

percyliang (approved):
Thanks!

PR #1371, opened 2019-09-07
Fix default cpu docker file

epicfaace:
The dockerfile wasn't able to build because `add-apt-repository ppa:deadsnakes/ppa` needs to run after `apt-get install python-software-properties`.


percyliang (approved):

percyliang:
Hm, wonder why this wasn't a problem before...

epicfaace:
I had accidentally added that line in #1014.

PR #1372, opened 2019-09-07
Merge python 3 changes into frontend

epicfaace:


percyliang (approved):
Is there anything here to note other than just a merge of the Python 3 stuff?

epicfaace:
There was one thing: When merging `master` into `frontend`, I noticed that rest/block_schemas.py (https://github.com/codalab/codalab-worksheets/blob/frontend/codalab/rest/block_schemas.py) was deleted in `master` but modified in `frontend`.

It looks like it was deleted as part of #1354.

However, [this commit](https://github.com/codalab/codalab-worksheets/commit/cb0ff27f87768357c29b9aeae84bff9be70ce60c#diff-1fba856d542aecefc6d48ecdd441aa0a) did modify block_schemas.py on frontend. Are we certain that it's actually unused?

percyliang:
I don't see it used.  The same change is also reflected in
`worksheet_block_schemas.py`, which is the version we should be using.

epicfaace:
Sounds good.

PR #1375, opened 2019-09-09
Remove event table

### LINKED ISSUE #518: reduce amount of logging
Currently, `event_log` is getting pretty big.  Consider cutting back / making this more efficient.

candicegjing:
Fixed #518 .

### DIFF
--- a/codalab/server/rest_server.py
+++ b/codalab/server/rest_server.py
@@ -2,10 +2,8 @@ from http.client import INTERNAL_SERVER_ERROR, BAD_REQUEST
 import datetime
 import json
 import os
-import re
 import sys
 import textwrap
-import time
 import traceback
 import logging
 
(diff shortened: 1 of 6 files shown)


percyliang (approved):

percyliang:
Just make sure tests pass.

codalab/server/rest_server.py
```diff
@@ -27,28 +26,12 @@
 
 from codalab.common import exception_to_http_error
 from codalab.lib import formatting, server_util
-import codalab.rest.account
```
epicfaace, line 31: Perhaps this change is what is causing the tests to fail -- maybe you need these imports in order to add the routes?

candicegjing: Hmm, let me check, somehow my IDE shows they are unused imports. I think you are right. Adding them back makes the build pass.

candicegjing: @epicfaace I think you are right. Adding them back makes the build working. I am just curious how are these imports get called in this code...

percyliang:
Can we put a comment on that line that says don't remove for now?  We
should figure out how to make this so subtle/confusing.

candicegjing:
I have added a comment for those imports. I was also trying to debug why we need those imports given nothing is calling them directly...

epicfaace (approved):
Any more changes needed before we can merge this?

PR #1381, opened 2019-09-10
Replace / with _ for codalab version

### LINKED ISSUE #1369: Travis fails on specially-named branches
For example, dependabot creates branches with slashes in their name. This causes the travis ci build to fail with the following message:

Status: Downloaded newer image for codalab/server:master
invalid argument "codalab/server:dependabot/npm_and_yarn/frontend/bootstrap-4.3.1" for "-t, --tag" flag: invalid reference format

epicfaace:
This fixes #1369, so that branches with "/" in their name do not fail CI.

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -94,7 +94,7 @@ def get_default_version():
     """Get the current git branch."""
     return subprocess.check_output(
         ['git', 'rev-parse', '--abbrev-ref', 'HEAD'], encoding='utf-8'
-    ).strip()
+    ).strip().replace("/", "_")
 
 
 def var_path(name):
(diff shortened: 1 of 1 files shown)


codalab_service.py
```diff
@@ -94,7 +94,7 @@ def get_default_version():
     """Get the current git branch."""
     return subprocess.check_output(
         ['git', 'rev-parse', '--abbrev-ref', 'HEAD'], encoding='utf-8'
-    ).strip()
+    ).strip().replace("/", "_")
```
percyliang: Add comment on why this is necessary.

percyliang (approved):

PR #1388, opened 2019-09-11
Don't decode return value from download_manager._get_read_response_string()

### LINKED ISSUE #1387: unicode decode error when viewing bundle that's running
Result of the Python 3 migration.  To reproduce, just start a running bundle like `while true; do date; sleep 1; done` and select the bundle.  It generates messages like: ```Error on request by pliang(0x5ed2416a59e041aeab72c991ac12020e):

GET /bundles/0x425ee7c0f4af4071a04bba42f1baef49/contents/blob/stderr

Query params:
head='50'
tail='50'
truncation_text='\n... [truncated] ...\n\n'
_='1568171042550'

Form params:

JSON body:
<none>

Local variables:
fileobj=<_io.BufferedReader name=-1>
response_socket_id=967```

epicfaace:
Fixes #1387, in which the `/rest/bundles/{bundleId}/contents/blob/stdout` and `/rest/bundles/{bundleId}/contents/blob/sterr` endpoints were returning 500 errors (I've deployed this branch to dev for now to confirm that the error is gone).

Note that the test I added doesn't actually cover this use case of the response being gzipped; but the test I did add is a good test to have anyway, so I just kept it. Any pointers on how we could write a test that might cover the functionality that broke in #1387?

Some context from my comment in that issue:

> The issue is this line of code: https://github.com/codalab/codalab-worksheets/blob/5b9390b5ecdead2ef736c4e43dc94037f2aeb161/codalab/lib/download_manager.py#L291
> 
> (In python 2, we didn't decode the data).
> 
> From [this link](https://stackoverflow.com/questions/44659851/unicodedecodeerror-utf-8-codec-cant-decode-byte-0x8b-in-position-1-invalid/44660123), it looks like "0x8b" byte is indicative of a gzipped file. We probably shouldn't be decoding the file here. I must have assumed that `_get_read_response_string` returns a string, so it must have `.decode()` in it; that assumption was wrong.

_Originally posted by @epicfaace in https://github.com/codalab/codalab-worksheets/issues/1387#issuecomment-530216143_

### DIFF
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -288,6 +288,7 @@ class DownloadManager(object):
 
     def _get_read_response_string(self, response_socket_id):
         with closing(self._get_read_response_stream(response_socket_id)) as fileobj:
+            # This function returns a binary string, as the file could be gzipped.
             return fileobj.read()
 
 
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -288,7 +288,7 @@ class DownloadManager(object):
 
     def _get_read_response_string(self, response_socket_id):
         with closing(self._get_read_response_stream(response_socket_id)) as fileobj:
-            return fileobj.read().decode()
+            return fileobj.read()
 
 
 class Deallocating(object):
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -165,7 +165,7 @@ class DownloadManager(object):
             try:
                 read_args = {'type': 'read_file_section', 'offset': offset, 'length': length}
                 self._send_read_message(worker, response_socket_id, uuid, path, read_args)
-                string = self._get_read_response_string(response_socket_id)
+                string = self._get_read_response(response_socket_id)
             finally:
                 self._worker_model.deallocate_socket(response_socket_id)
 
@@ -206,7 +206,7 @@ class DownloadManager(object):
                     'truncation_text': truncation_text,
                 }
                 self._send_read_message(worker, response_socket_id, uuid, path, read_args)
-                string = self._get_read_response_string(response_socket_id)
+                string = self._get_read_response(response_socket_id)
             finally:
                 self._worker_model.deallocate_socket(response_socket_id)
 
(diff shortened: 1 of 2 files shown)


codalab/lib/download_manager.py
```diff
@@ -288,7 +288,8 @@ def _get_read_response_stream(self, response_socket_id):
 
     def _get_read_response_string(self, response_socket_id):
```
percyliang: Should we rename this `_get_read_response_bytestring` to avoid confusion?

percyliang (approved):

epicfaace:
@percyliang any idea how to fix this? The new test I added only works with root access, it seems:

```
process: >> b'cl' b'download' b'0x3634b1f997314cd9ac985dcabf4a8c78/stdout'
process: 
process: Traceback (most recent call last):
process:   File "/usr/local/bin/cl", line 11, in <module>
process:     load_entry_point('codalab', 'console_scripts', 'cl')()
process:   File "/opt/codalab-worksheets/codalab/bin/cl.py", line 117, in main
process:     cli.do_command(sys.argv[1:])
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 854, in do_command
process:     structured_result = command_fn()
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 848, in <lambda>
process:     command_fn = lambda: args.function(self, args)
process:   File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 1296, in do_download_command
process:     with open(final_path, 'wb') as out:
process: PermissionError: [Errno 13] Permission denied: '/opt/codalab-worksheets/stdout'
process:  (exit code 1, expected 0 BAD)
```

https://travis-ci.org/codalab/codalab-worksheets/jobs/583480035#L3405

codalab/lib/download_manager.py
epicfaace: I'm not sure what "bytestring" would mean in python 3, so I think I'll just rename it to `_get_read_response` so that the name isn't misleading (calling it `_get_read_response_bytes` sounds superfluous because then we should rename all our functions to have the return type appended to its name).

percyliang:
Use `temp_path()`?

epicfaace:
BTW, changing this broke one of the netcat tests (https://travis-ci.org/codalab/codalab-worksheets/jobs/583813099#L4907) because it was expecting _get_read_response_string to return a string.

percyliang:
Merged into #1389

PR #1389, opened 2019-09-11
Fix unicode

percyliang:
There were more inconsistencies around reading files from the worker.  With the exception of summarize_file, everything returned should be in terms of bytes (since we can have binary files).

The inputs to netcat/write are string's, which is not fully general, but good enough for now.

### DIFF
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -288,6 +288,7 @@ class DownloadManager(object):
 
     def _get_read_response_string(self, response_socket_id):
         with closing(self._get_read_response_stream(response_socket_id)) as fileobj:
+            # This function returns a binary string, as the file could be gzipped.
             return fileobj.read()
 
 
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -224,11 +224,11 @@ class DownloadManager(object):
         )
         try:
             self._send_netcat_message(worker, response_socket_id, uuid, port, message)
-            bytestring = self._get_read_response(response_socket_id)
+            string = self._get_read_response(response_socket_id)
         finally:
             self._worker_model.deallocate_socket(response_socket_id)
 
-        return bytestring
+        return string
 
     def _is_available_locally(self, uuid):
         if self._bundle_model.get_bundle_state(uuid) in [State.RUNNING, State.PREPARING]:
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -224,11 +224,11 @@ class DownloadManager(object):
         )
         try:
             self._send_netcat_message(worker, response_socket_id, uuid, port, message)
-            string = self._get_read_response(response_socket_id)
+            bytestring = self._get_read_response(response_socket_id)
         finally:
             self._worker_model.deallocate_socket(response_socket_id)
 
-        return string
+        return bytestring
 
     def _is_available_locally(self, uuid):
         if self._bundle_model.get_bundle_state(uuid) in [State.RUNNING, State.PREPARING]:
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -288,7 +288,7 @@ class DownloadManager(object):
 
     def _get_read_response_string(self, response_socket_id):
         with closing(self._get_read_response_stream(response_socket_id)) as fileobj:
-            return fileobj.read().decode()
+            return fileobj.read()
 
 
 class Deallocating(object):
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -165,7 +165,7 @@ class DownloadManager(object):
             try:
                 read_args = {'type': 'read_file_section', 'offset': offset, 'length': length}
                 self._send_read_message(worker, response_socket_id, uuid, path, read_args)
-                string = self._get_read_response_string(response_socket_id)
+                string = self._get_read_response(response_socket_id)
             finally:
                 self._worker_model.deallocate_socket(response_socket_id)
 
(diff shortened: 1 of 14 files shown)


worker/codalabworker/file_util.py
```diff
@@ -180,6 +181,7 @@ def summarize_file(file_path, num_head_lines, num_tail_lines, max_line_length, t
     Summarizes the file at the given path, returning a string containing the
     given numbers of lines from beginning and end of the file. If the file needs
     to be truncated, places truncation_text at the truncation point.
+    Unlike other methods, this method treats everything as Unicodes string.
```
epicfaace:
```suggestion
    Unlike other methods, this method treats everything as a Unicode string.
```

worker/codalabworker/worker.py
```diff
@@ -105,9 +105,7 @@ def _checkin(self):
             elif action_type == 'read':
                 self._read(socket_id, response['uuid'], response['path'], response['read_args'])
             elif action_type == 'netcat':
-                self._netcat(
-                    socket_id, response['uuid'], response['port'], response['message'].encode()
-                )
+                self._netcat(socket_id, response['uuid'], response['port'], response['message'])
```
epicfaace, line 108: I think this change might be causing test failures -- perhaps _netcat requires the message to be bytes?

percyliang: Yeah, something's going on.  I think our convention should be that `_netcat` and `_write` take strings (for consistency).

codalab/lib/download_manager.py
```diff
@@ -165,7 +165,7 @@ def read_file_section(self, uuid, path, offset, length, gzipped):
             try:
                 read_args = {'type': 'read_file_section', 'offset': offset, 'length': length}
                 self._send_read_message(worker, response_socket_id, uuid, path, read_args)
-                string = self._get_read_response_string(response_socket_id)
+                string = self._get_read_response(response_socket_id)
```
epicfaace, line 168: Since we renamed `_get_read_response_string` to `_get_read_response` as it always returns bytes, should we rename `string` to `bytestring` instead wherever it is set equal to the result of calling `_get_read_response`?

codalab/rest/bundles.py
```diff
@@ -409,13 +409,11 @@ def _netcat_bundle(uuid, port):
     Send a raw bytestring into the specified port of the running bundle with uuid.
     Return the response from this bundle.
     """
-    check_bundles_have_read_permission(local.model, request.user, [uuid])
+    check_bundles_have_all_permission(local.model, request.user, [uuid])
```
epicfaace, line 412: Why did we change this from requiring just read permission to all permission? I guess the initial code was just an oversight?

codalab/server/rest_server.py
```diff
@@ -206,14 +208,14 @@ def report_exception(self, exc):
         ).format(request, aux_info, traceback.format_exc())
 
         # Both print to console and send email
-        print(message, file=sys.stderr)
+        logger.error(message)
```
epicfaace, line 211: Why use `logger.error` instead of printing to stderr?

tests/files/netcat-test.py
```diff
@@ -1,30 +1,27 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
epicfaace, line 1: Should this be python3.6? Or does it not matter?

test_cli.py
```diff
@@ -116,9 +116,9 @@ def sanitize(string, max_chars=256):
 def run_command(
     args, expected_exit_code=0, max_output_chars=256, env=None, include_stderr=False, binary=False
 ):
-    print(
-        ">>", *[a.encode('ascii', errors='replace') if type(a) is str else a for a in args], sep=" "
-    )
+    """If we don't care about the exit code, set `expected_exit_code` to None.
+    """
+    print(">>", *map(str, args), sep=" ")
```
epicfaace, line 121: The initial `.encode('ascii')` was there because Travis would behave strangely when trying to print non-ascii characters to console. Not sure if that's still an issue, though. After changing this code, did you notice any changes in the travis logs?

worker/codalabworker/rest_client.py
```diff
@@ -49,29 +49,38 @@ def _make_request(
...
-        if data is not None and isinstance(data, dict):
+        if isinstance(data, dict):
             headers['Content-Type'] = 'application/json'
-            data = json.dumps(data)
-        headers['X-Requested-With'] = 'XMLHttpRequest'
-        if query_params is not None:
-            path = path + '?' + urllib.parse.urlencode(query_params)
+            data = json.dumps(data)  # Turn dict into string
+        if isinstance(data, str):
+            data = data.encode()  # Turn string into bytes
```
epicfaace, line 70: Is it fine that `data` can be either string or bytes here?

epicfaace (approved):

codalab/lib/download_manager.py
percyliang: I initially did that, but there are many places where `string` should be `bytestring`.  I think that can be another PR.

codalab/rest/bundles.py
percyliang: You really want something in between where you might allow other people to connect to your service, but don't want people to delete/edit your bundle.  But this is just safer.

codalab/server/rest_server.py
percyliang: For some reason, stderr wasn't being printed.  Also, trying to standardize to logger everywhere.

tests/files/netcat-test.py
percyliang: Shouldn't matter.

test_cli.py
percyliang: It didn't complain.  Looking at the logs, there are some weird characters printed out, but that's true of master right now too.

worker/codalabworker/rest_client.py
percyliang: It definitely works...maybe a bit confusing to accept both types, but that's what we have right now.

PR #1391, opened 2019-09-14
don't crash when summarizing binary files

percyliang:


epicfaace (commented):
Which summarize file function is affected by this? I'm wondering if we can write a unit test for it (or make a note to do so later).

percyliang:
This is from download_manager.summarize_file...when you try to view an uploaded binary bundle in the web interface.  Yeah, we don't have a framework for testing this, since it's not called by the CLI.  We should add something though.

worker/codalabworker/file_util.py
```diff
@@ -198,9 +200,12 @@ def ensure_ends_with_newline(lines, remove_line_without_newline=False):
...
                 # To ensure that the last line is a whole line, we remove the
                 # last line if it doesn't have a newline character.
-                head_lines = fileobj.read(num_head_lines * max_line_length).splitlines(True)[
-                    :num_head_lines
-                ]
+                try:
+                    head_lines = fileobj.read(num_head_lines * max_line_length).splitlines(True)[
+                        :num_head_lines
+                    ]
+                except UnicodeDecodeError:
```
epicfaace, line 207: I'm guessing that this exception is triggered whenever fileobj is binary. Why is that the case? I don't see anything specific for expecting string output in the line above.

test_cli.py
```diff
@@ -794,6 +798,14 @@ def test(ctx):
     check_contains('::home-', run_command([cl, 'ls', '-w', '/']))
 
 
+@TestModule.register('binary')
+def test(ctx):
+    # Upload a binary file and test it
+    uuid = run_command([cl, 'upload', '/bin/ls'])
+    run_command([cl, 'cat', uuid], binary=True)
```
epicfaace: Can we test the output of the cat command to make sure it returns what we expect?

codalab/lib/formatting.py
```diff
@@ -23,16 +22,7 @@ def contents_str(input_string, verbose=False):
...
-    # Unicode is always printable
-    if isinstance(input_string, str):
-        return input_string
-
-    # Assume string is UTF-8 or else it contains arbitrary binary data that should not be rendered
-    try:
-        return input_string.decode('utf-8')
-    except UnicodeDecodeError:
-        return BINARY_PLACEHOLDER if verbose else ''
+    return str(input_string)
```
epicfaace, line 25: Why did we return BINARY_PLACEHOLDER for non-utf8-binary files before? Was this the previous behavior in Python 2, or something that got introduced in the Python 3 transition?

epicfaace (approved):
Looks good!

worker/codalabworker/file_util.py
percyliang: because the file is opened as a text file

test_cli.py
percyliang: Done

codalab/lib/formatting.py
percyliang: Before, this function could take bytes, but now we assume it can't.

PR #1393, opened 2019-09-15
gzip should support taking bytes, not string; rename things

percyliang:
Don't assume input to gzip is a string.  Should be bytestring.  Also, rename from `string` to `bytestring` in a lot of places to avoid confusion.

### DIFF
--- a/worker/codalabworker/local_run/local_reader.py
+++ b/worker/codalabworker/local_run/local_reader.py
@@ -124,7 +124,7 @@ class LocalReader(Reader):
                     args['num_tail_lines'],
                     args['max_line_length'],
                     args['truncation_text'],
-                )
+                ).encode()
             )
             reply_fn(None, {}, bytestring)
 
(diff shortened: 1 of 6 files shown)


epicfaace (approved):

worker/codalabworker/local_run/local_reader.py
```diff
@@ -102,7 +102,9 @@ def read_file_section(self, run_state, path, dep_paths, args, reply_fn):
         """
 
         def read_file_section_thread(final_path):
-            bytestring = gzip_string(read_file_section(final_path, args['offset'], args['length']))
+            bytestring = gzip_bytestring(
+                read_file_section(final_path, args['offset'], args['length'])
```
epicfaace, line 106: I'm assuming this was the piece of code that was failing before this PR, correct?

percyliang: Yeah.

PR #1399, opened 2019-09-17
Fix account activation and password reset emails over https

### LINKED ISSUE #1386: password resent link from email should be https if SSL enabled instead of http

### LINKED ISSUE #1397: [USER ISSUE] Account activation links are not working
The account activation links sent by e-mail during the sign up are not working as reported in [this issue](https://github.com/codalab/codalab-competitions/issues/2695).

epicfaace:
Fixes the password reset and email verification emails when codalab runs over ssl by doing the following:
- Using the right scheme in email verification and password reset emails (fixes #1386)
- Re-adding the http -> https redirect in the nginx config (fixes #1397)


percyliang:
Actually we should fix this by the nginx redirect, because some instances
of codalab are http.

epicfaace:
Makes sense, although ideally, we would also want the email to link directly to https:// if the site is on https://, not http:// redirecting to https://.

percyliang:
I guess we should pass in the appropriate URL.

candicegjing:
LGTM!

epicfaace:
Seems like the way Django detects whether the site is http or https is to check for a header set by our proxy such as "HTTP_X_FORWARDED_PROTO" -- see https://docs.djangoproject.com/en/2.2/ref/settings/#secure-proxy-ssl-header (this header is used in [HTTPRequest.is_secure()](https://docs.djangoproject.com/en/2.0/_modules/django/http/request/#HttpRequest.is_secure))

percyliang:
In our stack, we have the CODALAB_USE_SSL environment variable.

itachiCheng:
Great! Thank you!

epicfaace:
> In our stack, we have the CODALAB_USE_SSL environment variable.

I've changed it to use the X-Forwarded-Proto header because it's parallel to X-Forwarded-Host, and it seems to be set properly in the nginx config: https://github.com/codalab/codalab-worksheets/blob/71189d6b34792bdc497f8c44da497b515cfdf409/docker/compose_files/files/nginx.conf.ssl#L113

docker/compose_files/files/nginx.conf.ssl
```diff
@@ -68,6 +68,14 @@ http {
     server rest-server:2900;
   }
 
+  server {
```
percyliang, line 72: Write a comment that says this redirects all http traffic to https.

percyliang (approved):
Nice!

PR #1402, opened 2019-09-18
Remove codalabworker package

### LINKED ISSUE #1346: Get rid of codalabworker package and just have one codalab package
This will simplify the currently convoluted dependencies between codalab and codalabworker.

candicegjing:
Fixed #1346 .
I re-created this PR as the previous one was end up in *partial-merge-pull* state.


docker/dockerfiles/Dockerfile.worker
```diff
@@ -6,22 +6,26 @@ ENV DEBIAN_FRONTEND noninteractive
...
-# Install dependencies again
-RUN python3.6 -m pip install --user --upgrade pip; \
-    python3.6 -m pip install -e /opt/worker;
+RUN python3.6 -m pip install -e .
 
 # Allow non-root to read everything
-RUN chmod -R og=u-w /opt/worker
+RUN chmod -R og=u-w .
+
+EXPOSE 2900
```
percyliang: Why is this here?  this is from the rest server

percyliang (approved):

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -186,7 +186,7 @@ def _cleanup(self):
             with self._global_lock:
                 self._acquire_all_locks()
                 bytes_used = sum(dep.size_bytes for dep in self._dependencies.values())
-                serialized_length = len(codalabworker.pyjson.dumps(self._dependencies))
+                serialized_length = len(codalab.pyjson.dumps(self._dependencies))
```
epicfaace: Back to my comment on the original PR: why do we have codalab.pyjson instead of codalabworker.pyjson?

candicegjing: This is different than the original PR. I merged something wrong in this PR. It should be `codalab.worker`

docker/dockerfiles/Dockerfile.worker
candicegjing: It shouldn't be here. I was testing something else and left it here. I removed it just now.

PR #1403, opened 2019-09-18
Set autofocus on username when user enters into login page

### LINKED ISSUE #1348: when click sign in, autofocus should be on username field

candicegjing:
Fixed #1348 .

### DIFF
--- a/frontend/src/components/Login.js
+++ b/frontend/src/components/Login.js
@@ -56,7 +56,7 @@ class Login extends React.Component {
                                 name='username'
                                 placeholder='Username or e-mail'
                                 type='text'
-                                autoFocus=''
+                                autofocus="true"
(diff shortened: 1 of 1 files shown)


frontend/src/components/Login.js
```diff
@@ -56,7 +56,7 @@ class Login extends React.Component {
                                 name='username'
                                 placeholder='Username or e-mail'
                                 type='text'
-                                autoFocus=''
+                                autofocus="true"
```
percyliang: Prefer `autofocus={true}`.  Also, shouldn't the react property be `autoFocus` (the HTML property is `autofocus` I think)...

candicegjing: I think that's right. I have updated `autofocus` to be `autoFocus`

PR #1407, opened 2019-09-20
fix netcat

percyliang:
- Make netcat and netcurl work properly again (fix some Unicode issues and actually interpreting the HTTP response headers).
- If worker has a problem, don't fail the job, just hang the worker instead (will get worker_offline).
- Some random cleanups from previous PRs.


codalab/rest/bundles.py
```diff
@@ -409,7 +410,7 @@ def _netcat_bundle(uuid, port):
     Send a raw bytestring into the specified port of the running bundle with uuid.
     Return the response from this bundle.
     """
-    check_bundles_have_all_permission(local.model, request.user, [uuid])
+    check_bundles_have_read_permission(local.model, request.user, [uuid])
```
percyliang, line 418: Change it back so that one can launch a service that others can hit (for demos).  In the future, we want to explicitly expose ports.

codalab/rest/bundles.py
```diff
@@ -449,23 +450,39 @@ def _netcurl_bundle(uuid, port, path=''):
...
+        new_response.begin()
+        # Copy the headers over
+        for k in new_response.headers:
+            response.headers[k] = new_response.headers[k]
+        # Return the response (which sends back the body)
+        return new_response
+
     except Exception:
-        print("{}".format(request.environ), file=sys.stderr)
+        logger.error("{}".format(request.environ), file=sys.stderr)
```
percyliang, line 486: Trying to use logger everywhere for consistency.

codalab/worker/local_run/local_run_state.py
```diff
@@ -254,9 +265,11 @@ def _transition_from_PREPARING(self, run_state):
                 runtime=self.docker_runtime,
             )
             self.worker_docker_network.connect(container)
-        except docker_utils.DockerException as e:
-            run_state.info['failure_message'] = 'Cannot start Docker container: {}'.format(e)
-            return run_state._replace(stage=LocalRunStage.CLEANING_UP, info=run_state.info)
+        except Exception as e:
+            message = 'Cannot start Docker container: {}'.format(e)
```
percyliang, line 269: Don't fail the bundle because it's probably the worker's fault.

codalab/worker/worker.py
```diff
@@ -58,16 +58,18 @@ def start(self):
...
                     self._stop = True
                     break
 
             except Exception:
                 self._last_checkin_successful = False
                 traceback.print_exc()
-                time.sleep(1)
+                # Sleep for a long time so we don't keep on failing.
+                logger.error('Sleeping for 1 hour due to exception...please help me!')
+                time.sleep(1 * 60 * 60)
```
percyliang, line 72: Important to not keep on looping or else we create a billion docker containers and eventually bring down the worker machine.

docker/dockerfiles/Dockerfile.server
```diff
@@ -19,9 +20,6 @@ RUN apt-get update; apt-get install -y \
     python3-software-properties \
     zip;
  
-### Without this Python thinks we're ASCII and unicode chars fail
```
percyliang, line 22: Don't think we need this since we're on Python 3 now.

docker/dockerfiles/Dockerfile.server
```diff
@@ -7,6 +7,7 @@ RUN apt-get update && apt-get install -y software-properties-common && add-apt-r
 
 RUN apt-get update; apt-get install -y \
     build-essential \
+    curl \
```
percyliang, line 10: Needed for the netcurl test.

codalab/rest/bundles.py
epicfaace: Could you add a comment / TODO for this?

codalab/worker/local_run/local_run_state.py
epicfaace: Can you add this as a comment?

Also, it looks like it is failing the bundle, since there is a `raise`?

docker/dockerfiles/Dockerfile.server
epicfaace: I remember needing this on Python 3, because otherwise non-ascii command line arguments would not be properly decoded. We could try removing this in a separate PR though.

codalab/rest/bundles.py
percyliang: Done

codalab/worker/local_run/local_run_state.py
percyliang: A general comment is at the top of the class.

docker/dockerfiles/Dockerfile.server
percyliang: Added this back anyway, because it fails travis tests.

epicfaace (approved):
Looks good. Were these issues specifically caused by the Python 3 migration, or were they there even before that?

percyliang:
The bytes/str stuff was caused by Python 3, but the netcurl stuff seems unrelated and was from before.

PR #1408, opened 2019-09-21
Worker manager

percyliang:
Introduces a worker manager, which is a lightweight process that just essentially launches workers when there are staged jobs.  By keeping the process stateless and decoupling the worker system from the choice of backend (AWS, Azure, etc.), we can hope to get a general auto-scaling system that just works.

Currently, only the AWS Batch is supported.  There are still some rough edges and TODOs.  Once this is ironed out, we can do Azure Batch and Slurm.


codalab/worker/main.py
```diff
@@ -1,8 +1,3 @@
-#!/usr/bin/env python3.6
```
epicfaace: Why remove this line?

codalab_service.py
```diff
@@ -213,6 +219,30 @@ def has_callable_default(self):
     CodalabArg(name='use_ssl', help='Use HTTPS instead of HTTP', type=bool, default=False),
     CodalabArg(name='ssl_cert_file', help='Path to the cert file for SSL'),
     CodalabArg(name='ssl_key_file', help='Path to the key file for SSL'),
+    ### Worker manager
+    CodalabArg(name='worker_manager_type', help='Type of worker manager (e.g., aws, azure, slurm)'),
```
epicfaace: Can you add a note in the "help" saying that only aws is supported?

codalab/worker_manager/main.py
```diff
@@ -0,0 +1,63 @@
...
+    args = parser.parse_args()
+
+    # Set up logging.
+    if args.verbose:
+        logging.basicConfig(format='%(asctime)s %(message)s', level=logging.DEBUG)
+    else:
+        logging.basicConfig(format='%(asctime)s %(message)s', level=logging.INFO)
+
+    # Choose the worker manager type.
+    if args.worker_manager_type == 'aws':
```
epicfaace: Perhaps it would be better to rename this to "aws_batch". In case we want to support other compute environments like aws lambda...

codalab/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class WorkerManager(object):
+    """
+    The abstract class for a worker manager.  Different backends like AWS,
+    Azure, Slurm should override `get_worker_jobs` and `start_worker_job`.
+
+    The basic architecture of the WorkerManager is extremely simple: to a
+    first-order approximation, it simply launches `cl-worker`s as AWS/Azure
+    batch jobs as long as there staged bundles.
```
epicfaace:
```suggestion
    batch jobs as long as there are staged bundles.
```

codalab/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,167 @@
...
+
+    The simplicity means that we don't need to manage state about how workers
```
epicfaace, line 24: Architecturally, I'm wondering whether it would be better to have the worker manager handle this complexity / state instead of `cl-worker`. I'm not sure about the benefits or drawbacks of either approach, though -- do you have any thoughts?

codalab_service.py
```diff
@@ -213,6 +219,30 @@ def has_callable_default(self):
...
+        default='codalab-cpu',
+    ),
+    CodalabArg(
+        name='worker_manager_gpu_queue',
+        help='Name of queue to submit GPU jobs',
+        default='codalab-gpu',
+    ),
+    CodalabArg(
+        name='worker_manager_max_cpu_workers',
+        help='Maximum number of CPU workers',
```
epicfaace: This is the maximum number of CPU workers _per_ worker manager (not overall), right?

codalab_service.py
```diff
@@ -177,7 +179,7 @@ def has_callable_default(self):
     CodalabArg(
         name='bundle_mount',
         help='Path to bundle data (just for mounting into Docker)',
-        default='/tmp',
+        default=var_path(''),  # Put any non-empty path here
```
epicfaace, line 182: Why did this change? Did this have to do with making worker managers work?

codalab/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,167 @@
...
+    and `max_workers`.  It will also monitor the staged bundles that satisfy a
+    certain `search` criterion.  If there are staged bundles then it will issue
+    a `start_worker_job()` call, provided some other conditions are met (e.g.,
+    don't start workers too fast).
+
+    The WorkerManager is all client-side code, so it can be customized as one
+    sees fit.
+
+    Notes:
+    - The worker manager is not visible via CodaLab.  One needs to monitor the
```
epicfaace: What specifically do you mean by "not visible via CodaLab"? One can still view the docker containers when running the system.

codalab/worker/main.py
```diff
@@ -94,6 +89,12 @@ def main():
         action='store_true',
         help='If specified the worker quits if it finds itself with no jobs after a checkin',
     )
+    parser.add_argument(
+        '--idle-seconds',
+        help='Not running anything for this many seconds constitutes idle',
+        type=int,
+        default=0,
```
epicfaace, line 101: Is it ok to have a default of 0 for this? Or should we increase the default?

codalab/worker/worker.py
```diff
@@ -40,12 +41,15 @@ def __init__(
         self._work_dir = work_dir
         self._bundle_service = bundle_service
         self._exit_when_idle = exit_when_idle
+        self._idle_seconds = idle_seconds
         self._stop = False
         self._last_checkin_successful = False
         self._run_manager = create_run_manager(self)
 
     def start(self):
+        """Return whether ran anything."""
```
epicfaace:
```suggestion
        """Return whether we ran anything."""
```

epicfaace (commented):
Looks good! I think it would be good to also add some tests for `cl-worker-manager` (probably mocking/stubbing the AWS batch dependencies)

codalab/worker/main.py
percyliang: We never should execute this script directly

codalab_service.py
percyliang: Done

codalab/worker_manager/main.py
percyliang: Good point

codalab/worker_manager/worker_manager.py
percyliang: We have tried putting things into `cl-worker` and it has always turned into a nightmare.

Currently the implementation is pretty hands off (doesn't associate workers with jobs), but we could also attach particular bundle uuids to workers when we launch them, so I think we're not really losing on generality.

codalab_service.py
percyliang: Clarified.

codalab_service.py
percyliang: It's kind of unrelated...sharing a /tmp directory just pollutes the host's tmp, so it was a bad choice.

codalab/worker_manager/worker_manager.py
percyliang: What I mean is that you can't see from CodaLab that the worker manager is trying to do something (e.g., trying to launch jobs) or even whether it died.

codalab/worker/main.py
percyliang: This gets overridden by worker manager.  0 makes it backward compatible, and sometimes you want to just run one job with a worker manually and have it quit.

codalab_service.py
bkgoksel: Probably best to use a `choices` list here, I think that way all available choices get printed in the generated help text.

codalab/worker/file_util.py
```diff
@@ -193,7 +194,11 @@ def ensure_ends_with_newline(lines, remove_line_without_newline=False):
             else:
                 lines[-1] += '\n'
 
-    file_size = os.stat(file_path).st_size
```
bkgoksel, line 196: It looks like not handling exceptions here would cause issues but do we know how/when/if this was causing issues?

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -0,0 +1,113 @@
...
+            'Workers: {}'.format(
+                ' '.join(job['jobId'] + ':' + job['status'] for job in jobs) or '(none)'
+            )
+        )
+        return [WorkerJob() for job in jobs]
+
+    def start_worker_job(self):
+        image = 'codalab/worker:' + os.environ.get('CODALAB_VERSION', 'latest')
+        logger.debug('Starting worker with image {}'.format(image))
+        job_definition_name = 'codalab-worker-4'  # This is just an arbitrary identifier.
```
bkgoksel, line 36: Can we use a less magic magic number for the job definiton name? Or is this also something we'd read from a config?

bkgoksel (approved):
LGTM. From attempting a similar thing for Slurm in the past, I think it is a good idea to not try to store the state of worker <-> bundle attachment.

codalab/worker/file_util.py
percyliang: When things were misconfigured, that file didn't exist, which crashed everything.  Just makes it more robust.

codalab/worker_manager/aws_batch_worker_manager.py
percyliang: This whole part needs to be cleaned up a bit, but literally this is pretty arbitrary.  Could also put CODALAB_VERSION here.

codalab/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,169 @@
...
+            logger.info(
+                'Want to launch a worker because we are under the minimum ({} < {})'.format(
+                    len(worker_jobs), self.args.min_workers
+                )
+            )
+            want_more_workers = True
+
+        # 2) There is a staged bundle AND we have made progress running the
+        # previously staged bundles.
+        # To expand on this, if we just started a worker and we haven't made any
```
bkgoksel: There seems to be some room for error here. On the test deployment on SM, the gpu-worker-manager is stuck in a loop of no progress made while bundles are waiting staged. 

I think one plausible scenario for this is:
1. Worker manager is started, there are neither staged bundles nor worker jobs
2. Worker manager launches 1 worker job (to reach `min_workers`)
3. `wait_for_progress` is set to `True` since a new worker was launched
4. A run bundle is submitted and assigned to this newly created worker
5. Worker manager starts its new iteration, but isn't aware that its recently created worker picked up a bundle, because that bundle was never in its `staged_uuids` list
6. Another run bundle is submitted to Codalab and starts waiting in the `staged` state.
7. The worker manager refuses to launch another worker job despite the staged bundle because it doesn't think any progress was made on its first worker.
8. The second run bundle has to wait for the first run bundle to finish before it can be run.

The problem here is that a bundle might get submitted and immediately assigned to a new worker before the worker-manager can register it as a staged bundle first and then realize it's no longer staged (ie from the worker manager's perspective, its latest worker is doing nothing, and the staged bundles cannot be assigned to it for some mysterious reason).

Right now I'm trying to wrap my head around whether we can get to such a degenerate state some other way, but this is definitely one way the current system can get stuck.

percyliang:
Hm...maybe the wait_for_progress is not great.  The simplest thing would be
to get rid of it, and pay the cost of accidentally launching more workers
than we need (but should be mitigated by the minimum time between
workers).  That would simplify the logic, which is a good thing.

bkgoksel:
> Hm...maybe the wait_for_progress is not great. The simplest thing would be to get rid of it, and pay the cost of accidentally launching more workers than we need (but should be mitigated by the minimum time between workers). That would simplify the logic, which is a good thing.

Yeah, and it seems like `wait_for_progress` can block spawning workers even when there are 0 workers right now. I'll remove the mechanisms and redeploy and see how it goes.

codalab/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,167 @@
...
+        self.staged_uuids = new_staged_uuids
+        logger.info(
+            'Staged bundles [{}]: {}'.format(
+                ' '.join(keywords), ' '.join(self.staged_uuids) or '(none)'
+            )
+        )
+
+        # Get worker jobs
+        worker_jobs = self.get_worker_jobs()
+        pending_jobs, active_jobs = [], []
```
percyliang: Can we rename to `pending_worker_jobs` and `active_worker_jobs`?  `worker` is more important than `job` here, because `job` could mean bundle.

codalab/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,167 @@
...
+            if seconds_since_last_worker < self.args.min_seconds_between_workers:
+                logger.info(
+                    'Don\'t launch becaused waited {} < {} seconds since last worker'.format(
+                        seconds_since_last_worker, self.args.min_seconds_between_workers
+                    )
+                )
+                return
+
+            if len(pending_jobs) >= len(self.staged_uuids):
+                # Make sure we don't queue up more workers than staged UUIDs if there are
```
percyliang: To match the style around, move comment before the `if`.

PR #1409, opened 2019-09-21
Flush username prompt when authenticating

nelson-liu:
On my machine, when I run `cl work`, I see:

```
$ cl work

```

Note that the "username" prompt that usually appears is not there.

After hitting ctrl-c, it shows up.

```
$ cl work
^CUsername: Terminated by Ctrl-C
```

(I can still authenticate by typing my username into the void)

Seems like its reading the username properly, the text just isn't being immediately flushed because of `sys.stdout.write`'s buffering.


percyliang (approved):
Thanks!

PR #1415, opened 2019-09-23
Add cl-competiond as an entry_point to setup.py

### LINKED ISSUE #1286: add `cl-competitiond` endpoint to `setup.py`
Instead of running `scripts/competitiond.py`, install `cl-competitiond` to make it more consistent with how we invoke other endpoints.

Then we should update the docs.

candicegjing:
Fixed #1286

### DIFF
--- a/docs/Competitions.md
+++ b/docs/Competitions.md
@@ -34,21 +34,15 @@ We provide a small application built on the [CodaLab Worksheets API](REST-API-Re
 
 ## Installation and setup
 
-Checkout the CodaLab Worksheets repository (in the future, this should be available via pip):
+Install codalab package and its dependencies:
 
-    git clone https://github.com/codalab/codalab-worksheets
-
-Install all the dependencies:
-
-    cd codalab-worksheets
-    virtualenv -p python3 venv
-    venv/bin/pip install -e .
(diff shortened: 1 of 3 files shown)


docs/Competitions.md
```diff
@@ -42,13 +42,13 @@ Install all the dependencies:
 
     cd codalab-worksheets
     virtualenv -p python3.6 venv
```
percyliang: Don't need to git clone or make a venv anymore.  The whole point is that you can now just `pip install codalab` and type `cl-competitiond ...`

percyliang (changes requested):

docs/Competitions.md
percyliang: But should check that this works.

candicegjing: The only thing is that users may not have all the package requirements installed in their local machine.

percyliang: `pip install codalab` should install the dependencies.

candicegjing: That's true. I installed it in local mode when I was testing the code change. It didn't install the dependencies. I'll update the code.

percyliang (approved):

PR #1416, opened 2019-09-23
Specify dependencies in stopping/deleting docker services

### LINKED ISSUE #1373: when codalab_service.py delete, stop monitor script first to avoid error messages
In general, add `depends_on` in the docker compose file so services are stopped in the right order.  Don't have to worry about starting, since `codalab_service.py` handles that manually.

candicegjing:
Fixed #1373

### DIFF
--- a/docker/compose_files/docker-compose.yml
+++ b/docker/compose_files/docker-compose.yml
@@ -53,24 +53,33 @@ services:
     command: cl server
     <<: *codalab-base
     <<: *codalab-server
+    depends_on:
+      - mysql
 
   bundle-manager:
     image: codalab/server:${CODALAB_VERSION}
     command: cl bundle-manager
     <<: *codalab-base
     <<: *codalab-server
+    depends_on:
+      - mysql
 
   frontend:
     image: codalab/frontend:${CODALAB_VERSION}
     command: serve -s build -l ${CODALAB_FRONTEND_PORT}
     <<: *codalab-base
     <<: *codalab-root
+    depends_on:
+      - rest-server
 
   nginx:
     image: nginx:1.12.0
     command: nginx
     <<: *codalab-base
     <<: *codalab-root
+    depends_on:
+      - frontend
+      - rest-server
     ports:
       - ${CODALAB_HTTP_PORT}:80
     volumes:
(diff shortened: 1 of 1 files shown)


candicegjing:
I didn't see any error in my local machine without specifying all those depends_on items. However, based on docker documentation, specifying depends_on will make sure the order of execution.

docker/compose_files/docker-compose.yml
```diff
@@ -106,6 +106,12 @@ services:
     <<: *codalab-base
     depends_on:
       - rest-server
+      - bundle-manager
+      - frontend
```
percyliang: Doesn't depend on frontend

docker/compose_files/docker-compose.yml
```diff
@@ -106,6 +106,12 @@ services:
     <<: *codalab-base
     depends_on:
       - rest-server
+      - bundle-manager
```
percyliang: Doesn't depend on bundle-manager

docker/compose_files/docker-compose.yml
```diff
@@ -106,6 +106,12 @@ services:
...
+      - frontend
+      - nginx
+      - mysql
+      - worker
```
percyliang: Doesn't depend on worker

docker/compose_files/docker-compose.yml
```diff
@@ -106,6 +106,12 @@ services:
     <<: *codalab-base
     depends_on:
       - rest-server
+      - bundle-manager
+      - frontend
+      - nginx
+      - mysql
```
percyliang: mysql is technically a dependency, but mysql should be a dependency of rest-server and bundle-manager

percyliang (changes requested):
Might as well and go through each service and write down exactly what services it directly depends on.

docker/compose_files/docker-compose.yml
candicegjing: I think I misunderstood the meaning of *stopping monitor service first*. I have updated the code and split all the dependencies for each service.

percyliang (approved):

PR #1421, opened 2019-09-24
Frontend - fix crashing when selecting rows

epicfaace:
The issue was that the condition used to say `this._numTableRows(item.props.item) != null`, but I changed it to `this._numTableRows(item.props.item) !== null` to fix lint issues.

However, this meant that when `this._numTableRows(item.props.item)` is equal to undefined, it would end up evaluating the if statement when it wasn't actually supposed to.


percyliang (approved):
What was this returning before?

epicfaace:
> What was this returning before?

You mean `this._numTableRows(item.props.item)`? It was returning `undefined` in the cases when the crash was happening.

PR #1429, opened 2019-09-26
Upgrade database revision to the latest before any other operations d…

### LINKED ISSUE #1345: codalab_service.py: migrate first
Currently, deployment doesn't work when we require a migration because we try to call `create-root-user.py` then do the alembic migrate, but then we're not using the right database schema for the former.

Instead we need to do:
- If database exists, alembic upgrade
- create-root-user
- alembic stamp

candicegjing:
Fixed #1345

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -586,6 +586,12 @@ class CodalabServiceManager(object):
             ]
             self.run_service_cmd(' && '.join(commands))
 
+            print_header('Initializing the database with alembic')
+            # We need to upgrade the current database revision to the most recent revision before any other database operations.
+            self.run_service_cmd(
+                'if [ $(alembic current | wc -l) -gt 0 ]; then echo upgrade; alembic upgrade head; fi'
+            )
+
             print_header('Creating root user')
             self.run_service_cmd(
                 wait_mysql(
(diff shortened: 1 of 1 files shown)


codalab_service.py
```diff
@@ -586,18 +586,22 @@ def wait_rest_server(cmd):
...
                     'python3 scripts/create-root-user.py {}'.format(self.args.codalab_password)
                 )
             )
 
-            print_header('Initializing/migrating the database with alembic')
-            # The first time, we need to stamp; after that upgrade.
-            self.run_service_cmd(
-                'if [ $(alembic current | wc -l) -gt 0 ]; then echo upgrade; alembic upgrade head; else echo stamp; alembic stamp head; fi'
-            )
+            print_header('Stamping the database with alembic')
```
percyliang, line 602: This is only necessary when the `upgrade` condition is false, so maybe we want to add the if statement in there.

percyliang (approved):
Please double check that this works fine both with a new install and an actual migration (just manually roll back the DB via alembic and then run this script).

epicfaace (commented):
Is it ok to change the order to `upgrade` first and then `stamp` (initially it was `stamp` and then `upgrade`)? Does the order matter?

percyliang:
Order is crucial.  `stamp` then `upgrade` will not even perform migrations because you already stamp the current DB as the head.

percyliang (approved):

PR #1434, opened 2019-09-27
Add web-cli back to frontend

### LINKED ISSUE #1427: add web CLI back

epicfaace:
Fixes #1427.

Looks like this now:

image

### DIFF
--- a/frontend/src/components/worksheets/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet.js
@@ -791,7 +791,7 @@ class Worksheet extends React.Component {
         var editPermission = info && info.edit_permission;
         var canEdit = this.canEdit() && this.state.editMode;
 
-        var searchClassName = !this.state.showActionBar ? 'search-hidden' : '';
+        var searchClassName = 'search-hidden';
         var editableClassName = canEdit ? 'editable' : '';
         var viewClass = !canEdit && !this.state.editMode ? 'active' : '';
         var rawClass = this.state.editMode ? 'active' : '';
--- a/frontend/src/components/worksheets/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet.js
@@ -11,6 +11,7 @@ import 'bootstrap';
 import 'jquery-ui-bundle';
 import WorksheetHeader from './WorksheetHeader';
 import { NAVBAR_HEIGHT } from '../../constants';
+import WorksheetActionBar from './WorksheetActionBar';
 
 /*
 Information about the current worksheet and its items.
@@ -108,6 +109,7 @@ class Worksheet extends React.Component {
             activeComponent: 'list', // Where the focus is (action, list, or side_panel)
             editMode: false, // Whether we're editing the worksheet
             editorEnabled: false, // Whether the editor is actually showing (sometimes lags behind editMode)
+            showActionBar: true, // Whether the action bar is shown
             focusIndex: -1, // Which worksheet items to be on (-1 is none)
             subFocusIndex: 0, // For tables, which row in the table
             numOfBundles: -1, // Number of bundles in this worksheet (-1 is just the initial value)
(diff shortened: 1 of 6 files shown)


percyliang (changes requested):
Great!  The only problem I noticed is that if you scroll down to the end of the worksheet, then the web terminal is not visible.

frontend/src/components/worksheets/Worksheet.js
```diff
@@ -318,6 +336,22 @@ class Worksheet extends React.Component {
             }.bind(this),
         );
 
+        // Show/hide web terminal (action bar)	
+        Mousetrap.bind(
```
Levilian, line 340: Can you double check if this command actually works? Not on my mac keyboard.

frontend/src/components/worksheets/WorksheetActionBar.js
```diff
@@ -0,0 +1,254 @@
...
+            },	
+            {	
+                greetings:	
+                    "Click here to enter commands (e.g., help, run '<bash command>', rm <bundle>, kill <bundle>, etc.).",	
+                name: 'command_line',	
+                height: ACTIONBAR_MINIMIZE_HEIGHT,	
+                prompt: 'CodaLab> ',	
+                history: true,	
+                keydown: function(event, terminal) {	
+                    if (event.keyCode === 67 && event.ctrlKey) {
```
Levilian, line 71: I think `event.ctrlKey` is the direct cause of the error I pointed out here https://github.com/codalab/codalab-worksheets/blob/f664347652249842147f3ceb2671785dfa7101f4/frontend/src/components/worksheets/Worksheet.js#L318-L324

Levilian (changes requested):
Beyond my individual comments, there are two other things that you might consider (or merge this and file another PR)

1) fix the position of the terminal on top. In this case when you scroll down the terminal is still visible. This is the expected behavior in the old front-end so we'd better keep it the same. 
2) according to the code review guidelines, "Notes: we are still using CSS and jquery due to unfortunate legacy issues, but the plan is to get rid of this.", you might want to make sure no new changes uses jQuery.

frontend/src/components/worksheets/Worksheet.js
epicfaace: I just added in the code that was there initially before the web terminal was removed ¯\_(ツ)_/¯ -- will take a look though, thanks

percyliang (approved):

frontend/src/components/worksheets/WorksheetActionBar.js
epicfaace: What do you mean?

frontend/src/components/worksheets/Worksheet.js
epicfaace: You were right -- forgot to add in the functionality for showing/hiding the action bar. I've added it now -- it turned out to be a bit more complicated than originally implemented because of the <Sticky> component on the header. Please take a look!

frontend/src/components/worksheets/WorksheetActionBar.js
Levilian: Shouldn't it be `shift` not `ctrl` key?

epicfaace: I think this piece of code doesn't have to do with the shift+c keyboard shortcut to show/hide action bar.

epicfaace: i.e. I think it refers to copying with ctrl + c

Levilian: Got it!

Levilian (approved):

epicfaace:
> 2\. you might want to make sure no new changes uses jQuery.

To this point, the reason why this has jQuery is because I'm adding in the code that was previously there. In this case, the least destructive change would be to add in the old jQuery-using code, then later do a refactoring in a separate PR (along with the other jQuery refactorings needed).

Thanks!

PR #1437, opened 2019-09-27
Name bundle containers by the UUIDs

bkgoksel:
This hopefully helps avoid name clashes and will help us better debug
the issue that sometimes happens on the workers. Also good for general
transparency

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -141,9 +141,11 @@ def start_bundle_container(
         # nvidia-docker runtime uses this env variable to allocate GPUs
         environment['NVIDIA_VISIBLE_DEVICES'] = ','.join(gpuset) if gpuset else 'all'
 
+    # Name the container same as the UUID to avoid network name clashes
     container = client.containers.run(
         image=docker_image,
         command=docker_command,
+        name=uuid,
         network=network,
         mem_limit=memory_bytes,
         cpuset_cpus=cpuset_str,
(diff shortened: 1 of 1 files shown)


percyliang (approved):
The name clash only happened once in a while...do you know what the root problem there was?

bkgoksel:
Ok diving a bit deeper, I think I found the underlying cause of that issue, which is unrelated to this PR, but I think this is still a good thing to have for future debugging etc.

epicfaace:
What were bundle containers named previously?

bkgoksel:
> What were bundle containers named previously?

Unnamed, so they would get random names that docker assigns by default to containers (things like `suspicious_dijkstra` or `homely_falcon`. I'm guessing there aren't actually collisions because I found another explanation for the worker problem, so this is more of an aesthetic choice now (whether we want UUID named containers or more human-readable but randomly assigned names)

bkgoksel:
I'm leaving this up to you guys to merge or not.

codalab/worker/docker_utils.py
```diff
@@ -141,9 +141,11 @@ def start_bundle_container(
         # nvidia-docker runtime uses this env variable to allocate GPUs
         environment['NVIDIA_VISIBLE_DEVICES'] = ','.join(gpuset) if gpuset else 'all'
 
+    # Name the container same as the UUID to avoid network name clashes
     container = client.containers.run(
         image=docker_image,
         command=docker_command,
+        name=uuid,
```
percyliang: maybe prefix this with `run-`?

epicfaace: Do we want to prefix this with `codalab` something?

percyliang: `codalab-run-`?

epicfaace: I think we should use `codalab_run_` because we use underscores to name other containers (`codalab_frontend_1`, etc.)

percyliang: sounds good

bkgoksel: Done

PR #1438, opened 2019-09-27
Change name of general worker docker network

bkgoksel:
Previously the name of the general worker docker network that contains
all the containers was a prefix of the names of the specific internal
and external networks that determine the network connection status.

This caused problems when workers were started on machines with these
networks already existing. The Docker SDK's network search method
returns networks for which the search term is a substring (and in
apparent random order from call to call), so it was
possible for us to pick one of the external or internal networks when
trying to reuse the general network.

In that situation we would try to connect the container to the same
network twice, leading to the annoying endpoint already exists errors.
With this PR, no network name should be a substring of another, so even
when we reuse networks, we should not pick the wrong network.


percyliang (approved):
Ah!

bkgoksel:
Looks like the worker went offline after the `resources` test run, which is a bit strange. Could anyone else try to debug this (ie maybe run the resources test locally on this branch and see what's going on)?

bkgoksel:
Ok so somehow this breaks `netcat`. I'm investigating right now.

bkgoksel:
Had to make changes to the `docker-compose` file because the worker container itself was still being added to a Docker network with just the prefix as the name.

Right now the `_general` suffix is hardcoded to match in the compose and worker code files. This is a hacky solution but I don't have bandwidth to investigate a more thorough solution at the moment.

percyliang (approved):
Nice!

PR #1439, opened 2019-09-30
Adding logic to prevent pushing images to dockerhub when user credent…

### LINKED ISSUE #1378: don't try to push to Docker Hub if don't have credentials
If someone not on the CodaLab team submits a PR, then tests fail because travis tries to push to docker hub.  If `$CODALAB_DOCKER_USERNAME` is not set, then we shouldn't try to do this.

There are several places where we could add this logic...maybe `travis-deploy.sh`?

candicegjing:
Fixed #1378

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,9 @@ jobs:
         - sudo mv docker-compose /usr/local/bin
         - sudo service mysql stop
       install:
-        - python3.6 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull --push
+        - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
       script:
-        - python3.6 codalab_service.py start --services default monitor test --version ${TRAVIS_BRANCH}
+        - python3 codalab_service.py start --services default monitor test --version ${TRAVIS_BRANCH}
     - stage: deploy
       script: echo "Deploying"
       language: python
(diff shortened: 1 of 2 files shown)


percyliang (changes requested):
Also need to change the `--push` in `.travis.yml`.

percyliang (commented):
We need that `--push` in there for us, otherwise builds are really slow...

scripts/travis-deploy.sh
```diff
@@ -9,8 +9,16 @@
 
 tag=$1
 
-python3 codalab_service.py build --version $tag --pull --push
+# Check if ENV variable CODALAB_DOCKER_USERNAME is set. If not, it means this build was triggered
+# by an external user. Then we shouldn't push the docker image to docker hub.
+if [[ -z "${CODALAB_DOCKER_USERNAME}" ]]; then
```
epicfaace: It might look better to do something like
`PUSH_FLAG=$([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")`

to be consistent with the code in `.travis.yml`.

epicfaace (changes requested):

.travis.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         - sudo mv docker-compose /usr/local/bin
         - sudo service mysql stop
       install:
-        - python3.6 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull --push
+        - python3.6 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
```
percyliang: Do we need this to be python3.6?  We were trying to replace everything with python3.

percyliang (approved):

Issue #1342, opened 2019-09-01
[USER ISSUE] size parameter on `display image` no longer works


robinjia:
**Describe the bug**
The `size=XYZ` parameter no longer does anything for a `display image` row item.

**Codalab Information**
- *Your username*: robinjia
- *UUID of worksheet where problem occurred*: 0xe3eb416773ed4883bb737662b31b4948
 - *Browser (if on Web)*; Chrome (also confirmed on Firefox)

**To Reproduce**
Steps to reproduce the behavior :
1. Go to https://worksheets.codalab.org/worksheets/0xe3eb416773ed4883bb737662b31b4948

**Expected behavior**
The image is supposed to be 400 pixels wide, but it is not. I'm very confident that when I made that worksheet originally, I made it sized properly, so it's not like I typed a directive wrong.

**Screenshots**
![image](https://user-images.githubusercontent.com/7267349/64072293-d846c280-cc40-11e9-9e32-0473d48e0515.png)

yipenghe:
Will be fixed by #1441, thanks!

PR #1441, opened 2019-10-01
Fix size parameters for display image

yipenghe:
Fixing #1342 When passing a dictionary for the styles, the character ; is not needed like in hard coding strings. Changing the height and width parameters now work
display-200-200-cmd
display-200-200-image
display-400-400-cmd
display-400-400-image

### DIFF
--- a/frontend/src/components/worksheets/items/ImageItem.js
+++ b/frontend/src/components/worksheets/items/ImageItem.js
@@ -22,12 +22,11 @@ class ImageItem extends React.Component {
         var src = 'data:image/png;base64,' + this.props.item.image_data;
         var styles = {};
         if (this.props.item.hasOwnProperty('height')) {
-            styles['height'] = this.props.item.height + 'px;';
+            styles['height'] = this.props.item.height + 'px';
         }
         if (this.props.item.hasOwnProperty('width')) {
-            styles['width'] = this.props.item.width + 'px;';
+            styles['width'] = this.props.item.width + 'px';
         }
-
         return (
             <div className='ws-item' onClick={this.handleClick}>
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/items/ImageItem.js
```diff
@@ -22,12 +22,12 @@ class ImageItem extends React.Component {
...
         if (this.props.item.hasOwnProperty('height')) {
-            styles['height'] = this.props.item.height + 'px;';
+            styles['height'] = this.props.item.height + 'px';
         }
         if (this.props.item.hasOwnProperty('width')) {
-            styles['width'] = this.props.item.width + 'px;';
+            styles['width'] = this.props.item.width + 'px';
         }
-
+
```
epicfaace, line 30: Can you remove this space?
```suggestion

```

epicfaace (approved):
Nice catch!

frontend/src/components/worksheets/items/ImageItem.js
yipenghe: The space was originally here actually. Accidentally deleted it so added it back, it separates the code block & return block so I think we should keep it

epicfaace: It wasn't originally there. Only the newline was there, but the tab wasn't. I meant to remove the tab, but keep the newline.

yipenghe: Oh I see, I will fix this in a new pull request, sorry about that

yipenghe: Removing it #1448 , will be more careful next time

Issue #1349, opened 2019-09-05
empty worksheet title cannot be edited


percyliang:

Levilian:
http://g.recordit.co/idB5z20hoU.gif
Which part of the editing is the issue referenced here?

percyliang:
Ah, this is a problem with `master`, not the new `frontend`.  This is a bad
bug we should address ASAP.

Levilian:
#1443

PR #1443, opened 2019-10-01
Set empty worksheet title name to "untitled"

Levilian:
Now the default name of worksheet title is "untitled" and therefore an empty name is editable.

### DIFF
--- a/frontend/src/components/EditableField.js
+++ b/frontend/src/components/EditableField.js
@@ -103,7 +103,7 @@ export class EditableField extends React.Component<{
         if (!this.state.editing) {
             return (
                 <a className='editable editable-click' onClick={this.onClick}>
-                    {this.state.value === '' ? 'untitled' : this.state.value}
+                    {this.state.value === '' ? '(empty)' : this.state.value}
                 </a>
             );
         } else {
--- a/frontend/src/components/EditableField.js
+++ b/frontend/src/components/EditableField.js
@@ -103,7 +103,7 @@ export class EditableField extends React.Component<{
         if (!this.state.editing) {
             return (
                 <a className='editable editable-click' onClick={this.onClick}>
-                    {this.state.value === '' ? '<none>' : this.state.value}
+                    {this.state.value === '' ? 'untitled' : this.state.value}
                 </a>
             );
         } else {
--- a/frontend/src/components/EditableField.js
+++ b/frontend/src/components/EditableField.js
@@ -103,7 +103,7 @@ export class EditableField extends React.Component<{
         if (!this.state.editing) {
             return (
                 <a className='editable editable-click' onClick={this.onClick}>
-                    {this.state.value === '' ? '(empty)' : this.state.value}
+                    {this.state.value === '' ? '<none>' : this.state.value}
                 </a>
             );
         } else {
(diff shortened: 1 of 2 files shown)


Levilian:
Fixes #1349 . 
Now it looks like 
Screen Shot 2019-10-01 at 12 34 33 AM

Screen Shot 2019-10-01 at 12 34 48 AM

frontend/src/components/EditableField.js
```diff
@@ -103,7 +103,7 @@ export class EditableField extends React.Component<{
         if (!this.state.editing) {
             return (
                 <a className='editable editable-click' onClick={this.onClick}>
-                    {this.state.value === '' ? '<none>' : this.state.value}
+                    {this.state.value === '' ? 'untitled' : this.state.value}
```
percyliang: 'untitled' doesn't quite make sense for general fields...(oh, that's a problem too?)
Maybe '(empty)'?
I'd put parentheses to make it look not like the actual value.

frontend/src/components/worksheets/Worksheet.js
```diff
@@ -872,7 +872,7 @@ class Worksheet extends React.Component {
                                                             key={'title' + this.canEdit()}
                                                             canEdit={this.canEdit()}
                                                             fieldName='title'
-                                                            value={info && info.title}
+                                                            value={info && info.title || "untitled"}
```
percyliang: Let's do '(untitled)'.

Nit: we tend to use single quotes everywhere so try to be consistent (even though the codebase isn't quite consistent yet).

percyliang (changes requested):

frontend/src/components/worksheets/Worksheet.js
epicfaace: I think "Untitled Worksheet" might look better, what do you think? (based off making "Untitled Document" in Microsoft Word)

epicfaace: I think "Untitled Worksheet" might look better, what do you think? (based off making "Untitled Document" in Microsoft Word)

frontend/src/components/EditableField.js
epicfaace: Whatever we choose, we need to be consistent. We use `<none>` for editable fields too (https://github.com/codalab/codalab-worksheets/blob/5b9390b5ecdead2ef736c4e43dc94037f2aeb161/frontend/src/components/EditableField.js#L105-L107) and for empty files (https://github.com/codalab/codalab-worksheets/blob/ce7a845567055f0f5ac262fe1498c4ac5cdc79be/codalab/lib/formatting.py#L10-L25) and we don't use "(empty)" anywhere else in the codebase.

I would suggest keeping this to `<none>` and then perhaps changing the default placeholder in a separate PR if we want to do that?

epicfaace (commented):
Looks good, just added some replies.

frontend/src/components/worksheets/Worksheet.js
Levilian: The only issue is that our new front-end has a different set of naming systems with `untitled` and `unnamed`. I think for consistency we should make sure there's a unified naming system for both the old and new version (i.e., backward compatibility)? @percyliang what do you think?

percyliang: I don't feel strongly about this..."Untitled Worksheet" seems fine to me.  I guess the difference is that the worksheet title is actually not "Untitled Worksheet"...

epicfaace: Oh, I see, since the worksheet title is actually not "Untitled Worksheet", then it makes sense to have a non-capitalized placeholder like "untitled" / "(untitled)".

percyliang (approved):

PR #1445, opened 2019-10-01
create cl-server and cl-bundle-manager entry_points

### LINKED ISSUE #1380: split off `cl server` and `cl bundle-manager` into `cl-server` and `cl-bundle-manager`
Currently, these are all living in `cl.py` for no good reason.

candicegjing:
Fixed #1380

### DIFF
--- a/codalab/bin/bundle_manager.py
+++ b/codalab/bin/bundle_manager.py
@@ -1,0 +1,27 @@
+# Main entry point for CodaLab cl-bundle-manager.
+import signal
+import argparse
+from codalab.lib.codalab_manager import CodaLabManager
+from codalab.server.bundle_manager import BundleManager
+
+
+def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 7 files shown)


epicfaace (commented):
Looks good. Could you add some quick tests in https://github.com/codalab/codalab-worksheets/blob/2f9dfdd7f0546469cb1f4dd83d358015422642ec/test_cli.py#L1255-L1263 that just run "cl-server" and "cl-bundle-manager" and make sure no errors pop up?

candicegjing:
@percyliang Where do we want to add documentation for cl-server and cl-bundle-manager? They don't quite belong to the CLI reference page based on the current context, as it mentioned `cl <command> <arguments>` as its usage.

percyliang (approved):
Great.  I think we can leave out the documentation since users don't need to run it, and if you're running a server, you'd use codalab_service.py and not invoke these command directly.

Issue #1425, opened 2019-09-25
[Frontend] Clicking outside a worksheet row should deselect it.


epicfaace:
Steps:
1. Click on a worksheet row. It gets highlighted in blue.
2. Click somewhere else, like the margins.

One would expect that the row gets unhighlighted, so it is in the same state as it was when the worksheet first loaded. However, this does not happen -- the row keeps being blue.

This issue needs to be fixed because in order for someone to easily add a row to the end of the worksheet (after the changes in #1424 are merged), they must be able to click out of the worksheet to deselect the current row.

PR #1447, opened 2019-10-01
Deselect when clicking outside worksheet row items

yipenghe:
[This is in new frontend]
Implements the deselection mentioned in #1425 
event.stopPropagation() prevents the click event on child items (any worksheet row items) to be propagated to parent components (inner, outer, desktop etc.) which keeps the focus functionality. 

A demo (zipped mp4, github can't just play it):
[deselect-demo.zip](https://github.com/codalab/codalab-worksheets/files/3678909/deselect-demo.zip)

To do:
Should we allow to double click a worksheet row item to deselect as well?

### DIFF
--- a/frontend/src/components/worksheets/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet.js
@@ -140,7 +140,10 @@ class Worksheet extends React.Component {
     }
 
     handleClickForDeselect = (event) => {
-        this.setFocus(-1, 0);
+        //Deselecting when clicking outside worksheet_items component
+        if (event.target === event.currentTarget){
+            this.setFocus(-1, 0);
+        }
     };
 
     setFocus = (index, subIndex, shouldScroll) => {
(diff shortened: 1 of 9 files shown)


epicfaace (changes requested):
Thanks!

Can you not use `stopPropagation()` though? It can have unintended side effects / bugs; see https://css-tricks.com/dangers-stopping-event-propagation/.

A safer way to prevent clicks on a child from propagating to its parent is to compare `event.currentTarget` (the child element) and `event.target` (the parent element). See this for an example implementation of it (the author of this commit initially used `stopPropagation()` but then changed it to this): https://github.com/rjsf-team/react-jsonschema-form/pull/1336/files#diff-07a63ede6fe55fca89398d42da737369R223

yipenghe:
> Thanks!
> 
> Can you not use `stopPropagation()` though? It can have unintended side effects / bugs; see https://css-tricks.com/dangers-stopping-event-propagation/.
> 
> A safer way to prevent clicks on a child from propagating to its parent is to compare `event.currentTarget` (the child element) and `event.target` (the parent element). See this for an example implementation of it (the author of this commit initially used `stopPropagation()` but then changed it to this): https://github.com/rjsf-team/react-jsonschema-form/pull/1336/files#diff-07a63ede6fe55fca89398d42da737369R223

Thanks, updated to use the suggested way, new recorded demo
[deselect-demo.zip](https://github.com/codalab/codalab-worksheets/files/3679493/deselect-demo.zip)

epicfaace (approved):
Looks good!

PR #1451, opened 2019-10-02
doc: add macOS instructions

### LINKED ISSUE #1450: Add macOS server setup doc

Levilian:
Fix: #1450


docs/Server-Setup.md
```diff
@@ -83,6 +83,17 @@ Start the CodaLab service as follows:
 
     ./codalab_service.py start -bd
 
+If your default python version is not python 3.6+, then you might have to use:
+
+    python3 codalab_service.py start -bd
```
percyliang: Don't need this since `python3.6` is specified in `codalab_service.py` already.

docs/Server-Setup.md
```diff
@@ -83,6 +83,17 @@ Start the CodaLab service as follows:
...
+
+For macOS, you might come across an error with `gunicorn` when running the above command.
```
percyliang: Maybe this should be put in a section at the end called Troubleshooting so the main instructions don't get too long.

percyliang (changes requested):

docs/Server-Setup.md
Levilian: I still receive the following error if I just run ./codalab_service.py
`bash: ./codalab_service.py: /usr/bin/python3: bad interpreter: No such file or directory`

percyliang: replace the first line with `#!/usr/bin/env python3`

Levilian: Is it because I used anaconda instead of pip to install my python3?

Levilian: Yes I originally changed the line to `#!/usr/bin/env python3` but then I don't know if I should update the master branch with this change. I think `#!/usr/bin/env python3` is preferred in shebang as it doesn't assume that it's running on a Ubuntu system. Any reason why we chose to use `#!/usr/bin/python3`?

percyliang: It should be `#!/usr/bin/env python3`

Levilian: Ok I will create a separate PR and fix that!

docs/Server-Setup.md
```diff
@@ -6,8 +6,8 @@ CodaLab features.
 # Requirements
 
 All you need is a version of Docker and docker-compose compatible with docker-compose file version 3.5.
-Note that while Docker/docker-compose should be cross-platform, we have only tested Ubuntu.
-If you try to use CodaLab on MacOS or Windows and run into issues, feel free to let us know, but those platforms
+Note that while Docker/docker-compose should be cross-platform, we have only tested Ubuntu and MacOS.
```
epicfaace, line 9: macOS

Levilian: Nice catch

PR #1452, opened 2019-10-02
Remove the unused waiting_for_worker_startup state

bkgoksel:
This is a bundle state that exists in the codebase, but is not ever
used. In fact the method that sets the budle to be in that state is not
even called once in our codebase, so we can safely remove it for a bit
of simplification. This is part of my effort to break up and extract the
good parts of #983


bkgoksel:
Closed as it is subsumed by #1453

PR #1453, opened 2019-10-02
Refactor bundle model

bkgoksel:
* Unify the naming of server-side Bundle state transition methods.
* Reorganize method ordering in `bundle_model.py` logically
* Fix pylint nits in `bundle_model.py`:
  * Turning `if <cond>: return <val> elif <cond2>: return <val2>` kind of statements into `if <cond>: return <val> if <cond2>: return <val2>`
  * Getting rid of unused variables
  * Getting rid of `if True` and the ilk
  * renaming local variable called `type` to `item_type` as `type` is a protected word now

* Unify docstring comment delimiters as `"""` since that seems more
common in our codebase.
* Remove the unused `waiting_for_worker_startup` Bundle state references
from our codebase. There was no transition into this state anyways as
things were before. 

Part of my effort to break up and extract good bits of #983

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1770,11 +1770,7 @@ class BundleModel(object):
         """
         Helper function to serialize DataTime
         """
-        return (
-            obj.isoformat()
-            if isinstance(obj, datetime.datetime) or isinstance(obj, datetime.date)
-            else None
-        )
+        return obj.isoformat() if isinstance(obj, (datetime.date, datetime.datetime)) else None
 
     def add_chat_log_info(self, query_info):
         """
@@ -1810,7 +1806,7 @@ class BundleModel(object):
         """
         user_id1 = query_info.get('user_id')
         if user_id1 is None:
-            return
+            return None
         limit = query_info.get('limit')
         with self.engine.begin() as connection:
             query = select(
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1770,7 +1770,11 @@ class BundleModel(object):
         """
         Helper function to serialize DataTime
         """
-        return obj.isoformat() if isinstance(obj, (datetime.date, datetime.datetime)) else None
+        return (
+            obj.isoformat()
+            if isinstance(obj, datetime.datetime) or isinstance(obj, datetime.date)
+            else None
+        )
 
     def add_chat_log_info(self, query_info):
         """
@@ -1806,7 +1810,7 @@ class BundleModel(object):
         """
         user_id1 = query_info.get('user_id')
         if user_id1 is None:
-            return None
+            return
         limit = query_info.get('limit')
         with self.engine.begin() as connection:
             query = select(
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -70,6 +70,10 @@ class BundleModel(object):
         self.public_group_uuid = ''
         self.create_tables()
 
+    # ==========================================================================
+    # Database helper methods
+    # ==========================================================================
+
     # TODO: Remove these methods below when all appropriate table columns have
     # been converted to the appropriate types that perform automatic encoding.
     # (See tables.py for more details.)
@@ -96,10 +100,10 @@ class BundleModel(object):
         self._create_default_groups()
         self._create_default_clients()
 
-    def do_multirow_insert(self, connection, table, values):
+    @staticmethod
+    def do_multirow_insert(connection, table, values):
         """
         Insert multiple rows into the given table.
-
         This method may be overridden by models that use more powerful SQL dialects.
         """
         # This is a lowest-common-denominator implementation of a multi-row insert.
@@ -111,15 +115,15 @@ class BundleModel(object):
             with connection.begin():
                 connection.execute(table.insert(), values)
 
-    def make_clause(self, key, value):
+    @staticmethod
+    def make_clause(key, value):
         if isinstance(value, (list, set, tuple)):
             if not value:
                 return False
             return key.in_(value)
-        elif isinstance(value, LikeQuery):
+        if isinstance(value, LikeQuery):
             return key.like(value)
-        else:
-            return key == value
+        return key == value
 
     def make_kwargs_clause(self, table, kwargs):
         """
@@ -132,6 +136,29 @@ class BundleModel(object):
             clauses.append(self.make_clause(getattr(table.c, key), value))
         return and_(*clauses)
 
+    @staticmethod
+    def _render_query(query):
+        """
+        Return string representing SQL query.
+        """
+        query = query.compile()
+        s = str(query)
+        for k, v in query.params.items():
+            s = s.replace(':' + k, str(v))
+        return s
+
+    def _execute_query(self, query):
+        """
+        Execute the given query and return the first matching row
+        """
+        with self.engine.begin() as connection:
+            rows = connection.execute(query).fetchall()
+        return [row[0] for row in rows]
+
+    # ==========================================================================
+    # Bundle info accessor methods
+    # ==========================================================================
+
     def get_bundle(self, uuid):
         """
         Retrieve a bundle from the database given its uuid.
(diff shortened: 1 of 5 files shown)


codalab/model/bundle_model.py
```diff
@@ -96,10 +100,10 @@ def create_tables(self):
         self._create_default_groups()
         self._create_default_clients()
 
-    def do_multirow_insert(self, connection, table, values):
+    @staticmethod
```
bkgoksel, line 103: Turned all the little helper methods here that didn't use the self reference into static methods here to make it clearer.

codalab/model/bundle_model.py
```diff
@@ -132,6 +136,29 @@ def make_kwargs_clause(self, table, kwargs):
             clauses.append(self.make_clause(getattr(table.c, key), value))
         return and_(*clauses)
 
+    @staticmethod
```
bkgoksel, line 139: These were just moved from elsewhere, this is not new code.

codalab/model/bundle_model.py
```diff
@@ -588,19 +614,6 @@ def get_bundle_uuids(self, conditions, max_results):
 
         return self._execute_query(query)
 
-    # Helper function: return string representing SQL query.
```
bkgoksel, line 591: These just moved up, not removed.

codalab/model/bundle_model.py
```diff
@@ -643,30 +656,15 @@ def batch_get_bundles(self, **kwargs):
         ]
         return bundles
 
-    def set_waiting_for_worker_startup_bundle(self, bundle, job_handle):
```
bkgoksel, line 646: This got renamed, had its comment updated and moved through the file. No changes to business logic.

codalab/model/bundle_model.py
```diff
@@ -643,30 +656,15 @@ def batch_get_bundles(self, **kwargs):
...
-            ).fetchone()
-            if not row:
-                # The user deleted the bundle.
-                return
+    # ==========================================================================
+    # Server-side bundle state machine methods
+    # ==========================================================================
 
-            bundle_update = {
-                'state': State.WAITING_FOR_WORKER_STARTUP,
```
bkgoksel, line 661: Git diff is confused, this is just part of the earlier function.

codalab/model/bundle_model.py
```diff
@@ -643,30 +656,15 @@ def batch_get_bundles(self, **kwargs):
...
-                'metadata': {'job_handle': job_handle, 'last_updated': int(time.time())},
-            }
-            self.update_bundle(bundle, bundle_update, connection)
-
-    def set_starting_bundle(self, bundle, user_id, worker_id):
+    def transition_bundle_starting(self, bundle, user_id, worker_id):
```
bkgoksel, line 663: Renamed and comment updated.

codalab/model/bundle_model.py
```diff
@@ -688,40 +686,11 @@ def set_starting_bundle(self, bundle, user_id, worker_id):
 
             return True
 
-    def set_offline_bundle(self, bundle):
```
bkgoksel, line 691: Similarly this moved, got renamed, and had its comment updated. No logic changes.

codalab/model/bundle_model.py
```diff
@@ -765,41 +735,11 @@ def start_bundle(self, bundle, user_id, worker_id, start_time, remote):
 
         return True
 
-    def bundle_checkin(self, bundle, bundle_update, user_id, worker_id):
```
bkgoksel, line 768: This just got moved, no other changes.

codalab/model/bundle_model.py
```diff
@@ -830,19 +770,45 @@ def resume_bundle(self, bundle, bundle_update, row, user_id, worker_id, connecti
...
+        return True
+
+    def transition_bundle_finalizing(self, bundle, user_id, exitcode, failure_message, connection):
+        """
+        Transitions bundle to FINALIZING state:
+            Saves the failure message and exit code from the worker
+            If the user running the bundle was the CodaLab root user,
+            increments the time used by the bundle owner.
         """
-        state = State.FAILED if failure_message or exitcode else State.READY
```
bkgoksel, line 840: This `state` var wasn't used in this function since this is for finalizing, the state is actually computed in the `finish_bundle` (now `transition_bundle_finished` function)

codalab/model/bundle_model.py
```diff
@@ -879,6 +846,43 @@ def finish_bundle(self, bundle):
                 cl_worker_run.delete().where(cl_worker_run.c.run_uuid == bundle.uuid)
             )
 
+    # ==========================================================================
```
bkgoksel, line 849: This just moved from above, no new code here.

codalab/model/bundle_model.py
```diff
@@ -688,40 +686,11 @@ def set_starting_bundle(self, bundle, user_id, worker_id):
...
-            bundle_update = {
-                'state': State.WORKER_OFFLINE,
-                'metadata': {'last_updated': int(time.time())},
-            }
-            self.update_bundle(bundle, bundle_update, connection)
-        return True
-
-    def restage_bundle(self, bundle):
+    def transition_bundle_staged(self, bundle):
         """
```
bkgoksel, line 690: Renamed and comment updated

codalab/model/bundle_model.py
```diff
@@ -743,11 +712,12 @@ def restage_bundle(self, bundle):
 
             return True
 
-    def start_bundle(self, bundle, user_id, worker_id, start_time, remote):
+    def transition_bundle_preparing(self, bundle, user_id, worker_id, start_time, remote):
         """
```
bkgoksel, line 716: Renamed and comment updated

codalab/model/bundle_model.py
```diff
@@ -830,19 +770,45 @@ def resume_bundle(self, bundle, bundle_update, row, user_id, worker_id, connecti
 
         return True
 
-    def finalize_bundle(self, bundle, user_id, exitcode, failure_message, connection):
+    def transition_bundle_worker_offline(self, bundle):
```
bkgoksel, line 773: Finalize moved below `transition_bundle_worker_offline`.
The code for `transition_bundle_worker_offline` is the same as the old `set_bundle_offline`, which just moved down here from above. None of the code diff below is changed code.

codalab/model/bundle_model.py
```diff
@@ -859,11 +825,12 @@ def finalize_bundle(self, bundle, user_id, exitcode, failure_message, connection
 
         return True
 
-    def finish_bundle(self, bundle):
```
bkgoksel, line 862: Renamed and comment updated

bkgoksel:
Added some comments @percyliang @epicfaace to make it easy to navigate this PR. Also might help to hide whitespace changes. There's actually very little change in terms of any real code. It's mostly renaming and moving things around.

percyliang (approved):
Can the linting be part of black?

bkgoksel:
> Can the linting be part of black?

Black doesn't care about logical nits (for example code blocks wrapped in `if True`, instance methods that could be static methods, unused variables, `True if condition else False` expressions etc.)

On second thought I'm not sure which tool exactly in my dev environment captured these errors, they were just marked in my Vim and I fixed them as I passed by.

epicfaace:
@bkgoksel could you please merge `master` into `frontend`? There are merge conflicts because this commit (https://github.com/codalab/codalab-worksheets/commit/33f74b23810a65a9ea2ee9ebe51419be3ac73c32) touched some of the same methods/files, so I thought it'd be best for you to merge this since you worked with the code and know it best...

bkgoksel:
sure thing!

bkgoksel:
Hmm I'm seeing only one merge conflict in the file I touched (`bundle_model.py`) but that's a method I have not modified at all. There are also a couple merge conflicts on the frontend code itself. Feels like I'm even less qualified to do this merge, but it doesn't seem like the conflict is big, and I'm guessing we just need to discard the `master` code and go with the frontend for this.

It'd be great if you can double check and confirm though @epicfaace .

epicfaace:
Hmm, you're right. I had assumed this PR had touched that method because the merge conflict only came up recently, but it must be another PR. will check and merge

PR #1454, opened 2019-10-02

### LINKED ISSUE #940: Bundle time use calculation inconsistent

bkgoksel:
This is what I believe to be the core change from #983 that should fix most inconsistencies we've had with bundle runtime calculation in the past.


bkgoksel:
Closed in favor of #1457 since the upstream branch name here breaks our docker builds.

PR #1455, opened 2019-10-02
redirect error report to slack channel

### LINKED ISSUE #1449: Send monitoring reports which contain system error to Slack channel instead of email
Currently, monitoring reports are sent through email. We need to redirect the reports to Slack when there is a system error.

candicegjing:
Fixed #1449 
Still need to decide whose token to use.


monitor.py
```diff
@@ -1,16 +1,14 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
```
percyliang: I think using `/usr/bin/env` is better

monitor.py
```diff
@@ -104,6 +105,32 @@ def send_email(subject, message):
     s.quit()
 
 
+def send_slack_notification(subject, message):
```
percyliang: Should factor this out into a separate `slack.py` file that can be used by other parts of the codebase.

monitor.py
```diff
@@ -140,9 +167,9 @@ def error_logs(error_type, s):
     last_t = last_sent[error_type]
     t = time.time()
 
-    # Send email only every 4 hours
+    # Send Slack notification only every 4 hours
     if t > last_t + 60 * 60 * 4:
-        send_email('%s [%d times]' % (error_type, n), s.split('\n'))
```
percyliang: I don't think we want to get rid of email.  There are some installations where people don't use slack

percyliang (changes requested):
There are other parts of the codebase that send email (e.g., when there are exceptions thrown), which should have slack added.

epicfaace:
Looking into this and maybe another way to do this is to _keep_ the e-mail integration. Slack can give you a special email address, and if you send emails to that address, the messages will show up in a particular channel (https://codalab.slack.com/apps/A0F81496D-email)

Not sure which way is better, though. It might be easier for deployment if we don't have to manage more variables such as the slack API key, but on the other hand, if we are going to have more deep slack integration in the future, it would be good to continue integrating directly via Slack's API.

(it's worth noting that the Email add-on is available only to workspaces with a "Slack Standard Plan" or above, which won't be a problem for us, but might be a problem if someone else with a free slack workspace wants to integrate with it)

percyliang:
Oh, I think if we can just do it via email, maybe we should just do that, since it minimizes the number of dependencies we have?

candicegjing:
Yes, I agree. If we can do it through email, we should use it. @percyliang Then for this ticket, should I just close this PR and set up the email configuration?

percyliang:
Yes, maybe convert this ticket into just some documentation on how to setup
Slack via email.

candicegjing:
Screen Shot 2019-10-11 at 1 21 56 PM

epicfaace (commented):
Will we need to modify https://github.com/codalab/codalab-worksheets/blob/master/monitor.py too?

percyliang (changes requested):
I don't think we need a code change here...just change our admin_email to point to slack.

docker/compose_files/docker-compose.yml
```diff
@@ -133,6 +133,22 @@ services:
       - service
       - worker
 
+  worker-shared-file-system:
```
percyliang: Why does this show up in this PR?

docs/Server-Setup.md
```diff
@@ -249,4 +249,21 @@ The issue is documented in detail [here](https://github.com/benoitc/gunicorn/iss
 is to unselect `/tmp` in Docker's preferences under file sharing.
 
 The image below shows where the file sharing pane is.
-<img src="images/docker-preferences.png" />
\ No newline at end of file
+<img src="images/docker-preferences.png" />
+
+## Sending Slack notification in monitoring service
+
+If you need to send Slack notification from monitoring service, there are two ways to configure your system.
```
percyliang: "Slack notifications from the `monitor.py` service"

docs/Server-Setup.md
```diff
@@ -249,4 +249,21 @@ The issue is documented in detail [here](https://github.com/benoitc/gunicorn/iss
...
 
 The image below shows where the file sharing pane is.
-<img src="images/docker-preferences.png" />
\ No newline at end of file
+<img src="images/docker-preferences.png" />
+
+## Sending Slack notification in monitoring service
+
+If you need to send Slack notification from monitoring service, there are two ways to configure your system. 
+1. Using Slack rest API
```
percyliang: We don't support this, right?  Delete?

docs/Server-Setup.md
```diff
@@ -249,4 +249,21 @@ The issue is documented in detail [here](https://github.com/benoitc/gunicorn/iss
...
+     * System prerequisite: install slackclient package on the platform where hosts monitoring service
+                `pip3 install slackclient`  
+     * Generate Slack API token via [create-api-token](https://slack.com/help/articles/215770388-create-and-regenerate-api-tokens)
+     * Figure out which channel you want to send notification to: SlACK_CHANNEL_ID
+     * Integrate your Slack messaging function to monitoring service: [Slack API basic_usage](https://slack.dev/python-slackclient/basic_usage.html)
+     * Start your monitoring service
+     * Note that this method doesn't have any restriction on your Slack plans.
+1. Using Slack Email App
```
percyliang: Give more context: say "System notifications (e.g., monitor.py) are sent to `$ADMIN_EMAIL`.  If you're using Slack, then you can set $ADMIN_EMAIL to a special email address which will show up in a designated slack channel.  See X for more details.  Note that this works only with the standard plan."

percyliang (changes requested):
I think this should not be labeled as ADMIN versus SLACK but rather as MONITOR (automatic messages that we don't reply to - happen to go to slack) and HELP (messages from users - happen to go to gmail).

percyliang (commented):
The prod/dev distinction should not appear in the codalab-worksheets repo.  It should be a configuration in our deployment repo.

codalab_service.py
```diff
@@ -215,7 +215,8 @@ def has_callable_default(self):
         default=100,
     ),
     ### Email
-    CodalabArg(name='admin_email', help='Email to send admin notifications to (e.g., monitoring)'),
+    CodalabArg(name='monitor_email', help='Monitoring email to send notifications to'),
+    CodalabArg(name='help_email', help='Help email to send user questions to'),
```
percyliang: Maybe call this `admin_email` and `support_email` to use consistent terminology with what's used in the `config.json` file.

candicegjing:
Closed via #1449

PR #1457, opened 2019-10-02

### LINKED ISSUE #940: Bundle time use calculation inconsistent

bkgoksel:
Taken the core fix from #983 without the cruft to calculate bundle runtime at one place only and avoid earlier inconsistency problems.


bkgoksel:
Closed in favor of #1458

PR #1458, opened 2019-10-02

### LINKED ISSUE #940: Bundle time use calculation inconsistent

bkgoksel:
This is what I believe to be the core change from #983 that should fix most inconsistencies we've had with bundle runtime calculation in the past. Same as #1454 but turns out our branch names can't include the # character in them or our docker builds break :(


codalab/model/bundle_model.py
```diff
@@ -817,7 +817,9 @@ def resume_bundle(self, bundle, bundle_update, row, user_id, worker_id, connecti
         metadata_update = {
             'run_status': bundle_update['run_status'],
             'last_updated': int(time.time()),
-            'time': time.time() - bundle_update['start_time'],
```
bkgoksel, line 820: Basically this was the line causing all the issues, trying to recompute time used with each update.
Now the time is only calculated by the worker. I guess technically someone could write an adversarial worker that always reports 0 time usage, but if we want to do server-side time calculation, that would be a bigger change and probably on a separate effort.

codalab/bundles/run_bundle.py
```diff
@@ -35,9 +35,9 @@ class RunBundle(DerivedBundle):
 
     METADATA_SPECS.append(MetadataSpec('actions', list, 'Actions (e.g., kill) that were performed on this run.', generated=True))
 
-    METADATA_SPECS.append(MetadataSpec('time', float, 'Amount of time (seconds) used by this run (total).', generated=True, formatting='duration'))
-    METADATA_SPECS.append(MetadataSpec('time_user', float, 'Amount of time (seconds) by user.', generated=True, formatting='duration'))
-    METADATA_SPECS.append(MetadataSpec('time_system', float, 'Amount of time (seconds) by the system.', generated=True, formatting='duration'))
+    METADATA_SPECS.append(MetadataSpec('time', float, 'Amount of time (seconds) used by this run command in total (user+system).', generated=True, formatting='duration'))
```
percyliang: Just to confirm, this doesn't include the CodaLab preparing/finalizing time?

percyliang: Actually, from below, it seems like `time` is `time_total`, which is some sort of wall clock time.  Should probably have a comment somewhere that explains this in more detail.

percyliang (changes requested):

codalab/bundles/run_bundle.py
```diff
@@ -35,9 +35,9 @@ class RunBundle(DerivedBundle):
 
     METADATA_SPECS.append(MetadataSpec('actions', list, 'Actions (e.g., kill) that were performed on this run.', generated=True))
 
-    METADATA_SPECS.append(MetadataSpec('time', float, 'Amount of time (seconds) used by this run (total).', generated=True, formatting='duration'))
-    METADATA_SPECS.append(MetadataSpec('time_user', float, 'Amount of time (seconds) by user.', generated=True, formatting='duration'))
-    METADATA_SPECS.append(MetadataSpec('time_system', float, 'Amount of time (seconds) by the system.', generated=True, formatting='duration'))
+    METADATA_SPECS.append(MetadataSpec('time', float, 'Amount of wall clock time (seconds) used by this run command in total. [Runtime of the Docker container excluding CodaLab related steps such as preparing/uploading results]', generated=True, formatting='duration'))
```
bkgoksel: Added the clarifying info here that this is the runtime of the Docker container for the job.

codalab/worker/local_run/local_run_manager.py
```diff
@@ -204,15 +204,18 @@ def create_run(self, bundle, resources):
             bundle=bundle,
             bundle_path=os.path.realpath(bundle_path),
             resources=resources,
-            start_time=now,
+            bundle_start_time=now,
+            container_start_time=None,
```
bkgoksel, line 208: Differentiated bundle and container start times here.

codalab/worker/local_run/local_run_manager.py
```diff
@@ -204,15 +204,18 @@ def create_run(self, bundle, resources):
...
+            container_time_total=0,
+            container_time_user=0,
+            container_time_system=0,
```
bkgoksel, line 211: Clarified the naming of these that they are runtimes of the docker container.

codalab/worker/local_run/local_run_state.py
```diff
@@ -306,22 +309,35 @@ def check_resource_utilization(run_state):
...
 
-            run_state = run_state._replace(time_used=time_used)
             run_state = run_state._replace(
                 max_memory=max(run_state.max_memory, run_stats.get('memory', 0))
             )
             run_state = run_state._replace(
                 disk_utilization=self.disk_utilization[bundle_uuid]['disk_utilization']
             )
 
+            container_time_total = time.time() - run_state.container_start_time
```
bkgoksel, line 320: Unified the logic for updating the time state here.

bkgoksel:
@percyliang did a larger overhaul, I think it is much clearer now and amenable to changing the time calculation logic in the future if we so want. (for example calculate and expose a 'total codalab time' like value)

codalab/bundles/run_bundle.py
percyliang: Don't think 'run command' is necessary - just say 'run'.

codalab/worker/local_run/local_run_manager.py
```diff
@@ -204,15 +204,18 @@ def create_run(self, bundle, resources):
             bundle=bundle,
             bundle_path=os.path.realpath(bundle_path),
             resources=resources,
-            start_time=now,
+            bundle_start_time=now,
+            container_start_time=None,
+            container_time_total=0,
```
percyliang, line 209: Semantically, shouldn't these be `None`?

percyliang (approved):
This is much clearer - thanks!

codalab/worker/local_run/local_run_manager.py
bkgoksel: This state object generally acts as a source of information as of the last time the run manager checked the run. My reasoning with setting the runtime variables to 0 was that at the moment of creation the bundle has not spent any time yet, and the container runtime stays at 0 until the container starts, which makes downstream cases a lot simpler (not having to deal with None checking etc.)

codalab/bundles/run_bundle.py
bkgoksel: done

PR #1469, opened 2019-10-04
Get rid of code duplication in bundle state updates

bkgoksel:
Make `last_updated` a metadata field of all named bundles, and update it every time bundle information is changed in the bundle model. Concretely, this is done by adding it as a metadata update field every time the `update_bundle` method is called. Previously, every caller of `update_bundle` basically had to manually insert the `last_updated` field into the `update_bundle` call.

Final part of the effort to extract the good bits of #983


codalab/bundles/named_bundle.py
```diff
@@ -21,6 +21,7 @@ class NamedBundle(Bundle):
         MetadataSpec('description', str, 'Full description of the bundle.', short_key='d'),
         MetadataSpec('tags', list, 'Space-separated list of tags used for search (e.g., machine-learning).', metavar='TAG',),
         MetadataSpec('created', int, 'Time when this bundle was created.', generated=True, formatting='date'),
+        MetadataSpec('last_updated', int, 'Time when this bundle was last updated.', generated=True, formatting='date'),
```
percyliang, line 24: In this context, need to clarify what 'updated' means...it's not just updating metadata or contents, it's more associated with workers, no?

codalab/model/bundle_model.py
```diff
@@ -753,26 +751,24 @@ def transition_bundle_running(self, bundle, bundle_update, row, user_id, worker_
...
-        metadata_update = {
-            'run_status': bundle_update['run_status'],
-            'last_updated': int(time.time()),
-            'time': time.time() - bundle_update['start_time'],
-            'remote': bundle_update['remote'],
+        bundle_update = {
+            'state': worker_bundle_update['state'],
+            'metadata': {
+                'run_status': worker_bundle_update['run_status'],
+                'time': time.time() - worker_bundle_update['start_time'],
```
percyliang: Why isn't this based on the `container_time_total` field?

percyliang (commented):
Generally good to consolidate, but is the setting of `last_updated` now too implicit?  I'm not quite sure of the semantics of it and how much we can control it.

codalab/model/bundle_model.py
bkgoksel: The other PR isn't merged yet :) I'll update depending on which gets merged first.

codalab/bundles/named_bundle.py
bkgoksel: Hmm I guess it depends on what this field is used for.
If it's more about tracking when a bundle was last updated by a worker, we should probably clarify this as that wasn't obvious to me and I thought it was a broader notion of updated.

bkgoksel: And if it really is `last_worker_checkin` or something, then this PR is probably not good, and we should go back to having the worker-related bundle state update methods adding this to the metadata manually.

percyliang: I think it's basically `last_worker_checkin` (not 100% sure), so maybe we can eliminate the duplicate code in the worker update (and make it clearer what this field is).

bkgoksel:
Closing for now since the semantics of `last_updated` are not very clear/it is not obvious this is what we want.

PR #1471, opened 2019-10-04
updated shebang line for cross-platform portability

### LINKED ISSUE #1470: Shebang line is not portable across different systems
In `codalab_service.py`, we currently use the following shebang line:
`#!/usr/bin/python3`

However, while this is the default on Ubuntu systems, it does not work on 
other OSs such as macOS.

Levilian:
fix #1470


candicegjing:
Looks good!

PR #1472, opened 2019-10-04
Pull newly updated image when cl run

### LINKED ISSUE #1355: when update docker images, `cl run` should pull
When run a job, should do the equivalent of `docker pull` to always make sure we have the latest version of a docker image.

candicegjing:
Fixed #1355

Questions:
The current change is too simple, I am not sure if we want to add some logic to check if there is a new image created in Dockerhub before pulling the image to local?

### DIFF
--- a/codalab/worker/local_run/docker_image_manager.py
+++ b/codalab/worker/local_run/docker_image_manager.py
@@ -125,7 +125,14 @@ class DockerImageManager:
 
     def get(self, image_spec):
         """
-        Request the docker image, starting its download if it's not available on the system
+        Always request the newest docker image from Dockerhub if it's not in downloading thread and return the current
+        downloading status(READY, FAILED, or DOWNLOADING).
+        When the requested image in the following states:
+        1. If it's not available on the platform, we download the image and return DOWNLOADING status.
+        2. If another thread is actively downloading it, we return DOWNLOADING status.
+        3. If another thread was downloading it but not active by the time the request was sent, we return the following status:
+            * READY if the image was downloaded successfully.
+            * FAILED if the image wasn't able to be downloaded due to any reason.
         :param image_spec: Repo image_spec of docker image being requested
         :returns: A DockerAvailabilityState object with the state of the docker image
         """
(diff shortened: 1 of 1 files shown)


bkgoksel (changes requested):
It is safe to pull every time since Docker is smart and doesn't re-download all the layers if they're found locally.

However making a direct raw call to `pull` here is bad, because it would mean for runs requesting new docker images, the entire worker would block until the pull is complete (the `pull` method is synchronous and blocks).

This logic needs to go into the `DockerImageManager`'s `get` method, where regardless of the state of the docker image in the cache, it gets moved into the downloading state. 

In fact, we can use this as an opportunity to really simplify the Docker image manager. We can get rid of the entire "I already have this image, so no need to pull" logic, which has caused issues in the past.

bkgoksel:
Happy to help with understanding/navigating the Docker image manager for this PR.

candicegjing:
Thanks @bkgoksel! I think I still have some questions regarding this ticket and first I want to make sure I understand what I should do for this ticket.
After reading through `DockerImageManager.get(image_spec)`, my understanding of this function is that if the image specified by `image_spec` is not found locally, then we will try to pull it from DockerHub, return states in success or failure depending on if it was able to download the image successfully. Then what this ticket wants to do is to change the order of the logic? Currently, we search for the image in local and then try to download it from Dockerhub if not found. But now, we always want first to download it from Dockerhub, and then if there is no such image in Dockerhub, we return failure; otherwise, it will succeed?

bkgoksel:
I think that sounds about right. More concretely we always want to call `pull_or_report` every time an image is requested. Of course we might need to make some implementation changes in there too.

candicegjing:
@bkgoksel What if the requested image is still in `DOWNLOADING` state, do we still want to re-download it again?

candicegjing:
@bkgoksel I updated the PR and not sure if keep checking on `self._downloading` dict to see if the downloading process has completed is desired before return the final state. Please let me know.

candicegjing:
@bkgoksel I have updated the logic. Let me know if there is anything i need to change.

bkgoksel (approved):
LGTM! 
Maybe update the method docstring as well.
Also should probably test this manually a bit, maybe you can deploy this branch on dev and fire some runs that request different images, fire a run that requests an image and after the image is downloaded fire another run that requests the same image. Even pushing a new image to the same tag and ensuring we get the updated one afterwards.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -125,7 +125,7 @@ def _cleanup(self):
 
     def get(self, image_spec):
         """
-        Request the docker image, starting its download if it's not available on the system
+        Always request the newest docker image from Dockerhub if it's not in downloading thread and return the current downloading status
```
percyliang: Please write more comprehensive documentation documenting the logic about the possible cases.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -143,82 +143,60 @@ def get(self, image_spec):
...
+                        self._downloading[image_spec]['success'] = True
+                        self._downloading[image_spec]['message'] = "Downloading image"
+                    except (docker.errors.APIError, docker.errors.ImageNotFound) as ex:
+                        logger.debug('Download for Docker image %s failed: %s', image_spec, ex)
+                        self._downloading[image_spec]['success'] = False
+                        self._downloading[image_spec][
+                            'message'
+                        ] = "Can't download image: {}".format(ex)
+
+                self._downloading.add_if_new(image_spec, threading.Thread(target=download, args=[]))
```
percyliang, line 200: Something seems weird here - `download` already sets `self._downloading`, so what is `add_if_new` doing on `self._downloading`?  Can you explain this more?

candicegjing: I think `add_if_new` is to add `image_spec` to the dictionary but without any fields (`success` or `message`) being set. By the time when the `download` function gets executed and `pull` function finished, `self._downloading` will be able to set all the fields (`success`, `message`) for `image_spec`.

candicegjing: I think `add_if_new` is to add `image_spec` to the dictionary but without any fields (`success` or `message`) being set. By the time when the download function gets executed and `pull` function finished, `self._downloading` will set all the fields (`success`, `message`) based on downloading status for `image_spec`.

bkgoksel: `self._downloading` is actually a `ThreadDict` which is a convenience class I wrote ([docs here](https://docs.google.com/document/d/1BaclAazh-XmPfzIwLKCUMJTFKNiV6YrcDsIbUe_VMgk/edit#heading=h.sfbdizxrwv8n)). It basically lets us package threads with metadata keys in the form of a nested dict, so the thread itself is a value of the dict, alongside other key-value pairs (like `success` or `message`)

Like Jing said, `add_if_new`:
1. checks if we already have a download thread for this image spec (does nothing if that's the case)
2. creates and starts the download thread for this image spec
3. stores it as a value in the `ThreadDict`.

The `download` function itself only modifies other associated fields of the dict.

PR #1474, opened 2019-10-05
Add expand button to control bundle row detail showing

yipenghe:
Implementing #1272 
1. Clicking on a bundle row highlights the row but doesn't expand automatically
2. Clicking on the expand button highlights the row and expands the detail
3. Icon switches between expand more & expand less
Screenshot from 2019-10-05 02-23-47
Screenshot from 2019-10-05 02-23-51
Screenshot from 2019-10-05 02-23-54

Should use () => {} than function(){} for class methods, the later doesn't bind this to the class

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -216,9 +216,9 @@ class BundleRow extends Component {
             }
             if (url)
                 rowContent = (
-                        <a href={url} className='bundle-link' target='_blank' style={{ display: 'inline-block', width: 60 }}>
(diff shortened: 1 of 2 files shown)


percyliang (changes requested):
Could we put the button on the far left?  Note that with different table schemas, the first column might not be the uuid.  Also, in the future, we will have Gmail-style check boxes to the left, so wonder how we can make it not too crowded...

yipenghe:
Made the button to be on the left, If we just want one more checkbox I think the spacing is still OK 
Screenshot from 2019-10-06 20-16-03

percyliang (approved):
Great!

Issue #1199, opened 2019-06-11
Improve rendering of run details panel


percyliang:
The run time should be formatted better (using the format function) and right aligned so it's not the first thing you see (it's not the most important thing).

PR #1475, opened 2019-10-05
Change display of run time for bundles

yipenghe:
Addressing #1199 
1. Changed to display run time below the command and positioned to right so it won't be the first information that comes to eye
2. Formatted to hh:mm:ss for run time > 1s, display 0.xxxxxs for run time less than 1 second
2. Truncate the time to show at most 5 decimals, run time smaller than 0.00001s will just show '< 0.00001s' instead
3. Added a clock icon

Screenshot from 2019-10-05 14-24-56
Screenshot from 2019-10-05 14-51-36

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/MainContent.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/MainContent.jsx
@@ -7,6 +7,7 @@ import Grid from '@material-ui/core/Grid';
 import { withStyles } from '@material-ui/core/styles';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
 import AccessTimeIcon from '@material-ui/icons/AccessTime';
+import {renderDuration} from '../../../util/worksheet_utils';
 
 import { FileBrowserLite } from '../../FileBrowser';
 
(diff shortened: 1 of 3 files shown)


yipenghe:
Formatted run time

percyliang (changes requested):
We should use the same rendering as the logic for rendering durations in `formatting.py`.  We can either pass string versions from the backend, or if that's inconvenient, just have a `Formatting.js` that parallels `formatting.py`.

yipenghe:
> We should use the same rendering as the logic for rendering durations in `formatting.py`. We can either pass string versions from the backend, or if that's inconvenient, just have a `Formatting.js` that parallels `formatting.py`.

Found a exisiting render duration function and used that

frontend/src/components/worksheets/BundleDetail/MainContent.jsx
```diff
@@ -30,21 +32,23 @@ class MainContent extends React.Component<
 		const stateSpecClass = bundleInfo.state === 'failed'
             ? 'failedState'
             : (bundleInfo.state === 'ready' ? 'readyState' : 'otherState');
-		
+        //Get the correct run time display
```
percyliang, line 36: nit: space before "Get"

frontend/src/components/worksheets/BundleDetail/MainContent.jsx
```diff
@@ -30,21 +32,23 @@ class MainContent extends React.Component<
 		const stateSpecClass = bundleInfo.state === 'failed'
             ? 'failedState'
             : (bundleInfo.state === 'ready' ? 'readyState' : 'otherState');
-		
+        //Get the correct run time display
+        let bundleRunTime;
```
percyliang: Simplify: `const bundleRunTime = bundleInfo.metadata.time ? renderDuration(...) : ...`;

percyliang (approved):

frontend/src/components/worksheets/BundleDetail/MainContent.jsx
yipenghe: Thanks, addressed!

PR #1479, opened 2019-10-07
Fix Python3 migration issue

### LINKED ISSUE #1478: [USER ISSUE] competitiond.py Python3 migration issue
`competitiond.py` was failed in running Python3 with the following errors;
1. `Traceback (most recent call last):
  File "competitiond.py", line 848, in <module>
    main()
  File "competitiond.py", line 844, in main
    comp.run_once()
  File "competitiond.py", line 755, in run_once
    self.generate_leaderboard(num_total_submissions, num_period_submissions)
  File "competitiond.py", line 718, in generate_leaderboard
    reverse=True,
TypeError: '<' not supported between instances of 'NoneType' and 'float'`
2. `
  File "competitiond.py", line 848, in <module>
    main()
  File "competitiond.py", line 844, in main
    comp.run_once()
  File "competitiond.py", line 729, in run_once
    submissions, num_total_submissions, num_period_submissions = self.collect_submissions()
  File "competitiond.py", line 428, in collect_submissions
    submissions, previous_submission_ids, num_total_submissions, num_period_submissions
  File "competitiond.py", line 384, in _filter_submissions
    for key, bundle in (submissions.items()):
RuntimeError: dictionary changed size during iteration
`

candicegjing:
Fixed #1478

### DIFF
--- a/scripts/competitiond.py
+++ b/scripts/competitiond.py
@@ -381,7 +381,7 @@ class Competition(object):
         self, submissions, previous_submission_ids, num_total_submissions, num_period_submissions
     ):
         # Drop submission if user has exceeded their quota
-        for key, bundle in submissions.items():
+        for key, bundle in list(submissions.items()):
             # Drop submission if we already ran it before
             if bundle['id'] in previous_submission_ids:
                 logger.debug(
@@ -708,7 +708,10 @@ class Competition(object):
 
         # Sort by the scores, descending
         leaderboard.sort(
-            key=lambda e: tuple(e['scores'][spec['name']] for spec in self.config['score_specs']),
+            key=lambda e: [
+                (e['scores'][spec['name']] is not None, e['scores'][spec['name']])
+                for spec in self.config['score_specs']
+            ],
             reverse=True,
         )
 


percyliang (commented):
Does this change the JSON that's output?  If so, it'd be better to change the sorting function to not crash on None.  In general, I think it's better to keep backward compatibility if possible and also have values that are semantically accurate (here, `None` more accurately reflects that there was no score).

scripts/competitiond.py
```diff
@@ -381,7 +381,7 @@ def _filter_submissions(
         self, submissions, previous_submission_ids, num_total_submissions, num_period_submissions
     ):
         # Drop submission if user has exceeded their quota
-        for key, bundle in submissions.items():
+        for key, bundle in list(submissions.items()):
```
epicfaace, line 384: Good catch 👍

scripts/competitiond.py
```diff
@@ -708,7 +708,9 @@ def generate_leaderboard(self, num_total_submissions, num_period_submissions):
 
         # Sort by the scores, descending
         leaderboard.sort(
-            key=lambda e: tuple(e['scores'][spec['name']] for spec in self.config['score_specs']),
+            key=lambda e: tuple(
```
epicfaace: Yeah, this looks like it will return an array of booleans instead of an array of scores. What was the initial error here?

candicegjing: This one: `Traceback (most recent call last): File "competitiond.py", line 848, in <module> main() File "competitiond.py", line 844, in main comp.run_once() File "competitiond.py", line 755, in run_once self.generate_leaderboard(num_total_submissions, num_period_submissions) File "competitiond.py", line 718, in generate_leaderboard reverse=True, TypeError: '<' not supported between instances of 'NoneType' and 'float'` 
Python3 doesn't support this None value comparsion, but Python2 does.

scripts/competitiond.py
```diff
@@ -708,7 +708,9 @@ def generate_leaderboard(self, num_total_submissions, num_period_submissions):
...
+                e['scores'][spec['name']] is not None for spec in self.config['score_specs']
```
epicfaace: Maybe this would fix it @candicegjing ?

```suggestion
                e['scores'][spec['name']] or 0 for spec in self.config['score_specs']
```

candicegjing: I think with `is not None`, we sort all the bundles with `null` scores at the end of the result and all the `not null` scores on the top.

epicfaace (approved):

PR #1482, opened 2019-10-08
dialog box has higher z-index over sticky worksheet title and cli

Levilian:
Fix: #1460

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -520,7 +520,6 @@ const styles = (theme) => ({
         height: 36,
         borderBottom: '2px solid #ddd',
     },
-
     highlight: {
         backgroundColor: `${theme.color.primary.lightest} !important`,
     },
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -6,10 +6,11 @@ import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
 import IconButton from '@material-ui/core/IconButton';
 import Button from '@material-ui/core/Button';
-import Modal from '@material-ui/core/Modal';
-import Paper from '@material-ui/core/Paper';
-import Typography from '@material-ui/core/Typography';
-
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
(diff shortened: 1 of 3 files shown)


Levilian:
Screen Shot 2019-10-08 at 12 36 33 AM
Now the dialog box looks much better (I used Dialog instead of Model), and the weird highlight of CLI and worksheet title is gone.

frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -522,20 +520,7 @@ const styles = (theme) => ({
...
-        padding: theme.spacing.larger,
-        transform: 'translateY(-50%) translateX(-50%)',
-    },
-    flexRow: {
-        display: 'flex',
-        flexDirection: 'row',
-        alignItems: 'center',
-        justifyContent: 'flex-end',
-    },
+
```
yipenghe: Looks like an extra line

frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -342,31 +343,28 @@ class BundleRow extends Component {
...
+                                <DialogActions>
+                                    <Button color='primary' onClick={this.toggleDeletePopup}>
+                                        CANCEL
+                                    </Button>
+                                    <Button color='primary' onClick={this.deleteItem} autoFocus>
+                                        DELETE
+                                    </Button>
+                                </DialogActions>
+                            </Dialog>
+
```
yipenghe, line 367: ditto

frontend/src/css/imports.css
```diff
@@ -5562,7 +5562,7 @@ button.close {
   position: absolute;
   top: 0;
   left: 0;
-  z-index: 1060;
+  z-index: 1090;
```
yipenghe: Doesn't seem like any element is using this class, is this change for future usage? (to popover the header?)

Levilian: Yeah we use .popover somewhere else and just because I made changes to the worksheet title and CLI, I wanted to assign a higher z-index to .popover than any of the others. 

Is this a good practice? @epicfaace

yipenghe: Sounds good

epicfaace: Yeah, that's good.

epicfaace (approved):
Looks good. Quick question -- would it have been better to keep the z-indexes in imports.css constant and just change the z-index in WorksheetHeader.js to be 1059?

epicfaace:
> Looks good. Quick question -- would it have been better to keep the z-indexes in imports.css constant and just change the z-index in WorksheetHeader.js to be 1059?

@Levilian since you went with this suggestion, then I don't think you needed to change the z-indexes in imports.css.

PR #1487, opened 2019-10-09
Fill owner column with real value

### LINKED ISSUE #1344: "owner column" doesn't show owner information properly in "What's running on CodaLab?" page
**Describe the bug**
"Owner column" doesn't show owner information properly in "What's running on CodaLab?" page

**Codalab Information**
- *UUID of worksheet where problem occurred*: 0xa590fd1b68944a1a95c1c40c4931dc7b

**To Reproduce**
1. Go to https://worksheets.codalab.org/worksheets/0xa590fd1b68944a1a95c1c40c4931dc7b
2. Check the column "owner" in all tables.
3. See the owner information is not properly displayed.

**Expected behavior**
owner column should display owner information

**Screenshots**
Screen Shot 2019-09-03 at 10 02 45 PM

candicegjing:
Fixed #1344. Currently `owner_name` doesn't exist in `bundle_info` dictionary. If we want to have a real name return back, I'll find a way to transfer `owner_id` to `owner_name` by populating a new field directory to `bundle_info`.

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -332,18 +332,27 @@ def parse_worksheet_form(form_result, model, user, worksheet_uuid):
 
 
 def is_file_genpath(genpath):
-    # Return whether the genpath is a file (e.g., '/stdout') or not (e.g., 'command')
+    """
+    Determine whether the genpath is a file (e.g., '/stdout') or not (e.g., 'command')
+    :param genpath: a generalized path
+    :return: a boolean value indicating if the genpath is a file.
+    """
     return genpath.startswith('/')
 
 
-def interpret_genpath(bundle_info, genpath):
+def interpret_genpath(bundle_info, genpath, db_model=None, owner_cache=None):
     """
     Quickly interpret the genpaths (generalized path) that only require looking
     bundle_info (e.g., 'time', 'command').  The interpretation of generalized
     paths that require reading files is done by interpret_file_genpath.
+    If genpath is referring to a file, then just returns instructions for fetching that file rather than actually doing it.
+    :param bundle_info: dictionary which contains metadata of current bundle's information, e.g. uuid, bundle_type, owner_id, etc.
+    :param genpath: a generalized path, e.g. column names(summary, owner_name, etc.), args.
+    :param db_model (optional): database model which is used to query database
+    :param owner_cache (optional): a dictionary stores mappings from owner_id to owner_name
+    :return: the interpretation of genpath
     """
-    # If genpath is referring to a file, then just returns instructions for
-    # fetching that file rather than actually doing it.
+
     if is_file_genpath(genpath):
         return (bundle_info['uuid'], genpath)
 
@@ -434,6 +443,15 @@ def interpret_genpath(bundle_info, genpath):
             # FIXME(sckoo): we will be passing the old permissions format into this
             # which has been updated to accommodate the new formatting
             return group_permissions_str(bundle_info['group_permissions'])
+    elif genpath == 'owner_name':
+        if 'owner_id' in bundle_info:
+            if owner_cache is not None and bundle_info['owner_id'] in owner_cache:
+                return owner_cache[bundle_info['owner_id']]
+            else:
+                # We might batch this database operation in the future
+                owner = db_model.get_user(user_id=bundle_info['owner_id'])
+                owner_cache[bundle_info['owner_id']] = owner.user_name
+                return owner.user_name
 
     # Bundle field?
     value = bundle_info.get(genpath)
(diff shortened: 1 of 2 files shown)


yipenghe (commented):
Can you attach a screenshot on how it looks like right now?

candicegjing:
Screen Shot 2019-10-09 at 8 54 18 AM

It's currently filled in with `owner_id` but I feel it's probably better to fill in with `owner_name`. I am trying to find a way to add that field in.

epicfaace (commented):
Looks good -- can you also add a test? It could go something like this (with the right directives replaced, though) -- and you can test at the end to see if the rendered table contains the owner names as opposed to the ids:

```
    wuuid = run_command([cl, 'work', '-u'])
    run_command([cl, 'add', 'text', '% search'])
    check_contains(['...', run_command([cl, 'print']))
```

I think you could add it to the "worksheet_search" test module in test_cli.py

candicegjing:
I have updated the `owner_id` to `owner_name` in `owner` column. 
Screen Shot 2019-10-10 at 10 14 04 AM

codalab/lib/worksheet_util.py
```diff
@@ -434,6 +434,12 @@ def friendly_render_dep(dep):
             # FIXME(sckoo): we will be passing the old permissions format into this
             # which has been updated to accommodate the new formatting
             return group_permissions_str(bundle_info['group_permissions'])
+    elif genpath == 'owner_name':
+        from bottle import local
```
yipenghe: Imports are better put on top of file, see https://www.python.org/dev/peps/pep-0008/#imports

codalab/lib/worksheet_util.py
```diff
@@ -434,6 +434,12 @@ def friendly_render_dep(dep):
...
+
```
yipenghe: extra line

yipenghe (changes requested):

percyliang (changes requested):
worksheet_util.py shouldn't depend on bottle.  An intermediate solution is passing the `model` in for now.  More generally, we need to think about efficiency...here, we have to make a DB request for each row, which might be too slow, and we might want to batch DB queries.

candicegjing:
> worksheet_util.py shouldn't depend on bottle. An intermediate solution is passing the `model` in for now. More generally, we need to think about efficiency...here, we have to make a DB request for each row, which might be too slow, and we might want to batch DB queries.

candicegjing:
For each time we load a worksheet, I personally think the bundles listed under the current worksheet are more likely to belong to the owner. Based on this intuition, I updated the PR by passing a cached dictionary to the interpret_genpath function. When we parse the `bundle_info` one by one, if we see a previously visited `owner_id`, we can directly return it, instead of making a db call. 
Another thought is that we can change the way of parsing all the entries of `bundle_infos` from `row by row` to `column by column`. Then we can get a full list of `owner` for all the bundles and send a one-time db call.

codalab/lib/worksheet_util.py
```diff
@@ -336,7 +336,7 @@ def is_file_genpath(genpath):
     return genpath.startswith('/')
 
 
-def interpret_genpath(bundle_info, genpath):
+def interpret_genpath(bundle_info, genpath, db_model=None, owner_cache=None):
     """
     Quickly interpret the genpaths (generalized path) that only require looking
```
percyliang, line 345: Document the new arguments.  Specifically, note what the input/output of `owner_cache` is.

codalab/lib/worksheet_util.py
```diff
@@ -755,11 +763,18 @@ def raise_genpath_usage_error():
             header = tuple(name for (name, genpath, post) in schema)
             rows = []
             processed_bundle_infos = []
+            # Cache the mapping between owner_id to owner_name on current worksheet
+            owner_cache = dict()
```
percyliang: nit: just do `{}`

percyliang (approved):
LGTM modulo a few comments.

codalab/lib/worksheet_util.py
```diff
@@ -434,6 +434,14 @@ def friendly_render_dep(dep):
             # FIXME(sckoo): we will be passing the old permissions format into this
             # which has been updated to accommodate the new formatting
             return group_permissions_str(bundle_info['group_permissions'])
+    elif genpath == 'owner_name':
+        if 'owner_id' in bundle_info:
+            if owner_cache is not None and bundle_info['owner_id'] in owner_cache:
+                return owner_cache[bundle_info['owner_id']]
+            else:
+                owner = db_model.get_user(user_id=bundle_info['owner_id'])
```
percyliang, line 452: Put a comment that in the future, we might want to batch this.

percyliang:
On the other hand, the main reason you'd want to display the owner column (e.g., on https://worksheets.codalab.org/worksheets/0xa590fd1b68944a1a95c1c40c4931dc7b) is when there are different owners (otherwise, why bother).

PR #1489, opened 2019-10-09
Fix mathjax not defined error

### LINKED ISSUE #1210: [Frontend] Cannot read "Hub" of undefined
Just kind of randomly got this error.

Probably because of this line of code? https://github.com/codalab/codalab-worksheets/blob/dda05c16207e566451a20aa5d413e5e1f525bd8d/frontend/public/index.html#L35

image

epicfaace:
Fixes #1210. I actually wasn't able to reproduce this error, but added this just in case.


yipenghe (approved):
Yeh the error is hard to reproduce, this should fix it

PR #1501, opened 2019-10-14
Merge two bundle manager files and simplify run message

bkgoksel:
Gets rid of the bundle_manager/default_bundle_manager dichotomy by
merging the two classes.
Also renames the `request_` prefix from the resource fields of the run
message the bundle manager sends to the worker (ie `request_cpus`
becomes `cpus`, `request_network` becomes `network` etc) for simplicity.


codalab/server/bundle_manager.py
```diff
@@ -29,16 +29,11 @@ class BundleManager(object):
...
+    def __init__(self, codalab_manager):
         config = codalab_manager.config.get('workers')
         if not config:
             print('config.json file missing a workers section.', file=sys.stderr)
-            exit(1)
-
-        from codalab.server.default_bundle_manager import DefaultBundleManager
-
-        self = DefaultBundleManager()
+            sys.exit(1)
```
bkgoksel, line 36: [It's bad to use `exit()` to exit from programs.](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python)

codalab/server/bundle_manager.py
```diff
@@ -329,8 +322,6 @@ def _schedule_run_bundles_on_workers(self, workers, user_owned):
             for worker in workers_list:
                 if self._try_start_bundle(workers, worker, bundle):
                     break
-                else:
-                    continue  # Try the next worker.
```
bkgoksel, line 333: This code doesn't do anything since the for loop naturally continues at this point.

codalab/server/bundle_manager.py
```diff
@@ -324,13 +317,13 @@ def _schedule_run_bundles_on_workers(self, workers, user_owned):
                     continue  # Don't start this bundle yet
                 workers_list = workers.user_owned_workers(self._model.root_user_id)
 
-            workers_list = self._filter_and_sort_workers(workers_list, bundle)
+            self._deduct_worker_resources(workers_list)
```
bkgoksel: Felt like better abstraction to pass an up-to-date workers list into the `filter` function rather than having that function deduct its worker resources inside.

codalab/server/bundle_manager.py
```diff
@@ -339,11 +332,13 @@ def _deduct_worker_resources(self, workers_list):
         for worker in workers_list:
             for uuid in worker['run_uuids']:
                 bundle = self._model.get_bundle(uuid)
-                worker['cpus'] -= self._compute_request_cpus(bundle)
-                worker['gpus'] -= self._compute_request_gpus(bundle)
-                worker['memory_bytes'] -= self._compute_request_memory(bundle)
+                bundle_resources = self._compute_bundle_resources(bundle)
```
bkgoksel, line 329: While this adds a line of code I think it's just better to have method that does all resource computations rather than having to call individual functions for individual fields every time we need them.

codalab/server/bundle_manager.py
```diff
@@ -355,26 +350,22 @@ def _filter_and_sort_workers(self, workers_list, bundle):
...
-        # Filter by memory.
-        request_memory = self._compute_request_memory(bundle)
-        if request_memory:
+        if bundle_resources.gpus:
             workers_list = [
-                worker for worker in workers_list if worker['memory_bytes'] >= request_memory
+                worker for worker in workers_list if worker['gpus'] >= bundle_resources.gpus
             ]
 
+        # Filter by memory.
```
bkgoksel, line 359: There's no `free-for-all` workers anymore so all runs request some amount of memory and cpu by default.

codalab/server/bundle_manager.py
```diff
@@ -522,18 +516,19 @@ def _construct_run_message(self, worker, bundle):
...
-        resources['request_memory'] = self._compute_request_memory(bundle)
-        resources['request_disk'] = self._compute_request_disk(bundle)
-        resources['request_network'] = bundle.metadata.request_network
-
-        return message
+        bundle_resources = self._compute_bundle_resources(bundle)
+        message['resources'] = bundle_resources.to_dict()
+
+    def _compute_bundle_resources(self, bundle):
+        return RunResources(
```
bkgoksel, line 525: We have an object that encapsulates all bundle resources rather than a dict that comes with serialization/deserialization methods. This is so that we don't shoot ourselves in the foot by missing a dict field and only noticing when things go wrong etc.

tests/server/bundle_manager_test.py
```diff
@@ -3,16 +3,11 @@
...
 from codalab.bundles import RunBundle
 
 
 class BundleManagerTest(unittest.TestCase):
-    def setUp(self):
-        self.manager = BundleManager()
-        self.manager._default_request_cpus = 1
-        self.manager._default_request_gpus = 0
-        self.manager._default_request_queue = None
-
```
bkgoksel, line 15: No need to set up a full on `BundleManager` object anymore since we're just testing the static worker sorting method.

codalab/worker/bundle_state.py
```diff
@@ -31,3 +31,52 @@ class State(object):
...
+    def __init__(self, cpus, gpus, docker_image, time, memory, disk, network):
+        self.cpus = cpus  # type: int
+        self.gpus = gpus  # type: str
+        self.docker_image = docker_image  # type: str
+        self.time = time  # type: int
+        self.memory = memory  # type: int
+        self.disk = disk  # type: int
+        self.network = network  # type: bool
+
+        if ":" not in self.docker_image:
```
percyliang: Should we be adding this here or closer to when things need to get resolved?  I'm wary of having transformations of user input in too many places...

codalab/worker/bundle_state.py
```diff
@@ -31,3 +31,52 @@ class State(object):
...
     ACTIVE_STATES = {MAKING, STARTING, RUNNING, FINALIZING, PREPARING}
     FINAL_STATES = {READY, FAILED, KILLED}
+
+
+class RunResources(object):
+    """
+    Defines all the resource fields the server propagates to the worker for its runs
+    """
+
+    def __init__(self, cpus, gpus, docker_image, time, memory, disk, network):
```
fabeschan, line 41: might be a good idea to start inserting type annotations in new code (now that we've moved to py3)

percyliang (commented):
Great, this makes things cleaner!  I have some reservations about the dropping of `request`...because it might introduce more chance for confusion with bare `time`...  More accurately, it should be `max_time` or something, but the `request` prefix filled the role of specifying the `max` part...

codalab/worker/bundle_state.py
bkgoksel: Hmm, I've been slowly adding comment-annotations to keep py2 compatibility. I guess won't hurt to keep doing that.

codalab/worker/bundle_state.py
bkgoksel: Moved it to bundle manager to have it be in a more expected place. I agree this is a weird location to modify parameters.

bkgoksel:
> Great, this makes things cleaner! I have some reservations about the dropping of `request`...because it might introduce more chance for confusion with bare `time`... More accurately, it should be `max_time` or something, but the `request` prefix filled the role of specifying the `max` part...

The renaming is only done for server <-> worker communication and on the worker-side, mainly to reduce verbosity (also because these are all fields of a `resources` object now, so there's less room for confusion)

codalab/server/bundle_manager.py
bkgoksel: Also I just realized that before this change, we actually were not filtering for memory when the request memory was the default value (ie 2gb). Shouldn't have been a practical issues since our workers rarely run out of memory but still, might've caused issues in the future.

codalab/worker/bundle_state.py
bkgoksel: Doing this in another PR that replaces a few more of these unstructured dicts with objects

percyliang:
Are there going to be problems using an old worker against the new server if names are changed?

percyliang (approved):

bkgoksel:
> 
> 
> Are there going to be problems using an old worker against the new server if names are changed?

Yeah you're right this would be a breaking change. But so was #1458, and we may have already made  a cut with it so we may have broken backwards compatibility of the worker already.

percyliang:
Can we have some sort of version check where if the worker is too old, then
there's at least a meaningful error message?

PR #1502, opened 2019-10-15
Update DockerImageManager cleanup function

### LINKED ISSUE #1030: Dependency cleanup doesn't seem to work
Today (2/22/19) one of our gpuworkers became unresponsive as its dependency cache became full and it seemed like it couldn't clear out the cache appropriately. Potentially related to:
- #890 
- #413 
- #986 
- #952

candicegjing:
Fixed #1030 . 
Based on the new get function(which is in the reviewing stage and hasn't been merged to master yet), since we always want to pull the image when it's requested, I feel the `image_cache` and `LRU` algorithm in `cleanup()` function could be simplified. `image_cache` could still be used to decide when to prune the system.   
**Question is that I am not sure if we want to only remove the image that is downloaded from `DockerImageManager`, as docker image prune will remove any image that's unused.**

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -30,7 +30,11 @@ def wrap_exception(message):
                 return f(*args, **kwargs)
             except DockerException as e:
                 raise DockerException(message + ': ' + str(e))
-            except (docker.errors.APIError, docker.errors.ImageNotFound) as e:
+            except (
+                docker.errors.APIError,
+                docker.errors.ImageNotFound,
+                docker.errors.NotFound,
+            ) as e:
                 raise DockerException(message + ': ' + str(e))
 
         return wrapper
--- a/codalab/worker/local_run/docker_image_manager.py
+++ b/codalab/worker/local_run/docker_image_manager.py
@@ -79,8 +79,9 @@ class DockerImageManager:
     def _cleanup(self):
         """
         Prunes the image cache for runs.
-        1. Only care about images we (this DockerImageManager) downloaded and know about
-        2. We use sum of VirtualSize's, which is an upper bound on the disk use of our images:
+        1. Only care about images we (this DockerImageManager) downloaded and know about.
+        2. We also try to prune any dangling docker images on the system.
+        3. We use sum of VirtualSize's, which is an upper bound on the disk use of our images:
             in case no images share any intermediate layers, this will be the real disk use,
             however if images share layers, the virtual size will count that layer's size for each
             image that uses it, even though it's stored only once in the disk. The 'Size' field
@@ -91,8 +92,9 @@ class DockerImageManager:
             but because of (1) we don't want to use that.
         """
         while not self._stop:
+            # Sort the image cache in LRU order
             deletable_entries = set(self._image_cache.values())
-            disk_use = sum(cache_entry.virtual_size for cache_entry in deletable_entries)
+            disk_use = sum(entry.virtual_size for entry in deletable_entries)
             while disk_use > self._max_image_cache_size:
                 entry_to_remove = min(deletable_entries, key=lambda entry: entry.last_used)
                 logger.info(
@@ -102,13 +104,10 @@ class DockerImageManager:
                     entry_to_remove.digest,
                 )
                 try:
-                    image_to_delete = self._docker.images.get(entry_to_remove.id)
-                    tags_to_delete = image_to_delete.tags
-                    for tag in tags_to_delete:
-                        self._docker.images.remove(tag)
-                    # if we successfully removed the image also remove its cache entry
+                    # Delete images from image cache in LRU order
+                    self._docker.images.remove(entry_to_remove.id)
                     del self._image_cache[entry_to_remove.digest]
-                except docker.errors.NotFound:
+                except docker.errors.ImageNotFound:
                     # image doesn't exist anymore for some reason, stop tracking it
                     del self._image_cache[entry_to_remove.digest]
                 except docker.errors.APIError as err:
--- a/codalab/worker/local_run/local_dependency_manager.py
+++ b/codalab/worker/local_run/local_dependency_manager.py
@@ -78,6 +78,8 @@ class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager)
         # DependencyKey -> WorkerThread(thread, success, failure_message)
         self._downloading = ThreadDict(fields={'success': False, 'failure_message': None})
         self._load_state()
+        # Sync states between dependency-state.json and dependency directories on the local file system.
+        self._sync_state()
 
         self._stop = False
         self._main_thread = None
@@ -87,44 +89,76 @@ class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager)
             self._state_committer.commit({'dependencies': self._dependencies, 'paths': self._paths})
 
     def _load_state(self):
+        """
+        Load states from dependencies-state.json, which contains information about bundles (e.g., state, dependencies,
+        last used, etc.) and populates values for self._dependencies, self._dependency_locks, and self._paths
+        """
         state = self._state_committer.load(default={'dependencies': {}, 'paths': set()})
+
         dependencies = {}
         dependency_locks = {}
-        paths = set()
-        for dep_key, dep_state in state['dependencies'].items():
-            full_path = os.path.join(self.dependencies_dir, dep_state.path)
-            if os.path.exists(full_path):
-                dependencies[dep_key] = dep_state
-                dependency_locks[dep_key] = threading.RLock()
-            else:
-                logger.info(
-                    "Dependency {} in loaded state but its path {} doesn't exist in the filesystem".format(
-                        dep_key, full_path
-                    )
-                )
-            if dep_state.path not in state['paths']:
-                state['paths'].add(dep_state.path)
-                logger.info(
-                    "Dependency {} in loaded state but its path {} is not in the loaded paths {}".format(
-                        dep_key, dep_state.path, state['paths']
-                    )
-                )
-        for path in state['paths']:
-            full_path = os.path.join(self.dependencies_dir, path)
-            if os.path.exists(full_path):
-                paths.add(path)
-            else:
-                logger.info(
-                    "Path {} in loaded state but doesn't exist in the filesystem".format(full_path)
-                )
+
+        for dep, dep_state in state['dependencies'].items():
+            dependencies[dep] = dep_state
+            dependency_locks[dep] = threading.RLock()
 
         with self._global_lock, self._paths_lock:
             self._dependencies = dependencies
             self._dependency_locks = dependency_locks
-            self._paths = paths
+            self._paths = state['paths']
+
         logger.info(
-            '{} dependencies, {} paths in cache.'.format(len(self._dependencies), len(self._paths))
+            'Loaded {} dependencies, {} paths from cache.'.format(
+                len(self._dependencies), len(self._paths)
+            )
         )
+
+    def _sync_state(self):
+        """
+        Synchronize dependency states between dependencies-state.json and the local file system as follows:
+        1. self._dependencies, self._dependency_locks, and self._paths: populated from dependencies-state.json in function _load_state()
+        2. directories on the local file system: the bundle contents
+        This function forces the 1 and 2 to be in sync by taking the intersection (e.g., deleting bundles from the
+        local file system that don't appear in the dependencies-state.json and vice-versa)
+        """
+        # Get all the dependency directories on the local file system under self.dependencies_dir
+        local_directories = set(os.listdir(self.dependencies_dir))
+
+        # Remove the orphaned dependencies from self._dependencies and
+        # self._dependency_locks if they don't exist on the local file system
+        dependencies_to_remove = [
+            dep
+            for dep, dep_state in self._dependencies.items()
+            if dep_state.path not in local_directories
+        ]
+        for dep in dependencies_to_remove:
+            logger.info(
+                "Dependency {} in dependency state but its path {} doesn't exist on the local file system. "
+                "Removing it from dependency state.".format(
+                    dep, os.path.join(self.dependencies_dir, self._dependencies[dep].path)
+                )
+            )
+            del self._dependencies[dep]
+            del self._dependency_locks[dep]
+
+        # Get the paths that exist in dependency state, loaded path and the local file system
+        paths_in_loaded_state = [dep_state.path for dep_state in self._dependencies.values()]
+        self._paths = self._paths.intersection(paths_in_loaded_state).intersection(
+            local_directories
+        )
+
+        # Remove the orphaned directories from the local file system
+        directories_to_remove = local_directories - self._paths
+        for dir in directories_to_remove:
+            full_path = os.path.join(self.dependencies_dir, dir)
+            logger.info(
+                "Remove orphaned directory {} from the local file system.".format(full_path)
+            )
+            remove_path(full_path)
+
+        # Save the current state back to the state file: dependency-state.json as
+        # the current state might have been changed during the state syncing phase
+        self._save_state()
 
     def start(self):
         logger.info('Starting local dependency manager')
@@ -346,7 +380,7 @@ class LocalFileSystemDependencyManager(StateTransitioner, BaseDependencyManager)
 
     def _store_dependency(self, dependency_path, fileobj, target_type):
         """
-        Copy the dependency fileobj to its path in the local filesystem
+        Copy the dependency fileobj to its path on the local filesystem
         Overwrite existing files by the same name if found
         (may happen if filesystem modified outside the dependency manager,
          for example during an update if the state gets reset but filesystem


percyliang:
You you please summarize what was wrong with the old logic that's fixed, and why is this only a partial fix (because it's only image dependencies)?

candicegjing:
I split up the original git issue into two PRs is because I feel it might be easier for the reviewer to review two different cleanup logics.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -94,33 +94,26 @@ def _cleanup(self):
...
-                    image_to_delete = self._docker.images.get(entry_to_remove.id)
-                    tags_to_delete = image_to_delete.tags
-                    for tag in tags_to_delete:
-                        self._docker.images.remove(tag)
-                    # if we successfully removed the image also remove its cache entry
-                    del self._image_cache[entry_to_remove.digest]
-                except docker.errors.NotFound:
-                    # image doesn't exist anymore for some reason, stop tracking it
-                    del self._image_cache[entry_to_remove.digest]
+                    deleted_entries = self._docker.images.prune()
```
bkgoksel: Using docker's `prune` changes the logic here significantly. 
What `prune` does is it deletes any image that's not used by a container at this time. This includes images that could've been pulled by other Docker users on this machine. While on our workers that's probably fine, in shared-compute deployment scenarios like Slurm or AWS batch this is actually somehow dangerous, so we should probably not use it.

bkgoksel (changes requested):
I'm concerned about using `prune` since it modifies the system docker state outside our reach.

codalab/worker/local_run/docker_image_manager.py
candicegjing: I also had concerns about using `prune` here as I mentioned at the beginning of the ticket.  wasn't sure if we will share the environment with anyone else. But if we will use a shared environment with others, we shouldn't use this directly.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -91,8 +92,9 @@ def _cleanup(self):
             but because of (1) we don't want to use that.
         """
         while not self._stop:
+            # We sort the image cache in LRU order
             deletable_entries = set(self._image_cache.values())
-            disk_use = sum(cache_entry.virtual_size for cache_entry in deletable_entries)
+            disk_use = sum(entry.virtual_size for entry in deletable_entries)
```
candicegjing, line 97: I changed the variable name here just for consistency. At the bottom of this function, there is a similar logic that uses `entry`.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -102,13 +104,20 @@ def _cleanup(self):
                     entry_to_remove.digest,
                 )
                 try:
-                    image_to_delete = self._docker.images.get(entry_to_remove.id)
-                    tags_to_delete = image_to_delete.tags
-                    for tag in tags_to_delete:
-                        self._docker.images.remove(tag)
-                    # if we successfully removed the image also remove its cache entry
+                    # We try to delete images from image cache in LRU order
+                    self._docker.images.remove(entry_to_remove.id)
```
candicegjing, line 108: From the docker rest api documentation, it seems that we can directly remove the docker image given image id provided.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -102,13 +104,20 @@ def _cleanup(self):
...
                     del self._image_cache[entry_to_remove.digest]
-                except docker.errors.NotFound:
+
+                    # We try to remove any dangling docker images
+                    deleted_dangling_entries = self._docker.images.prune(filters={'dangling': True})
```
candicegjing: I added this part to remove any dangling images which might be helpful a little bit?

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -102,13 +104,20 @@ def _cleanup(self):
...
+                    if deleted_dangling_entries['ImagesDeleted'] is not None:
+                        for deleted_entry in deleted_dangling_entries['ImagesDeleted']:
+                            if deleted_entry['Deleted'] in self._image_cache:
+                                # if we successfully removed the dangling image which exists in cache entry,
+                                # we should also remove its cache entry
+                                del self._image_cache[deleted_entry['Deleted']]
+                                deletable_entries.remove(deleted_entry['Deleted'])
+                except docker.errors.ImageNotFound:
```
candicegjing, line 110: `NotFound` exception has been changed to `ImageNotFound` lately.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -102,13 +104,20 @@ def _cleanup(self):
...
                     # image doesn't exist anymore for some reason, stop tracking it
                     del self._image_cache[entry_to_remove.digest]
                 except docker.errors.APIError as err:
```
candicegjing, line 113: Also, forcing removing a running docker image won't give us any benefit, as it only removes the tag of the running image, not the image itself. So when `APIError` happens, I guess we can just keep retrying?

codalab/worker/local_run/docker_image_manager.py
bkgoksel: Neat!

codalab/worker/local_run/docker_image_manager.py
bkgoksel: I think dangling images are mainly caused by locally building images and not pulling finished images from Dockerhub like we do, so I don't think this will make a big difference.

codalab/worker/docker_utils.py
```diff
@@ -220,7 +220,7 @@ def container_exists(container):
     try:
         client.containers.get(container.id)
         return True
-    except docker.errors.NotFound:
+    except docker.errors.ImageNotFound:
```
bkgoksel: This is for finding a `Container` not an `Image`. Can you double check which error trying to `get` a nonexistent container throws?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -154,6 +155,24 @@ def _process_dependencies(self):
...
+        # Find the dependencies to remove
+        dependencies_to_remove = dependencies - valid_dependencies
+        # Remove all the invalid dependency paths from local file system
+        for dep in dependencies_to_remove:
+            remove_path(os.path.join(self.dependencies_dir, dep))
+        # Remove all the invalid dependencies from self._dependency_locks and self._dependencies
+        for dep in self._dependency_locks.keys():
+            if dep not in valid_dependencies:
+                del self._dependency_locks[dep]
+                del self._dependencies[dep]
```
bkgoksel: We may have a `remove_dependency` helper in here that does this. If so, better to use that.

codalab/worker/local_run/local_run_manager.py
```diff
@@ -105,7 +105,7 @@ def load_state(self):
                     run_state = run_state._replace(
                         container=self._docker.containers.get(run_state.container_id)
                     )
-                except docker.errors.NotFound as ex:
+                except docker.errors.ImageNotFound as ex:
```
bkgoksel: Again, double check that this is a `Container` `get`, not `Image`

codalab/worker/local_run/local_run_manager.py
```diff
@@ -184,7 +184,7 @@ def process_runs(self):
                 try:
                     container = self._docker.containers.get(container_id)
                     container.remove(force=True)
-                except (docker.errors.NotFound, docker.errors.NullResource):
+                except (docker.errors.ImageNotFound, docker.errors.NullResource):
```
bkgoksel: Same as above

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -127,6 +127,7 @@ def _load_state(self):
 
     def start(self):
         logger.info('Starting local dependency manager')
+        self._sync_dependency_state()
```
bkgoksel: A partial version of this logic is implemented on lines 100-120ish of this file. Can you maybe merge that logic into your function and call this function there? (It might be good to log deleted entries etc like we do on lines 100-120 right now)

bkgoksel (changes requested):
Should be better without the `prune` call. Might be a good idea to get rid of deleting dangling images just because we shouldn't be creating them in the first place, and probably shouldn't mess with host docker environment too much.

Besides that, all requested changes are quite localized and small scoped.

codalab/worker/local_run/local_run_manager.py
candicegjing: Fixed.

codalab/worker/local_run/local_run_manager.py
candicegjing: Fixed.

candicegjing:
@bkgoksel I ended up adding all the syncing dependencies logic in `_load_state()`. Please take a look and let me know if I need to change anything.

bkgoksel (approved):
LGTM now!

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -91,8 +92,9 @@ def _cleanup(self):
             but because of (1) we don't want to use that.
         """
         while not self._stop:
+            # We sort the image cache in LRU order
```
percyliang: nit: simply write in the imperative form: 'Sort the ...'

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -102,13 +104,10 @@ def _cleanup(self):
                     entry_to_remove.digest,
                 )
                 try:
-                    image_to_delete = self._docker.images.get(entry_to_remove.id)
-                    tags_to_delete = image_to_delete.tags
-                    for tag in tags_to_delete:
-                        self._docker.images.remove(tag)
-                    # if we successfully removed the image also remove its cache entry
+                    # We try to delete images from image cache in LRU order
```
percyliang: 'Try to ...'

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -91,15 +91,18 @@ def _load_state(self):
...
         paths = set()
-        for dep_key, dep_state in state['dependencies'].items():
-            full_path = os.path.join(self.dependencies_dir, dep_state.path)
-            if os.path.exists(full_path):
-                dependencies[dep_key] = dep_state
-                dependency_locks[dep_key] = threading.RLock()
+
+        # List all the local directories under self.dependencies_dir
+        local_directories = set(os.listdir(self.dependencies_dir))
+        # Filter in all the dependencies that exist in local file system
```
percyliang: "Filter by"

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -102,13 +104,10 @@ def _cleanup(self):
...
+                    self._docker.images.remove(entry_to_remove.id)
```
percyliang, line 108: Can you explain in this PR what this change does?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -91,15 +91,18 @@ def _load_state(self):
...
         dependency_locks = {}
         paths = set()
-        for dep_key, dep_state in state['dependencies'].items():
-            full_path = os.path.join(self.dependencies_dir, dep_state.path)
-            if os.path.exists(full_path):
-                dependencies[dep_key] = dep_state
-                dependency_locks[dep_key] = threading.RLock()
+
+        # List all the local directories under self.dependencies_dir
+        local_directories = set(os.listdir(self.dependencies_dir))
```
percyliang: Provide a PR comment on what this change does

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -109,19 +112,32 @@ def _load_state(self):
...
-            if os.path.exists(full_path):
-                paths.add(path)
-            else:
-                logger.info(
-                    "Path {} in loaded state but doesn't exist in the filesystem".format(full_path)
+
+        state_paths = state['paths']
+        # Find out the valid loaded paths by intersecting them with local_directories.
+        paths = state_paths.intersection(local_directories)
+        # Filter in those invalid loaded paths and generate logging message
```
percyliang: Make this comment clearer: "Remove invalid loaded paths..."?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -109,19 +112,32 @@ def _load_state(self):
...
+        invalid_paths = state_paths - paths
+        for path in invalid_paths:
+            logger.info(
+                "Path {} in loaded state but doesn't exist in the filesystem".format(
+                    os.path.join(self.dependencies_dir, path)
                 )
+            )
+        # Remove the redundant dependencies from local file system
```
percyliang: These aren't 'redundant' right?  These are 'orphaned' (not accounted for in dependencies)?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -91,15 +91,18 @@ def _load_state(self):
         dependencies = {}
         dependency_locks = {}
         paths = set()
-        for dep_key, dep_state in state['dependencies'].items():
-            full_path = os.path.join(self.dependencies_dir, dep_state.path)
-            if os.path.exists(full_path):
-                dependencies[dep_key] = dep_state
-                dependency_locks[dep_key] = threading.RLock()
+
```
percyliang, line 100: This function now is doing more complicated things like cleaning up directories...Can you write method level documentation describing what the logic is.  Ideally, we would be able to separate the loading logic from the cleaning up logic so that this method doesn't get too complicated to reason about.  But probably not worth doing that refactor now.  But we should be clear about what the invariants of the system are - what is the relationship between the local directory and the dependencies file?

percyliang (changes requested):

codalab/worker/local_run/local_dependency_manager.py
candicegjing: I initially wanted to add a separate function to clean up orphaned directories, but once I read through `_load_state()`, I noticed that there are steps that could be combined and make the code more efficient. 
I tried to keep the same structure of the original logic and only changed the directory lookup by checking `local_directories`(`/mnt/scratch/bundles/`) instead of calling `os.path.join` all the times. Even though this is trivial, but I think it saves a bit of time. Line 107-136 is the same as original logic. Line 137-142 is to remove orphaned local dependency directories. Line 149-150 is to save the final synced state file back to `/mnt/scratch/bundles`.

codalab/worker/local_run/local_dependency_manager.py
candicegjing: This change will add all the local directories that exist under `self.dependencies_dir`. In real system, it will be all the directories under `/mnt/scratch/bundles/`.

codalab/worker/local_run/docker_image_manager.py
candicegjing: The original code tries to delete the image by looping through and deleting all the tags associated with the image one by one. I found a simpler API when I read through docker's rest api documentation: given that we already have `image.id` in the variable `entry_to_remove`, we can directly remove the image using `image.id`.

codalab/worker/local_run/local_dependency_manager.py
percyliang: Right, but can you contrast this with the old logic?

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -80,7 +80,8 @@ def _cleanup(self):
         """
         Prunes the image cache for runs.
         1. Only care about images we (this DockerImageManager) downloaded and know about
-        2. We use sum of VirtualSize's, which is an upper bound on the disk use of our images:
+        2. We also try to prune any dangling docker images on the system
```
percyliang: add period

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,83 @@ def _save_state(self):
             self._state_committer.commit({'dependencies': self._dependencies, 'paths': self._paths})
 
     def _load_state(self):
+        """
+        Load states from dependencies-state.json file and sync them with local directories specified by self.dependencies_dir.
+        Populate values for self._dependencies, self._dependency_locks and self._paths by
```
percyliang: replace " by" with "as follows:"

codalab/worker/local_run/local_dependency_manager.py
percyliang: In the future, put the comments directly on the line ranges now that GitHub supports multiline comments.

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,83 @@ def _save_state(self):
...
             self._dependency_locks = dependency_locks
             self._paths = paths
+
         logger.info(
             '{} dependencies, {} paths in cache.'.format(len(self._dependencies), len(self._paths))
         )
 
+    def _sync_state(self):
+        """
+        Sync states between loaded paths and local file system by removing any paths that doesn't exist in
```
percyliang: I think it's clearer to write: Remove any path that exists in the local file system but not in `dependency-state.json` file.

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,83 @@ def _save_state(self):
...
+        # All the local directories under self.dependencies_dir
+        local_directories = set(os.listdir(self.dependencies_dir))
+        # Remove the orphaned dependencies from local file system
+        directories_to_remove = local_directories - self._paths
+        for dir in directories_to_remove:
+            full_path = os.path.join(self.dependencies_dir, dir)
+            logger.info("Remove {} from the filesystem".format(full_path))
+            remove_path(full_path)
+        # Save the current state back to the state file: dependency-state.json
+        self._save_state()
```
percyliang, line 161: Why do we need to save again?  It seems like we just deleted things from the local disk.

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,83 @@ def _save_state(self):
...
                 )
-        for path in state['paths']:
-            full_path = os.path.join(self.dependencies_dir, path)
-            if os.path.exists(full_path):
-                paths.add(path)
-            else:
-                logger.info(
-                    "Path {} in loaded state but doesn't exist in the filesystem".format(full_path)
+
+        state_paths = state['paths']
```
percyliang: This logic seems overlapping with the `_sync_state` - merge

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,83 @@ def _save_state(self):
             self._state_committer.commit({'dependencies': self._dependencies, 'paths': self._paths})
 
     def _load_state(self):
+        """
+        Load states from dependencies-state.json file and sync them with local directories specified by self.dependencies_dir.
+        Populate values for self._dependencies, self._dependency_locks and self._paths by
+        1. Load states from dependencies-state.json file and local directory self.dependencies_dir.
```
percyliang: 'states' is used in a vague way here...be more precise (but succinct) about what is loaded.  In general, this description seems like it's overly complicated.  I think it'd be better to describe the purpose of this function rather than a step by step procedure - what invariants are maintained?  I'd write something like:

There are two pieces of information:
(1) dependencies-state.json: contains bundles with some information (e.g., size, last_used)
(2) local disk: the bundle contents
Ideally, these would be in sync, but life happens and we want to robust to them not being in sync.  Therefore, when we load the state, we keep only the bundles that appear in both (1) and (2).

percyliang (changes requested):
I think this logic could be simplified and made clearer.  I'd move a lot of the logic out of `_load_state` so all it does is load the state from dependencies-state.json.  Then `_sync_state` should do the intersection logic (see comment that I put at the top of `_load_state`).  That way, you separate out the part that mutates from the part that doesn't mutate (which can be easily changed with a flag to not mutate and only print out warnings if we want).

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
             self._state_committer.commit({'dependencies': self._dependencies, 'paths': self._paths})
 
     def _load_state(self):
+        """
+        Load states from dependencies-state.json file and populate values for self._dependencies,
+        self._dependency_locks and self._paths as follows:
+        1. dependencies-state.json: contains bundles with other information (e.g., stage, dependency, last_used, etc)
+        2. directories in local file system: the bundle contents
+        Ideally, 1. and 2. would be in sync, but life happens and we want to robust to them not being in sync.
+        Therefore, when we load the state from json file, we keep only the bundles that appear in both (1) and (2).
```
percyliang: This comment is not accurate since this function doesn't actually intersect (1) and (2).

Also, be consistent (1. versus (1)).

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
+        :return:
```
percyliang: Return what?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
         logger.info(
-            '{} dependencies, {} paths in cache.'.format(len(self._dependencies), len(self._paths))
+            'Loaded {} dependencies, {} paths from cache.'.format(
+                len(self._dependencies), len(self._paths)
+            )
+        )
+
+    def _sync_state(self):
+        """
+        Remove any path that exists in the local file system but not in dependency-state.json file and
```
percyliang: The comment about the two states from `_load_state` should be moved here.

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
+        save the synchronized dependency state and loaded paths back to dependency-state.json file
+        :return:
```
percyliang: Return what?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
+        # self._dependency_locks if they don't exist in local file system
+        dependencies_to_remove = [
+            dep
+            for dep, dep_state in self._dependencies.items()
+            if dep_state.path not in local_directories
+        ]
+        for dep in dependencies_to_remove:
+            logger.info(
+                "Dependency {} in loaded state but its path {} doesn't exist in the file system. "
+                "Remove it from loaded state.".format(
```
percyliang: `Removing it`

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
+        # Remove the orphaned dependencies from self._dependencies and
+        # self._dependency_locks if they don't exist in local file system
+        dependencies_to_remove = [
+            dep
+            for dep, dep_state in self._dependencies.items()
+            if dep_state.path not in local_directories
+        ]
+        for dep in dependencies_to_remove:
+            logger.info(
+                "Dependency {} in loaded state but its path {} doesn't exist in the file system. "
```
percyliang: Replace 'loaded state' with 'dependency state' (this comment applies more broadly)

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
+        # Get the paths that exist in loaded state, loaded path and local file system
+        self._paths = self._paths.intersection(paths_in_loaded_state).intersection(
+            local_directories
         )
 
+        # Remove the orphaned dependencies from local file system
+        directories_to_remove = local_directories - self._paths
+        for dir in directories_to_remove:
+            full_path = os.path.join(self.dependencies_dir, dir)
+            logger.info("Remove orphaned directory {} from the file system.".format(full_path))
```
percyliang: 'local file system' (be consistent)

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,79 @@ def _save_state(self):
...
+            logger.info(
+                "Dependency {} in loaded state but its path {} doesn't exist in the file system. "
+                "Remove it from loaded state.".format(
+                    dep, os.path.join(self.dependencies_dir, self._dependencies[dep].path)
+                )
+            )
+            del self._dependencies[dep]
+            del self._dependency_locks[dep]
+
+        # Get all the paths from loaded state
```
percyliang: Just replace with one comment that summarizes this block

percyliang (changes requested):
Logic looks good!  Just a few comments on the comments...

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -78,6 +78,8 @@ def __init__(self, commit_file, bundle_service, worker_dir, max_cache_size_bytes
         # DependencyKey -> WorkerThread(thread, success, failure_message)
         self._downloading = ThreadDict(fields={'success': False, 'failure_message': None})
         self._load_state()
+        # Sync states between dependency-state.json and dependency directories in local file system.
```
percyliang: "on the local file system"

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,77 @@ def _save_state(self):
             self._state_committer.commit({'dependencies': self._dependencies, 'paths': self._paths})
 
     def _load_state(self):
+        """
+        Load states from dependencies-state.json file that contains bundles with other information (e.g., stage,
```
percyliang: "Load from dependencies-state.json, which contains information about bundles (e.g., state, dependencies, last used, etc.)"

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,77 @@ def _save_state(self):
...
+        dependency, last_used, etc.) and populates values for self._dependencies, self._dependency_locks, and self._paths
+        """
         state = self._state_committer.load(default={'dependencies': {}, 'paths': set()})
+
         dependencies = {}
```
percyliang, line 98: These variables don't seem to be used, remove?

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -87,45 +89,77 @@ def _save_state(self):
...
+            )
         )
 
+    def _sync_state(self):
+        """
+        Synchronize dependency states between dependencies-state.json and the local file system as follows:
+        1. self._dependencies, self._dependency_locks, and self._paths: populated from dependencies-state.json in function _load_state()
+        2. directories in the local file system: the bundle contents
+        Ideally, 1. and 2. would be in sync, but life happens and we want to be robust to them when they are not in sync.
+        Therefore, when we load the state from json file, we keep only the bundles that appear in both 1 and 2.
```
percyliang: "This function forces the two to be in sync by taking the intersection (e.g., deleting bundles from the local file system that don't appear in the dependencies-state.json and vice-versa)"

percyliang (approved):
Just a few more improvements to the comments...

codalab/worker/local_run/local_dependency_manager.py
candicegjing: They are used in the next few lines.

PR #1503, opened 2019-10-15
Fix worksheet insert

### LINKED ISSUE #1462: Make sure worksheet editing works

epicfaace:
Fixes #1462. Removing `* 2 ` did the trick. (it wasn't as simple to figure out as it seems haha)

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1287,6 +1287,10 @@ class BundleModel(object):
             worksheet.id = result.lastrowid
 
     def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replace=[]):
+        """
+        Add worksheet items *items* to the position *after_sort_key* to the worksheet,
+        removing items specified by *replace* if necessary.
+        """
         with self.engine.begin() as connection:
             if len(replace) > 0:
                 # Remove the old items.
@@ -1298,7 +1302,7 @@ class BundleModel(object):
                 return
             if after_sort_key is not None:
                 after_sort_key = int(after_sort_key)
-                # Shift existing items' sort_keys for items that original came after
+                # Shift existing items' sort_keys for items that originally came after
                 # the after_sort_key
                 offset = len(items)
                 clause = and_(
@@ -1329,13 +1333,13 @@ class BundleModel(object):
                             'subworksheet_uuid': item.subworksheet_uuid,
                             'value': item.value,
                             'type': item.type,
-                            'sort_key': item_sort_key(item) * 2 + offset,
+                            'sort_key': item_sort_key(item) + offset,
                         }
                         for item in after_items
                     ]
                     self.do_multirow_insert(connection, cl_worksheet_item, new_after_items)
             # Insert new items
-            items_2_insert = [
+            items_to_insert = [
                 {
                     'worksheet_uuid': worksheet_uuid,
                     'bundle_uuid': bundle_uuid,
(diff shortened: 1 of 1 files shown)


codalab/model/bundle_model.py
```diff
@@ -1346,7 +1350,7 @@ def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replac
                 }
                 for idx, (bundle_uuid, subworksheet_uuid, value, type) in enumerate(items)
             ]
-            self.do_multirow_insert(connection, cl_worksheet_item, items_2_insert)
+            self.do_multirow_insert(connection, cl_worksheet_item, items_to_insert)
```
percyliang, line 1353: Worries me a bit that the delete above and this insert aren't atomic.

codalab/model/bundle_model.py
```diff
@@ -1329,13 +1333,13 @@ def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replac
                             'subworksheet_uuid': item.subworksheet_uuid,
                             'value': item.value,
                             'type': item.type,
-                            'sort_key': item_sort_key(item) * 2 + offset,
+                            'sort_key': item_sort_key(item) + offset,
```
percyliang, line 1336: Was this a bug?  Can you explain what goes wrong / how things are fixed?

percyliang (approved):

codalab/model/bundle_model.py
epicfaace: I agree; probably belongs in another issue though because we don't make multiple insertions atomic anyway. I've made https://github.com/codalab/codalab-worksheets/issues/1504

codalab/model/bundle_model.py
epicfaace: This line of code moves the items _after_ the location where you're inserting down by `offset`, by adding `offset` to the sort key. However, the previous code was first doubling the sort key and then adding `offset` to it.

I think the new code gets the logic right; not sure why one would want to double the sort key as in the old code. I'm not exactly sure how doubling it led to the completely unexpected behavior documented in #1462, though.

yipenghe: Looking from the if clause in the code, 1323 (in commit) or 1319(in original), 
the condition min(item_sort_key(item) for item in after_items) - after_sort_key <= offset indicates that the distance between the sort_key of the original after_items and after_sort_key is less than offset, which means there aren't enough integer for sort_keys between min(all after_items_sort_key) and after_sort_key, I think that's the original reason why the code doubled item_sort_key, which seems unreasonable

yipenghe: <s>I not sure about just changing the sort_key to sort_key + offset, since the if condition means sort_key + offset < after_sort_key + len(items), which sounds like the new sort key would overlap with some of the inserted items' sort key, not sure if this will cause any new problem</s>

yipenghe: Actually my original thoughts were wrong, if we look at the error comment example in the issue #1462 we see  : 

{"items":["## **My running bundles**","These are bundles that are currently running or queued to be run.","a","b","d","c",""],"after_sort_key":-28,"ids":[676,677]}

sort keys can be assigned negative numbers, which is why *2 can mess up the ordering, because we place smaller sort_key in front and larger sort_key on the back, if the sort key was assigned negative number, *2 makes it smaller and move it to the front and thus the mess up.

Example: 
row 1: key -27
row 2: key -28
we want to insert an item between row 1 & row 2, the original code will cause the row 2 key be -56 + 1 = -55 and thus will be placed before row 1 after insertion

So the original code was a bug and removing *2 is good.

yipenghe (approved):

codalab/model/bundle_model.py
epicfaace: Great explanation!

PR #1505, opened 2019-10-15
Get rid of dicts in worker

bkgoksel:
Represent state in worker using objects and named tuples instead of dicts to make the code more manageable.


codalab/rest/workers.py
```diff
@@ -37,9 +37,9 @@ def checkin(worker_id):
         request.json["dependencies"],
     )
 
-    for uuid, run in request.json["runs"].items():
+    for run in request.json["runs"]:
```
bkgoksel, line 41: No need for this to be a dict since the uuid is part of the run state.

codalab/worker/bundle_service_client.py
```diff
@@ -21,26 +21,29 @@ def wrapper(*args, **kwargs):
...
+                        raise BundleServiceException(
+                            message
+                            + ': '
+                            + http.client.responses[e.code]
+                            + ' - '
+                            + json.dumps(client_error),
+                            e.code >= 400 and e.code < 500,
+                        )
+                except json.decoder.JSONDecodeError:
+                    raise BundleServiceException(message + ': ' + str(e), False)
```
bkgoksel, line 46: Error 500 codes were causing JSON decode errors in the earlier code block, so this surfaces the actual error from the server instead.

codalab/worker/bundle_state.py
```diff
@@ -33,19 +36,102 @@ class State(object):
...
+DependencyKey = namedtuple('DependencyKey', 'parent_uuid parent_path')
+Dependency = namedtuple(
+    'Dependency', 'parent_name parent_path parent_uuid child_path child_uuid location'
+)
+
+
+class BundleInfo(object):
+    """
+    Defines the bundle info passed to the worker by the server.
+    Refer to codalab/lib/bundle_util.py#bundle_to_bundle_info() for how the
```
bkgoksel, line 52: Maybe this class needs to be more tightly coupled with that function (ie move this class there or move that function here.

codalab/worker/bundle_state.py
```diff
@@ -33,19 +36,102 @@ class State(object):
...
-        self.network = network  # type: bool
+    def __init__(
+        self,
+        cpus,  # type: int
+        gpus,  # type: int
+        docker_image,  # type: str
+        time,  # type: int
+        memory,  # type: int
+        disk,  # type: int
+        network,  # type: bool
```
bkgoksel, line 130: Apparently it's better to type-annotate the args of the init function.

codalab/worker/local_run/local_dependency_manager.py
```diff
@@ -13,11 +13,12 @@
 import codalab.worker.pyjson
 from codalab.worker.worker_thread import ThreadDict
 from codalab.worker.state_committer import JsonStateCommitter
+from codalab.worker.bundle_state import DependencyKey
 
 logger = logging.getLogger(__name__)
 
 DependencyState = namedtuple(
-    'DependencyState', 'stage dependency path size_bytes dependents last_used message killed'
+    'DependencyState', 'stage dependency_key path size_bytes dependents last_used message killed'
```
bkgoksel, line 21: This will be a breaking change for the workers as well.
During next release we should bring down all workers, clear all their caches and bring them up from scratch.

codalab/worker/local_run/local_run_manager.py
```diff
@@ -94,8 +95,13 @@ def create_or_get_network(name, internal):
 
     def save_state(self):
         # Remove complex container objects from state before serializing, these can be retrieved
-        simple_runs = {uuid: state._replace(container=None) for uuid, state in self._runs.items()}
-        self._state_committer.commit(simple_runs)
+        runs = {
+            uuid: state._replace(
+                container=None, bundle=state.bundle.to_dict(), resources=state.resources.to_dict()
```
bkgoksel, line 100: Need to convert these to dicts for serialization.

codalab/worker/worker.py
```diff
@@ -100,7 +101,7 @@ def _checkin(self):
             'free_disk_bytes': self._run_manager.free_disk_bytes,
             'dependencies': self._run_manager.all_dependencies,
             'hostname': socket.gethostname(),
-            'runs': self._run_manager.all_runs,
+            'runs': [run.to_dict() for run in self._run_manager.all_runs],
```
bkgoksel, line 104: Serialize before network request

codalab/worker/worker.py
```diff
@@ -135,6 +136,8 @@ def _run(self, bundle, resources):
         start_message = {'hostname': socket.gethostname(), 'start_time': int(now)}
 
         if self._bundle_service.start_bundle(self.id, bundle['uuid'], start_message):
+            bundle = BundleInfo.from_dict(bundle)
+            resources = RunResources.from_dict(resources)
```
bkgoksel, line 140: Deserialize

codalab/worker/bundle_state.py
```diff
@@ -33,19 +36,102 @@ class State(object):
     FINAL_STATES = {READY, FAILED, KILLED}
 
 
+DependencyKey = namedtuple('DependencyKey', 'parent_uuid parent_path')
```
percyliang, line 41: This almost feels like it should be generic to CodaLab, not just the worker...  Maybe with the exception of `location`...maybe should say what this is...

codalab/worker/local_run/local_dependency_manager.py
percyliang: If we're going to do this, can we put in some mechanism to make workers more robust to changes in version and fail gracefully?  Want to minimize disruption.  Also, somewhere we should make a note of these breaking changes so that the next person who deploys is aware.  They are probably not going to see this comment.

percyliang (approved):

codalab/worker/local_run/local_dependency_manager.py
bkgoksel: Yeah, that's a good thing to add. Can someone else from the Codalab team pick up the task (see if the server already responds with the version number to worker checkin and if so get the worker to log a warning that it's out of date). Again I can guide with that. Of course it won't help people with old workers who don't have the warning code, but hopefully there aren't too many personal cl-workers out there.

As for release notes, it may be a good idea to make a PR template like we do at SM, and then release people can just go over the PR notes of all the commits merged since last release to compile this stuff.

codalab/worker/bundle_state.py
bkgoksel: Will add comments explaining for now. I agree that there's more room for canonicalizing these across Codalab, but I don't have the bandwidth to undertake it all at the moment, and once the object definitions are in, it is less work/breaking changes to move them around and use them in more places.

PR #1506, opened 2019-10-15
Allow adding items to empty worksheets

### LINKED ISSUE #1480: "Text" button cannot be launched in new, empty worksheets
**Describe the bug**
In a newly created worksheet, nothing happens when clicking "Text". 

**To Reproduce**
1. Create a new worksheet in the new front-end
2. Click on "Text"
3. Nothing happens

**Expected behavior**
Clicking "Text" would launch a new text box. 

**Screenshots**
http://g.recordit.co/MieNqsCTOC.gif

epicfaace:
Fixes #1480.

This is done by rendering an item on the frontend
with the `isDummyItem` attribute equal to true.

This PR also:
- Hides the text editor upon pressing the "Escape" key
- Adds an item to the end of the worksheet if no item is selected

image

### DIFF
--- a/frontend/src/components/worksheets/WorksheetItemList.js
+++ b/frontend/src/components/worksheets/WorksheetItemList.js
@@ -208,7 +208,9 @@ class WorksheetItemList extends React.Component {
             var worksheet_items = [];
             info.items.forEach(
                 function(item, index) {
-                    var focused = index === this.props.focusIndex;
+                    var focused = (index === this.props.focusIndex) ||
+                        // If nothing is focused, append to the end by default.
+                        (this.props.focusIndex === -1 && index === info.items.length - 1);
                     var props = {
                         worksheetUUID: info.uuid,
                         item: item,
--- a/frontend/src/components/worksheets/WorksheetItemList.js
+++ b/frontend/src/components/worksheets/WorksheetItemList.js
@@ -211,7 +211,7 @@ class WorksheetItemList extends React.Component {
             var worksheet_items = [];
             info.items.forEach(
                 function(item, index) {
-                    var focused = (index === this.props.focusIndex) ||
+                    const focused = (index === this.props.focusIndex) ||
                         // If nothing is focused, append to the end by default.
                         (this.props.focusIndex === -1 && index === info.items.length - 1);
                     var props = {
--- a/frontend/src/components/worksheets/WorksheetItemList.js
+++ b/frontend/src/components/worksheets/WorksheetItemList.js
@@ -205,7 +205,20 @@ class WorksheetItemList extends React.Component {
         var items_display;
         var info = this.props.ws.info;
         if (info && info.items.length === 0) { // Create a "dummy" item at the beginning so that only empty text can be added.
-            info.items = [{"isDummyItem": true, "text":"","mode":"markup_block","sort_keys":[-1],"ids":[null],"is_refined":true}];
+            info.items = [
+                {
+                    "isDummyItem": true,
+                    "text": "",
+                    "mode": "markup_block",
+                    "sort_keys": [
+                        -1
+                    ],
+                    "ids": [
+                        null
+                    ],
+                    "is_refined": true
+                }
+            ];
         }
         if (info && info.items.length > 0) {
             var worksheet_items = [];
--- a/frontend/src/components/worksheets/WorksheetItemList.js
+++ b/frontend/src/components/worksheets/WorksheetItemList.js
@@ -204,6 +204,9 @@ class WorksheetItemList extends React.Component {
         // Create items
         var items_display;
         var info = this.props.ws.info;
+        if (info && info.items.length === 0) { // Create a "dummy" item at the beginning so that only empty text can be added.
+            info.items = [{"isDummyItem": true, "text":"","mode":"markup_block","sort_keys":[-1],"ids":[null],"is_refined":true}];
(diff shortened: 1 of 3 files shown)


frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -204,11 +204,16 @@ class WorksheetItemList extends React.Component {
         // Create items
         var items_display;
         var info = this.props.ws.info;
+        if (info && info.items.length === 0) { // Create a "dummy" item at the beginning so that only empty text can be added.
+            info.items = [{"isDummyItem": true, "text":"","mode":"markup_block","sort_keys":[-1],"ids":[null],"is_refined":true}];
```
percyliang: Make spacing consistent, also might be easier to read with multiple lines.

frontend/src/components/worksheets/items/ItemWrapper.js
```diff
@@ -59,11 +59,15 @@ class ItemWrapper extends React.Component {
             isWorkSheetItem = false;
         }
 
+        const {isDummyItem} = item;
```
percyliang, line 62: these two variables don't seem like they type check

frontend/src/components/worksheets/items/ItemWrapper.js
```diff
@@ -59,11 +59,15 @@ class ItemWrapper extends React.Component {
...
+
         return (
             <div
-                className={classes.container}
+                className={isDummyItem ? "": classes.container}
```
percyliang, line 66: what's the standard here - `null`?

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -204,11 +204,16 @@ class WorksheetItemList extends React.Component {
...
+        }
         if (info && info.items.length > 0) {
             var worksheet_items = [];
             info.items.forEach(
                 function(item, index) {
-                    var focused = index === this.props.focusIndex;
+                    var focused = (index === this.props.focusIndex) ||
```
percyliang: try to use `const` as you go

percyliang (changes requested):
Why do we need a dummy element (would be nice to avoid special casing)?  You just need to be able to add a text block to an empty worksheet, no?

epicfaace:
> Why do we need a dummy element (would be nice to avoid special casing)? You just need to be able to add a text block to an empty worksheet, no?

The text / run / upload blocks themselves are actually part of the `ItemWrapper` component -- so they can only show up _in relation to_ an existing item. That's why we need an existing `ItemWrapper` component (which stores dummy data) to get this to work.

frontend/src/components/worksheets/WorksheetItemList.js
epicfaace:
```suggestion
                    const focused = (index === this.props.focusIndex) ||
```

frontend/src/components/worksheets/items/ItemWrapper.js
epicfaace: I don't think there's a standard -- either `null` or `""` work.

frontend/src/components/worksheets/items/ItemWrapper.js
epicfaace: What do you mean? This is the same as saying

`const isDummyItem = item.isDummyItem;`

frontend/src/components/worksheets/WorksheetItemList.js
epicfaace:
```suggestion
            info.items = [
                {
                    "isDummyItem": true,
                    "text": "",
                    "mode": "markup_block",
                    "sort_keys": [
                        -1
                    ],
                    "ids": [
                        null
                    ],
                    "is_refined": true
                }
            ];
```

epicfaace: Updated, although for these things, I was thinking of adding `prettier` to our `lint-staged` step (https://github.com/codalab/codalab-worksheets/blob/frontend/frontend/package.json#L70). This would need to wait until the frontend is merged though so that the diff between master and frontend isn't too great.

frontend/src/components/worksheets/items/ItemWrapper.js
percyliang: Oh, misread, never mind.

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -204,11 +204,29 @@ class WorksheetItemList extends React.Component {
...
+                        -1
+                    ],
+                    "ids": [
+                        null
+                    ],
+                    "is_refined": true
+                }
+            ];
+        }
         if (info && info.items.length > 0) {
```
percyliang, line 223: `info.items.length > 0` should always hold now with the above.

percyliang:
I'm still confused - we only want the upload/run/buttons at the header, not showing up between items (which was confusing, right?).  So I think we can get rid of ItemWrapper actually, no?

epicfaace:
> I'm still confused - we only want the upload/run/buttons at the header, not showing up between items (which was confusing, right?). So I think we can get rid of ItemWrapper actually, no?

No, ItemWrapper doesn't show the actual upload/run/text buttons -- it shows the actual upload screen / run screen / text box screens.

PR #1507, opened 2019-10-16
Make contiguous blocks of text appear on different lines

### LINKED ISSUE #1268: Contiguous blocks of text appear on the same line
- Click "TEXT" and add a single text block.
- Click "TEXT" again and add another text block.
- When you save the second block, both text blocks get "merged" instead of being separate blocks on separate lines, as expected.

image

epicfaace:
Fixes #1268 by adding a newline to the beginning of new text
items, so that they stay separate from the previous item.


yipenghe (approved):
Tested locally, looks good

Issue #781, opened 2017-06-11
Drag and drop reordering for worksheet items


kashizui:
Support drag and drop manipulation of worksheet items. Or at least moving them up and down (with little arrow buttons? I'm no designer)

In general, this is for working towards a more "dynamic" and flexible user experience when using worksheets, which otherwise feels super static.

May require significant code changes, such as overhauling worksheet interpretation, making table schemas separate from worksheet items, etc.

epicfaace:
On hold -- #1509 implemented this, but it feels unintuitive because all worksheet items (such as tables themselves) don't have sort keys and are not draggable.

For bundle rows, we will do https://github.com/codalab/codalab-worksheets/issues/1227#issuecomment-550561598 instead.

percyliang:
Specifically, you will be able to hit 'm' or click 'move' to move all the selected rows to the current cursor position.  This will be more powerful / faster for moving lots of non-adjacent bundles long distances.

percyliang:
We now have cut/copy/paste.

PR #1508, opened 2019-10-16
Upgrade react to 16.8

epicfaace:
This is so that we can use React Hooks. This will allow us to 
use libraries such as [React DnD](https://react-dnd.github.io/react-dnd/about),
which will be useful for #781.


yipenghe (approved):
Looks good

Issue #1436, opened 2019-09-27
[USER ISSUE] Weird characters in worksheet name


robinjia:
**Describe the bug**
This is minor, but it seems like special characters aren't displayed properly in worksheet names in the dashboard.

**Codalab Information**
- *Your username*: robinjia
- *UUID of worksheet where problem occurred*: 0xb843335f02a9443995db2013fbe2e7f4
- *UUID of problem bundle (please don't delete)*: affects the listing of the worksheet 0x9a15a170809f4e2cb7940e1f256dee55 
- *Error message received from Codalab (if applicable)*: The following is displayed:

Know What You Donât Know: Unanswerable Questions for SQuAD (ACL 2018) [squadrun-acl2018]

 - *OS (if on CLI)*:
 - *Codalab CLI Version (if on CLI)*: 
 - *Browser (if on Web)*; Chrome

**To Reproduce**
Steps to reproduce the behavior :
1. Go to https://worksheets.codalab.org/worksheets/0xb843335f02a9443995db2013fbe2e7f4/
**Expected behavior**
Should say "Don't" instead of these weird characters.

**Screenshots**
![image](https://user-images.githubusercontent.com/7267349/65803789-89ebed00-e134-11e9-816d-2159919456a0.png)

**Additional context**

yipenghe:
Hi, is it still displaying like this? Can't see the display myself since the worksheet is your dashboard

robinjia:
Yes, in fact if you just visit this worksheet you'll see the same thing: https://worksheets.codalab.org/worksheets/0x9a15a170809f4e2cb7940e1f256dee55

yipenghe:
Thanks! I will work on it

yipenghe:
Thanks for reporting this, this will be fixed in the new frontend

yipenghe:
It will also be fixed in current frontend in next release #1568

PR #1511, opened 2019-10-16
Allow unicode in title

yipenghe:
#1436 See latest comment from me below
<s>Supporting unicode in title & name for new frontend, the trick is to encode the way special characters get encoded in url and store that to database. 

The root cause of unicode error seems to be related to database, I tried several ways to change the character set of the database, but character-set-server is still latin1. It seems to be tricky when we are trying to change an existing database's character set...</s>
Screenshot from 2019-10-16 11-24-41

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1048,6 +1048,7 @@ class BundleModel(object):
         # Set tags
         for value in worksheet_values.values():
             value['tags'] = []
+            value['title'] = self.decode_str(value['title'])
         for row in tag_rows:
             worksheet_values[row.worksheet_uuid]['tags'].append(row.tag)
         if fetch_items:
@@ -1267,6 +1268,8 @@ class BundleModel(object):
         for row in rows:
             row = str_key_dict(row)
             row['group_permissions'] = uuid_group_permissions[row['uuid']]
+            if row['title']:
+                row['title'] = self.decode_str(row['title'])
             row_dicts.append(row)
 
         return row_dicts
@@ -1444,6 +1447,8 @@ class BundleModel(object):
             worksheet.frozen = info['frozen']
         if 'owner_id' in info:
             worksheet.owner_id = info['owner_id']
+        if 'title' in info:
+            info['title'] = self.encode_str(info['title'])
         worksheet.validate()
         with self.engine.begin() as connection:
             if 'tags' in info:
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1048,7 +1048,8 @@ class BundleModel(object):
         # Set tags
         for value in worksheet_values.values():
             value['tags'] = []
-            value['title'] = self.decode_str(value['title'])
+            if value['title']:
+                value['title'] = self.decode_str(value['title'])
         for row in tag_rows:
             worksheet_values[row.worksheet_uuid]['tags'].append(row.tag)
         if fetch_items:
--- a/codalab/objects/worksheet.py
+++ b/codalab/objects/worksheet.py
@@ -33,7 +33,6 @@ class Worksheet(ORMObject):
         spec_util.check_uuid(self.uuid)
         spec_util.check_name(self.name)
 
-
     def __repr__(self):
         return 'Worksheet(uuid=%r, name=%r)' % (self.uuid, self.name)
 
--- a/codalab/objects/worksheet.py
+++ b/codalab/objects/worksheet.py
@@ -31,6 +31,8 @@ class Worksheet(ORMObject):
         and its name.
         '''
         spec_util.check_uuid(self.uuid)
+        spec_util.check_name(self.name)
+
 
     def __repr__(self):
         return 'Worksheet(uuid=%r, name=%r)' % (self.uuid, self.name)
--- a/codalab/objects/worksheet.py
+++ b/codalab/objects/worksheet.py
@@ -31,7 +31,6 @@ class Worksheet(ORMObject):
         and its name.
         '''
         spec_util.check_uuid(self.uuid)
-        spec_util.check_name(self.name)
 
     def __repr__(self):
         return 'Worksheet(uuid=%r, name=%r)' % (self.uuid, self.name)
--- a/docker/compose_files/files/mysqlconf/mysql.cnf
+++ b/docker/compose_files/files/mysqlconf/mysql.cnf
@@ -1,9 +1,0 @@
-[client]
-default-character-set = utf8mb4
-
-[mysql]
-default-character-set = utf8mb4
-
-[mysqld]
-character_ser_server=utf8mb4
-collation_server=utf8mb4_0900_ai_ci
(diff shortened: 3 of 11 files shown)


codalab/objects/worksheet.py
```diff
@@ -31,7 +31,6 @@ def validate(self):
         and its name.
         '''
         spec_util.check_uuid(self.uuid)
-        spec_util.check_name(self.name)
```
percyliang: Why are these removed?

docker/compose_files/files/mysqlconf/mysql.cnf
```diff
@@ -0,0 +1,9 @@
+[client]
```
percyliang: Do we need this to work?  We have to be careful, because on prod, the mysql server is setup not through docker but is on another server.

frontend/src/components/EditableField.js
```diff
@@ -53,11 +53,10 @@ class EditableFieldBase extends React.Component<{
 
         this.setState({ editing: false });
         event.preventDefault();
-
         $.ajax({
             type: this.props.method,
             url: this.props.url,
-            data: JSON.stringify(this.props.buildPayload(this.state.value)),
+            data: JSON.stringify(this.props.buildPayload(encodeURI(this.state.value))),
```
percyliang: These are not necessarily URIs though, right?

yipenghe: Yeh, should actually be using encodeURIComponent, which is safe to use on strings

docker/compose_files/files/mysqlconf/mysql.cnf
yipenghe: Added accidentally, will remove

codalab/objects/worksheet.py
yipenghe: These check if the name contains ASCII patterns

percyliang: I think the worksheet name should still be restricted (like an identifier).  The title can be free-form.

percyliang (commented):
Just to clarify: only the title should be unicode / arbitrary.  The name (for both bundles and worksheets) should still be constrained.

yipenghe:
> Just to clarify: only the title should be unicode / arbitrary. The name (for both bundles and worksheets) should still be constrained.

Ok, I will update accordingly

epicfaace (changes requested):
I would suggest using the `encode_str` and `decode_str` functions for consistency with how we handle storing worksheet item contents with unicode in them. (Otherwise, we will have two workarounds in place -- for one, we will be using `encode_str` and `decode_str` to store worksheet item content, and two, we will store uri-encoded versions of unicode text for worksheet titles, which is the approach in this PR. It seems better to pick one way of doing this and be consistent, so that in the future, we can more easily change the unicode encoding method altogether if need be.)

yipenghe:
Well I'm definitely much more familiar with all these codes now...
1. removed all the uri-escaping encode
2. use bundle_model level encode&decode str for encoding unicode, this allows both web interface editting & cli editting (using the wedit command)
3. still restrict worksheet name to be ascii characters only & follow name pattern
Screenshot from 2019-10-22 01-44-07
Screenshot from 2019-10-22 01-43-38

codalab/model/bundle_model.py
```diff
@@ -1048,6 +1048,7 @@ def batch_get_worksheets(self, fetch_items, **kwargs):
         # Set tags
         for value in worksheet_values.values():
             value['tags'] = []
+            value['title'] = self.decode_str(value['title'])
```
yipenghe: This takes care of general worksheet fetching

codalab/model/bundle_model.py
```diff
@@ -1267,6 +1268,8 @@ def make_condition(field, value):
         for row in rows:
             row = str_key_dict(row)
             row['group_permissions'] = uuid_group_permissions[row['uuid']]
+            if row['title']:
+                row['title'] = self.decode_str(row['title'])
```
yipenghe, line 1273: This takes care of worksheet fetching through 'wsearch' command line, the None check is needed as some legacy title might be None for some reason (i.e before didn't prevent title being None)

frontend/src/components/EditableField.js
```diff
@@ -85,10 +85,16 @@ class EditableFieldBase extends React.Component<{
         }
     };
 
-    handleChange = (event) => {
+    handleAsciiChange = (event) => {
```
yipenghe, line 88: This change makes EditableField togglable for Ascii restriction

frontend/src/components/EditableField.js
```diff
@@ -85,10 +85,16 @@ class EditableFieldBase extends React.Component<{
...
+        //only ascii
```
percyliang: nit: add space after `//`

frontend/src/components/EditableField.js
```diff
@@ -112,7 +118,7 @@ class EditableFieldBase extends React.Component<{
                         autoFocus
                         value={this.state.value}
                         onBlur={this.onBlur}
-                        onChange={this.handleChange}
+                        onChange={this.props.allowASCII? this.handleFreeChange: this.handleAsciiChange}
```
percyliang: nit: put space before '?' and ':'

percyliang (approved):
Could you add a comment in front of `decode_str` and `encode_str` saying how these functions are meant to be used (needed right now because we want to use unicode but DB stores serialized bytes...).  Currently, these are used only for worksheet title and worksheet lines, right?  I feel like inserting these calls might be a bit error-prone since it's very decentralized.

yipenghe:
> Could you add a comment in front of `decode_str` and `encode_str` saying how these functions are meant to be used (needed right now because we want to use unicode but DB stores serialized bytes...). Currently, these are used only for worksheet title and worksheet lines, right? I feel like inserting these calls might be a bit error-prone since it's very decentralized.

Updated. It seems like the unicode change is failing a test for unicode constraints? Where should I change that test/ can I still merge directly?

percyliang:
`test_cli.py`?  Always fix tests before merging.

yipenghe:
@percyliang Actually, do we want this fix for master as well or is it good to have it released along with new frontend? I can open a new PR for fixing it in master since this is working

percyliang:
Ah, would have been good to merge into master, if it's not too much work.

test_cli.py
```diff
@@ -1675,8 +1675,8 @@ def test(ctx):
     run_command([cl, 'wedit', wuuid, '--title', 'nonunicode'])
     check_contains('nonunicode', run_command([cl, 'print', wuuid]))
 
-    # TODO: enable with Unicode support.
-    run_command([cl, 'wedit', wuuid, '--title', 'fáncy ünicode 你好世界😊'], 1)
+    # unicode in worksheet title
+    run_command([cl, 'wedit', wuuid, '--title', 'fáncy ünicode 你好世界😊'], 0)
     # check_contains('fáncy ünicode 你好世界😊', run_command([cl, 'print']))
```
epicfaace, line 1680: @yipenghe You didn't uncomment out the second part of the test, which actually checks that the title is shown as unicode. Can you uncomment this line and make a PR?

yipenghe: Good catch, will do! Thanks!

PR #1512, opened 2019-10-16
Fix problem in cl detaching last item

### LINKED ISSUE #1467: cl detach didn't work on detaching the last bundle
**Describe the bug**
I was trying to deleting the bundles within this worksheet: https://worksheets.codalab.org/worksheets/0x3a6c93e8a90a46eda43fb682d139645d, where it contains 2 same bundles with uuid `0xb30d93`, which was a bundle owned by another user and has been referenced twice in the current worksheet. I used `cl detach 0xb30d93 -n 1` to remove the first one. This was done successfully. However, when I tried to detach it using `cl detach 0xb30d93` again as there is only 1 bundle left, it failed doing so. I also tried `cl detach 0xb30d93 -n 2` in case it still remember the previous state and `cl detach 0xb30d93 -n 1`. Both of them failed.

**Codalab Information**
- *UUID of worksheet where problem occurred*: https://worksheets.codalab.org/worksheets/0x3a6c93e8a90a46eda43fb682d139645d
- *UUID of problem bundle (please don't delete)*: 0xb30d937a18574073903bb38b382aab03
 - *Browser (if on Web)*; Firefox

**To Reproduce**
1. Go to https://worksheets.codalab.org/worksheets/0x3a6c93e8a90a46eda43fb682d139645d
2. Type `cl detach 0xb30d93` in browser 
3. See if the bundle got deleted.

**Expected behavior**
The bundle didn't get detached from the host worksheet as it is supposed to.

candicegjing:
Fixed  #1467.
After investigating on the issue #1467, we found out that the reason why the very last item in a worksheet cannot be detached is due to the for-loop logic, which will never be executed when there are no elements in `new_items` in function `create_worksheet_items()`.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1354,6 +1354,7 @@ class BundleCLI(object):
                         'worksheet': JsonApiRelationship('worksheets', dest_worksheet_uuid),
                         'bundle': JsonApiRelationship('bundles', source_bundle_uuid),
                     },
+                    params={'uuid': dest_worksheet_uuid},
                 )
             return
 
@@ -1755,7 +1756,9 @@ class BundleCLI(object):
             if not detach:
                 new_items.append(item)
 
-        client.create('worksheet-items', data=new_items, params={'replace': True})
+        client.create(
+            'worksheet-items', data=new_items, params={'replace': True, 'uuid': worksheet_uuid}
+        )
 
     @Commands.command(
         'rm',
@@ -1902,6 +1905,7 @@ class BundleCLI(object):
                     }
                     for bundle in bundles
                 ],
+                params={'uuid': worksheet_uuid},
             )
             worksheet_info = client.fetch('worksheets', worksheet_uuid)
             print(
@@ -2755,6 +2759,7 @@ class BundleCLI(object):
                             'worksheet': JsonApiRelationship('worksheets', dest_worksheet_uuid),
                             'value': item_spec[1:].strip(),
                         },
+                        params={'uuid': dest_worksheet_uuid},
                     )
                 else:
                     dest_client.create(
@@ -2764,6 +2769,7 @@ class BundleCLI(object):
                             'worksheet': JsonApiRelationship('worksheets', dest_worksheet_uuid),
                             'value': item_spec,
                         },
+                        params={'uuid': dest_worksheet_uuid},
                     )
 
         elif args.item_type == 'bundle':
@@ -2804,6 +2810,7 @@ class BundleCLI(object):
                         'worksheet': JsonApiRelationship('worksheets', dest_worksheet_uuid),
                         'subworksheet': JsonApiRelationship('worksheets', subworksheet_uuid),
                     },
+                    params={'uuid': dest_worksheet_uuid},
                 )
 
     @Commands.command(
@@ -3203,7 +3210,11 @@ class BundleCLI(object):
         # Save all items to the destination worksheet
         for item in source_items:
             item['worksheet'] = JsonApiRelationship('worksheets', dest_worksheet_uuid)
-        dest_client.create('worksheet-items', source_items, params={'replace': args.replace})
+        dest_client.create(
+            'worksheet-items',
+            source_items,
+            params={'replace': args.replace, 'uuid': dest_worksheet_uuid},
+        )
 
         # Copy over the bundles
         for item in source_items:
--- a/codalab/rest/worksheets.py
+++ b/codalab/rest/worksheets.py
@@ -13,6 +13,7 @@ from codalab.lib.server_util import (
     decoded_body,
     json_api_include,
     query_get_bool,
+    query_get_type,
     query_get_json_api_include_set,
     query_get_list,
 )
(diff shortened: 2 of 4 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1746,8 +1746,11 @@ def do_detach_command(self, args):
                             detach = True
             if not detach:
                 new_items.append(item)
-
-        client.create('worksheet-items', data=new_items, params={'replace': True})
+        if len(new_items) > 0:
```
candicegjing: `cl detach` last item failed when the `new_items` array is empty. So here I added some logic to handle the rest call when it sends an empty array over.

codalab/rest/worksheets.py
```diff
@@ -211,24 +212,29 @@ def create_worksheet_items():
...
         if replace:
-            # Replace items in the worksheet
-            update_worksheet_items(
-                worksheet_info, [Worksheet.Item.as_tuple(i) for i in items], convert_items=False
-            )
-        else:
-            # Append items to the worksheet
-            for item in items:
-                add_worksheet_item(worksheet_uuid, Worksheet.Item.as_tuple(item))
+            worksheet_uuid = query_get_type(str, 'uuid')
```
candicegjing: When `new_items` array is empty, I added logic here to update the host worksheet with empty data. Please let me know if this behavior is desired.

epicfaace (commented):
Looks good. I think because this is a CLI operation, we can actually add a test in `test_cli.py` to verify this functionality?

codalab/lib/bundle_cli.py
percyliang: I don't understand what `new_items` has to do with passing in `uuid`.

percyliang (changes requested):
Is there a way to do this without special casing the empty `new_items` case?  The fact that the old logic was breaking suggests that there might be a deeper problem with it, so it'd be nice not to form a completely different code path...

codalab/lib/bundle_cli.py
candicegjing: When I was debugging the problem worksheet, I noticed that the `new_items` is always `empty` when it passed to `client.fetch`. Then in `client.fetch` side, there is a for loop on `new_items` to generate `worksheet_to_items`. Since `new_items` is empty, `worksheet_to_items` is also empty, then the logic of checking the `replace` variable will never be executed.
So I tried to split the `replace` logic from original for loop. Then I feel if the request is looking for replacing a worksheet with no data inside, we just need to update the worksheet with empty data. Hence, I pass an empty list and the current worksheet's uuid to `update_worksheet_items` function.

candicegjing:
1. I think another way to investigate this is to check what happened during 
https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1706, as all the calculations after this line are correct.  
2. For handling empty `new_items`, I will need the worksheet uuid, which is not present in the input parameters of the current function call. If we don't have a special case for empty `new_items`, we might always include `uuid` in the function call.

candicegjing:
@percyliang I did several checks on the logic and tried to find a different way to avoid using the special `empty` worksheet logic. However, based on [L1740-L1763](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1740-L1763), it only stores the items that will not be detached in the `new_item` list which will be passed to `client.fetch` later and will be used to refresh the current worksheet. So when it tries to detach the last item, the `new_item` list is empty, then `create_worksheet_items` can't work properly as it doesn't know which worksheet to update as no `uuid` is given. I think this is an edge case that is missing in the original logic. Please let me know how you think.

codalab/lib/bundle_cli.py
percyliang: I still feel like we don't need to branch on `len(new_items) > 0`.  Can't we just always pass the `uuid` in?

codalab/rest/worksheets.py
```diff
@@ -211,24 +212,29 @@ def create_worksheet_items():
...
-
     new_items = WorksheetItemSchema(strict=True, many=True).load(request.json).data
 
-    worksheet_to_items = {}
-    for item in new_items:
-        worksheet_to_items.setdefault(item['worksheet_uuid'], []).append(item)
-
-    for worksheet_uuid, items in worksheet_to_items.items():
-        worksheet_info = get_worksheet_info(worksheet_uuid, fetch_items=True)
+    if len(new_items) > 0:
```
percyliang: I'd try to avoid the branching, because there's nothing special about `new_items` and more code paths makes it harder to maintain.

The reason this is more complicated than it needs to be is because in the current code, each item in `new_items` is associated with the worksheet that it should be added to, rather than having a single `uuid` where all of `new_items` gets added to.  Can we change all the places where `client.create('worksheet-items', ...` is called to obey the logic where we pass in a uuid and a list of items (which themselves don't have the associated uuid).  This should vastly simply things.  (I don't know why we had this more complicated and powerful way of specifying worksheet uuids...)

percyliang (changes requested):

codalab/rest/worksheets.py
candicegjing: Yes, I agree. Always passing the `uuid` in would be a cleaner way to handle detached items. I think I should also check if there is any other function is currently using `create_worksheet_items` and see how they will be affected by this change.

percyliang (changes requested):
Can you add a test for detach (to `test_cli.py`) that stresses the thing that this PR fixes?

Also, can you provide a high-level description of what the issue was and how this PR addresses it?

candicegjing:
> Can you add a test for detach (to `test_cli.py`) that stresses the thing that this PR fixes?
> 
> Also, can you provide a high-level description of what the issue was and how this PR addresses it?

Added!

percyliang (approved):
Looks good...is there a test for `test_cli.py` that makes sense to add?

candicegjing:
> Looks good...is there a test for `test_cli.py` that makes sense to add?

Added test cases in `test_cli.py` to verify the very last item of a worksheet can be detached successfully.

PR #1534, opened 2019-10-17
Enable shared filesystem workers again

bkgoksel:
After the worker rewrite, workers had no longer support for shared filesystem deployment even though server-side support was there.
This PR adds that mode back on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ jobs:
       install:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
-        - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
       script:
         - python3 codalab_service.py test --version ${TRAVIS_BRANCH} default
         - docker stop codalab_worker_1
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,7 @@ jobs:
       services:
         - docker
       env:
-        - CODALAB_SHARED_FILESYTEM=false CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
-        - CODALAB_SHARED_FILESYTEM=true CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
-      python: 3.6
+        - CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
       before_install:
         - sudo rm /usr/local/bin/docker-compose
         - curl -L https://github.com/docker/compose/releases/download/1.23.2/docker-compose-`uname -s`-`uname -m` > docker-compose
@@ -24,6 +22,8 @@ jobs:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
       script:
         - python3 codalab_service.py start --services default monitor test --version ${TRAVIS_BRANCH}
+        - python3 codalab_service.py stop
+        - python3 codalab_service.py start --services default monitor test --version ${TRAVIS_BRANCH} --shared-file-system
     - stage: deploy
       script: echo "Deploying"
       language: python
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ jobs:
         - docker
       env:
         - CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
+      python: 3.6
       before_install:
         - sudo rm /usr/local/bin/docker-compose
         - curl -L https://github.com/docker/compose/releases/download/1.23.2/docker-compose-`uname -s`-`uname -m` > docker-compose
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ jobs:
         - chmod +x docker-compose
         - sudo mv docker-compose /usr/local/bin
         - sudo service mysql stop
+        - pip install psutil
       install:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ jobs:
       services:
         - docker
       env:
-        - CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
+        - CODALAB_SHARED_FILESYTEM=false CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
+        - CODALAB_SHARED_FILESYTEM=true CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
       python: 3.6
       before_install:
         - sudo rm /usr/local/bin/docker-compose
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ jobs:
         - chmod +x docker-compose
         - sudo mv docker-compose /usr/local/bin
         - sudo service mysql stop
-        - pip install psutil
       install:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
       script:
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -176,36 +176,6 @@ class UploadManager(object):
         path_util.rename(child_path, path)
         path_util.remove(temp_path)
 
-    def update_metadata_and_save(self, bundle, enforce_disk_quota=False):
-        """
-        Updates the metadata about the contents of the bundle, including
-        data_size as well as the total amount of disk used by the user.
-
-        If |new_bundle| is True, saves the bundle as a new bundle. Otherwise,
-        updates it.
-        """
-        bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
-
-        dirs_and_files = None
-        if os.path.isdir(bundle_path):
-            dirs_and_files = path_util.recursive_ls(bundle_path)
-        else:
-            dirs_and_files = [], [bundle_path]
-
-        data_hash = '0x%s' % (path_util.hash_directory(bundle_path, dirs_and_files))
-        data_size = path_util.get_size(bundle_path, dirs_and_files)
-        if enforce_disk_quota:
-            disk_left = self._bundle_model.get_user_disk_quota_left(bundle.owner_id)
-            if data_size > disk_left:
-                raise UsageError(
-                    "Can't save bundle, bundle size %s greater than user's disk quota left: %s"
-                    % (data_size, disk_left)
-                )
-
-        bundle_update = {'data_hash': data_hash, 'metadata': {'data_size': data_size}}
-        self._bundle_model.update_bundle(bundle, bundle_update)
-        self._bundle_model.update_user_disk_used(bundle.owner_id)
-
     def has_contents(self, bundle):
         return os.path.exists(self._bundle_store.get_bundle_location(bundle.uuid))
 
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -826,15 +826,14 @@ class BundleModel(object):
             self.update_bundle(bundle, bundle_update, connection)
         return True
 
-    def transition_bundle_finalizing(
-        self, bundle, bundle_location, user_id, worker_run, connection
-    ):
+    def transition_bundle_finalizing(self, bundle, user_id, worker_run, connection):
         """
         Transitions bundle to FINALIZING state:
             Saves the failure message and exit code from the worker
             If the user running the bundle was the CodaLab root user,
             increments the time used by the bundle owner.
         """
+
         failure_message, exitcode = worker_run.failure_message, worker_run.exitcode
         if failure_message is None and exitcode is not None and exitcode != 0:
             failure_message = 'Exit code %d' % exitcode
@@ -852,18 +851,9 @@ class BundleModel(object):
         if user_id == self.root_user_id:
             self.increment_user_time_used(bundle.owner_id, getattr(bundle.metadata, 'time', 0))
 
-        worker = self.get_bundle_worker(bundle.uuid)
-        if worker['shared_file_system']:
-            # Update metadata and save loops over each file in the bundle to calculate size on disk
-            # so it might potentially take a very long time, so we just fire this thread and respond
-            metadata_update_thread = threading.Thread(
-                self.update_metadata_and_save, args=(bundle, bundle_location)
-            )
-            metadata_update_thread.start()
-
         return True
 
-    def transition_bundle_finished(self, bundle):
+    def transition_bundle_finished(self, bundle, bundle_location):
         """
         Transitions bundle to READY or FAILED state:
             The final state is determined by whether a failure message or exitcode
@@ -875,6 +865,10 @@ class BundleModel(object):
         state = State.FAILED if failure_message or exitcode else State.READY
         if failure_message == 'Kill requested':
             state = State.KILLED
+
+        worker = self.get_bundle_worker(bundle.uuid)
+        if worker['shared_file_system']:
+            self.update_metadata_and_save(bundle, bundle_location)
 
         metadata = {'run_status': 'Finished', 'last_updated': int(time.time())}
 
@@ -909,7 +903,7 @@ class BundleModel(object):
         self.update_bundle(bundle, bundle_update)
         self.update_user_disk_used(bundle.owner_id)
 
-    def bundle_checkin(self, bundle, worker_run, user_id, worker_id, bundle_location):
+    def bundle_checkin(self, bundle, worker_run, user_id, worker_id):
         """
         Updates the database tables with the most recent bundle information from worker
         """
@@ -927,7 +921,7 @@ class BundleModel(object):
                     bundle, worker_run, row, user_id, worker_id, connection
                 )
                 return self.transition_bundle_finalizing(
-                    bundle, bundle_location, user_id, worker_run, connection
+                    bundle, user_id, worker_run, connection
                 )
 
             if worker_run.state in [State.PREPARING, State.RUNNING]:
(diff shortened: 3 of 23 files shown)


codalab/worker/local_run/local_run_manager.py
```diff
@@ -364,6 +373,8 @@ def all_dependencies(self):
         """
         Returns a list of all dependencies available in this RunManager
         """
+        if self._shared_file_system:
+            return []
```
bkgoksel, line 387: Should probably add a comment here

codalab/worker/local_run/local_run_state.py
```diff
@@ -422,15 +455,16 @@ def _transition_from_CLEANING_UP(self, run_state):
...
+                self.dependency_manager.release(run_state.bundle.uuid, dep_key)
 
             child_path = os.path.join(run_state.bundle_path, dep.child_path)
             try:
                 remove_path(child_path)
             except Exception:
                 logger.error(traceback.format_exc())
 
-        if run_state.has_contents:
+        if not self.shared_file_system and run_state.has_contents:
```
bkgoksel, line 461: Maybe add a comment here that if filesystem is shared, no need to upload results.

codalab/worker/main.py
```diff
@@ -108,9 +108,7 @@ def main():
     parser.add_argument(
         '--shared-file-system',
         action='store_true',
-        help='Internal use: Whether the file system containing '
-        'bundle data is shared between the bundle service '
-        'and the worker.',
+        help='Internal use: If specified, the worker expects to find bundles mounted on its filesystem.',
```
bkgoksel: Accidentally deleted and rewrote this help string, can probably stash this change.

bkgoksel (commented):
This is a start! I'm testing this right now on the Semantic Machines (test) deployment.
It might be a good idea to also add tests that test the shared filesystem deployment. It should not be _too_ hard. Probably need to spin up another codalab instance this time with shared filesystem workers and rerun worker-related tests.

codalab/worker/local_run/local_run_state.py
```diff
@@ -209,30 +212,42 @@ def _transition_from_PREPARING(self, run_state):
 
         # All dependencies ready! Set up directories, symlinks and container. Start container.
         # 1) Set up a directory to store the bundle.
-        remove_path(run_state.bundle_path)
-        os.mkdir(run_state.bundle_path)
+        if self.shared_file_system:
+            try:
+                self._wait_for_bundle_folder(run_state.bundle_path)
```
bkgoksel: This needs to be async/not-blocking with a proper timeout mechanism

codalab/worker/local_run/local_run_manager.py
```diff
@@ -126,7 +130,8 @@ def start(self):
         """
         self.load_state()
         self._image_manager.start()
-        self._dependency_manager.start()
+        if not self._shared_file_system:
+            self._dependency_manager.start()
```
fabeschan, line 136: does it make sense to still have a dependency manager on shared filesystem? Say, if the filesystem has a size limit for example and some cleanup process is needed to not exceed it

fabeschan: just realized shared fs means shared among not only workers but the server as well, so cleanup doesn't make sense there

do you think there'll be a use for a fs shared only among workers?

codalab/worker/local_run/local_run_state.py
```diff
@@ -209,30 +212,42 @@ def _transition_from_PREPARING(self, run_state):
 
         # All dependencies ready! Set up directories, symlinks and container. Start container.
         # 1) Set up a directory to store the bundle.
-        remove_path(run_state.bundle_path)
-        os.mkdir(run_state.bundle_path)
+        if self.shared_file_system:
+            try:
+                self._wait_for_bundle_folder(run_state)
+            except self.NoBundleFolderException as ex:
+                message = 'Bundle folder unavailable: %s' % str(ex)
```
fabeschan: I assume the user will see this failure message. Is it actionable? (i.e. user should delete and retry) If so, specify that in the message as well

codalab/worker/local_run/local_run_manager.py
```diff
@@ -41,17 +41,20 @@ def __init__(
...
         self._docker = docker.from_env()
-        self._bundles_dir = os.path.join(work_dir, LocalRunManager.BUNDLES_DIR_NAME)
-        if not os.path.exists(self._bundles_dir):
-            logger.info('{} doesn\'t exist, creating.'.format(self._bundles_dir))
-            os.makedirs(self._bundles_dir, 0o770)
+        self._shared_file_system = shared_file_system
+        if not shared_file_system:
+            self._bundles_dir = os.path.join(work_dir, LocalRunManager.BUNDLES_DIR_NAME)
+            if not os.path.exists(self._bundles_dir):
+                logger.info('%s doesn\'t exist, creating.', self._bundles_dir)
```
fabeschan, line 58: any plans for an ELK-esque logging setup? say if you can make every logging message traceable to a bundle-uuid/worker-id/server-action etc should make debugging production issues much easier

.travis.yml
```diff
@@ -12,7 +12,6 @@ jobs:
         - docker
       env:
         - CODALAB_USERNAME=codalab CODALAB_PASSWORD=codalab CI=false
-      python: 3.6
```
percyliang: What's the rationale for removing this?

.travis.yml
```diff
@@ -23,6 +22,8 @@ jobs:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
       script:
         - python3 codalab_service.py start --services default monitor test --version ${TRAVIS_BRANCH}
+        - python3 codalab_service.py stop
+        - python3 codalab_service.py start --services default monitor test --version ${TRAVIS_BRANCH} --shared-file-system
```
percyliang: We're running the tests again?  This is going to make our builds really slow...

codalab/worker/local_run/local_run_manager.py
percyliang: where `[]` means all bundles?  Might be confusing...

codalab/worker/local_run/local_run_state.py
```diff
@@ -209,30 +212,42 @@ def _transition_from_PREPARING(self, run_state):
 
         # All dependencies ready! Set up directories, symlinks and container. Start container.
         # 1) Set up a directory to store the bundle.
-        remove_path(run_state.bundle_path)
-        os.mkdir(run_state.bundle_path)
+        if self.shared_file_system:
+            try:
+                self._wait_for_bundle_folder(run_state)
```
percyliang: "folder" => "directory"

codalab/worker/local_run/local_run_state.py
percyliang: Yes

codalab/worker/main.py
percyliang: Shouldn't be purely just internal use though since people might run `cl-worker` on the NLP mchines...

docker/compose_files/docker-compose.yml
```diff
@@ -133,6 +133,22 @@ services:
       - service
       - worker
 
+  shared-filesystem-worker:
```
percyliang: It's unfortunate that this has to be a completely separate thing...

percyliang (changes requested):

codalab/worker/local_run/local_run_manager.py
bkgoksel: It is kind of confusing, but short of changing the worker API or adding a special const value to report here like `ALL` (sounds smelly to me) there's no better solution here.

Also, since in the current model `shared_filesystem_worker` is both a server-side AND worker-side setting, the same server will never have both shared filesystem and regular workers, so the dependencies field will be essentially meaningless for shared filesystem setups. Given all this, I think it is safe to just report no dependencies here.

I should add a comment explaining this though.

codalab/worker/main.py
bkgoksel: Yep, that's just how it was initially named, will remove that caveat.

codalab/worker/local_run/local_run_manager.py
bkgoksel: I don't see a use case in the short term, since the main new deployment regime we're aiming for is the job-queue system where workers are expected to be relatively short lived, and caches aren't used that much.

That said, there might be longer term uses...

codalab/worker/local_run/local_run_state.py
bkgoksel: Yeah, although something along the lines of 'ensure your bundle store is properly mounted to your worker host machine' would be better.

codalab/worker/local_run/local_run_manager.py
bkgoksel: This is a great idea, although probably out of scope for this PR. Can you create an issue for this? And then someone can pick it up in the future.

codalab/worker/local_run/local_run_manager.py
percyliang: Could we return `None` at least so it can't be confused with no dependencies?

.travis.yml
bkgoksel: Accidentally removed, adding back.

.travis.yml
bkgoksel: Just created #1536 so I can change this to only run worker-related tests

docker/compose_files/docker-compose.yml
bkgoksel: True, but:
1. There's not really a good way to pass bool arguments to python programs that are fired from docker-compose (with other arguments we can pass env variables but for bool arguments we'd have to roll our own bool argument parsing logic that would be different from python default and that would be bad too)
2. Shared fs workers require mounting the bundle store as a volume as well, so there's some difference in the container setup too. We could make all workers mount that volume, but I think this way the differences are more concrete between the two setups.

Besides, this way it's obvious from the container name whether the worker is shared fs or not.

percyliang: Yeah, that's fine.

bkgoksel:
One more thing:
Currently the server needs to be configured to work with shared filesystem workers as well.
This is done in config.json, and once it is set up, the server only works with shared filesystem workers.

As part of my Slurm Worker PR, I wrote a different mechanism that allows workers themselves to report whether they're shared filesystem to the server, and then the server could treat those workers differently. This allowed both shared filesystem and non shared filesystem workers on the same server (and meant we didn't have to change the server config to allow it to use shared fs workers)

However, this change requires another change to the server <-> worker API and also running a db migration (migration is on the `workers` table, so it is trivial if we bring down all workers, run the migration, and bring up all-updated workers)

What do you think @percyliang , should I add those changes here too? Or do you think the current way it is done is fine?

percyliang:
I think having the flexibility to have shared file system workers + non-shared file system workers would definitely be nice, because being able to connect a cl-worker locally to any codalab instance is really useful (you can go into the docker container and muck around / debug - we should advertise this more).  Also, not having to configure the server is a plus too.  The only thing is making sure that this doesn't make things terribly complicated (which I don't think is the case?)

bkgoksel:
Ok, added those changes to this PR as well

codalab/worker/local_run/local_run_manager.py
bkgoksel: Currently none of the upstream code is written to handle a worker that just doesn't report None for dependencies. Besides, even if we did that, the end result would be that the worker dependencies row for that worker would be empty on the database, so it wouldn't change much if we did that through reporting a None/null field on the checkin vs reporting no dependencies.

One thing that might be good to add is to update the worker filtering code to assume all dependencies are locally cached for all runs on shared fs workers, but I can make that change in the filtering logic rather than here.

bkgoksel: Just pushed the latter change I mentioned to the `filter_and_sort_workers` method.

codalab/worker/local_run/local_run_state.py
fabeschan: hmm. On one hand, if the worker is user-owned, 'ensure your bundle store is properly mounted to your worker host machine' would be a helpful message. On the other, if the user doesn't own the worker, that same message would be less helpful.

.travis.yml
percyliang: If shared-file-system is a worker-only config flag, then we don't need to restart the server stack, but just the worker.

codalab/model/tables.py
```diff
@@ -313,6 +313,9 @@
         'checkin_time', DateTime, nullable=False
     ),  # When the worker last checked in with the bundle service.
     Column('socket_id', Integer, nullable=False),  # Socket ID worker listens for messages on.
+    Column(
+        'shared_file_system', Boolean, nullable=False
+    ),  # Whether the worker is on a shared FS with server.
```
percyliang: Replace: "Whether the worker and the server have a shared filesystem"

codalab/server/bundle_manager.py
```diff
@@ -509,7 +510,7 @@ def _construct_run_message(self, worker, bundle):
         message = {}
         message['type'] = 'run'
         message['bundle'] = bundle_util.bundle_to_bundle_info(self._model, bundle)
-        if self._worker_model.shared_file_system and worker['user_id'] == self._model.root_user_id:
+        if worker['shared_file_system']:
```
percyliang, line 515: How come the logic changed?

codalab/worker/local_run/local_run_state.py
```diff
@@ -209,30 +212,50 @@ def _transition_from_PREPARING(self, run_state):
...
-                self.dependency_manager.dependencies_dir, dependency_path
-            )
-
             docker_dependency_path = os.path.join(docker_dependencies_path, dep.child_path)
-
-            os.symlink(docker_dependency_path, child_path)
+            if self.shared_file_system:
+                # On a shared FS, we know where the dep is stored and can get the contents directly
+                dependency_path = os.path.realpath(
+                    os.path.join(os.path.realpath(dep.location), dep.parent_path)
```
percyliang: do we need to call `os.path.realpath` twice?

codalab/worker/local_run/local_run_state.py
```diff
@@ -283,6 +306,24 @@ def _transition_from_PREPARING(self, run_state):
             gpuset=gpuset,
         )
 
+    class NoBundleFolderException(Exception):
```
percyliang: Folder => Directory

codalab_service.py
```diff
@@ -665,7 +666,10 @@ def wait_rest_server(cmd):
         self.bring_up_service('worker-manager-gpu')
         self.bring_up_service('frontend')
         self.bring_up_service('nginx')
-        self.bring_up_service('worker')
+        if self.args.shared_file_system:
+            self.bring_up_service('shared-filesystem-worker')
```
percyliang: Let's call this `worker-shared-file-system` to be consistent.

percyliang (approved):
Great!  One thing that's kind of tricky about this code is that there's a lot of checks of `shared_file_system` spread throughout the codebase.  Not sure if there's anything to do about it though.

codalab/server/bundle_manager.py
bkgoksel: I don't why that root user check was there in the first place, but just thinking about the situation now, there's no reason why a non-root user cannot start a shared filesystem worker (as in the Slurm use case). So I have no explanation for the old logic.

codalab/worker/local_run/local_run_state.py
bkgoksel: Probably not, this was an artifact from the old worker code, but I can change it.

codalab_service.py
bkgoksel: good point, will do

codalab/lib/upload_manager.py
```diff
@@ -176,36 +176,6 @@ def _simplify_directory(self, path, child_path=None):
         path_util.rename(child_path, path)
         path_util.remove(temp_path)
 
-    def update_metadata_and_save(self, bundle, enforce_disk_quota=False):
```
bkgoksel, line 179: Moved this function to `bundle_model` because otherwise calculating and saving the disk use of shared filesystem bundles would be very difficult.
I also think this kind of metadata calculation makes a lot more sense to be in a less specific module than `upload_manager` in a world where bundles could be created without being uploaded.

codalab/model/bundle_model.py
```diff
@@ -852,7 +888,28 @@ def transition_bundle_finished(self, bundle):
     # Bundle state machine helper functions
     # ==========================================================================
 
-    def bundle_checkin(self, bundle, worker_run, user_id, worker_id):
+    def update_metadata_and_save(self, bundle, bundle_location, enforce_disk_quota=False):
```
bkgoksel: Should add a comment here.

codalab/model/worker_model.py
```diff
@@ -172,34 +181,14 @@ def get_workers(self):
...
                 and self._deserialize_dependencies(row.dependencies),
+                'shared_file_system': row.shared_file_system,
             }
             for row in worker_rows
         }
         for row in worker_run_rows:
             worker_dict[(row.user_id, row.worker_id)]['run_uuids'].append(row.run_uuid)
         return list(worker_dict.values())
 
-    def get_bundle_worker(self, uuid):
```
bkgoksel, line 182: This was moved to `bundle_model` since this is more a property of the bundle (conceptually) and practically it was needed in there, and any client of this method has access to both bundle and worker models.

codalab/worker/local_run/local_run_state.py
bkgoksel: Yeah I guess that + contact the administrator should cover both cases

.travis.yml
bkgoksel: Hmm yes, after those changes, this can just restart the worker.

codalab/worker/local_run/local_run_manager.py
```diff
@@ -290,17 +299,17 @@ def mark_finalized(self, uuid):
             with self._lock:
                 self._runs[uuid] = self._runs[uuid]._replace(finalized=True)
 
-    def read(self, run_state, path, dep_paths, args, reply):
+    def read(self, run_state, path, args, reply):
```
bkgoksel: Make these take `uuids` like in the Slurm Run PR

codalab/worker/local_run/local_run_state.py
```diff
@@ -209,30 +212,48 @@ def _transition_from_PREPARING(self, run_state):
 
         # All dependencies ready! Set up directories, symlinks and container. Start container.
         # 1) Set up a directory to store the bundle.
-        remove_path(run_state.bundle_path)
-        os.mkdir(run_state.bundle_path)
+        if self.shared_file_system:
+            try:
+                self._wait_for_bundle_directory(run_state)
```
bkgoksel: Make this an async PREPARING step like waiting for docker image downloads

PR #1536, opened 2019-10-18
Add a test command to be able to run particular tests

bkgoksel:
This is mainly so that we can test shared filesystem workers on Travis without running all the tests from scratch. Also helpful for locally developing backend features as it makes it possible to only rerun failing tests.

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -24,6 +24,8 @@ import os
 import socket
 import subprocess
 
+from test_cli import TestModule
+
 DEFAULT_SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker', 'init']
 
 ALL_SERVICES = DEFAULT_SERVICES + ['test', 'monitor', 'worker-manager-cpu', 'worker-manager-gpu']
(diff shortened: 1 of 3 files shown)


codalab_service.py
```diff
@@ -316,7 +330,7 @@ def _get_parser():
                 default=False,
             )
             cmd.add_argument(
-                'images',
+                '--images',
```
percyliang: Doesn't this make it so you can't do `./codalab_service.py build frontend` anymore?

percyliang (changes requested):
Now we have two ways of invoking tests...through `test` and through `start`.  The latter is useful to have still, but could we factor out the common piece as much as possible?

Maybe we should make `test` an option just like `--build` is an option for `start`, so that it's consistent that options are for invoking other commands other than `start`.

codalab_service.py
bkgoksel: Yes it does.
The problem is that we add this option to every command, and that means we can't add any other positional arguments to any other command.

I think I can make it such that it stay a positional argument for the `build` command and named argument for the other commands. Let me try that.

bkgoksel:
I think with the latest changes, `test` now acts pretty much the same as `build`. I added some conditional logic to make their respective arguments positional for the explicit commands and named for the `start` command

bkgoksel:
Now one problem is this:
1. To list all the runs that can be run, I import the `TestModule`.
2. The `TestModule` imports `CodaLabManager` to make API calls directly
3. `CodaLabManager` imports other codalab modules
4. We end up needing to install a bunch of external packages like `psutil` or `sqlalchemy` just to be able to use the `codalab_service.py`

Right now the only simple option I see around this is to hard code the list of possible tests into the `codalab_service` file and get rid of the import `test_cli` but that doesn't feel great either.

As another solution, I moved the `CodaLabManager` import to the init function of the test module so we can import from the `test_cli` file without having to import the rest of Codalab. Not great cause we end up with not-a-top-level-import but this feels like the cleanest solution without having to install a whole bunch of stuff

percyliang:
Yeah, not great to have `codalab_service.py` have so many dependencies
(maybe put a comment at the top).

Why does codalab_service.py need to know about the list of tests?  I
usually do `docker exec codalab_rest_server python3 test_cli.py`...we could
even do that directly in travis.

In general, I'm wary of making codalab_service.py too monolithic if it's
just passing commands through.

bkgoksel:
Mainly to print the list of tests on the `help` and do arg checking so we don't fail downstream.
I think it runs fine now with the import-inside-init trick I did on the test cli file, so if we're fine with that we can keep it like this.

Otherwise list of tests can be a freeform string argument and could be transparently passed to the test cli file.

percyliang (approved):

Issue #1254, opened 2019-06-24
jobs take a long time to go through the staged,starting,preparing,running pipeline


percyliang:
Sometimes something just seems hung for no good reason.

gumulka:
Could this be the reason, that the cl-worker with the  --exit-when-idle parameter does not execute anything?

Whenever I start the worker with  --exit-when-idle, it does not execute my runs, but when I start it without the flag, it runs smoothly. 

Sometimes, when starting with  --exit-when-idle i see a "starting" in my worksheet, but it reverts to "staged" instead of "preparing".

percyliang:
That seems like a separate bug since in your case, the worker never runs it, not just runs it slowly.

candicegjing:
One issue I noticed in my local environment is the transition from `staged` to `running` takes a long time. While I was debugging the above slow transition, I found out that this line is the bottleneck: https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/worker_model.py#L152-L158. When I submitted a run via CLI, the available worker is always empty and hence no job would be assigned with a valid worker. The job then got stuck in `staged` state.
For this specific run : `cl run time`
`stage` to `running` takes 44 minutes. Basically, it keeps looping on checking available workers and each of the phase takes the following time (unit is second):
```
2019-10-17 22:55:32,945 schedule bundles takes time = 0.08561134338378906
2019-10-17 22:55:32,950 failed bundles takes time = 0.0048503875732421875
2019-10-17 22:55:33,456 stage bundles takes time = 0.004965543746948242
2019-10-17 22:55:33,460 make bundles takes time = 0.00351715087890625
```
Once the bundle started running:
```
2019-10-17 22:55:34,522 Starting run bundle 0xe3d377b53b6b49cb81ff27ffbb00e608
2019-10-17 22:55:34,522 schedule bundles takes time = 0.4564027786254883
2019-10-17 22:55:34,526 failed bundles takes time = 0.003636598587036133
2019-10-17 22:55:35,034 stage bundles takes time = 0.005255699157714844
2019-10-17 22:55:35,037 make bundles takes time = 0.003418445587158203
```

percyliang:
Can you quantify how much time this takes compared to the rest of the code?

candicegjing:
From my local database instance:
```mysql> select * from worker;
+---------+-----------------------------+------+------+------+--------------+-----------------+---------------------+-----------+
[...]

PR #1537, opened 2019-10-18
Update sql outerjoin query

candicegjing:
This will fix slow transition from `staged` to `preparing` in  #1254 . 
This sql query change resolved the slow transition issue from `staged` to `preparing` on my local machine. The original outerjoin(no join key specified) somehow doesn't return any valid worker even if there is a valid one. I waited for 45 minutes to get a simple `cl run date` running. Once I added the join key, it ran immediately.

### DIFF
--- a/codalab/model/worker_model.py
+++ b/codalab/model/worker_model.py
@@ -152,7 +152,10 @@ class WorkerModel(object):
         with self._engine.begin() as conn:
             worker_rows = conn.execute(
                 select([cl_worker, cl_worker_dependency.c.dependencies]).select_from(
-                    cl_worker.outerjoin(cl_worker_dependency)
+                    cl_worker.outerjoin(
+                        cl_worker_dependency,
+                        cl_worker.c.worker_id == cl_worker_dependency.c.worker_id,
+                    )
                 )
             ).fetchall()
             worker_run_rows = conn.execute(cl_worker_run.select()).fetchall()


yipenghe:
<s>What is the testing, do you need it to be merged into master?</s>
just saw slack, will test it out tomorrow

percyliang (commented):
Can you provide timing statistics on how long things take before versus after?

@Levilian to verify this

candicegjing:
I put the timing statistics in the original comment: **I waited for 45 minutes to get a simple cl run date running. Once I added the join key, it ran immediately.**

percyliang:
Locally, even without this change, `cl run date` runs in ~5 seconds.  When it takes 45 minutes, what state is taking that time?

percyliang:
With one worker and everything started from scratch, things seem to be fine (though they could probably be a bit faster)...

candicegjing:
45 minutes was in `staged` state. Not sure what's the difference between our local machines. I tested several times, it was always the same for me.

bkgoksel:
I'm confused how this outerjoin was working in the first place without a key to join on. 

Edit: after digging further [into the docs](https://docs-sqlalchemy.readthedocs.io/ko/latest/core/metadata.html?highlight=outerjoin#sqlalchemy.schema.Table.outerjoin.params.onclause) it looks like SQLAlchemy automatically joins on foreign key columns if no other on clause is specified. That sounds like this change only makes explicit what was implicitly there before, and it should not really affect performance.

That is unless we're hitting some edge case or bug in sqlalchemy.

codalab/model/worker_model.py
```diff
@@ -152,7 +152,10 @@ def get_workers(self):
...
             worker_rows = conn.execute(
                 select([cl_worker, cl_worker_dependency.c.dependencies]).select_from(
-                    cl_worker.outerjoin(cl_worker_dependency)
+                    cl_worker.outerjoin(
+                        cl_worker_dependency,
+                        cl_worker.c.worker_id == cl_worker_dependency.c.worker_id,
+                    )
                 )
             ).fetchall()
             worker_run_rows = conn.execute(cl_worker_run.select()).fetchall()
```
bkgoksel, line 161: It is also strange to me that we do an outerjoin for the `worker_dependency` table but do this second loop for the `worker_run` table, even though they feel like they should be analogous to me. 
Git blame shows @kashizui made this change a while back. Maybe he can chime in?

candicegjing: @bkgoksel I think the `worker_id` returned from `work_rows` is a superset of `worker_run_rows`. When we build the `worker_dict`, we use the superset. However, when we fill the `run_uuids` in this superset, we get that information from `worker_run_rows`. So in the end, the `get_workers()` function will return reconstructed `worker_dict` which contains all the workers from `worker` table, but only some of them will have not-none `run_uuids`.

candicegjing:
@percyliang and @bkgoksel I guess even if this issue doesn't have much impact on your guys' machine, but it definitely improves readability of the SQL query. Let me know if we still want to merge this change, otherwise, I'll just close it.

bkgoksel (approved):
I agree that it's better to have the join key be explicit in case maintainers in the future are not aware of the automatic foreign key joining of these tables.

percyliang (approved):

PR #1538, opened 2019-10-19
Cleaning up the hovering insert panels

yipenghe:
Fixing #1518 
The upload & run button the nav bar already works when a there's a selected bundle, just need some code cleaned up in bundle row.

Test: checkout & hover to bundle rows, should no longer appear the insert panel on any bundle table

p.s: the feeling of just deleting, wow...


epicfaace (approved):
Looks good. Because of this change, though, we can no longer add a bundle to the _beginning_ of a table (you have to insert after an existing row), but that should be ok.

yipenghe:
> Looks good. Because of this change, though, we can no longer add a bundle to the _beginning_ of a table (you have to insert after an existing row), but that should be ok.

If we ultimately have dragging work it should be good

PR #1544, opened 2019-10-21
Test restarting worker in Travis tests

bkgoksel:
This is to make sure worker restarts are fine and there is no
serialization/deserialization problems in worker state committing on
disk.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ jobs:
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
       script:
         - python3 codalab_service.py test default --version ${TRAVIS_BRANCH}
-          # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization
-        - docker stop codalab_worker_1 && python3 codalab_service.py start --services worker --version ${TRAVIS_BRANCH} && python3 codalab_service.py test run --version ${TRAVIS_BRANCH}
+          # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization)
+        - docker restart codalab_worker_1 && python3 codalab_service.py test run --version ${TRAVIS_BRANCH}
     - stage: deploy
       script: echo "Deploying"
       language: python
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ jobs:
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
       script:
         - python3 codalab_service.py test default --version ${TRAVIS_BRANCH}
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
       script:
         - python3 codalab_service.py test default --version ${TRAVIS_BRANCH}
+          # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization
+        - docker stop codalab_worker_1 && python3 codalab_service.py start --services worker --version ${TRAVIS_BRANCH} && python3 codalab_service.py test run --version ${TRAVIS_BRANCH}
```
bkgoksel: Maybe running the `run` tests again isn't necessary.

.travis.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
       script:
         - python3 codalab_service.py test default --version ${TRAVIS_BRANCH}
+          # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization
```
percyliang: missing ')'

percyliang (commented):
What problem is this reacting to?

bkgoksel:
Not a specific problem, but while working on #1505 I realized I introduced some serialization/deserialization bugs that were not caught by the tests and I only caught them after doing this simple worker restart.  I also added some unit tests after that, but I think adding an end-to-end state serialization/deserialization test by doing this restart wouldn't hurt.

.travis.yml
percyliang: What about just doing `docker restart codalab_worker_1`?

I guess you need to do something to test that the worker is alive and well...

.travis.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
       script:
         - python3 codalab_service.py test default --version ${TRAVIS_BRANCH}
+          # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization)
+        - docker restart codalab_worker_1 && python3 codalab_service.py test run --version ${TRAVIS_BRANCH}
```
epicfaace, line 28: Why only run the `run` tests here?

bkgoksel: I don't think we'd get any extra signal from the other tests at this point since we don't have any dedicated tests for things like dependency/docker image caches or worker state resumption. This is just a sanity check that all the worker sub-components have initialized successfully that they can complete basic runs.

percyliang (approved):

PR #1545, opened 2019-10-22
changed the default disk space to empty

### LINKED ISSUE #1202: frontend requests too much disk quota
Default disk quota is 10G.  Frontend tries to request 10G, so if you have any existing bundles, runs fail immediately.

Long term, should make a call to get the amount of disk quota first and prepopulate (maybe).

For now, leave the default blank, and put a comment that says if it's blank, then the default is all remaining disk quota.

Levilian:
Fix #1202

### DIFF
--- a/frontend/src/components/worksheets/NewRun/NewRun.jsx
+++ b/frontend/src/components/worksheets/NewRun/NewRun.jsx
@@ -165,7 +165,7 @@ class NewRun extends React.Component<{
         name: 'untitled-run',
         description: '',
         tags: [],
-        disk: kDefaultDisk,
+        disk: '',
         memory: kDefaultMemory,
         cpu: kDefaultCpu,
         gpu: kDefaultGpu,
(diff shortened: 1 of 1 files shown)


yipenghe:
Can you add some explanation on how it fixes the issue & also attach a screenshot

percyliang:
Related: can you also say how this was tested?

Levilian:
@yipenghe @percyliang This is a front-end change where insteading of setting the default value to "10g", which was the case before, I set it to "". 

See the image below:
Screen Shot 2019-10-22 at 3 28 04 PM

frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -165,7 +165,7 @@ class NewRun extends React.Component<{
         name: 'untitled-run',
         description: '',
         tags: [],
-        disk: kDefaultDisk,
+        disk: '',
```
yipenghe, line 167: If it just needs a frontend change it's good to go

Nit: Remove the kdefault constant declaration on line 131 before merging, since it's not used anymore

yipenghe (approved):

frontend/src/components/worksheets/NewRun/NewRun.jsx
Levilian: Thanks Yipeng!

PR #1546, opened 2019-10-22
fix the text when item gets selected; beautified the box shadow

Levilian:
Fix: #1516

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -325,7 +325,6 @@ class BundleRow extends Component {
                     className={classNames({
                         [classes.contentRow]: true,
                         [classes.highlight]: this.props.focused,
-                        [classes.cursor]: this.props.focused,
                         [classes.lowlight]: !this.props.focused && this.state.showDetail,
                     })}
                 >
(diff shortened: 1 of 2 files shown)


yipenghe:
Thanks for working on this. Can you attach screenshots next time in frontend related PRs? It would be easier to review. The code part looks good to me. I think adding some padding would be better. What do you think? @percyliang 
Screenshot from 2019-10-21 22-30-34

epicfaace:
I don't like the new box shadow. I think the old blurred border looked better as it seemed like you were really "selecting" a worksheet item:

image

I think the main issue here with #1516 was that the dark blue bar that comes to the left of the text upon selection of a worksheet item should actually be a bit more to the left, or be a bit closer to the text -- that way, it doesn't move the text to the right upon selection.

Levilian:
> I don't like the new box shadow. I think the old blurred border looked better as it seemed like you were really "selecting" a worksheet item:
> 
> image
> 
> I think the main issue here with #1516 was that the dark blue bar that comes to the left of the text upon selection of a worksheet item should actually be a bit more to the left, or be a bit closer to the text -- that way, it doesn't move the text to the right upon selection.

Ok. Maybe this is too big of a change. 

What I wanted to do here is to shoot for consistency across the page. Since we use box shadow for the header, I used box shadow for each item as well. The blurred blue box seems out of sync with the overall design. 

I will go ahead and remove/shift the blue line.

Levilian:
The demo is here: https://recordit.co/ArtZSYX8Vm

yipenghe (approved):
Looks good to me

PR #1549, opened 2019-10-22
changed how worksheet title and  web cli looks

Levilian:
#1520 : made cli more transparent
#1532 : made navbar above the web cli, which is hidden by default
#1548 : changed how title and name look like and their respective hover behavior

### DIFF
--- a/frontend/src/css/imports.css
+++ b/frontend/src/css/imports.css
@@ -6264,6 +6264,7 @@ fieldset[disabled] .btn-secondary.active {
 #ws_search {
   margin: 10px 0 20px;
   height: auto;
+  box-shadow: 0 1px 10px 0 rgba(0,0,0,0.12), 0 2px 4px -1px rgba(0,0,0,0.4);
 }
 #ws_search:focus {
   background: white;
@@ -6743,10 +6744,10 @@ body,
   position: relative;
   height: calc(100% - 20px - 25px);
 }
-#worksheet .editable-field:hover {
+.editable-field:hover {
   cursor: text;
-  border-radius: 4px;
-  box-shadow: 1px 1px 1px 1px gray;
+  border: 2px solid #eeeeee;
+  border-radius: 2px;
 }
 #ws_search {
   position: fixed;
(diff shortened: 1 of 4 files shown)


frontend/src/css/imports.css
```diff
@@ -7736,9 +7737,16 @@ fieldset[disabled] .worksheet_content .header-row .controls .edit-features .btn-
   }
   #ws_search,
   #worksheet .ws-panel,
-  #worksheet .dragbar {
+  #terminal-button {
```
percyliang, line 7740: We should really get off of this css file soon before we start hacking this too much...

epicfaace:
Oh, it appears this PR fixes #1549 as well.

epicfaace (commented):
Looks good.

I think you should probably keep the "dashboard" and "Codalab Dashboard" text in a blue color, though (just remove the underline upon hover so it doesn't look like a link). Otherwise, the text blends in with other text around it and it's not obvious that it's editable -- how would the user know that "dashboard" is editable, but "by codalab" isn't?

image

frontend/src/css/imports.css
```diff
@@ -6743,18 +6744,18 @@ body,
   position: relative;
   height: calc(100% - 20px - 25px);
 }
-#worksheet .editable-field:hover {
+.editable-field:hover {
```
epicfaace, line 6747: When you hover on an editable field, the text moves slightly off around 1px. Can you fix this?

frontend/src/css/imports.css
```diff
@@ -6743,18 +6744,18 @@ body,
...
+  border-radius: 2px;
 }
 #ws_search {
   position: fixed;
   width: 100%;
   margin: 0;
   padding: 10px 0;
   text-align: center;
-  background-color: #F8F9FA;
+  background: rgba(245, 245, 245, 0.55);
```
epicfaace: Can you make this a bit more opaque? Right now it's too transparent and a bit hard to read:

image

epicfaace (changes requested):

frontend/src/css/imports.css
Levilian: I see. In terms of hiding CLI when the screen is smaller than a certain size, is that expected behavior? Should I remove this restriction completely? 

Right now what I'm doing here is to hide the Open/Hide Terminal toggle bar when the screen becomes too small.

PR #1553, opened 2019-10-22
fix scrolling with keyboard shortcuts

### LINKED ISSUE #1528: Keyboard shortcuts, when going up/down, does not scroll the screen.

epicfaace:
fixes #1528

You can now scroll when using up and down arrows. However, using "j" and "k" still does not scroll. Is this intended behavior @percyliang or should it also scroll when using "j" and "k" to navigate?

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -41,7 +41,7 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <div id="root" style="height: 100%"></div>
+    <div id="root"></div>
     <!--
       This HTML file is a template.
       If you open it directly in the browser, you will see an empty page.
--- a/frontend/src/components/worksheets/ContextMenu.js
+++ b/frontend/src/components/worksheets/ContextMenu.js
@@ -64,12 +64,10 @@ class ContextMenu extends React.Component {
 
     handleOpenEvent = (payload) => {
         this.setState(Immutable({ ...payload, mouseState: mouse }));
-        $('#worksheet_container').css({ overflow: 'hidden' });
     };
 
     closeMenu = () => {
         this.setState(this.getInitialState());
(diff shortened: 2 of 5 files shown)


percyliang:
Yes, we want j/k to scroll - basically make it mimic the old frontend.

yipenghe:
A minor bug, if the focus is at the first item in worksheet and we press up arrow, the focus changes to the last item in the worksheet

epicfaace:
> A minor bug, if the focus is at the first item in worksheet and we press up arrow, the focus changes to the last item in the worksheet

Will do this as part of #1515.

epicfaace:
This PR has also changed the actual scrolling element to be just the worksheet (as opposed to the body -- note how the scrollbar only starts after the header). This is more in line with the old frontend.

image


old frontend:

image

yipenghe:
> > A minor bug, if the focus is at the first item in worksheet and we press up arrow, the focus changes to the last item in the worksheet
> 
> Will do this as part of #1515.

<s>Thanks for fixing this in #1560 . One other thing, after that fix, currently if we use down to move all the way down, the focus will stop at the last item, but if we use up to move all the way up, it will lose focus for the first one. I remember we added a dummy item to allow adding a new cell, is this behaviour related (so the focus actually went to the dummy item we can't see)? 

I think we should be consistent on both directions (so either don't focus on anything when we press up/down on the first/last item, or let the focus stay on the first/last item)</s>
It actually is the behaviour of the current frontend, good to keep it for now then

yipenghe (approved):

epicfaace:
> I think we should be consistent on both directions (so either don't focus on anything when we press up/down on the first/last item, or let the focus stay on the first/last item)

Good point. Even though it wasn't in the old version, I think it's a good change -- and I've went ahead and done it.

PR #1554, opened 2019-10-22
re-add context menu

### LINKED ISSUE #1517: Right clicking on a bundle should show the context menu

epicfaace:
fixes #1517

This was in the backlog, but when I was working on #1553, I found the fix to this so made a PR for this anyway.

image


percyliang (approved):
Note: we'll want to remove this menu eventually in favor of gmail-style checkboxes.  I think the right-click menu is very non-discovarable.

yipenghe (approved):
Looks good, not blocking, but can we change the order & put "add to my home worksheet" at top?

PR #1556, opened 2019-10-22
Add database migration for the shared filesystem workers

bkgoksel:
@percyliang looks like you are set as the default git author on the SM codalab machines :)


bkgoksel:
Merged since tested locally and trivial

PR #1557, opened 2019-10-22
Respect minimum number of workers on worker manager

bkgoksel:
An earlier change made it such that we would never launch a worker if
there were no staged bundles. This updates the logic slightly so we
still launch workers to reach the minimum worker count even when there
are no bundles waiting.

### DIFF
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -110,27 +110,27 @@ class WorkerManager(object):
         )
 
         # There are two conditions under which we want more workers.
-        want_more_workers = False
-
         # 1) We have fewer than min_workers.
+        less_than_min_workers = False
         if len(worker_jobs) < self.args.min_workers:
             logger.info(
                 'Want to launch a worker because we are under the minimum ({} < {})'.format(
                     len(worker_jobs), self.args.min_workers
                 )
             )
-            want_more_workers = True
+            less_than_min_workers = True
 
         # 2) There is a staged bundle AND there aren't any workers that are still booting up/starting
+        staged_bundles = False
         if len(self.staged_uuids) > 0:
             logger.info(
                 'Want to launch a worker because we have {} > 0 staged bundles'.format(
                     len(self.staged_uuids)
                 )
             )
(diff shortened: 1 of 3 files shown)


codalab/worker_manager/worker_manager.py
```diff
@@ -143,7 +143,8 @@ def run_one_iteration(self):
 
             # Make sure we don't queue up more workers than staged UUIDs if there are
             # more workers still booting up than staged bundles
-            if len(pending_worker_jobs) >= len(self.staged_uuids):
+            # However override this if we have less than minimum number of workers
+            if len(pending_worker_jobs) >= len(self.staged_uuids) and not less_than_min_workers:
```
percyliang: Is there a way to make this logic a bit easier to follow, since `less_than_min_workers` showing up again here makes things kind of coupled?

bkgoksel: Ok changed the logic a little bit.

codalab/worker_manager/worker_manager.py
```diff
@@ -160,8 +151,18 @@ def run_one_iteration(self):
                         len(worker_jobs), self.args.max_workers
                     )
                 )
-                return
+                want_workers = False
+
+        # We have fewer than min_workers, so launch one regardless of other constraints
```
yipenghe: Is this override also intended for the time constraint? The new added comment on line 136 sounds like we only wanted to override that situation before. I understand that the max worker constraint won't be override assuming args.min_workers < args.max_workers

codalab/worker_manager/worker_manager.py
```diff
@@ -160,8 +151,18 @@ def run_one_iteration(self):
...
+        if len(worker_jobs) < self.args.min_workers:
```
yipenghe, line 156: Is this override also intended for the time constraint? The new added comment on line 136 sounds like we only wanted to override the queue situation before. I understand that the max worker constraint won't be override assuming args.min_workers < args.max_workers

percyliang (approved):
This looks better, thanks.

Issue #1525, opened 2019-10-16
When inserting text, remove the highlighting of the existing row, and add highlighting to the new row when it is added.


epicfaace:

Levilian:
@epicfaace can you say more about this issue

Levilian:
#1558

PR #1559, opened 2019-10-23
add highlight to the right row when adding/deleting new rows

Levilian:
Fix: 
#1525 
#1558

### DIFF
--- a/frontend/src/components/worksheets/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet.js
@@ -721,6 +721,8 @@ class Worksheet extends React.Component {
                         var focus = this.getFocusAfterBundleRemoved(items);
                         this.setFocus(focus[0], focus[1]);
                     } else {
+                        // if the change has no impact on bundles, but on adding an item
+                        // then we want the focus to be the one below the current focus
                         this.setFocus(this.state.focusIndex + 1, this.state.subFocusIndex);
                     }
                     this.setState({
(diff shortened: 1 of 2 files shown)


Levilian:
Adding rows: https://recordit.co/kB1B7LYeZE

Levilian:
Deleting rows: https://recordit.co/08cKS9MJzv

Levilian:
Edges cases: 
deleting the last row
adding the first row
https://recordit.co/6DuT6v0esE

frontend/src/components/worksheets/Worksheet.js
```diff
@@ -718,8 +718,10 @@ class Worksheet extends React.Component {
                         // If the number of bundles decreases, then focus should be on the same bundle as before
                         // unless that bundle doesn't exist anymore, in which case we select the closest bundle that does exist,
                         // where closest means 'next' by default or 'last' if there is no next bundle.
-                        var focus = this.getFocusAfterBundleRemoved();
+                        var focus = this.getFocusAfterBundleRemoved(items);
                         this.setFocus(focus[0], focus[1]);
+                    } else {
+                        this.setFocus(this.state.focusIndex + 1, this.state.subFocusIndex);
```
yipenghe: Overall looks good, can you add a comment here explaining when this else condition happens and why the setFocus indices? (like the comments from 718-720) The code is getting more complex so would be good to have comments to follow

yipenghe (changes requested):
Thanks for recording the examples! Just add the comments and good to go

yipenghe (approved):
Add the comments before merging then good to go

frontend/src/components/worksheets/Worksheet.js
Levilian: Thanks Yipeng!

PR #1563, opened 2019-10-23
Fix ctrl+enter to save a bundle

### LINKED ISSUE #1521: Bundle run screen: Ctrl + Enter should run a bundle

epicfaace:
Fixes #1521. This error happens on the bundle run
screen. The fix makes sure that the `onKeyPress`
prop is passed on and uses the right key code
(13) for enter.


yipenghe:
Looks like an accidentally opened PR?

PR #1566, opened 2019-10-23
Raise error when dependency is empty in CLI

### LINKED ISSUE #1500: desugar_command:  error concatenating 'NoneType' and 'str'
```
POST /cli/command

Query params:


Form params:


JSON body:
{
    "command": "cl run : :src 'python src/predict.py < valid_image_paths.csv'",
    "worksheet_uuid": "0x5b880b4b3bf44d56911789b4adeca28b"
}

Local variables:
subpath=None
bundle=None
_=None
val=''
key=None
dep=':'
val2key={'': None}
target_spec=[]
key2val={None: ''}

    rv = callback(*a, **ka)
  File "/opt/codalab-worksheets/codalab/rest/cli.py", line 48, in post_worksheets_command
    return general_command(query['worksheet_uuid'], query['command'])
  File "/opt/codalab-worksheets/codalab/rest/cli.py", line 149, in general_command
    structured_result = cli.do_command(args)
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 852, in do_command
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 846, in <lambda>
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 1518, in do_run_command
    args.target_spec, args.command = desugar_command(args.target_spec, args.command)
  File "/opt/codalab-worksheets/codalab/lib/cli_util.py", line 118, in desugar_command
    get(spec)
  File "/opt/codalab-worksheets/codalab/lib/cli_util.py", line 113, in get
    target_spec.append(key + ':' + val)
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

candicegjing:
Fixed #1500

### DIFF
--- a/codalab/lib/cli_util.py
+++ b/codalab/lib/cli_util.py
@@ -1,5 +1,5 @@
 import re
-import datetime
+import os
 
 from codalab.common import precondition, UsageError
 
@@ -8,6 +8,29 @@ WORKSHEET_SEPARATOR = "//"
 
 TARGET_KEY_REGEX = r"(?<=^)(?:([^:]*?)\:(?!:))?(.*(?=$))"
 TARGET_REGEX = r"(?<=^)(?:(.*?)\:\:)?(?:(.*?)\/\/)?(.+?)(?:\/(.*?))?(?=$)"
+
+# Formatting Constants
+ADDRESS_SPEC_FORMAT = "(<alias>|<address>)"
+BASIC_SPEC_FORMAT = '(<uuid>|<name>)'
+BASIC_BUNDLE_SPEC_FORMAT = '(<uuid>|<name>|^<index>)'
+
+GLOBAL_SPEC_FORMAT = "[%s%s]%s" % (ADDRESS_SPEC_FORMAT, INSTANCE_SEPARATOR, BASIC_SPEC_FORMAT)
+WORKSHEET_SPEC_FORMAT = GLOBAL_SPEC_FORMAT
+
+BUNDLE_SPEC_FORMAT = '[%s%s]%s' % (
+    WORKSHEET_SPEC_FORMAT,
+    WORKSHEET_SEPARATOR,
+    BASIC_BUNDLE_SPEC_FORMAT,
+)
+
+WORKSHEETS_URL_SEPARATOR = '/worksheets/'
+
+TARGET_SPEC_FORMAT = '%s[%s<subpath within bundle>]' % (BUNDLE_SPEC_FORMAT, os.sep)
+RUN_TARGET_SPEC_FORMAT = '[<key>]:' + TARGET_SPEC_FORMAT
+MAKE_TARGET_SPEC_FORMAT = '[<key>:]' + TARGET_SPEC_FORMAT
+GROUP_SPEC_FORMAT = '(<uuid>|<name>|public)'
+PERMISSION_SPEC_FORMAT = '((n)one|(r)ead|(a)ll)'
+UUID_POST_FUNC = '[0:8]'  # Only keep first 8 characters
 
 
 def nested_dict_get(obj, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


yipenghe:
Can you add a screenshot on the error message getting thrown?

candicegjing:
@yipenghe This is how it looks when dependency isn't correctly specified. 
Screen Shot 2019-10-23 at 6 03 44 PM

codalab/lib/cli_util.py
```diff
@@ -109,6 +109,11 @@ def get(dep):  # Return the key
                     'key %s exists with multiple values: %s and %s' % (key, key2val[key], val)
                 )
         else:
+            if key is None:
+                raise UsageError(
+                    'Dependency is empty. Please provide a valid dependency in the format of :<dependency>.'
```
percyliang: It's really should be '[<key>]:<dependency>' or something - can you use the actual format of a 'target' for consistency, especially, if you can just use it as a variable'...

candicegjing: What do you mean by `[]:`?

percyliang: Oh, `[&lt;key&gt;]:`...(angle brackets got swallowed up)

bkgoksel: There should be a constant somewhere in our codebase that specifies the actual format. All the format strings like that are called `*_SPEC` and this one might be called `DEPENDENCY_SPEC` or something like that. It would be good to find and use that for consistency.

candicegjing: @percyliang @bkgoksel Do you guys mean this one `[<key>]:[[(<alias>|<address>)::](<uuid>|<name>)//](<uuid>|<name>|^<index>)[/<subpath within bundle>]`?

percyliang: Yeah

codalab/lib/cli_util.py
```diff
@@ -109,6 +109,12 @@ def get(dep):  # Return the key
                     'key %s exists with multiple values: %s and %s' % (key, key2val[key], val)
                 )
         else:
+            if key is None:
+                raise UsageError(
+                    'Dependency is empty. Please provide a valid dependency in the format of [<key>]:'
+                    '[[(<alias>|<address>)::](<uuid>|<name>)//](<uuid>|<name>|^<index>)[/<subpath within bundle>].'
```
percyliang: Can you not duplicate this here but instead use the `*_SPEC_FORMAT` variables defined in `bundle_cli.py`?  You might have to move those definitions.

candicegjing: Updated. If we want to move all the *SPEC_FORMAT from `bundle_cli` to `cli_util` to avoid circular importing, let me know

codalab/lib/cli_util.py
```diff
@@ -109,6 +111,13 @@ def get(dep):  # Return the key
                     'key %s exists with multiple values: %s and %s' % (key, key2val[key], val)
                 )
         else:
+            if key is None:
+                raise UsageError(
+                    'target_spec is empty. Please provide a valid target_spec in the format of {}.'.format(
+                        codalab.lib.bundle_cli.RUN_TARGET_SPEC_FORMAT
```
percyliang: This introduces a circular dependency since bundle_cli depends on cli_util.  Move the definitions to a new file and have both files include that.

candicegjing: Can we move them to `cli_util.py` as there are already some formatting variables definied there? This way we can keep the orignial importing.

percyliang: Sure.

candicegjing: Done.

percyliang (approved):
Great, thanks!

PR #1567, opened 2019-10-24
Install AWS Elastic Container Registry Login helper on Worker image

bkgoksel:
This is a package that allows the Docker client of the worker to use
private AWS Docker image repositories via setting simple environment
variables.

Preinstalling this on our image means any downstream user of Codalab can
launch a worker that can use their private registries without having to
do any modifications beforehands beyond setting environment variables at
 runtime.

### DIFF
--- a/docker/dockerfiles/Dockerfile.worker
+++ b/docker/dockerfiles/Dockerfile.worker
@@ -1,3 +1,12 @@
+FROM golang as ecr-login-installation
+# Install the Amazon ECR Docker Credential Helper
+# Use Docker multi-stage builds to get rid of intermediate layers related to installation
+
+ENV GOPATH /root/go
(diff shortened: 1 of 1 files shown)


percyliang (commented):
It's slightly unfortunate that all this AWS stuff is now part of the main worker Docker image, which is kind of irrelevant for Azure, for which we'd presumably need something else.  But having multiple Docker worker images would make things more complicated.  Just wanted to bring this up for discussion.

bkgoksel:
Yeah, I was also considering making an aws-worker image that would take the worker image as base and just install the login helper on top, but the modifications to get it to work felt small enough that I went ahead with this approach.

Given that there are probably only up to 3 cloud providers we'd support in this way, I feel like this won't be a huge issue, but we could also go with having a separate aws-worker image, and either maintained in the main codalab repo or elsewhere for downstream users.

percyliang (approved):

PR #1569, opened 2019-10-24
Add a mechanism to configure worker-manager specific values

bkgoksel:
Before this PR, configurable values specific to a worker manager type
like AWS Batch had to be hard coded in the respective worker manager
files.

This PR allows worker manager types to define their own config
structures, and the main worker manager wrapper only takes in a file path
and passes it on to them for them to parse as they see fit.

Another way to do this might have been defining parsers in the specific
worker manager subclasses, then importing these parsers in the wrapper
class and adding them as subcommands to the worker-manager command.

That would enable specific worker manager types to define their own
command line arguments without having to modify the wrapper class, but
it would be a bit more complicated to code compared to offloading
everything to JSON files.


docker/dockerfiles/Dockerfile.worker
```diff
@@ -1,3 +1,12 @@
+FROM golang as ecr-login-installation
```
bkgoksel, line 1: Changes to this file are irrelevant, I just based this branch off of my other PR branch since that'll hopefully be merged soon.

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -32,11 +63,6 @@ def get_worker_jobs(self):
 
     def start_worker_job(self):
         image = 'codalab/worker:' + os.environ.get('CODALAB_VERSION', 'latest')
-        logger.debug('Starting worker with image {}'.format(image))
```
percyliang, line 35: Why remove the debug statement?

codalab/worker_manager/main.py
```diff
@@ -45,6 +44,11 @@ def main():
         help='Minimum time to wait between launching workers',
         default=1 * 60,
     )
+    parser.add_argument(
+        '--queue-config-file',
+        help='Path to a JSON config file that sets config values specific '
+        'to your Worker Manager type. (eg AWS Batch)',
```
percyliang: 'to your worker manager type (e.g., AWS Batch)'

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -7,20 +8,50 @@
...
+        self.queue = config.get('queue', self.DEFAULT_CONFIG['queue'])
+
+
 class AWSBatchWorkerManager(WorkerManager):
     def __init__(self, args):
         super().__init__(args)
-        if not args.queue:
-            raise Exception('Missing queue for AWS Batch')
-        self.batch_client = boto3.client('batch', region_name='us-east-1')
+        self.config = AWSBatchWorkerManagerConfig(args.queue_config_file)
```
percyliang: rename this just `config_file`

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -7,20 +8,50 @@
...
+            with open(config_filename, 'r') as config_file:
+                config = json.load(config_file)
+        except (IOError, ValueError, json.decoder.JSONDecodeError) as ex:
+            logger.error(
+                "Problem loading config file [%s]: %s. Using the default config.",
+                config_filename,
+                str(ex),
+            )
+            config = AWSBatchWorkerManagerConfig.DEFAULT_CONFIG
+        self.region = config.get('region', self.DEFAULT_CONFIG['region'])
```
percyliang: Can we make this a bit more automated?

percyliang (changes requested):
Can you describe how the config file will be specified in a deployment...since I didn't see any change in the docker compose files....

codalab/worker_manager/aws_batch_worker_manager.py
bkgoksel: It just felt strange to me that we print the image as a debug statement. Doesn't feel too informative to me. Will re-add with the ID as well.

bkgoksel:
Added a way to define config files. As usual there's not a good way to define optional/conditional volumes on Docker compose, so the best way is to define a default file that will always be mounted, and allow users to overwrite the path to that file. This is still not great, but best I could think of right now.

docker/compose_files/files/gpu_worker_manager_config.json
```diff
@@ -0,0 +1,8 @@
+{
```
percyliang: Are these meant to be samples?

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -7,20 +8,47 @@
 logger = logging.getLogger(__name__)
 
 
+class AWSBatchWorkerManagerConfig:
+    DEFAULT_CONFIG = {
```
percyliang: We don't need this since we already have JSON files in the docker compose files.

percyliang (commented):
Currently, CodaLab instances are configured via environment variables (and optional command-line arguments), which is nice because everything is in one place and there's one mechanism for configuration.  But this introduces a config file, which I feel like makes things more complicated (and you can't easily change one part)...can we not still define everything via environment variables?

bkgoksel:
Ok I agree it's better to not get files involved here. I think I have decent-ish way to handle this inside the worker manager classes.
But then the next problem is that we need to maintain two separate argument lists in sync (one in `codalab_service` and one in `worker_manager`), this kind of balloons the argument list in `codalab_service` a bit, but maybe that's unavoidable for now.

codalab/worker_manager/main.py
```diff
@@ -45,6 +37,19 @@ def main():
         help='Minimum time to wait between launching workers',
         default=1 * 60,
     )
+    subparsers = parser.add_subparsers(
+        title='Worker Manager to run',
+        description='Which worker manager to run (AWS Batch etc.)',
+        dest='worker_manager_name',
+    )
+
+    # Each worker manager class defines its NAME, which is the subcommand the users user
```
bkgoksel: nit: typo in comment

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -8,19 +10,49 @@
...
+        parser.add_argument(
+            '--user', type=str, default='root', help='User to run the Batch jobs as'
+        )
+        parser.add_argument(
+            '--job-queue',
+            type=str,
+            default='codalab-batch-cpu',
+            help='Name of the AWS Batch job queue to use',
+        )
+
```
percyliang, line 42: Don't we need to return something from a function called `get_args`?

percyliang: I think this function should be called `add_parser`.

docker/compose_files/docker-compose.yml
```diff
@@ -27,11 +28,20 @@ x-codalab-env: &codalab-env
...
-  - CODALAB_WORKER_MANAGER_MAX_GPU_WORKERS=${CODALAB_WORKER_MANAGER_MAX_GPU_WORKERS}
+  - CODALAB_WORKER_MANAGER_IDLE_SECONDS=${CODALAB_WORKER_MANAGER_IDLE_SECONDS}
+  - CODALAB_WORKER_MANAGER_SECONDS_BETWEEN_WORKERS=${CODALAB_WORKER_MANAGER_SECONDS_BETWEEN_WORKERS}
+  - CODALAB_WORKER_MANAGER_AWS_REGION=${CODALAB_WORKER_MANAGER_AWS_REGION}
+  - CODALAB_WORKER_MANAGER_AWS_BATCH_JOB_DEFINITION_NAME=${CODALAB_WORKER_MANAGER_AWS_BATCH_JOB_DEFINITION_NAME}
+  - CODALAB_WORKER_MANAGER_AWS_DEFAULT_CPUS=${CODALAB_WORKER_MANAGER_AWS_DEFAULT_CPUS}
+  - CODALAB_WORKER_MANAGER_AWS_DEFAULT_MEMORY_MB=${CODALAB_WORKER_MANAGER_AWS_DEFAULT_MEMORY_MB}
   - CODALAB_WORKER_MANAGER_CPU_QUEUE=${CODALAB_WORKER_MANAGER_CPU_QUEUE}
   - CODALAB_WORKER_MANAGER_GPU_QUEUE=${CODALAB_WORKER_MANAGER_GPU_QUEUE}
-  - HOSTNAME=${HOSTNAME}
```
percyliang, line 34: Why is this deleted?

docker/compose_files/docker-compose.yml
```diff
@@ -76,13 +86,13 @@ services:
 
   worker-manager-cpu:
     image: codalab/server:${CODALAB_VERSION}
-    command: cl-worker-manager --worker-manager-type ${CODALAB_WORKER_MANAGER_TYPE} --max-workers ${CODALAB_WORKER_MANAGER_MAX_CPU_WORKERS} --server ${CODALAB_SERVER} --queue ${CODALAB_WORKER_MANAGER_CPU_QUEUE} --search request_gpus=0
+    command: cl-worker-manager ${CODALAB_WORKER_MANAGER_TYPE} --min-workers ${CODALAB_WORKER_MANAGER_MIN_CPU_WORKERS} --max-workers ${CODALAB_WORKER_MANAGER_MAX_CPU_WORKERS} --server ${CODALAB_SERVER} --job-queue ${CODALAB_WORKER_MANAGER_CPU_QUEUE} --worker-tag ${CODALAB_WORKER_MANAGER_CPU_TAG} --sleep-time ${CODALAB_WORKER_MANAGER_SECONDS_BETWEEN_WORKERS} --worker-idle-seconds ${CODALAB_WORKER_MANAGER_IDLE_SECONDS} --region ${CODALAB_WORKER_MANAGER_AWS_REGION} --job-definition-name ${CODALAB_WORKER_MANAGER_AWS_BATCH_JOB_DEFINITION_NAME} --cpus ${CODALAB_WORKER_MANAGER_AWS_DEFAULT_CPUS} --memory-mb ${CODALAB_WORKER_MANAGER_AWS_DEFAULT_MEMORY_MB} --search request_gpus=0
```
percyliang: Can we wrap this command?

docker/compose_files/docker-compose.yml
```diff
@@ -76,13 +86,13 @@ services:
...
     <<: *codalab-base
     <<: *codalab-server
 
   worker-manager-gpu:
     image: codalab/server:${CODALAB_VERSION}
-    command: cl-worker-manager --worker-manager-type ${CODALAB_WORKER_MANAGER_TYPE} --max-workers ${CODALAB_WORKER_MANAGER_MAX_GPU_WORKERS} --server ${CODALAB_SERVER} --queue ${CODALAB_WORKER_MANAGER_GPU_QUEUE} --search request_gpus=1
+    command: cl-worker-manager ${CODALAB_WORKER_MANAGER_TYPE} --min-workers ${CODALAB_WORKER_MANAGER_MIN_GPU_WORKERS} --max-workers ${CODALAB_WORKER_MANAGER_MAX_GPU_WORKERS} --server ${CODALAB_SERVER} --job-queue ${CODALAB_WORKER_MANAGER_GPU_QUEUE} --worker-tag ${CODALAB_WORKER_MANAGER_GPU_TAG} --sleep-time ${CODALAB_WORKER_MANAGER_SECONDS_BETWEEN_WORKERS} --worker-idle-seconds ${CODALAB_WORKER_MANAGER_IDLE_SECONDS} --region ${CODALAB_WORKER_MANAGER_AWS_REGION} --job-definition-name ${CODALAB_WORKER_MANAGER_AWS_BATCH_JOB_DEFINITION_NAME} --cpus ${CODALAB_WORKER_MANAGER_AWS_DEFAULT_CPUS} --memory-mb ${CODALAB_WORKER_MANAGER_AWS_DEFAULT_MEMORY_MB} --search request_gpus=1
```
percyliang: And this one

percyliang (changes requested):
Mostly looks good - just a few suggestions.

docker/compose_files/docker-compose.yml
bkgoksel: Just moved to the top of the list as the only non-codalab env var.

docker/compose_files/docker-compose.yml
```diff
@@ -76,13 +86,41 @@ services:
...
+    command: |
+      cl-worker-manager
+      --server ${CODALAB_SERVER}
+      --min-workers ${CODALAB_WORKER_MANAGER_MIN_CPU_WORKERS}
+      --max-workers ${CODALAB_WORKER_MANAGER_MAX_CPU_WORKERS}
+      --search request_gpus=0
+      --worker-tag ${CODALAB_WORKER_MANAGER_CPU_TAG}
+      --sleep-time ${CODALAB_WORKER_MANAGER_SECONDS_BETWEEN_WORKERS}
+      --worker-idle-seconds ${CODALAB_WORKER_MANAGER_IDLE_SECONDS}
+      ${CODALAB_WORKER_MANAGER_TYPE}
```
bkgoksel, line 98: Args that apply to all worker manager types need to go before the worker manager type and args that are specific to this worker manager type need to go after the argument. 

The problem with this setup is that if we ever add another worker manager type in the future, we'd need to add a separate entry in `docker-compose.yml` for it with the correct argument structure for it + we'd need to add all its arguments as arguments to `codalab_service.py` as well.

I don't know if there's an ideal solution here. Maybe making `--worker-manager-args` a single freeform string argument to `codalab_service` and passing just that in `docker-compose.yml` to the worker manager? Or getting rid of the subcommand structure for the `worker-manager` types and adding every argument for every type of worker manager into the same argument namespace and passing in everything that's set. 

I also feel like this decision can be  postponed to when a second worker manager type is about to be added.

codalab/worker_manager/worker_manager.py
```diff
@@ -46,6 +46,16 @@ class WorkerManager(object):
       burden of deciding on the user.
     """
 
+    # Subcommand name to use for this worker manager type
+    NAME = 'worker-manager'
+
+    @staticmethod
+    def add_parser(subparsers):
+        """
+        Add any arguments specific to this worker manager to the given subparser
```
percyliang, line 56: It might be more user-friendly for the sub-classes to have a `add_arguments_to_subparser(subparser)` function, and then create the parser on the outside (which might help with sharing code).

docker/compose_files/docker-compose.yml
percyliang: Can the order of the arguments be documented?

codalab/worker_manager/worker_manager.py
bkgoksel: So the actual sub-command is created by the `main` script and each sub-class just adds its arguments to this sub-parser, rather than creating the sub-parser in the sub-class as well?

docker/compose_files/docker-compose.yml
percyliang: I guess there's going to be Slurm and Azure Batch...yeah, it's a bit unfortunate that we're going to get the 2 x 3 explosion for CPU and GPU types.  Actually, maybe a better solution would be to have the worker manager read the environment variables directly, so that we can just have two entries here.

codalab/worker_manager/worker_manager.py
percyliang: Yes, that way we can enforce more standardization.

docker/compose_files/docker-compose.yml
bkgoksel: I still think it might be better to decide this when the next set of worker managers are actually added and we can decide what's more ergonomic with a concrete case at hand.

percyliang (approved):
Great, thanks!

Issue #1504, opened 2019-10-15
Make transactions on backend atomic


epicfaace:
Use sqlalchemy's `session` as documented here: https://docs.sqlalchemy.org/en/13/orm/session_transaction.html

percyliang:
Do we have any bugs currently as a result of this?

epicfaace:
No, but I'd imagine it might end up creating some silent bugs / inconsistencies.

PR #1571, opened 2019-10-27
Fix slow transition issue due to doublecounting the resource of run_uuid

candicegjing:
The issue in logic here https://github.com/codalab/codalab-worksheets/blob/codalab/server/bundle_manager.py#L335-L344 is: 
double-counting the cpu/gpu/memory resources => negative values of resources on available workers => no available workers => jobs stuck in staged state. 
Here is the logging messages that shows the transition of `cpu` and `memory_bytes`:
The value of `cpu` is substracted 1 a time. 
The value of `memory_bytes` is substracted `2g` a time. 
```
########################################Current bundle uuid = 0x56105637af6c4d8c8c20198bfdd5f884
2019-10-26 23:33:43,758 get_user_parallel_run_quota_left takes time = 0.007198333740234375
2019-10-26 23:33:43,758 ORG_workers_list = [{'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': 8, 'gpus': 0, 'memory_bytes': 17179869184}
2019-10-26 23:33:43,779 [_deduct_worker_resources] uuid = 0x12895b688fce44e7b7dc669a818d304c, bundle_resources cpu = 1, memory = 0
2019-10-26 23:33:43,779 [_deduct_worker_resources] worker = {'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': 8, 'gpus': 0, 'memory_bytes': 17179869184}
2019-10-26 23:33:43,800 [_deduct_worker_resources] uuid = 0x04f18d7b5e3140a5be2aa95db3d563d0, bundle_resources cpu = 1, memory = 0
2019-10-26 23:33:43,800 [_deduct_worker_resources] worker = {'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': 7, 'gpus': 0, 'memory_bytes': 15032385536}
2019-10-26 23:33:43,821 [_deduct_worker_resources] uuid = 0x9fa54604662c41ecbc72aa38b112ea9b, bundle_resources cpu = 1, memory = 0
2019-10-26 23:33:43,821 [_deduct_worker_resources] worker = {'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': 6, 'gpus': 0, 'memory_bytes': 12884901888}
...
2019-10-26 23:33:43,843 [_deduct_worker_resources] uuid = 0xbb03d8ed12504591a19b95d050278a72, bundle_resources cpu = 1, memory = 0
2019-10-26 23:33:43,935 [_deduct_worker_resources] worker = {'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': 1, 'gpus': 0, 'memory_bytes': 2147483648}
2019-10-26 23:33:43,958 [_deduct_worker_resources] uuid = 0xd757daa59b18418da11b71cb1e41a90a, bundle_resources cpu = 1, memory = 0
2019-10-26 23:33:43,958 [_deduct_worker_resources] worker = {'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': 0, 'gpus': 0, 'memory_bytes': 0}
2019-10-26 23:33:43,980 [_deduct_worker_resources] uuid = 0x6ea7d0acdc994a1c9e22aa1e2fe691f6, bundle_resources cpu = 1, memory = 0
2019-10-26 23:33:43,980 [_deduct_worker_resources] worker = {'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': -1, 'gpus': 0, 'memory_bytes': -2147483648}
2019-10-26 23:33:43,980 deduct_worker_resources Workers_list = [{'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': -2, 'gpus': 0, 'memory_bytes': -4294967296}
2019-10-26 23:33:43,996 Bundle_resources memory = 2147483648, cpu = 1, disk = 0
2019-10-26 23:33:43,997 [server/bundle_manager.py] _filter_and_sort_workers Check filter out MEMORY = [{'user_id': '0', 'worker_id': 'Stanfords-MacBook-Pro.local(31814)', 'tag': 'mojolady', 'cpus': -2, 'gpus': 0, 'memory_bytes': -4294967296}
2019-10-26 23:33:43,997 [server/bundle_manager.py]  Caution: this is a special queue request tag = <_sre.SRE_Match object; span=(0, 13), match='tag=mojolady2'>
```
I added a variable to track those `run_uuids`, so that each of them will be calculated once. Then there will be no double-counting and no negative values generated from `def _deduct_worker_resources()` function.

### DIFF
--- a/codalab/server/worker_info_accessor.py
+++ b/codalab/server/worker_info_accessor.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+import copy
 
 
 class WorkerInfoAccessor(object):
@@ -25,7 +26,11 @@ class WorkerInfoAccessor(object):
         return None
 
     def user_owned_workers(self, user_id):
-        return list(self._user_id_to_workers[user_id])
+        # A deep copy is necessary here due to the following facts:
+        # 1. assignment statements in Python do not copy objects, meaning it generates a shallow copy
+        # 2. deep copy is only necessary for compound objects which contains other objects, like lists or class instances
+        # 3. a deep copy will guarantee that one can change one copy without changing the other
+        return list(copy.deepcopy(worker) for worker in self._user_id_to_workers[user_id])
 
     def remove(self, worker):
         self._workers.remove(worker)
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -293,7 +293,7 @@ class BundleModel(object):
 
     def search_bundles(self, user_id, keywords):
         """
-        Returns a  bundle search result dict where:
+        Returns a bundle search result dict where:
             result: list of bundle uuids matching search criteria in order
                           specified for bundle searches
                     single number value for aggregate searches(.count, .sum)


bkgoksel:
I'm not sure I understand where the double-counting is happening. Are there duplicate entries in the `worker['run_uuids']` list? Or is it that we're deducting resources of `FAILED` and `KILLED` bundles because they're still in `worker['run_uuids']` and your change filters those out explicitly?

codalab/server/bundle_manager.py
```diff
@@ -327,6 +330,11 @@ def _deduct_worker_resources(self, workers_list):
         """
         for worker in workers_list:
             for uuid in worker['run_uuids']:
+                # If the resource of uuid has been subtracted before, continue to next uuid.
+                # This logic will ensure that the resource of each uuid will only be subtracted once.
+                if uuid in self._track_run_uuids:
+                    continue
+                self._track_run_uuids.add(uuid)
                 bundle = self._model.get_bundle(uuid)
```
bkgoksel: I think we shouldn't add to the global state here and add a bunch more functions.
Converting the UUID list to a set in the beginning, and then locally skipping failed/killed bundles should be enough like the following:
```suggestion
    worker_run_uuids = set(worker['run_uuids'])
    for uuid in worker_run_uuids:
        bundle = self._model.get_bundle(uuid)
        if bundle.state in (State.KILLED, State.FAILED):
            continue
```

bkgoksel (changes requested):

codalab/server/bundle_manager.py
```diff
@@ -327,6 +330,11 @@ def _deduct_worker_resources(self, workers_list):
         """
         for worker in workers_list:
             for uuid in worker['run_uuids']:
+                # If the resource of uuid has been subtracted before, continue to next uuid.
```
bkgoksel: We should also check whether this bundle still exists here. Currently there's a potential race condition where the bundle might get deleted while the bundle manager execution is right between the lines 314 and 316 of this file (ie after we fetch the workers list from the DB and before we call deduct worker resources)

This race condition is causing #1572 

I guess ideally we'd use something like a session (as suggested by #1504) to make sure the bundles don't get deleted while we're in the middle of reading that table, but the alternate solution of checking for bundle existence here seems good enough for now.

candicegjing:
@bkgoksel We currently schedule jobs using the following logic:
1. in `_schedule_run_bundles_on_workers()`, we loop through all the `staged` bundles at here: [Line 304](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L304.) For each `staged` bundle, we first check if it is `user_owned`. If yes, we create a `workers_list` that only belongs to this particular user. Otherwise, we return the default `workers_list`. 
2. Then based on the `workers_list` we got from 1st, we loop through each worker and for each worker, we loop through each running bundle at line:[Line 329](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L329). Based on the `uuid`, we fetch the metadata of current bundle from database and generated `bunlde_resources` at [Line 331](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L331). Next, we subtract each `bunlde_resources`'s `cpus`, `gpus`, and `memory` value from the current worker.

Then the double-counting logic happens as follows:
Suppose there is only 1 worker which contains 2 CPUs and 6g memory, and 1 user submitted 10 jobs. Based on the above logic, initially suppose there are 2 `staged` jobs (8 jobs left in `staged` state) can proceed as there are plenty of computing resources available on this worker. Then based on 1st, we start to loop through these 8 `staged` jobs. Since there is only one worker, 2 bundles are running on this worker (and they may run for a while), based on 2nd, we subtract these two running bundles `bundle_resources` from the current worker 2 times. This is where double-counting happens. Then the number of `cpus` and `memory_bytes` will be easily changed to a negative value, which indicates insufficient computing resources on the worker. Then no available worker will be returned for those `staged` jobs for a while.
We can extend the above pattern to multiple workers as well. Let me know if this is clear. If I overlooked other details, please let me know.

bkgoksel:
> @bkgoksel We currently schedule jobs using the following logic:
> 
>     1. in `_schedule_run_bundles_on_workers()`, we loop through all the `staged` bundles at here: [Line 304](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L304.) For each `staged` bundle, we first check if it is `user_owned`. If yes, we create a `workers_list` that only belongs to this particular user. Otherwise, we return the default `workers_list`.
> 
>     2. Then based on the `workers_list` we got from 1st, we loop through each worker and for each worker, we loop through each running bundle at line:[Line 329](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L329). Based on the `uuid`, we fetch the metadata of current bundle from database and generated `bunlde_resources` at [Line 331](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L331). Next, we subtract each `bunlde_resources`'s `cpus`, `gpus`, and `memory` value from the current worker.
> 
> 
> Then the double-counting logic happens as follows:
> Suppose there is only 1 worker which contains 2 CPUs and 6g memory, and 1 user submitted 10 jobs. Based on the above logic, initially suppose there are 2 `staged` jobs (8 jobs left in `staged` state) can proceed as there are plenty of computing resources available on this worker. Then based on 1st, we start to loop through these 8 `staged` jobs. Since there is only one worker, 2 bundles are running on this worker (and they may run for a while), based on 2nd, we subtract these two running bundles `bundle_resources` from the current worker 8 times. This is where double-counting happens. Then the number of `cpus` and `memory_bytes` will be easily changed to a negative value, which indicates insufficient computing resources on the worker. Then no available worker will be returned for those `staged` jobs for a while.
> We can extend the above pattern to multiple workers as well. Let me know if this is clear. If I overlooked other details, please let me know.

I thought that we fetch a fresh workers list for each bundle we loop over (line 306 and [line 314](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L314)) My understanding is that the `workers.user_owned_workers()` function returns a brand new list every time [looking at its definition.](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L27)

Is it that changes we make to the local `workers_list` variable actually change the values for the future iterations as well?

candicegjing:
> > @bkgoksel We currently schedule jobs using the following logic:
> > ```
> > 1. in `_schedule_run_bundles_on_workers()`, we loop through all the `staged` bundles at here: [Line 304](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L304.) For each `staged` bundle, we first check if it is `user_owned`. If yes, we create a `workers_list` that only belongs to this particular user. Otherwise, we return the default `workers_list`.
> > 
> > 2. Then based on the `workers_list` we got from 1st, we loop through each worker and for each worker, we loop through each running bundle at line:[Line 329](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L329). Based on the `uuid`, we fetch the metadata of current bundle from database and generated `bunlde_resources` at [Line 331](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L331). Next, we subtract each `bunlde_resources`'s `cpus`, `gpus`, and `memory` value from the current worker.
> > ```
> > 
> > 
> > Then the double-counting logic happens as follows:
> > Suppose there is only 1 worker which contains 2 CPUs and 6g memory, and 1 user submitted 10 jobs. Based on the above logic, initially suppose there are 2 `staged` jobs (8 jobs left in `staged` state) can proceed as there are plenty of computing resources available on this worker. Then based on 1st, we start to loop through these 8 `staged` jobs. Since there is only one worker, 2 bundles are running on this worker (and they may run for a while), based on 2nd, we subtract these two running bundles `bundle_resources` from the current worker 8 times. This is where double-counting happens. Then the number of `cpus` and `memory_bytes` will be easily changed to a negative value, which indicates insufficient computing resources on the worker. Then no available worker will be returned for those `staged` jobs for a while.
> > We can extend the above pattern to multiple workers as well. Let me know if this is clear. If I overlooked other details, please let me know.
> 
> I thought that we fetch a fresh workers list for each bundle we loop over (line 306 and [line 314](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L314)) My understanding is that the `workers.user_owned_workers()` function returns a brand new list every time [looking at its definition.](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L27)
> 
> Is it that changes we make to the local `workers_list` variable actually change the values for the future iterations as well?

**There was a typo in my previous comments. Updated here from 8 times to 2 times (assuming there are 2 running bundles) -> *based on 2nd, we subtract these two running bundles bundle_resources from the current worker `2` times.***

I initially had the same thoughts as you, but once I saw the following logging message, I realized that there is something off in the logic.   
The following is the logging message I printed out from code when there are 2 running bundles on the target worker. Check for the value of `cpus` and `memory_bytes`. The value `cpus` got subtracted 1 for twice (there are 2 bundles running) from `worker` and the same for `memory_bytes`. 
Note that line of `Current staged bundle uuid` is printed out right after the for loop on bundles in function `_schedule_run_bundles_on_workers()`  
```
########################################Current staged bundle uuid = 0xd9a121e319b0427aa030b2d00aa27b3d
2019-10-28 20:11:37,636 get_user_parallel_run_quota_left takes time = 0.2799339294433594
2019-10-28 20:11:37,637 ORG_workers_list = [{'user_id': '0', 'worker_id': 'vm-clws-dev-worker-0', 'tag': None, 'cpus': -6, 'gpus': 0, 'memory_bytes': -6791790592, ...}]
2019-10-28 20:11:37,657 [_deduct_worker_resources] uuid = 0x0597c81e5e644862a04d6d8032543822, bundle_resources cpu = 1, memory = 2147483648
2019-10-28 20:11:37,657 [_deduct_worker_resources] worker_id = vm-clws-dev-worker-0, cpus = -6, memory = -6791790592, run_uuids = ['0x0597c81e5e644862a04d6d8032543822', '0x8032be2b03b14bc8aa635f9b4feb3cc1']
2019-10-28 20:11:37,677 [_deduct_worker_resources] uuid = 0x8032be2b03b14bc8aa635f9b4feb3cc1, bundle_resources cpu = 1, memory = 2147483648
2019-10-28 20:11:37,677 [_deduct_worker_resources] worker_id = vm-clws-dev-worker-0, cpus = -7, memory = -8939274240, run_uuids = ['0x0597c81e5e644862a04d6d8032543822', '0x8032be2b03b14bc8aa635f9b4feb3cc1']
2019-10-28 20:11:37,677 Workers_list = [{'user_id': '0', 'worker_id': 'vm-clws-dev-worker-0', 'tag': None, 'cpus': -8, 'gpus': 0, 'memory_bytes': -11086757888, ...}]
2019-10-28 20:11:37,692 Bundle_resources memory = 2147483648, cpu = 1, disk = 0 
2019-10-28 20:11:37,692 [server/bundle_manager.py] _filter_and_sort_workers Check CPU = []
2019-10-28 20:11:37,692 [server/bundle_manager.py] _filter_and_sort_workers Check GPU = []
2019-10-28 20:11:37,692 [server/bundle_manager.py] _filter_and_sort_workers Check filter out MEMORY = []

########################################Current staged bundle uuid = 0x3e29040b08d544cfbf162b5ae7185efd
2019-10-28 20:11:37,692 get_user_parallel_run_quota_left takes time = 0.3358433246612549
2019-10-28 20:11:37,692 ORG_workers_list = [{'user_id': '0', 'worker_id': 'vm-clws-dev-worker-0', 'tag': None, 'cpus': -8, 'gpus': 0, 'memory_bytes': -11086757888, ...}]
2019-10-28 20:11:37,714 [_deduct_worker_resources] uuid = 0x0597c81e5e644862a04d6d8032543822, bundle_resources cpu = 1, memory = 2147483648
2019-10-28 20:11:37,714 [_deduct_worker_resources] worker_id = vm-clws-dev-worker-0, cpus = -8, memory = -11086757888, run_uuids = ['0x0597c81e5e644862a04d6d8032543822', '0x8032be2b03b14bc8aa635f9b4feb3cc1']
2019-10-28 20:11:37,737 [_deduct_worker_resources] uuid = 0x8032be2b03b14bc8aa635f9b4feb3cc1, bundle_resources cpu = 1, memory = 2147483648
2019-10-28 20:11:37,737 [_deduct_worker_resources] worker_id = vm-clws-dev-worker-0, cpus = -9, memory = -13234241536, run_uuids = ['0x0597c81e5e644862a04d6d8032543822', '0x8032be2b03b14bc8aa635f9b4feb3cc1']
2019-10-28 20:11:37,737 Workers_list = [{'user_id': '0', 'worker_id': 'vm-clws-dev-worker-0', 'tag': None, 'cpus': -10, 'gpus': 0, 'memory_bytes': -15381725184, ...}]
2019-10-28 20:11:37,754 Bundle_resources memory = 2147483648, cpu = 1, disk = 0 
2019-10-28 20:11:37,755 [server/bundle_manager.py] _filter_and_sort_workers Check CPU = []
2019-10-28 20:11:37,755 [server/bundle_manager.py] _filter_and_sort_workers Check GPU = []
2019-10-28 20:11:37,755 [server/bundle_manager.py] _filter_and_sort_workers Check filter out MEMORY = []
```

bkgoksel:
I see. This is really strange and thanks for catching it! It's a very subtle bug (and I bet it wasn't a huge problem before because the filtering code wasn't actually applying the memory filter for the jobs with the default memory request, but after my changes we started enforcing that) and hopefully fixing this will unblock our bundle manager again.

As for the solution, I think instead of having a global set so we visit every bundle once, we should go to the source of the problem and make sure we just work on a copy of the workers list. I think it'd be nicer to avoid adding another instance-level variable and associated methods and try to solve the issue locally.

Also would be good to add the bundle existence check I mentioned we need in the earlier comment, and this should be good to go!

candicegjing:
> I see. This is really strange and thanks for catching it! It's a very subtle bug (and I bet it wasn't a huge problem before because the filtering code wasn't actually applying the memory filter for the jobs with the default memory request, but after my changes we started enforcing that) and hopefully fixing this will unblock our bundle manager again.
> 
> As for the solution, I think instead of having a global set so we visit every bundle once, we should go to the source of the problem and make sure we just work on a copy of the workers list. I think it'd be nicer to avoid adding another instance-level variable and associated methods and try to solve the issue locally.
> 
> Also would be good to add the bundle existence check I mentioned we need in the earlier comment, and this should be good to go!

I am still trying to figure out why the worker keeps getting updated all the time.

bkgoksel:
@candicegjing Ah, I think I figured out why this is happening. If you look at [the definition of the `WorkerInfoAccessor` class](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L12), the `_user_id_to_workers` field is a `Dict[str, List[Dict[str, str]]]` (ie a dict from user ids to a list of dicts that represent the workers).

The [`user_owned_workers` method](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L27) creates a brand new list of the worker dicts of the given user and returns that. So the modifications to the list itself don't affect anything else, but since the elements of this list are the worker dicts, any change made to these dicts actually affect the `_user_id_to_workers` of the `WorkerInfoAccessor`.

I think the solution here would be to have the [`user_owned_workers` method](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L27) return a deep copied list.

candicegjing:
> @candicegjing Ah, I think I figured out why this is happening. If you look at [the definition of the `WorkerInfoAccessor` class](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L12), the `_user_id_to_workers` field is a `Dict[str, List[Dict[str, str]]]` (ie a dict from user ids to a list of dicts that represent the workers).
> 
> The [`user_owned_workers` method](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L27) creates a brand new list of the worker dicts of the given user and returns that. So the modifications to the list itself don't affect anything else, but since the elements of this list are the worker dicts, any change made to these dicts actually affect the `_user_id_to_workers` of the `WorkerInfoAccessor`.
> 
> I think the solution here would be to have the [`user_owned_workers` method](https://github.com/codalab/codalab-worksheets/blob/6ac009e34316d030125e521a8471539922412c76/codalab/server/worker_info_accessor.py#L27) return a deep copied list.

I think you are right! I was also looking at this function and found something else in the same time. I'll make the changes and then you can take a look.

codalab/server/bundle_manager.py
```diff
@@ -581,10 +581,9 @@ def _schedule_run_bundles(self):
         self._bring_offline_stuck_running_bundles(workers)
         self._fail_on_too_many_resources()
         self._acknowledge_recently_finished_bundles(workers)
-
-        # Schedule, preferring user-owned workers.
-        self._schedule_run_bundles_on_workers(workers, user_owned=True)
-        self._schedule_run_bundles_on_workers(workers, user_owned=False)
+        
+        # Schedule bundles to run on workers
+        self._schedule_run_bundles_on_workers(workers)
```
candicegjing: My understanding of `user_owned` variable is used to schedule jobs that are requested to run on personal workers first. However, after reading the logic inside `self._schedule_run_bundles_on_workers()`, I noticed that it always loop through the staged bundles fetched from the database without actually scheduling those `user_owned` bundles first. Then there are several unnecessary loops that happened before the real jobs start. I remove `user_owned` logic and created a new list named with `ordered_staged_bundles` to make sure those user_owned bundles will be scheduled first.

codalab/server/worker_info_accessor.py
```diff
@@ -24,8 +26,19 @@ def worker_with_id(self, user_id, worker_id):
                 return worker
         return None
 
-    def user_owned_workers(self, user_id):
-        return list(self._user_id_to_workers[user_id])
+    def user_owned_workers(self, user_id, request_queue=None):
+        workers_list = self._user_id_to_workers[user_id]
```
candicegjing: Apart from changing the shallow copy to deep copy, I also changed the logic of this function to be: 
1. If `request_queue` is specified for this bundle, the function will return all the personal workers that belong to this user_id.
2. If `request_queue` is not specified for this bundle, the function will return all the public workers.

codalab/server/bundle_manager.py
bkgoksel: The `user_owned` here refers to the workers, not the bundles. So, the desired logic is that for any bundle submitted by a user, we should first try to schedule that bundle on a worker that also belongs to that user before trying any of the Codalab workers. This is regardless of whether the user requested a queue or not in the bundle.

I think there's room to refactor this like you started, but we need to keep the logic of preferring user-owned workers over codalab workers for all bundles. Maybe you can try to move that logic to the sort and filter workers function.

But I think it's also fine to keep this part unchanged for the sake of getting this PR finished sooner.

codalab/server/worker_info_accessor.py
bkgoksel: I find it a bit strange that this function does these two separate things. I think the filtering by tag belongs more to the filter function than here (and this function can stay a short, simple function)

codalab/server/bundle_manager.py
```diff
@@ -363,17 +374,6 @@ def _filter_and_sort_workers(workers_list, bundle, bundle_resources):
...
-        # Filter by tag.
-        request_queue = bundle.metadata.request_queue
-        if request_queue:
-            tagm = re.match('tag=(.+)', request_queue)
-            if tagm:
-                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
-            else:
-                # We don't know how to handle this type of request queue
-                # argument.
-                return []
```
bkgoksel: We should add a logging statement here about how we got a malformed `request_queue`

codalab/server/bundle_manager.py
```diff
@@ -295,15 +295,26 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
         """
-        for bundle in self._model.batch_get_bundles(state=State.STAGED, bundle_type='run'):
-            if user_owned:
-                workers_list = workers.user_owned_workers(bundle.owner_id)
+        staged_bundles = self._model.batch_get_bundles(state=State.STAGED, bundle_type='run')
+        # Returns all the staged bundles that are requested to run on personal workers
+        user_owned_bundles = [b for b in staged_bundles if b.metadata.request_queue is not None]
+        # Returns all the staged bundles that are requested to run on CodaLab-owned public workers
+        codalab_owned_bundles = [b for b in staged_bundles if b.metadata.request_queue is None]
+
```
bkgoksel: One way to refactor here is to try to get both the user AND codalab workers here, then filter and sort the user workers, and if that turns out to be empty after the filtering, then move on to the codalab workers (and also do the parallel run quota check)

bkgoksel: See my comment below for why I suggest this.

bkgoksel (changes requested):
We need to keep the old logic of always preferring user owned workers, for all bundles of that user. I also think the `user_owned_workers` logic should stay simple without getting the `request_queue` involved there. However there's probably a nice refactor that we can do there so we don't have to call the schedule function twice.

codalab/server/bundle_manager.py
candicegjing: @bkgoksel  Oh, I see. I originally thought `user_owned` means the bundles that were scheduled to run on personal workers, but it's good to get it cleared now. 
I also had the same thought for the desired logic. The logic that I changed here is to save one round of for loop. _With the logic of sorting on `staged_bundle.metadata.request_queue`, it schedules the jobs that were requested to run on personal workers first and then the jobs on public shared workers_, which I think is more intuitive in a way that we just need to go through all the `staged` jobs once and saves a bit of time. But if we don't want to do this right now, I can revert it back. Then we can at least deploy other parts asap.

codalab/server/bundle_manager.py
candicegjing: @bkgoksel Do you mean to change the return of this function? https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L171

codalab/server/bundle_manager.py
candicegjing: @bkgoksel If you see my above comments, please let me know!

bkgoksel: Hmm, so `request_queue` should not matter here. We want a given user's bundles to run on their own workers whenever that's possible, regardless of whether `request_queue` is `None` or not on the bundle. In other words, not setting `request_queue` doesn't mean scheduling the bundle to run on codalab workers.

What determines whether a run runs on a user-owned worker or public worker is solely whether there's a user-owned worker available at the time we're trying to run that bundle.

As a result, the thing to do here is not sorting the bundles by the `request_queue` parameter. I think the ideal solution would be something like for each staged bundle: 1. get all user owned workers and public workers. 2. sort and filter these workers, and while sorting give first precedence to user-owned workers. 3. In the end, if your top worker is NOT user owned, enforce the parallel run quota.

bkgoksel: But I think just as a way to move things forward, it might be a good idea to just make a tiny PR where the `user_owned_workers` method does a deepcopy and merge that in, so that prod can be unblocked.
Feel free to make a separate PR for that and tag me on it as well.

candicegjing: @bkgoksel I see. I just updated the code. Please check and let me know if it looks good!

bkgoksel (approved):
LGTM! Thanks for debugging this.

codalab/server/worker_info_accessor.py
```diff
@@ -25,7 +26,7 @@ def worker_with_id(self, user_id, worker_id):
         return None
 
     def user_owned_workers(self, user_id):
-        return list(self._user_id_to_workers[user_id])
+        return list(copy.deepcopy(worker) for worker in self._user_id_to_workers[user_id])
```
percyliang, line 33: Can the importance of this deepcopy (what is mutable, etc.) be documented somewhere?

candicegjing: @percyliang Just added.

PR #1574, opened 2019-10-28
Update sql query to fetch running bundles that are requested to run o…

### LINKED ISSUE #1573: parallel_run_quota doesn't work properly.
The `parallel_run_quota` doesn't seem to work correctly while I was testing the slow transition issue (#1254). 
When I was locally testing the slow transition changes I made, I noticed all the jobs (more than 10) were running at once. However, my user `parallel_run_quota` is 3 fetching from the database. So we need to correct this to avoid one user taking too many resources from the shared system.

candicegjing:
Fixed #1573.
When I was locally testing the slow transition (#1254) changes I made, I noticed all the jobs (more than 10) were running at once. However, my user parallel_run_quota is 3 fetching from the database. So we need to correct this to avoid one user taking too many resources from the shared system. 
I found out that the original sql query always returns empty entries. So I updated the SQL query to 
```
SELECT worker_run.run_uuid
From worker_run JOIN worker
ON worker.user_id=worker_run.user_id
WHERE worker_run.user_id = user_id (input parameter) and  worker.user_id = user_id (input parameter) and worker.tag IS NULL;
```
In order to get the `request_queue` information from tables, I used the `worker` table, which contains the `tag` column that indicates the `request_queue` from the user, to join with `worker_run` table. Then I will be able to get a list of running bundles that belong to the user_id passed from the input parameter and were requested to run on personal worker.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -2234,19 +2234,30 @@ class BundleModel(object):
     def get_user_parallel_run_quota_left(self, user_id):
         user_info = self.get_user_info(user_id)
         parallel_run_quota = user_info['parallel_run_quota']
+
+        # The following sqlalchemy query is equivalent to
+        # SELECT worker_run.run_uuid
+        # From worker_run JOIN worker
+        # ON worker.user_id=worker_run.user_id
+        # WHERE worker.tag IS NULL and worker.user_id = <user_id> and worker_run.user_id = <user_id>;
+        # Note: <user_id> is an input parameter
+        join = cl_worker_run.join(cl_worker, cl_worker_run.c.user_id == cl_worker.c.user_id)
+
         with self.engine.begin() as connection:
             # Get all the runs belonging to this user whose workers are not personal workers
             # of the user themselves
             active_runs = connection.execute(
-                select([cl_worker_run.c.run_uuid]).where(
+                select([cl_worker_run.c.run_uuid])
+                .select_from(join)
+                .where(
                     and_(
-                        cl_worker_run.c.run_uuid.in_(
-                            select([cl_bundle.c.uuid]).where(cl_bundle.c.owner_id == user_id)
-                        ),
-                        cl_worker_run.c.user_id != user_id,
+                        cl_worker_run.c.user_id == user_id,
+                        cl_worker.c.user_id == user_id,
+                        cl_worker.c.tag == None,
                     )
                 )
             ).fetchall()
+
         return parallel_run_quota - len(active_runs)
 
     def update_user_last_login(self, user_id):


codalab/model/bundle_model.py
```diff
@@ -2234,19 +2234,30 @@ def get_user_time_quota_left(self, user_id):
...
         user_info = self.get_user_info(user_id)
         parallel_run_quota = user_info['parallel_run_quota']
+
+        # The following sqlalchemy query is equivalent to
+        # SELECT worker_run.run_uuid
+        # From worker_run JOIN worker
+        # ON worker.user_id=worker_run.user_id
+        # WHERE worker.tag IS NULL and worker.user_id = <user_id> and worker_run.user_id = <user_id>;
+        # Note: <user_id> is an input parameter
+        join = cl_worker_run.join(cl_worker, cl_worker_run.c.user_id == cl_worker.c.user_id)
```
bkgoksel, line 2244: Can you rename this variable to a more descriptive name?

bkgoksel (changes requested):
2 things:
* Can you try to add a simple test for `parallel_run_quota`? I really should've done it at the time of making the PR but somehow skipped it 😿 
* We want to count the users' runs that are NOT running on workers that also belong to them. The old logic had a `cl_worker_run.c.user_id != user_id` to test for this. It seems like the new logic does the opposite and counts the number of runs a user has on their own workers. That looks like the wrong number. Am I misunderstanding the new logic?

candicegjing:
> 2 things:
> 
>     * Can you try to add a simple test for `parallel_run_quota`? I really should've done it at the time of making the PR but somehow skipped it 😿
> 
>     * We want to count the users' runs that are NOT running on workers that also belong to them. The old logic had a `cl_worker_run.c.user_id != user_id` to test for this. It seems like the new logic does the opposite and counts the number of runs a user has on their own workers. That looks like the wrong number. Am I misunderstanding the new logic?

Hmm, I thought this function `get_user_parallel_run_quota_left` is to get how many `parallel_run_quota` left for a particular user. Based on [Line 2250](https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L2250), `active_runs` should return the number of active running bundles on this worker. Then after it gets subtracted by original `parallel_run_quota`, it will return the number of parallel jobs that can be launched on this worker. . I thought the calling function in [bundle_manager.py](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L308) means this as well. Also when I was testing the changes in bundle_manger.py, `active_runs` somehow always returns zero, so that `get_user_parallel_run_quota_left` returns a valid non-zero number, then jobs always get run as long as there are enough computing resources. Could you please double-check and let me know? Let's try to get these PRs merged today.

codalab/model/bundle_model.py
candicegjing: Oh, I was actually following another outer join clause in our codebase for consisitency at [bundle_model.py](https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L619). Please let me know if we want to change both of them to something else. Maybe we could rename them to be something containing tables' names?

bkgoksel: I think in these cases it's better to try make the new code better styled then stay consistent. Feel free to update the other names as well, but looking at your link that looks like an annoying place to edit.

bkgoksel:
> > 2 things:
> > ```
> > * Can you try to add a simple test for `parallel_run_quota`? I really should've done it at the time of making the PR but somehow skipped it 😿
> > 
> > * We want to count the users' runs that are NOT running on workers that also belong to them. The old logic had a `cl_worker_run.c.user_id != user_id` to test for this. It seems like the new logic does the opposite and counts the number of runs a user has on their own workers. That looks like the wrong number. Am I misunderstanding the new logic?
> > ```
> 
> Hmm, I thought this function `get_user_parallel_run_quota_left` is to get how many `parallel_run_quota` left for a particular user. Based on [Line 2250](https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L2250), `active_runs` should return the number of active running bundles on this worker. Then after it gets subtracted by original `parallel_run_quota`, it will return the number of parallel jobs that can be launched on this worker. . I thought the calling function in [bundle_manager.py](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L308) means this as well. Also when I was testing the changes in bundle_manger.py, `active_runs` somehow always returns zero, so that `get_user_parallel_run_quota_left` returns a valid non-zero number, then jobs always get run as long as there are enough computing resources. Could you please double-check and let me know? Let's try to get these PRs merged today.

Your understanding is correct, however one important distinction is whether the user's runs are running on a worker they own, or a worker the `codalab` root user owns. We don't want to count runs happening on their own workers, and only count runs happening on `codalab` root user's workers. That's why we need to check the worker's owner id is different from the user id. 

By that logic, if you're submitting runs as the root `codalab` user (while testing for example), `active_runs` should  always be 0, since the runs technically always run on `codalab`'s own workers. Or if you're non-root, but you also started a worker yourself, again `active_runs` should be 0 as long as your runs are only running on your own worker.

Technically, even with a `parallel_run_quota` of 3 for example, you could have >3 runs at a time if most of these runs are running on your own workers. And for the `codalab` root user, `parallel_run_quota` is completely meaningless since none of its runs could count against this quota.

I hope this clarifies!

bkgoksel:
One additional observation: I'm running a test deployment off of master at SM, and I'm currently observing that the parallel run quota is being correctly enforced on my user, so maybe there's no reason to alarm here. However it would be good to add documentation that the parallel run quota 1) does not apply to users' own workers 2) does not apply to the root user

candicegjing:
@bkgoksel Yes, I think the main reason that I created this ticket is when I was running jobs under `codalab` account. If this is the correct behavior, we can close this ticket.

bkgoksel:
Sorry about the confusion! Maybe you can make an issue or add a small comment near the parallel run quota method that explains this behavior, so things are clearer for future maintainers.

Issue #1064, opened 2019-03-11
Support bulk bundle operations


w0nche0l:
_From @fabeschan on May 27, 2017 6:4_

Perhaps like a checkbox then gear icon-> bulk operations

_Copied from original issue: codalab/codalab-worksheets#324_

- Delete
- Kill
- Detach

w0nche0l:
_From @robinjia on November 20, 2017 17:57_

Or even checkboxes + keyboard shortcuts would be nice, mostly for `rm` and `kill`.  Otherwise the workflow is, first look on the web interface to find the UUID's of the first and last bundle in the range I want deleted, change to CLI, `cl ls` to figure out what bundle numbers those ID's correspond to, then `cl rm/kill ^X-Y`...and I've definitely accidentally deleted bundles this way, either because I typed the numbers wrong or I accidentally ran the command twice.

Bonus would be reorganizing the worksheet (move these 10 bundles to different place, or create a new table with schema=XYZ containing these 10 bundles) but that sounds like more work.

w0nche0l:
_From @percyliang on November 20, 2017 20:7_

+1

PR #1579, opened 2019-10-30
Bundle bulk operation on web ui

yipenghe:
Implementing #1064 
> Supports bulk bundle operation from the web UI
> Can use 'd' for detach, 'backspace/del' for delete, 'v' for kill on multiple selected bundles, they open up a dialog for confirmation, can 'f' to select force deletion on deletion dialog, can use 'enter' to confirm action. It's the same as running cl rm, cl detach & cl kill on selected bundles
> show/hides bulk bundle action bar when at least one/none is selected, hides the worksheet action bar when bundles are selected (can still use shortcuts to perform)
> Most shortcuts are disabled when a dialog is opened
> Currently the checkmark get refreshed when a new bundle row is added / a bundle operation is successfully performed, the selection persists if bundle operation failed
> some minor changes to bundle row style

### DIFF
--- a/frontend/src/components/worksheets/BundleBulkActionMenu.js
+++ b/frontend/src/components/worksheets/BundleBulkActionMenu.js
@@ -13,6 +13,9 @@ import DialogContentText from '@material-ui/core/DialogContentText';
 import DialogTitle from '@material-ui/core/DialogTitle';
 import Checkbox from '@material-ui/core/Checkbox';
 import * as Mousetrap from '../../util/ws_mousetrap_fork';
+import Tooltip from '@material-ui/core/Tooltip';
+import IconButton from '@material-ui/core/IconButton';
+import InfoIcon from '@material-ui/icons/InfoOutlined';
 
 class BundleBulkActionMenu extends React.Component {
     constructor(props) {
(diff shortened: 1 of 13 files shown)


epicfaace:
When deleting (or bulk deleting) a bundle, show an extra dialog (warning -- "these bundles are attached to more than 1 worksheet. are you sure you want to delete --force?" if force delete is needed

epicfaace:
- Checkbox in the header to "select all"
- keyboard shortcuts (could do later)

epicfaace:
Change "bundles" dropdown to show the four buttons; only show them when at least one checkbox is selected.

yipenghe:
The selection logic should be working now, following the behaviours of gmail style...bulk delete & kill are working
Remaining: 1. Force delete logic 2. attach/detach command 3. style the four buttons

yipenghe:
Some issues:
When select ‘Delete’, want to be able to hit ENTER to confirm, but currently, this triggers something in the back

Message ‘Failed to perform this actions’ shows up in a popup box but unclear how to close it (should have an ‘x’ in the corner)

Force delete - should have an [i] with a tooltip that explains what this is

epicfaace (approved):

PR #1603, opened 2019-11-06
Refactor CSS

epicfaace:
- refactor a bunch of CSS
- upgrade react-scripts to allow using SASS


docker/compose_files/docker-compose.yml
```diff
@@ -88,7 +88,7 @@ services:
 
   frontend:
     image: codalab/frontend:${CODALAB_VERSION}
-    command: serve -s build -l ${CODALAB_FRONTEND_PORT}
+    command: npm run build --production && serve -s build -l ${CODALAB_FRONTEND_PORT}
```
percyliang: Why this change?  The files should already be built in the Docker image for production, and in dev mode, we hotload don't want to build the files anyway.

frontend/package.json
```diff
@@ -12,7 +12,7 @@
     "@material-ui/icons": "^3.0.2",
     "ace-builds": "^1.4.2",
     "autosuggest-highlight": "^3.1.1",
-    "bootstrap": "^4.3.1",
+    "bootstrap": "^3.3.6",
```
percyliang, line 15: Why are we going back in version?

percyliang (approved):
Thanks for doing this!  Everything looks good to me except the docker change (which probably should be a different PR anyway).

percyliang:
@epicfaace can we merge this soon?

frontend/package.json
epicfaace: We didn't use the `bootstrap` node module at all earlier; instead, we had just copied-and-pasted the bootstrap v3 styles into the imports.css file.

docker/compose_files/docker-compose.yml
epicfaace: Let me move that to another PR. The problem was that the way it worked before, whenever you started up a codalab instance in dev mode, you would have to wait for the files to build in production mode, which takes a long time.

PR #1605, opened 2019-11-06
Update time_used logic to consider time overlap between bundles

### LINKED ISSUE #1533: time used 161d20h but only registered a month
Date Joined (UTC)
Sun Sep 15 14:49:30 2019
Time Quota
60d0h
Time Used
161d20h

candicegjing:
Fixed #1533 .
In this PR, I am trying to fix the unusual time quota increasing problem. 

- Current logic: 
For each bundle, when it starts to finalize, we update the `user['time_used']` by directly adding the amount of wall clock running time from `bundle.metadata` to `time_used`. This logic doesn't consider the overlapping time windows that could be generated from other bundles owned by this user. So those certain users who created a large number of bundles in a short time window will get their time quota reached very quickly.

- New logic:
In this PR, I added a new column that tracks the timestamp of the last bundle from each user to `user` table, named with `last_bundle_finished_time`. During the time that we update `user['time_used']`, we compare if the start time of the current bundle is earlier the `last_bundle_finished_time`. If yes, then there is an overlapping period which we shouldn't add to the final result. If no, we can use the bundle start time as the left boundary of the current time window to calculate the amount of time to be added to `user['time_used']`.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -847,7 +847,7 @@ class BundleModel(object):
         self.update_bundle(bundle, bundle_update, connection)
 
         if user_id == self.root_user_id:
-            self.increment_user_time_used(bundle.owner_id, getattr(bundle.metadata, 'time', 0))
+            self.increment_user_time_used(bundle.owner_id, getattr(bundle.metadata, 'started', 0))
 
         return True
 
@@ -2009,6 +2009,7 @@ class BundleModel(object):
                         "disk_used": 0,
                         "affiliation": affiliation,
                         "url": None,
+                        "last_bundle_finished_time": now,
                     }
                 )
             )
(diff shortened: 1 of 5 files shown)


percyliang (commented):
I'm confused why we want this.  If a user has two runs that overlap that each take an hour, it should count 2 hours towards the quota.

codalab/model/bundle_model.py
```diff
@@ -2217,12 +2218,32 @@ def update_user_info(self, user_info):
                 cl_user.update().where(cl_user.c.user_id == user_info['user_id']).values(user_info)
             )
 
-    def increment_user_time_used(self, user_id, amount):
```
yipenghe, line 2220: I think we should be checking how 'amount' is calculated, as in meeting we wanted to make sure amount should only be the 'runnning' time of a run bundle

PR #1625, opened 2019-11-10
Respect .gitignore when archiving

### LINKED ISSUE #1312: cl upload use gitignore
We package things using tar, so we can do this:
https://stackoverflow.com/questions/39514510/exclude-directories-from-tar-archive-with-a-tarignore-file

Also need to ignore `.git`.

This way, we can do `cl upload` on a git repo and not have unwanted binary files uploaded.

teetone:
Fixes #1312 

Changes:
- When archiving, ignore entries specified by `.gitignore` by default. It also works for `.gitignore`'s in subdirectories.
- Exclude `.git/` always
- Added `-i` option for `cl upload` where the user can specify a different ignore file.

How to test:
- `cl upload some_directory` where `some_directory` has `.gitignore`.
- `cl upload some_directory -i='.tarignore'`  should ignore `.tarignore`.
- `cl upload some_directory -i=''`  shouldn't ignore.

I also added a test that covers this logic in `file_util_test.py`.


codalab/worker/file_util.py
```diff
@@ -25,18 +30,24 @@ def tar_gzip_directory(
     exclude_patterns: Any directory entries with the given names at any depth in
                       the directory structure are excluded.
     """
-    args = ['tar', 'czf', '-', '-C', directory_path]
+    # always ignore entries specified by the ignore file (e.g. .gitignore)
```
percyliang: nit: capitalize comments

codalab/worker/file_util.py
```diff
@@ -10,10 +10,15 @@
...
+GIT_PATTERN = '.git'
 
 
 def tar_gzip_directory(
-    directory_path, follow_symlinks=False, exclude_patterns=[], exclude_names=[]
+    directory_path,
+    follow_symlinks=False,
+    exclude_patterns=[],
+    exclude_names=[],
+    ignore_file_name='.gitignore',
```
percyliang: Maybe this should be `None` just to maintain backward compatibility?

codalab/worker/file_util.py
```diff
@@ -25,18 +30,24 @@ def tar_gzip_directory(
     exclude_patterns: Any directory entries with the given names at any depth in
                       the directory structure are excluded.
     """
```
percyliang, line 29: document `ignore_file_name`

tests/worker/file_util_test.py
```diff
@@ -65,3 +65,18 @@ def test_bz2_file(self):
...
+
+    def test_tar_exclude_ignore(self):
+        dir = os.path.join(os.path.dirname(os.path.dirname(__file__)), 'files/dir3')
+        temp_dir = tempfile.mkdtemp()
+        self.addCleanup(lambda: remove_path(temp_dir))
+        output_dir = os.path.join(temp_dir, 'output')
+
+        un_tar_directory(tar_gzip_directory(dir, ignore_file_name='.tarignore'), output_dir, 'gz')
+        output_dir_entries = os.listdir(output_dir)
+        self.assertIn('dir4', output_dir_entries)
```
percyliang: Could we name the files/directories in such a way to make it more obvious which ones are meant to be included versus excluded?

percyliang (changes requested):
Thanks!  One thing I'm worried about is breaking backward compatibility, where a user might be trying to upload a data file that is in `.gitignore`...  In that case, the failure will be quite mysterious...  The other option is to have an explicit flag in `cl upload` that asks to use `.gitignore`.  Or maybe we can have `.gitignore` as a default, but print out a message that makes it clear we're using it.  Thoughts?

Note that if people will be uploading directories in the future from the web interface, it would be nice to have this logic be consistent.

codalab/worker/file_util.py
teetone: @percyliang: Regarding your comments about backward compatibility, I prefer respecting `.gitignore` as the default, since I think most users do not want to see binary files and extraneous files in CodaLab. Additionally, I can add the optional flag to `cl upload` to give users the freedom to turn it off and print a message based on the value of the flag.

tests/worker/file_util_test.py
teetone: Yes, that makes it much more clear.

codalab/lib/bundle_cli.py
```diff
@@ -1120,6 +1120,13 @@ def do_workers_command(self, args):
                 help='Upload to this worksheet (%s).' % WORKSHEET_SPEC_FORMAT,
                 completer=WorksheetsCompleter,
             ),
+            Commands.Argument(
+                '-i',
+                '--ignore-gitignore',
+                help='Ignores exclusion patterns specified by.gitignore when archiving.',
```
percyliang: missing space

codalab/lib/bundle_cli.py
```diff
@@ -1120,6 +1120,13 @@ def do_workers_command(self, args):
                 help='Upload to this worksheet (%s).' % WORKSHEET_SPEC_FORMAT,
                 completer=WorksheetsCompleter,
             ),
+            Commands.Argument(
```
percyliang, line 1123: ignore-gitignore is counterintuitive...I'd just make this `--ignore` with the default being '.gitignore'.  People can override this by passing in `''`.

codalab/lib/zip_util.py
```diff
@@ -141,9 +147,17 @@ def resolve_source(source):
         source = sources[0]
         filename = os.path.basename(source)
         if os.path.isdir(sources[0]):
-            archived = tar_gzip_directory(
-                source, follow_symlinks=follow_symlinks, exclude_patterns=exclude_patterns
-            )
+            if use_gitignore:
```
percyliang: Shouldn't need to branch here...just pass the ignore file through

codalab/worker/file_util.py
```diff
@@ -24,19 +29,28 @@ def tar_gzip_directory(
                    are not included.
     exclude_patterns: Any directory entries with the given names at any depth in
                       the directory structure are excluded.
+    ignore_file_name: Name of the file where exclusion patterns are read from.
```
percyliang: Would prefer `ignore_filename` or even just `ignore_file` to highlight the structure of the `ignore` part.

tests/worker/file_util_test.py
```diff
@@ -65,3 +65,24 @@ def test_bz2_file(self):
 
     def test_gzip_bytestring(self):
         self.assertEqual(un_gzip_bytestring(gzip_bytestring(b'contents')), b'contents')
+
+    def test_tar_exclude_ignore(self):
+        dir = os.path.join(
+            os.path.dirname(os.path.dirname(__file__)), 'files/dirToTestExcludeIgnore'
```
percyliang: In general, can we make these directory names less Java-like?  (i) shorter and (ii) using snake_case...  I think 'ignore_test' would suffice.

percyliang (changes requested):

codalab/lib/bundle_cli.py
teetone: @percyliang: Just having a `--ignore` argument is cleaner to implement, but this would allow the user to specify any ignore file. I'm not sure if that flexibility is needed, since most users will just rely on `.gitignore` and most likely won't add another ignore file like `.clignore` to their repos. And, if that's the case, I think it's cleaner to just rely on a boolean flag to turn off the functionality instead of passing in an empty string.

codalab/lib/bundle_cli.py
teetone: Sorry about that, will update.

codalab/lib/bundle_cli.py
percyliang: I'm not sure about that...people might have different ignore files for different purposes (e.g., `.dockerignore`).  As I mentioned before, you might want to upload files that are not checked into git (like datasets).  So I think we should go with `--ignore` since it's simpler, easier to think about, and more powerful.

teetone: @percyliang: That makes sense. I have updated the PR.

codalab/lib/bundle_cli.py
```diff
@@ -1120,6 +1120,12 @@ def do_workers_command(self, args):
...
+                '-i',
+                '--ignore',
+                help='Name of the file to read exclusion patterns from when archiving.',
```
percyliang: 'when archiving' is an implementation detail, should leave out.  I'd say 'Name of file containing patterns matching files to exclude from upload'.

codalab/worker/file_util.py
```diff
@@ -10,10 +10,15 @@
...
+GIT_PATTERN = '.git'
 
 
 def tar_gzip_directory(
-    directory_path, follow_symlinks=False, exclude_patterns=[], exclude_names=[]
+    directory_path,
+    follow_symlinks=False,
+    exclude_patterns=[],
+    exclude_names=[],
+    ignore_file='.gitignore',
```
percyliang: Perhaps this default should be `None`.  In general, should have as little magic/processing as possible.

percyliang (approved):
Great, thanks!  Just a few more small comments.

epicfaace:
@teetone why did we decide to make `.gitignore` the default? To me, `.gitignore` sounds like "these files should not be checked into git" which is not the same as "these files should not be archived" (in which case, something like `.tarignore` might make more sense).

percyliang:
I guess most people have a `.gitignore`, but no one has a `.tarignore`.  If anything, maybe it should be a `.clignore`.  We can take a poll in the CodaLab slack channel.

teetone:
I still like the idea of having `.gitignore` as the default (since it's most commonly used) and encourage users to have a `.clignore` if they want an ignore file specific for CodaLab.

PR #1627, opened 2019-11-11
Not to raise exception when bundle is not found on target worker when…

### LINKED ISSUE #1485: [USER ISSUE] Worker finalizing bundles can be very slow
**Describe the bug**

I'm consistently seeing run bundles take an unreasonably long time to finalize, given their size (15 minutes for a 129k file).

**Codalab Information**
- *Your username*: nfliu
- *UUID of worksheet where problem occurred*: 0xf6bae43872f14e4a88bb2c549abfe873
- *UUID of problem bundle (please don't delete)*: [0x587da02384d541a38cf2b02d67901c0b](https://worksheets.codalab.org/bundles/0x587da02384d541a38cf2b02d67901c0b)
 - *OS (if on CLI)*: Mac OS Mojave
 - *Codalab CLI Version (if on CLI)*: `master`

**To Reproduce**
1. Run a `cl-worker` on master
2. Run a bundle.

**Expected behavior**
Finalizing the bundle should not take 15 minutes

**Screenshots**

Here's what I'm seeing in the worker logs:

```
2019-10-08 20:26:29,885 Uploading results for run with UUID 0x587da02384d541a38cf2b02d67901c0b
2019-10-08 20:40:40,248 Received mark_finalized message: {'type': 'mark_finalized', 'uuid': '0x587da02384d541a38cf2b02d67901c0b', 'meta': {'version': '0.4.3'}}
```

Note the timestamps, it takes nearly 15 minutes to "upload results" for a 129k bundle. I don't think it takes actually 15 minutes, so it'd be good to get some more information about what else is happening in the `finalizing` state.

**Additional context**

I have a `--max-work-dir-size` of `20g` on this worker.

candicegjing:
Fixed #1485 . 
When we tried to acknowledge recently finished bundles, we always try to get a list of bundles that are in `finalizing` state and loop through each of them to see if we can find the current bundle is running on a worker by scanning the `worker_run` table. When we cannot find the current bundle running on any workers (this could be due to a worker lost connection to our server or any other possible reasons), we send this bundle to go through a `precondition` check, which will then raise a `PreconditionViolation` error. This raised error will block the rest of the bundles that were trying to get acknowledged. I update the precondition check here to be non-blocking for the rest of the bundles that were in `finalizing` state.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -7,6 +7,8 @@ import datetime
 import os
 import re
 import time
+import logging
+
 from uuid import uuid4
 
 from sqlalchemy import and_, or_, not_, select, union, desc, func
@@ -48,6 +50,8 @@ from codalab.objects.user import User
 from codalab.rest.util import get_group_info
 from codalab.worker.bundle_state import State
 
+
+logger = logging.getLogger(__name__)
 
 SEARCH_KEYWORD_REGEX = re.compile('^([\.\w/]*)=(.*)$')
 
@@ -221,9 +225,11 @@ class BundleModel(object):
             row = conn.execute(
                 cl_worker_run.select().where(cl_worker_run.c.run_uuid == uuid)
             ).fetchone()
-            precondition(
-                row, 'Trying to find worker for bundle {} that is not running.'.format(uuid)
-            )
+
+            if not row:
+                logger.info('Trying to find worker for bundle {} that is not running.'.format(uuid))
+                return None
+
             worker_row = conn.execute(
                 cl_worker.select().where(
                     and_(cl_worker.c.user_id == row.user_id, cl_worker.c.worker_id == row.worker_id)


codalab/model/bundle_model.py
```diff
@@ -221,9 +225,11 @@ def get_bundle_worker(self, uuid):
             row = conn.execute(
                 cl_worker_run.select().where(cl_worker_run.c.run_uuid == uuid)
             ).fetchone()
-            precondition(
-                row, 'Trying to find worker for bundle {} that is not running.'.format(uuid)
-            )
+
+            if len(row) == 0:
+                logger.info('Trying to find worker for bundle {} that is not running.'.format(uuid))
```
percyliang, line 230: `logger.warn`?

codalab/model/bundle_model.py
```diff
@@ -221,9 +225,11 @@ def get_bundle_worker(self, uuid):
...
+                return None
```
percyliang, line 231: Can the caller deal with `None`?

percyliang (approved):

codalab/model/bundle_model.py
candicegjing: Yes, the caller handles `None` return result at https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L266

codalab/model/bundle_model.py
```diff
@@ -221,9 +225,11 @@ def get_bundle_worker(self, uuid):
             row = conn.execute(
                 cl_worker_run.select().where(cl_worker_run.c.run_uuid == uuid)
             ).fetchone()
-            precondition(
```
yipenghe, line 224: When you say the precondition error blocks the rest workers from getting acknowledged, how does that block actually work?

bkgoksel: I'm guessing it throws an exception, and the upstream caller can't handle that exception.

candicegjing: It's from here [common.py](https://github.com/codalab/codalab-worksheets/blob/28e3c011f0cc426ea0dc27c7eeba844869b8159c/codalab/common.py#L93,) which raises an exception that will be eventually caught by [bundle_manager.py](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L66)

yipenghe: Ah so is the catching of this exception taking a lot of time? What happens after, does the bundle that raise this exception continue to try getting acknowledgement and keep raising the exception, which therefore blocks the finalizing?

bkgoksel (approved):
LGTM. I'm not super familiar with the logic here, and we used to sporadically see these `PreconditionViolation` errors on our logs, so having something more descriptive and non-blocking like this is good, especially if this precondition might get violated during common usage.

codalab/model/bundle_model.py
candicegjing: It's once the exception is thrown, the rest of the code will be skipped and the main process will return back to this part [bundle_manager.py](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L66)

yipenghe: I see

yipenghe (approved):

Issue #1614, opened 2019-11-07
Refreshing issues


epicfaace:
- Don't change the cursor location after refreshing (shift + R)
- Show some visual feedback / loading sign when refreshing

PR #1629, opened 2019-11-12
cursor remains the same after refresh; signposting for refresh

Levilian:
fix: #1614


frontend/src/components/worksheets/Worksheet.js
```diff
@@ -511,9 +514,17 @@ class Worksheet extends React.Component {
             Mousetrap.bind(
                 ['shift+r'],
                 function(e) {
-                    // refresh the worksheet and update the
-                    // focus to the first item
-                    this.reloadWorksheet(undefined, 'end');
+                    var focusIndex = this.state.focusIndex;
+                    var subFocusIndex = this.state.subFocusIndex;
+                    this.reloadWorksheet(focusIndex, subFocusIndex);
```
yipenghe: Looking at the reloadworksheet, this doesn't seem the correct arguments (passing focusIndex as partialUpdateItems?).  

I think a better way might be to add a new argument (similar to rawIndexAfterEditMode) for refresh, after refetching the worksheet, setFocus on the original focus&subfocus. We actually want to refetch the worksheet after shift+r I think, so partialUpdateItem should be undefined for this call

frontend/src/components/worksheets/Worksheet.js
```diff
@@ -511,9 +514,17 @@ class Worksheet extends React.Component {
...
+                    toast.success('🦄 Worksheet refreshed!', {
```
yipenghe: I like the toast!!!

frontend/src/components/worksheets/Worksheet.js
```diff
@@ -230,7 +233,7 @@ class Worksheet extends React.Component {
         // The uuid are recorded by handleCheckBundle
         // Refreshes the checkbox after commands
         // If the action failed, the check will persist
-        let force_delete = cmd=='rm' && this.state.forceDelete ? '--force' : null;
+        let force_delete = cmd === 'rm' && this.state.forceDelete ? '--force' : null;
```
yipenghe: Great catches!

yipenghe (changes requested):

yipenghe (approved):

PR #1631, opened 2019-11-12
Refactor bundle_manager.py to have less db call and reuse some variables

candicegjing:
This PR addresses partial problems discovered in #1254 .
In this PR, it made changes to the following items:
1. Reordered `_acknowledge_recently_finished_bundles` and `_fail_on_too_many_resources` as they don't depend on each other. 
2. Created a return value for `_fail_on_too_many_resources` so that it can be reused in the function `_schedule_run_bundles_on_workers` later and save a little bit computing time.
3. In `_schedule_run_bundles_on_workers`:
   1. removed `user_owned` flag.
   2. Sorted the `staged_bundles` with `request_queue` so that those bundles which were requested to run on a personal worker will be scheduled first.
   3. Replaced `user_owned` logic but kept the same logic: always schedule bundles to run on *user owned workers* if there is any. If there is no *user owned workers*, schedule jobs to run on CodaLab's public worker.
4. In `_filter_and_sort_workers`:
   1. moved the logic of tag filtering up so that user owned worker will be examined first.
   2. removed the logic of returning an empty list when the tag is not found in `workers_list`. The reason that I was thinking is that even if there is no matched worker found in `workers_list`, if there is any available workers out there, we could schedule the job to run instead of keeping it in `staged` state all the time.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -299,28 +299,40 @@ class BundleManager(object):
                 logger.info('Bringing bundle offline %s: %s', bundle.uuid, failure_message)
                 self._model.transition_bundle_worker_offline(bundle)
 
-    def _schedule_run_bundles_on_workers(self, workers, user_owned):
+    def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run):
         """
         Schedules STAGED bundles to run on the given workers. If user_owned is
         True, then schedules on workers run by the owner of each bundle.
         Otherwise, uses CodaLab-owned workers, which have user ID root_user_id.
         """
-        for bundle in self._model.batch_get_bundles(state=State.STAGED, bundle_type='run'):
-            if user_owned:
-                workers_list = workers.user_owned_workers(bundle.owner_id)
-            else:
+        # Reorder the stage_bundles so that bundles which were requested to run on a personal worker
+        # will be scheduled to run first
+        staged_bundles_to_run.sort(
+            key=lambda b: (b[0].metadata.request_queue is not None, b[0].metadata.request_queue),
+            reverse=True,
+        )
+
+        for bundle, bundle_resources in staged_bundles_to_run:
+            # Get user_owned workers.
+            workers_list = workers.user_owned_workers(bundle.owner_id)
+
+            # If there is no user_owned worker, try to schedule the current bundle to run on a CodaLab's public worker.
+            if not workers_list:
+                # Check if there is enough parallel run quota left for this user
                 if not self._model.get_user_parallel_run_quota_left(bundle.owner_id):
                     logger.info(
                         "User %s has no parallel run quota left, skipping job for now",
                         bundle.owner_id,
                     )
-                    continue  # Don't start this bundle yet
+                    # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
+                    continue
+                # Get all the CodaLab's public workers
                 workers_list = workers.user_owned_workers(self._model.root_user_id)
 
             workers_list = self._deduct_worker_resources(workers_list)
-            bundle_resources = self._compute_bundle_resources(bundle)
             workers_list = self._filter_and_sort_workers(workers_list, bundle, bundle_resources)
 
+            # Try starting bundles on the workers that have enough computing resources
             for worker in workers_list:
                 if self._try_start_bundle(workers, worker, bundle):
                     break
@@ -351,12 +363,21 @@ class BundleManager(object):
         Filters the workers to those that can run the given bundle and returns
         the list sorted in order of preference for running the bundle.
         """
-
         # keep track of which workers have GPUs
         has_gpu = {}
         for worker in workers_list:
             worker_id = worker['worker_id']
             has_gpu[worker_id] = worker['gpus'] > 0
+
+        # Filter by tag.
+        request_queue = bundle.metadata.request_queue
+        if request_queue:
+            tagm = re.match('tag=(.+)', request_queue)
+            if tagm:
+                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
+            else:
+                # We don't know how to handle this type of request queue argument
+                return []
 
         # Filter by CPUs.
         workers_list = [
@@ -373,17 +394,6 @@ class BundleManager(object):
         workers_list = [
             worker for worker in workers_list if worker['memory_bytes'] >= bundle_resources.memory
         ]
-
-        # Filter by tag.
-        request_queue = bundle.metadata.request_queue
-        if request_queue:
-            tagm = re.match('tag=(.+)', request_queue)
-            if tagm:
-                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
-            else:
-                # We don't know how to handle this type of request queue
-                # argument.
-                return []
 
         # Sort workers list according to these keys in the following succession:
         #  - whether the worker is a CPU-only worker, if the bundle doesn't request GPUs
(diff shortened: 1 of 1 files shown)


codalab/server/bundle_manager.py
```diff
@@ -299,16 +299,25 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
+        # will be scheduled to run first
+        staged_bundles_to_run.sort(
+            key=lambda b: (b.metadata.request_queue is not None, b.metadata.request_queue),
+            reverse=True,
+        )
+
+        for bundle, bundle_resources in staged_bundles_to_run:
+            # Check if there is an user_owned worker.
+            # If so, schedule the current bundle to run on a user_owned worker first
+            workers_list = workers.user_owned_workers(bundle.owner_id)
```
bkgoksel, line 317: What if there are user owned workers but none of them can run this bundle (ie I have my own CPU workers but I submitted a GPU run, and there is a public GPU worker that could take my run)
The fallback to public workers should come after we filter the workers.

codalab/server/bundle_manager.py
```diff
@@ -374,17 +388,6 @@ def _filter_and_sort_workers(workers_list, bundle, bundle_resources):
...
-        # Filter by tag.
-        request_queue = bundle.metadata.request_queue
-        if request_queue:
-            tagm = re.match('tag=(.+)', request_queue)
-            if tagm:
-                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
-            else:
-                # We don't know how to handle this type of request queue
-                # argument.
-                return []
```
bkgoksel, line 386: Removing this + prioritizing `request_queue` runs means I can submit a run, set some bogus string as the `request_queue` and get ahead of all the other runs (it won't trigger the tag matching, so instead in the filtering step we will continue with all the workers)

codalab/server/bundle_manager.py
```diff
@@ -627,7 +629,11 @@ def _fail_on_too_many_resources(self):
         Fails bundles that request more resources than available for the given user.
         Note: allow more resources than available on any worker because new
         workers might get spun up in response to the presence of this run.
+
+        :return: a list of tuple which contains valid staged bundles and their bundle_resources.
         """
+        # Keep track of staged bundles that have valid resources requested
+        valid_staged_bundles_to_run = []
```
bkgoksel:
```suggestion
        staged_bundles_to_run = []
```

codalab/server/bundle_manager.py
```diff
@@ -673,3 +679,7 @@ def _fail_on_too_many_resources(self):
                     bundle,
                     {'state': State.FAILED, 'metadata': {'failure_message': failure_message}},
                 )
+            else:
+                valid_staged_bundles_to_run.append(tuple(bundle, bundle_resources))
```
bkgoksel:
```suggestion
                staged_bundles_to_run.append(tuple(bundle, bundle_resources))
```

codalab/server/bundle_manager.py
```diff
@@ -673,3 +679,7 @@ def _fail_on_too_many_resources(self):
...
+
+        return valid_staged_bundles_to_run
```
bkgoksel:
```suggestion
        return staged_bundles_to_run
```

codalab/server/bundle_manager.py
```diff
@@ -592,12 +595,11 @@ def _schedule_run_bundles(self):
...
         self._restage_stuck_starting_bundles(workers)
         self._bring_offline_stuck_running_bundles(workers)
-        self._fail_on_too_many_resources()
         self._acknowledge_recently_finished_bundles(workers)
+        staged_bundles_to_run = self._fail_on_too_many_resources()
 
         # Schedule, preferring user-owned workers.
-        self._schedule_run_bundles_on_workers(workers, user_owned=True)
-        self._schedule_run_bundles_on_workers(workers, user_owned=False)
+        self._schedule_run_bundles_on_workers(workers, staged_bundles_to_run)
```
bkgoksel, line 609: I like the idea of getting the bundles once and caching them, but maybe there's room for even more improvement here. Right now (ie before this PR) every operation here re-fetches everything from the database before trying to operate on them, but this doesn't really prevent race conditions and makes us waste a bunch of time re-getting everything from the database.

I wonder if we can juts get everythin (workers, bundles etc) once per bundle manager iteration, and explicitly skip cases where we notice  some disparity, hoping that next iteration they will be handled fine.

bkgoksel (changes requested):
Added a couple concerns, but this looks good. I'm wondering if we can cache even more values for each bundle manager iteration as I commented below.

codalab/server/bundle_manager.py
candicegjing: I think this is a good point! We should always be better at utilizing all the current resources we have and unblocking users' jobs. I have updated the code to fall back to public workers so that even if there is no valid private workers available, we can still check our public workers to see if the current bundle can be run there.

codalab/server/bundle_manager.py
candicegjing: Variable names are all updated.

codalab/server/bundle_manager.py
candicegjing: That's true. We need to cache more variables so that we don't need to keep fetching them from the database.

codalab/server/bundle_manager.py
```diff
@@ -351,13 +372,19 @@ def _filter_and_sort_workers(workers_list, bundle, bundle_resources):
...
         for worker in workers_list:
             worker_id = worker['worker_id']
             has_gpu[worker_id] = worker['gpus'] > 0
 
+        # Filter by tag.
+        request_queue = bundle.metadata.request_queue
+        if request_queue:
+            tagm = re.match('tag=(.+)', request_queue)
+            if tagm:
+                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
```
bkgoksel, line 377: We still need the `else return []` statement here. We don't want to silently ignore malformed request queue arguments.

codalab/server/bundle_manager.py
```diff
@@ -299,28 +299,49 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
                         bundle.owner_id,
                     )
-                    continue  # Don't start this bundle yet
-                workers_list = workers.user_owned_workers(self._model.root_user_id)
-
-            workers_list = self._deduct_worker_resources(workers_list)
-            bundle_resources = self._compute_bundle_resources(bundle)
-            workers_list = self._filter_and_sort_workers(workers_list, bundle, bundle_resources)
+                    # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
+                    continue
```
bkgoksel, line 328: I think this whole block needs to be the `else` clause of the `if` statement on line 341. (ie if workers list is empty at the end of the first iteration then we're surely moving on to public workers and want to enforce this limit there).

codalab/server/bundle_manager.py
```diff
@@ -592,12 +608,11 @@ def _schedule_run_bundles(self):
         self._cleanup_dead_workers(workers)
         self._restage_stuck_starting_bundles(workers)
         self._bring_offline_stuck_running_bundles(workers)
-        self._fail_on_too_many_resources()
         self._acknowledge_recently_finished_bundles(workers)
+        staged_bundles_to_run = self._fail_on_too_many_resources()
```
bkgoksel: This method should be renamed to something like `get_staged_bundles_to_run`

codalab/server/bundle_manager.py
bkgoksel: Otherwise, I can go over my run quota like the following:
Say there's a parallel run quota of 1, and  I already have a running bundle. I can start a dummy private worker with 100mb of ram, and then submit a job requesting 4gb of ram. It would pass this check because I have a nonempty list of private workers, but then later on that private worker would get filtered away, and my run would get scheduled to the public workers. I could keep submitting as many runs as I want like that.

bkgoksel (changes requested):
This makes sense now, just noticed two edge cases and one method should be renamed.

candicegjing:
@bkgoksel I removed the logic related to how we should handle the case when no private workers are available, as it would be better to keep this PR small and have a new PR handling that logic separately. Please check the code and let me know if I need to change anything.

Issue #1638, opened 2019-11-12
Improve checkbox style


yipenghe:
>Can we copy Gmail's style for the gmail checkboxes more?  In particular, the checkboxes are more grayed out by default, and become dark only on the entry where the cursor is (along with the blue bar).  Also, when you hover, it creates more of a border around the item rather than changing the color, which I think is better.

Issue #1613, opened 2019-11-07
Selecting a bundle row is very faint


epicfaace:
Maybe show a blue bar, like when you select a worksheet item.

PR #1639, opened 2019-11-12
Styles bundle rows

yipenghe:
#1638 & #1613 
Contains all changes from fix-functions, making the base branch fix-functions to see difference easier.

Check out the bundle row style, it's cool!!!
Screenshot from 2019-11-12 00-47-45

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -341,6 +341,7 @@ const styles = (theme) => ({
     },
     lowlight: {
         backgroundColor: `${theme.color.grey.light} !important`,
+        borderLeft: '3px solid transparent',
     },
 });
 
--- a/frontend/src/components/worksheets/items/TableItem/BundleRow.js
+++ b/frontend/src/components/worksheets/items/TableItem/BundleRow.js
@@ -326,6 +326,7 @@ const styles = (theme) => ({
     },
     highlight: {
         backgroundColor: `${theme.color.primary.lightest} !important`,
+        borderLeft: '3px solid #1d91c0',
     },
     lowlight: {
         backgroundColor: `${theme.color.grey.light} !important`,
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -122,7 +123,7 @@ class BundleRow extends Component {
             if (col === 0) {
                 url = baseUrl;
                 checkBox = <Checkbox
-                                icon={<CheckBoxOutlineBlankIcon fontSize="small" />}
+                                icon={<CheckBoxOutlineBlankIcon color={this.props.focused || this.state.hovered? 'action': 'disabled'} fontSize="small" />}
```
percyliang: nit: space before `?`

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -104,7 +105,7 @@ class TableItem extends React.Component<{
                                 checked={this.state.checked}
                                 onChange={this.handleSelectAllClick}
                                 value="checked"
-                                icon={<CheckBoxOutlineBlankIcon fontSize="small" />}
+                                icon={<CheckBoxOutlineBlankIcon color={this.state.hovered? 'action':'disabled'} fontSize="small" />}
```
percyliang: ditto, and space between `:`

percyliang (approved):
This looks great!  One remaining thing is that the row that the cursor is on has the blue bar, which causes a larger indent than a non-selected row...which causes the rows to shift around a bit.  Can we make the indent the same to avoid the jumpiness?

yipenghe:
> This looks great! One remaining thing is that the row that the cursor is on has the blue bar, which causes a larger indent than a non-selected row...which causes the rows to shift around a bit. Can we make the indent the same to avoid the jumpiness?

Added a transparent border to non-focused rows, it won't jump around now

yipenghe:
Force pushed to make changes here clean and only contain the styling part

Issue #1643, opened 2019-11-12
Double/multiple clicking on same worksheet rows except  for bundle rows doesn't trigger keyboards event


yipenghe:

yipenghe:
will by solved by #1649

PR #1644, opened 2019-11-13
Fix 'enter' event after multiple clicks on worksheet row

yipenghe:
Fixing #1643

Not entirely sure what's going on, but I think the setState even when index, subIndex didn't change will cause the focus of a click change. Adding the check solves this problem and - it's good practice to return early anyway

Also a nit change to scope 'enter' binding

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -338,9 +338,9 @@ class Worksheet extends React.Component {
 // BULK OPERATION RELATED CODE ABOVE======================================
 
     setFocus = (index, subIndex, shouldScroll = true) => {
-        if (index === this.state.focusIndex && subIndex === this.state.subFocusIndex){
-            return;
-        }
+        // if (index === this.state.focusIndex && subIndex === this.state.subFocusIndex){
+        //     return;
+        // }
         var info = this.state.ws.info;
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/Worksheet.js
```diff
@@ -669,16 +672,18 @@ class Worksheet extends React.Component {
...
-        Mousetrap.bind(['enter'], function(e) {
-            if (this.state.openDelete){
-                this.executeBundleCommandNoEvent('rm');
-            }else if (this.state.openKill){
-                this.executeBundleCommandNoEvent('kill');
-            }else if (this.state.openDetach){
-                this.executeBundleCommandNoEvent('detach');
-            }
-        }.bind(this));
+        if (this.state.showBundleOperationButtons && (this.state.openDelete||this.state.openKill||this.state.openDelete)){
```
teetone: It looks like we're checking `this.state.openDelete` twice.

teetone (changes requested):

frontend/src/components/worksheets/Worksheet.js
yipenghe: Good catch, thanks!

yipenghe:
rebasing onto fix-worksheet-import branch

teetone (approved):

yipenghe:
Seems like #1649 is fixing it as well, I will close this one

PR #1647, opened 2019-11-13
Don't do a production build in dev mode

epicfaace:
The problem was that whenever you started up a codalab instance
in dev mode, you would have to wait for the files to build in
production mode, which takes a long time. This is partly why
it takes so long to set up our demos. Additionally, when you
switch between branches that have different node_modules installed,
when you do `npm install` it resets your node_modules directory,
so essentially you have to wait for a full production build each
time you switch between branches (whereas if you could just reuse
the same node_modules folder, you wouldn't have to do that).

This change changes that by making sure that the production build
happens only in production mode, and that dev mode only spins up
the dev server.

### DIFF
--- a/docker/compose_files/docker-compose.yml
+++ b/docker/compose_files/docker-compose.yml
@@ -88,7 +88,7 @@ services:
 
   frontend:
     image: codalab/frontend:${CODALAB_VERSION}
-    command: serve -s build -l ${CODALAB_FRONTEND_PORT}
+    command: npm run build --production && serve -s build -l ${CODALAB_FRONTEND_PORT}
     <<: *codalab-base
     <<: *codalab-root
     depends_on:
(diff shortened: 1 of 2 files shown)


percyliang:
If you do `./codalab_service.py start -d`, then it shouldn't build the dev images, right?

epicfaace:
That's a good point. I've always been adding in the `-b` flag when building, as I assumed that even though `-b` would build the Docker images, it wouldn't force a production build of the js files. Not including the `-b` flag is a good workaround.

Still, when we include both `-b` and `-d`, it would make sense not to build the production files, right?

percyliang:
Ideally yes.  But I'm worried about adding extra build time for starting the CodaLab service in production.

epicfaace:
Probably don't need this.

PR #1650, opened 2019-11-13
Remove text-align prop

epicfaace:
Removes warning: "Warning: React does not recognize the `textAlign` prop on a DOM element"

### DIFF
--- a/frontend/src/routes/HomePage.js
+++ b/frontend/src/routes/HomePage.js
@@ -106,7 +106,7 @@ class HomePage extends React.Component<{
                     <Grid item xs={12} className={classes.inner} container>
                         {/** Summary. */}
                         <Grid item xs={12} className={classNames(classes.textBox, classes.spacer)}>
-                            <Typography variant='h5' color='textSecondary' align='center'>
+                            <Typography variant='h5' color='textSecondary'>
                                 Run your machine learning experiments in the cloud. Manage them in a
                                 digital lab notebook. Publish them so other researchers can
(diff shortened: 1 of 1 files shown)


frontend/src/routes/HomePage.js
```diff
@@ -106,7 +106,7 @@ class HomePage extends React.Component<{
                     <Grid item xs={12} className={classes.inner} container>
                         {/** Summary. */}
                         <Grid item xs={12} className={classNames(classes.textBox, classes.spacer)}>
-                            <Typography variant='h5' color='textSecondary' textAlign='center'>
+                            <Typography variant='h5' color='textSecondary' align='center'>
```
yipenghe: The makes the text align differently from the current homepage (second image is current)
Screenshot from 2019-11-13 10-00-11
Screenshot from 2019-11-13 10-00-19

epicfaace: Good point. I assume that was the intended look and feel though when the frontend was first designed, since otherwise it wouldn't have used the `text-align` prop. If we want to keep it the same look, though, I can just go ahead and remove the `align` prop.

thoughts @percyliang ?

percyliang: I don't have a strong preference.  Maybe left aligned?

yipenghe: Left aligned looks better IMO

epicfaace:
@yipenghe can you review?

PR #1651, opened 2019-11-13
Fix undefined element error

epicfaace:
Fix error:

"Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined."


yipenghe (approved):
lgtm

PR #1665, opened 2019-11-13
Remove margin for text edit cell

### LINKED ISSUE #1609: When editing a text cell, borders should match
Borders (grey and blue) should be at the same size

epicfaace:
Fixes #1609.

image


yipenghe (approved):
LGTM

PR #1669, opened 2019-11-13
Update genpath of owner_name to be owner instead

### LINKED ISSUE #1628: owner column still doesn't have valid value filled
Owner column was filled in ticket #1344. However, in production, the value is still `null` as following:
Screen Shot 2019-11-11 at 1 40 04 PM
This ticket is to fix this issue.

candicegjing:
Fixed #1628 .


percyliang (approved):
How was this working before on non-prod?

candicegjing:
> How was this working before on non-prod?

@percyliang My local testing environment's markdown maps the `owner` to `owner_name` by default. Is this the same to everyone? Or is there a setting that I can change that?

percyliang:
There should not be a difference...

PR #1671, opened 2019-11-14
Control modal entirely with react props

epicfaace:
Removing jquery from one more place! Fixes #1670

PR #1672, opened 2019-11-14
Use props instead of jquery for error message handling

epicfaace:
Also uses a dismissable bootstrap "Alert" for error messages.
Eventually, we should replace the loading sign with a similar
type of change. This is one step in the path of removing
jquery.

This PR builds off #1671, so merge this after that PR is merged.


epicfaace:
@yipenghe can you review?

yipenghe:
Code looks good, what should I be looking for on the view of frontend?

epicfaace:
> Code looks good, what should I be looking for on the view of frontend?

Hmm, I guess you could test by setting a hard-coded value to be set on this line: https://github.com/codalab/codalab-worksheets/pull/1672/files#diff-f36be677374a909f3276bc2317babfcdR1285

yipenghe (approved):
Make sure to be careful with the merge conflicts though

PR #1673, opened 2019-11-14
Add travis checks for frontend style/formatting

epicfaace:
Previously we didn't have this; we just had a precommit
hook that did this. I've also run prettier on all frontend files
so that the existing files do conform to the travis prettier
checks.

Also, add prettier to the precommit hook.


percyliang (approved):
We should get this in soon.  Can we also add something to `pre-commit.sh`?

epicfaace:
> We should get this in soon. Can we also add something to `pre-commit.sh`?

I don't think we'll need to do this -- because it will already run during precommit as long as node/npm is installed, through the `lint-staged` package.

PR #1675, opened 2019-11-14
Update bundle detail during runs

yipenghe:
Fetch new data to update bundle detail panel
including, stdout, stderr & file contents & metadata (run time)
#1611


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -260,6 +260,15 @@ class Worksheet extends React.Component {
             worksheet_uuid,
         )
             .done(() => {
+                Object.keys(this.state.checkedBundles).forEach((uuid) => {
```
epicfaace, line 263: I changed `.map` to `.forEach` because otherwise it caused an eslint error (since the function doesn't return anything)

q
```diff
@@ -0,0 +1,23 @@
+  autocomplete-master[m
```
epicfaace: remove this file

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -533,6 +542,9 @@ class Worksheet extends React.Component {
     toggleGlossaryModal = () => {
         this.setState({ showGlossaryModal: !this.state.showGlossaryModal });
     };
+    toggleGlossaryModal = () => {
```
epicfaace: remove this function (duplicate declaration)

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -57,6 +59,18 @@ class BundleDetail extends React.Component<
         this.fetchBundleContents();
     }
 
+    componentDidUpdate() {
```
epicfaace: I think it might be cleaner to put this logic as a call to `setInterval` in `componentDidMount` -- is there a particular reason why you put it in `componentDidUpdate`?

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -57,6 +59,18 @@ class BundleDetail extends React.Component<
         this.fetchBundleContents();
     }
 
+    componentDidUpdate() {
+        let newTime = new Date();
+        if (newTime.getTime() - this.state.lastFetchTime > 3000 && !this.state.fetching){
+            this.setState({fetching: true});
+            setTimeout(()=>{
+                this.setState({lastFetchTime: newTime.getTime(), fetching: false})
+                this.fetchBundleMetaData();
```
epicfaace, line 62: Do we need to fetch both bundle metadata and bundle contents? Might it be more efficient to only fetch metadata, and then we might be able to tell if the contents could change?

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
yipenghe: I will try this, just had it for demo

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
yipenghe: If you only fetch one of them, the other remains the same and won't be able to update, same as for fileBrower data

yipenghe:
This addresses #1681 #1611

Issue #1589, opened 2019-11-01
Performance bottleneck study on bundle_manager.py


candicegjing:
Some stats of running `bunlde_manager.py`.
Calling stack: 
1. `_run_iteration()`
    1. `_stage_bundles()`
    1. `_make_bundles()`
    1. `_schedule_run_bundles()`
          1. `_cleanup_dead_workers()`
          1. `_restage_stuck_starting_bundles()`
          1. `_bring_offline_stuck_running_bundles()`
          1. `_fail_on_too_many_resources()`
          1. `_acknowledge_recently_finished_bundles()`
          1. `_schedule_run_bundles_on_workers()`
    1. `_fail_unresponsive_bundles()`

The following table shows the performance stats that I collected from both production env and stanging env using roughly around 80 `created` bundles and 100 `staged` bundles. For stats collected in column (1), unlike the bundles in production, those bundles that I created in staging env don't have any dependencies. Time is measured in seconds.

| **Function Calls**                        | **Production**        |    **(1) Staging (Dev DB)**          |  **(2) Staging (Prod DB 11/04/2019)** |
|-------------------------------| ---------------- | ---------------------- | ---------------------- |
| _stage_bundles()                          | 4.893916368484497     |  0.007596492767333984   | 0.954134464263916 |
| _make_bundles()                           | 0.014322280883789062  |  0.00735783576965332    | 0.007351398468017578 | 
| _schedule_run_bundles()                   | 411.18663215637207    |  6.098901033401489      | 267.14445996284485 |
| _fail_unresponsive_bundles()              | 0.11637234687805176   |  0.07699465751647949    | 0.08106207847595215 |
| In _schedule_run_bundles() | | | |
| _cleanup_dead_workers()                   | 3.790855407714844e-05 |  6.67572021484375e-06   | 1.5974044799804688e-05 | 
| _restage_stuck_starting_bundles()         | 0.006481170654296875  |  0.0030193328857421875  | 0.003535747528076172  | 
| _bring_offline_stuck_running_bundles()    | 0.026740312576293945  |  0.006380558013916016   | 0.00631403923034668 | 
| _fail_on_too_many_resource()              | 299.062668800354      |  2.5921273231506348     | 129.30349040031433 |
| _acknowledge_recently_finished_bundles()  | 0.006404876708984375  |  0.003915309906005859   | 0.35086989402770996 |
| In _fail_on_too_many_resource() | |  | |
[...]

PR #1676, opened 2019-11-14
Make stage bundles run in background

candicegjing:
This is a **temporary solution** to let `stage_bundle` function not blocked by later functions. 
Based on the previous study #1589 , the function that dispatches jobs takes most of the time and blocks `stage_bundle` from moving newly `created` bundles to `staged` state. As a temporary solution here, we can set `stage_bundle` to run as a separate thread, so that newly created bundles will be fetched into the loop of checking their prerequisite of moving to the next stage (`staged` state).


candicegjing:
The commits went wrong in this PR. I am closing it and recreating another one.

PR #1595, opened 2019-11-05
Use saved disk_used when checking quota (fix #1589)

bkgoksel:
Thanks @candicegjing for figuring this out.

PR #1677, opened 2019-11-14
Make stage_bundles running in background

candicegjing:
This is a temporary solution to let stage_bundle function not blocked by later functions.
Based on the previous study #1589 , the function that dispatches jobs takes most of the time and blocks stage_bundle from moving newly created bundles to staged state. As a temporary solution here, we can set stage_bundle to run as a separate thread, so that newly created bundles will be fetched into the loop of checking their prerequisite of moving to the next stage (staged state).

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -62,6 +62,7 @@ class BundleManager(object):
         logger.info('Bundle manager running!')
         while not self._is_exiting():
             try:
+                self._run_stage_bundles_in_background(sleep_time)
                 self._run_iteration()
             except Exception:
                 traceback.print_exc()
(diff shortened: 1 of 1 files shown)


codalab/server/bundle_manager.py
```diff
@@ -62,6 +62,7 @@ def run(self, sleep_time):
         logger.info('Bundle manager running!')
         while not self._is_exiting():
             try:
+                self._run_stage_bundles_in_background(sleep_time)
```
percyliang, line 65: If we want to put this in a thread, we should create a thread once at the top (not in this loop).

codalab/server/bundle_manager.py
```diff
@@ -80,11 +81,32 @@ def _is_exiting(self):
...
+        :param sleep_time: number of seconds to sleep
+        :return:
+        """
+
+        def run():
+            while True:
+                self._stage_bundles()
+                time.sleep(sleep_time)
+
+        running_threads = [thread.getName() for thread in threading.enumerate()]
```
percyliang, line 101: This seems unnecessary; just create a thread once and be done with it.

percyliang (commented):
We should think more generally about restructuring this code in general, not just handling this one case.

Also, are there any atomicity / race conditions to worry about when we're introducing threading?

codalab/server/bundle_manager.py
candicegjing: I initially had the same thought. But what if something happened to this thread and then the newly submitted the jobs will never be moved from `created` to `staged` state. Then I started to put a check under the loop.

bkgoksel:
Hmm I'm not sure if this will drastically improve the end user experience. From what I understand, the actual time spent is mostly in actually scheduling `staged` bundled to workers and getting them on the `running`/`preparing` state. Making the `created` -> `staged` transition faster in the background wouldn't really reduce the time from bundle submission to getting bundles actually running in this case, right?

Also, did the performance profile of the bundle manager change at all after #1595 ? It would be really strange if it hasn't?

PR #1680, opened 2019-11-15
138: support host_worksheet for display table

### LINKED ISSUE #138: Add host_worksheets to genpath
We should be able to do

```
% display table host_worksheet
[bundle]
```

to show a column with a link to the host worksheet (if there is more than one, just return the first one).

To do this, several things need to happen:

1) We need to have get_bundle_infos return host_worksheet (similar to how we need the `-v` in  `cl info -f host_worksheets <bundle> -v`).

2) The returned value from worksheet_util.interpret_genpath should be a dict {type: 'worksheet', uuid: ..., name: ...} (or something like this; need to sync with displaying bundle links).

3) On the frontend, a link should be displayed.

teetone:
Fixes #138 

Changes:
Support `host_worksheet` for `display table` markup.

How to test:
Use `display table host_worksheet` or `display table host_worksheets` for bundles.

Screenshots:


frontend/src/components/worksheets/items/TableItem.js
```diff
@@ -141,14 +152,21 @@ class TableRow extends React.Component {
         var baseUrl = this.props.url;
         var uuid = this.props.uuid;
         var columnWithHyperlinks = this.props.columnWithHyperlinks;
+        var worksheet_name = this.props.worksheet_name;
```
percyliang: Use camelCase in Javascript.

frontend/src/components/worksheets/items/TableItem.js
```diff
@@ -141,14 +152,21 @@ class TableRow extends React.Component {
...
+        var worksheet_url = this.props.worksheet_url;
         var rowCells = this.props.headerItems.map(function(headerKey, col) {
             var rowContent = rowItems[headerKey];
 
             // See if there's a link
             var url;
-            if (col == 0) {
+            if (headerKey === 'host_worksheets' || headerKey === 'host_worksheet') {
```
percyliang: Would be better to pick one (e.g., `host_worksheet`) so that people are encouraged to be consistent.

frontend/src/components/worksheets/items/TableItem.js
```diff
@@ -89,9 +89,18 @@ class TableItem extends React.Component {
             if (rowItems[0][x] && rowItems[0][x]['path']) columnWithHyperlinks.push(x);
         });
         var bodyRowsHtml = rowItems.map(function(rowItem, rowIndex) {
-            var rowRef = 'row' + rowIndex;
-            var rowFocused = self.props.focused && rowIndex == self.props.subFocusIndex;
-            var url = '/bundles/' + bundleInfos[rowIndex].uuid;
+            let bundleInfo = bundleInfos[rowIndex];
+            let worksheets = bundleInfo.host_worksheets;
```
percyliang: Move this closer to where it's used.

percyliang (changes requested):
One thing I'm worried about is a possible performance hit when a bundle that shows up on 10000 worksheets.  The DB query should just get one host worksheet if that's all we're using (though we have to be careful not to alter other functionality).

codalab/model/bundle_model.py
```diff
@@ -258,11 +258,59 @@ def get_children_uuids(self, uuids):
...
+        """
+        if n <= 0:
+            return dict()
+        if n == 1:
+            result = self.get_single_host_worksheet_uuid(bundle_uuids)
+            return dict(
+                (uuid, [worksheet]) for uuid, worksheet in result.items() if worksheet is not None
+            )
+
+        with self.engine.begin() as connection:
```
teetone, line 273: @percyliang: I looked into lots of different approaches and If we wanted to try `n` different worksheet uuid's per bundle, I think this is the best solution. Other solutions require `order_by` worksheet_uuids across the bundles.

codalab/model/bundle_model.py
```diff
@@ -258,11 +258,59 @@ def get_children_uuids(self, uuids):
             result[row.parent_uuid].append(row.child_uuid)
         return result
 
+    def get_n_host_worksheet_uuids(self, bundle_uuids, n):
+        """
+        Get up to n host_worksheet uuids per bundle uuid. n of 0 or less will return an empty dictionary.
```
percyliang: n shouldn't ever be negative.  I'd rather assert that than accept that.

codalab/model/bundle_model.py
```diff
@@ -258,11 +258,59 @@ def get_children_uuids(self, uuids):
...
+                    cl_worksheet_item.c.bundle_uuid.in_(bundle_uuids)
+                )
+            ).fetchall()
+        result = dict((uuid, set()) for uuid in bundle_uuids)
+        for row in rows:
+            worksheets = result[row.bundle_uuid]
+            if len(worksheets) < n:
+                result[row.bundle_uuid].add(row.worksheet_uuid)
+        # Convert set into a list
+        for uuid in result.keys():
```
percyliang: don't need `.keys()`

codalab/model/bundle_model.py
percyliang: Don't we need to put a limit in the actual query so that if a bundle shows up on a ton of worksheets but we're only getting 5, we're not too wasteful?  Maybe something in here would help? https://www.xaprb.com/blog/2006/12/07/how-to-select-the-firstleastmax-row-per-group-in-sql/

percyliang: I'm worried about a performance hit since I believe this logic is going to be exercised on every call now...can you provide some indication of the change in performance (on dev, where there's a lot of data)?

codalab/model/bundle_model.py
```diff
@@ -258,11 +258,59 @@ def get_children_uuids(self, uuids):
...
+            result[uuid] = list(result[uuid])
+        return result
+
+    def get_single_host_worksheet_uuid(self, bundle_uuids):
+        """
+        Get any single host_worksheet uuid per bundle uuid. This is achieved by a simple group by.
```
percyliang: Can we unify this function and the n-best?  It seems like it should be a special case...

percyliang (changes requested):
I'm mostly worried about performance of the N-bundle-per-worksheet logic.  If it turns out to be too complicated, then I'd just go with at most one bundle per worksheet and do the filtering.  Most of the time I think a bundle should only show up on one worksheet...and at least this will be an improvement over what we have now, which is no functionality.

codalab/model/bundle_model.py
teetone: @percyliang: On my local machine, I created ~1000 bundles, where each bundle has ~1000 host worksheets.

- Single worksheet retrieval + checking read permissions took ~0.055 seconds on average.
- Retrieving all but limiting to 5 worksheets + checking read permissions took ~0.078 seconds on average.
- Retrieving all (no limit) + checking read permissions took ~0.4 seconds on average.

I'm leaning towards the limiting to 5 worksheets approach. It will most likely yield correct results to the user and fetching all worksheet uuid's for multiple bundles was just a bit slower than the group by.

percyliang: Cool.  But does the current logic actually retrieve 5 *per* bundle (see my first comment)?  I wonder still if we can just have a single function instead of two (n=1 and n>1)?

teetone: @percyliang: Unfortunately it doesn't. It still retrieves all. I just limit the read permission checks to 5, making it faster than checking all. 

Next step, I can try `order by` + `limit` to get exactly 5 and measure its performance compared to the current implementations. I speculate it will be slower because of the `order by`. If it's fast, I can use it to handle both n=1 and n>1 cases.

teetone: @percyliang: I discovered a new solution that's cleaner and faster than the previous solutions. For 5 worksheets + checks it takes about ~0.041 seconds on average.

yipenghe:
It's working on dashboard
Screenshot from 2019-11-26 00-53-46

But not on public home..
Screenshot from 2019-11-26 00-50-09

teetone:
In public home, you have it as host_worksheets. It should be host_worksheet.
    On Tuesday, November 26, 2019, 12:55:08 AM PST, yipenghe <notifications@github.com> wrote:  
 
 
It's working on dashboard


But not on public home..

yipenghe:
> In public home, you have it as host_worksheets. It should be host_worksheet. On Tuesday, November 26, 2019, 12:55:08 AM PST, yipenghe <notifications@github.com> wrote: It's working on dashboard But not on public home.. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Yeah I figured...

frontend/src/components/worksheets/items/TableItem.js
```diff
@@ -141,14 +151,21 @@ class TableRow extends React.Component {
...
         var columnWithHyperlinks = this.props.columnWithHyperlinks;
+        var worksheetName = this.props.worksheetName;
+        var worksheetUrl = this.props.worksheetUrl;
         var rowCells = this.props.headerItems.map(function(headerKey, col) {
             var rowContent = rowItems[headerKey];
 
             // See if there's a link
             var url;
-            if (col == 0) {
+            if (headerKey === 'host_worksheet' && worksheetUrl !== undefined) {
```
yipenghe, line 161: Would it be better to check if worksheetUrl evaluates to true ( && worksheetUrl), so it handles null/''/undefined?

codalab/rest/util.py
```diff
@@ -114,39 +120,40 @@ def get_bundle_infos(
                 if child_perms[child_uuid] >= GROUP_OBJECT_PERMISSION_READ
             ]
 
+    if get_single_host_worksheet:
+        # Get back a dict of bundle uuids -> list of worksheet uuids (max of 5)
```
yipenghe: Slightly weird comment, it should explain the whole block instead of just the get_n_host_worksheet_uuids function right?  Would be better to explain the block logic (first get at most 5 host worksheet, filter with permission and ok to have no worksheet returned if none of 5 qualifies)

codalab/rest/util.py
```diff
@@ -46,29 +46,35 @@ def resolve(keyword):
...
     ignore_not_found=True,
+    model=None,
 ):
     """
     Return a map from bundle uuid to info.
 
     :param Collection[str] uuids: uuids of bundles to fetch
     :param bool get_children: include children
+    :param bool get_single_host_worksheet: include one host_worksheet per bundle uuid
     :param bool get_host_worksheets: include host worksheets
```
yipenghe: nit: maybe we should adjust this comment to be "include all host worksheets"

yipenghe (commented):
Overall looks good to me

codalab/model/bundle_model.py
```diff
@@ -258,11 +258,38 @@ def get_children_uuids(self, uuids):
             result[row.parent_uuid].append(row.child_uuid)
         return result
 
+    def get_n_host_worksheet_uuids(self, bundle_uuids, n):
```
percyliang: I'd rename to `get_host_worksheet_uuids` and then `n` to `max_worksheets`.

percyliang (approved):
Cool!

yipenghe:
Since this is on the old ui, can you check it works with the current frontend?

frontend/src/components/worksheets/items/TableItem.js
teetone: I will leave it as is. In this situation, `worksheetUrl` can only be `undefined`. 

In general, we should avoid falsies (`&& worksheetUrl`). This is not idiomatic in JavaScript and in the lower level, it starts a chain of casting types, since in isolation it suggests `worksheetUrl` is a boolean, which in this case it is not.

teetone:
> Since this is on the old ui, can you check it works with the current frontend?

@yipenghe: Good call, I just cherry-picked it and it works. Thanks.

PR #1687, opened 2019-11-19
fix-time-quota-miscalculation

candicegjing:
**Note that I haven't been able to test this PR locally yet due to both of the laptops I have are not working functionally in different ways.** 
Since I don't want to block the issue gets fixed, I read through the code and did some investigation on the docker api and those users who experienced unusual time quota increases, then I came up with this PR. 
Basically, I did a bunch of checks on those finished containers and got an impression that the stats returning from `docker inspect` is more reliable than manually calculating the time. There are some open questions we still need to decide:
1. Should we keep updating the `container_time_total` when the container is still `running`? The current logic (before this PR) seems to do so. 
2. This question is related to 1st, if we do want to keep updating the `container_time_total` even if the container is still `running`, then when we inspect the docker container, should we calculate the difference between now and start time? Or give it a default value?

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -9,7 +9,8 @@ created if not.
 import logging
 import os
 import docker
-
+from dateutil import parser, tz
+import datetime
 from codalab.lib.formatting import parse_size
 
 
@@ -256,3 +257,25 @@ def check_finished(container):
             failure_msg = 'Memory limit exceeded.'
         return (True, exitcode, failure_msg)
     return (False, None, None)
+
+
+@wrap_exception('Unable to check Docker container running time')
+def get_container_running_time(container):
+    # Get the current container
+    container = client.containers.get(container.id)
+    # Load this container from the server again and update attributes with the new data.
+    container.reload()
+    # Calculate the start_time of the current container
+    start_time = container.attrs['State']['StartedAt']
+    # Calculate the end_time of the current container. If 'Status' of the current container is not 'exited',
+    # then using the current time as end_time
+    end_time = (
+        container.attrs['State']['FinishedAt']
+        if container.attrs['State']['Status'] == 'exited'
+        else str(datetime.datetime.now().replace(tzinfo=tz.tzutc()))
+    )
+    # Docker reports both the start_time and the end_time in ISO format. We currently use dateutil.parser.isoparse to
+    # parse them. In Python3.7 or above, the built-in function datetime.fromisoformat() can be used to parse ISO
+    # formatted datetime string directly.
+    container_running_time = parser.isoparse(end_time) - parser.isoparse(start_time)
+    return container_running_time.total_seconds()
--- a/codalab/worker/local_run/local_run_state.py
+++ b/codalab/worker/local_run/local_run_state.py
@@ -339,7 +339,7 @@ class LocalRunStateMachine(StateTransitioner):
                 disk_utilization=self.disk_utilization[run_state.bundle.uuid]['disk_utilization']
             )
 
-            container_time_total = time.time() - run_state.container_start_time
+            container_time_total = docker_utils.get_container_running_time(run_state.container)
             run_state = run_state._replace(
                 container_time_total=container_time_total,
                 container_time_user=run_stats.get(
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ six==1.11.0
 SQLAlchemy==1.0.8
 watchdog==0.8.3
 fusepy==2.0.4
+python-dateutil


codalab/worker/docker_utils.py
```diff
@@ -252,3 +254,17 @@ def check_finished(container):
...
+    # Identical to the docker inspect container command.
+    container_stats = apiclient.inspect_container(container.id)
+    start_time = container_stats['State']['StartedAt']
+    end_time = (
+        container_stats['State']['FinishedAt']
+        if container_stats['State']['Status'] == 'exited'
+        else datetime.datetime.now().replace(tzinfo=tz.tzutc())
+    )
+    container_running_time = parser.isoparse(end_time) - parser.isoparse(start_time)
+    return container_running_time.seconds
```
bkgoksel: Instead of creating a raw API client and calling `inspect_container`, you can call `container.reload()` then use the `container.attrs` dict to access the inspection fields of the container. For this purpose the following should work:
```
container.reload()
start_time = container.attrs['State']['StartedAt']
. . .
```
https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container

bkgoksel (changes requested):
I think this can be done without using an `APIClient`, so let's try to go that way. LGTM otherwise.

BTW what was the problem with the old way of doing `time.time() - run_state.container_start_time`?

bkgoksel:
For 1, I think the answer can stay yes as it is now, as long as that calculation isn't grossly wrong. And for 2, the first option sounds reasonable to me, again as long as we can calculate that time without going wrong.

Looks like this PR is consistent with this set of answers, so I think it is good to go.

candicegjing:
@bkgoksel There is something wrong with the time quota calculation. Some users experienced an unusual time quota increase, like this #1533

codalab/worker/docker_utils.py
```diff
@@ -252,3 +254,17 @@ def check_finished(container):
...
+
+@wrap_exception('Unable to check Docker container running time')
+def get_container_running_time(container):
+    # Identical to the docker inspect container command.
+    container_stats = apiclient.inspect_container(container.id)
+    start_time = container_stats['State']['StartedAt']
+    end_time = (
+        container_stats['State']['FinishedAt']
+        if container_stats['State']['Status'] == 'exited'
+        else datetime.datetime.now().replace(tzinfo=tz.tzutc())
```
percyliang: Can you explain the timezone logic here?

percyliang (commented):
I still don't quite understand why the logic wasn't working before...

Also, when you do `cl run date`, is the container_time_total going to be 0 seconds?  It isn't now, but it'd be nice if it were.

codalab/worker/docker_utils.py
candicegjing: This is because the DateTime received from docker rest API is using the UTC timezone. I just want to match the current time with what we received from the docker rest API so that the calculation of the total time used is correct.

candicegjing: For example, this is the timestamp received from rest api: `"StartedAt": "2019-09-29T03:00:57.558381851Z"`. It's in ISO 8601 format and `Z` stands for UTC timezone.

percyliang: What about using `datetime.utcnow()`?  Also, seems like `end_time` can either be a string or a `datetime` object, which is non-ideal.

candicegjing: I tried that earlier today, but got the following:
```
>>> now = datetime.datetime.utcnow()
>>> now.isoformat()
'2019-11-19T19:45:56.941517'
>>> parser.isoparse(now)
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/dateutil/parser/isoparser.py", line 37, in func
    return f(self, str_in, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/dateutil/parser/isoparser.py", line 134, in isoparse
    components, pos = self._parse_isodate(dt_str)
  File "/usr/local/lib/python3.6/site-packages/dateutil/parser/isoparser.py", line 208, in _parse_isodate
    return self._parse_isodate_common(dt_str)
  File "/usr/local/lib/python3.6/site-packages/dateutil/parser/isoparser.py", line 213, in _parse_isodate_common
    len_str = len(dt_str)
TypeError: object of type 'datetime.datetime' has no len()
```
But let me double check.

candicegjing:
I did more testing on this and the `cl run date` commands returns the following result. 
Screen Shot 2019-11-25 at 5 17 51 PM
From the backend worker log, I got the following timestamps:
```
2019-11-26 01:15:57,625 >>>StartedAt = 2019-11-26T01:15:54.7762612Z
2019-11-26 01:15:57,627 >>>parsing StartedAt= 2019-11-26 01:15:54.776261+00:00
2019-11-26 01:15:57,627 >>>FinishedAt = 2019-11-26T01:15:54.8272588Z
2019-11-26 01:15:57,628 >>>parsing FinishedAt= 2019-11-26 01:15:54.827258+00:00
2019-11-26 01:15:57,629 >>>Container total running time = 0.050997
```
From the UI, we rounded the value `0.050997` sec to `0.1` sec. Let me know if we want to not round the value.

codalab/worker/docker_utils.py
```diff
@@ -256,3 +257,25 @@ def check_finished(container):
...
+    # Load this container from the server again and update attrs with the new data.
+    container.reload()
+    # Calculate the start_time of the current container
+    start_time = container.attrs['State']['StartedAt']
+    # Calculate the end_time of the current container. If 'Status' of the current container is not 'exited',
+    # then using the current time as end_time
+    end_time = (
+        container.attrs['State']['FinishedAt']
+        if container.attrs['State']['Status'] == 'exited'
+        else datetime.datetime.now().replace(tzinfo=tz.tzutc()).__str__()
```
percyliang: At least do `str(...)` rather than `__str__`.  But it's better to convert `start_time` and `end_time` once from strings into `datetime` rather than converting to `str` and back.

percyliang (changes requested):

codalab/worker/docker_utils.py
percyliang: Also, why not `datetime.datetime.utcnow()`?

candicegjing: Because I want to exactly match the format of `end_time` with `start_time` when we parse them using `parser.isoparse()`. See the following example:
1. `start_time` is copied directly from attribute `container.attrs['State']['StartedAt']`. When we parse the `start_time` using `parser.isoparse`, there is the `tzinfo` in the return value.
```
>>> start = "2019-11-26T01:15:54.7762612Z"
>>> parser.isoparse(start)
datetime.datetime(2019, 11, 26, 1, 15, 54, 776261, tzinfo=tzutc())
```
2. However, if we use `datetime.datetime.utcnow()` without specifing `tzinfo`, there is no `tzinfo` attribute returned in the datetime object.
```
>>> parser.isoparse(str(datetime.datetime.utcnow().isoformat()))
datetime.datetime(2019, 11, 26, 5, 38, 38, 745842)
```
3. When we explicitly specified `tzinfo` in `datetime` object, we also get the `tzinfo` in the return result.
```
>>> end=str(datetime.datetime.now().replace(tzinfo=tz.tzutc()))
>>> parser.isoparse(end)
datetime.datetime(2019, 11, 26, 5, 36, 41, 91737, tzinfo=tzutc())
```
Therefore, I decided to choose `str(datetime.datetime.now().replace(tzinfo=tz.tzutc()))` instead of `datetime.datetime.utcnow()`. Let me know if we want to swith over.

percyliang (approved):

codalab/worker/docker_utils.py
percyliang: Ok, that's fine.

PR #1688, opened 2019-11-20
Use local image when pull from docker failed

### LINKED ISSUE #1685: new run should work without Internet
Currently, it tries to contact docker hub, and if it fails, the run crashes, presumably because we do a pull every time.  Don't fail hard on the pull, so that if we're not connected to the Internet and the docker image exists locally, we just use that.

This is mostly useful for demoing CodaLab when the wireless isn't working properly...

candicegjing:
Fixed #1685.
This PR added the option to use the local image if pulling from the docker hub failed due to any reasons. 
**I still haven't got a chance to test this PR as my laptops are all broken >.<**

### DIFF
--- a/codalab/worker/local_run/docker_image_manager.py
+++ b/codalab/worker/local_run/docker_image_manager.py
@@ -135,6 +135,32 @@ class DockerImageManager:
         :param image_spec: Repo image_spec of docker image being requested
         :returns: A DockerAvailabilityState object with the state of the docker image
         """
+
+        def image_availability_state(image_spec, success_message, failure_message):
+            """
+            Try to get the image specified by image_spec from host machine.
+            Return ImageAvailabilityState.
+            """
+            try:
+                image = self._docker.images.get(image_spec)
+                digests = image.attrs.get('RepoDigests', [image_spec])
+                digest = digests[0] if len(digests) > 0 else None
+                with self._lock:
+                    self._image_cache[digest] = ImageCacheEntry(
+                        id=image.id,
+                        digest=digest,
+                        last_used=time.time(),
+                        virtual_size=image.attrs['VirtualSize'],
+                        marginal_size=image.attrs['Size'],
+                    )
+                return ImageAvailabilityState(
+                    digest=digest, stage=DependencyStage.READY, message=success_message
+                )
+            except Exception:
+                return ImageAvailabilityState(
+                    digest=None, stage=DependencyStage.FAILED, message=failure_message
+                )
+
         if ':' not in image_spec:
             # Both digests and repo:tag kind of specs include the : character. The only case without it is when
             # a repo is specified without a tag (like 'latest')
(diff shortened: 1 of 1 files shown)


codalab/worker/local_run/docker_image_manager.py
```diff
@@ -136,6 +136,32 @@ def get(self, image_spec):
         :param image_spec: Repo image_spec of docker image being requested
         :returns: A DockerAvailabilityState object with the state of the docker image
         """
+
+        def get_local_image(image_spec, success_message, failure_message):
+            """
+            Try to get the image specified by image_spec from local.
```
percyliang: What do you mean "from local"?

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -136,6 +136,32 @@ def get(self, image_spec):
         :param image_spec: Repo image_spec of docker image being requested
         :returns: A DockerAvailabilityState object with the state of the docker image
         """
+
+        def get_local_image(image_spec, success_message, failure_message):
```
percyliang: Maybe to be more precise, should be image_availability_state?

percyliang (commented):
Can you explain (in comments) how this code implements the desired logic of using the local image when pull fails?

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -160,24 +186,20 @@ def get(self, image_spec):
...
+                                failure_message='Image {} was downloaded successfully, but it can not be found locally due to unknown reasons'.format(
+                                    image_spec
+                                ),
                             )
                         else:
-                            status = ImageAvailabilityState(
-                                digest=None,
-                                stage=DependencyStage.FAILED,
-                                message=self._downloading[image_spec]['message'],
+                            status = get_local_image(
```
candicegjing: When pulling image from DockerHub fails, it then tries to scan local images to see if it can find one specified by `image_spec` in function `get_local_image()`.

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -136,6 +136,32 @@ def get(self, image_spec):
...
+                        id=image.id,
+                        digest=digest,
+                        last_used=time.time(),
+                        virtual_size=image.attrs['VirtualSize'],
+                        marginal_size=image.attrs['Size'],
+                    )
+                return ImageAvailabilityState(
+                    digest=digest, stage=DependencyStage.READY, message=success_message
+                )
+            except Exception:
```
candicegjing, line 159: There are only two exceptions here based on documentation of [docker rest API:get(image_id) ](https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.get)

codalab/worker/local_run/docker_image_manager.py
candicegjing: from local machine...

epicfaace: @candicegjing can you update the docstring to clarify that?

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -136,6 +136,32 @@ def get(self, image_spec):
...
+                return ImageAvailabilityState(
+                    digest=None, stage=DependencyStage.FAILED, message=failure_message
```
epicfaace, line 161: Would we instead want to send out the specific failure message for this operation as opposed to a generic failure message ("due to unknown reasons")?

codalab/worker/local_run/docker_image_manager.py
```diff
@@ -160,24 +186,20 @@ def get(self, image_spec):
...
+                                ),
                             )
                         else:
-                            status = ImageAvailabilityState(
-                                digest=None,
-                                stage=DependencyStage.FAILED,
-                                message=self._downloading[image_spec]['message'],
+                            status = image_availability_state(
+                                image_spec,
+                                success_message='Image {} can not be downloaded from the DockerHub but it is found locally'.format(
```
epicfaace:
```suggestion
                                success_message='Image {} could not be downloaded from DockerHub but was found locally'.format(
```

codalab/worker/local_run/docker_image_manager.py
candicegjing: @epicfaace Updated.

codalab/worker/local_run/docker_image_manager.py
candicegjing: @epicfaace Why do we want to change to `could`? We are sure that the image can't be downloaded at this point. Let me know.

codalab/worker/local_run/docker_image_manager.py
candicegjing: Yes, we are sending a specific failure message right now for different cases.

candicegjing:
@percyliang @bkgoksel @epicfaace This PR is tested in local machine without connecting to any internet services. The testing bundle ran successfully.

PR #1690, opened 2019-11-20
Display title as Loading when info is not ready

### LINKED ISSUE #1661: [New Frontend] Display a more descriptive title when a worksheet is loading?
The worksheet name says "Untitled" when you're waiting for it to load. Maybe instead display something like "Loading worksheet data..." or some other icon to show it's loading.

epicfaace:
Fixes #1661.


yipenghe:
The placeholder doesn't seem to be working due to #1689 , plus wouldn't this cause untitled worksheets all display 'Loading'?

yipenghe:
I think for this, we should do the following: check 
1. if this.info && this.info.title => set to info.title 
2. if this.info && !this.info.title => leave undefined (actually no title set before)
3. no this.info => Loading (info data isn't ready yet)

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -55,7 +55,7 @@ export default ({
                                 canEdit={canEdit}
                                 fieldName='title'
                                 value={info && info.title}
```
yipenghe:
```suggestion
                                value={info ? info.title : 'Loading...'}
```

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -55,7 +55,7 @@ export default ({
...
-                                placeholder='Untitled'
+                                placeholder='Loading...'
```
yipenghe:
```suggestion
```

yipenghe (commented):
This should do the trick

yipenghe:
Also would be good to have #1693 when info is not ready, maybe first merge that PR and then add {!info && \<Loading />} to Worksheet.js after {this.state.updating && \<Loading />}

PR #1695, opened 2019-11-20
Refactor fetchInterpretedWorksheet calls into separate "actions.js" file

epicfaace:
This is the first step in maintaining a separation from react components
with the network actions that they call. This will make it easier
to test, maintain, etc.


epicfaace:
@yipenghe what do you think about this structure for refactoring network calls? If you're good with it I  can do it for all the network calls that exist currently

yipenghe:
> @yipenghe what do you think about this structure for refactoring network calls? If you're good with it I can do it for all the network calls that exist currently

Looks great, more organized, thanks!

yipenghe (approved):
Let's merge this first so when if we have new network calls we can add to this file

percyliang:
Why use the frontend branch as opposed to master?

epicfaace:
Oops, that was a mistake. Thanks for catching

PR #1707, opened 2019-11-22
Improve checking resource failure function

### LINKED ISSUE #1706: [Bundle Manager] Improve the way of resource checking in _fail_on_too_many_resources()
_fail_on_too_many_resources() function currently checks for resources failure multiple times in different aspects(e.g. memory, disk quota, user quota) for a single bundle. To make the bundle manager run faster, the logic could be changed to as long as one check failed, we fail the bundle right away.

candicegjing:
Fixed #1706 .
This PR is to make the function `_fail_on_too_many_resources()` in the bundle manager run a bit faster.
The original logic tries to check for three conditions (memory, disk quota, and time quota) for each bundle to see if it was requested more resources than the system currently has. The logic could be changed to as long as one check failed, we fail the bundle right away, which can improve the performance a bit. It might be helpful to understand which resource check fails more frequently than others. Then we could reorder those checks to make it run faster.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -630,46 +630,33 @@ class BundleManager(object):
         """
         for bundle in self._model.batch_get_bundles(state=State.STAGED, bundle_type='run'):
             bundle_resources = self._compute_bundle_resources(bundle)
-            failures = []
 
-            failures.append(
-                self._check_resource_failure(
+            failure_message = self._check_resource_failure(
                     bundle_resources.disk,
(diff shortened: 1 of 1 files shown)


percyliang (commented):
I think we should keep the original logic.  This doesn't seem to affect efficiency by much, and it's nice to have all the failure messages.

candicegjing:
> I think we should keep the original logic. This doesn't seem to affect efficiency by much, and it's nice to have all the failure messages.

If we want to have all failure messages returning to user, then this logic shouldn't be changed. Closing this PR.

Issue #1709, opened 2019-11-22
[New Frontend] Shortcut `i` does not work to add a cell below


teetone:
**Describe the bug**
Shortcut `i` key press event is ignored for the new front end. It should "add a cell right below the current focus".

**Codalab Information**
-Browser: Chrome

**To Reproduce**
Steps to reproduce the behavior :
1. Go to a worksheet.
2. Focus on a table cell.
3. Press `i`.

**Expected behavior**
Event should be handled as expected.

percyliang:
Don't 'a t', 'a u', 'a r' cover it for creating text, upload and run bundles?

yipenghe:
> Don't 'a t', 'a u', 'a r' cover it for creating text, upload and run bundles?

It might be just the keyboard shortcut help haven't been updated, I will double check

yipenghe:
Yeh this is just the help haven't been updated, closing

Issue #1702, opened 2019-11-21
[New Frontend] Update keyboard shortcuts help


candicegjing:
Keyboard shortcuts help needs to be updated (enter, u, a, don’t do what is written)

Issue #1701, opened 2019-11-21
[New Frontend] Bug in deleting markdown block


candicegjing:
If delete markdown block, there’s a bug where the cursor is pointed to the right place, but hitting delete again triggers the deletion of bundles further down.

yipenghe:
I think this shouldn't be too blocking for production

yipenghe:
It's actually pretty annoying if you wan't to delete multiple markdown items

PR #1718, opened 2019-11-22
Update shortcut help

yipenghe:
Addressing #1702 to avoid issues like #1709 
The code is more or less just moving around and change the text, so you can see the screenshot for review instead.
I think it might be better to classify these shortcuts into sections, opening this PR just to gather opinions
Screenshot from 2019-11-21 23-42-05

### DIFF
--- a/frontend/src/components/worksheets/ExtraWorksheetHTML/ExtraWorksheetHTML.js
+++ b/frontend/src/components/worksheets/ExtraWorksheetHTML/ExtraWorksheetHTML.js
@@ -34,31 +34,31 @@ const GlossaryModal = ({ show, toggle }) => (
                     </tr>
                     <tr>
                         <td>
-                            <kbd>g g (double click)</kbd>
+                            <kbd>g g</kbd>
                         </td>
                         <td>Move to beginning of worksheet</td>
                     </tr>
                     <tr>
                         <td>
-                            <kbd>t</kbd>
+                            <kbd>a t</kbd>
                         </td>
                         <td>Add a cell right below the current focus</td>
                     </tr>
                     <tr>
                         <td>
-                            <kbd>u</kbd>
+                            <kbd>a u</kbd>
                         </td>
                         <td>Upload a file right below the current focus</td>
                     </tr>
                     <tr>
                         <td>
-                            <kbd>a+r</kbd>
+                            <kbd>a r</kbd>
                         </td>
                         <td>Add a new run</td>
                     </tr>
                     <tr>
                         <td>
-                            <kbd>e</kbd>
+                            <kbd>shift+e</kbd>
                         </td>
                         <td>Edit worksheet in source mode</td>
                     </tr>
@@ -79,7 +79,7 @@ const GlossaryModal = ({ show, toggle }) => (
                     </tr>
                     <tr>
                         <td>
-                            <kbd>ctrl/command+enter</kbd>
+                            <kbd>ctrl/cmd+enter</kbd>
                         </td>
                         <td>Save current edit</td>
                     </tr>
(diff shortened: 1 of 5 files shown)


percyliang:
Some comments:
1) 'double click' is weird - this is supposed to be keyboard shortcuts - remove?
2) should be 'a t', 'a u' to be consistent with 'a r'.
3) given that we're relying less on raw edit, maybe we should change 'e' to 'shift+e' instead...?
4) can we replace 'command' with the apple sign to make it look cleaner?
5) backspace/del similarly doesn't look good - maybe use the icons?  Also, the description isn't quite right...it doesn't exit markdown editing...

In general, I think some of these are confusing because they only apply in certain contexts...I think organizing it is a good idea.

There are also a few things missing like 'a k' for kill selected (let's not call it 'v', which is counterintuitive) and detach.  Perhaps sync with @Levilian to make sure this documentation reflects reality.

yipenghe:
@Levilian Can you take a look?

epicfaace (approved):

yipenghe:
> Some comments:
> 
> 1. 'double click' is weird - this is supposed to be keyboard shortcuts - remove?
> 2. should be 'a t', 'a u' to be consistent with 'a r'.
> 3. given that we're relying less on raw edit, maybe we should change 'e' to 'shift+e' instead...?
> 4. can we replace 'command' with the apple sign to make it look cleaner?
> 5. backspace/del similarly doesn't look good - maybe use the icons?  Also, the description isn't quite right...it doesn't exit markdown editing...
> 
> In general, I think some of these are confusing because they only apply in certain contexts...I think organizing it is a good idea.
> 
> There are also a few things missing like 'a k' for kill selected (let's not call it 'v', which is counterintuitive) and detach. Perhaps sync with @Levilian to make sure this documentation reflects reality.

1. will remove
2. I can change this, but t and u are quite handy...
3. I can also change this
4 & 5 I think I will hold on this as long as the key-description pair is correct

I don't think reusing 'k' for two quite different functionalities is a good idea given that 'k' might be used very frequently to traverse...but if we really want 'a k' to do it I can do that as well..

yipenghe:
Screenshot from 2019-11-25 15-14-02

percyliang:
True, 't' and 'u' are more convenient...but consistency is important...

I think I'd still prefer 'a k' because it's much easier to remember than 'v'.

space for selecting everything in the current cell seems quite aggressive - seems easy to hit by accident and doesn't seem that common.  Should we remove this?

epicfaace:
@yipenghe one thing to add/note: "backspace/del" doesn't only "Delete selected bundles", but it also deletes the focused worksheet item. Actually, to be more precise, "backspace" deletes the focus worksheet item, but "delete" doesn't do so (this is probably a bug?)

yipenghe:
> @yipenghe one thing to add/note: "backspace/del" doesn't only "Delete selected bundles", but it also deletes the focused worksheet item. Actually, to be more precise, "backspace" deletes the focus worksheet item, but "delete" doesn't do so (this is probably a bug?)

I think it might be related to this issue https://github.com/codalab/codalab-worksheets/issues/1701 do you have time to take a look?

yipenghe:
> True, 't' and 'u' are more convenient...but consistency is important...
> 
> I think I'd still prefer 'a k' because it's much easier to remember than 'v'.
> 
> space for selecting everything in the current cell seems quite aggressive - seems easy to hit by accident and doesn't seem that common. Should we remove this?

sure, I will comment out the select all keyboard shortcut, but leaving it in case we want this in the future

epicfaace (approved):
Looks good. I'd suggest removing the "space" code altogether instead of commenting it out -- we can always add it back later.

Issue #1708, opened 2019-11-22
[New Frontend] Error: Please provide worksheet uuid. Request for a better 404 page.


teetone:
**Describe the bug**
When going to https://worksheets-dev.codalab.org/worksheets, you will run into this following unexpected error screen. A user reported this issue when he was logged in, but I just reproduced it without logging in.

![worksheets](https://user-images.githubusercontent.com/16793796/69402738-9b252600-0cad-11ea-85dd-ec7fac5a68a5.png)

**Codalab Information**
- Environment: Dev
- Browser: Chrome

**Expected behavior**
We should show a 404 invalid url error page instead since it is indeed an invalid url. We also should show a better 404 page instead of the current one (see below).

Current:
![Screen Shot 2019-11-21 at 10 30 43 PM](https://user-images.githubusercontent.com/16793796/69403097-93b24c80-0cae-11ea-9ecf-0284315fa1e6.png)

yipenghe:
I think this is a redirection issue, we should redirect to the dashboard worksheet, good catch

PR #1719, opened 2019-11-22
redirect to home/dashboard when going to /worksheets link

yipenghe:
#1708 when user types in /worksheets, redirect to public home (not logged in) or their dashboard (logged in) instead of asking for a uuid

### DIFF
--- a/frontend/src/components/worksheets/WorksheetNameSearch.js
+++ b/frontend/src/components/worksheets/WorksheetNameSearch.js
@@ -12,11 +12,11 @@ export default class extends React.Component {
     }
     async componentDidMount() {
         let { name } = queryString.parse(this.props.location.search);
-        if (name === undefined && this.props.auth.isAuthenticated){
+        if (name === undefined && this.props.auth.isAuthenticated) {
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/WorksheetNameSearch.js
```diff
@@ -11,12 +11,20 @@ export default class extends React.Component {
         };
     }
     async componentDidMount() {
-        const { name } = queryString.parse(this.props.location.search);
+        let { name } = queryString.parse(this.props.location.search);
+        if (name === undefined && this.props.auth.isAuthenticated) {
```
teetone: nit: Would it be better to check if the name is undefined once and nest the if-else statement that checks `isAuthenticated` or use a ternary operator?

teetone (approved):

frontend/src/components/worksheets/WorksheetNameSearch.js
yipenghe: Yeah that makes sense! I will change this before merging

PR #1739, opened 2019-11-25
Use placeholder for empty worksheet titles

### LINKED ISSUE #1721: [New Frontend] Add new worksheet, title field is empty and cannot edit, should be <none> instead
Discussed in meeting, when adding a new worksheet, title field is empty and cannot edit, should be \<none > instead

Adding the issue to better track it

yipenghe:
fix #1721 , test by adding a new worksheet, verify that the title shows <none> instead of not having the component

### DIFF
--- a/frontend/src/components/EditableField.js
+++ b/frontend/src/components/EditableField.js
@@ -111,7 +111,7 @@ class EditableFieldBase extends React.Component<{
                     onClick={this.onClick}
                     style={{ color: '#225ea8' }}
                 >
-                    {this.state.value === '' ? '<none>' : this.state.value}
+                    {!this.state.value ? '<none>' : this.state.value}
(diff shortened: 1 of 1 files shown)


frontend/src/components/EditableField.js
```diff
@@ -111,7 +111,7 @@ class EditableFieldBase extends React.Component<{
                     onClick={this.onClick}
                     style={{ color: '#225ea8' }}
                 >
-                    {this.state.value === '' ? '<none>' : this.state.value}
+                    {!this.state.value ? '<none>' : this.state.value}
```
epicfaace:
```suggestion
                    {this.state.value || '<none>'}
```

epicfaace (approved):

PR #1741, opened 2019-11-26
Avoid db call in deduct_worker_resources for every single running bundle

### LINKED ISSUE #1735: [Bundle Manager] Avoid querying the database for every single bundle when calculating available resources on each worker
This ticket is to avoid querying the database for every single bundle when calculating available resources on each worker. As we can see from the following code [_deduct_worker_resources()](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L328-L346), for each `uuid`, it checks for bundle's existence by querying the database, which is time-consuming. This ticket is to find an alternative way to improve the performance of the bundle manager overall.
```
    def _deduct_worker_resources(self, workers_list):
        """
        From each worker, subtract resources used by running bundles. Modifies the list.
        """
        for worker in workers_list:
                try:
                except NotFoundError:
                    logger.info(
                        'Bundle %s in WorkerInfoAccessor but no longer found. Skipping for resource deduction.',
                        uuid,
                    )
                    continue
                bundle_resources = self._compute_bundle_resources(bundle)
                worker['cpus'] -= bundle_resources.cpus
                worker['gpus'] -= bundle_resources.gpus
                worker['memory_bytes'] -= bundle_resources.memory
        return workers_list
```

candicegjing:
Fixed  #1735.

This PR is to reduce calling DB for every single running bundle when checking the current bundle's existence, so that it improves the overall performance of the bundle manager.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -312,6 +312,12 @@ class BundleManager(object):
             reverse=True,
         )
 
+        # Get bundles in RUNNING state from the bundle table
+        running_bundles = self._model.batch_get_bundles(state=State.RUNNING)
+        # Build a dictionary which maps from uuid to bundle
+        uuid_to_running_bundles = {bundle.uuid: bundle for bundle in running_bundles}
+
+        # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
             # Get user_owned workers.
             workers_list = workers.user_owned_workers(bundle.owner_id)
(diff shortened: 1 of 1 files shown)


codalab/server/bundle_manager.py
```diff
@@ -329,29 +335,29 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run):
...
         """
         From each worker, subtract resources used by running bundles. Modifies the list.
         """
+
         for worker in workers_list:
             for uuid in worker['run_uuids']:
-                try:
-                    bundle = self._model.get_bundle(uuid)
-                except NotFoundError:
+                # Verify if the current bundle exists in both Worker table and Bundle table
```
percyliang: fix grammar/terminology: "the `worker` table and the `bundle` table" (comment applies throughout)

percyliang (approved):
Looks good.  How much of an improvement do you get?

codalab/server/bundle_manager.py
candicegjing: This database call takes around 50% to 70% of the total time for each iteration. So this change should be able to help a bit when dispatching jobs.

PR #1743, opened 2019-11-26
Fix multiple deletion and deletion focus issues, pressing down on empty worksheet

### LINKED ISSUE #1701: [New Frontend] Bug in deleting markdown block

epicfaace:
Fixes #1701 by doing the following:
-   wait for a single deletion to complete before a keyboard press can trigger another deletion
-   keep focus the same upon deletion

Also fixes an issue in which pressing down on an empty worksheet crashes the page.


yipenghe (commented):
Overall looks good, one bug to fix: delete last element, the focus disappears and pressing up/down after that breaks

yipenghe (approved):
Stamping, just fix the empty worksheet problem

PR #1746, opened 2019-11-27
add mysqlclient dependency

yipenghe:


yipenghe:
Actually gonna just merge master to frontend

PR #1720, opened 2019-11-22
make upload-cancel-upload work

yipenghe:
fix #1710 , test with click upload -> click cancel/hit 'esc' ->click upload again, working

PR #1747, opened 2019-11-27
Handle File cancel properly

yipenghe:
The fix #1720 was not complete - it actually fails to upload things...
Found a solution here: https://stackoverflow.com/questions/49627109/reactjs-how-to-detect-when-cancel-is-clicked-on-file-input

Test:
1. upload-cancel-upload-cancel-upload sequence works fine, that is, the upload pops up every time
2. can actually upload things using the UI

In general the file input cancel event doesn't expose to browser, so needs the trick to properly handle this, hopefully material-ui will have their file upload component ready soon (it's in their road map)

### DIFF
--- a/frontend/src/components/worksheets/NewUpload/NewUpload.jsx
+++ b/frontend/src/components/worksheets/NewUpload/NewUpload.jsx
@@ -173,7 +173,7 @@ class NewUpload extends React.Component<{
     }
 
     // Handling cancel
-    // When the user clicks, it focuses back on the window and triggers onClose 
+    // When the user cancels, will focus back on the window and triggers onClose()
     handleFocusBack=()=>{
         this.props.onClose();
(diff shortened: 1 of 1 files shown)


teetone (approved):

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -171,6 +172,18 @@ class NewUpload extends React.Component<{
         this.setState({ percentComplete: 0, uploading: false });
     }
 
+    // Handling cancelling
+    // Based on: https://stackoverflow.com/questions/49627109/reactjs-how-to-detect-when-cancel-is-clicked-on-file-input
```
teetone: Nit: I'm not sure if we need to include the StackOverflow link here. I think just a brief description of the fix should suffice.

PR #1759, opened 2019-11-30
Add download to run bundles; show 'Contents' instead of 'Files' when bundle is not a directory; kill button only at proper state

### LINKED ISSUE #1762: Add download button to run bundles

### LINKED ISSUE #1763: Change files to contents if bundle is not a directory in bundle details

### LINKED ISSUE #1772: "Kill" button is shown even for bundles that have finished running
# Expected behavior
"Kill" button should be shown only for bundles that are preparing/running/etc. (i.e. not in the "ready state").

# Observed behavior
"Kill" button is shown for a bundle that has finished running and is in the "ready" state. Clicking on it doesn't do anything.

image

yipenghe:
fixes #1762, fixes #1763 , fixes #1772 
Test: 
1. can download a run bundle that has finished
2. can download a running bundle (while true; do date; done)

Screenshot from 2019-11-29 17-53-52

Contents:
Screenshot from 2019-11-30 11-13-55

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
@@ -51,6 +51,13 @@ class BundleActions extends React.Component<
 	            >
 	            	Kill
 	            </Button>
+				<Button
+					variant='contained'
+					color='primary'
+					onClick={ () => { window.location.href = bundleDownloadUrl; } }
+				>
+					Download
+				</Button>
 	            <Button variant='contained' color='primary'
 	            	onClick={ this.rerun }
 	            >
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
```diff
@@ -51,6 +51,13 @@ class BundleActions extends React.Component<
 	            >
 	            	Kill
 	            </Button>
+				<Button
```
yipenghe, line 61: Locally it looks nice...
Screenshot from 2019-11-30 10-07-52

epicfaace (commented):
One issue: the "Download" button still shows up when the bundle is in a "Preparing..." state. Then, clicking on the "Download" button replaces the screen with an error message such as the following:

image

epicfaace (approved):

Issue #1756, opened 2019-11-29
[New Frontend] Referencing a worksheet using bundle syntax (i.e. {worksheet_uuid}) causes page to crash


teetone:
**Describe the bug**
Referencing a worksheet using bundle syntax (i.e. {worksheet_uuid}) causes page to crash.

**To Reproduce**
Steps to reproduce the behavior :
1. Go to a worksheet.
2. Add the following line: {worksheet_uuid}
3. Page will try to load and then crash (see screenshot below).

See also the tiny worksheet to reproduce this issue: 
https://worksheets-dev.codalab.org/worksheets/0x3b4d0d28aa324f13b196d401482deaa8

**Expected behavior**
Page should not crash, we should just output an error.

**Screenshots**

![page_crashed](https://user-images.githubusercontent.com/16793796/69887379-7b889180-129b-11ea-9277-e7ceaef7af2f.png)

yipenghe:
This bug is not usual to trigger but very fatal once triggered, no way to access the worksheet after triggered. We should at least be able to access the source if we can get info back from the rest server

yipenghe:
Closing for now, opening another issue for refactoring that part of code

PR #1760, opened 2019-11-30
Prevent page crashing when referencing a worksheet uuid as bundle uuid

yipenghe:
Address #1756 

Not the cleanest, but it works, we should find ways to handle error better, easier to debug

When we pass in worksheet_uuid as a bundle uuid into get_bundle_infos, bundles (constructed on line 71) will filter out invalid ones, but the newly added get_host_worksheet_uuids function doesn't filter invalid ones out, so there will be a key error thrown on line 136 because bundle_infos doesn't contain the invalid worksheet uuid key, but host_worksheets does.

Line HERE 1: printed uuids before line 71
HERE 2: printed bundles right after line 71
HERE 3: printed (host_uuid, worksheet_names, bundle_uuid, bundles_info) 

Screenshot from 2019-11-30 01-08-30

Not entirely sure if 151-152 needs it too, but from the way host_worksheets is constructed, we should just add it to make sure

To test:
1. stop local service, checkout to `master`, restart, create a worksheet, edit source, add a line []{worksheet_uuid} (use any valid worksheet uuid), the page will crash and return worksheet not found
2. checkout to this branch, restart (it's a backend change so should restart backend to see difference), the page should show again

Ideally we should show it's tell not a valid bundle uuid, but not now

### DIFF
--- a/codalab/rest/util.py
+++ b/codalab/rest/util.py
@@ -129,6 +129,8 @@ def get_bundle_infos(
         worksheet_names = _get_readable_worksheet_names(model, worksheet_uuids)
 
         for bundle_uuid, host_uuids in host_worksheets.items():
+            if bundle_uuid not in bundle_infos:
+                continue
             for host_uuid in host_uuids:
                 if host_uuid in worksheet_names:
                     bundle_infos[bundle_uuid]['host_worksheet'] = {
--- a/codalab/rest/util.py
+++ b/codalab/rest/util.py
@@ -148,6 +148,8 @@ def get_bundle_infos(
 
         # Fill the info
         for bundle_uuid, host_uuids in host_worksheets.items():
+            if bundle_uuid not in bundle_infos:
+                continue
             bundle_infos[bundle_uuid]['host_worksheets'] = [
                 {'uuid': host_uuid, 'name': worksheet_names[host_uuid]}
                 for host_uuid in host_uuids


codalab/rest/util.py
```diff
@@ -129,6 +129,8 @@ def get_bundle_infos(
         worksheet_names = _get_readable_worksheet_names(model, worksheet_uuids)
 
         for bundle_uuid, host_uuids in host_worksheets.items():
+            if bundle_uuid not in bundle_infos:
```
teetone, line 132: I'm a little confused as to why `bundle_uuid` would not be in `bundle_infos`. Do you know when this is the case? We build `bundle_infos` from all the bundles we want to query and the keys of `host_worksheets` (on line 131) should only have bundle uuid's queried from the database itself (without any worksheet uuid's I believe).

teetone (approved):

codalab/rest/util.py
yipenghe: I think it's the query difference in get_host_worksheet_uuids and batch_get_bundles, if we query an invalid bundle uuid in batch_get_bundles, we get an empty list, and it returns here https://github.com/codalab/codalab-worksheets/blob/ada3fa9950136595a7226f2996d14c4d887c27f2/codalab/model/bundle_model.py#L685, maybe need to do some digging in the query

percyliang (commented):
Should `host_worksheets` be constructed to filter out invalid bundles in the first place so you don't need a check?

yipenghe:
> Should `host_worksheets` be constructed to filter out invalid bundles in the first place so you don't need a check?

I'm not familiar with how the queries work, maybe @teetone can do some more digging and open a PR to change that?

Issue #1684, opened 2019-11-18
keyboard shortcut for Rerun


percyliang:
Currently have to open the bundle details and click 'Rerun'.  Should be able to trigger this from the keyboard with 'a a' (in contrast to 'a r' for creating a run).

Furthermore, when the new run box pops up, the cursor should be at the end of the command-line, not the beginning.

The 'Rerun' button should perhaps say something like 'Edit and Rerun' to emphasize that you can make modifications.

Levilian:
- [x] Trigger Rerun from the keyboard with 'a a' (in contrast to 'a r' for creating a run).

- [x] When the new run box pops up, the cursor should be at the end of the command-line, not the beginning.

- [x] The 'Rerun' button should say something like 'Edit and Rerun' to emphasize that you can make modifications.

PR #1773, opened 2019-11-30
keyboard shortcut for rerun

Levilian:
-  Trigger Rerun from the keyboard with 'a a' (in contrast to 'a r' for creating a run).
-  When the new run box pops up, the cursor should be at the end of the command-line, not the beginning.
-  The 'Rerun' button should say something like 'Edit and Rerun' to emphasize that you can make modifications.

#1684

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
@@ -1,7 +1,6 @@
 // @flow
 import * as React from 'react';
 import Button from '@material-ui/core/Button';
-
 import { buildTerminalCommand } from '../../../util/worksheet_utils';
 import { executeCommand } from '../../../util/cli_utils';
 
@@ -38,6 +37,13 @@ class BundleActions extends React.Component<
 		});
 	}
 
+	componentDidUpdate = () => {
+		const { showNewRerun } = this.props;
+		if (showNewRerun) {
+			this.rerun();
+		}
+	}
+
 	render() {
 		const { bundleInfo } = this.props;
 		const bundleDownloadUrl = '/rest/bundles/' + bundleInfo.uuid + '/contents/blob/';
@@ -54,7 +60,7 @@ class BundleActions extends React.Component<
 	            <Button variant='contained' color='primary'
 	            	onClick={ this.rerun }
 	            >
-	            	Rerun
+	            	Edit and Rerun
 	            </Button>
 	        </div>
 	        : <Button
--- a/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
@@ -45,7 +45,7 @@ class BundleActions extends React.Component<
 	}
 
 	render() {
-		const { bundleInfo } = this.props;
+		const { bundleInfo, editPermission } = this.props;
 		const bundleDownloadUrl = '/rest/bundles/' + bundleInfo.uuid + '/contents/blob/';
 		const isRunBundle = bundleInfo.bundle_type === 'run';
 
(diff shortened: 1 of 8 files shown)


percyliang:
Thinking about it more, I wonder if 'a a' is not good because it might be triggered accidentally.  Maybe 'a u'?

Another thought: have 'a r' trigger new run when there are no bundles that are checked and rerun when there is a bundle that is checked.  But this requires checking rather than just selecting a bundle, which is an extra keypress and is not so natural when more than one bundle is checked...

Thoughts?

yipenghe:
> Thinking about it more, I wonder if 'a a' is not good because it might be triggered accidentally. Maybe 'a u'?
> 
> Another thought: have 'a r' trigger new run when there are no bundles that are checked and rerun when there is a bundle that is checked. But this requires checking rather than just selecting a bundle, which is an extra keypress and is not so natural when more than one bundle is checked...
> 
> Thoughts?

'a u' is used for upload already

I think triggering on the current focused bundle row (with detail opened) is better and more intuitive

yipenghe:
1. If you use 'a a' to trigger the rerun, and then cancel with 'esc'/ clicking with cancel, and open the detail again, it will stay in rerun. The cancel button/esc both can't get us back to the original bundle detail panel (no problem doing it with clicking the edit and rerun button) 

2. Currently clicking cance/esc also collapse the bundle detail panel, Ideally after clicking cancel, it should just go back to the original bundle detail state instead of collapsing, maybe just change it in your PR as well

3. after selecting a bundle, even if I click back on the bundle detail panel, can't trigger rerun with 'a a'

yipenghe (changes requested):
There are a few things that needs to be addressed

1. If you use 'a a' to trigger the rerun, and then cancel with 'esc'/ clicking with cancel, and open the same bundle detail again, it will stay in rerun. The cancel button/esc both can't get us back to the original bundle detail panel after that (no problem doing it with just clicking the edit and rerun button) 

2. Currently clicking cance/esc also collapse the bundle detail panel, Ideally after clicking cancel, it should just go back to the original bundle detail state instead of collapsing, maybe just change it in your PR as well

3. after selecting a bundle, even if I click back on the bundle detail panel, can't trigger rerun with 'a a'

percyliang:
Oh, right, forgot about upload...then maybe 'a n'?

Levilian:
> There are a few things that needs to be addressed
> 
> 1. If you use 'a a' to trigger the rerun, and then cancel with 'esc'/ clicking with cancel, and open the same bundle detail again, it will stay in rerun. The cancel button/esc both can't get us back to the original bundle detail panel after that (no problem doing it with just clicking the edit and rerun button)
> 2. Currently clicking cance/esc also collapse the bundle detail panel, Ideally after clicking cancel, it should just go back to the original bundle detail state instead of collapsing, maybe just change it in your PR as well
> 3. after selecting a bundle, even if I click back on the bundle detail panel, can't trigger rerun with 'a a'

For 3, we lock all keyboard shortcuts (including enter into detail view) as of now. Do you want to open them?

yipenghe:
> > There are a few things that needs to be addressed
> > 
> > 1. If you use 'a a' to trigger the rerun, and then cancel with 'esc'/ clicking with cancel, and open the same bundle detail again, it will stay in rerun. The cancel button/esc both can't get us back to the original bundle detail panel after that (no problem doing it with just clicking the edit and rerun button)
> > 2. Currently clicking cance/esc also collapse the bundle detail panel, Ideally after clicking cancel, it should just go back to the original bundle detail state instead of collapsing, maybe just change it in your PR as well
> > 3. after selecting a bundle, even if I click back on the bundle detail panel, can't trigger rerun with 'a a'
> 
> For 3, we lock all keyboard shortcuts (including enter into detail view) as of now. Do you want to open them?

Ah I see...let's leave as is then, I spotted a bug when we issue rerun while selecting a bundle row. After I fix that I will open up the rerun shortcut

yipenghe (approved):
Works great, can you also hide the rerun/kill button/disable the shortcut when the user doesn't have edit permission to the worksheet. See this commit https://github.com/codalab/codalab-worksheets/commit/f2fdd4b5ca1ec4f722742e62231b8fff410c11bf for reference.

Issue #1780, opened 2019-12-01
Non-editable worksheets should have edit buttons disabled


epicfaace:
On, non-editable worksheets we should disable the text, upload, run, edit/delete buttons for each worksheet item (to mirror how "Edit Source" becomes "View Source"). Otherwise, it is confusing because the user will think that they are able to edit a worksheet, when they really can't.

This is especially important to fix because it applies to the dashboard.

![image](https://user-images.githubusercontent.com/1689183/69918439-c682e000-1426-11ea-8606-2f2e582c902a.png)

PR #1783, opened 2019-12-01
[Frontend] Disable edit related bttons according to permission 

yipenghe:
If user doesn't have edit permission on worksheet, disable +text, upload, run, markdown item edit/delete buttons & shortcuts. 
Also don't show checkbox/disable 'x' when no edit worksheet permission. 
Disabling worksheet deletion as well.

<s>Bundle operations are trickier in this case, because not having permission to edit worksheet doesn't mean the user can't have permission to the bundles in the worksheet, thoughts? Currently the execution respects the bundle permission on backend side</s>

After thinking a bit, we should just respect 'view' mode should only view, not edit. We don't have bulk attach operation for now, so the selection is mostly useless to view mode users

#1780

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -1288,7 +1288,7 @@ class Worksheet extends React.Component {
                     size='small'
                     color='inherit'
                     aria-label='Delete Worksheet'
-                    disabled={!info}
+                    disabled={!editPermission}
                 >
                     <Tooltip
                         disableFocusListener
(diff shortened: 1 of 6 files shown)


epicfaace (commented):
Looks good.

Can you also disable (actually, probably just hide) the edit / delete buttons that show up when hovering over a worksheet item?

image

yipenghe:
> Looks good.
> 
> Can you also disable (actually, probably just hide) the edit / delete buttons that show up when hovering over a worksheet item?
> 
> image

I think I removed that, can you still see it after logging out?

epicfaace (approved):
Yeah, you're right, it's working.

PR #1784, opened 2019-12-01
Add permission check to /add-items

### LINKED ISSUE #1782: /add-items does not respect permissions
Any logged-in user can open the dashboard (which is read-only to them), and then add or delete items.

epicfaace:
Fixes #1782


yipenghe:
Does it just throw an error and not proceed if the worksheet doesn't have permission (outside currently, afk)

epicfaace:
> Does it just throw an error and not proceed if the worksheet doesn't have permission (outside currently, afk)

Yeah, it should do that. That's how the function is used in other API endpoint handlers.

yipenghe (approved):

PR #1787, opened 2019-12-01
Run tests in parallel

epicfaace:


percyliang (commented):
How is the output of all the tests collated?

epicfaace:
They weren't collated. I've just added a change that collates them, though.

epicfaace:
Doesn't seem like the tests are designed so they can work in parallel

percyliang:
What's the limitation?  You'd ideally want to be able to run in parallel.  We would just need different state for each one.

Issue #1791, opened 2019-12-02
Table header alignment for non-editable worksheet


yipenghe:
The header is slightly off when viewing a non-editable worksheet, the bug is introduced by #1783, need to adjust padding for non-editable worksheet

PR #1793, opened 2019-12-02
Fix table item padding

yipenghe:
#1791 Currently when you view a table of a non-editable worksheet, the header-value alignment is off, fixing that

To test: log in & log out, the table header-value alignment should both be nice


yipenghe:
@epicfaace @teetone Can you review when you have time?

teetone (approved):

PR #1798, opened 2019-12-02
Don't use subprocess for cli tests

epicfaace:
This will greatly speed up functionality tests (45 minutes -> 27 minutes)

### DIFF
--- a/test_cli.py
+++ b/test_cli.py
@@ -149,8 +149,8 @@ def run_command(
             # than opening a new subprocess to do so.
             stderr = io.StringIO()  # Not used; we just don't want to redirect stderr to f.
             f = FakeStdout()
+            cli = BundleCLI(CodaLabManager(), stdout=f, stderr=stderr)
             try:
-                cli = BundleCLI(CodaLabManager(), stdout=f, stderr=stderr)
                 cli.do_command(args[1:])
                 exitcode = 0
             except SystemExit as e:
--- a/test_cli.py
+++ b/test_cli.py
@@ -144,7 +144,7 @@ def run_command(
             kwargs = dict(kwargs, encoding="utf-8")
         if include_stderr:
             kwargs = dict(kwargs, stderr=subprocess.STDOUT)
-        if isinstance(args, list) and args[0] == "cl":
+        if isinstance(args, list) and args[0] == cl:
             # In this case, run the codalab CLI directly, which is much faster
             # than opening a new subprocess to do so.
             stderr = io.StringIO()  # Not used; we just don't want to redirect stderr to f.
--- a/test_cli.py
+++ b/test_cli.py
@@ -31,8 +31,6 @@ import sys
 import time
 import traceback
 from unittest.mock import patch
-from codalab.lib.bundle_cli import BundleCLI
-from codalab.lib.codalab_manager import CodaLabManager
 
 
 global cl
@@ -133,6 +131,12 @@ class FakeStdout(io.StringIO):
 def run_command(
     args, expected_exit_code=0, max_output_chars=1024, env=None, include_stderr=False, binary=False
 ):
+    # We import the following imports here because codalab_service.py imports TestModule from
+    # this file. If we kept the imports at the top, then anyone who ran codalab_service.py
+    # would also have to install all the dependencies that BundleCLI and CodaLabManager use.
+    from codalab.lib.bundle_cli import BundleCLI
+    from codalab.lib.codalab_manager import CodaLabManager
+
     """If we don't care about the exit code, set `expected_exit_code` to None.
     """
     print(">>", *map(str, args), sep=" ")
--- a/test_cli.py
+++ b/test_cli.py
@@ -125,7 +125,15 @@ class FakeStdout(io.StringIO):
         self.buffer = io.BytesIO()
 
     def getvalue(self):
-        return super().getvalue() + self.buffer.getvalue().decode()
+        """If self.buffer has a non-unicode value, return that value.
+        Otherwise, decode the self.buffer value and append it
+        to self.getvalue().
+        """
+        try:
+            buffer_value = self.buffer.getvalue().decode()
+        except UnicodeDecodeError:
+            return self.buffer.getvalue()
+        return super().getvalue() + buffer_value
 
 
 def run_command(
--- a/test_cli.py
+++ b/test_cli.py
@@ -177,9 +177,7 @@ def run_command(
                 exitcode = e.code
             output = stdout.getvalue()
         else:
-            output = subprocess.check_output(
-                [a.encode() for a in args], **kwargs
-            )
+            output = subprocess.check_output([a.encode() for a in args], **kwargs)
             exitcode = 0
     except subprocess.CalledProcessError as e:
         output = e.output
--- a/test_cli.py
+++ b/test_cli.py
@@ -125,9 +125,21 @@ class FakeStdout(io.StringIO):
         self.buffer = io.BytesIO()
 
     def getvalue(self):
-        """If self.buffer has a non-unicode value, return that value.
+        """
+        If self.buffer has a non-unicode value, return that value.
         Otherwise, decode the self.buffer value and append it
         to self.getvalue().
+
+        This is because this function is mimicking the behavior of `sys.stdout`.
+        `sys.stdout` can be read as either a string or bytes.
+
+        When a string is written to `sys.stdout`, it returns a string when doing `getvalue()`.
+
+        When bytes are written to `sys.stdout` (by writing to `sys.stdout.buffer`),
+        it returns bytes when doing `getvalue()`. The reason we need to account for this
+        case is that there are tests in which a binary file is uploaded, then it is
+        printed out (by writing to `sys.stdout.buffer`), and then the test reads what's
+        printed out and makes sure it matches the original file.
         """
         try:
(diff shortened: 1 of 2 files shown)


test_cli.py
```diff
@@ -126,10 +129,22 @@ def run_command(
...
+                    run_cli_command(args[1:])
+                    exitcode = 0
+                except SystemExit as e:
+                    exitcode = e
+                except:
+                    exitcode = 1
+            output = f.getvalue()
+        else:
+            output = subprocess.check_output(
+                [a.encode() if type(a) is str else a for a in args], **kwargs
```
percyliang: `isinstance(a, str)`

test_cli.py
```diff
@@ -126,10 +129,22 @@ def run_command(
             kwargs = dict(kwargs, encoding="utf-8")
         if include_stderr:
             kwargs = dict(kwargs, stderr=subprocess.STDOUT)
-        output = subprocess.check_output(
-            [a.encode() if type(a) is str else a for a in args], **kwargs
-        )
-        exitcode = 0
+        if type(args) is list and args[0] == "cl":
```
percyliang: `isinstance(args, list)`

percyliang: Add a general comment on why we're special casing this and the motivation.

epicfaace:
@percyliang ready for review now

test_cli.py
```diff
@@ -924,7 +975,9 @@ def test(ctx):
     # Modify to non-ascii tags
     # TODO: enable with Unicode support.
     non_ascii_tags = ['你好世界😊', 'fáncy ünicode']
-    run_command([cl, 'wedit', wname, '--tags'] + non_ascii_tags, 1)
+    run_command(
+        [cl, 'wedit', wname, '--tags'] + non_ascii_tags, 1, force_subprocess=True
+    )  # TODO: find a way to make this work without force_subprocess
```
epicfaace, line 990: This needs to eventually be fixed -- for some reason this command only succeeds when using `force_subprocess`.

test_cli.py
```diff
@@ -126,10 +162,25 @@ def run_command(
...
-            [a.encode() if type(a) is str else a for a in args], **kwargs
-        )
-        exitcode = 0
+        if not force_subprocess and isinstance(args, list) and args[0] == cl:
+            # In this case, run the codalab CLI directly, which is much faster
+            # than opening a new subprocess to do so.
+            # We skip doing this if force_subprocess is set to true (which forces
+            # us to use subprocess even for cl commands.)
+            _ = io.StringIO()  # Not used; we just don't want to redirect cli.stderr to f.
+            f = FakeStdout()
```
percyliang: Why don't we just name these `stdout` and `stderr` instead of `_` and `f`?

test_cli.py
```diff
@@ -126,10 +162,25 @@ def run_command(
             kwargs = dict(kwargs, encoding="utf-8")
         if include_stderr:
             kwargs = dict(kwargs, stderr=subprocess.STDOUT)
-        output = subprocess.check_output(
-            [a.encode() if type(a) is str else a for a in args], **kwargs
-        )
-        exitcode = 0
+        if not force_subprocess and isinstance(args, list) and args[0] == cl:
```
percyliang: why would `args` not be a list?

test_cli.py
```diff
@@ -126,10 +162,25 @@ def run_command(
...
+            cli = BundleCLI(CodaLabManager(), stdout=f, stderr=_)
+            try:
+                cli.do_command(args[1:])
+                exitcode = 0
+            except SystemExit as e:
+                exitcode = e.code
+            output = f.getvalue()
+        else:
+            output = subprocess.check_output(
+                [a.encode() if isinstance(a, str) else a for a in args], **kwargs
```
percyliang: When is `a` not a `str`?  The list should be all `bytes`, right?

test_cli.py
```diff
@@ -924,7 +975,9 @@ def test(ctx):
     # Modify to non-ascii tags
     # TODO: enable with Unicode support.
```
percyliang, line 986: Is this comment out of date?

test_cli.py
```diff
@@ -126,10 +162,25 @@ def run_command(
...
-        )
-        exitcode = 0
+        if not force_subprocess and isinstance(args, list) and args[0] == cl:
+            # In this case, run the codalab CLI directly, which is much faster
+            # than opening a new subprocess to do so.
+            # We skip doing this if force_subprocess is set to true (which forces
+            # us to use subprocess even for cl commands.)
+            _ = io.StringIO()  # Not used; we just don't want to redirect cli.stderr to f.
+            f = FakeStdout()
+            cli = BundleCLI(CodaLabManager(), stdout=f, stderr=_)
```
percyliang: Might not be great to have stderr just disappear?  Why not redirect to `f`?

test_cli.py
```diff
@@ -112,9 +114,43 @@ def sanitize(string, max_chars=256):
...
+    def getvalue(self):
+        """If self.buffer has a non-unicode value, return that value.
+        Otherwise, decode the self.buffer value and append it
+        to self.getvalue().
+        """
+        try:
+            buffer_value = self.buffer.getvalue().decode()
+        except UnicodeDecodeError:
+            return self.buffer.getvalue()
+        return super().getvalue() + buffer_value
```
percyliang, line 148: I'm confused...this function can either return `str` or `bytes`?  I don't understand how both can be fine...

epicfaace: This is because this function is mimicking the behavior of `sys.stdout`. `sys.stdout` can be read as either a string or bytes.

When a string is written to `sys.stdout`, it returns a string when doing `getvalue()`.

When bytes are written to `sys.stdout` (by writing to `sys.stdout.buffer`), it returns bytes when doing `getvalue()`. The reason I had to account for this case is that there are tests in which a binary file is uploaded, then it is printed out (by writing to `sys.stdout.buffer`), and then the test reads what's printed out and makes sure it matches the original file.

test_cli.py
epicfaace: Yeah, looks like `args` is always a list. Changing.

epicfaace:
```suggestion
        if not force_subprocess and args[0] == cl:
```

test_cli.py
```diff
@@ -126,10 +162,25 @@ def run_command(
...
+            # us to use subprocess even for cl commands.)
+            _ = io.StringIO()  # Not used; we just don't want to redirect cli.stderr to f.
+            f = FakeStdout()
+            cli = BundleCLI(CodaLabManager(), stdout=f, stderr=_)
+            try:
+                cli.do_command(args[1:])
+                exitcode = 0
+            except SystemExit as e:
+                exitcode = e.code
+            output = f.getvalue()
```
epicfaace:
```suggestion
            output = stdout.getvalue()
```

test_cli.py
epicfaace: Actually, that's how it worked before. Our call to `subprocess.check_output` discarded stderr, except for the case in which `include_stderr=True` is passed. That is passed only in the test module `gen-readthedocs` (https://github.com/codalab/codalab-worksheets/blob/4d0ef5b009dc4420559d6803b3b6ddd3a029d3b9/test_cli.py#L589) which will use the subprocess anyway.

If you redirect stderr to `f`, it fails several other tests because they currently don't expect to see stderr input suddenly coming into stdout.

epicfaace:
```suggestion
            cli = BundleCLI(CodaLabManager(), stdout=stdout, stderr=stderr)
```

test_cli.py
epicfaace: Done.

epicfaace:
```suggestion
            stdout = FakeStdout()
```

test_cli.py
```diff
@@ -126,10 +162,25 @@ def run_command(
...
-        output = subprocess.check_output(
-            [a.encode() if type(a) is str else a for a in args], **kwargs
-        )
-        exitcode = 0
+        if not force_subprocess and isinstance(args, list) and args[0] == cl:
+            # In this case, run the codalab CLI directly, which is much faster
+            # than opening a new subprocess to do so.
+            # We skip doing this if force_subprocess is set to true (which forces
+            # us to use subprocess even for cl commands.)
+            _ = io.StringIO()  # Not used; we just don't want to redirect cli.stderr to f.
```
epicfaace:
```suggestion
            stderr = io.StringIO()  # Not used; we just don't want to redirect cli.stderr to stdout.
```

test_cli.py
epicfaace: Changing.

epicfaace:
```suggestion
                [a.encode() for a in args], **kwargs
```

test_cli.py
epicfaace: I'm not sure. It appears that we still run `validate_ascii` on tag names (https://github.com/codalab/codalab-worksheets/blob/1c2b326cf42e42f5b98657050cddcef69ca357cb/codalab/rest/schemas.py#L118) so I don't recall us ever adding support for unicode tags. Need to investigate this further, but that can probably be done separate from this PR.

test_cli.py
percyliang: Can you add this to the comment?

percyliang (approved):
Great!

PR #1801, opened 2019-12-02
filter out unqualified bundles and update run_status to inform user i…

### LINKED ISSUE #1800: [Bundle Manager] Filter out unqualified staged bundles and updated their staged_status in UI
This ticket is to filter out unqualified staged bundles that are requested to run a personal worker which hasn't been discovered in the system, meaning either the worker is connected using a different name or it isn't connected to the system.

candicegjing:
Fixed #1800 
Added a new metadata field called `staged_status` (open to discuss of renaming or reusing `run_status`). 
This ticket will filter out staged bundles that are requested to run on the personal worker which hasn't been connected to CodaLab server yet.
 New frontend will look like this:
Screen Shot 2019-12-02 at 12 53 20 PM

### DIFF
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -44,6 +44,8 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('started', int, 'Time when this bundle started executing.', generated=True, formatting='date'))
     METADATA_SPECS.append(MetadataSpec('last_updated', int, 'Time when information about this bundle was last updated.', generated=True, formatting='date'))
     METADATA_SPECS.append(MetadataSpec('run_status', str, 'Execution status of the bundle.', generated=True))
+    METADATA_SPECS.append(MetadataSpec('staged_status', str, 'Information about the status of the staged bundle.', generated=True))
+
 
     # Information about running
     METADATA_SPECS.append(MetadataSpec('docker_image', str, 'Which docker image was used to run the process.', generated=True, hide_when_anonymous=True))
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -27,6 +27,9 @@ BUNDLE_TIMEOUT_DAYS = 60
 # When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
 # access to the host machine's memory, which we have decided to not allow
 MINIMUM_REQUEST_MEMORY_BYTES = 4 * 1024 * 1024
+# CodaLab public workers: a comma-separated list of worker ids to monitor.
+# Example: vm-clws-prod-worker-0,vm-clws-prod-worker-1
+CODALAB_PUBLIC_WORKERS = os.environ['CODALAB_PUBLIC_WORKERS'].split(',')
 
 
 class BundleManager(object):
@@ -50,6 +53,8 @@ class BundleManager(object):
 
         self._make_uuids_lock = threading.Lock()
         self._make_uuids = set()
+        # A set of UUID of bundles without their requested private worker discovered in the system
+        self._bundles_without_matched_workers = set()
 
         def parse(to_value, field):
             return to_value(config[field]) if field in config else None
@@ -327,7 +332,6 @@ class BundleManager(object):
         for bundle, bundle_resources in staged_bundles_to_run:
             # Get user_owned workers.
             workers_list = workers.user_owned_workers(bundle.owner_id)
-
             # If there is no user_owned worker, try to schedule the current bundle to run on a CodaLab's public worker.
             if len(workers_list) == 0:
                 # Check if there is enough parallel run quota left for this user
@@ -353,7 +357,6 @@ class BundleManager(object):
         """
         From each worker, subtract resources used by running bundles. Modifies the list.
         """
-
         for worker in workers_list:
             for uuid in worker['run_uuids']:
                 # Verify if the current bundle exists in both the worker table and the bundle table
@@ -369,8 +372,7 @@ class BundleManager(object):
                 worker['memory_bytes'] -= bundle_resources.memory
         return workers_list
 
-    @staticmethod
-    def _filter_and_sort_workers(workers_list, bundle, bundle_resources):
+    def _filter_and_sort_workers(self, workers_list, bundle, bundle_resources):
         """
         Filters the workers to those that can run the given bundle and returns
         the list sorted in order of preference for running the bundle.
@@ -384,12 +386,7 @@ class BundleManager(object):
         # Filter by tag.
         request_queue = bundle.metadata.request_queue
         if request_queue:
-            tagm = re.match('tag=(.+)', request_queue)
-            if tagm:
-                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
-            else:
-                # We don't know how to handle this type of request queue argument
-                return []
+            workers_list = self._get_matched_workers(request_queue, workers_list)
 
         # Filter by CPUs.
         workers_list = [
@@ -500,25 +497,27 @@ class BundleManager(object):
             return formatting.parse_size('2g')
         return formatting.parse_size(bundle.metadata.request_memory)
 
-    def _compute_request_disk(self, bundle):
+    def _compute_request_disk(self, bundle, user_info=None):
         """
         Compute the disk limit used for scheduling the run.
         The default is min(disk quota the user has left, global max)
         """
         if not bundle.metadata.request_disk:
             return min(
-                self._model.get_user_disk_quota_left(bundle.owner_id) - 1, self._max_request_disk
+                self._model.get_user_disk_quota_left(bundle.owner_id, user_info) - 1,
+                self._max_request_disk,
             )
         return formatting.parse_size(bundle.metadata.request_disk)
 
-    def _compute_request_time(self, bundle):
+    def _compute_request_time(self, bundle, user_info=None):
         """
         Compute the time limit used for scheduling the run.
         The default is min(time quota the user has left, global max)
         """
         if not bundle.metadata.request_time:
             return min(
-                self._model.get_user_time_quota_left(bundle.owner_id) - 1, self._max_request_time
+                self._model.get_user_time_quota_left(bundle.owner_id, user_info) - 1,
+                self._max_request_time,
             )
         return formatting.parse_duration(bundle.metadata.request_time)
 
@@ -560,16 +559,16 @@ class BundleManager(object):
         message['resources'] = bundle_resources.as_dict
         return message
 
-    def _compute_bundle_resources(self, bundle):
+    def _compute_bundle_resources(self, bundle, user_info=None):
         return RunResources(
             cpus=self._compute_request_cpus(bundle),
             gpus=self._compute_request_gpus(bundle),
             docker_image=self._get_docker_image(bundle),
             # _compute_request_time contains database queries that may reduce efficiency
-            time=self._compute_request_time(bundle),
+            time=self._compute_request_time(bundle, user_info),
             memory=self._compute_request_memory(bundle),
             # _compute_request_disk contains database queries that may reduce efficiency
-            disk=self._compute_request_disk(bundle),
+            disk=self._compute_request_disk(bundle, user_info),
             network=bundle.metadata.request_network,
         )
 
@@ -617,7 +616,7 @@ class BundleManager(object):
         self._restage_stuck_starting_bundles(workers)
         self._bring_offline_stuck_running_bundles(workers)
         self._acknowledge_recently_finished_bundles(workers)
-        staged_bundles_to_run = self._get_staged_bundles_to_run()
+        staged_bundles_to_run = self._get_staged_bundles_to_run(workers)
 
         # Schedule, preferring user-owned workers.
         self._schedule_run_bundles_on_workers(workers, staged_bundles_to_run)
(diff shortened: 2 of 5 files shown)


codalab/server/bundle_manager.py
```diff
@@ -692,6 +693,25 @@ def _get_staged_bundles_to_run(self):
                     bundle,
                     {'state': State.FAILED, 'metadata': {'failure_message': failure_message}},
                 )
+            elif bundle.metadata.request_queue:
+                # For those bundles that were requested to run on a private worker, if the private worker does not exist
+                # temporarily, we dispatch those bundles until the private worker connected to CodaLab server.
```
percyliang: Comment should make it clear that what we're doing here is simply updating the `staged_status` field, not actually carrying out any logic.

codalab/server/bundle_manager.py
```diff
@@ -692,6 +693,25 @@ def _get_staged_bundles_to_run(self):
...
+                request_queue = bundle.metadata.request_queue
+                if request_queue:
+                    tagm = re.match('tag=(.+)', request_queue)
```
percyliang: We should have a utility function for this (there must be other places where we have this logic).

percyliang: This could also return `None`, don't crash on that.

codalab/server/bundle_manager.py
```diff
@@ -692,6 +693,25 @@ def _get_staged_bundles_to_run(self):
...
+                if not ([worker for worker in workers.workers() if worker['tag'] == tagm.group(1)]):
```
percyliang: use `any(...)` and use more local variables for intermediate quantities to make it more readable.

codalab/server/bundle_manager.py
```diff
@@ -692,6 +693,25 @@ def _get_staged_bundles_to_run(self):
...
+                    # TODO: update bundles with batch operation: implement function batch_update_bundles()
```
percyliang, line 755: Not sure you can do that if the message is different for each bundle

codalab/server/bundle_manager.py
```diff
@@ -692,6 +693,25 @@ def _get_staged_bundles_to_run(self):
...
+                    self._model.update_bundle(
+                        bundle,
+                        {
+                            'metadata': {
+                                'staged_status': 'Bundle {} is requested to run on the private worker {} which has '
```
percyliang: The uuid does not need to be in the mesage

percyliang (changes requested):
Also, should modify the frontend so `staged_status` shows up there.

codalab/server/bundle_manager.py
candicegjing: Found some information online related to this: [link 1](https://stackoverflow.com/questions/54365873/sqlalchemy-update-multiple-rows-in-one-transaction), [link 2](https://docs.sqlalchemy.org/en/13/_modules/examples/performance/bulk_updates.html), [link 3](https://www.tutorialspoint.com/sqlalchemy/sqlalchemy_orm_updating_objects.htm)

codalab/bundles/run_bundle.py
```diff
@@ -44,6 +44,8 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('started', int, 'Time when this bundle started executing.', generated=True, formatting='date'))
     METADATA_SPECS.append(MetadataSpec('last_updated', int, 'Time when information about this bundle was last updated.', generated=True, formatting='date'))
     METADATA_SPECS.append(MetadataSpec('run_status', str, 'Execution status of the bundle.', generated=True))
+    METADATA_SPECS.append(MetadataSpec('staged_status', str, 'Detailed information about the status of the staged bundle.', generated=True))
```
percyliang: "Detailed information" => "Information" (simplify)

codalab/server/bundle_manager.py
```diff
@@ -378,11 +377,10 @@ def _filter_and_sort_workers(workers_list, bundle, bundle_resources):
...
         if request_queue:
-            tagm = re.match('tag=(.+)', request_queue)
-            if tagm:
-                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
-            else:
-                # We don't know how to handle this type of request queue argument
+            workers_list = self._get_matched_workers(request_queue, workers_list)
+            # Technically this shouldn't happen any more as all the bundles that are specified to run on
+            # invalid private workers are filtered out in _get_staged_bundles_to_run()
+            if not workers_list:
```
percyliang: I'd prefer to be more explicit (`workers_list is None` or `len(workers_list) == 0`)

codalab/server/bundle_manager.py
```diff
@@ -637,20 +635,33 @@ def _check_resource_failure(
                 return global_fail_string % (pretty_print(value), pretty_print(global_max))
         return None
 
-    def _get_staged_bundles_to_run(self):
+    def _get_matched_workers(self, request_queue, workers):
+        """
+        Get all of the workers that match with the name of the requested worker
+        :param request_queue: the requested worker
```
percyliang: Be more precise - `request_queue` can specify a tag that can match workers

codalab/server/bundle_manager.py
```diff
@@ -637,20 +635,33 @@ def _check_resource_failure(
...
+        :param workers: a list of workers
+        :return: a list of matched workers
+        """
+        tagm = re.match('tag=(.+)', request_queue)
+        matched_workers = [worker for worker in workers.workers() if worker['tag'] == tagm.group(1)]
```
percyliang: Add `worker_tag = tagm.group(1)`

codalab/server/bundle_manager.py
```diff
@@ -637,20 +635,33 @@ def _check_resource_failure(
...
+        if not matched_workers:
+            return []
```
percyliang: Seems like this is not necessary when `matched_workers = []` (again, be explicit about types)

codalab/server/bundle_manager.py
```diff
@@ -692,6 +703,24 @@ def _get_staged_bundles_to_run(self):
...
                     {'state': State.FAILED, 'metadata': {'failure_message': failure_message}},
                 )
+            elif bundle.metadata.request_queue:
+                # For those bundles that were requested to run on a private worker, if the private worker does not exist
+                # temporarily, we filter out those bundles so that they won't be dispatched to run on workers.
+                request_queue = bundle.metadata.request_queue
+                matched_workers = self._get_matched_workers(request_queue, workers)
+                if not matched_workers:
+                    # TODO: update bundles with batch operation: implement function batch_update_bundles()
+                    self._model.update_bundle(
```
percyliang, line 756: Seems like this is going to get executed a lot even though the `staged_status` doesn't change?  Not sure what to do about that though...

tests/server/bundle_manager_test.py
```diff
@@ -5,9 +5,24 @@
 from codalab.server.bundle_manager import BundleManager
 from codalab.worker.bundle_state import RunResources
 from codalab.bundles import RunBundle
+from codalab.lib.codalab_manager import CodaLabManager
 
 
 class BundleManagerTest(unittest.TestCase):
+    def setUp(self):
```
percyliang, line 12: What does this have to do with the main functionality of this PR?

percyliang (changes requested):

codalab/server/bundle_manager.py
candicegjing: That's a good point. We can cache those bundles. They shouldn't take too much space.

tests/server/bundle_manager_test.py
candicegjing: When I refactored the tag matching logic to `_get_matched_workers` as a separate function, I noticed that it can't be called from `_filter_and_sort_workers` as `_get_matched_workers` is a class function but `_filter_and_sort_workers` is declared as a static function. Then I searched through our codebase, I found out that the static function `_filter_and_sort_workers` is only used by `BundleManagerTest`. If this is only created for unit testing, I think it might not be necessary to keep only this function as a static function within the entire `BundleManager` class. Moreover, it makes more sense to make it as a class function instead of a static function. So I removed the `staticmethod` declaration for `_filter_and_sort_workers` and added missing settings for `BundleManagerTest`.

codalab/server/bundle_manager.py
candicegjing: This is necessary, as the function will return early so that it won't need to spend time going through the rest of the logic.

codalab/server/bundle_manager.py
```diff
@@ -45,6 +45,8 @@ def __init__(self, codalab_manager):
 
         self._make_uuids_lock = threading.Lock()
         self._make_uuids = set()
+        # A set of uuid of bundles whose requested private worker is not discovered in the system
+        self._bundles_wo_matched_workers = set()
```
percyliang: Write it out: "without"

codalab/server/bundle_manager.py
percyliang: The rest of the function is just `return matched_workers`, no?

codalab/server/bundle_manager.py
percyliang: I don't think you can cache because a bundle can eventually move out of this set.

percyliang (changes requested):

codalab/server/bundle_manager.py
candicegjing: Added logic to handle this.

codalab/server/bundle_manager.py
```diff
@@ -27,6 +27,8 @@
 # When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
 # access to the host machine's memory, which we have decided to not allow
 MINIMUM_REQUEST_MEMORY_BYTES = 4 * 1024 * 1024
+# CodaLab public workers regex on both dev and prod environment
+CODALAB_PUBLIC_WORKER_REGEX = re.compile(r"^vm-clws-(prod|dev)-(worker|gpuworker)-(\d)$")
```
percyliang: I really don't think we should have this hard-coded in here.

codalab/server/bundle_manager.py
percyliang: To clarify, call it `_bundles_without_matched_workers`

codalab/server/bundle_manager.py
```diff
@@ -386,11 +388,10 @@ def _filter_and_sort_workers(workers_list, bundle, bundle_resources):
...
         if request_queue:
-            tagm = re.match('tag=(.+)', request_queue)
-            if tagm:
-                workers_list = [worker for worker in workers_list if worker['tag'] == tagm.group(1)]
-            else:
-                # We don't know how to handle this type of request queue argument
+            workers_list = self._get_matched_workers(request_queue, workers_list)
+            # Technically this shouldn't happen any more as all the bundles that are specified to run on
+            # invalid private workers are filtered out in _get_staged_bundles_to_run()
+            if len(workers_list) == 0:
```
percyliang: This check shouldn't be needed, right?  I don't think there's any performance wins and I'd avoid having too many special cases.

codalab/server/bundle_manager.py
```diff
@@ -714,7 +739,27 @@ def _get_staged_bundles_to_run(self):
...
+                        {
+                            'metadata': {
+                                'staged_status': 'Bundle is requested to run on the private worker {} which has '
+                                'not been connected to the CodaLab server yet.'.format(
+                                    request_queue
+                                )
+                            }
+                        },
+                    )
+                    self._bundles_wo_matched_workers.add(bundle.uuid)
```
percyliang: This state is only for efficiency, right?  Did we check that it makes a difference in practice?  Again, try to get rid of state as much as possible.

codalab/server/bundle_manager.py
```diff
@@ -714,7 +739,27 @@ def _get_staged_bundles_to_run(self):
...
                     {'state': State.FAILED, 'metadata': {'failure_message': failure_message}},
                 )
+            elif bundle.metadata.request_queue and len(matched_workers) == 0:
+                # For those bundles that were requested to run on a private worker, if the private worker does not exist
+                # temporarily, we filter out those bundles so that they won't be dispatched to run on workers.
+                if bundle.uuid not in self._bundles_wo_matched_workers:
+                    # TODO: update bundles with batch operation: implement function batch_update_bundles()
+                    self._model.update_bundle(
+                        bundle,
+                        {
```
percyliang, line 758: Where do we actually unset `staged_status`?

percyliang (commented):
In general, how was this code tested?  Can you include some description of that in the PR?

codalab/server/bundle_manager.py
candicegjing: I agree that we shouldn't make special state as much as possible. I think what we are doing here is based on a point brought up earlier which is to avoid updating the database with the exact same information. In reality, the number of bundles that don't have private workers matched could vary from time to time. And this is something we can't control. By timing the `update_bundle` function, one query takes `0.0023462772369384766 seconds` when the database is in the same instance as rest-server/bundle-manager/worker. In production, this number will be larger for sure, as the database is hosted in a separate machine. Technically this will improve the performance for sure. But in terms of how much it will improve the performance overall will fully depend on how many bad bundles submitted to the system.

codalab/server/bundle_manager.py
candicegjing: Good point! Updated

candicegjing:
Rebasing this PR seems to be quite messy (Some changes in other PRs show up here). Closing this PR and creating another one.

PR #1788, opened 2019-12-01
Run performance tests

epicfaace:
Just to test this out

PR #1810, opened 2019-12-03
1347: Add script to stress test CodaLab backend

### LINKED ISSUE #1347: stress test script
Create a script that loops and hammers the server with large uploads, runs that eat up a lot of disk space, memory, time, etc.  Make sure the server can handle it, and try to break the worker system, make it run out of disk space.

teetone:
Resolves #1347 

Usage:
Screen Shot 2019-12-03 at 1 02 19 AM


epicfaace (approved):

epicfaace:
Can you make sure stress_test.py is copied to the docker container as in https://github.com/codalab/codalab-worksheets/pull/1788/files#diff-4a7fdf57d68d6fd4470c72984e1d15fe ? That way, it will be easy to run it on the dev server since it will already be there.

percyliang (commented):
Can you comment on what happens when you run the script on dev?  What does the output look like?

teetone:
> Can you make sure stress_test.py is copied to the docker container as in https://github.com/codalab/codalab-worksheets/pull/1788/files#diff-4a7fdf57d68d6fd4470c72984e1d15fe ? That way, it will be easy to run it on the dev server since it will already be there.

@epicfaace: I see what you mean now. It was a bit of a hassle to get everything set up. Thanks again for running the tests for me before. I will make the change to `Dockerfile.server` soon.

epicfaace (approved):

teetone:
> Can you comment on what happens when you run the script on dev? What does the output look like?

@percyliang : I still haven't had a successful run with full, heavy stress tests on dev. It either crashes dev or the heartbeats fail.

However, I did a handful of light and custom runs, which were successful. Here is a small subset of how the output will look like:
[light_output.log](https://github.com/codalab/codalab-worksheets/files/3936685/light_output.log)

stress_test.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, file_name, size_mb=1, content=None):
 
     def _make_random_file(self):
         with open(self._file_name, 'wb') as file:
-            file.seek(self._size_mb * 1024)  # seek takes in file size in terms of bytes
+            file.seek(self._size_mb * 1024000)  # seek takes in file size in terms of bytes
```
percyliang: Make this 1024 * 1024

stress_test.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def _start_heartbeat(self):
+        # Start heartbeats in the background. Each heartbeat creates a worksheet and prints its content.
+        t = Thread(target=self._heartbeat)
+        t.daemon = True
+        t.start()
+
+    def _heartbeat(self):
+        while True:
+            # Create a worksheet in a separate thread and check if it times out or not.
```
percyliang: If you just want a heartbeat, why not choose something that doesn't mutate?

percyliang (commented):
Where do you clean up any worksheet/bundle created by stress tests?  Should behave properly when there are failures...

stress_test.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def get_infinite_gpu_run_count(self):
+        return args.infinite_gpu_runs_count
+
+    def get_disk_write_count(self):
+        return args.large_disk_write_count
+
+
+def main():
+    if args.light:
```
percyliang: I think the default should be light, not heavy because the stress test could do damage.

stress_test.py
```diff
@@ -0,0 +1,490 @@
...
+        print('Created a LightStressTestRunner...')
+        runner = LightStressTestRunner(cl)
+    elif args.custom:
```
percyliang: These flags should override a light or heavy configuration, not be it's own thing.

stress_test.py
teetone: I replaced it with a read-only query instead.

teetone:
@percyliang  The clean up happens at the end of the stress testing. I currently keep track of all bundles and worksheets and just delete them at the end.

stress_test.py
```diff
@@ -0,0 +1,502 @@
...
+        return ''.join(
+            random.choice(string.ascii_lowercase + string.ascii_uppercase) for _ in range(24)
+        )
+
+    def _run_bundle(self, args):
+        uuid = run_command(args)
+        self._bundles.append(uuid)
+        return uuid
+
+    def _cleanup(self):
```
teetone: @percyliang I added the clean up logic here.

percyliang (changes requested):
1) Stress tests can most certainly fail and we might want to cancel things in the middle, so it's important to be able to be able to cleanup easily even if that happens, whereas right now, if the script dies halfway through, then we can't cleanup.  I'd suggest tagging any bundle or worksheet that's created by this script with something like `stress-test`, and then cleanup can be something like `cl search tag=stress-test .limit=500 | xargs cl`.  We should print out instructions on how to clean up.  Finally, we should cleanup after each test.

2) I still think the light/heavy config is too verbose and can be simplified.  First, I don't think there's any point in having 3 default configs (light, heavy, and command-line args).  Second, I don't think the configs should be child classes of `StressTestRunner` - they should be simply data.  So I'd just define the command-line arguments as you had them before with the light defaults.  If `--heavy` is set, then just simply set the relevant command-line arguments.  Then pass the `args` object into the `StressTestRunner`.

teetone:
> 1. Stress tests can most certainly fail and we might want to cancel things in the middle, so it's important to be able to be able to cleanup easily even if that happens, whereas right now, if the script dies halfway through, then we can't cleanup.  I'd suggest tagging any bundle or worksheet that's created by this script with something like `stress-test`, and then cleanup can be something like `cl search tag=stress-test .limit=500 | xargs cl`.  We should print out instructions on how to clean up.  Finally, we should cleanup after each test.
> 2. I still think the light/heavy config is too verbose and can be simplified.  First, I don't think there's any point in having 3 default configs (light, heavy, and command-line args).  Second, I don't think the configs should be child classes of `StressTestRunner` - they should be simply data.  So I'd just define the command-line arguments as you had them before with the light defaults.  If `--heavy` is set, then just simply set the relevant command-line arguments.  Then pass the `args` object into the `StressTestRunner`.

@percyliang  I am thinking of leaving the current clean up logic that will clean up after the stress tests on a happy path and also adding an additional parameter (something like `--clean-up-only`) that gives the option to just clean up the tagged bundles and worksheets.

One follow up question I had is do we have the capability to search for bundles by their tags i.e. `cl search tag=stress-test`?  `cl wsearch tag=stress-test` works, but I'm not sure if the `tag` option for bundle searching is currently supported or not.

percyliang:
The cleanup logic in both cases should exercise the same code path for consistency.

As for searching bundles by tags, it should work (if not, we should make it work).  Note that the search has a limit, so we have to put a loop around it.

teetone:
@percyliang It turns out when searching for bundles, the metadata field name has to be used, so I had to use `tags` for bundles and `tag` for worksheets.

I removed the child classes and cleaned up the configuration logic, so if `heavy` is set to true, I override the default, light configuration. I also added an additional option to just do the cleanup only, which uses the same code path as the post-stress tests cleanup.

stress_test.py
```diff
@@ -0,0 +1,444 @@
...
+            # Wait until the bundle finishes and then delete it
+            run_command([self._cl, 'wait', uuid], force_subprocess=True)
+            run_command([self._cl, 'rm', uuid, '--force'], force_subprocess=True)
+        for uuid in worksheets:
+            if len(uuid) == 0:
+                continue
+            run_command([self._cl, 'wrm', uuid, '--force'], force_subprocess=True)
+        print('Removed {} bundles and {} worksheets.'.format(len(bundles), len(worksheets)))
+
+    def _should_bypass_cleanup(self):
```
percyliang: Instead of these accessors, can't you just directly write `self._args.bypass_cleanup`?  Trying to minimize the amount of boilerplate...

teetone: I got rid of the accessors, but I wrapped the args around an enum so I have to access the values a bit differently using a dict.

teetone: If the preferred way is to use the namespace object directly, I can get rid of the enum as well.

stress_test.py
```diff
@@ -0,0 +1,413 @@
...
+- Run that writes infinitely to disk
+- Many runs that all use up plenty of disk (1GB each)
+- Run that uses infinite memory
+- Run that tries to use up all the GPUs (+ memory)
+- Many runs that use all sorts of large Docker images
+- Call CodaLab repeatedly to get information to check if the system is still responsive
+"""
+
+
+class StressTestArg(Enum):
```
percyliang: This is not necessary since `argparse` does it for you...you can `add_argument('--large-file-size-gb', ...)` and then access the arguments with `args.large_file_size_gb`.  This should simplify things quite a bit.

percyliang (changes requested):

stress_test.py
```diff
@@ -0,0 +1,390 @@
...
+    def _run_bundle(self, args):
+        args.append('--tags=%s' % StressTestRunner._TAG)
+        return run_command(args, force_subprocess=True)
+
+    def cleanup(self):
+        if self._args.bypass_cleanup:
+            return
+        print('Cleaning up...')
+        bundles = run_command(
+            [self._cl, 'search', 'tags=%s' % StressTestRunner._TAG, '.limit=10000', '--uuid-only'],
```
percyliang: 10000 is a lot, worried about it timing out.  I'd put this in a loop: I'd cap it at 1000 and put it in a loop.

stress_test.py
```diff
@@ -0,0 +1,390 @@
...
+        self._test_large_bundle()
+        self._test_many_bundle_uploads()
+        self._test_many_worksheet_copies()
+        self._test_parallel_runs()
+        self._test_many_docker_runs()
+        self._test_infinite_memory()
+        self._test_infinite_gpu()
+        self._test_infinite_disk()
+        self._test_many_disk_writes()
+        self.cleanup()
```
percyliang, line 118: We should call `cleanup` after each `test` to keep things under control.

stress_test.py
```diff
@@ -0,0 +1,390 @@
...
+
+    def cleanup(self):
+        if self._args.bypass_cleanup:
+            return
+        print('Cleaning up...')
+        bundles = run_command(
+            [self._cl, 'search', 'tags=%s' % StressTestRunner._TAG, '.limit=10000', '--uuid-only'],
+            force_subprocess=True,
+        ).split('\n')
+        worksheets = run_command(
```
percyliang: Might as well do it for the worksheets to be consistent.

percyliang (approved):
This is much better!  Can you also attach as a PR comment a transcript of this script running against some server?  It'd be nice to know what the expected output is and how longs take to run.

PR #1812, opened 2019-12-03
Do not fail worker when insufficient memory happened for starting a d…

### LINKED ISSUE #1811: Do not fail worker when there is no sufficient memory to start a docker container
This issue has happened several times in production. When there is not sufficient memory to start a docker container, the worker will throw an exception and go offline. This ticket is to change the behavior to keep the worker alive even if there is an exception thrown for a particular bundle so that the worker can still be used to run other jobs. 
An example of the error message from one production worker:
```
vm-clws-prod-gpuworker-5:~% docker logs codalab_worker_1 --tail 10
    runtime=self.docker_runtime,
  File "/opt/codalab/worker/docker_utils.py", line 33, in wrapper
codalab.worker.docker_utils.DockerException: Unable to start Docker container: Bundle 0xf97322adbb404e75801275d0e6cdf30b: minimum memory to start pytorch/pytorch:latest must be 4m (4194304 bytes), current available memory is 20 bytes
```

candicegjing:
Fixed #1811 
The final result would be look like in the following picture
Screen Shot 2019-12-03 at 1 43 29 PM

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -22,6 +22,11 @@ WORKER_TIMEOUT_SECONDS = 60
 SECONDS_PER_DAY = 60 * 60 * 24
 # Fail unresponsive bundles in uploading, staged and running state after this many days.
 BUNDLE_TIMEOUT_DAYS = 60
+# Impose a minimum container request memory 4mb (4 * 1024 * 1024 bytes), same as docker's minimum allowed value
+# https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory
+# When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
+# access to the host machine's memory, which we have decided to not allow
+MINIMUM_REQUEST_MEMORY_BYTES = 4 * 1024 * 1024
 
 
 class BundleManager(object):
@@ -51,6 +56,9 @@ class BundleManager(object):
 
         self._max_request_time = parse(formatting.parse_duration, 'max_request_time') or 0
         self._max_request_memory = parse(formatting.parse_size, 'max_request_memory') or 0
+        self._min_request_memory = (
+            parse(formatting.parse_size, 'min_request_memory') or MINIMUM_REQUEST_MEMORY_BYTES
+        )
         self._max_request_disk = parse(formatting.parse_size, 'max_request_disk') or 0
 
         self._default_cpu_image = config.get('default_cpu_image')
@@ -621,12 +629,14 @@ class BundleManager(object):
         global_fail_string=None,
         user_max=None,
         global_max=None,
+        global_min=None,
         pretty_print=lambda x: str(x),
     ):
         """
         Returns a failure message in case a certain resource limit is not respected.
         If value > user_max, user_fail_string is formatted with value and user_max in that order
-        If value > global_max, global_fail_strintg is formatted with value and global_max in that order
+        If value > global_max, global_fail_string is formatted with value and global_max in that order
+        If value < global_min, global_fail_string is formatted with value and global_min in that order
         Pretty print is applied to both the value and max values before they're passed on to the functions
         The strings should expect string inputs for formatting and pretty_print should convert values to strings
         """
@@ -635,6 +645,8 @@ class BundleManager(object):
                 return user_fail_string % (pretty_print(value), pretty_print(user_max))
             elif global_max and value > global_max:
                 return global_fail_string % (pretty_print(value), pretty_print(global_max))
+            elif global_min and value < global_min:
+                return global_fail_string % (pretty_print(value), pretty_print(global_min))
         return None
 
     def _get_staged_bundles_to_run(self):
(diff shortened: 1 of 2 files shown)


yipenghe:
We should get this in soon

percyliang (commented):
One concern I have is if that a worker is broken, then this change will end up failing a bunch of runs, which is worse than just essentially bringing the worker offline while we figure out what's wrong.

yipenghe:
> One concern I have is if that a worker is broken, then this change will end up failing a bunch of runs, which is worse than just essentially bringing the worker offline while we figure out what's wrong.

Then maybe we should just reject memory request less than, say, 10m? Not sure happens when memory is enough to run start container but not enough for running a run though

percyliang:
Ideally, we would target particular errors (like not enough memory)...not sure if we can read that out of the Exception.  Or maybe better, we should make sure we set a minimum memory before starting the container.

yipenghe:
> Ideally, we would target particular errors (like not enough memory)...not sure if we can read that out of the Exception. Or maybe better, we should make sure we set a minimum memory before starting the container.

We probably should either 1. replace insufficient memory request with minimum memory or 2. Reject the run immediately after reading the request

percyliang:
We should reject the request on the backend, because it seems like this only happens when someone deliberately does `--request-memory 20`.

yipenghe:
> We should reject the request on the backend, because it seems like this only happens when someone deliberately does `--request-memory 20`.

Might also be forgot to add the unit (b, m , g etc.), but I agree we should reject it on the backend. How much should the minimum memory be, maybe like ~100m?

percyliang:
Sure, `100m` sounds good.

candicegjing:
> One concern I have is if that a worker is broken, then this change will end up failing a bunch of runs, which is worse than just essentially bringing the worker offline while we figure out what's wrong.

Can you explain more on why this change would fail a bunch of runs when a worker is broken? I think this change will only affect the bundle which has insufficient memory requested.

candicegjing:
> Ideally, we would target particular errors (like not enough memory)...not sure if we can read that out of the Exception. Or maybe better, we should make sure we set a minimum memory before starting the container.

Yes. There is a piece of code that checks if the current requested memory is sufficient enough to start a container. The exception is thrown from that logic. Also, we can target that particular errors.

candicegjing:
Also failing the worker because of exception is thrown from one job doesn't seem to be a good idea. Right now the exception is due to insufficient memory, but that `try except` block will also catch other docker exceptions as well. Then the worker will stop accepting new jobs right away, which is not quite good.

percyliang:
If there is an error and it's the worker's fault, then we want the worker to fail so that we don't keep on failing all jobs that are assigned to that bad worker.  If it's the job's fault, we should fail the job.

candicegjing:
> If there is an error and it's the worker's fault, then we want the worker to fail so that we don't keep on failing all jobs that are assigned to that bad worker. If it's the job's fault, we should fail the job.

I think the confusion for this issue is that if the exception represents a broken worker or a bad job request. In this particular case, it's a bad job request (the error message might be reworded to *requested memory* instead of *available memory*), and the exception is from Docker not from the worker itself. I think rejecting the job when it requests insufficient memory is also helpful. We need to decide at which stage we want to reject it. For example, should we reject it during the time of request submission, reject it before assigning it to a worker or after assigning it to a worker? Also since there are other computing resources which will be requested from a user (CPUs, GPUs) and could potentially have invalid values like `0 CPU`, do we also consider adding a sanity check on those value and reject them early?

codalab/server/bundle_manager.py
```diff
@@ -22,6 +22,11 @@
 SECONDS_PER_DAY = 60 * 60 * 24
 # Fail unresponsive bundles in uploading, staged and running state after this many days.
 BUNDLE_TIMEOUT_DAYS = 60
+# Impose a minimum container request memory 4mb (4096 bytes), same as docker's minimum allowed value
+# https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory
+# When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
+# access to the host machine's memory, which we have decided to not allow
+MINIMUM_REQUEST_MEMORY = 4096
```
percyliang: Make type explicit: `MINIMUM_REQUEST_MEMORY_BYTES`

codalab/server/bundle_manager.py
```diff
@@ -22,6 +22,11 @@
 SECONDS_PER_DAY = 60 * 60 * 24
 # Fail unresponsive bundles in uploading, staged and running state after this many days.
 BUNDLE_TIMEOUT_DAYS = 60
+# Impose a minimum container request memory 4mb (4096 bytes), same as docker's minimum allowed value
```
percyliang: 4mb is not 4096 bytes

codalab/server/bundle_manager.py
percyliang: write `4 * 1024 * 1024`

percyliang (approved):

yipenghe:
Kind of confused, how did #2514 happen after this PR?

PR #1814, opened 2019-12-04
1813: get rid of the default for cl upload --ignore

### LINKED ISSUE #1813: [USER ISSUE] --exclude-ignore is only supported by GNU tar library
Mac users cannot run `cl upload [folder]` due to the fact that the default tar library is BSD. 

I think the best fix for now would be to get rid of the default value  of `--ignore` for `cl upload`. I will create a follow up issue to find a better to specify ignore files for BSD tar libraries.

teetone:
Fixes #1813 

Currently, `cl upload` can't be run on Macs since `--exclude-ignore` is only supported by GNU tar library and not BSD tar library. The quick and safest fix for now will be to just get rid of the default value of `--ignore` for `cl upload`.

I will create another issue to figure how we can specify ignore files for BSD tar library.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1111,7 +1111,7 @@ class BundleCLI(object):
                 '-i',
                 '--ignore',
                 help='Name of file containing patterns matching files and directories to exclude from upload. '
-                'This option is currently only supported with GNU tar library.',
+                'This option is currently only supported with the GNU tar library.',
             ),
         )
         + Commands.metadata_arguments([UploadedBundle])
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1110,8 +1110,8 @@ def do_workers_command(self, args):
             Commands.Argument(
                 '-i',
                 '--ignore',
-                help='Name of file containing patterns matching files and directories to exclude from upload.',
-                default='.gitignore',
+                help='Name of file containing patterns matching files and directories to exclude from upload. '
+                'This option is currently only supported with GNU tar library.',
```
percyliang: "with the GNU tar library"

percyliang (approved):

candicegjing (approved):
LGTM! Also please squash the commits before merging to master!

PR #1818, opened 2019-12-04
Heart check on workers

### LINKED ISSUE #1821: Monitor worker status
Currently, admins can't know if all workers are running in a healthy state. This ticket is to check if all workers are running online and notify admins when there are workers offline.

candicegjing:
Fixed #1821 
This PR is to add worker monitoring function so that admin will get notified when there is worker offline.


monitor.py
```diff
@@ -254,6 +255,24 @@ def check_disk_space(paths):
...
+    online_workers = set()
+    for line in workers_info:
+        online_workers.add(line.split()[0].strip())
+
+    workers_intersection = set(public_workers).intersection(online_workers)
+    offline_public_workers = public_workers - workers_intersection
+    if len(offline_public_workers) > 0:
+        error_logs(
+            'worker offline',
+            'The following public workers are offline: \n{}.'.format(
```
percyliang: remove space after `:`

monitor.py
```diff
@@ -74,6 +72,9 @@
 if not os.path.exists(args.backup_path):
     os.mkdir(args.backup_path)
 
+# List of public workers to monitor
+public_workers = os.environ['CODALAB_PUBLIC_WORKERS'].split(',')
```
percyliang, line 76: Be more explicit that these are worker ids?  But the worker ids are assumed to be hosts here and thus stable?  At some point they had hosts + a unique ID.  Make this explicit.

percyliang (changes requested):
Note that in the future when we use Azure Batch, we won't need this anymore...

monitor.py
candicegjing: Updated! The env variable `CODALAB_PUBLIC_WORKERS` will be a string containing workers_id `'vm-clws-prod-worker-0` separated by `,`.

monitor.py
```diff
@@ -74,6 +72,17 @@
 if not os.path.exists(args.backup_path):
     os.mkdir(args.backup_path)
 
+# List of public workers to monitor. For instance, in production environment, the list will include the following workers:
```
percyliang: Technically, this is a list, so you should write: "Comma-separated list of worker ids to monitor.  Example: vm-clws-prod-worker-0,vm-clws-prod-worker-1"

monitor.py
```diff
@@ -74,6 +72,17 @@
...
+# 1. 'vm-clws-prod-worker-0'
+# 2. 'vm-clws-prod-worker-1'
+# 3. 'vm-clws-prod-gpuworker-0'
+# 4. 'vm-clws-prod-gpuworker-1'
+# 5. 'vm-clws-prod-gpuworker-2'
+# 6. 'vm-clws-prod-gpuworker-3'
+# 7. 'vm-clws-prod-gpuworker-4'
+# 8. 'vm-clws-prod-gpuworker-5'
+public_workers = os.environ['CODALAB_PUBLIC_WORKERS'].split(',')
```
percyliang, line 76: Don't assume this environmental variable exists...

percyliang (changes requested):

percyliang (approved):

Issue #1820, opened 2019-12-04
When opening bundle details, the loading page is shown first


Levilian:
![emmgAgNkK6](https://user-images.githubusercontent.com/13169407/70185803-5f239500-169f-11ea-9e44-777deb3cd49e.gif)

Issue #1792, opened 2019-12-02
Better description when bundle detail is loading


yipenghe:
differentiate 'no bundle information'/'private bundle'/'loading bundle'

PR #1833, opened 2019-12-09
Remove detail loading message

yipenghe:
When the bundle detail is loading, don't show the 'not available message'
#1820 #1792


epicfaace (approved):
Looks good! Consider returning `null` instead of `<div></div>`, though it probably doesn't really matter.

Issue #1607, opened 2019-11-07
ctrl+enter does not work to stop editing when caps lock is mapped to ctrl


epicfaace:

yipenghe:
...too many issues this week, fixing it next time since this is not blocking..

PR #1835, opened 2019-12-09
Enable  caplocks+enter/s when caplocks is mapped as ctrl on Ubuntu

yipenghe:
#1607 
Replaced checking keycode with mousetrap mechanism


percyliang (approved):
Great, thanks!  Should we move ESC to Mousetrap to have consistent handling?

Also, I noticed that Ctrl+S doesn't work since it triggers Save...should we remove this?

yipenghe:
The '/' in the branch name causes error in the test, opened another PR to complete this#1871

PR #1837, opened 2019-12-09
Remove unreachable logic

### LINKED ISSUE #1836: [Bundle Manager]Remove unreachable logic.
This PR is to remove the following unreachable logic in bundle_manager.py:
https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L79

candicegjing:
Fixed #1836 .

This [logic]( https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L79-L80) seems to be not reachable during the iteration. The option is to either move it inside [while loop](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L71) or completely remove it. Based on the situation in production, I remove that piece of unreachable logic for now. Open to discuss this.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -76,9 +76,6 @@ class BundleManager(object):
 
             time.sleep(sleep_time)
 
-        while self._is_making_bundles():
-            time.sleep(sleep_time)
-
     def signal(self):
         with self._exiting_lock:
             self._exiting = True


percyliang (commented):
Doesn't this get executed when we exit?

candicegjing:
> Doesn't this get executed when we exit?

Yes, only when the code exit it will run. I guess I am quite confused about the reason we put this logic here. I originally thought this logic is to release some computing resources when we are making bundles... 
Do you mean this is a feature is specifically designed for the situation when we stop the bundle manager and want to let the making bundle finish?

percyliang:
Yeah, my guess is that if you Ctrl+C, you want to wait for making bundles to complete to avoid leaving things in a broken state.

PR #1588, opened 2019-11-01
Check for bundle existence before deducting resources

bkgoksel:
Fixes #1572

PR #1842, opened 2019-12-10
Do not fail jobs without enough resources

### LINKED ISSUE #1841: [USER ISSUE] bundle fail due to unavailable GPU
**Describe the bug**
When submitting multiple bundles where each bundle requests 1 GPU, only one bundle runs and the rest fails. It used to stay in the `staged` status but now directly fails. I could have manually submitted them one by one, but we're using the competition scripts that submit bundles in one batch.

**Codalab Information**
- *Your username*: hhe
- *UUID of worksheet where problem occurred*: https://worksheets.codalab.org/worksheets/0x1976df1c790b48c79eb8519c97d130dc
- *UUID of problem bundle (please don't delete)*: 0xcc97e8
- *Error message received from Codalab (if applicable)*: (Cannot assign enough resources: Requested more GPUs (1) than available (0 currently out of 1 on the machine))
 - *OS (if on CLI)*:
 - *Browser (if on Web)*;

**To Reproduce**

**Expected behavior**
It used to stay in the `staged` status which is fine, but now directly fails.

**Screenshots**

**Additional context**

candicegjing:
Fixed #1841 
I tested the functionality by setting the requested_cpu larger than the total available number of cpus. Bundle will stay in `PREPARING` state if not enough computing resources can be assigned with. `run_status` will be updated with a message similar as `Cannot assign enough resources: Requested more CPUs (5) than available (4 currently out of 4 on the machine)`
Not sure if we want to put the bundle back to `STAGED` state, I searched through our code base, didn't find a state that can be set back to `STAGED` in worker side. If I missed anything, @bkgoksel please do let me know.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -320,9 +320,13 @@ class BundleManager(object):
             reverse=True,
         )
 
-        # Get bundles in RUNNING state from the bundle table
-        running_bundles = self._model.batch_get_bundles(state=State.RUNNING)
-        # Build a dictionary which maps from uuid to bundle
+        # Get all the run_uuids from workers
+        run_uuids = []
+        for worker in workers.workers():
+            run_uuids.extend(worker["run_uuids"])
+        # Get the running bundles that exist in the bundle table
+        running_bundles = self._model.batch_get_bundles(uuid=run_uuids)
+        # Build a dictionary which maps from uuid to running bundle
         uuid_to_running_bundles = {bundle.uuid: bundle for bundle in running_bundles}
 
         # Dispatch bundles
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -270,7 +270,7 @@ class RunStateMachine(StateTransitioner):
             message = "Cannot assign enough resources: %s" % str(e)
             logger.error(message)
             logger.error(traceback.format_exc())
-            return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
+            return run_state._replace(run_status=message)
 
         # 4) Start container
         try:


percyliang:
If we don't have enough resources, we shouldn't even be sending it to the worker, no?

hhexiy:
Hi! Just want to ask about the progress on this fix. Currently we cannot submit multiple bundles in batch since only one GPU job can be run and the rest are failed.

hhexiy:
Hi, just want to ask about the progress on this fix. Currently we cannot submit multiple GPU jobs/bundles in one batch as only one can be run and the rest is failed.

candicegjing:
@hhexiy Sorry for the delay. I was off for the past few days. Will be working on this today and get it fixed asap.

hhexiy:
Thanks! I tried the PR, but seems it stills fails some jobs instead of sending them to the `preparing` queue.

candicegjing:
> jobs

Hmm, I tested locally with the previous change, it works for me. Did you attach your personal worker to run the tests?

codalab/server/bundle_manager.py
```diff
@@ -320,9 +320,11 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run):
             reverse=True,
         )
 
-        # Get bundles in RUNNING state from the bundle table
-        running_bundles = self._model.batch_get_bundles(state=State.RUNNING)
```
candicegjing, line 324: There is a race condition when fetching bundles in `state=running`. When a bundle moves from `staged` to `running` state, it took around 3 seconds to get that bundle back from `batch_get_bundles` function. If `deduct_bundle_resources` happened during this 3 seconds, the resource calculation will be wrong. Switching to fetching `run_uuids` will be more accurate.

codalab/worker/local_run/local_run_state.py
```diff
@@ -275,7 +275,7 @@ def _transition_from_PREPARING(self, run_state):
             message = "Cannot assign enough resources: %s" % str(e)
             logger.error(message)
             logger.error(traceback.format_exc())
-            return run_state._replace(stage=LocalRunStage.CLEANING_UP, failure_message=message)
+            return run_state._replace(run_status=message)
```
percyliang: Why this change?

codalab/server/bundle_manager.py
```diff
@@ -320,9 +320,11 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run):
...
-        # Build a dictionary which maps from uuid to bundle
+        # Get the running bundles that exist in the bundle table
+        running_bundles = []
+        for worker in workers.workers():
+            running_bundles.extend(self._model.batch_get_bundles(uuid=worker["run_uuids"]))
```
percyliang: This makes many more DB calls, which is not ideal.

percyliang (commented):
This wasn't a problem before, so what introduced this bug?

codalab/worker/local_run/local_run_state.py
candicegjing: This change will ensure jobs won't fail when there are not enough computing resources that could be assigned. This was changed two months ago from non-failing to failing state. But please let me know if we want to fail it on the worker side.

percyliang: This case shouldn't happen as the bundle manager really shouldn't be assigning jobs to workers that can't run them.  If we just kick it back, then won't we just end up in a loop?

candicegjing: Yes, I agree that we shouldn't assign jobs to a worker if there are not enough resources that can be allocated to run it. However, this piece of logic has been there for a long time and I guess it would be good to have a secondary check from the worker side to prevent any error from bundle manager so that we keep this logic here? 
Removing the `stage=LocalRunStage.CLEANING_UP` will let bundle stay in `preparing` stage. Once there are enough resources to run, the bundle will move to `running` stage gain.

percyliang: So job just stays in `preparing` on that worker?  Can the user kill the job in `preparing`?

candicegjing: User can definitely `kill` the job in `preparing` if they want. I think the key is to decide if a job should stay alive if there aren't enough computing resources that could be found on the worker side (when bundle manager didn't work as expected).  In this PR `run_status` will be updated with a message similar to `Cannot assign enough resources: Requested more GPUs (10) than available (0 currently out of 0 on the machine)` and the job will stay in `preparing` state.

candicegjing:
@percyliang The timeline related to this issue:
1. there was some logic added for checking the existence of bundle during resource deduction: [#1588 ](https://github.com/codalab/codalab-worksheets/commit/557a18b0cb5e8468ac619fbc96db838502394e28#diff-623d90f755b41b69c721e18042ffa858)
2. the above changes introduced performance overhead, then it changed to the new way of checking the existence of bundle: [#1741](https://github.com/codalab/codalab-worksheets/commit/20ec054f0e236b26bfb32ebd1bd822c2241d9c56#diff-623d90f755b41b69c721e18042ffa858)
3. during investigating this PR, there is a 3 seconds delay during the resource deduction when fetching bundles in `state=RUNNING` which may cause issues during resource calculation.

codalab/worker/local_run/local_run_state.py
percyliang: Ok, this sounds fine to me.

codalab/server/bundle_manager.py
percyliang: You mean that this single DB call took 3 seconds?  That seems like an incredibly long time...

But how does this make resource allocation more accurate?

percyliang: You mean that this single DB call took 3 seconds?  That seems like an incredibly long time...

But how does this make resource allocation more accurate?

candicegjing: I mean by the time we assign a bundle to run on a worker successfully in [Line 351](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L351), that bundle's `UUID` has been added to `workrer['run_uuid']`'s in-memory object in [Line 449](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L449). Then in the next round of iteration in [Line 324](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L324), the fetched results from the database don't have that bundle reflected in `running` state yet. Since that bundle has already been added to `worker['run_uuid']` and we don't get its information from the bundle table, when we do resource deduction in [Line 362](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L362), it will just be skipped, and the returned `workers_list` in [Line 372](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L372) will eventually end up having more resources left once the calculation is done. Then those workers will be assigned with more jobs than their real capacity eventually.

candicegjing: With the new SQL statement, it won't look for bundles in `running` state anymore. Instead, it will only search for bundles that exist in `run_uuid` and ignore the state condition check, which increases the chance to be able to fetch the actual bundle back. Then during the resource deduction, skipping checks for one bundle in [Line 362](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L362) won't happen frequently and hence increase the resource calculation accuracy.

percyliang (approved):

PR #1844, opened 2019-12-10
Remove unused container_start_time

### LINKED ISSUE #1843: Remove unused variable container_start_time
Since we move to the new logic to calculate time_used per job, container_start_time is no longer necessary. Removing it from code base.

candicegjing:
Fixed #1843
This is to remove `container_start_time`, as we have already moved to the new logic of calculating `time_used` per job. It's not necessary to have any more.


codalab/worker/local_run/local_run_state.py
```diff
@@ -75,7 +75,6 @@ class LocalRunStage(object):
         'bundle_dir_wait_num_tries',  # Optional[int]
         'resources',  # RunResources
         'bundle_start_time',  # int
```
percyliang, line 77: What's the difference between `bundle_start_time` and `container_start_time`?

percyliang (approved):

codalab/worker/local_run/local_run_state.py
candicegjing: After scanning the code, `bundle_start_time` was set at `create_run` function which is used to process a new run with the given bundle and resources. However, this variable seems to be not used in any other calculations. If this is not necessary, I'll remove it in a different ticket.

PR #1847, opened 2019-12-11
Keyboard shortcut improvements

Levilian:
- g g and shift+g can work under all conditions
- 'a s' is the shortcut for downloading a bundle
- disable shortcuts that do not apply to certain cells/rows


Levilian:
fix: #1846, #1764, #1765

PR #1852, opened 2019-12-12
Fix parallel run quota condition check

### LINKED ISSUE #1828: investigating parallel_run_quota
On prod, one user is running too many jobs which doesn't seem to be correct. This ticket is to investigate if `parallel_run_quota` is enforced for all users.
Screen Shot 2019-12-04 at 9 14 26 PM

candicegjing:
Fixed #1828 
This PR is to fix the problem of parallel run quota on Prod. 
When `self._model.get_user_parallel_run_quota_left(bundle.owner_id)` returns `negative` value, statement of `not <negative>` within `if` clause won't work. This is to make the logic more precise. 
**The solution might not be optimal. But it worked.**

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -331,9 +331,9 @@ class BundleManager(object):
             workers_list = workers.user_owned_workers(bundle.owner_id)
 
             # If there is no user_owned worker, try to schedule the current bundle to run on a CodaLab's public worker.
-            if not workers_list:
+            if len(workers_list) == 0:
                 # Check if there is enough parallel run quota left for this user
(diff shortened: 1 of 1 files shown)


candicegjing:
Another instance of this issue:
Screen Shot 2019-12-12 at 11 21 24 AM

candicegjing:
@bkgoksel Do you remember at which condition this issue happened? [bundle exists in worker_run table but not in bundle table](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L362-L366)

percyliang (approved):
Great.  Can we make sure we the bug where the function is returning negative values?

PR #1857, opened 2019-12-17
Fix failures due to duplicate container names

### LINKED ISSUE #1704: Prod worker can not be started due to "exec: \"bash\": executable file not found in $PATH": unknown"
```

  File "/opt/codalab/worker/worker.py", line 56, in start
    runtime=self.docker_runtime,
  File "/opt/codalab/worker/docker_utils.py", line 34, in wrapper
codalab.worker.docker_utils.DockerException: Unable to start Docker container: 409 Client Error: Conflict ("Conflict. The container name "/codalab_run_0xc2559ab6e42f446b9beea50c74c1222f" is already in use by container "3d823a5f30232fb833b193da2d8839d380ac8c16114b60f0cc8f007caf4d10f7". You have to remove (or rename) that container to be able to reuse that name.")
2019-11-21 02:20:18,429 Sleeping for 1 hour due to exception...please help me!
```
This issue might need some investigation as I saw it several times in production.

candicegjing:
Fixed #1704.

There are two temporary options to fix this problem when we discovered that there is an existing container with the same container name as the new one:
1. Delete the old container and return the new container.
2. Return the old container without trying to starting a new container.

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -144,7 +144,6 @@ def start_bundle_container(
     container_name = 'codalab_run_%s' % uuid
 
     if container_exists(container_name):
-        client.containers.remove(container_name)
         return client.get(container_name)
 
     container = client.containers.run(
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -142,6 +142,11 @@ def start_bundle_container(
 
     # Name the container with the UUID for readability
     container_name = 'codalab_run_%s' % uuid
+
+    if container_exists(container_name):
+        client.containers.remove(container_name)
+        return client.get(container_name)
+
     container = client.containers.run(
         image=docker_image,
         command=docker_command,


codalab/worker/docker_utils.py
```diff
@@ -142,6 +142,11 @@ def start_bundle_container(
 
     # Name the container with the UUID for readability
     container_name = 'codalab_run_%s' % uuid
+
+    if container_exists(container_name):
+        client.containers.remove(container_name)
```
percyliang: What happens if this fails?  This really shouldn't happen, so if it does, something might be wrong anyway, and the container might be actually running...  We should at least send us a notification when this happens.

percyliang (commented):
This duplicate container really shouldn't happen, so I think we should get to the root of the problem.

codalab/worker/docker_utils.py
candicegjing: This error was thrown from [_transition_from_PREPARING](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/local_run/local_run_state.py#L153). I searched through the entire worker code base, didn't find any trace indicating that it's trying to restart the same container twice. Maybe @bkgoksel would have some suggestions?

bkgoksel:
I know this PR is quite old, but can you share some more details about this issue? How often is it happening? Do you see the "Started container for X" log message on the logs and if so is there anything on the logs between the first time the container is started and the second time it is attempted. Is it possible that the worker might've been restarted in between the two instances.

candicegjing:
This PR is invalid as the root cause will be fixed by #1917

PR #1860, opened 2019-12-18
Rename WorkerRun to BundleCheckinState

bkgoksel:
Better represents what it is used for (part of my broader worker simplification work)

### DIFF
--- a/codalab/worker/bundle_state.py
+++ b/codalab/worker/bundle_state.py
@@ -155,7 +155,8 @@ class RunResources(object):
 
 class BundleCheckinState(object):
     """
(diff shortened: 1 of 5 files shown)


codalab/worker/bundle_state.py
```diff
@@ -153,7 +153,7 @@ def from_dict(cls, dct):
         )
 
 
-class WorkerRun(object):
+class BundleCheckinState(object):
     """
     Defines all the field the worker needs to check in with the server for its runs
```
percyliang: Update comment (seems like this should be about one bundle).

percyliang (approved):

PR #1861, opened 2019-12-18
Rename to_dict to as_dict and make it a property for worker state objects

bkgoksel:
It makes more sense to me as properties, just came up while I was working on simplifying worker.


percyliang (approved):
Is there any benefit to doing it this way aside from aesthetics?

bkgoksel:
No. Purely aesthetics. My thinking was that it doesn't change any state and is purely a different deterministic view on the same object. But it's completely optional and isn't necessary for any of the rest of the changes.

PR #1865, opened 2019-12-18
No more run manager

bkgoksel:
Fuses RunManager and Worker classes.

PR #1866, opened 2019-12-19
Threaded netcat write

bkgoksel:
These operations may take a long time and we really don't want to
block on them. On the flipside we don't have to manage these threads
carefully since it's OK if they fail due to larger run state changes.

Depends on #1865


percyliang (approved):
Cool!  Could you add a comment in the code about not caring too much about larger run_state changes?

PR #1867, opened 2019-12-19
No more worker lock

bkgoksel:
With the simplified single-file worker, I realized that the lock in the worker isn't really necessary as there's always only one thread manipulating the run states directly now.


codalab/worker/worker.py
```diff
@@ -257,9 +255,11 @@ def checkin(self):
                     self.read_run_missing(socket_id)
                 return
             if action_type == 'kill':
-                self.kill(uuid)
+                self.runs[uuid] = self.runs[uuid]._replace(
```
percyliang: Perhaps better to leave the functions `kill` and `mark_finalized` to parallel the other `action_type`s, which have their own functions.

percyliang (approved):

PR #1868, opened 2019-12-20
Worker cleanup

bkgoksel:
Nits cleaned around worker
Depends on #1865


bkgoksel:
Rebased and ready for a review @percyliang mostly small changes.

codalab/worker/main.py
```diff
@@ -45,15 +45,15 @@ def parse_args():
     )
     parser.add_argument(
         '--cpuset',
-        type=str,
+        type=cpuset_args,
```
bkgoksel: This automates the logic of passing the string form of the arg to the `type` function, and sets the arg value as the return value of that function. (in this case our custom parsing function)

codalab/worker/main.py
```diff
@@ -147,44 +148,48 @@ def connect_to_codalab_server(server, password_file):
...
+            os.path.join(args.work_dir, 'dependencies-state.json'),
+            bundle_service,
+            args.work_dir,
+            args.max_work_dir_size,
+        )
+    # Set up local directories
     if not os.path.exists(args.work_dir):
         logging.debug('Work dir %s doesn\'t exist, creating.', args.work_dir)
         os.makedirs(args.work_dir, 0o770)
+    if local_bundles_dir and not os.path.exists(local_bundles_dir):
```
bkgoksel, line 184: Create both local dirs here for consistency (before, the latter was created within the constructor of the Worker class)

codalab/worker/main.py
```diff
@@ -147,44 +148,48 @@ def connect_to_codalab_server(server, password_file):
...
-    )
-
+    if args.shared_file_system:
+        # No need to store bundles locally if filesystems are shared
+        local_bundles_dir = None
+        # Also no need to download dependencies if they're on the filesystem already
+        dependency_manager = None
+    else:
+        local_bundles_dir = os.path.join(args.work_dir, 'runs')
+        dependency_manager = DependencyManager(
```
bkgoksel, line 174: Don't create a dependency manager if one is not needed.

codalab/worker/main.py
```diff
@@ -219,18 +224,18 @@ def parse_cpuset_args(arg):
         try:
             cpuset = [int(s) for s in arg.split(',')]
         except ValueError:
-            raise ValueError(
+            raise argparse.ArgumentTypeError(
```
bkgoksel, line 239: This is the correct error to raise when you can't parse your argument type.

codalab/worker/worker.py
```diff
@@ -38,55 +38,52 @@ class Worker:
     NETCAT_BUFFER_SIZE = 4096
     # Number of seconds to wait for bundle kills to propagate before forcing kill
     KILL_TIMEOUT = 100
-    # Directory name to store running bundles in worker filesystem
```
bkgoksel, line 41: I omitted this constant now because it was added here after this PR was opened. We can re-add it as a constant to the main file, but it's a lot less jarring there as a hard-coded value than here.

codalab/worker/worker.py
```diff
@@ -38,55 +38,52 @@ class Worker:
...
         self.tag = tag
+
         self.work_dir = work_dir
-        self.bundle_service = bundle_service
+        self.local_bundles_dir = local_bundles_dir
+        self.shared_file_system = shared_file_system
+
         self.exit_when_idle = exit_when_idle
         self.idle_seconds = idle_seconds
+
```
bkgoksel, line 81: Just re-ordered the arg initialization for more logical consistency.

codalab/worker/worker.py
```diff
@@ -147,37 +144,32 @@ def load_state(self):
...
             )
 
+    @property
+    def should_stop_because_idle(self):
+        now = time.time()
+        if len(self.runs) > 0 or self.last_time_ran is None:
+            self.last_time_ran = now
+        is_idle = now - self.last_time_ran > self.idle_seconds
+        return self.exit_when_idle and is_idle and self.last_checkin_successful
+
```
bkgoksel: Separated this logic into its own function.

codalab/worker/main.py
```diff
@@ -205,7 +210,7 @@ def main():
     worker.start()
 
 
-def parse_cpuset_args(arg):
+def cpuset_args(arg):
```
percyliang: Can we keep the `parse_cpuset_args` name to be consistent with the `parse_size` functions?

codalab/worker/worker.py
```diff
@@ -147,37 +144,32 @@ def load_state(self):
                 resources=RunResources.from_dict(run_state.resources),
             )
 
+    @property
+    def should_stop_because_idle(self):
+        now = time.time()
+        if len(self.runs) > 0 or self.last_time_ran is None:
+            self.last_time_ran = now
```
percyliang, line 158: Not great that the property mutates...

codalab/worker/worker.py
```diff
@@ -147,37 +144,32 @@ def load_state(self):
                 resources=RunResources.from_dict(run_state.resources),
             )
 
+    @property
+    def should_stop_because_idle(self):
+        now = time.time()
```
percyliang, line 156: Add docstring?

percyliang (approved):

codalab/worker/worker.py
bkgoksel: Right. I'll rename + make it not a property and add a docstring.

bkgoksel:
Are the tests currently flaky? I only added a docstring, renamed a couple methods and removed the @ property decoration from one method and now one of the tests is failing with some error that's happening during the cleanups while the other runs fine. Same thing happened in my other PR after I rebased on master.

PR #1871, opened 2019-12-21
Enable caplocks+enter/s when caplocks is mapped as ctrl on Ubuntu

yipenghe:
same as #1835, but with a different branch name. The '/' in the original branch name seems to cause a problem when pulling latest codebase (always error around ~43s), changed a different branch name and it passes

Replaced the ctrl+enter and esc with mousetrap mechanism

> #1607 
> Replaced checking keycode with mousetrap mechanism


yipenghe:
merging as the original PR was approved

PR #1877, opened 2019-12-28
Set CODALAB_PUBLIC_WORKERS in codalab_service.py

candicegjing:
This PR will set the value of `CODALAB_PUBLIC_WORKERS` in Docker container from `codalab_service.py` during deployment. 
Note that **master** has more commits than the **staging** branch. I'll sync them once the PR is approved to make sure those changes are in the right order.

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -266,6 +266,8 @@ CODALAB_ARGUMENTS = [
         default=2048,
         help='Default memory (in MB) for each worker in the AWS Batch Worker Manager',
     ),
+    ### Public workers
+    CodalabArg(name='public_workers', help='Comma-separated list of worker ids to monitor'),
 ]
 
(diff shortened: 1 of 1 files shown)


codalab_service.py
```diff
@@ -696,6 +698,7 @@ def start_services(self):
                         'server/default_user_info/parallel_run_quota',
                         self.args.user_parallel_run_quota,
                     ),
+                    ('server/public_workers', self.args.public_workers),
```
percyliang: This shouldn't be necessary, right?  We just need to pass `CODALAB_PUBLIC_WORKERS` through, which defining a `CodalabArg` already does.

candicegjing: Yes. That's true. I removed this line.

percyliang (approved):

Issue #1853, opened 2019-12-13
Bundle failing with "Error while uploading: 'str' object has no attribute 'decode'"


teetone:
There are a few bundles that fail with "Error while uploading: 'str' object has no attribute 'decode'", which is an error from CodaLab:

https://worksheets.codalab.org/bundles/0x7eefa36a6c5747538ce74bb5e9f8c055
https://worksheets.codalab.org/bundles/0xa2f0624ad5a0432e87578117d62c90c1

We need to investigate why some bundles fail with this error. Does rerunning the bundle resolve this error?

Issue #1829, opened 2019-12-05
Prod CPU worker 0 went offline due to HTTP Error 400: Bad Request in _upload_with_chunked_encoding


candicegjing:
prodcpu0 went offline with the following error:
```
2019-12-05 05:21:27,969 http://localhost:None "GET /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8/json HTTP/1.1" 200 None
2019-12-05 05:21:27,981 http://localhost:None "GET /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8/json HTTP/1.1" 200 None
2019-12-05 05:21:27,982 Finished run with UUID 0x030b0d84f576464d9958e1aafb947c88, exitcode 0, failure_message None
2019-12-05 05:21:33,231 http://localhost:None "GET /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8/json HTTP/1.1" 200 None
2019-12-05 05:21:33,254 http://localhost:None "GET /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8/json HTTP/1.1" 200 None
2019-12-05 05:21:33,261 http://localhost:None "GET /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8/logs?stderr=1&stdout=0&timestamps=0&follow=0&tail=all HTTP/1.1" 200 None
2019-12-05 05:21:33,278 http://localhost:None "GET /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8/json HTTP/1.1" 200 None
2019-12-05 05:21:33,304 http://localhost:None "DELETE /v1.35/containers/7b8dd192457b3711b3ea3ae15c1e8ff15b47ecb15a389a3c070f5dfcf8b212c8?v=False&link=False&force=True HTTP/1.1" 204 0
2019-12-05 05:21:36,673 Uploading results for run with UUID 0x030b0d84f576464d9958e1aafb947c88
[...]

PR #1884, opened 2020-01-06
fix-stucked-bundle-in-uploading-state

### LINKED ISSUE #1875: Jobs stuck at uploading stage when user doesn't have enough disk space left
Screenshot from 2019-12-25 06-48-37

Jobs can't end when user don't have enough disk space left, the worker should fail the job

candicegjing:
Fixed #1875 and #1829 
During investigating on #1875, there are two issues were discovered:
1. `AttributeError: 'str' object has no attribute 'decode'`: might be related to #1853 
```
  File "/usr/local/bin/cl", line 11, in <module>
    load_entry_point('codalab', 'console_scripts', 'cl')()
  File "/opt/codalab-worksheets/codalab/bin/cl.py", line 10, in main
    cli.do_command(sys.argv[1:])
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 839, in do_command
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 833, in <lambda>
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 1229, in do_upload_command
    progress_callback=progress.update,
  File "/opt/codalab-worksheets/codalab/client/json_api_client.py", line 18, in wrapper
    error_body = e.read().decode()
AttributeError: 'str' object has no attribute 'decode'
```
2. Bundle failed to move to `failed` state for a very long time. It kept moving from `failed` --> `running` --> `failed` and caused the situation described in #1875. This is due to a very large retry number defined at [COMMAND_RETRY_SECONDS](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L35). When the current bundle failed from uploading, it keeps retrying until the number of retries becomes a negative value, see logic here [execute_bundle_service_command_with_retry](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L530). Hence, the bundle got stuck in uploading state will eventually move to a permanent `failed` state. However, this will take a very long time. 
**We need to decide how many times of retries should be used for function** `execute_bundle_service_command_with_retry` **as the existing value seems to be quite big.**

### DIFF
--- a/codalab/worker/bundle_service_client.py
+++ b/codalab/worker/bundle_service_client.py
@@ -22,27 +22,18 @@ def wrap_exception(message):
                 raise BundleServiceException(message + ': ' + str(e), e.client_error)
             except urllib.error.HTTPError as e:
                 try:
-                    client_error = json.loads(e.read().decode())
-
-                    if client_error['error'] == 'invalid_grant':
+                    client_error = e.read()
+                    if e.reason == 'invalid_grant':
                         raise BundleAuthException(
-                            message
-                            + ': '
-                            + http.client.responses[e.code]
-                            + ' - '
-                            + json.dumps(client_error),
+                            message + ': ' + http.client.responses[e.code] + ' - ' + client_error,
                             True,
                         )
                     else:
                         raise BundleServiceException(
-                            message
-                            + ': '
-                            + http.client.responses[e.code]
-                            + ' - '
-                            + json.dumps(client_error),
+                            message + ': ' + http.client.responses[e.code] + ' - ' + client_error,
                             e.code >= 400 and e.code < 500,
                         )
-                except json.decoder.JSONDecodeError:
+                except json.decoder.JSONDecodeError as e:
                     raise BundleServiceException(message + ': ' + str(e), False)
             except (urllib.error.URLError, http.client.HTTPException, socket.error) as e:
                 raise BundleServiceException(message + ': ' + str(e), False)
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -15,7 +15,7 @@ def wrap_exception(message):
                 return f(*args, **kwargs)
             except urllib.error.HTTPError as e:
                 # Translate known errors to the standard CodaLab errors
-                error_body = e.read().decode()
+                error_body = e.read()
                 exc = http_error_to_exception(e.code, error_body)
                 # All standard CodaLab errors are subclasses of UsageError
                 if isinstance(exc, UsageError):
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -31,8 +31,9 @@ but they expect the platform specific RunManagers they use to implement a common
 
 
 class Worker:
-    # Number of seconds to retry a bundle service client command
-    COMMAND_RETRY_SECONDS = 60 * 12
+    # Number of retries when a bundle service client command failed to execute. Defining a large number here
+    # would allow offline workers to patiently wait until connection to server is re-established.
+    COMMAND_RETRY_ATTEMPTS = 720
     # Network buffer size to use while proxying with netcat
     NETCAT_BUFFER_SIZE = 4096
     # Number of seconds to wait for bundle kills to propagate before forcing kill
@@ -524,7 +525,7 @@ class Worker:
 
     @staticmethod
     def execute_bundle_service_command_with_retry(cmd):
-        retries_left = Worker.COMMAND_RETRY_SECONDS
+        retries_left = Worker.COMMAND_RETRY_ATTEMPTS
         while True:
             try:
                 retries_left -= 1


codalab/worker/bundle_service_client.py
```diff
@@ -22,27 +22,20 @@ def wrapper(*args, **kwargs):
...
+                    e.seek(0)
+                    client_error = e.read()
+                    if e.reason == 'invalid_grant':
                         raise BundleAuthException(
-                            message
-                            + ': '
-                            + http.client.responses[e.code]
-                            + ' - '
-                            + json.dumps(client_error),
+                            message + ': ' + http.client.responses[e.code] + ' - ' + client_error,
```
percyliang, line 28: Why is `json.dumps` gone now?

candicegjing: 1. In the original logic [e.read().decode()](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/bundle_service_client.py#L25), `e.read()` was treated as a bytes object that can be decoded. However, in Python3, `e.read()` actually returns a string object which doesn't have `decode()` function. This is where the error `AttributeError: 'str' object has no attribute 'decode'` was coming from. So I removed the `decode()` method.
2. Since `e.read()` will return a plain string in Python3 now, the structure of parsing object `urllib.error.HTTPError` changed quite a lot comparing with the original way. The original `client_error` will return something similar to ```{
    "error": "bad client",
    "msg": "error message"
}```
However, right now, `e.read()` will directly return the error message as a Unicode string. I think it might not be necessary to convert it back and forth between a json object and string. (But I am open to discuss this, as one benefit of `json.dumps` is that it can ensure all the characters are ASCII encoded.)

codalab/worker/bundle_service_client.py
```diff
@@ -22,27 +22,20 @@ def wrapper(*args, **kwargs):
                 raise BundleServiceException(message + ': ' + str(e), e.client_error)
             except urllib.error.HTTPError as e:
                 try:
-                    client_error = json.loads(e.read().decode())
-
-                    if client_error['error'] == 'invalid_grant':
+                    # Sets the file's current position at the offset 0. Otherwise, e.read() will return an empty string.
+                    e.seek(0)
```
percyliang: This part is a bit strange...the offset is not always at 0?

percyliang (approved):

codalab/worker/bundle_service_client.py
candicegjing: There are two options:
1. Use `e.read()` but with `seek(0)` right before `e.read()`. Otherwise, `e.read()` will return an empty string from my experiments. 
2. Use `e.getvalue()`. This will directly return the real error message. It's one method belongs to `StringIO`, see [io.StringIO.getvalue](https://docs.python.org/3/library/io.html#io.StringIO.getvalue)
They are essentially the same. Open to change this, as I also feel it's quite strange in doing this.

candicegjing:
@percyliang We also need to decide whether we want to keep the value of [COMMAND_RETRY_SECONDS](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L35), as it's actually using as a counter instead of seconds in retrying the command and it makes the jobs stuck in `running` state for a very long time.

codalab/worker/bundle_service_client.py
percyliang: `e.getvalue()` seems much more principled / simpler than having to `seek` and `read`, so I'd go with that if possible.

percyliang:
@candicegjing Can you say more about `COMMAND_RETRY_SECONDS` (sounds like it's mis-named), what it's set to, and what behavior that causes, as well as what you think we should set it to?

candicegjing:
> @candicegjing Can you say more about `COMMAND_RETRY_SECONDS` (sounds like it's mis-named), what it's set to, and what behavior that causes, as well as what you think we should set it to?

Oh, I put more detailed information in [the original PR comments](https://github.com/codalab/codalab-worksheets/pull/1884#issue-359750668). Basically, `COMMAND_RETRY_SECONDS` is using as a counter to track the number of retries when a bundle service client command failed at [execute_bundle_service_command_with_retry](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L527). I think normally a few times (e.g. 3 to 10 times) should be enough. I have updated the value to be `10`. But if I missed any background information, please let me know.

percyliang (commented):
I think the reason we wanted a very long (perhaps infinite) wait time is to handle the case where the worker goes offline (maybe the server goes down for upgrade) - perhaps for days, and we want the worker to patiently wait until the connection is re-established.  This should probably be documented, and maybe there should be an exponential backoff.

codalab/worker/bundle_service_client.py
candicegjing: Actually, I just realized that I was wrong on this. `e.read()` without `e.seek(0)` should be the right way to handle it. The file pointer will originally be set at 0 when running `e.read()`, but will be changed after `e.read()` get executed. When I was testing, I tried to print out the logging information by calling `e.read()` once, then assigned the value to `client_error`, so the pointer changed its position.

percyliang: If we do `x = e.read()` once, we can just use `x` and not depend on the cursor position or `e` for the error message, right?

candicegjing:
> I think the reason we wanted a very long (perhaps infinite) wait time is to handle the case where the worker goes offline (maybe the server goes down for upgrade) - perhaps for days, and we want the worker to patiently wait until the connection is re-established. This should probably be documented, and maybe there should be an exponential backoff.

Yes. I agree with that case. #1875 happens when users' disk quota is full. Maybe I'll add a comment, update the variable naming and keep the value as the same as before for now, as either way the variable name isn't reflecting what it actually does in code logic?

codalab/worker/bundle_service_client.py
candicegjing: Yes, that's right. I have updated the code to use the standard `e.read()` in both files that have issues related to this.

Issue #1885, opened 2020-01-07
Page crashes when displaying records with React Errors


teetone:
**Describe the bug**
Page crashes when displaying records with react errors.

![page_crash](https://user-images.githubusercontent.com/16793796/71891197-eb8c0200-30fa-11ea-9d23-c46f1f4f12aa.png)

**To Reproduce**
Steps to reproduce the behavior :
1. Go to https://worksheets.codalab.org/worksheets/0x47273b9990a4448ea2055ddb67933d2b
2. Page crashes.

**Expected behavior**
I would expect the page not to crash.

yipenghe:
any chance knowing what's inside the worksheet?

teetone:
```
# Introduction
This is the **tiny** sample worksheet.

## Worksheet References

#### Valid Worksheet References
[worksheet valid_worksheet_adhuWVPjMczmRBqFwwdEBolm]{{0xb39daa5f257a476d94892406966afe68}} 
[worksheet valid_worksheet_SqzlnzGzgOLgWxGzhVZtSCwW]{{0x185e1055d96e41fd88a21e25cd3b3f02}}
[worksheet valid_worksheet_WFbNzsGFfhAeecPxLajQQjWX]{{0x934c42188e944a62985aa78e48ef8903}}

#### Private Worksheet References
[worksheet valid_private_worksheet_adhuWVPjMczmRBqFwwdEBolm]{{0xb716292b31b84f4b96b4b057d0e51a00}}
[worksheet valid_private_worksheet_SqzlnzGzgOLgWxGzhVZtSCwW]{{0xc66dd08532414e8e900fbd1fcf2e80ce}}
[worksheet valid_private_worksheet_WFbNzsGFfhAeecPxLajQQjWX]{{0x4af5cc3ad0b74883a0455c2f049ae8b0}}

## Bundle References

#### Valid Bundle References
[run run-echo : echo codalab rules!]{0x2598250a1ac74d31a5d525b86a9a6783} 
[run run-echo : echo codalab rules!]{0x52b4d8131ec64c099e1e056069b2694b} 
[run run-echo : echo codalab rules!]{0xc18099f4d41e4ae58e2a9c96756fae4a} 

#### Private Bundle References
[run run-echo : echo private run]{0xa3ec4e5861634baca9aba3e1bc8a8fed}
[run run-echo : echo private run]{0x94c42e8c85d54a728cef065e410d12b0}
[...]

PR #1886, opened 2020-01-07
1641: Stress Test Sample Worksheets Script

### LINKED ISSUE #1641: Add a script that creates tiny and giant worksheet from scratch (stress test front end rendering)
Let us create a big worksheet that tries to exercise all the worksheet rendering logic (math, markdown, bundles, subworksheets, search directives, comments, invalid directives), bundles with invalid permissions, Unicode, etc.

teetone:
Resolves #1641.
Depends on #1810 for the cleanup logic.

Usage:
Screen Shot 2020-01-16 at 12 20 52 AM

Running the script created the following tiny worksheet:
https://worksheets.codalab.org/worksheets/0xc2234172f064468197d2107120b0dfe4

The giant worksheet is pretty much the same as the tiny worksheet, but with a lot more bundle references, worksheets, markdown, etc.


sample_worksheet.py
```diff
@@ -0,0 +1,338 @@
...
+        self._add_text('% display table valid_schema')
+        self._add_bundles(self._valid_bundles)
+
+        self._add_subheader('Image')
+        for uuid in self._search_bundles('.png'):
+            self._add_text('% display image / width=500')
+            self._add_bundle(uuid)
+
+        self._add_subheader('Record')
+        self._add_text('% display record valid_schema')
```
teetone: Line 175 and 176 causes the page to crash with React errors. I have to comment out these two lines when creating the worksheet.

I filed a separate issue to investigate: https://github.com/codalab/codalab-worksheets/issues/1885

docker/dockerfiles/Dockerfile.server
```diff
@@ -50,6 +50,7 @@ COPY monitor.py monitor.py
 COPY scripts scripts
 COPY setup.py setup.py
 COPY test_cli.py test_cli.py
+COPY sample_worksheet.py sample_worksheet.py
```
percyliang: I'd move this to `scripts` since it's not an essential script to avoid cluttering the top-level, and name it `create-sample-worksheet.py` for consistency.

sample_worksheet.py
```diff
@@ -0,0 +1,338 @@
+import argparse
+import random
+import string
+
+from stress_test import cleanup
+from test_cli import run_command
+
+"""
+Script to create tiny and giant worksheets in any instance.
```
percyliang: Say more about what the purpose of these worksheets are.

percyliang (changes requested):
Can we output the worksheet contents to a temporary file, so that we can create the worksheet in one `cl wedit -f ...` command?  This will be much faster than having so many calls to `cl`, and also will be easier to debug (can quickly see what the worksheet will look like before actually creating it).  Of course, the worksheet depends on uuids.  There should be a preview mode where these uuids are fake so we can see what the worksheet markdown will look like without making any `cl` calls, and then an upload flag which actually makes the bundles that can then be referenced.

Actually, to the extent

teetone:
> Can we output the worksheet contents to a temporary file, so that we can create the worksheet in one `cl wedit -f ...` command? This will be much faster than having so many calls to `cl`, and also will be easier to debug (can quickly see what the worksheet will look like before actually creating it). Of course, the worksheet depends on uuids. There should be a preview mode where these uuids are fake so we can see what the worksheet markdown will look like without making any `cl` calls, and then an upload flag which actually makes the bundles that can then be referenced.

@percyliang: In my latest changes, I leveraged `wedit` to make the script run faster. I ran into validation errors when referencing fake UUIDs with `wedit`, so instead I referenced existing bundles and worksheets when in preview mode.

In preview mode, the script creates a tiny worksheet in ~8 seconds and a giant worksheet in ~10 seconds.

scripts/create-sample-worksheet.py
```diff
@@ -0,0 +1,383 @@
+import argparse
+import os
+import random
+import string
+import sys
+import time
+
+sys.path.append('..')
+from stress_test import cleanup
```
percyliang: Can we put `cleanup` and `run_command` in a separate utilities file?  Creating a worksheet shouldn't depend on tests.

scripts/create-sample-worksheet.py
```diff
@@ -0,0 +1,383 @@
...
+from test_cli import run_command
+
+"""
+Script to create tiny and giant worksheets in any instance to stress test the front end. The purpose of
```
percyliang: Maybe call `small` and `large` everywhere?  It's not that tiny...

percyliang (changes requested):

codalab/rest/bundle_actions.py
```diff
@@ -19,11 +19,7 @@ def create_bundle_actions():
     check_bundles_have_all_permission(local.model, request.user, [a['uuid'] for a in actions])
     for action in actions:
         bundle = local.model.get_bundle(action['uuid'])
-        if bundle.state in [
```
teetone: This change is currently causing master to fail.

test_cli.py
```diff
@@ -279,33 +170,21 @@ def wait_until_substring(fp, substr):
...
-        return cls._colorize(string, "GREEN")
-
-    @classmethod
-    def yellow(cls, string):
-        return cls._colorize(string, "YELLOW")
-
-    @classmethod
-    def cyan(cls, string):
-        return cls._colorize(string, "CYAN")
+def _run_command(
```
epicfaace, line 173: Would it be better just to get rid of this function and call `run_command` from all the tests instead?

scripts/create-sample-worksheet.py
```diff
@@ -0,0 +1,379 @@
+import argparse
+import os
+import random
+import string
+import sys
+import time
+
+sys.path.append('..')
```
epicfaace: Why do you need this line?

test_cli.py
teetone: This is because of the `force_subprocess` check we had previously.  I don't want to check `force_subprocess` or have to pass in the `cl` instance everytime I call the `run_command`.

scripts/create-sample-worksheet.py
teetone: I originally had the util on the root level, but now it's not needed since I moved it to `scripts`. I updated.

scripts/test_util.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def run_command(
+    args,
+    expected_exit_code=0,
+    max_output_chars=1024,
+    env=None,
+    include_stderr=False,
+    binary=False,
+    use_cli_directly=False,
```
percyliang: This name is not intuitive to me since CLI can mean either invoking `cl` or creating `BundleCLI`.  Why change it from `force_subprocess`?

percyliang (approved):

scripts/test_util.py
teetone: @percyliang  I will change it back before merging.

PR #1887, opened 2020-01-07
1883: Get rid of MySQL operational errors during worker checkin's

### LINKED ISSUE #1883: MySQL OperationalError: (MySQLdb._exceptions.OperationalError)
We need to fix the following error:

OperationalError: (MySQLdb._exceptions.OperationalError) (1048, "Column 'shared_file_system' cannot be null") [SQL: 'INSERT INTO worker (user_id, worker_id, tag, cpus, gpus, memory_bytes, free_disk_bytes, checkin_time, socket_id, shared_file_system) VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s)'] [parameters: ('0xf62bb09b33574bb291d40cbbf9a0ea50', 'deep24.stanford.edu(24705)', 'deep', 8, 4, 135170699264, 768393965568, datetime.datetime(2020, 1, 6, 7, 24, 38, 2736), 12687725, None)]

teetone:
Fixes #1883. 

`shared_file_system` is a relatively new field that old workers most likely do not specify. The errors happen when inserting null values for `shared_file_system`. This fix gets rid of these errors by using a default value of `False` if the value of the field is unspecified.

### DIFF
--- a/codalab/rest/workers.py
+++ b/codalab/rest/workers.py
@@ -23,8 +23,7 @@ def checkin(worker_id):
     """
     WAIT_TIME_SECS = 3.0
 
-    # Old workers might not have all the fields, so allow subsets to be
-    # missing.
+    # Old workers might not have all the fields, so allow subsets to be missing.
     socket_id = local.worker_model.worker_checkin(
         request.user.user_id,
         worker_id,
@@ -34,9 +33,8 @@ def checkin(worker_id):
         request.json.get("memory_bytes"),
         request.json.get("free_disk_bytes"),
         request.json["dependencies"],
-        request.json.get("shared_file_system"),
+        request.json.get("shared_file_system", False),
     )
-
     for run in request.json["runs"]:
         try:
             worker_run = BundleCheckinState.from_dict(run)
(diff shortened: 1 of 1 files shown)


codalab/rest/workers.py
```diff
@@ -34,7 +33,7 @@ def checkin(worker_id):
         request.json.get("memory_bytes"),
         request.json.get("free_disk_bytes"),
         request.json["dependencies"],
-        request.json.get("shared_file_system"),
+        request.json.get("shared_file_system", False),
```
percyliang, line 36: For consistency with the other fields, could we just pass `None` back?

bkgoksel: The database schema for this field does not accept null values. The cleanest solution would be to update all the workers so they simply report this field, but if that's not possible defaulting to False is the right thing to do.

candicegjing:
I was actually wondering how this problem could happen as we saw this error every time right after we deployed a new release to production and it will disappear after a few hours? Based on the error message, the MySQL insertion request came from a private worker `deep23.stanford.edu(20855)` that wasn't connected to our rest-server at the time when the error was thrown (verified in our database as well). Moreover, I also verified that this particular user listed in the insertion record didn't try to connect as well.

percyliang (approved):

PR #1888, opened 2020-01-07
Do not depend on the mysql service in Docker compose

bkgoksel:
Using depends_on to make other services depend on mysql doesn't make
sense if we want to allow remote mysql hosts since docker compose
wouldn't know about them. Additionally, we don't really make use of the
depends_on statements given the way we use docker compose (ie by
bringing up services one by one using an external management script)

### DIFF
--- a/docker/compose_files/docker-compose.yml
+++ b/docker/compose_files/docker-compose.yml
@@ -75,16 +75,12 @@ services:
     command: cl-server
     <<: *codalab-base
     <<: *codalab-server
-    depends_on:
-      - mysql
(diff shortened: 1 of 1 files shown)


percyliang (commented):
Is the `depends_on` used for teardown, in which case order does matter...?

This hasn't been a problem on the public codalab instance, where mysql is on a different box...

bkgoksel:
Looks like bringing the rest up without mysql works even with the dependency. Strange but in any case this PR isn't needed then.

PR #1890, opened 2020-01-08
update datetime get time function

### LINKED ISSUE #1840: Different timezones for bundles in Worksheet vs Bundle View
The timezone used for datetime fields for bundles in a worksheet is different from the timezone used in Bundle View. This can be confusing to users. See the screenshots below for an example.

Screen Shot 2019-12-10 at 2 33 20 AM
Screen Shot 2019-12-10 at 2 33 30 AM

Levilian:
- following rules are followed
1. utcnow() is preferred when the time is not user-facing, and will be used to calculate the difference
2. now() is preferred when the time will be displayed to the user
3. in general, utcnow() is much faster than now(), and takes into account timezone differences

fix #1840

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2957,7 +2957,7 @@ class BundleCLI(object):
                 owner = client.fetch('users', args.owner_spec)
                 info['owner'] = JsonApiRelationship('users', owner['id'])
             if args.freeze:
-                info['frozen'] = datetime.datetime.utcnow()
+                info['frozen'] = datetime.datetime.utcnow().isoformat()
             if args.anonymous is not None:
                 info['is_anonymous'] = args.anonymous
 
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2957,7 +2957,7 @@ class BundleCLI(object):
                 owner = client.fetch('users', args.owner_spec)
                 info['owner'] = JsonApiRelationship('users', owner['id'])
             if args.freeze:
-                info['frozen'] = datetime.datetime.utcnow().isoformat()
+                info['frozen'] = datetime.datetime.utcnow()
             if args.anonymous is not None:
                 info['is_anonymous'] = args.anonymous
 
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -530,7 +530,7 @@ class CodaLabManager(object):
         epoch_str = formatting.datetime_str(datetime.datetime.utcfromtimestamp(0))
         last_check_str = self.state.get('last_check_version_datetime', epoch_str)
         last_check_dt = formatting.parse_datetime(last_check_str)
-        now = datetime.datetime.now()
+        now = datetime.datetime.utcnow()
         if (now - last_check_dt) < datetime.timedelta(days=1):
             return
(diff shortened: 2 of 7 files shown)


codalab/worker/docker_utils.py
```diff
@@ -258,10 +258,14 @@ def get_container_running_time(container):
...
     # then using the current time as end_time
+    # Because naive datetime objects are treated by many datetime methods as local times,
+    # it is preferred to use aware datetimes to represent times in UTC. As such, the
+    # recommended way to create an object representing the current time in UTC is
+    # by calling datetime.now(timezone.utc).
     end_time = (
         container.attrs['State']['FinishedAt']
         if container.attrs['State']['Status'] == 'exited'
-        else str(datetime.datetime.now().replace(tzinfo=tz.tzutc()))
+        else str(datetime.now(timezone.utc))
```
candicegjing: It's better to not change the docker container running time calculation to `timezone.utz`, as the start time received from Docker rest API uses a special format that sets the `tzinfo`, see the discussion here [PR #1687 ](https://github.com/codalab/codalab-worksheets/pull/1687#discussion_r350552900). Keep both the start time and end time in the exact same format will ensure us to achieve better accuracy in calculating total running time for a container.

codalab/lib/bundle_cli.py
```diff
@@ -2957,7 +2957,7 @@ def do_wedit_command(self, args):
                 owner = client.fetch('users', args.owner_spec)
                 info['owner'] = JsonApiRelationship('users', owner['id'])
             if args.freeze:
-                info['frozen'] = datetime.datetime.utcnow().isoformat()
+                info['frozen'] = datetime.datetime.utcnow()
```
candicegjing: `datetime.datetime.utcnow()` will return a datetime object, but with `isoformat()`, it will return a string representing the date in ISO 8601 format. Maybe we should keep the `isoformat()`?

codalab/worker/docker_utils.py
Levilian: Thanks for the reminder!

codalab/lib/bundle_cli.py
Levilian: @candicegjing The reason why I changed it is because https://github.com/codalab/codalab-worksheets/blob/179e117b3ee4b4a51013fa4c6509e1d83ec58cac/codalab/rest/worksheets.py#L375-L378 previously we used .now(), which still gives a datetime. Should we make them consistent?

candicegjing: Yes. Keep them consistent would be better. And it also depends on what we want to save in `info[frozen]`

Levilian: Do you know if we want a string or a datetime?

Levilian: Right now `info[frozen]` is not really used anywhere. metadata is also separate from the value.

codalab/lib/codalab_manager.py
```diff
@@ -530,7 +530,7 @@ def check_version(self, server_version):
         epoch_str = formatting.datetime_str(datetime.datetime.utcfromtimestamp(0))
         last_check_str = self.state.get('last_check_version_datetime', epoch_str)
```
candicegjing, line 531: We probably also need to check what's the timezone when generating `last_check_version_datetime`, as the calculation in the next two lines will depend on this.

PR #1896, opened 2020-01-13
1779: Handle preceding and trailing white spaces in worksheets

### LINKED ISSUE #1779: [Frontend] Trailing whitespaces for worksheet or bundle references causes rendering problems
**To Reproduce**
1. Go to a worksheet with worksheet or bundle references.
2. Edit the source and add some trailing spaces to some references.
3. The references from step 2 just get treated as plain text (see screenshot below).

Link: https://worksheets-dev.codalab.org/worksheets/0xc98f4a352680448fb1ff498993d9f1ce

**Expected behavior**
We should ignore trailing white space. This can be confusing to users.

**Screenshots**
Screen Shot 2019-12-01 at 2 35 02 AM

teetone:
Fixes #1779 

I updated the regex to match for bundles, worksheets and directives that might have some trailing or preceding whitespace in a given worksheet. My original fix for this issue relied on `strip()`, but I later realized that it may strip away valid, preceding whitespace for some markdown.

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -62,11 +62,11 @@ TYPE_WORKSHEET = 'worksheet'
 WORKSHEET_ITEM_TYPES = (TYPE_MARKUP, TYPE_DIRECTIVE, TYPE_BUNDLE, TYPE_WORKSHEET)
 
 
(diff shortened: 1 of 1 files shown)


percyliang (approved):

candicegjing (approved):
Looks good!

PR #1897, opened 2020-01-13
Set maximum docker image size before pulling from docker hub

### LINKED ISSUE #1894: set maximum docker image size
We should have an environment variable or command-line argument on the workers that limits the maximum docker image.  A good default is 10 GB.

candicegjing:
Fixed #1894.
Final failure message will be showing as follows:
Screen Shot 2020-01-13 at 12 25 56 PM

Issues:
Tried using [docker-py](https://docker-py.readthedocs.io/en/stable/images.html#registrydata-objects) at the beginning as I want to reuse the existing libraries instead of introducing a new one. However, by examining the results return from `RegistryData` object as following:
```
>>> import docker
>>> client = docker.from_env()
>>> r = client.images.get_registry_data("codalab/default-cpu")
>>> r.attrs
{'Descriptor': {'mediaType': 'application/vnd.docker.distribution.manifest.v2+json', 'digest': 'sha256:ff7de716ac5f94d9b4ff8d6341d60166ed5b0163cc05d8d51da9418511caf5e1', 'size': 4544}, 'Platforms': [{'architecture': 'amd64', 'os': 'linux'}]}
```
The returned `size` element is quite different than its actual value [(5G in docker hub)](https://hub.docker.com/layers/codalab/default-cpu/latest/images/sha256-ff7de716ac5f94d9b4ff8d6341d60166ed5b0163cc05d8d51da9418511caf5e1). 
I also tried to find any documentation about the `size` returned from `RegistryData.attrs` and correlation between the size return by `RegistryData.attrs` and the actual value. Nothing helpful has been discovered. (**If anyone has suggestions of this, please feel free to let me know.**)
Then, I decided to move to Docker HTTP API V2 to query the size information of an image before pulling from docker hub.

### DIFF
--- a/codalab/worker/docker_image_manager.py
+++ b/codalab/worker/docker_image_manager.py
@@ -5,14 +5,18 @@ import traceback
 import logging
 
 import docker
+import codalab.worker.docker_utils as docker_utils
 
 from codalab.worker.fsm import DependencyStage
 from codalab.worker.state_committer import JsonStateCommitter
 from codalab.worker.worker_thread import ThreadDict
+from codalab.lib.formatting import size_str
+
 
 logger = logging.getLogger(__name__)
 
-# Stores the download state of a Docker image (also includes the digest being pulled, digest string, DependencyStage and relevant status message from the download)
+# Stores the download state of a Docker image (also includes the digest being pulled, digest string,
+# DependencyStage and relevant status message from the download)
 ImageAvailabilityState = namedtuple('ImageAvailabilityState', ['digest', 'stage', 'message'])
 # Stores information relevant about caching about docker images
 ImageCacheEntry = namedtuple(
@@ -21,11 +25,12 @@ ImageCacheEntry = namedtuple(
 
 
 class DockerImageManager:
-    def __init__(self, commit_file, max_image_cache_size):
+    def __init__(self, commit_file, max_image_cache_size, max_image_size):
         """
         Initializes a DockerImageManager
         :param commit_file: String path to where the state file should be committed
         :param max_image_cache_size: Total size in bytes that the image cache can use
+        :param max_image_size: Total size in bytes that the image can have
         """
         self._state_committer = JsonStateCommitter(commit_file)  # type: JsonStateCommitter
         self._docker = docker.from_env()  # type: DockerClient
@@ -34,6 +39,7 @@ class DockerImageManager:
             fields={'success': False, 'status': 'Download starting.'}, lock=True
         )
         self._max_image_cache_size = max_image_cache_size
+        self._max_image_size = max_image_size
         self._lock = threading.RLock()
 
         self._stop = False
@@ -218,12 +224,34 @@ class DockerImageManager:
                             'message'
                         ] = "Can't download image: {}".format(ex)
 
-                self._downloading.add_if_new(image_spec, threading.Thread(target=download, args=[]))
+                # Check docker image size before pulling from Docker Hub.
+                # Do not download images larger than self._max_image_size
+                image_size_bytes = docker_utils.get_image_size_without_pulling(image_spec)
+                if image_size_bytes is None:
+                    failure_msg = (
+                        "Unable to find image: " + image_spec + " from Docker HTTP API V2."
+                    )
+                elif image_size_bytes < self._max_image_size:
+                    self._downloading.add_if_new(
+                        image_spec, threading.Thread(target=download, args=[])
+                    )
+                    return ImageAvailabilityState(
+                        digest=None,
+                        stage=DependencyStage.DOWNLOADING,
+                        message=self._downloading[image_spec]['status'],
+                    )
+                else:
+                    failure_msg = (
+                        "The size of "
+                        + image_spec
+                        + ": {} exceeds the maximum image size allowed {}.".format(
+                            size_str(image_size_bytes), size_str(self._max_image_size)
+                        )
+                    )
                 return ImageAvailabilityState(
-                    digest=None,
-                    stage=DependencyStage.DOWNLOADING,
-                    message=self._downloading[image_spec]['status'],
+                    digest=None, stage=DependencyStage.FAILED, message=failure_msg
                 )
+
         except Exception as ex:
             return ImageAvailabilityState(
                 digest=None, stage=DependencyStage.FAILED, message=str(ex)
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -11,13 +11,14 @@ import os
 import docker
 from dateutil import parser, tz
 import datetime
-from codalab.lib.formatting import parse_size
+import requests
 
 
 MIN_API_VERSION = '1.17'
 NVIDIA_RUNTIME = 'nvidia'
 DEFAULT_RUNTIME = 'runc'
 DEFAULT_TIMEOUT = 720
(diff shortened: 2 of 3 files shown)


codalab/worker/docker_utils.py
```diff
@@ -268,3 +269,31 @@ def get_container_running_time(container):
     # formatted datetime string directly.
     container_running_time = parser.isoparse(end_time) - parser.isoparse(start_time)
     return container_running_time.total_seconds()
+
+
+@wrap_exception('Unable to check image size before pulling from Docker Hub')
+def check_image_size_before_pulling(image_spec):
```
percyliang: To make this more modular and reusable, I'd break this up into two functions, one that just gets the size of the image, and another one that checks the size.  In fact, the checking logic doesn't need to be in this file at all since it's not core to docker.

codalab/worker/docker_utils.py
```diff
@@ -11,17 +11,18 @@
...
-
+import requests
+from codalab.lib import formatting
 
 MIN_API_VERSION = '1.17'
 NVIDIA_RUNTIME = 'nvidia'
 DEFAULT_RUNTIME = 'runc'
 DEFAULT_TIMEOUT = 720
-
+MAXIMUM_DOCKER_IMAGE_SIZE = '10g'
```
percyliang: I'd make this a command-line flag.

codalab/worker/docker_image_manager.py
```diff
@@ -218,6 +219,15 @@ def download():
                             'message'
                         ] = "Can't download image: {}".format(ex)
 
+                # Check docker image size before pulling from docker hub.
+                # Images with size larger than '10g' will be rejected to download.
+                size_validated, failure_msg = docker_utils.check_image_size_before_pulling(
```
percyliang: The logic here can just be get the size, and just check whether we're exceeding the limit.

percyliang (changes requested):

codalab/worker/docker_utils.py
candicegjing: Agree. It's updated.

codalab/worker/docker_utils.py
candicegjing: Splitted the logic.

codalab/worker/docker_image_manager.py
candicegjing: Updated.

codalab/worker/docker_image_manager.py
```diff
@@ -21,11 +25,12 @@
 
 
 class DockerImageManager:
-    def __init__(self, commit_file, max_image_cache_size):
+    def __init__(self, commit_file, max_image_cache_size, max_image_size):
         """
         Initializes a DockerImageManager
         :param commit_file: String path to where the state file should be committed
         :param max_image_cache_size: Total size in bytes that the image cache can use
+        :param max_image_size: Total size in bytes that the image can has
```
percyliang: typo

codalab/worker/docker_image_manager.py
```diff
@@ -218,12 +224,34 @@ def download():
                             'message'
                         ] = "Can't download image: {}".format(ex)
 
-                self._downloading.add_if_new(image_spec, threading.Thread(target=download, args=[]))
+                # Check docker image size before pulling from docker hub.
+                # Images with size larger than '10g' will be rejected to download.
```
percyliang: Update the 10g

codalab/worker/docker_image_manager.py
```diff
@@ -218,12 +224,34 @@ def download():
                             'message'
                         ] = "Can't download image: {}".format(ex)
 
-                self._downloading.add_if_new(image_spec, threading.Thread(target=download, args=[]))
+                # Check docker image size before pulling from docker hub.
```
percyliang: 'Docker Hub'

codalab/worker/docker_image_manager.py
percyliang: fix grammar: 'Do not download images larger than `max_image_size`.'

codalab/worker/docker_image_manager.py
```diff
@@ -218,12 +224,34 @@ def download():
...
+                # Images with size larger than '10g' will be rejected to download.
+                image_size_bytes = docker_utils.get_image_size_without_pulling(image_spec)
+                if 0 < image_size_bytes < self._max_image_size:
```
percyliang: Why is `0 < ` necessary?  Should be always positive no?

codalab/worker/docker_image_manager.py
```diff
@@ -218,12 +224,34 @@ def download():
...
+                    self._downloading.add_if_new(
+                        image_spec, threading.Thread(target=download, args=[])
+                    )
+                    return ImageAvailabilityState(
+                        digest=None,
+                        stage=DependencyStage.DOWNLOADING,
+                        message=self._downloading[image_spec]['status'],
+                    )
+                elif image_size_bytes == 0:
```
percyliang: Don't use `0` to denote failure since it's too cryptic / easy to miss that case....I'd throw an exception or return `None`.

codalab/worker/docker_utils.py
```diff
@@ -268,3 +268,20 @@ def get_container_running_time(container):
     # formatted datetime string directly.
     container_running_time = parser.isoparse(end_time) - parser.isoparse(start_time)
     return container_running_time.total_seconds()
+
+
+@wrap_exception('Unable to get image size without pulling from Docker Hub')
+def get_image_size_without_pulling(image_spec):
+    logger.info("Downloading tag information for {}".format(image_spec))
```
percyliang, line 281: Say why you're not using the dockerpy API.

percyliang (changes requested):

codalab/worker/docker_image_manager.py
candicegjing: There is a chance that we don't get the record of the requested image from the docker registry, then the return image size will be empty. In this particular case, I was defining it as `0`. I guess `None` will be better.

codalab/worker/docker_image_manager.py
```diff
@@ -218,12 +224,34 @@ def download():
                             'message'
                         ] = "Can't download image: {}".format(ex)
 
-                self._downloading.add_if_new(image_spec, threading.Thread(target=download, args=[]))
+                # Check docker image size before pulling from Docker Hub.
+                # Do not download images larger than self._max_image_size
+                image_size_bytes = docker_utils.get_image_size_without_pulling(image_spec)
+                if image_size_bytes and image_size_bytes < self._max_image_size:
```
percyliang: make this epxlicit: `is not None`

codalab/worker/docker_image_manager.py
```diff
@@ -218,12 +224,34 @@ def download():
...
+                    self._downloading.add_if_new(
+                        image_spec, threading.Thread(target=download, args=[])
+                    )
+                    return ImageAvailabilityState(
+                        digest=None,
+                        stage=DependencyStage.DOWNLOADING,
+                        message=self._downloading[image_spec]['status'],
+                    )
+                elif not image_size_bytes:
```
percyliang: Put this first (`image_size_bytes is None`), so that you don't need the `image_size_bytes` condition above.

percyliang (approved):

PR #1905, opened 2020-01-16
1895: Resolve key error

### LINKED ISSUE #1895: Python KeyError  for GET /interpret/worksheet/[uuid]
Seeing the following error in DEV during stress testing.

  rv = callback(*a, **ka)
File "/opt/codalab-worksheets/codalab/rest/interpret.py", line 178, in fetch_interpreted_worksheet
  worksheet_info['items'] = expand_raw_items(worksheet_info['items'])
File "/opt/codalab-worksheets/codalab/rest/interpret.py", line 579, in expand_raw_items
  return list(chain.from_iterable([expand_raw_item(raw_item) for raw_item in raw_items]))
File "/opt/codalab-worksheets/codalab/rest/interpret.py", line 579, in <listcomp>
  return list(chain.from_iterable([expand_raw_item(raw_item) for raw_item in raw_items]))
File "/opt/codalab-worksheets/codalab/rest/interpret.py", line 613, in expand_raw_item
  raw_items.append(bundle_item(bundle_infos[bundle_uuid]) + (None, None))
KeyError: ‘0xe8278af10e8941bbb6e9ed9c5cd7b6ee’

The error bubbles up to the UI:

Screen Shot 2020-01-12 at 7 17 56 PM

teetone:
Fixes #1895 

In a few places, we query for bundle UUIDs and then query for their corresponding bundle infos. However, there is a chance that bundles get deleted in between queries, which in turn, causes key errors to surface. This is a simple fix to prevent these errors. An alternative solution would be to change how we query for bundles.

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -280,7 +280,8 @@ def _create_bundles():
     bundles_dict = get_bundle_infos(created_uuids)
 
     # Return bundles in original order
-    # There is a chance that a bundle is created and deleted right after, while fetching bundles info's
+    # Need to check if the UUID is in the dict, since there is a chance that a bundle is deleted
+    # right after being created.
     bundles = [bundles_dict[uuid] for uuid in created_uuids if uuid in bundles_dict]
     return BundleSchema(many=True).dump(bundles).data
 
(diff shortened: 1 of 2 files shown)


candicegjing (approved):
Looks good! So this [database query](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/util.py#L71) is the line that causes the discrepancy, as those bundles fetched based on given uuids could be a subset of the original uuids?

teetone:
> Looks good! So this [database query](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/util.py#L71) is the line that causes the discrepancy, as those bundles fetched based on given uuids could be a subset of the original uuids?

@candicegjing  Yes, exactly and in some places, we assume that we will get a bundle for each uuid.

percyliang (approved):

PR #1917, opened 2020-01-21
Set /bin/bash to bash shell in docker command.

### LINKED ISSUE #1704: Prod worker can not be started due to "exec: \"bash\": executable file not found in $PATH": unknown"

candicegjing:
Fixed #1704. 
This PR will fix the problem caused by the following error: 
```"exec: \"bash\": executable file not found in $PATH": unknown"```

This error had periodically happened in production, shown as follows:
```
docker.errors.APIError: 400 Client Error: Bad Request ("OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"bash\": executable file not found in $PATH": unknown")
```
I had done some investigation, findings are all documented at the original ticket. Currently, PATH is set to have the following paths:
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
Explicitly setting the path of bash shell in environment variable $PATH will help the program to navigate the right bash shell easier.

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -119,7 +119,9 @@ def start_bundle_container(
 ):
     if not command.endswith(';'):
         command = '{};'.format(command)
-    docker_command = ['bash', '-c', '( %s ) >stdout 2>stderr' % command]
+    # Explicitly specifying "/bin/bash" instead of "bash" for bash shell to avoid the situation when
+    # the program can't find the symbolic link (default is "/bin/bash") of bash in the environment
+    docker_command = ['/bin/bash', '-c', '( %s ) >stdout 2>stderr' % command]
     docker_bundle_path = '/' + uuid
     volumes = get_bundle_container_volume_binds(bundle_path, docker_bundle_path, dependencies)
     environment = {'HOME': docker_bundle_path, 'CODALAB': 'true'}

### CODE REVIEW

File: docker/dockerfiles/Dockerfile.worker [hunk 1]
```diff
@@ -11,6 +11,7 @@ FROM ubuntu:16.04
 MAINTAINER CodaLab Worksheets <codalab.worksheets@gmail.com>
 
 ENV DEBIAN_FRONTEND noninteractive
+ENV PATH="/bin/bash:${PATH}"
```
percyliang: Why is this necessary, especially since you have `/bin/bash` in the docker command above?

percyliang:
How come this wasn't an issue before / when does this error occur?

docker/dockerfiles/Dockerfile.worker
candicegjing: Theoretically yes. However, the existing `bash` is already linked to `/bin/bash`, but the program still can't find it when executing. I did several ways of testing, they are documented at [here](https://github.com/codalab/codalab-worksheets/issues/1704#issuecomment-575926090). It appears that setting only the `PATH` in the environment variable still can't help the program to navigate to the right executable. So I feel the safest way is to add BASH executable directly to `PATH` and call `/bin/bash` in the code.

candicegjing:
> How come this wasn't an issue before / when does this error occur?

This issue actually happened in the past quarter when we had an error about **duplicate container name**. The **duplicate container name**  exception is triggered by **the unknown bash executable exception**. There seem to be many discussions about this issue. Some posts mentioned this is related to the specific version of the docker engine version: `18.09.*`, which is the same one we are using. Some of the properties within the docker config shouldn't be set (e.g. `MountFlags=slave`). But this is not our issue. We don't have that configuration in our docker service config file. Then another possible reason is from the command itself. The command may evoke a new environment that doesn't have the linked path in the original environment (e.g. subshell).

percyliang (commented):
Could you please summarize the consequences of this change and any outstanding points that someone in the future can address?  Thanks!

candicegjing:
> Could you please summarize the consequences of this change and any outstanding points that someone in the future can address? Thanks!

I added a comment above the line that declares the `/bin/bash`.  I think explicitly specifying this won't affect anything. The only thing could be bad is when `/bin/bash` doesn't exist in the system. But this should happen rarely.

PR #1918, opened 2020-01-21
Fix travis CI random failures due to cl kill command

candicegjing:
This is to fix Travis CI tests failed to `kill` bundles. 
Bundles in `staged` state can't be killed directly. It needs to be removed instead.

### DIFF
--- a/codalab/rest/bundle_actions.py
+++ b/codalab/rest/bundle_actions.py
@@ -17,20 +17,34 @@ def create_bundle_actions():
     actions = BundleActionSchema(strict=True, many=True).load(request.json).data
 
     check_bundles_have_all_permission(local.model, request.user, [a['uuid'] for a in actions])
-
     for action in actions:
         bundle = local.model.get_bundle(action['uuid'])
-        if bundle.state not in [State.RUNNING, State.PREPARING]:
-            raise UsageError('Cannot execute this action on a bundle that is not running.')
+        if bundle.state in [
+            State.READY,
+            State.FAILED,
+            State.KILLED,
+        ]:
+            raise UsageError(
+                'Cannot execute this action on a bundle that is in the following states: ready, failed, killed. '
+                'Kill action can be executed on bundles in created, uploading, staged, making, starting, '
+                'running, preparing, or finalizing state.'
+            )
 
         worker = local.model.get_bundle_worker(action['uuid'])
-        precondition(
-            local.worker_model.send_json_message(worker['socket_id'], action, 60),
-            'Unable to reach worker.',
-        )
-
         new_actions = getattr(bundle.metadata, 'actions', []) + [BundleAction.as_string(action)]
-        db_update = {'metadata': {'actions': new_actions}}
-        local.model.update_bundle(bundle, db_update)
+
+        # The state updates of bundles in PREPARING, RUNNING, or FINALIZING state will be handled on the worker side.
+        if worker:
+            precondition(
+                local.worker_model.send_json_message(worker['socket_id'], action, 60),
+                'Unable to reach worker.',
+            )
+            local.model.update_bundle(bundle, {'metadata': {'actions': new_actions}})
+        else:
+            # The state updates of bundles in CREATED, UPLOADING, MAKING, STARTING or STAGED state
+            # will be handled on the rest-server side.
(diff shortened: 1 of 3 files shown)


percyliang (commented):
What's wrong with killing staged bundles?  We kind of want those to not get started in the meantime before we `rm` it...

candicegjing:
> What's wrong with killing staged bundles? We kind of want those to not get started in the meantime before we `rm` 

The following examples are captured from[ the latest CI build](https://travis-ci.org/codalab/codalab-worksheets/jobs/639754030?utm_medium=notification&utm_source=github_status) in master.
Screen Shot 2020-01-21 at 3 24 45 PM
Screen Shot 2020-01-21 at 3 24 58 PM
Based on the logic in [bundle-action](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/bundle_actions.py#L23), only the bundles in `PREPARING` and `RUNNING` state can be killed. Killing a bundle in any other states will throw an exception and not work as expected. 
In test_cli.py, restricting killing bundles in `PREPARING` and `RUNNING` state will avoid those unnecessary failed assertions, which may trigger CI build failure.

codalab/client/json_api_client.py
```diff
@@ -14,8 +14,11 @@ def wrapper(*args, **kwargs):
             try:
                 return f(*args, **kwargs)
             except urllib.error.HTTPError as e:
+                # Standardize the data type of error_body generated from different objects (e.g. bottle.HTTPError or
+                # urllib.error.HTTPError) to string. The data type of error message generated by
+                # urllib.error.HTTPError is string by default. The one generated by bottle.HTTPError is bytes.
+                error_body = e.read().decode() if isinstance(e.read(), bytes) else e.read()
```
candicegjing: We have two resources that generate this exception. One is from `bottle.HTTPError`. Another is from `urllib.error.HTTPError`. The type of error message generated by `bottle.HTTPError` is **bytes**. The other one is **string**. We can either choose to handle the different data types at here or raise the same `HTTPError` throughout our codebase. Open to discuss this.

percyliang: Seems like this should be in another PR?

percyliang: I think you want to call `e.read()` only once.

codalab/rest/bundle_actions.py
```diff
@@ -17,20 +17,36 @@ def create_bundle_actions():
...
-            raise UsageError('Cannot execute this action on a bundle that is not running.')
+        if bundle.state not in [
+            State.CREATED,
+            State.UPLOADING,
+            State.STAGED,
+            State.PREPARING,
+            State.RUNNING,
+            State.FINALIZING,
+        ]:
+            raise UsageError(
```
percyliang, line 27: At this point, isn't better to define this in terms of terminal states (ready, failed)?

percyliang (changes requested):

codalab/rest/bundle_actions.py
candicegjing: Sure. I was debating if I should list all the states here or just using terminal states. Updating it.

codalab/client/json_api_client.py
candicegjing: Will create a different PR.

candicegjing:
@percyliang This is updated and retested. Let me know if you think I need to change anything else.

docs/CLI-Reference.md
```diff
@@ -253,7 +253,8 @@ This file is auto-generated from the output of `cl help -v` and provides the lis
           -v, --verbose                Display verbose output.
 
       kill:
-        Instruct the appropriate worker to terminate the running bundle(s).
+        Instruct the appropriate worker to terminate the bundle(s) in the following states:
+         created, uploading, staged, making, starting, running, preparing, or finalizing.
```
bkgoksel: Since we can kill bundle before they're even assigned a worker now, let's change the language here to reflect that and maybe something like "terminate the bundle execution if it's being executed or remove it from the execution queue" (but maybe with better wording)

bkgoksel (approved):
Great! I think beyond travis, it's a better UX to be able to kill any non-finished bundle as well.

PR #1927, opened 2020-01-24
 Use docker image tags to keep track of cached images

bkgoksel:
Previously we used an in-memory list and a JSON file to commit to disk
but these occasionally fall out of sync with the reality on the machine
and cause problems. This was the risk of falling out of sync should be
much lower.

### DIFF
--- a/codalab/worker/docker_image_manager.py
+++ b/codalab/worker/docker_image_manager.py
@@ -25,6 +25,9 @@ ImageCacheEntry = namedtuple(
 
 
 class DockerImageManager:
+
+    CACHE_TAG = 'codalab-image-cache/last-used'
+
     def __init__(self, commit_file, max_image_cache_size, max_image_size):
         """
         Initializes a DockerImageManager
@@ -34,37 +37,26 @@ class DockerImageManager:
         """
         self._state_committer = JsonStateCommitter(commit_file)  # type: JsonStateCommitter
         self._docker = docker.from_env()  # type: DockerClient
-        self._image_cache = {}  # type: Dict[str, ImageCacheEntry]
         self._downloading = ThreadDict(
             fields={'success': False, 'status': 'Download starting.'}, lock=True
         )
         self._max_image_cache_size = max_image_cache_size
         self._max_image_size = max_image_size
-        self._lock = threading.RLock()
 
         self._stop = False
         self._sleep_secs = 10
         self._cleanup_thread = None
 
-        self._load_state()
-
-    def _save_state(self):
-        with self._lock:
-            self._state_committer.commit(self._image_cache)
-
-    def _load_state(self):
-        with self._lock:
-            self._image_cache = self._state_committer.load()
-
     def start(self):
         logger.info("Starting docker image manager")
+
         if self._max_image_cache_size:
+            self._tag_local_images()
 
             def cleanup_loop(self):
                 while not self._stop:
                     try:
                         self._cleanup()
-                        self._save_state()
                     except Exception:
                         traceback.print_exc()
                     time.sleep(self._sleep_secs)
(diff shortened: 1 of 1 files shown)


codalab/worker/docker_image_manager.py
```diff
@@ -151,14 +148,11 @@ def image_availability_state(image_spec, success_message, failure_message):
...
-                    self._image_cache[digest] = ImageCacheEntry(
-                        id=image.id,
-                        digest=digest,
-                        last_used=time.time(),
-                        virtual_size=image.attrs['VirtualSize'],
-                        marginal_size=image.attrs['Size'],
-                    )
+                timestamp = str(time.time())
+                for tag in image.tags:
+                    if tag.split(":")[0] == "codalab-image-cache/last-used":
```
percyliang: Create a function that encapsulates the tag logic to avoid duplication with the same thing above.  The string `"codalab-image-cache/last-used"` should ideally only happen once.

codalab/worker/docker_image_manager.py
```diff
@@ -151,14 +148,11 @@ def image_availability_state(image_spec, success_message, failure_message):
...
+                        self._docker.images.remove(tag)
+                image.tag("codalab-image-cache/last-used", tag=timestamp)
```
percyliang: Should we put this line above the removal, so if something fails, we don't end up with a situation where an image that should be managed by codalab falls off the radar?

percyliang (commented):
Nice!  Could we add a test for this logic?

codalab/worker/docker_image_manager.py
bkgoksel: Good point, I wanted to avoid deleting the new tag, but that's easily avoidable with a string comparison.

codalab/worker/docker_image_manager.py
bkgoksel: ah yeah, I was planning to make it a constant but missed it when copy pasting from my old code.

candicegjing (approved):
LGTM! We just need to make sure when we first deploy this change to prod, we might need to manually tag all the existing docker images on each worker using the `CACHE_TAG` pattern. Then those images should be picked up by the program without any issue.

bkgoksel:
@candicegjing could you test this this a little bit more by deploying this branch on some system, overriding max cache size to be pretty low and making sure it actually deletes images. I don't think we have tests for the cleanup process and it would be bad to catch it only on prod.

candicegjing:
@bkgoksel I have added additional logging messages at line 111 and updated line 120 and 129.

candicegjing:
@bkgoksel Also added `image.reload()` after each tagging operation, as sometimes I noticed the new tag doesn't show up automatically.

codalab/worker/docker_image_manager.py
```diff
@@ -82,6 +74,29 @@ def stop(self):
             self._cleanup_thread.join()
         logger.info("Stopped docker image manager")
 
+    def _get_cache_use(self):
+        return sum(
+            float(image.attrs['VirtualSize']) for image in self._docker.images.list(self.CACHE_TAG)
+        )
+
+    def _tag_local_images(self):
```
candicegjing, line 82: @bkgoksel I added a new function `_tag_local_images` for the very first release. This function will tag all the existing images. Then those images will be picked up in the cleanup phase properly. Otherwise, only newly requested docker images will be considered during the cleanup phase.

codalab/worker/docker_image_manager.py
```diff
@@ -151,20 +176,20 @@ def image_availability_state(image_spec, success_message, failure_message):
...
-                    self._image_cache[digest] = ImageCacheEntry(
-                        id=image.id,
-                        digest=digest,
-                        last_used=time.time(),
-                        virtual_size=image.attrs['VirtualSize'],
-                        marginal_size=image.attrs['Size'],
-                    )
+                new_timestamp = str(time.time())
+                image.tag(self.CACHE_TAG, tag=new_timestamp)
+                image.reload()
```
candicegjing: @bkgoksel I also added a `reload()` function after tagging an image. Sometimes I noticed that the new tag doesn't show up automatically. Reloading the image can help resolve this issue.

codalab/worker/docker_image_manager.py
```diff
@@ -99,33 +113,43 @@ def _cleanup(self):
         """
         while not self._stop:
```
candicegjing: It seems that we have 2 `while not self._stop` here. One is within `_cleanup()` function, another is in `cleanup_loop()`. Do we have keep both of them? Let me know if I missed anything.

codalab/worker/docker_image_manager.py
bkgoksel: I think this makes too strong an assumption that the worker is responsible for all the docker images on its host machine. This makes sense in the public codalab workers, but in a shared tenancy deployment (like in Slurm or AWS Batch) there might be docker images downloaded by other users/system components that we do not care about/should not mess with. Calling this would wrongly make our worker try to prune those images later on.

I think it's better to manually tag existing images on worker update only for this update (or prune all existing images on public workers and start afresh with a new cache once the update is done).

codalab/worker/docker_image_manager.py
bkgoksel: Haha good catch! We don't need the 2 loops and this is likely a bug that may be causing issues. We should get rid of this one (in the `_cleanup` method)

candicegjing: Okay, I removed the inner `while loop` and left the outer `while loop` there.

codalab/worker/docker_image_manager.py
candicegjing: Yes. I agree with you on this. The more that I was thinking is that I think this function can be used to run only during the very first release. Then all the deployment processes can just be automated during the next release. Since we don't need to do this after the first release, once the first release is done, we can remove this function from `start()`. In the future, if there is any change that people need to tag their local images, they can just manually tag their images using this function.

candicegjing: Created a github issue to track this https://github.com/codalab/codalab-worksheets/issues/1949

Issue #1932, opened 2020-01-27
Random failures in CI tests: cl wait


candicegjing:
There are two issues for `cl wait`:
1. Running `cl run main.pl:0xce8fae62584b4716bd7d835de64bfb08 perl main.pl 2 50 10 --request-time 10 --request-memory 200m --request-disk 2m`
<img width="958" alt="Screen Shot 2020-01-27 at 5 14 11 PM" src="https://user-images.githubusercontent.com/1483372/73227472-8723f780-4128-11ea-8424-307da65b877e.png">
2. Running `cl run ping -c 1 google.com --request-network`
<img width="569" alt="Screen Shot 2020-01-27 at 10 18 19 AM" src="https://user-images.githubusercontent.com/1483372/73201658-5e7f0c00-40ee-11ea-981f-fae47c069efe.png">

candicegjing:
Close this for now as Travis CI build seems more stable these days. Will reopen if things get stuck again in the future.

PR #1936, opened 2020-01-28
Fix disk usage check

candicegjing:
Fixed the first issue in #1932

When comparing the disk resources at [`disk_utilization`](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L390), the value returned from [self.disk_utilization](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L386-L388) can sometimes be 0, even if the bundle is finished (`state.finished=True`). Then the `request-disk` constraint becomes invalid. So we see random failures in test_cli.py described in #1932 . Basically the calculating disk utilization in a separate thread here could be inaccurate when the bundle reaches `finished` state but the thread is in sleeping mode so that the main process never gets the final disk usage of the bundle. 
Before this PR, disk utilization calculation is running as a single thread. However, this piece of logic could be simplified by just adding the condition check whenever the bundle is in `RUNNING` state at [self._transition_from_RUNNING](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L307). Then as soon as `disk_utilization` exceeds the upper limit of `request-disk`, then the `run_state` will be set to failure.

### DIFF
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -341,9 +341,6 @@ class RunStateMachine(StateTransitioner):
                     'Time limit exceeded. (Container uptime %s > time limit %s)'
                     % (duration_str(container_time_total), duration_str(run_state.resources.time))
                 )
-            logger.info("max_memory = {}".format(run_state.max_memory))
-            logger.info("run_state.resources.memory = {}".format(run_state.resources.memory))
-            logger.info("run_state.exitcode = {}".format(run_state.exitcode))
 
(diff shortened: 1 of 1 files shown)


percyliang (commented):
The disk utilization is deliberately in a thread because it can be slow and we don't want to block everything else.

candicegjing:
Closing this PR now, as we are going to explore another direction to solve this problem.

PR #1938, opened 2020-01-28
1854: Display a more useful failure message when a run is killed

### LINKED ISSUE #1854: Some bundles fail with an unknown error message
The following bundles fail with the same, unknown error messages and nothing useful outputted to stderr: 

(bash: line 1: 6 Killed)
https://worksheets.codalab.org/bundles/0xa1bf711247d74d9d877a31d38ad6098d
https://worksheets.codalab.org/bundles/0x0d9bd192da0041e2810afcd0afe2e0a1
https://worksheets.codalab.org/bundles/0x5289c96f21bf4149aecf01f5e3976d66
https://worksheets.codalab.org/bundles/0x38159f78ed3f4b8d90cf002ab8ada179

(bash: line 1: 11 Killed)
https://worksheets.codalab.org/bundles/0x68917d35b482458ab01ffb1613a49352

Dev:
https://worksheets-dev.codalab.org/bundles/0x6ebf82c0e9414fcf954a3c77786f1acc
https://worksheets-dev.codalab.org/bundles/0x3e60232e0f2248e18872839dc864dc20
https://worksheets-dev.codalab.org/bundles/0x4ed8bd8b5cac4db284b95bb69126a7a2

teetone:
Fixes #1854 

Display a more useful error message when a run is killed by a worker. Before this fix, the error message was something like "bash: line 1: 6 Killed". After the fix, it displays the reason why the run was killed.

Screen Shot 2020-01-28 at 12 43 03 AM

### DIFF
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -525,7 +525,7 @@ class RunStateMachine(StateTransitioner):
             failure_message = (
                 "{}. {}".format(run_state.failure_message, run_state.kill_message)
                 if run_state.failure_message
-                else run_state.kill_messsage
+                else run_state.kill_message
             )
             run_state = run_state._replace(failure_message=failure_message)
         return run_state._replace(stage=RunStage.FINALIZING, run_status="Finalizing bundle")
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -350,7 +350,7 @@ class RunStateMachine(StateTransitioner):
                     % (duration_str(container_time_total), duration_str(run_state.resources.time))
                 )
 
-            if run_state.max_memory > run_state.resources.memory or run_state.exitcode == '137':
+            if run_state.max_memory > run_state.resources.memory or run_state.exitcode == 137:
                 kill_messages.append(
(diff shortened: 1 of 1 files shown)


codalab/worker/worker_run_state.py
```diff
@@ -521,7 +520,8 @@ def finalize_run(self, run_state):
         """
         Prepare the finalize message to be sent with the next checkin
         """
-        if not run_state.failure_message and run_state.is_killed:
+        if run_state.is_killed:
+            # Set the failure message as the kill_message, which contains useful information on why a run was killed.
```
percyliang: If `failure_message` exists, we should concatenate the kill message to the failure message so we don't lose information.

percyliang (changes requested):

codalab/worker/worker_run_state.py
teetone: @percyliang That's a good idea. Are you okay with something like this:

`/bin/bash: line 1: 7 Killed ( python test.py ) > stdout 2> stderr . Memory limit 1g exceeded.`

percyliang (approved):
Great!

In retropect, we should make `failure_message` a list of messages so we maintain the structure rather than doing string shenanigans...

PR #1930, opened 2020-01-27
Cache parallel run quota

candicegjing:
This PR caches `user_info` to reduce database calls in `get_user_parallel_run_quota_left`:
The following status is collected using 100 staged jobs in staging and `get_user_parallel_run_quota_left` check is forced to run. In the real production environment, the situation might be different. But basically, this change will reduce around 40% of the time in `get_user_parallel_run_quota_left`, especially when a user who submitted a large number of jobs.
```
Sun Jan 26 18:57:48 2020    _schedule_run_bundles_on_workers.profile

         534465 function calls (466277 primitive calls) in 0.861 seconds

   Ordered by: cumulative time
   List reduced from 302 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.001    0.001    0.862    0.862 /opt/codalab-worksheets/codalab/server/bundle_manager.py:310(_schedule_run_bundles_on_workers)
      101    0.004    0.000    0.717    0.007 /opt/codalab-worksheets/codalab/model/bundle_model.py:2334(get_user_parallel_run_quota_left)
      101    0.002    0.000    0.289    0.003 /opt/codalab-worksheets/codalab/model/bundle_model.py:2283(get_user_info)
      203    0.001    0.000    0.258    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/base.py:1974(execute)
      203    0.000    0.000    0.257    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/base.py:846(execute)
      203    0.000    0.000    0.256    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/sql/elements.py:322(_execute_on_connection)
      203    0.001    0.000    0.256    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/base.py:975(_execute_clauseelement)
      203    0.001    0.000    0.200    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/threadlocal.py:93(__exit__)
      203    0.002    0.000    0.185    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/base.py:1061(_execute_context)
      203    0.000    0.000    0.160    0.001 /usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/default.py:449(do_execute)
[...]

PR #1944, opened 2020-01-29
Add missing parameter to get_user_parallel_run_quota_left

candicegjing:
Fix as part of #1930

This issue was discovered after the release of v0.5.7 on production. However, both dev and staging didn't show the problem. Thinking through the dev/staging testing process, when we login as the root user to test out changes, some logic related to the regular users might not be exercised (e.g. `get_user_parallel_run_quota_left`). Then the testing will not be sufficient enough to discover all the potential bugs. Need to use non-root user for testing in the future.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -2332,8 +2332,9 @@ class BundleModel(object):
         time_used = user_info['time_used']
         return time_quota - time_used
 
-    def get_user_parallel_run_quota_left(self, user_id):
-        user_info = self.get_user_info(user_id)
+    def get_user_parallel_run_quota_left(self, user_id, user_info=None):
+        if not user_info:
+            user_info = self.get_user_info(user_id)
         parallel_run_quota = user_info['parallel_run_quota']
         with self.engine.begin() as connection:
             # Get all the runs belonging to this user whose workers are not personal workers


codalab/model/bundle_model.py
```diff
@@ -2332,8 +2332,9 @@ def get_user_time_quota_left(self, user_id, user_info=None):
         time_used = user_info['time_used']
         return time_quota - time_used
 
-    def get_user_parallel_run_quota_left(self, user_id):
-        user_info = self.get_user_info(user_id)
+    def get_user_parallel_run_quota_left(self, user_id, user_info=None):
```
yipenghe, line 2335: Won't ` if not user_info` always be true since nowhere is passing the argument?

candicegjing: Yes. That's the case when we want to make a database query.

yipenghe: Then how is it different from the original code? Like isn't the change
`
call()
`
to
`
if true:
    call()
`
?
Am I missing something?

candicegjing: The point is to make less database query so that its performance will be improved.

candicegjing: The parameter is passing from bundle_manager.py

yipenghe: But this PR doesn't change bundle_manager right🧐

PR #1945, opened 2020-01-29
Fix chrome uploads by triggering uploads with a label

### LINKED ISSUE #1799: Can't upload files on Chrome
**Describe the bug**
the request for uploading files doesn't work.

**Codalab Information**
- *Your username*: wangfan
- *UUID of worksheet where problem occurred*: 
 - *OS (if on CLI)*:
 - *Browser (if on Web)*;

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

epicfaace:
Fix uploads on Chrome by changing how uploads work. Fixes #1799 

Now, the `NewUpload` component is always rendered a single time, and clicking on the "Upload" button actually clicks a "label" element that then triggers the `<input type="file">` component to upload.

Previously, the issue was that the `handleFocusBack` function was called even when a file was selected for upload, thus causing the component to call `onClose` before the file could be uploaded. Additionally, Chrome does not call an `onchange` event when the value of an `<input type="file">` component does not change (if the same file is selected, or "Cancel" is selected when the input already has no value) -- this caused some race cases with the other logic. This PR simplifies a lot of the upload logic and should cause less issues than the previous approach.

### DIFF
--- a/frontend/src/components/worksheets/WorksheetItemList.js
+++ b/frontend/src/components/worksheets/WorksheetItemList.js
@@ -2,7 +2,7 @@ import * as React from 'react';
 import Immutable from 'seamless-immutable';
 import $ from 'jquery';
 import * as Mousetrap from '../../util/ws_mousetrap_fork';
-import { buildTerminalCommand } from '../../util/worksheet_utils';
+import { buildTerminalCommand, getMinMaxKeys } from '../../util/worksheet_utils';
 import { ContextMenuEnum, ContextMenuMixin } from './ContextMenu';
 import ContentsItem from './items/ContentsItem';
 import GraphItem from './items/GraphItem';
@@ -12,6 +12,7 @@ import RecordItem from './items/RecordItem';
 import TableItem from './items/TableItem';
 import WorksheetItem from './items/WorksheetItem';
 import ItemWrapper from './items/ItemWrapper';
+import NewUpload from './NewUpload/NewUpload';
 
 ////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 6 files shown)


yipenghe:
Can you deploy on dev so I can test, I don't use windows for development so can't build locally there. I can review the code still.

@candicegjing @Levilian can you test out Mac chrome?

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -251,7 +257,22 @@ class WorksheetItemList extends React.Component {
                     );
                 }.bind(this),
             );
-            items_display = worksheet_items;
+            items_display = (
+                <>
+                    {worksheet_items}
+                    <NewUpload
+                        key={this.state.newUploadKey}
```
Levilian, line 264: While changing the key is a very smart way to deal with the rerendering issue, I wonder if there's a more readable way to deal with issues like this.

Levilian (approved):
One minor thing is this comment I left about resetting the key when you upload the same file. Everything else is pretty clear and has simplified the logic of the code a lot. Thanks Ashwin!

Is there a better way to deal with this? 
https://github.com/valor-software/ng2-file-upload/issues/220

frontend/src/components/worksheets/WorksheetItemList.js
epicfaace: I believe the other way you linked to recommends setting `el.target.value` to an empty string. It seems like a hack and not documented behavior (https://stackoverflow.com/questions/20549241/how-to-reset-input-type-file) so I think it's better to change the key just to be safe, so that the entire elemnt re-renders.

PR #1946, opened 2020-01-29
Add staged_status and filter out unqualified bundles before scheduling

### LINKED ISSUE #1800: [Bundle Manager] Filter out unqualified staged bundles and updated their staged_status in UI

candicegjing:
Original PR #1801 . Fixed #1800 
1. Added a new metadata field called `staged_status`
2. Filtered out those bundles which are requested to run on a private worker that hasn't been connected to the server yet. (Currently only considering bad `request-queue`, will add filters for invalid `cpus`/`gpus`/`memory` in another PR)
3. Added metadata delete operation in `update_bundles`

Testing is done locally in the following steps:
1. Start local development instances and stop the worker container.
2. Submit a bundle which is requested to run on a worker with `--request-queue a`
3. Go to the bundle's web page, verify that the bundle is staged and `staged_status` shows the correct information
4. Attach my local machine as a private worker with `tag=a`
5. Verify that the previously staged job is able to proceed and the `staged_status` column is removed from metadata list.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -988,10 +988,9 @@ class BundleModel(object):
 
     def update_bundle(self, bundle, update, connection=None):
         """
-        Update a bundle's columns and metadata in the database and in memory.
-        The update is done as a diff: columns that do not appear in the update dict
-        and metadata keys that do not appear in the metadata sub-dict are unaffected.
-
+        For each key-value pair in the update dictionary, add or update key-value pair. Note
+        that metadata keys not in the update dictionary are not affected in the update operation.
+        Also, delete any metadata key-value pairs when the value specified is None.
         This method validates all updates to the bundle, so it is appropriate
         to use this method to update bundles based on user input (eg: cl edit).
         """
@@ -1000,31 +999,50 @@ class BundleModel(object):
         # Apply the column and metadata updates in memory and validate the result.
         metadata_update = update.pop('metadata', {})
         bundle.update_in_memory(update)
-        for (key, value) in metadata_update.items():
-            bundle.metadata.set_metadata_key(key, value)
+
+        # Generate a list of metadata keys that will be deleted and udpate metadata key-value pair
+        metadata_delete_keys = []
+        for key, value in metadata_update.items():
+            if value is None:
+                bundle.metadata.remove_metadata_key(key)
+                metadata_delete_keys.append(key)
+            else:
+                bundle.metadata.set_metadata_key(key, value)
+
+        # Delete metadata keys from metadata_update dictionary
+        for key in metadata_delete_keys:
+            del metadata_update[key]
+
         bundle.validate()
         # Construct clauses and update lists for updating certain bundle columns.
         if update:
             clause = cl_bundle.c.uuid == bundle.uuid
         if metadata_update:
-            metadata_clause = and_(
+            metadata_update_clause = and_(
                 cl_bundle_metadata.c.bundle_uuid == bundle.uuid,
                 cl_bundle_metadata.c.metadata_key.in_(metadata_update),
             )
-            metadata_values = [
+            metadata_update_values = [
                 row_dict
                 for row_dict in bundle.to_dict().pop('metadata')
                 if row_dict['metadata_key'] in metadata_update
             ]
+        if metadata_delete_keys:
+            metadata_delete_clause = and_(
+                cl_bundle_metadata.c.bundle_uuid == bundle.uuid,
+                cl_bundle_metadata.c.metadata_key.in_(metadata_delete_keys),
+            )
 
-        # Perform the actual updates.
+        # Perform the actual updates and deletes.
         def do_update(connection):
             try:
                 if update:
                     connection.execute(cl_bundle.update().where(clause).values(update))
                 if metadata_update:
-                    connection.execute(cl_bundle_metadata.delete().where(metadata_clause))
-                    self.do_multirow_insert(connection, cl_bundle_metadata, metadata_values)
+                    connection.execute(cl_bundle_metadata.delete().where(metadata_update_clause))
+                    self.do_multirow_insert(connection, cl_bundle_metadata, metadata_update_values)
+                if metadata_delete_keys:
+                    connection.execute(cl_bundle_metadata.delete().where(metadata_delete_clause))
             except UnicodeError:
                 raise UsageError("Invalid character detected; use ascii characters only.")
 
--- a/codalab/objects/metadata.py
+++ b/codalab/objects/metadata.py
@@ -56,6 +56,11 @@ class Metadata(object):
         self._metadata_keys.add(key)
         setattr(self, key, value)
 
+    def remove_metadata_key(self, key):
+        if key in self._metadata_keys:
+            self._metadata_keys.remove(key)
+            delattr(self, key)
+
     @classmethod
     def collapse_dicts(cls, metadata_specs, rows):
         '''
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -10,7 +10,7 @@ import time
 import traceback
 
 from codalab.objects.permission import check_bundles_have_read_permission
-from codalab.common import PermissionError, NotFoundError
+from codalab.common import PermissionError
 from codalab.lib import bundle_util, formatting, path_util
 from codalab.server.worker_info_accessor import WorkerInfoAccessor
 from codalab.worker.file_util import remove_path
@@ -51,6 +51,8 @@ class BundleManager(object):
 
         self._make_uuids_lock = threading.Lock()
         self._make_uuids = set()
+        # Set of bundle UUIDs with an unknown requested worker
+        self._bundles_without_matched_workers = set()
 
         def parse(to_value, field):
             return to_value(config[field]) if field in config else None
@@ -356,7 +358,9 @@ class BundleManager(object):
                         )
                         # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
                         continue
-            if not workers_list: # Length is 0 (private user with no workers) or is None (root user)
+            if (
+                not workers_list
+            ):  # Length is 0 (private user with no workers) or is None (root user)
                 workers_list = get_available_workers(self._model.root_user_id)
 
             # Try starting bundles on the workers that have enough computing resources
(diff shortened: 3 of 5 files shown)


epicfaace:
What is an unqualified bundle?

candicegjing:
> What is an unqualified bundle?

This PR addresses those bundles which are requested to run on a private worker that hasn't been connected to Codalab server yet. Basically, we don't want to schedule those bundles to run when their requested workers are not present in the system. This PR is to filter them out before the real dispatching algorithm.

codalab/model/bundle_model.py
```diff
@@ -986,11 +986,13 @@ def save_bundle(self, bundle):
...
 
-    def update_bundle(self, bundle, update, connection=None):
+    def update_bundle(self, bundle, update, connection=None, delete=False):
         """
-        Update a bundle's columns and metadata in the database and in memory.
-        The update is done as a diff: columns that do not appear in the update dict
+        Update a bundle's columns and metadata (default) or delete a bundle's metadata in the database and in memory.
+        1. The update is done as a diff: columns that do not appear in the update dict
         and metadata keys that do not appear in the metadata sub-dict are unaffected.
+        2. The delete is done as a diff on metadata dict only:
```
percyliang: I think using a `delete` flag to signal that only the keys in `update` to be used is confusing; usually, you want arguments to be as independent as possible, and here the semantics of the arguments are deeply intertwined.

One option is just to assume that keys with `None` values in `update` are deleted.  I *think* this should be safe, but you should double check if a null value is ever needed in the database apart from not existing.  Then you don't need `delete` at all as a flag.

codalab/objects/metadata.py
```diff
@@ -56,6 +56,10 @@ def set_metadata_key(self, key, value):
         self._metadata_keys.add(key)
         setattr(self, key, value)
 
+    def remove_metadata_key(self, key):
+        if key in self._metadata_keys:
+            self._metadata_keys.remove(key)
```
percyliang, line 61: Don't you need to do `delattr` (mirroring what `set_metadata_key` does)?

codalab/server/bundle_manager.py
```diff
@@ -50,6 +52,8 @@ def __init__(self, codalab_manager):
 
         self._make_uuids_lock = threading.Lock()
         self._make_uuids = set()
+        # A set of UUID of bundles without their requested private worker discovered in the system
```
percyliang: "Set of bundle UUIDs with an unknown requested worker" (more general than just private worker, discovered in the system)

codalab/server/bundle_manager.py
```diff
@@ -660,14 +658,28 @@ def _check_resource_failure(
                 return global_fail_string % (pretty_print(value), pretty_print(global_min))
         return None
 
-    def _get_staged_bundles_to_run(self, user_info_cache):
+    def _get_matched_workers(self, request_queue, workers):
```
percyliang: this can be a static function

percyliang (changes requested):
I think you don't need specific logic for private workers.  If worker requests *any* worker that isn't in the system, then we should print out the appropriate message that says whatever worker was being requested is not available.  This is more general and requires fewer concepts (i.e., private/public) in the code.

codalab/model/bundle_model.py
candicegjing: Yeah, I also thought about this several times during the development and I do feel that adding `delete=True` is somehow confusing and redundant in the way it looks like right now. Originally I was also wondering if I should directly set the specific field that I want to delete to be `None`, then using this `None` value as an indicator of deletion operation. However, what if a user originally set `request-queue=a`, later for some reason he wants to run his jobs on CodaLab's public workers, then he removes the corresponding value of `request-queue` from UI, meaning the `request-queue` now becomes `None`. With our deletion logic, the metadata key `request-queue` will be removed out from the metadata list and not shows up in UI eventually. Based on the definition of `request-queue` specified [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/bundles/run_bundle.py#L32), it is default set to `None`, which is confusing as well if we use it as an indicator of deletion operation.

codalab/objects/metadata.py
candicegjing: Good catch! Updated.

codalab/model/bundle_model.py
candicegjing: I guess the function naming is a bit confusing as well, especially when describing a database operation. There are three types of operations when making changes to a table: `insert`, `update` and `delete`. We here merged `insert` with `update` as a single function. I think splitting these three operations in a certain way could definitely help improving the readability for this piece of code. 
One potential way could be to include three operations in the input parameters? e.g. `def change_bundle(self, bundle, update=None, delete=None, insert=None,...)` Or something equivalent to switch case statement for those operations. Let me know how you think about this.

candicegjing:
@percyliang This PR is ready for another review. For the flag which can be used to indicate delete operation, `None` is used as the default value for several fields in the run_bundle table, e.g. [request-queue](https://github.com/codalab/codalab-worksheets/blob/master/codalab/bundles/run_bundle.py#L32). So currently I changed it to string **"DELETE"** so that we can avoid using an additional parameter for this function. Let me know if you have any thoughts about this.

codalab/model/bundle_model.py
```diff
@@ -1000,22 +1002,37 @@ def update_bundle(self, bundle, update, connection=None):
...
         metadata_update = update.pop('metadata', {})
         bundle.update_in_memory(update)
+
+        # Currently we only allow deleting fields in metadata dict using "DELETE" as the operation indicator.
+        # When the value of a metadata key is "DELETE", delete the key, value pair from the bundle.metadata dict
+        # Otherwise, update the (key, value) pair in the bundle.metadata dict
+        delete = False
         for (key, value) in metadata_update.items():
-            bundle.metadata.set_metadata_key(key, value)
+            if value == 'DELETE':
```
percyliang: Don't use 'DELETE' because the value could be 'DELETE'.  Either use `None` or have a separate parameter passed into this function which is just the set of keys to delete.

codalab/model/bundle_model.py
```diff
@@ -988,9 +988,11 @@ def save_bundle(self, bundle):
 
     def update_bundle(self, bundle, update, connection=None):
         """
-        Update a bundle's columns and metadata in the database and in memory.
-        The update is done as a diff: columns that do not appear in the update dict
+        Update a bundle's columns and metadata (default) or delete a bundle's metadata in the database and in memory.
```
percyliang: Improve the docstring: Say: for each key: value in `update`, add metadata pair (key, value), replacing any old value.  Note that metadata keys not in `update` are not affected.  Also, delete any metadata entries corresponding to keys `delete_keys`.  [See comment below]

percyliang (changes requested):

codalab/model/bundle_model.py
candicegjing: @percyliang I chose to update the function parameter to have an additional input parameter: `delete_keys`, as `None` is used as a default value for those fields, e.g.[request-queue](https://github.com/codalab/codalab-worksheets/blob/master/codalab/bundles/run_bundle.py#L32) which could be confusing if using `None` to indicate the delete operation. The code and updated and tested against several test cases.

codalab/model/bundle_model.py
```diff
@@ -986,11 +986,15 @@ def save_bundle(self, bundle):
             self.do_multirow_insert(connection, cl_bundle_metadata, metadata_values)
             bundle.id = result.lastrowid
 
-    def update_bundle(self, bundle, update, connection=None):
+    def update_bundle(self, bundle, update={}, connection=None, delete_keys=[]):
```
percyliang: put `delete_keys` before `connection`

codalab/model/bundle_model.py
```diff
@@ -986,11 +986,15 @@ def save_bundle(self, bundle):
...
         """
-        Update a bundle's columns and metadata in the database and in memory.
-        The update is done as a diff: columns that do not appear in the update dict
-        and metadata keys that do not appear in the metadata sub-dict are unaffected.
+        For each (key, value) pair in update dictionary, add or update metadata pair (key, value).
+        Note that metadata keys not in the update dictionary are not affected. Also, delete any
```
percyliang: This is not quite true due to `delete_keys`.

codalab/model/bundle_model.py
```diff
@@ -986,11 +986,15 @@ def save_bundle(self, bundle):
             self.do_multirow_insert(connection, cl_bundle_metadata, metadata_values)
             bundle.id = result.lastrowid
 
-    def update_bundle(self, bundle, update, connection=None):
+    def update_bundle(self, bundle, update={}, connection=None, delete_keys=[]):
         """
-        Update a bundle's columns and metadata in the database and in memory.
-        The update is done as a diff: columns that do not appear in the update dict
-        and metadata keys that do not appear in the metadata sub-dict are unaffected.
+        For each (key, value) pair in update dictionary, add or update metadata pair (key, value).
```
percyliang: Actually, just realized that `update` contains "metadata".  Make this more precise.  Then `delete_metadata_keys` is more precise.

percyliang (changes requested):

codalab/model/bundle_model.py
percyliang: Question: what goes wrong if we use `None` (which is a bit slicker)?  Is there a reason we need to distinguish between deleting and setting a metadata entry with a `None`?

Are there currently metadata entries in the database with a `None` value?

codalab/model/bundle_model.py
candicegjing: It's updated!

codalab/model/bundle_model.py
candicegjing: @percyliang I updated the order of the input parameters and all other places that could be affected.

codalab/model/bundle_model.py
```diff
@@ -988,9 +988,13 @@ def save_bundle(self, bundle):
 
     def update_bundle(self, bundle, update, connection=None):
         """
-        Update a bundle's columns and metadata in the database and in memory.
-        The update is done as a diff: columns that do not appear in the update dict
-        and metadata keys that do not appear in the metadata sub-dict are unaffected.
+        For each key-value pair in the update dictionary, add or update key-value pair. Note
+        that metadata keys not in the update dictionary are not affected in the update operation.
+        Also, delete any metadata key-value pairs when the value specified is None.
+        1. The update is done as a diff: columns that do not appear in the update dictionary
```
percyliang: 1. and 2. seem redundant - delete.

percyliang (approved):

candicegjing:
Closed via #1800

PR #1947, opened 2020-01-30
1891: Print warning when CodaLab is not installed in $PATH

teetone:
For #1891 

Print a warning when CodaLab is not installed in $PATH via pip. Below is an example output of the warning message:

output

Updated doc:
Screen Shot 2020-02-03 at 11 33 04 PM

--
I have looked into this issue for awhile and I believe this is the best solution. As a Python package, we have no control on how the user installs CodaLab. `pip install codalab` can drop the executable not in the PATH (can happen in both Python2 and Python3). Additionally, the `--user` option will install CodaLab on the user's path, which can vary. There is a pip issue that talks more about `--user` https://github.com/pypa/pip/issues/3813 in depth. The corresponding fix for it (https://github.com/pypa/pip/pull/4553) is similar to this one.

A couple of alternative solutions were:
1. Update our docs and let the users know that they might run into `command not found error` and show them how to resolve  this problem. I can also make the doc change in this PR.
2. Take the path where CodaLab is installed and include in the user's PATH by adding it to a `~/.bashrc` for example. However, this change is a bit invasive as we are modifying the user's environment for them.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ from setuptools.command.install import install
 
 import os
 import setuptools
+import sys
 
 
 # Should match codalab/common.py#CODALAB_VERSION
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,15 +1,32 @@
...
 
-if int(setuptools.__version__.split('.')[0]) < 25:
-    print(
-        "WARNING: Please upgrade setuptools to a newer version, otherwise installation may break. "
-        "Recommended command: `pip3 install -U setuptools`"
+
+class Install(install):
+    _WARNING_TEMPLATE = (
+        '\n\n\033[1m\033[93mWarning! CodaLab was installed at {}, which is not\n'
+        'one of the following paths in PATH:\n\n{}\n\nConsider adding {} to be in the path in order\n'
```
percyliang: Rewrite: Warning! CodaLab was installed at {}, which is not one of the following paths in $PATH.  Consider adding {} to $PATH to use the CodaLab CLI.  You can do this by appending the following line to your `.bashrc`: {}

percyliang (approved):

candicegjing (approved):

setup.py
teetone: @percyliang I'm not sure if can say the last sentence. I think the process is different for different OS like Windows. Should I add an entry to the FAQ of our doc? I think I can be more explicit there and give examples on how to fix the path issue in unix systems and Windows.

percyliang: If you're slick, maybe you can detect the OS can give the appropriate message.

teetone:
@percyliang: I added additional logic to check for the user's OS.

Some bad news is that later versions of pip will actually swallow print statements made out to both stdout and stderr (see https://github.com/pypa/pip/issues/2933). That means the users will only see the warning messages in `setup.py` if they added the `--verbose` option when running `pip install`.

The good news is the enhancement made by pip (see https://github.com/pypa/pip/pull/4553) to warn users when the executable is not on their PATH will always show. I added additional troubleshooting instructions to the FAQ section of our doc.

percyliang (approved):

teetone:
Tested with https://test.pypi.org/project/codalab/

Issue #1906, opened 2020-01-16
Worker went offline due to image pruning failure


candicegjing:
Production GPU3 went offline when running https://worksheets.codalab.org/bundles/0x91ea4c751d424cdeab9706911d060f32 .
System docker containers' running status is listed below:
```
vm-clws-prod-gpuworker-3:~% docker ps -a
CONTAINER ID        IMAGE                                   COMMAND                  CREATED             STATUS                     PORTS               NAMES
6b4bb73e70d1        sellpy/python3-jupyter-sklearn:latest   "bash -c '( python3 …"   34 hours ago        Exited (255) 2 hours ago                       codalab_run_0x91ea4c751d424cdeab9706911d060f32
89e5736f0d57        codalab/worker:v0.5.4                   "cl-worker --server …"   7 days ago          Exited (255) 2 hours ago                       codalab_worker_1
```
In `codalab_run_0x91ea4c751d424cdeab9706911d060f32`: no error message reported from docker.
In `codalab_worker_1`:
```
2020-01-15 16:19:19,481 http://localhost:None "GET /v1.35/containers/a6ea55bbf5dfac8d856ce3e0a3aa5da78c7905d5d7f079102174eeb31043a562/logs?stderr=1&stdout=0&timestamps=0&follow=0&tail=all HTTP/1.1" 200 None
2020-01-15 16:19:19,484 http://localhost:None "DELETE /v1.35/images/sha256:fc78c9a627043147baba782807c3a5250a9b876c98b71878b5c286a42d9d27c0?force=False&noprune=False HTTP/1.1" 409 127
2020-01-15 16:19:19,484 Cannot remove image floydhub/pytorch@sha256:c43afff6a9d7e817fb40dd517ef1e6ec0bcec3d0f20de5c0285a84ab3c89199b from cache: 409 Client Error: Conflict ("conflict: unable to delete fc78c9a
2020-01-15 16:19:19,484 Disk use (32512759562) > max cache size (32212254720), pruning image: floydhub/pytorch@sha256:c43afff6a9d7e817fb40dd517ef1e6ec0bcec3d0f20de5c0285a84ab3c89199b
2020-01-15 16:19:19,485 http://localhost:None "GET /v1.35/containers/a6ea55bbf5dfac8d856ce3e0a3aa5da78c7905d5d7f079102174eeb31043a562/json HTTP/1.1" 200 None
[...]

PR #1952, opened 2020-01-31
Improve docker image cleanup function

### LINKED ISSUE #1948: Differentiate "cannot be forced" and "must be forced" in docker image cleanup
There are three conditions that we can't remove docker images:
1. When an image is using by a running container:
```
2020-01-30 02:19:39,132 Cannot remove image codalab/worker:docker_tag_cache from cache: 409 Client Error: Conflict ("conflict: unable to delete f5d99f693ec3 (cannot be forced) - image is being used by running container fc4da58466cb")
```
2. When an image has dependents:
```
2020-01-30 23:50:50,422 Cannot remove image codalab_cache/last_used:123 from cache: 409 Client Error: Conflict ("conflict: unable to delete d5680e53a228 (cannot be forced) - image has dependent child images")
```
3. When an image is referenced in multiple repositories:
```
Error response from daemon: conflict: unable to delete 0e21a45a5660 (must be forced) - image is referenced in multiple repositories
```
We definitely don't want to remove images in 1 and 2 conditions, but images in 3rd condition can be removed technically. Need to find a way (either parsing the error message or other better ways) to differentiate those. Otherwise, we can only end up cleaning images that have either no dependencies or no references, which can be not sufficient for us to keep the system clean.

candicegjing:
Fixed #1948 and #1906.

This PR will improve the docker image cleanup mechanism. Basically, the only docker images that can be removed without the `-force` option are the ones that either don't tie to a running container or don't have any references to other repositories. Some of our images do have dependencies with others. Without forcibly removing those docker images on a host, the space that can be reclaimed is limited. Based on findings in #1948, there are two types of `409 Client Error` when removing an image. One is **must be forced**. Another is **cannot be forced**. The later one happens to those images that tie to a running container or have child dependencies that can't be removed. The former one is the one that we can actually try to clean up by using `-force` option. This PR is to add this option when encountering a `409 Client Error` to further clean up the docker images.
For instance, the following logging message copied from `m-clws-prod-worker-0` shows the situation described above (it's logged after the new release of `v0.5.8`):
```
2020-02-03 16:29:10,179 http://localhost:None "GET /v1.35/images/sha256:a307a5e9528a14dce69881257a807bd35fd89daa17f7803ab3c94bb48315b05f/json HTTP/1.1" 200 None
2020-02-03 16:29:10,181 http://localhost:None "GET /v1.35/images/sha256:aeff5a9860a391e0313f68c37ed368e524f53d6ecb6be0f2b70ef3922ebcda28/json HTTP/1.1" 200 None
2020-02-03 16:29:10,181 Disk use (56679073288.0) > max cache size (32212254720), pruning image: large64/docker-test:ruby
2020-02-03 16:29:10,184 http://localhost:None "DELETE /v1.35/images/sha256:e55c53059ea922fbded27082009c3d304d5890e2bb44a0680f319c5aa0efaf61?force=False&noprune=False HTTP/1.1" 409 118
2020-02-03 16:29:10,184 Cannot remove image large64/docker-test:ruby from cache: 409 Client Error: Conflict ("conflict: unable to delete e55c53059ea9 (must be forced) - image is referenced in multiple repositories")
2020-02-03 16:29:10,194 http://localhost:None "GET /v1.35/images/json?filter=codalab-image-cache%2Flast-used&only_ids=0&all=0 HTTP/1.1" 200 None
2020-02-03 16:29:10,196 http://localhost:None "GET /v1.35/images/sha256:45dccfd526b0b9df7a2f3033a3701f944e0a650091b17913988fe6277325194d/json HTTP/1.1" 200 None
```

### DIFF
--- a/codalab/worker/docker_image_manager.py
+++ b/codalab/worker/docker_image_manager.py
@@ -131,23 +131,34 @@ class DockerImageManager:
                 # We re-list all the images to get an updated total size since we may have deleted some
                 cache_use = self._get_cache_use()
                 if cache_use > self._max_image_cache_size:
+                    image_tag = (
+                        image.attrs['RepoTags'][-1]
+                        if len(image.attrs['RepoTags']) > 0
+                        else '<none>'
+                    )
                     logger.info(
                         'Disk use (%s) > max cache size (%s), pruning image: %s',
                         cache_use,
                         self._max_image_cache_size,
-                        image.attrs['RepoTags'][1],
+                        image_tag,
                     )
                     try:
-                        self._docker.images.remove(image.id)
+                        self._docker.images.remove(image.id, force=True)
                     except docker.errors.APIError as err:
-                        # Maybe we can't delete this image because its container is still running
-                        # (think a run that takes 4 days so this is the oldest image but still in use)
-                        # In that case we just continue with our lives, hoping it will get deleted once
-                        # it's no longer in use and the cache becomes full again
+                        # Two types of 409 Client Error can be thrown here:
+                        # 1. 409 Client Error: Conflict ("conflict: unable to delete <image_id> (cannot be forced)")
+                        #   This happens when an image either has a running container or has multiple child dependents.
+                        # 2. 409 Client Error: Conflict ("conflict: unable to delete <image_id> (must be forced)")
+                        #   This happens when an image is referenced in multiple repositories.
+                        # We can only remove images in 2rd case using force=True, but not the 1st case. So after we
+                        # try to remove the image using force=True, if it failed, then this indicates that we were
+                        # trying to remove images in 1st case. Since we can't do much for images in 1st case, we
+                        # just continue with our lives, hoping it will get deleted once it's no longer in use and
+                        # the cache becomes full again
                         logger.error(
-                            "Cannot remove image %s from cache: %s", image.attrs['RepoTags'][1], err
+                            "Cannot forcibly remove image %s from cache: %s", image_tag, err
                         )
-        logger.debug("Stopping docker image manager cleanup")
+            logger.debug("Stopping docker image manager cleanup")
 
     def get(self, image_spec):
         """


codalab/worker/docker_image_manager.py
```diff
@@ -22,6 +22,7 @@
 ImageCacheEntry = namedtuple(
     'ImageCacheEntry', ['id', 'digest', 'last_used', 'virtual_size', 'marginal_size']
 )
+DOCKER_ERROR_SUBSTRING = "(must be forced)"
```
bkgoksel: Since we use this constant only once, I would actually not store this as a constant variable but directly compare against the string literal. That way the string is at the same place as the comments that explain what is going on. I'm afraid if it's like this this constant would be confusing to someone looking at this codebase for the first time.

bkgoksel (commented):
Thanks for catching this! 

My main questions is: why can't we just make the initial `remove` call with `force=True`? What happens if we call delete with `force=True` in the first error case (cannot be forced)?

If trying to force in that case still throws an exception but doesn't break anything, then we can always try to force and when it throws, we just log and move on, rather than doing the string check, which is a bit hacky/brittle IMO (what if docker API changes the wording of the error message etc)

codalab/worker/docker_image_manager.py
candicegjing: Yes. I feel this is quite hacky as well. I'll remove this and compare it directly with the error message.

candicegjing:
> Thanks for catching this!
> 
> My main questions is: why can't we just make the initial `remove` call with `force=True`? What happens if we call delete with `force=True` in the first error case (cannot be forced)?
> 
> If trying to force in that case still throws an exception but doesn't break anything, then we can always try to force and when it throws, we just log and move on, rather than doing the string check, which is a bit hacky/brittle IMO (what if docker API changes the wording of the error message etc)

Originally I thought always forcibly removing docker image might be too dangerous, so I kept the first remove logic as it is. However, I just ran some checks based on your suggestions, it seems that if we add `force=True` to those images that tie to running containers or have dependent child images, it still returns us `cannot be forced`. If this is true then it's better to use `force=True` in the first removing logic and remove the later ones.

codalab/worker/docker_image_manager.py
```diff
@@ -131,23 +131,34 @@ def last_used(image):
...
+                        # try to remove the image using force=True, if it failed, then this indicates that we were
+                        # trying to remove images in 1st case. Since we can't do much for images in 1st case, we
+                        # just continue with our lives, hoping it will get deleted once it's no longer in use and
+                        # the cache becomes full again
                         logger.error(
-                            "Cannot remove image %s from cache: %s", image.attrs['RepoTags'][1], err
+                            "Cannot forcibly remove image %s from cache: %s", image_tag, err
                         )
-        logger.debug("Stopping docker image manager cleanup")
+            logger.debug("Stopping docker image manager cleanup")
```
bkgoksel, line 161: Did you mean to indent this?

bkgoksel (approved):
Looks great!

codalab/worker/docker_image_manager.py
candicegjing: Yeah, otherwise if the `cleanup` thread is running in the background, this logging message will show up all the time.

PR #1954, opened 2020-01-31
Add codalab-worker-scratch to gitignore

candicegjing:
Ignore `codalab-worker-scratch` in git commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ venv*/
 /frontend/node_modules
 pip-wheel-metadata
 site
+
+# Temporary bundle data
+codalab-worker-scratch


.gitignore
```diff
@@ -36,3 +36,6 @@ venv*/
 /frontend/node_modules
 pip-wheel-metadata
 site
+
+# Temporary bundle data
+codalab-worker-scratch
```
teetone, line 41: Can you explain what this bundle data is used for?

teetone (approved):

.gitignore
candicegjing: https://github.com/codalab/codalab-worksheets/blob/cf515131f68489ec66cdadd47a15c5ccbf7bf0da/codalab/worker/main.py#L38

percyliang (approved):

PR #1961, opened 2020-02-04
Increase shared memory to 500mb

epicfaace:


Levilian:
The bundle that caused the error is here: https://worksheets.codalab.org/bundles/0xbbd4b64ea7d549939e5f8807e24cbc9f

The error message is as follows:
`ERROR: Unexpected bus error encountered in worker. This might be caused by insufficient shared memory (shm).`

Issue #1911, opened 2020-01-18
CPU/GPU reporting issue


candicegjing:
The highlighted line seems to have unreasonable number of CPUs. Might need some investigation to fix this.
<img width="1074" alt="Screen Shot 2020-01-18 at 11 09 57 AM" src="https://user-images.githubusercontent.com/1483372/72669132-2c312880-39e3-11ea-9c1e-938072108681.png">

candicegjing:
Based on [this piece of code](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L338-L339): `cpuset` setting will always return the first few cores, meaning jobs will only be allowed to execute on the first few cores. For example, assuming available CPUs are `{0,1,2,3}`, if a user doesn't specify the `request-cpu` option, the number of CPUs to be allocated will be `1`, which leads to the returning `cpuset` to be `{0}` all the time. This doesn't seem to be correct, as we might want to fully utilize all cores at any time?

candicegjing:
For GPU reporting [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/workers.py#L153), I might need additional parameters to determine the number of GPUs in use. Currently, we only have 1 GPUs per GPU worker which could be easy to handle. But with GPU workers containing more than 1 GPU, we can't just sum all requested GPUs up, as GPU's architecture is different than CPU's

percyliang:
We should never be sharing GPUs across jobs.

percyliang:
Try to reproduce this on dev first, possibly with a private worker.

PR #1962, opened 2020-02-05
Modify job assignment algorithm

### LINKED ISSUE #1953: Job assignment sometimes doesn't work well
Sometimes there are two jobs that are assigned to the same worker, and it shows cpu7/6 or 2/1 gpu, while there are workers that are idling. Since it requires root access to see this, ping me for screenshots if unclear

candicegjing:
Fixed #1953 

This PR is to improve job assignment logic in `bundle_manager.py`. Based on the original ticket #1953, jobs are usually assigned to single workers but with many other workers in idle state. This PR has the following modifications:
1. Added the number of running jobs on each worker as one of the sorting keys to sort existing workers. 
2. `gpu_priority` can return `True`/`False` or a digit based on current logic. Since the returning value isn't quite intuitive, especially considering sorting some `Integer` values with `boolean` values, so I replaced this piece of logic to have only `Integer` value returned.
3. Adjusted the order of sorting keys: `gpu_priority` > number of running jobs > number of available dependencies > number of cpus >  the random seeds. (Open to discuss this order)

Here are some test results I did locally:
Assuming we have the following worker items in the `workers_list`:
```
workers_list = [
                {'worker_id': 6, 'gpus': 1, 'dep': 8, 'cpus':6, 'runs':[1,2,3]},
                {'worker_id': 5, 'gpus': 1, 'dep': 0, 'cpus':6, 'runs':[]},
                {'worker_id': 4, 'gpus': 1, 'dep': 8, 'cpus':6, 'runs':[1,2,3]},
                {'worker_id': 3, 'gpus': 1, 'dep': 2, 'cpus':6, 'runs':[1]},
                {'worker_id': 2, 'gpus': 1, 'dep': 4, 'cpus':6, 'runs':[1,2,3,4]},
                {'worker_id': 1, 'gpus': 0, 'dep': 5, 'cpus':8, 'runs':[1,2,3,4,5,6]},
                {'worker_id': 0, 'gpus': 0, 'dep': 7, 'cpus':8, 'runs':[]},]
```
1. When the current bundle requested for `0` gpus, sorting results become: 
```
stanford@Stanfords-MacBook-Pro competition % python3 sort.py
{'worker_id': 0, 'gpus': 0, 'dep': 7, 'cpus': 8, 'runs': []}
{'worker_id': 5, 'gpus': 1, 'dep': 0, 'cpus': 6, 'runs': []}
{'worker_id': 3, 'gpus': 1, 'dep': 2, 'cpus': 6, 'runs': [1]}
{'worker_id': 4, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1, 2, 3]}
{'worker_id': 6, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1, 2, 3]}
{'worker_id': 2, 'gpus': 1, 'dep': 4, 'cpus': 6, 'runs': [1, 2, 3, 4]}
{'worker_id': 1, 'gpus': 0, 'dep': 5, 'cpus': 8, 'runs': [1, 2, 3, 4, 5, 6]}
```
2. When the current bundle requested for `1` gpus, sorting results become:
```
stanford@Stanfords-MacBook-Pro competition % vim sort.py 
stanford@Stanfords-MacBook-Pro competition % python3 sort.py
{'worker_id': 5, 'gpus': 1, 'dep': 0, 'cpus': 6, 'runs': []}
{'worker_id': 3, 'gpus': 1, 'dep': 2, 'cpus': 6, 'runs': [1]}
{'worker_id': 6, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1, 2, 3]}
{'worker_id': 4, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1, 2, 3]}
{'worker_id': 2, 'gpus': 1, 'dep': 4, 'cpus': 6, 'runs': [1, 2, 3, 4]}
{'worker_id': 0, 'gpus': 0, 'dep': 7, 'cpus': 8, 'runs': []}
{'worker_id': 1, 'gpus': 0, 'dep': 5, 'cpus': 8, 'runs': [1, 2, 3, 4, 5, 6]}
```

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -383,12 +383,6 @@ class BundleManager(object):
         Filters the workers to those that can run the given bundle and returns
         the list sorted in order of preference for running the bundle.
         """
-        # keep track of which workers have GPUs
-        has_gpu = {}
-        for worker in workers_list:
-            worker_id = worker['worker_id']
-            has_gpu[worker_id] = worker['gpus'] > 0
-
         # Filter by tag.
         request_queue = bundle.metadata.request_queue
         if request_queue:
@@ -436,11 +430,21 @@ class BundleManager(object):
             else:
                 deps = set(worker['dependencies'])
                 num_available_deps = len(needed_deps & deps)
-            worker_id = worker['worker_id']
 
-            # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
-            gpu_priority = bundle_resources.gpus or not has_gpu[worker_id]
-            return (gpu_priority, num_available_deps, worker['cpus'], random.random())
+            # If the bundle requested for GPUs, prioritize only those workers that have GPUs.
+            # If the bundle didn't request for GPUs (a CPU only job), prioritize all workers based criteria as follows:
+            # 1. number of running jobs.
+            # 2. number of available dependencies.
+            # 3. number of CPUs.
+            # 4. random seeds.
+            gpu_priority = bundle_resources.gpus and worker['gpus']
+            return (
+                gpu_priority,
+                -len(worker['run_uuids']),
+                num_available_deps,
+                worker['cpus'],
+                random.random(),
+            )
 
         workers_list.sort(key=get_sort_key, reverse=True)
 
--- a/tests/server/bundle_manager_test.py
+++ b/tests/server/bundle_manager_test.py
@@ -5,14 +5,59 @@ from codalab.objects.metadata_spec import MetadataSpec
 from codalab.server.bundle_manager import BundleManager
 from codalab.worker.bundle_state import RunResources
 from codalab.bundles import RunBundle
+from codalab.lib.codalab_manager import CodaLabManager
 
 
 class BundleManagerTest(unittest.TestCase):
+    def setUp(self):
+        self.codalab_manager = Mock(CodaLabManager)
+        self.codalab_manager.config = {
+            "workers": {
+                'default_cpu_image': 'codalab/default-cpu:latest',
+                'default_gpu_image': 'codalab/default-gpu:latest',
+            }
+        }
+        self.bundle_manager = BundleManager(self.codalab_manager)
+        self.bundle = Mock(spec=RunBundle, metadata=Mock(spec=MetadataSpec))
+        self.bundle.dependencies = []
+        self.bundle.metadata.request_queue = None
+        self.bundle_resources = RunResources(
+            cpus=0, gpus=0, docker_image='', time=100, memory=1000, disk=1000, network=False
+        )
+        self.workers_list = self.get_sample_workers_list()
+
+    def tearDown(self):
+        del self.bundle
+        del self.bundle_manager
+        del self.codalab_manager
+        del self.bundle_resources
+        del self.workers_list
+
     def get_sample_workers_list(self):
         workers_list = [
             {
+                'worker_id': 0,
+                'cpus': 4,
+                'gpus': 1,
+                'memory_bytes': 4 * 1000,
+                'tag': None,
+                'run_uuids': [1, 2],
+                'dependencies': [],
+                'shared_file_system': False,
+            },
+            {
                 'worker_id': 1,
-                'cpus': 6,
+                'cpus': 4,
+                'gpus': 1,
+                'memory_bytes': 4 * 1000,
+                'tag': None,
+                'run_uuids': [1, 2, 3],
+                'dependencies': [],
+                'shared_file_system': False,
+            },
+            {
+                'worker_id': 2,
+                'cpus': 4,
                 'gpus': 1,
                 'memory_bytes': 4 * 1000,
                 'tag': None,
@@ -21,18 +66,18 @@ class BundleManagerTest(unittest.TestCase):
                 'shared_file_system': False,
             },
             {
-                'worker_id': 2,
-                'cpus': 4,
+                'worker_id': 3,
+                'cpus': 6,
                 'gpus': 0,
                 'memory_bytes': 4 * 1000,
                 'tag': None,
-                'run_uuids': [],
+                'run_uuids': [1],
                 'dependencies': [],
                 'shared_file_system': False,
             },
             {
-                'worker_id': 3,
-                'cpus': 4,
+                'worker_id': 4,
+                'cpus': 6,
                 'gpus': 0,
                 'memory_bytes': 2 * 1000,
                 'tag': None,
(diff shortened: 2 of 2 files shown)


yipenghe:
Is gpu_priority 1. number of gpus or 2. number of **available** gpus

For example, if a gpu worker got a assigned a cpu job, another worker got assigned a gpu job, all other is the same. A new gpu-related job comes in, will the order prioritize the first worker but not random seed?

candicegjing:
> Is gpu_priority 1. number of gpus or 2. number of **available** gpus
> 
> For example, if a gpu worker got a assigned a cpu job, another worker got assigned a gpu job, all other is the same. A new gpu-related job comes in, will the order prioritize the first worker but not random seed?

Yes. I see your concern. I thought about this as well. Technically, I think we should use **available** [gpus/cpus] counts. However, right now the data we have are **total** number of [gpus/cpus]. Another things is that how do we define **available** [cpus/gpus]. For example, even if 7 jobs each requires 1 cpu, are scheduled to run on a worker with only 4 cpus available, those 7 jobs can still run without a problem. This is also something I want to find a better way to handle it. I guess one option is to use the difference between **total** number of [gpus/cpus] and `running jobs` on each worker. But this sounds similar to what we have now.

yipenghe:
> > Is gpu_priority 1. number of gpus or 2. number of **available** gpus
> > For example, if a gpu worker got a assigned a cpu job, another worker got assigned a gpu job, all other is the same. A new gpu-related job comes in, will the order prioritize the first worker but not random seed?
> 
> Yes. I see your concern. I thought about this as well. Technically, I think we should use **available** [gpus/cpus] counts. However, right now the data we have are **total** number of [gpus/cpus]. Another things is that how do we define **available** [cpus/gpus]. For example, even if 7 jobs each requires 1 cpu, are scheduled to run on a worker with only 4 cpus available, those 7 jobs can still run without a problem. This is also something I want to find a better way to handle it. I guess one option is to use the difference between **total** number of [gpus/cpus] and `running jobs` on each worker. But this sounds similar to what we have now.

I think it's probably ok to overassign for cpu tasks, but I'm not sure how multiple process shares a gpu  work, and how much it will slow down task

yipenghe:
> > Is gpu_priority 1. number of gpus or 2. number of **available** gpus
> > For example, if a gpu worker got a assigned a cpu job, another worker got assigned a gpu job, all other is the same. A new gpu-related job comes in, will the order prioritize the first worker but not random seed?
> 
> Yes. I see your concern. I thought about this as well. Technically, I think we should use **available** [gpus/cpus] counts. However, right now the data we have are **total** number of [gpus/cpus]. Another things is that how do we define **available** [cpus/gpus]. For example, even if 7 jobs each requires 1 cpu, are scheduled to run on a worker with only 4 cpus available, those 7 jobs can still run without a problem. This is also something I want to find a better way to handle it. I guess one option is to use the difference between **total** number of [gpus/cpus] and `running jobs` on each worker. But this sounds similar to what we have now.

From the test result you posted, it seems like it's what we want

candicegjing:
> > > Is gpu_priority 1. number of gpus or 2. number of **available** gpus
> > > For example, if a gpu worker got a assigned a cpu job, another worker got assigned a gpu job, all other is the same. A new gpu-related job comes in, will the order prioritize the first worker but not random seed?
> > 
> > 
> > Yes. I see your concern. I thought about this as well. Technically, I think we should use **available** [gpus/cpus] counts. However, right now the data we have are **total** number of [gpus/cpus]. Another things is that how do we define **available** [cpus/gpus]. For example, even if 7 jobs each requires 1 cpu, are scheduled to run on a worker with only 4 cpus available, those 7 jobs can still run without a problem. This is also something I want to find a better way to handle it. I guess one option is to use the difference between **total** number of [gpus/cpus] and `running jobs` on each worker. But this sounds similar to what we have now.
> 
> I think it's probably ok to overassign for cpu tasks, but I'm not sure how multiple process shares a gpu work, and how much it will slow down task

That's true, GPU could be a problem. We might need to revisit this if GPU assignment keeps bothering us.

candicegjing:
@yipenghe Updated the test code and some comments, let me know if you think this is good to go.

codalab/server/bundle_manager.py
```diff
@@ -436,11 +430,21 @@ def get_sort_key(worker):
...
                 num_available_deps = len(needed_deps & deps)
-            worker_id = worker['worker_id']
 
-            # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
-            gpu_priority = bundle_resources.gpus or not has_gpu[worker_id]
-            return (gpu_priority, num_available_deps, worker['cpus'], random.random())
+            # If the bundle requested for GPUs, prioritize only those workers that have GPUs.
+            # If the bundle didn't request for GPUs (a CPU only job), prioritize all workers based criteria as follows:
+            # 1. number of running jobs.
+            # 2. number of available dependencies.
```
percyliang, line 437: Number of available dependencies should be prioritized as high as possible (ideally the sum of the sizes of the dependencies) because we want to leverage the local bundle cache as much as possible, of course subject to being able to run a job at all.

codalab/server/bundle_manager.py
```diff
@@ -436,11 +430,21 @@ def get_sort_key(worker):
...
+            # 3. number of CPUs.
+            # 4. random seeds.
+            gpu_priority = bundle_resources.gpus and worker['gpus']
```
percyliang, line 440: `and` - shouldn't this be an integer?

percyliang:
Isn't the main problem from the original ticket the fact that we have `7/6` gpus, which looks like a bug.  But this PR alters the scheduling logic, which seems more than just a bug fix?

candicegjing:
> Isn't the main problem from the original ticket the fact that we have `7/6` gpus, which looks like a bug. But this PR alters the scheduling logic, which seems more than just a bug fix?

Oh, I thought what @yipenghe reported is that the jobs aren't evenly distributed across workers, as jobs are mainly scheduled to run on some specific workers but with many other workers in idle state. Maybe @yipenghe could clarify this? 
Also we have another ticket to track the reporting issue #1911

codalab/server/bundle_manager.py
candicegjing: Yes. `1 and 1` returns `1`. `0 and 1` returns `0`.

candicegjing: Oh, i see. this won't work when number of `gpus > 1`

yipenghe:
> > Isn't the main problem from the original ticket the fact that we have `7/6` gpus, which looks like a bug. But this PR alters the scheduling logic, which seems more than just a bug fix?
> 
> Oh, I thought what @yipenghe reported is that the jobs aren't evenly distributed across workers, as jobs are mainly scheduled to run on some specific workers but with many other workers in idle state. Maybe @yipenghe could clarify this?
> Also we have another ticket to track the reporting issue #1911

I think it was more in the sense of #1911 , I thought you meant the current scheduling logic causes the job assignment to overassign a worker

candicegjing:
> > > Isn't the main problem from the original ticket the fact that we have `7/6` gpus, which looks like a bug. But this PR alters the scheduling logic, which seems more than just a bug fix?
> > 
> > 
> > Oh, I thought what @yipenghe reported is that the jobs aren't evenly distributed across workers, as jobs are mainly scheduled to run on some specific workers but with many other workers in idle state. Maybe @yipenghe could clarify this?
> > Also we have another ticket to track the reporting issue #1911
> 
> I think it was more in the sense of #1911 , I thought you meant the current scheduling logic causes the job assignment to overassign a worker

Technically yes. The current scheduling logic does contribute to the over assigning problem.

PR #1963, opened 2020-02-05
Async loading of worksheets

### LINKED ISSUE #1234: Make loading worksheets asynchronous
Currently, the entire worksheet (including fetching all information of all bundles) has to be loaded before anything renders (including the title of the worksheet).  Should instead get the worksheet and then asynchronously load all the parts of the worksheet.  This will require refactoring some backend APIs.

We should split the /interpret request into two request:
- get worksheet info and items
- get bundle_infos that are used for each item (i.e. which are rendered in the table)

epicfaace:
Does async loading of worksheets (fixes #1234). When a worksheet loads:
- First, it calls the endpoint `/rest/interpret/worksheet/...?brief=1` -- this gets the data but does not resolve / query for bundle infos.
- Then, it calls the endpoint `/rest/interpret/worksheet/...?brief=0` -- this gets all the data, including bundle infos, which is how the interpret endpoint behaved previously.

After the first endpoint is queried, everything except for the bundle data shows up. It shows animated placeholders for the actual bundle infos as below:

image

The above placeholder is loaded whenever a text item with text equal to `"<codalab_bundle_info_loading>"` is retrieved from the server.

This PR also refactors much of the worksheet fetch logic on the frontend to use async/await instead of callbacks.

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -35,6 +35,7 @@ import Grid from '@material-ui/core/Grid';
 import WorksheetDialogs from '../WorksheetDialogs';
 import { ToastContainer, toast, Zoom } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
+import queryString from 'query-string';
 
 /*
 Information about the current worksheet and its items.
@@ -50,31 +51,25 @@ var WorksheetContent = (function() {
         this.info = null; // Worksheet info
     }
 
-    WorksheetContent.prototype.fetch = function(props) {
+    WorksheetContent.prototype.fetch = async function(props) {
         // Set defaults
         props = props || {};
-        props.success = props.success || function(data) {};
-        props.error = props.error || function(xhr, status, err) {};
         if (props.async === undefined) {
             props.async = true;
         }
-
-        $.ajax({
+        const queryParams = { brief: props.brief ? 1 : 0 };
+        const info = await $.ajax({
             type: 'GET',
-            url: '/rest/interpret/worksheet/' + this.uuid,
+            url:
+                '/rest/interpret/worksheet/' + this.uuid + '?' + queryString.stringify(queryParams),
             // TODO: migrate to using main API
             // url: '/rest/worksheets/' + ws.uuid,
             async: props.async,
             dataType: 'json',
             cache: false,
-            success: function(info) {
-                this.info = info;
-                props.success(this.info);
-            }.bind(this),
-            error: function(xhr, status, err) {
-                props.error(xhr, status, err);
-            },
         });
+        this.info = info;
+        return info;
     };
 
     WorksheetContent.prototype.saveWorksheet = function(props) {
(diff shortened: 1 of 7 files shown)


candicegjing:
How's the performance improvement?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -506,47 +501,45 @@ class Worksheet extends React.Component {
...
+            });
+
+            await this.state.ws.fetch({ brief: false });
+            this.setState({
+                updating: false,
+                version: this.state.version + 1,
+                numOfBundles: this.getNumOfBundles(),
+                errorMessage: '',
+            });
+        } catch (e) {
```
yipenghe, line 537: what might cause error here?

yipenghe (commented):
Looks good in general, user will be able to view non-bundle row items on top faster, but still needs to wait for all bundles to load before scrolling

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1056,107 +1049,99 @@ class Worksheet extends React.Component {
...
-    reloadWorksheet = (
+    reloadWorksheet = async (
         partialUpdateItems,
         rawIndexAfterEditMode,
         { moveIndex = false, textDeleted = false } = {},
     ) => {
         if (partialUpdateItems === undefined) {
             $('#update_progress').show();
             this.setState({ updating: true });
-            this.state.ws.fetch({
```
percyliang, line 1067: There's a lot of frontend changes here...can you comment on what the nature of these changes are with respect to the async loading feature?

percyliang (commented):
Could you create a large worksheet with many bundles, and use a table schema that requires loading lots of files from disk?  This is where things become really slow.

There are multiple levels here:
1) Getting the raw items of the worksheet
2) Doing searches (`expand_raw_items`)
3) Getting more info about the items in the database (`interpret_items`)
4) Reading things from disk (`resolve_interpreted_blocks`)

Your brief mode only bypasses 3), but I was initially imagining that the brief mode would do 2) and 4) as well...  In my experience, 4) is really the big one.  I think it'd be useful to take a few different huge worksheets and profile where the slowness is coming from, so we can load as much as possible subject to a latency constraint.

frontend/src/components/worksheets/Worksheet/Worksheet.js
epicfaace: if the ajax request fails

frontend/src/components/worksheets/Worksheet/Worksheet.js
epicfaace: Mainly, I've refactored `this.state.ws.fetch` to return a promise instead of using callbacks. This way, this simplifies the work done in `componentDidMount()`, since `await this.state.ws.fetch({ brief: true });` can just be called once and then followed by `wait this.state.ws.fetch()`.

Nothing else in terms of frontend functionality was changed.

epicfaace:
> Your brief mode only bypasses 3), but I was initially imagining that the brief mode would do 2) and 4) as well... In my experience, 4) is really the big one.

The brief mode bypasses both 3 and 4 for all table items.

epicfaace:
check large worksheet @teetone

teetone:
@epicfaace Is there a way to get these changes into `dev`? That's the best way to test the `large` worksheet.

Link to large worksheet in `dev`: https://worksheets-dev.codalab.org/worksheets/0x248dff2eadff4694bf8ef90b7445d5fa

epicfaace:
@teetone it says "worksheet not found" on dev?

epicfaace:
I added some profiling in the async-profile branch. Here's what I get (this is in seconds, with a single worksheet with hundreds of bundles). What can I add to make it run even slower?

total_times	0.08689904999999998
get_worksheet_lines	0.0004367219999998895
expand_raw_items	0.028927035999999795
interpret_items	0.036016750000000375
resolve_interpreted_blocks	0.002023271999999743

epicfaace:
brief mode:

name | time
-- | --
total_times | 0.06227632700000002
get_worksheet_lines | 0.0005665909999999913
expand_raw_items | 0.025070216999999992
interpret_items | 0.011678946999999995
resolve_interpreted_blocks | 0.00006042900000000073


non-brief mode:

name | time
-- | --
total_times | 0.077385386
get_worksheet_lines | 0.00045964700000000747
expand_raw_items | 0.024892895000000026
interpret_items | 0.033394261000000036
resolve_interpreted_blocks | 0.0011049989999999954

epicfaace:
Turns out I was using the wrong metrics (I was using `time.process_time()`, which is not measured in seconds). here are better stats for the stress test worksheet:

https://worksheets-dev.codalab.org/worksheets/0xf1bddfa2f5634160a62db5ea781d41e9

Regular mode:

name | time
-- | --
get_worksheet_lines|0.005675554275512695
expand_raw_items| 10.567695617675781
interpret_items| 0.43700718879699707
resolve_interpreted_blocks| 1.9040861129760742
total| 14.141964673995972

Even in brief mode:

name | time
-- | --
get_worksheet_lines| 0.004877328872680664
expand_raw_items| 9.485067367553711
interpret_items| 0.08468079566955566
resolve_interpreted_blocks| 0.001806497573852539
total| 10.811790466308594

Seems like expand_raw_items is the major timesink, followed by resolve_interpreted_blocks.

epicfaace:
It appears that the only thing called in expand_raw_items is search/wsearch. I'll modify this PR to make the UI load search results in a different request.

https://github.com/codalab/codalab-worksheets/blob/166e45d9e1495ea59f4c849aa2c7a359eb080f8a/codalab/rest/interpret.py#L570-L590

epicfaace:
superseded by #2086

PR #1964, opened 2020-02-05
Mount backend files as a volume

epicfaace:
This way, when a backend file is changed, one can just re-start codalab in
dev mode for the backend files to be updated. Previously, one would have to
rebuild the docker images entirely when a backend file was updated.

### DIFF
--- a/docker/compose_files/docker-compose.dev.yml
+++ b/docker/compose_files/docker-compose.dev.yml
@@ -11,3 +11,4 @@ services:
   rest-server:
     volumes:
       - ./../../alembic/:/opt/codalab-worksheets/alembic/
+      - ./../../codalab/:/opt/codalab-worksheets/codalab/
--- a/docker/compose_files/docker-compose.dev.yml
+++ b/docker/compose_files/docker-compose.dev.yml
@@ -7,7 +7,7 @@ services:
     volumes:
       - ./../../frontend/src:/opt/frontend/src
   # 2) Hot load the alembic version files so that we can create new database migrations (which need to written out
-  # of Docker into the `alembic` source directory)
+  # of Docker into the `alembic` source directory), as well as the backend files in codalab/
   rest-server:
     volumes:
       - ./../../alembic/:/opt/codalab-worksheets/alembic/


percyliang (commented):
We didn't do this because building the volume for the backend was really fast (is that not the case?) and we needed to pip install, which reads from a different location.

epicfaace:
It's not very fast to build the server image -- it starts all the way from `Step 19/32 : COPY codalab codalab` and pip installs. Additionally, when running with the `-b` tag, you need to wait for all the other images to build as well (while all layers should be cached, this adds a non-negligible wait time).

All in all, it's noticeably faster to be able to just start without using the `-b` tag, especially since one has to restart to test any change in the backend code.

For the pip install issue, one could always re-build the docker images after changing the pip dependencies (which won't happen very often, anyway).

percyliang:
How long does it take for you to rebuild?  It was < 5 seconds for me.

I meant we `pip install codalab` so that we can just use `cl` as usual.

epicfaace:
It takes me 1 minute 50 seconds to start with the `-bd` option (after making a single change in the backend code).

It takes me 40 seconds to start with only the `-d` option.

percyliang:
Seems like the docker caching isn't working properly?

epicfaace:
Docker caching seems to be working fine. Maybe my computer is just slower than yours.

[logs with -bd](https://gist.github.com/epicfaace/6a935c6f625c67635089275100acbf4a)

[logs with -d](https://gist.github.com/epicfaace/696afe82506e4acc2639d314c50d4c74)

bkgoksel:
I don't remember exactly why we didn't do this back when we first added the service, probably because it was very low priority. I think if you can make it work, go for it. (although in my experience too the `start -bs default` takes 5-10 seconds max, having it be instantaneous is always better)

epicfaace:
@percyliang I don't see any drawback to having this merged, even if just doing `start -bs default` builds things faster, ask @bkgoksel said.

I don't see an issue with having to build the image again in order for `pip install codalab` to run so that we can use `cl` as usual. That seems better than the current scenario (having to build the image again for `pip install` to update `cl`, and also having to build the image again to update backend code).

percyliang:
The potential drawback is having two separate code paths and things not working as intended.  If you mount the backend files, and you run a `cl` command, does it actually pick it up?  I thought it wouldn't because you would need to `pip install` it first...

For me it takes about 5-10 seconds, but @epicfaace was saying that it was taking over a minute.

Separately, you can target certain services only by doing `start -bs rest-server`.

epicfaace:
Also, running codalab is now way faster after I switched to a new computer.

PR #1969, opened 2020-02-08
Reduce cache miss when deducting worker resources

### LINKED ISSUE #1968: Too many cache miss in deduct_worker_resources
Currently, there are too many cache misses in function [deduct_worker_resource](https://github.com/codalab/codalab-worksheets/blob/bb20c4e522c5c9d4fc2d50e2930eca4fd2c037a2/codalab/server/bundle_manager.py#L360-L378), where we check the existence of running bundle in `bundle` table. We need to fix this to avoid overestimating worker resources.

candicegjing:
Fixed #1968 

This PR will reduce cache miss when deducting worker resources for dispatching jobs. The cache miss is due to the fact that we don't consider those staged bundles that will be scheduled to run when we created the [running_bundles_info](https://github.com/codalab/codalab-worksheets/blob/bb20c4e522c5c9d4fc2d50e2930eca4fd2c037a2/codalab/server/bundle_manager.py#L760) object. 
Adding those staged bundles as part of the `running_bundles_info` will eliminate the cache miss. However, on the other side, we might be conservative in assigning jobs to workers, as there could be jobs that run extremely fast and finish before [the current dispatching for loop](https://github.com/codalab/codalab-worksheets/blob/bb20c4e522c5c9d4fc2d50e2930eca4fd2c037a2/codalab/server/bundle_manager.py#L330) finished.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -10,7 +10,7 @@ import time
 import traceback
 
 from codalab.objects.permission import check_bundles_have_read_permission
-from codalab.common import PermissionError
+from codalab.common import PermissionError, NotFoundError
 from codalab.lib import bundle_util, formatting, path_util
 from codalab.server.worker_info_accessor import WorkerInfoAccessor
 from codalab.worker.file_util import remove_path
@@ -327,7 +327,7 @@ class BundleManager(object):
         )
 
         # Build a dictionary which maps from uuid to running bundle and bundle_resources
-        running_bundles_info = self._get_running_bundles_info(workers)
+        running_bundles_info = self._get_running_bundles_info(workers, staged_bundles_to_run)
 
         # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
@@ -372,7 +372,6 @@ class BundleManager(object):
         """
         From each worker, subtract resources used by running bundles. Modifies the list.
         """
-
         for worker in workers_list:
             for uuid in worker['run_uuids']:
                 # Verify if the current bundle exists in both the worker table and the bundle table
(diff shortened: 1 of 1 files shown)


bkgoksel:
Can we try to fetch the bundle info from the database if it's not in the cache table? It seems like that's necessary for running a cache anyways. It might be good to also do what's done in this PR and add the staged bundles that will likely be scheduled to workers throughout the loop, but the core thing is that we should never skip runs' resources if we can find them in the database.

codalab/server/bundle_manager.py
```diff
@@ -361,16 +361,8 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
...
         for worker in workers_list:
             for uuid in worker['run_uuids']:
-                # Verify if the current bundle exists in both the worker table and the bundle table
-                if uuid not in running_bundles_info:
-                    logger.info(
-                        'Bundle {} exists on worker {} but no longer found in the bundle table. '
-                        'Skipping for resource deduction.'.format(uuid, worker['worker_id'])
-                    )
-                    continue
                 bundle_resources = running_bundles_info[uuid]["bundle_resources"]
```
bkgoksel: @candicegjing can you merge master? I made a change on master to this part where we still do the check and call the db if the bundle is not in the cache. Do you think we need that? Is it possible for a run to be on a worker and not in the cache, but still in the db?

bkgoksel (changes requested):

codalab/server/bundle_manager.py
candicegjing: @bkgoksel I saw your changes. I think your change is accurate. But the only thing is that it introduces many database queries which will slow down the performance of the scheduler.

bkgoksel: I agree, I think we want to have both your change (ie cache staged bundles) + query the database if somehow a bundle isn't in that cache either.

This is more of a correctness guarantee, and hopefully we shouldn't hit this case and make that many db queries often (when your changes are added to my change)

candicegjing: @bkgoksel The reason that I haven't rebased to master is mainly that I am not sure which side we should pick (accuracy vs performance). I wish there could be a way to fulfill both requirements. I was trying to recall the original reason that leads us to check for the bundle existence during resource deduction. Do you by any chance remember that it's because if the bundle doesn't exist in the Bundle table, there will be an exception thrown from `compute_bundle_resources` function as it tries to get this bundle from database? Then, when this happened, the error will cause bundle-manager to fail?

bkgoksel: @candicegjing for your specific question, I believe that's what was happening. 

As for performance vs accuracy, this is a spot where we can't afford to have any inaccuracy because if we assign too many jobs to a worker, they simply fail. Because of that, any caching/perf improvement has to guarantee not breaking the correctness of resource calculations for workers. 

In this case, I don't think falling back to the db call would affect performance anyways because you cache all the running bundles AND the staged bundles, so it's a very small chance we have a non-cached running bundle. And IF that happens, we  have no choice but to fetch the information for it from the database, as otherwise we'll potentially fail existing runs, which is the worst outcome from a user experience perspective.

candicegjing: Yes. I agree. I just rebased to master for now. Let me know if I need to change anything else before merging to master.

bkgoksel (approved):
Looks good!

PR #1975, opened 2020-02-09
Copy multiple bundle ids and form markdowns

### LINKED ISSUE #1802: Copy bundle uuid within and across tables in worksheet

Bulk attach bundles to a particular worksheet, would be handy

Similar to bulk detach/kill/delete, except ask for a worksheet uuid as arguments in dialog, current cmd line supports bulk attach I think

yipenghe:
First part of fix #1802 
Can also use 'a+c' and 'enter' to do the actions

Screenshot from 2020-02-09 09-47-34


yipenghe:
Future updates will be all addressed in #2009

Issue #1934, opened 2020-01-27
can't see the bundle details when click on image


percyliang:
In the old UI, the side panel would be able to see the bundle info when the cursor was over an image, but now you can't.  Ideally, would show the same bundle details when press enter on other display modes (text, image, etc.).

yipenghe:
I can take this, should it be a similar panel to the default mode?

percyliang:
By default mode, you mean table mode?  Yes.

PR #1976, opened 2020-02-09
View image detail by click and 'enter'

yipenghe:
#1934 
Open/click bundle detail when clicked on image item or hit 'enter' when focused


frontend/src/components/worksheets/items/ImageItem.js
```diff
@@ -1,23 +1,49 @@
...
+            showDetail: false,
+            bundleInfoUpdates: {},
+        };
     }
 
+    receiveBundleInfoUpdates = (update) => {
+        let { bundleInfoUpdates } = this.state;
+        // Use object spread to update.
+        bundleInfoUpdates = { ...bundleInfoUpdates, ...update };
+        this.setState({ bundleInfoUpdates });
```
epicfaace: I think this would be clearer
```suggestion
        this.setState({ bundleInfoUpdates: { ...bundleInfoUpdates, ...update } });
```

frontend/src/components/worksheets/items/ImageItem.js
```diff
@@ -1,23 +1,49 @@
...
+                (e) => {
+                    e.preventDefault();
+                    if (!this.props.confirmBundleRowAction(e.code)) {
+                        this.setState({ showDetail: !this.state.showDetail });
+                    }
+                },
+                'keydown',
+            );
+            // unbind shortcuts that are active for markdown_block and worksheet_block
+            Mousetrap.unbind('i');
```
epicfaace: Is `i` the only shortcut that needs to be unbound here?

frontend/src/components/worksheets/items/ImageItem.js
```diff
@@ -1,23 +1,49 @@
...
+        }
+        const { classes } = this.props;
+        var item = this.props.item;
```
epicfaace:
```suggestion
        const item = this.props.item;
```

frontend/src/components/worksheets/items/ImageItem.js
```diff
@@ -1,23 +1,49 @@
...
+        var bundleInfo = item.bundles_spec.bundle_infos[0];
```
epicfaace:
```suggestion
        const bundleInfo = item.bundles_spec.bundle_infos[0];
```

frontend/src/components/worksheets/items/ImageItem.js
```diff
@@ -27,12 +53,28 @@ class ImageItem extends React.Component {
         if (this.props.item.hasOwnProperty('width')) {
             styles['width'] = this.props.item.width + 'px';
         }
-
         return (
-            <div className='ws-item' onClick={this.handleClick}>
-                <div className={className} ref={this.props.item.ref}>
+            <div className='ws-item'>
+                <div className={className} ref={this.props.item.ref} onClick={this.handleClick}>
```
epicfaace, line 56: Why did you move the `onClick` handler to the child?

yipenghe: Otherwise the click will trigger when we are clicking on the bundle detail

frontend/src/components/worksheets/items/ImageItem.js
yipenghe: Actually it's not needed...copied from the bundle row implementation

PR #1977, opened 2020-02-10
Renew WorkerInfoAccessor to avoid stale cache

bkgoksel:
Currently we only renew this cache once per bundle manager loop. While we don't expect this loop to take as long as one worker timeout cycle, there's no guarantee for it and we wouldn't want to start setting workers as offline due to this.

Drive-by fix: Don't make workers sleep for an hour if check-in with server fails, keep trying.


codalab/lib/spec_util.py
```diff
@@ -11,9 +11,7 @@
 UUID_PREFIX_REGEX = re.compile('^0x[0-9a-f]{1,31}$')
 BEGIN_NAME_STR = '[a-zA-Z_]'
 NAME_STR = '[a-zA-Z_][a-zA-Z0-9_\.\-]*'
-NAME_PATTERN_STR = (
-    '[%\*a-zA-Z0-9_\.\-]+'
-)  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
+NAME_PATTERN_STR = '[%\*a-zA-Z0-9_\.\-]+'  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
```
bkgoksel: This was accidentally formatted by a wrong version of black on my earlier hotfix commit.

scripts/competitiond.py
```diff
@@ -421,9 +421,11 @@ def collect_submissions(self):
         """
         logger.debug("Collecting latest submissions")
         submissions = self._fetch_latest_submissions()
-        previous_submission_ids, num_total_submissions, num_period_submissions = (
-            self._fetch_submission_history()
-        )
+        (
```
bkgoksel, line 424: Same as first comment.

codalab/server/bundle_manager.py
```diff
@@ -356,7 +354,7 @@ def get_available_workers(user_id):
                         )
                         # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
                         continue
-            if workers_list is None:
+            if not workers_list: # Length is 0 (private user with no workers) or is None (root user)
```
bkgoksel: This was a heinous one.

scripts/competitiond.py
bkgoksel: (ie fixed earlier reformat with wrong version of black.

candicegjing:
Last time when bench-marked the jobs scheduler, the scheduling part takes under 1 second for 100 staged jobs which don't have any dependencies: [profiling results](https://github.com/codalab/codalab-worksheets/issues/1254#issuecomment-582025538).  I guess the feature implemented in this ticket would be safe to have. We just need to benchmark one time to double-check if this will affect performance.

codalab/server/bundle_manager.py
```diff
@@ -356,7 +354,9 @@ def get_available_workers(user_id):
                         )
                         # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
                         continue
-            if not workers_list: # Length is 0 (private user with no workers) or is None (root user)
+            if (
+                not workers_list
+            ):  # Length is 0 (private user with no workers) or is None (root user)
```
teetone: Is this because of black? The formatting looks off.

codalab/server/bundle_manager.py
```diff
@@ -632,7 +632,7 @@ def _schedule_run_bundles(self):
         READY / FAILED, no worker_run DB entry:
             Finished.
         """
-        workers = WorkerInfoAccessor(self._worker_model.get_workers())
+        workers = WorkerInfoAccessor(self._worker_model, WORKER_TIMEOUT_SECONDS - 5)
```
teetone, line 630: It's not clear to me what the -5 is for when setting the timeout.

codalab/server/bundle_manager.py
bkgoksel: I think black includes the comment in the line length, hence it breaks it. 
I'll move the comment above the condition so it doesn't break.

codalab/server/bundle_manager.py
bkgoksel: It's safer to renew slightly more frequently than the timeout to avoid a case where we get the worker states from this object say 59 seconds after it was last updated, but by the time we check the timestamps, it's already been 60 seconds. I can set it to be 1 just to show that it's a symbolic safety measure.

teetone (approved):

candicegjing (approved):

Issue #1967, opened 2020-02-07
[USER ISSUE] Failed to run command df codalab-worker-scratch: 'Available' is not in list


Ajanulk:
**Describe the bug**
I follow the official document [https://codalab-worksheets.readthedocs.io/en/latest/Execution/#running-your-own-worker](url) to run my own worker, but I find that it didn't work on my worker (I have installed Docker 19.03).

**Codalab Information**
- *Your username*:lrh001
- *UUID of worksheet where problem occurred*:0x395c6910f16f4221b28cc36d0cf7e874
- *Error message received from Codalab (if applicable)*:

2020-02-07 17:38:44,374 Loaded 0 dependencies, 0 paths from cache.
2020-02-07 17:38:44,410 http://localhost:None "GET /v1.35/version HTTP/1.1" 200 861
2020-02-07 17:38:44,411 Looking for auth config
2020-02-07 17:38:44,411 No auth config in memory - loading from filesystem
2020-02-07 17:38:44,412 Trying paths: ['/home/lrh/.docker/config.json', '/home/lrh/.dockercfg']
2020-02-07 17:38:44,412 No config file found
2020-02-07 17:38:44,412 Looking for auth entry for 'docker.io'
2020-02-07 17:38:44,412 No entry found
2020-02-07 17:38:44,412 No auth config found
2020-02-07 17:38:47,129 http://localhost:None "POST /v1.35/images/create?tag=9.0-cudnn7-devel-ubuntu16.04&fromImage=nvidia%2Fcuda HTTP/1.1" 200 None
2020-02-07 17:38:47,134 http://localhost:None "GET /v1.35/images/nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04/json HTTP/1.1" 200 None
2020-02-07 17:38:47,141 http://localhost:None "POST /v1.35/containers/create HTTP/1.1" 400 47
2020-02-07 17:38:47,141 Cannot initialize NVIDIA runtime, no GPU support: Problem getting NVIDIA devices: 400 Client Error: Bad Request ("Unknown runtime specified nvidia")
2020-02-07 17:38:47,142 Trying paths: ['/home/lrh/.docker/config.json', '/home/lrh/.dockercfg']
2020-02-07 17:38:47,142 No config file found
2020-02-07 17:38:47,142 Trying paths: ['/home/lrh/.docker/config.json', '/home/lrh/.dockercfg']
2020-02-07 17:38:47,142 No config file found
[...]

PR #1982, opened 2020-02-11
1967: check fourth header as a back up

teetone:
Follow up to user issue #1967.

The user had a different system language other than english, which caused failures when executing the `df` command to attach a worker. This fix just simply checks the fourth key (3rd index) if `Available` is not available.

### DIFF
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -368,7 +368,7 @@ class Worker:
                 lines = output.decode().split("\n")
                 headers = lines[0].split()
                 # The machine being attached as a worker may be using a different language other than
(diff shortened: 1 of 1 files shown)


codalab/worker/worker.py
```diff
@@ -366,7 +366,10 @@ def free_disk_bytes(self):
 
             if output:
                 lines = output.decode().split("\n")
-                index = lines[0].split().index("Available")
+                headers = lines[0].split()
+                # The machine being attached as a worker may be using a different language other than
+                # english, so check the 4th header if "Available" is not present.
```
percyliang: capitalize English

percyliang (approved):

PR #1986, opened 2020-02-11
Sleep less when worker-server checkin fails

bkgoksel:
Right now we sleep for 1h, making recovery difficult if there's a routine server maintenance restart.
This PR makes is such that we don't sleep the first failure, and only sleep 5 seconds if we fail two times in a row (only for checkins, for other exceptions, the 1h sleep is kept)

### DIFF
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -40,6 +40,8 @@ class Worker:
     KILL_TIMEOUT = 100
     # Number of loops to check for bundle directory creation by server on shared FS workers
     BUNDLE_DIR_WAIT_NUM_TRIES = 120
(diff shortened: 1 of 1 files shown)


percyliang (commented):
Why sleep after 2 instead of just 1?  Could that make the logic simpler?

bkgoksel:
> Why sleep after 2 instead of just 1? Could that make the logic simpler?

No particular reason besides to save a few seconds if checkin failed only due to a network glitch and no longer term error. Since we already do all of the work of keeping `last_checkin_successful` sleeping only on the second fail adds almost no complexity (ie it's just that if block0. I can get rid of it, it's not gaining us anything probably.

percyliang (approved):

PR #1988, opened 2020-02-12
make docker image size check optional

### LINKED ISSUE #1984: Support AWS image registry parsing for image size pre-check
Current image size pre-check only supports docker images. However, we need to parse AWS image registry for some use cases: https://docs.aws.amazon.com/AmazonECR/latest/userguide/Registries.html

candicegjing:
Fix #1984.
Changed the original request to make image size check optional. @bkgoksel when you pass `--max-image-size=0` it won't trigger the image size pre-check. Let me know if this is ok to you.
This PR should go together with https://github.com/codalab/deployment/pull/15

### DIFF
--- a/codalab/worker/docker_image_manager.py
+++ b/codalab/worker/docker_image_manager.py
@@ -240,21 +240,35 @@ class DockerImageManager:
                 # Check docker image size before pulling from Docker Hub.
                 # Do not download images larger than self._max_image_size
                 # Download images if size cannot be obtained
-                try:
-                    image_size_bytes = docker_utils.get_image_size_without_pulling(image_spec)
-                    if image_size_bytes > self._max_image_size:
-                        failure_msg = (
-                            "The size of "
-                            + image_spec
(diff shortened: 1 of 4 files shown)


bkgoksel (changes requested):
Thanks! 2 things:
1. We also want to change the default value for the argument in `worker.py` to be `None`
2. This still allows images to be downloaded if their size cannot be fetched. If you want to not allow that, you should return  the failed state in the `except` clause.

candicegjing:
> Thanks! 2 things:
> 
>     1. We also want to change the default value for the argument in `worker.py` to be `None`
> 
>     2. This still allows images to be downloaded if their size cannot be fetched. If you want to not allow that, you should return  the failed state in the `except` clause.

@bkgoksel With the current change, if you don't specify `max_image_size` option, it will be default to `None`. I ran test cases including setting the `max_image_size` and without setting it, both were running fine. Let me know if you think I missed anything else.

bkgoksel (approved):
LGTM

PR #1992, opened 2020-02-13
Update pre-commit script to cleanup venv on failure and add new troubleshooting steps

### LINKED ISSUE #1937: pre_commit.sh failed setup bug and missing dependencies

armantajback:
…leshooting steps


docs/Server-Setup.md
```diff
@@ -243,6 +244,17 @@ service on the same machine, be careful about the following:
...
+If you run the codalab_service script with root privileges and see an error about either of the following:
+
+    Unknown MySQL server host 'mysql' (0)
+    Host '*.*.*.*' is not allowed to connect to this MySQL server
+
+Make sure you can run docker commands without root privileges. If you cannot, make sure to run:
+
+    sudo usermod -aG docker ${USER}
+
+Exit the terminal window in which that command is run. Once you ensure docker commands can be run without sudo, Remove the `/var` folder from within the `/codalab-worksheets` directory. **Warning: Removing the `/var` folder leads to dataloss for your local Codalab instance.** Backup any necessary data first.
```
percyliang: `/var` => `var`
dataloss => data loss

pre-commit.sh
```diff
@@ -3,9 +3,9 @@
 # Run this script before you commit.
 
 if ! [ -e venv ]; then
-  virtualenv -p python3 venv || exit 1
-  venv/bin/pip install -r requirements-server.txt || exit 1
-  venv/bin/pip install -r requirements.docs.txt || exit 1
+  virtualenv -p python3 venv || rm -rf venv; exit 1
```
percyliang: Don't you need parentheses around `rm -rf venv; exit 1`?

percyliang (changes requested):

pre-commit.sh
armantajback: I added a {} since a parenthesis would spawn a new shell so the exit wouldn't kill the script (which is the behavior I was going for).

docs/Server-Setup.md
armantajback: Done

percyliang (approved):

pre-commit.sh
```diff
@@ -11,9 +11,9 @@ fi
 venv/bin/pip install -e .
 
 # Generate docs
-venv/bin/python scripts/gen-rest-docs.py || exit 1  # Outputs to `docs`
-venv/bin/python scripts/gen-cli-docs.py || exit 1  # Outputs to `docs`
-venv/bin/mkdocs build || exit 1  # Outputs to `site`
+venv/bin/python scripts/gen-rest-docs.py || { rm -rf venv; exit 1; }  # Outputs to `docs`
```
epicfaace, line 14: Why do we want to delete the venv folder on failure? Failure of `scripts/gen-rest-docs.py` doesn't necessarily mean that something was wrong with the venv -- I could imagine that automatic deletion of the env could serve as a cause of frustration for many, and it's not too difficult to do yourself.

armantajback: I saw that venv was corrupted (I'm not sure what about it was deleting it mitigated the issues I saw). I think removing it is a reasonable thing to do since it fixes the issue I was hitting on a clean ubuntu vm but I can remove this step if you'd like. FWIW, I thought I had merged this pr but seems I didn't.

Issue #1096, opened 2019-03-11
fix GitHub security vulnerabilities (pyyaml, marshmallow)


percyliang:
As a note, we need:
- [x] `marshmallow >= 2.15.1`
- [x] `pyyaml >= 5.1`
- [x] `sqlalchemy >= 1.3.0`

bkgoksel:
As a note, we need:
- `marshmallow >= 2.15.1`
- `pyyaml >= 4.1b4`

epicfaace:
finished by #2042

PR #1995, opened 2020-02-13
bump sqlalchemy

epicfaace:
Fixes vulnerabilities (#1096).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ argcomplete==1.9.4
 PyYAML==5.1
 psutil==5.4.6
 six==1.11.0
-SQLAlchemy==1.0.8
+SQLAlchemy==1.3.0
 watchdog==0.8.3
 fusepy==2.0.4
 python-dateutil


candicegjing (approved):
LGTM 👍

percyliang:
Is there any subtle changes in behavior to be aware of with the new version?

epicfaace:
Changes - https://docs.sqlalchemy.org/en/13/changelog/migration_13.html#key-changes-core

changes that will probably affect us:
- Coercion of string SQL fragments to text() fully removed (see https://github.com/codalab/codalab-worksheets/blob/8286b1c3439f113ce00840338efbdc5e4feb115f/codalab/model/bundle_model.py#L1428)

changes that we'll need to fix later, perhaps before upgrading to the next sqlalchemy version (1.4):
- “threadlocal” engine strategy deprecated
- convert_unicode parameters deprecated

epicfaace:
Actually, we seem to be fine. The coercion change only affects code such as `.where("somecolumn = 'value'")`, which we don't have.

PR #2000, opened 2020-02-14
Allow crawling up bundle dependencies

bkgoksel:
## Description
This PR allows accessing a bundle's dependencies via their keys, as if they are local directories within the bundle. This allows doing things like building worksheet tables that aggregate stats from bundles and their dependencies, or downloading dependency files without having to know the dependency bundles themselves.

At a high level, this is made possible by having `get_target_info` -which allows clients to get information about paths requested by users from the server- also return a `resolved_uuid` and a `resolved_path`, basically letting it redirect path requests to other bundles. 

Once a call is made to `get_target_info`, it still attempts to find the path info using the old code paths, but if that fails, makes an extra call to the database to get all the dependencies. It then tries to match the first component of the path on the dependency keys, and if it finds a match, recursively calls itself on that dependency bundle and the rest of the path.

This also allows arbitrary traversals since it's a recursive call, and in the end it resolves to one uuid and one subpath.

I modified the rest of the codebase to consume these `resolved_uuid` and `resolved_path` fields that `get_target_info` returns when making any file-related request, effectively obeying the redirect.

## Testing
The following were manually tested, on top of adding a couple unit tests:
- Frontend file content schemas
- Frontend table schemas (ie json reading)
- `cl cat`
- `cl down <bundle>/<dep>/<path>` downloads `<dep>/<path>`
- `cl down <bundle>` doesn't download `<dep>` by itself
- all read ops while the bundle is still running

## Screenshots

image

### DIFF
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -583,26 +583,25 @@ class JsonApiClient(RestClient):
         return response
 
     @wrap_exception('Unable to fetch contents info of bundle {1}')
-    def fetch_contents_info(self, bundle_id, target_path='', depth=0):
-        request_path = '/bundles/%s/contents/info/%s' % (bundle_id, urllib.parse.quote(target_path))
+    def fetch_contents_info(self, target, depth=0):
+        request_path = '/bundles/%s/contents/info/%s' % (target[0], urllib.parse.quote(target[1]))
         response = self._make_request('GET', request_path, query_params={'depth': depth})
         return response['data']
 
     @wrap_exception('Unable to fetch contents blob of bundle {1}')
-    def fetch_contents_blob(
-        self, bundle_id, target_path='', range_=None, head=None, tail=None, truncation_text=None
-    ):
+    def fetch_contents_blob(self, target, range_=None, head=None, tail=None, truncation_text=None):
         """
         Returns a file-like object for the target on the given bundle.
 
-        :param bundle_id: id of target bundle
-        :param target_path: path to target in bundle
+        :param target: A tuple of bundle_id, path where:
+            bundle_id: id of target bundle
+            path: path to target in bundle
         :param range_: range of bytes to fetch
         :param head: number of lines to summarize from beginning of file
         :param tail: number of lines to summarize from end of file
         :return: file-like object containing requested data blob
         """
-        request_path = '/bundles/%s/contents/blob/%s' % (bundle_id, urllib.parse.quote(target_path))
+        request_path = '/bundles/%s/contents/blob/%s' % (target[0], urllib.parse.quote(target[1]))
         headers = {'Accept-Encoding': 'gzip'}
         if range_ is not None:
             headers['Range'] = 'bytes=%d-%d' % range_
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -6,6 +6,7 @@ import urllib.request, urllib.parse, urllib.error
 
 from codalab.common import http_error_to_exception, precondition, UsageError
 from codalab.worker.rest_client import RestClient, RestClientException
+from codalab.worker.download_util import BundleTarget
 
 
 def wrap_exception(message):
@@ -594,6 +595,10 @@ class JsonApiClient(RestClient):
             urllib.parse.quote(target.subpath),
         )
         response = self._make_request('GET', request_path, query_params={'depth': depth})
+        # Deserialize the target. See /rest/bundles/_fetch_contents_info for serialization side
+        response['data']['resolved_target'] = BundleTarget.from_dict(
+            response['data']['resolved_target']
+        )
         return response['data']
 
     @wrap_exception('Unable to fetch contents blob of bundle {1}')
--- a/codalab/client/json_api_client.py
+++ b/codalab/client/json_api_client.py
@@ -584,7 +584,15 @@ class JsonApiClient(RestClient):
 
     @wrap_exception('Unable to fetch contents info of bundle {1}')
     def fetch_contents_info(self, target, depth=0):
-        request_path = '/bundles/%s/contents/info/%s' % (target[0], urllib.parse.quote(target[1]))
+        """
+        Calls download_manager.get_target_info server-side and returns the target_info.
+        For details on return value look at worker.download_util.get_target_info
+        :param target: a worker.download_util.BundleTarget
+        """
+        request_path = '/bundles/%s/contents/info/%s' % (
+            target.bundle_uuid,
+            urllib.parse.quote(target.subpath),
+        )
         response = self._make_request('GET', request_path, query_params={'depth': depth})
         return response['data']
 
@@ -593,15 +601,16 @@ class JsonApiClient(RestClient):
         """
         Returns a file-like object for the target on the given bundle.
 
-        :param target: A tuple of bundle_id, path where:
-            bundle_id: id of target bundle
-            path: path to target in bundle
+        :param target: A worker.download_util.BundleTarget
         :param range_: range of bytes to fetch
         :param head: number of lines to summarize from beginning of file
         :param tail: number of lines to summarize from end of file
         :return: file-like object containing requested data blob
         """
-        request_path = '/bundles/%s/contents/blob/%s' % (target[0], urllib.parse.quote(target[1]))
+        request_path = '/bundles/%s/contents/blob/%s' % (
+            target.bundle_uuid,
+            urllib.parse.quote(target.subpath),
+        )
         headers = {'Accept-Encoding': 'gzip'}
         if range_ is not None:
             headers['Range'] = 'bytes=%d-%d' % range_
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2423,7 +2423,7 @@ class BundleCLI(object):
             while run_state not in State.FINAL_STATES:
                 run_state = client.fetch('bundles', bundle_uuid)['state']
                 try:
-                    client.fetch_contents_info((bundle_uuid, subpath), 0)
+                    client.fetch_contents_info(BundleTarget(bundle_uuid, subpath), 0)
                 except NotFoundError:
                     time.sleep(SLEEP_PERIOD)
                     continue
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -5,8 +5,7 @@ a list of CodaLab bundle system command-line arguments and executes them.
 Each of the supported commands corresponds to a method on this class.
 This function takes an argument list and does the action.
 
-For example:
-
+For example: 
   cl upload foo
 
 results in the following:
@@ -84,6 +83,7 @@ from codalab.lib.completers import (
 from codalab.lib.bundle_store import MultiDiskBundleStore
 from codalab.lib.print_util import FileTransferProgress
 from codalab.worker.file_util import un_tar_directory
+from codalab.worker.download_util import BundleTarget
 
 from codalab.lib.spec_util import generate_uuid
 from codalab.worker.docker_utils import get_available_runtime, start_bundle_container
@@ -517,7 +517,9 @@ class BundleCLI(object):
         supported
         """
         return [
-            self.resolve_target(default_client, default_worksheet_uuid, spec, allow_remote)[2][0]
+            self.resolve_target(default_client, default_worksheet_uuid, spec, allow_remote)[
+                2
+            ].bundle_uuid
             for spec in target_specs
         ]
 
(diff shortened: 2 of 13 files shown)


codalab/lib/download_manager.py
```diff
@@ -57,6 +58,19 @@ def get_target_info(self, uuid, path, depth):
         bundle_state = self._bundle_model.get_bundle_state(uuid)
         # Raises NotFoundException if uuid is invalid
 
+        def check_dependencies(err):
```
bkgoksel: This is the core change

codalab/lib/bundle_cli.py
```diff
@@ -1277,6 +1277,8 @@ def do_download_command(self, args):
 
         # Do the download.
         target_info = client.fetch_contents_info(bundle_uuid, subpath, 0)
+        bundle_uuid = target_info.get('resolved_uuid', bundle_uuid)
```
bkgoksel: These are just the redirects I mentioned

codalab/lib/download_manager.py
```diff
@@ -89,7 +103,10 @@ def get_target_info(self, uuid, path, depth):
                         )
                     )
                 elif 'error_code' in result:
-                    raise http_error_to_exception(result['error_code'], result['error_message'])
+                    try:
+                        raise http_error_to_exception(result['error_code'], result['error_message'])
+                    except NotFoundError as e:
```
bkgoksel: This is the call made to the worker if the bundle is still `RUNNING`. In this case, the call may fail for a variety of reasons, but we care about the `NotFound` case, which means the worker successfully looked for the path and did not find it. Otherwise, we don't want to bother.

codalab/model/bundle_model.py
```diff
@@ -1052,6 +1053,15 @@ def do_update(connection):
             with self.engine.begin() as connection:
                 do_update(connection)
 
+    def get_bundle_dependencies(self, uuid):
```
bkgoksel, line 1056: This just allows the `get_target_info` function to make a lighter DB call to fetch only the dependencies, rather than fetching everything about the bundle.

codalab/lib/download_manager.py
```diff
@@ -57,6 +58,20 @@ def get_target_info(self, uuid, path, depth):
         bundle_state = self._bundle_model.get_bundle_state(uuid)
         # Raises NotFoundException if uuid is invalid
 
+        def check_dependencies(err):
+            # if path not in bundle, check if it matches one of its dependencies
+            child_path_to_dep = {
+                dep.child_path: dep for dep in self._bundle_model.get_bundle_dependencies(uuid)
+            }
+            matching_dep = child_path_to_dep.get(path.split(os.path.sep)[0], None)
```
percyliang: We always want to use '/' no matter what OS we're on, right?  Since otherwise CodaLab markdown isn't portable.

percyliang: Don't need `, None`

codalab/lib/download_manager.py
percyliang: Would be helpful to document the logic more thorougly with examples.

codalab/lib/download_manager.py
```diff
@@ -66,7 +81,7 @@ def get_target_info(self, uuid, path, depth):
             try:
                 return download_util.get_target_info(bundle_path, uuid, path, depth)
             except download_util.PathException as e:
-                raise NotFoundError(str(e))
+                return check_dependencies(NotFoundError(str(e)))
```
percyliang: I think this logic is a bit non-obvious, so documenting would be good.  Also, I wonder if the recursive structure here could be streamlined...I think it could be cleaner to have a separate helper function that shouldn't include the handling of other bundle states.

codalab/rest/bundles.py
```diff
@@ -541,18 +541,22 @@ def _fetch_bundle_contents_blob(uuid, path=''):
...
+            check_bundles_have_read_permission(local.model, request.user, [uuid])
     except NotFoundError as e:
         abort(http.client.NOT_FOUND, str(e))
     except Exception as e:
         abort(http.client.BAD_REQUEST, str(e))
 
     # Figure out the file name.
-    if not path and bundle.metadata.name:
-        filename = bundle.metadata.name
+    bundle_name = local.model.get_bundle(uuid).metadata.name
```
percyliang: Why is a new DB call needed here?  Any performance problems?

codalab/lib/bundle_cli.py
```diff
@@ -1277,6 +1277,8 @@ def do_download_command(self, args):
...
+        resolved_subpath = target_info.get('resolved_path', subpath)
```
percyliang: just use `subpath` to make the naming a bit more symmetric (since `bundle_uuid` gets overridden)

codalab/lib/bundle_cli.py
```diff
@@ -2428,6 +2434,8 @@ def follow_targets(self, client, bundle_uuid, subpaths, from_start=False):
                 # file and if so, initialize the offset.
                 if subpath_is_file[i] is None:
                     target_info = client.fetch_contents_info(bundle_uuid, subpaths[i], 0)
+                    bundle_uuid = info.get('resolved_uuid', bundle_uuid)
+                    subpath = info.get('resolved_path', subpath)
```
percyliang: Is there a way to avoid the proliferation of these redirections on the client side?  Maybe these should be wrapped or pushed into `fetch_contents_info`?  Not sure if that's the right call, but just bringing it up.

codalab/lib/download_manager.py
```diff
@@ -57,6 +58,20 @@ def get_target_info(self, uuid, path, depth):
...
+        def check_dependencies(err):
+            # if path not in bundle, check if it matches one of its dependencies
+            child_path_to_dep = {
+                dep.child_path: dep for dep in self._bundle_model.get_bundle_dependencies(uuid)
+            }
+            matching_dep = child_path_to_dep.get(path.split(os.path.sep)[0], None)
+            if matching_dep:
+                # The path actually belongs to a dependency of this bundle
+                parent_path = matching_dep.parent_path
+                parent_subpath = path.split(os.path.sep)[1:]
```
percyliang: See comment about '/' above

percyliang (changes requested):
Cool, will be great to have this feature in.  My main comment is that it might become hard to think about what a "target" is now, and this logic exists in many places, so I'm wondering if there's a way around that.

codalab/lib/bundle_cli.py
bkgoksel: I didn't override the variable `subpath` here because it is used to print the download source to the console in the CLI. If we overwrote it, `cl down <bundle-1>/<dep-key>/<path>` would print `<dep-bundle>/<path>` in the download progress. But on second thought, that seems reasonable too (I just didn't want to change what download printed at first)

codalab/lib/bundle_cli.py
bkgoksel: Despite all the code lines there's not really much "logic" on the client side here. It would be more aesthetically pleasing to maybe have all the other `contents`-related methods take a unified `target_info` object so the client doesn't have to do any variable extraction from the results of `get_target_info` and just passes its return value directly downstream, or maybe all other contents getting methods may do the `get_target_info` call server-side to double check. I'll look into the viability of either.

codalab/lib/download_manager.py
bkgoksel: Hmm but the rest of the `download_manager`/`download_util` code uses `os.path.sep`. I was trying to be consistent with that.
Also I don't even know if this matters since we're running our server on UNIX anyways.

codalab/rest/bundles.py
bkgoksel: This is just a moved db call, not an additional one. See the removed db call on line 544 of the before-changes side of the diff. I just realized there was a call to get `bundle` here and it seemed like it was used in multiple places but turns out it was only used here. Kind of a drive-by change.

codalab/lib/download_manager.py
bkgoksel: I'll take an attempt at this.

bkgoksel (commented):
I'll work on addressing the higher level concerns.

percyliang (commented):
This is cleaner, thanks!  I'm a bit worried about turning everything into a tuple, and using `[0]` and `[1]` to index things.  What about having a dict `target = {'uuid': ..., 'path': ...}`?  We could also define a "constructor" `Target(uuid, path)` which returns whatever the underlying representation is.

bkgoksel:
Dict didn't work because it was unhashable and we hash targets in one place. However I made it a proper hashable object and it feels pretty clean like this.

codalab/client/json_api_client.py
```diff
@@ -583,26 +584,38 @@ def fetch_interpreted_worksheet(self, worksheet_uuid):
...
+        """
+        request_path = '/bundles/%s/contents/info/%s' % (
+            target.bundle_uuid,
+            urllib.parse.quote(target.subpath),
+        )
         response = self._make_request('GET', request_path, query_params={'depth': depth})
+        # Deserialize the target. See /rest/bundles/_fetch_contents_info for serialization side
+        response['data']['resolved_target'] = BundleTarget.from_dict(
+            response['data']['resolved_target']
+        )
```
bkgoksel, line 601: Here's the ugly side of using a proper object: serialization

codalab/lib/download_manager.py
```diff
@@ -90,25 +132,32 @@ def get_target_info(self, uuid, path, depth):
                     )
                 elif 'error_code' in result:
                     raise http_error_to_exception(result['error_code'], result['error_message'])
-                return result['target_info']
+                target_info = result['target_info']
+                # Deserialize dict response sent over JSON
+                target_info['resolved_target'] = download_util.BundleTarget.from_dict(
+                    target_info['resolved_target']
+                )
```
bkgoksel, line 139: Here's more ugly serialization

codalab/rest/bundles.py
```diff
@@ -395,12 +396,16 @@ def _fetch_bundle_contents_info(uuid, path=''):
...
     if depth < 0:
         abort(http.client.BAD_REQUEST, "Depth must be at least 0")
 
     check_bundles_have_read_permission(local.model, request.user, [uuid])
     try:
-        info = local.download_manager.get_target_info(uuid, path, depth)
+        info = local.download_manager.get_target_info(target, depth)
+        # Object is not JSON serializable so submit its dict in API response
+        # The client is responsible for deserializing it
+        info['resolved_target'] = info['resolved_target'].__dict__
```
bkgoksel, line 408: Here's another place where we send this over the API and need to serialize to dict.

codalab/worker/reader.py
```diff
@@ -82,7 +84,9 @@ def get_target_info(self, run_state, path, args, reply_fn):
             target_info['contents'] = [
                 child for child in target_info['contents'] if child['name'] not in dep_paths
             ]
-
+        # Object is not JSON serializable so submit its dict in API response
+        # The client is responsible for deserializing it
+        target_info['resolved_target'] = target_info['resolved_target'].__dict__
```
bkgoksel, line 89: And the worker-side response serialization

bkgoksel (commented):
With the proper object we need to serialize into a dict before we can send it over the API, I annotated all the places where we manually serialize/deserialize. My thinking is that it's better to serialize as late as possible and deserialize as soon as possible.

codalab/worker/download_util.py
```diff
@@ -5,7 +5,32 @@ class PathException(Exception):
...
 
-def get_target_info(bundle_path, uuid, path, depth):
+class BundleTarget:
+    """
+        bundle_uuid: UUID of the bundle the path is actually found on. This is
+            used for when a path resolves to a dependency of a bundle.
+        subpath: the particular path to resolve in the bundle UUID in the end. If
+            the path resolves to a dependency, then the first component of the
+            path is the dependency key and should not be used within the actual
+            dependency bundle. This field strips that value.
```
percyliang, line 15: Explain this with an example

test_cli.py
```diff
@@ -1686,9 +1691,11 @@ def test(ctx):
     # Basic getting info and blob contents of a bundle
     path = test_path('a.txt')
     uuid = _run_command([cl, 'upload', path])
-    response = ctx.client.fetch_contents_info(uuid)
+    response = ctx.client.fetch_contents_info(BundleTarget(uuid, ''))
     check_equals(response['name'], uuid)
-    check_equals(open(path, 'rb').read(), ctx.client.fetch_contents_blob(uuid, '/').read())
+    check_equals(
+        open(path, 'rb').read(), ctx.client.fetch_contents_blob(BundleTarget(uuid, '/')).read()
```
percyliang: What's the difference between `subpath = ''` versus `subpath = '/'`?

percyliang (approved):
A lot cleaner with `BundleTarget`.  The serialization complexity is unfortunate, but I it's probably a net win.

test_cli.py
bkgoksel: honestly I'm not sure. I don't think `subpath='/'` makes any sense, I didn't really attention to each individual target I encountered and generally wrapped them with the constructor. This can probably be an empty path.

PR #2006, opened 2020-02-18
Upgrade package version due to Travis CI changes

### LINKED ISSUE #2005: Travis CI frontend failure
Travis CI frontend started to fail with the following error: https://travis-ci.org/codalab/codalab-worksheets/jobs/650233192?utm_medium=notification&utm_source=github_status. Need to fix this for deployment.

candicegjing:
Fixed #2005 

Not sure if this is the best way to fix the front end Travis CI failure. `node` version seems to be outdated in the Travis CI build. Updating the `node` version and installing the required package seems to help the build to pass. It's unclear to me why the steps of Travis CI are suddenly different than before. If anyone has any ideas, please advise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ jobs:
       install: pip install --upgrade black==18.9b0
       script: black . --check --diff
     - stage: check
-      language: javascript
+      language: node_js
+      node_js:
+        - 13
       install: cd frontend && npm install
       script:
         - npm run check-ci
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -46,6 +46,7 @@
   "devDependencies": {
     "flow-bin": "^0.91.0",
     "prettier": "^1.19.1",
+    "eslint": "^6.8.0",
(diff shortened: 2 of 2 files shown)


bkgoksel:
There must be a better way to get travis to give us a newer version of npm rather than installing it ourselves. Something like how we configure the python version for other stages.

.travis.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       script: black . --check --diff
     - stage: check
       language: javascript
-      install: cd frontend && npm install
+      install: cd frontend && nvm install node && npm install
```
epicfaace: You can use the `node_js` attribute in `.travis.yml` to specify the node version: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions

candicegjing: Updated!

yipenghe (approved):
Looks like its passing Travis test, lets get it in

PR #2007, opened 2020-02-19
Do not upload results for killed bundles

### LINKED ISSUE #1135: Problem with disk quota enforcement on workers
A couple issues seen around disk requests:

1- When a bundle goes over its disk quota, it is killed by the worker when that is caught, but the worker still uploads all the results, making the disk quota effectively useless.

2- When the disk quota is the same as the user's total disk quota left, this upload step causes a failure with an unparsable (No JSON could be decoded) error message.

candicegjing:
Fixed #1135.

This PR will fix the first issue in #1135: When a bundle goes over its disk quota, it is killed by the worker when that is caught, but the worker still uploads all the results, making the disk quota effectively useless. 

When uploading the results, we might need to check `run_state` to see if the bundle status is killed or it has any failures. If there is a failure, we won't upload the results.

### DIFF
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -449,7 +449,7 @@ class RunStateMachine(StateTransitioner):
             except Exception:
                 logger.error(traceback.format_exc())
 
-        if not self.shared_file_system and run_state.has_contents:
+        if not self.shared_file_system and run_state.has_contents and not run_state.is_killed:
             # No need to upload results since results are directly written to bundle store
             return run_state._replace(
                 stage=RunStage.UPLOADING_RESULTS, run_status='Uploading results', container=None
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -449,7 +449,12 @@ class RunStateMachine(StateTransitioner):
             except Exception:
                 logger.error(traceback.format_exc())
 
-        if not self.shared_file_system and run_state.has_contents and not run_state.is_killed:
+        if (
+            not self.shared_file_system
+            and run_state.has_contents
+            and not run_state.is_killed
+            and not run_state.failure_message
+        ):
             # No need to upload results since results are directly written to bundle store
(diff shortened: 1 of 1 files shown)


percyliang (commented):
We definitely want to still upload contents when there's a failure so that the user can debug.

candicegjing:
> We definitely want to still upload contents when there's a failure so that the user can debug.

Oh, I see. That's true. I removed the logic related to failure message

percyliang (commented):
You still might want to retain the results of killed bundles.  I don't think we strictly need this for correctness, right?

candicegjing:
> You still might want to retain the results of killed bundles. I don't think we strictly need this for correctness, right?

Do you mean results of killed bundles or `kill_message` of killed bundles? For `kill_message`, it will exist during the entire lifecycle of a killed bundle, see this logic [finalize_run](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L523-L530). But for results, I thought the original ticket wants to avoid uploading results when a bundle is killed?

percyliang:
I think we should just try to upload the results, whatever they are, and just make sure the upload doesn't exceed the disk quota and fails gracefully with an error message (Not enough disk space to complete upload).  It seems like the problem right now is that the user can circumvent the disk quota.

candicegjing:
> I think we should just try to upload the results, whatever they are, and just make sure the upload doesn't exceed the disk quota and fails gracefully with an error message (Not enough disk space to complete upload). It seems like the problem right now is that the user can circumvent the disk quota.

I think currently we are able to fail gracefully with an error message reported to UI when disk quota exceeds the user's upper limit. See the following example: (e.g. **Error while uploading: Unable to update bundle contents in bundle service: Bad Request - Upload failed: Can't save bundle, bundle size 10491125 greater than user's disk quota left: 300000.0**). So I'll cross out the first issue in this ticket then.

candicegjing:
Closing this ticket as this feature is not needed anymore.

PR #2009, opened 2020-02-19
Copy/Paste text to raw source with click button / keyboard

### LINKED ISSUE #1802: Copy bundle uuid within and across tables in worksheet

### LINKED ISSUE #1996: Paste bundles within and across worksheets

yipenghe:
continuation of #1975 
(implement) fix #1996 fix #1802 

implement a copy paste mechanism, changing the keyboard shortcut for opening terminal from "c" to "c c" so it does not collide with "a c"


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -778,28 +863,44 @@ class Worksheet extends React.Component {
...
                 this.togglePopupNoEvent('detach');
             });
             Mousetrap.bind(['a k'], () => {
-                if (this.state.openDetach || this.state.openDelete) {
+                if (this.state.openDetach || this.state.openDelete || this.state.openCopy) {
                     return;
                 }
                 this.togglePopupNoEvent('kill');
             });
+            Mousetrap.bind(['a c'], () => {
```
epicfaace, line 963: Can you update the PR description and the "keyboard shortcuts" popup/documentation since you've now implemented `a c` and `a v`?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -418,6 +454,56 @@ class Worksheet extends React.Component {
         this.setState({ deleteItemCallback: callback, openDeleteItem: true });
     };
 
+    pasteToWorksheet = () => {
+        // Unchecks all bundles after pasting
+        console.log('Pasting to worksheet');
```
epicfaace: remove

frontend/src/components/worksheets/WorksheetDialogs.js
```diff
@@ -141,6 +142,38 @@ class WorksheetDialogs extends React.Component {
...
                 </Dialog>
+                <Dialog
+                    open={this.props.openCopy}
+                    onClose={this.props.togglePopup('copy')}
+                    aria-labelledby='copy-title'
+                    aria-describedby='deletion-description'
+                >
+                    <DialogContent className={classes.copyDialog}>
+                        <DialogContentText id='alert-dialog-description'>
+                            The following bundle ids (excluded invalid ones) will be copied to
```
epicfaace:
```suggestion
                            The following bundle ids (excluding invalid ones) will be copied to
```

frontend/src/components/worksheets/WorksheetDialogs.js
```diff
@@ -141,6 +142,38 @@ class WorksheetDialogs extends React.Component {
...
+                            clipboard:
+                            <div style={{ whiteSpace: 'pre-wrap' }}>
+                                {this.props.copiedBundleIds.display}
+                            </div>
+                            You can use "paste" to move the copied bundles
```
epicfaace:
```suggestion
                            You can use "paste" to move the copied bundles.
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -357,20 +342,47 @@ class Worksheet extends React.Component {
         this.togglePopupNoEvent(cmd_type);
     };
 
+    addCopyBundleRowsCallBack = (tableID, callback) => {
```
```suggestion
    addCopyBundleRowsCallback = (tableID, callback) => {
```

Can you also rename `CallBack` to `Callback` everywhere?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -357,20 +342,47 @@ class Worksheet extends React.Component {
...
     togglePopup = (cmd_type) => () => {
         if (cmd_type === 'deleteItem') {
             this.setState({ openDeleteItem: !this.state.openDeleteItem });
         }
         if (!this.state.showBundleOperationButtons) {
             return;
         }
-        const { openKill, openDelete, openDetach } = this.state;
+        const { openKill, openDelete, openDetach, openCopy } = this.state;
         if (cmd_type === 'rm') {
```
epicfaace, line 389: Can you refactor this whole if/else statement to another function (as it gets copied verbatim later on)?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -862,13 +963,8 @@ class Worksheet extends React.Component {
             }
         } else {
             // Go into edit mode.
-            this.setState({
-                editMode: editMode,
-                uuidBundlesCheckedCount: {},
-                checkedBundles: {},
-                showBundleOperationButtons: false,
-                updating: false,
```
epicfaace, line 925: Why remove `updating: false`?

codalab/rest/interpret.py
```diff
@@ -218,8 +230,13 @@ def fetch_interpreted_worksheet(uuid):
                 if not is_relevant_block:
                     interpreted_blocks['blocks'][i] = None
 
+    # items: actual blocks in the worksheets, each item may group several raw items
```
epicfaace: Instead of adding these comments inline, can you instead add information about each key in the response (`raw_to_block`, `block_to_raw`, etc.) in the docstring at the top of this function?

epicfaace (dismissed):

frontend/src/components/worksheets/Worksheet/Worksheet.js
yipenghe: clearCheckedBundles will set updating: false

codalab/rest/interpret.py
```diff
@@ -158,6 +158,26 @@ def fetch_interpreted_worksheet(uuid):
...
+                                bundles will be grouped into table block items
+            raw-list:[list] raw source lines
+            raw_to_block-list:[list] raw_item to block 
+                          for example, assume the first resolved item is a bundle table block that has 2 elements,
+                            then the 2nd element in the list would be [0, 1]
+                            [0, 1]: 0 means the item belongs to the first block,
+                                    1 means the item is the second item of the block (2nd bundle in our example)
+            block_to_raw:[dict] Mostly a reverse mapping of raw_to_block
+                            by "Mostly": for example, 
+                            multiple text item may be grouped into one markdown block, so
```
epicfaace:
```suggestion
                            multiple text items may be grouped into one markdown block, so
```

codalab/rest/interpret.py
```diff
@@ -218,8 +248,13 @@ def fetch_interpreted_worksheet(uuid):
                 if not is_relevant_block:
                     interpreted_blocks['blocks'][i] = None
 
+    # items: actual blocks in the worksheets, each item may group several raw items
+    #        into one block i.e one table item groups multiple bundle raw items
     worksheet_info['items'] = resolve_interpreted_blocks(interpreted_blocks['blocks'])
+    # raw_to_block: an array that records how raw items are grouped into items.
+    # The value is of form [block_index, sub_block_index]
     worksheet_info['raw_to_block'] = interpreted_blocks['raw_to_block']
+    # block_to_raw: inverse mapping of raw_to_block
```
epicfaace: You can remove these inline comments now that you've added this info to the docstring at the beginning.

codalab/rest/interpret.py
```diff
@@ -173,10 +193,20 @@ def fetch_interpreted_worksheet(uuid):
...
     worksheet_info['raw'] = get_worksheet_lines(worksheet_info)
 
-    # Replace searches with raw items.
+    # Expand user-written searches (including bundle & worksheet) to raw items.
     # This needs to be done before get_worksheet_lines because this replaces
     # user-written raw items.
-    worksheet_info['items'] = expand_raw_items(worksheet_info['items'])
+    # Also stores the mapping from raw items to the source line that produced them.
+    # This allows us to manipulate the source lines easier on the frontend.
+    expanded_items_to_raw_lines = []
```
epicfaace: Can we be consistent with naming? either call both raw_item_to_source_line or expanded_items_to_raw_lines?

yipenghe:
@epicfaace so I found that Firefox actually doesn't support reading from clipboard unless it's a browser extension app. I thinking to change to use localStorage/sessionStorage to remember what was copied, this way we don't need a extra dialog anymore, what are your thoughts on this?

percyliang:
It seems like it's nice to actually have it end up on the clipboard, because then it's more interoperable with things outside CodaLab.

yipenghe:
> It seems like it's nice to actually have it end up on the clipboard, because then it's more interoperable with things outside CodaLab.

Copy works for both chrome and firefox, but reading from clipboard is not allowed by firefox. We can use local/session storage to complete the pasting (adding to the raw source lines), and still copy the contents to the clipboard as well

percyliang:
So if I copy a piece of markdown from a tutorial, I can't paste it in?

yipenghe:
> So if I copy a piece of markdown from a tutorial, I can't paste it in?

If we wanna support both Chrome and Firefox, then we can't do that (copy a markdown and paste), because Firefox doesn't allow us to read from the clipboard. I'm looking for some hacks on this

yipenghe:
TODO: in pasteboard, add keyboard shortcut to confirm paste directly in text box

frontend/src/components/worksheets/items/ActionButtons.js
```diff
@@ -133,9 +136,23 @@ class ActionButtons extends React.Component<{
...
+                    <Button
+                        size='small'
+                        color='inherit'
+                        aria-label='Paste'
+                        onClick={toggleCmdDialog('paste')}
+                        disabled={!editPermission}
+                        id='paste-button'
+                    >
+                        <NoteAddIcon className={classes.buttonIcon} />
+                        PasteBoard
```
epicfaace: Can we just call it "Paste"? "Pasteboard" sounds kind of unfamiliar / confusing.

```suggestion
                        Paste
```

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -95,6 +96,21 @@ class TableItem extends React.Component<{
...
     };
+
+    copyCheckedBundleRows = () => {
+        let bundleRowsData = this.props.item.rows;
+        let result = bundleRowsData.filter((item, index) => {
+            return this.state.childrenCheckState[index];
+        });
+        result = result.map((row) => {
+            let bundleIdName = {};
+            bundleIdName.uuid = row['uuid[0:8]'];
```
epicfaace: Will this header always be called `uuid[0:8]`?

docs/REST-API-Reference.md
```diff
@@ -849,6 +849,26 @@ Return information about a worksheet. Calls
...
+        raw_to_block-list:[list] raw_item to block 
+                      for example, assume the first resolved item is a bundle table block that has 2 elements,
+                        then the 2nd element in the list would be [0, 1]
+                        [0, 1]: 0 means the item belongs to the first block,
+                                1 means the item is the second item of the block (2nd bundle in our example)
+        block_to_raw:[dict] Mostly a reverse mapping of raw_to_block
+                        by "Mostly": for example, 
+                        multiple text items may be grouped into one markdown block, so
+                        the mapping skip some raw_item indices
+        expanded_items_to_raw_lines: [list] raw_item to its original source line
```
epicfaace: I'm not clear what expanded_items_to_raw_lines is based on the description. Is it just a list of items? If so, then why is it called "expanded_items_to_raw_lines" (which sounds like a dict)?

frontend/src/components/worksheets/ExtraWorksheetHTML/ExtraWorksheetHTML.js
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Alert, Modal } from 'react-bootstrap';
 import './ExtraWorksheetHTML.scss';
 
-const GlossaryModal = ({ show, toggle }) => (
+const KeyboardShortCutModal = ({ show, toggle }) => (
```
epicfaace: Prefer "Shortcut" over "ShortCut"

```suggestion
const KeyboardShortcutModal = ({ show, toggle }) => (
```

frontend/src/components/worksheets/ExtraWorksheetHTML/ExtraWorksheetHTML.js
```diff
@@ -99,6 +99,12 @@ const GlossaryModal = ({ show, toggle }) => (
                         </td>
                         <td>Edit and add a rerun in bundle details</td>
                     </tr>
+                    <tr>
+                        <td>
+                            <kbd>a v</kbd>
+                        </td>
+                        <td>Paste clipboard content to source after focused line</td>
```
epicfaace:
```suggestion
                        <td>Paste clipboard content to source after the current line</td>
```

epicfaace (commented):
- When I click on "Pasteboard" and paste, it works properly -- but when I try typing in "a v", nothing happens.
- When I focus on a bundle at the top of a table and paste a bundle, the bundle gets added to the bottom of the table instead of right after the focused bundle.
- We should not show the confirmation dialog when the user copies something empty. This is because, otherwise, it can cause confusing behavior if a user, say, focuses on a bundle row (but doesn't check the box) and presses "a c", and then gets an empty prompt saying "The following bundle ids (excluding invalid ones) will be copied to clipboard:"
- The keyboard shortcut for copy is "a c", but the keyboard shortcut for showing the web terminal is also "c". Thus, a lot of the time in which I am trying to type "a c" to copy something, the web terminal just shows up instead.
- This last point isn't a bug, but more of a general suggestion. I'm not quite comfortable with the idea of a Pasteboard. Isn't the whole idea of the visual editor that you _don't_ have to deal with source lines? Yet in the normal usage of copy/paste, you will have situations like this:

image

In effect, we are having a mini-"Edit Source" feature that is being added here to the visual editor. I think source lines have their place, but that's only in the "Edit Source" page -- the visual editor should try to avoid them if at all possible.

I think it would be good if we could find a more seamless way to copy/paste in which the user _**does not see the source lines at all**_ -- perhaps they could view a preview of the bundle table in the dialog (just like it shows up in the worksheet) instead of the raw source lines?

frontend/src/components/worksheets/items/TableItem/TableItem.js
yipenghe: Good catch, I changed to use bundleInfos which always contain 'uuid'

docs/REST-API-Reference.md
yipenghe: This should be addressed with #2145

yipenghe:
> * When I click on "Pasteboard" and paste, it works properly -- but when I try typing in "a v", nothing happens.
> * When I focus on a bundle at the top of a table and paste a bundle, the bundle gets added to the bottom of the table instead of right after the focused bundle.
> * We should not show the confirmation dialog when the user copies something empty. This is because, otherwise, it can cause confusing behavior if a user, say, focuses on a bundle row (but doesn't check the box) and presses "a c", and then gets an empty prompt saying "The following bundle ids (excluding invalid ones) will be copied to clipboard:"
> * The keyboard shortcut for copy is "a c", but the keyboard shortcut for showing the web terminal is also "c". Thus, a lot of the time in which I am trying to type "a c" to copy something, the web terminal just shows up instead.
> * This last point isn't a bug, but more of a general suggestion. I'm not quite comfortable with the idea of a Pasteboard. Isn't the whole idea of the visual editor that you _don't_ have to deal with source lines? Yet in the normal usage of copy/paste, you will have situations like this:
> 
> image
> 
> In effect, we are having a mini-"Edit Source" feature that is being added here to the visual editor. I think source lines have their place, but that's only in the "Edit Source" page -- the visual editor should try to avoid them if at all possible.
> 
> I think it would be good if we could find a more seamless way to copy/paste in which the user _**does not see the source lines at all**_ -- perhaps they could view a preview of the bundle table in the dialog (just like it shows up in the worksheet) instead of the raw source lines?

Thanks for all the comments. After changing to use localStorage, everything becomes much cleaner and simpler:
1. No more extra dialogs to confirm copy and paste, I just added a toast info after copying. Since we no longer overwrite users' clipboard, it's also more safer
2. No more source lines! All we do now is select the bundles, copy them, move to the spot we want to paste and then paste
3. since 'c' naturally associates with copy, I just changed the open full terminal shortcut to 'c c' (hit c twice in a row), so it won't trigger the terminal that easily any more

yipenghe:
@percyliang At the end I decided to go with local storage, which Ashwin also agrees

Here are some thoughts:

- Supporting bundle copy/cut/paste is what we initially want to support and is the most important thing for this feature. 

- While it would be great to be able to paste Text directly as well, it's not possible to do it directly at this point since firefox restricts the access to user clipboard.
- Thinking about text, one can always click "text" and paste it there, so probably not much value supporting text copy with a different mechanism. (With the dialog approach, we are basically opening another text box when pasting text)

- Moreover, we don't interfere with user clipboard, which I think is a plus as they don't need to worry about having the current clipboard overwritten.

- Last, the process is much simpler since we don't need to open a confirmation dialog

yipenghe:
> * When I click on "Pasteboard" and paste, it works properly -- but when I try typing in "a v", nothing happens.
> * When I focus on a bundle at the top of a table and paste a bundle, the bundle gets added to the bottom of the table instead of right after the focused bundle.
> * We should not show the confirmation dialog when the user copies something empty. This is because, otherwise, it can cause confusing behavior if a user, say, focuses on a bundle row (but doesn't check the box) and presses "a c", and then gets an empty prompt saying "The following bundle ids (excluding invalid ones) will be copied to clipboard:"
> * The keyboard shortcut for copy is "a c", but the keyboard shortcut for showing the web terminal is also "c". Thus, a lot of the time in which I am trying to type "a c" to copy something, the web terminal just shows up instead.
> * This last point isn't a bug, but more of a general suggestion. I'm not quite comfortable with the idea of a Pasteboard. Isn't the whole idea of the visual editor that you _don't_ have to deal with source lines? Yet in the normal usage of copy/paste, you will have situations like this:
> 
> image
> 
> In effect, we are having a mini-"Edit Source" feature that is being added here to the visual editor. I think source lines have their place, but that's only in the "Edit Source" page -- the visual editor should try to avoid them if at all possible.
> 
> I think it would be good if we could find a more seamless way to copy/paste in which the user _**does not see the source lines at all**_ -- perhaps they could view a preview of the bundle table in the dialog (just like it shows up in the worksheet) instead of the raw source lines?

About the adding to the end problem, I think it's copies it into the right place unless you are focusing on a bundle row generated by a search. There is focus issue that is not specific to copy/paste: when you add a new run, it adds after the current focused bundle row, but the focus then moves to the last row of the  bundle table. I think it's common focus bug and would be better to address in a different PR.

yipenghe:
turns out that block_to_raw is an accurate mapping from the blocks to the source lines, the only thing that is buggy and in accurate is raw_to_block. I added a mechanism to track which block is a search-expanded one so it would be able to use the right indexing.

Can @epicfaace review?

epicfaace:
- when you copy, we should deselect the checkbox immediately (because if you paste right afterwards and the checkbox is still checked, it's confusing)
- when you paste, we shouldn't clear the clipboard
- I think this is still kind of buggy -- for example, I tried having `markdown - markdown - bundle table` and pasted while selecting the first markdown block, then it added the bundle to the bundle table instead of creating a new bundle table after the first markdown block.

https://github.com/codalab/codalab-worksheets/pull/2190 is also ready for you to use with copy/paste, if we want to use it.

yipenghe:
> * when you copy, we should deselect the checkbox immediately (because if you paste right afterwards and the checkbox is still checked, it's confusing)
> * when you paste, we shouldn't clear the clipboard
> * I think this is still kind of buggy -- for example, I tried having `markdown - markdown - bundle table` and pasted while selecting the first markdown block, then it added the bundle to the bundle table instead of creating a new bundle table after the first markdown block.
> 
> #2190 is also ready for you to use with copy/paste, if we want to use it.

I think the reason is the add-item end point doesn't update the raw_to_block properly, in this case I will try to use the add-item endpoint, but cut and schema gets trickier

frontend/src/components/worksheets/ExtraWorksheetHTML/ExtraWorksheetHTML.js
```diff
@@ -121,7 +133,7 @@ const GlossaryModal = ({ show, toggle }) => (
                     </tr>
                     <tr>
                         <td>
-                            <kbd>c</kbd>
+                            <kbd>c c</kbd>
```
epicfaace, line 136: We should probably mention in the PR body and release notes that this keyboard shortcut has changed.

cc @percyliang

frontend/src/components/worksheets/ExtraWorksheetHTML/ExtraWorksheetHTML.js
```diff
@@ -177,12 +189,13 @@ let ExtraWorksheetHTML = ({
         <div id='update_progress' className='progress-message'>
             <img src='/img/Preloader_Small.gif' /> Updating...
         </div>
+        {/*TODO: Move all error message to worksheet dialog */}
```
epicfaace:
```suggestion
        {/* TODO: Move all error messages to worksheet dialog */}
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -446,6 +545,7 @@ class Worksheet extends React.Component {
                     blocks: Object.assign([], this.state.ws.info.blocks, { [focusIndex]: item }),
                 },
             },
+            searchExpandedBlock: searchBlocks,
```
epicfaace: What is `searchExpandedBlock` used for? I don't see it used anywhere in the code.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -436,7 +465,77 @@ class Worksheet extends React.Component {
...
+        const data = JSON.parse(window.localStorage.getItem('CopiedBundles'));
+        let bundleString = '';
+        let items = [];
+        data.forEach((bundle) => {
+            bundleString += '[]{' + bundle.uuid + '}\n';
+            items.push(bundle.uuid);
+        });
+        // remove the last new line character
+        bundleString = bundleString.substr(0, bundleString.length - 1);
+        let worksheetuuid = this.state.ws.uuid;
```
epicfaace: For consistency
```suggestion
        let worksheetUUID = this.state.ws.uuid;
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -436,7 +465,77 @@ class Worksheet extends React.Component {
...
+        let after_sort_key;
+        if (this.state.focusIndex !== -1 && this.state.focusIndex !== undefined) {
+            let currentFocusedBlock = this.state.ws.info.blocks[this.state.focusIndex];
+            console.log(this.state.subFocusIndex, currentFocusedBlock);
+            after_sort_key = getAfterSortKey(currentFocusedBlock, this.state.subFocusIndex);
+        }
+        let url = `/rest/worksheets/${worksheetuuid}/add-items`;
```
```suggestion
        let url = `/rest/worksheets/${worksheetUUID}/add-items`;
```

frontend/src/components/worksheets/items/PlaceholderItem.js
```diff
@@ -58,7 +58,7 @@ export default forwardRef((props, ref) => {
     if (error) {
         return <div ref={ref}>Error loading item.</div>;
     }
-    if (item === null) {
+    if (!item) {
```
epicfaace: Won't this change cause the placeholder to always say "No results found" (because `item` is initially undefined)? Let's change it back to

```suggestion
    if (item === null) {
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
yipenghe: Good catch, it was used before the sort key change

epicfaace (approved):

PR #2023, opened 2020-02-21
Disk quota failure

### LINKED ISSUE #973: Users with very high disk/time quotas have their runs fail
Now that we set the `request_disk` and `request_time` of runs to be whatever the user has left in their quota if they don't specify a particular value, if a user has a very high quota, their runs will request very high amounts of disk/time. 

Right now we have global per-bundle maximums and these users will have all their runs fail if their quota is higher than these. If we want to keep these global per-run maximums, we should set the default value to be min(user quota left, per-run maximum). Another solution would be to get rid of per-run maximums.

candicegjing:
Fixed #973 
The only thing is that when we add DISK_QUOTA_OFFSET_BYTES in calculating the max disk quota allowed, the reporting will also contain this offset.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -28,6 +28,9 @@ BUNDLE_TIMEOUT_DAYS = 60
 # When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
 # access to the host machine's memory, which we have decided to not allow
 MINIMUM_REQUEST_MEMORY_BYTES = 4 * 1024 * 1024
+# Deduct DISK_QUOTA_SLACK_BYTES from the max user disk quota bytes when computing the default amount of disk space to
+# request. Then the default max disk quota that can be requested becomes disk quota left - DISK_QUOTA_SLACK_BYTES.
+DISK_QUOTA_SLACK_BYTES = 0.5 * 1024 * 1024 * 1024
 
 
 class BundleManager(object):
@@ -660,11 +663,23 @@ class BundleManager(object):
         """
         if value:
             if user_max and value > user_max:
-                return user_fail_string % (pretty_print(value), pretty_print(user_max))
+                return user_fail_string % (
+                    pretty_print(value),
+                    pretty_print(user_max),
+                    pretty_print(value - user_max),
+                )
             elif global_max and value > global_max:
-                return global_fail_string % (pretty_print(value), pretty_print(global_max))
+                return global_fail_string % (
+                    pretty_print(value),
+                    pretty_print(global_max),
+                    pretty_print(value - global_max),
+                )
             elif global_min and value < global_min:
-                return global_fail_string % (pretty_print(value), pretty_print(global_min))
+                return global_fail_string % (
+                    pretty_print(value),
+                    pretty_print(global_min),
+                    pretty_print(global_min - value),
+                )
         return None
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


codalab/server/bundle_manager.py
```diff
@@ -707,8 +721,13 @@ def _get_staged_bundles_to_run(self, workers, user_info_cache):
             failures.append(
                 self._check_resource_failure(
                     bundle_resources.disk,
-                    user_fail_string='Requested more disk (%s) than user disk quota left (%s)',
-                    user_max=self._model.get_user_disk_quota_left(bundle.owner_id, user_info),
+                    user_fail_string='Requested more disk (%s) than user disk quota left (%s) by %s bytes',
```
percyliang: Can we pretty print the difference too instead of using bytes?

codalab/server/bundle_manager.py
```diff
@@ -28,6 +28,8 @@
 # When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
 # access to the host machine's memory, which we have decided to not allow
 MINIMUM_REQUEST_MEMORY_BYTES = 4 * 1024 * 1024
+# Offset max user disk quota bytes to 0.5G when computing the default amount of disk space to request
```
percyliang: Let's call it "slack" instead of "offset" to convey more of the meaning.  Update the comment to explain this better (The default disk request is disk quota left - slack.)

percyliang (changes requested):

codalab/server/bundle_manager.py
candicegjing: Yes, we can. When using pretty print, the final error message won't contain the unit `bytes`, which could be a little bit confusing. See the logic here: [size_str](https://github.com/codalab/codalab-worksheets/blob/b8b2efed243150dfd8398d4a6136b3d351bcb9fa/codalab/lib/formatting.py#L42). We could add `b` to `size_str` function, or if we think keeping the formatting as it's is fine, I'll update the function to use pretty print.

codalab/server/bundle_manager.py

codalab/server/bundle_manager.py
candicegjing: @percyliang Want to double-check with you on ^ options. Let me know!

codalab/server/bundle_manager.py
```diff
@@ -28,6 +28,9 @@
 # When using the REST api, it is allowed to set Memory to 0 but that means the container has unbounded
 # access to the host machine's memory, which we have decided to not allow
 MINIMUM_REQUEST_MEMORY_BYTES = 4 * 1024 * 1024
+# Deduct 0.5G from the max user disk quota bytes when computing the default amount of disk space to request.
```
percyliang: be consistent in comments: Deduct DISK_QUOTA_SLACK_BYTES

codalab/server/bundle_manager.py
```diff
@@ -707,8 +722,14 @@ def _get_staged_bundles_to_run(self, workers, user_info_cache):
             failures.append(
                 self._check_resource_failure(
                     bundle_resources.disk,
-                    user_fail_string='Requested more disk (%s) than user disk quota left (%s)',
-                    user_max=self._model.get_user_disk_quota_left(bundle.owner_id, user_info),
+                    user_fail_string='Requested more disk (%s) than user disk quota left (%s) by %s',
+                    # Return the larger number between 1.0 and disk quota after deducting
+                    # DISK_QUOTA_SLACK_BYTES to avoid float overflow
```
percyliang: How does max'ing with 1 deal with float overflow?  Shouldn't be necessary?

candicegjing: After deducting 0.5G from the original max disk quota (e.g. 0.1G - 0.5G), the result could become a negative value, which is probably not what we want when setting the `user_max` value.

percyliang: If the value is negative, then the job will fail anyway...  I'd just remove the max if the behavior is correct to keep things simple.

percyliang (approved):

PR #2024, opened 2020-02-21
Allow string and bytes when processing urllib.error.HTTPError object

### LINKED ISSUE #2012: The data type of urllib.error.HTTPError object could vary between String and bytes
Case 1: when a user types in a wrong password, which causes authentication error, the bundle_service_client.py throws an error complaining about the data type of [e.read](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/bundle_service_client.py#L25). From the logging message listed below, the data type of `e.read()` is bytes.
```
stanford@Stanfords-MacBook-Pro codalab-worksheets % cl-worker --server=http://localhost                                                                        
Username: codalab
Password: 
[code] = <class 'int'>, 401
[msg] = <class 'str'>, Unauthorized
[e.read] = <class 'bytes'>, b'{"error": "invalid_grant", "error_description": "Invalid credentials given."}'
[e.code] = <class 'int'>, 401
  File "/Users/stanford/PycharmProjects/codalab-worksheets/codalab/worker/bundle_service_client.py", line 18, in wrapper
    return f(*args, **kwargs)
  File "/Users/stanford/PycharmProjects/codalab-worksheets/codalab/worker/bundle_service_client.py", line 105, in _authorize
    with closing(urllib.request.urlopen(request)) as response:
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 641, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 651, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 401: Unauthorized

  File "/usr/local/bin/cl-worker", line 11, in <module>
    load_entry_point('codalab', 'console_scripts', 'cl-worker')()
  File "/Users/stanford/PycharmProjects/codalab-worksheets/codalab/worker/main.py", line 162, in main
    bundle_service = connect_to_codalab_server(args.server, args.password_file)
  File "/Users/stanford/PycharmProjects/codalab-worksheets/codalab/worker/main.py", line 151, in connect_to_codalab_server
    bundle_service = BundleServiceClient(server, username, password)
  File "/Users/stanford/PycharmProjects/codalab-worksheets/codalab/worker/bundle_service_client.py", line 78, in __init__
    self._authorize()
  File "/Users/stanford/PycharmProjects/codalab-worksheets/codalab/worker/bundle_service_client.py", line 35, in wrapper
    message + ': ' + http.client.responses[e.code] + ' - ' + client_error,
TypeError: can only concatenate str (not "bytes") to str
```

Case 2: in function `upload_with_chunk_encoding`: [StringIO(response.read().decode())](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/rest_client.py#L160-L165) will be received as String in `e.read()`, which causes discrepancy in data type between case 1.  This error will be thrown when a user's disk utilization exceeds the upper limit of the disk quota allowed.

candicegjing:
Fixed #2012 

For this dynamic typing issue: bytes vs string. Since the error could be thrown from any place (e.g. case 1 in #2012 is thrown from a low-level python library), I am really not sure what's the best way to handle this when parsing the urllib.error.HTTPError. If anyone has any ideas, please do let me know.

### DIFF
--- a/codalab/worker/bundle_service_client.py
+++ b/codalab/worker/bundle_service_client.py
@@ -8,7 +8,7 @@ import time
 import urllib.request, urllib.parse, urllib.error
 
 from .rest_client import RestClient, RestClientException
-from .file_util import tar_gzip_directory
+from .file_util import tar_gzip_directory, BINARY_PLACEHOLDER
 
 
 def wrap_exception(message):
@@ -22,7 +22,8 @@ def wrap_exception(message):
                 raise BundleServiceException(message + ': ' + str(e), e.client_error)
             except urllib.error.HTTPError as e:
                 try:
-                    client_error = e.read()
+                    # Ensure the type of urllib.error.HTTPError response to be string
+                    client_error = ensure_str(e.read())
                     if e.reason == 'invalid_grant':
                         raise BundleAuthException(
                             message + ': ' + http.client.responses[e.code] + ' - ' + client_error,
@@ -41,6 +42,20 @@ def wrap_exception(message):
         return wrapper
 
     return decorator
+
+
+def ensure_str(response):
+    """
+    Ensure the data type of input response to be string
+    :param response: a response in bytes or string
+    :return: the input response in string
+    """
+    if isinstance(response, str):
+        return response
+    try:
+        return response.decode()
+    except UnicodeDecodeError:
+        return BINARY_PLACEHOLDER
 
 
 class BundleAuthException(RestClientException):


codalab/worker/bundle_service_client.py
```diff
@@ -43,6 +43,10 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def parse_dynamic_error_type(error):
```
percyliang: This function should be called `ensure_str` (if you need a function at all) since it's generic.  Document what the function does, and more generally why it is needed.

percyliang (changes requested):

codalab/worker/bundle_service_client.py
```diff
@@ -43,6 +43,10 @@ def wrapper(*args, **kwargs):
...
+    return error if isinstance(error, str) else error.decode()
```
epicfaace: Looks good. I remember I had initially had this in the python 2->3 migration but removed it because I thought it was unnecessary (and wanted to avoid the str-bytes multiple types case) -- but it appears that in this case, we have to do this for HTTPError

epicfaace (commented):
Is it at all possible to make a test that throws an HTTPError with bytes, so we can make sure this issue is covered by our tests?

codalab/worker/bundle_service_client.py
candicegjing: Yes, I think this is mainly due to we manually throw an HTTPError in our code base. I tried to look for methods to send bytes instead through `urllib.error.HTTPError`, but haven't found anything useful yet.

epicfaace (approved):

codalab/worker/bundle_service_client.py
```diff
@@ -43,6 +43,15 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def ensure_str(error):
+    '''
+    Ensure the type of urllib.error.HTTPError response to be string
```
percyliang: Comment should belong at the call site...this function is completely generic.

codalab/worker/bundle_service_client.py
```diff
@@ -43,6 +43,15 @@ def wrapper(*args, **kwargs):
...
+    :param error: urlbli.error.HTTPError response
+    :return: HTTPError response in string type
+    '''
+    return error if isinstance(error, str) else error.decode()
```
percyliang: Why would the `decode()` work if the underlying data is `bytes`?

candicegjing: `bytes.deocde()` will change to `string`

codalab/worker/bundle_service_client.py
```diff
@@ -43,6 +44,19 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def ensure_str(response):
+    """
+    Ensure the type of response to be string
+    :param response: an response in bytes or string
+    :return: the input response in string type
+    """
+    try:
```
percyliang, line 55: Rewrite: put `if isinstance(response, str): return response` as the first line.  Then just return `response.decode()` inside the `try`.

codalab/worker/bundle_service_client.py
```diff
@@ -43,6 +44,21 @@ def wrapper(*args, **kwargs):
...
+def ensure_str(response):
+    """
+    Ensure the data type of input response to be string
+    :param response: a response in bytes or string
+    :return: the input response in string
+    """
+    if isinstance(response, str):
+        return response
+    try:
+        response = response.decode()
```
epicfaace: You can just do `return response.decode()` here

epicfaace (approved):

codalab/worker/bundle_service_client.py
candicegjing: Updated!

PR #2002, opened 2020-02-17
547: Test Worksheets (Backend)

teetone:
Fixes #547 

- Use regex to validate the output of `cl print` for the sample worksheet
- Hook it up to `test_cli`

I had to modify some of the example search criteria in order to not add any new dependencies (e.g. search for `ready` runs instead of `failed` runs).

PR #2028, opened 2020-02-23
2001: Set up frontend automation using Selenium

### LINKED ISSUE #2001: [Frontend] Test Malformed Worksheets
Set up a mechanism to test frontend worksheets.

teetone:
Resolves #2001 .

- Set up Selenium to test CodaLab in Chrome and Firefox for Travis. Use `diffimg` for screenshot comparisons.
- Added screenshots test for small sample worksheet. The new test takes ~20 seconds to run (Chrome and Firefox combined).
- Included a `README.md` with instructions on how to maintain existing tests and add new tests

I will add more tests in a separate PR for #2019 .


.travis.yml
```diff
@@ -29,13 +29,27 @@ jobs:
       install:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || [ "$TRAVIS_PULL_REQUEST_BRANCH" != "" ] || echo "--push")
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
+        - pip install -r requirements.txt
+      before_script:
+        # For testing CodaLab in Chrome
```
epicfaace: should we add these scripts to a separate file (such as `test-setup.sh`) so that it's easy to set up and run Selenium tests locally?

.travis.yml
```diff
@@ -29,13 +29,27 @@ jobs:
       install:
         - python3 codalab_service.py build services --version ${TRAVIS_BRANCH} --pull $([ -z "${CODALAB_DOCKER_USERNAME}" ] || [ "$TRAVIS_PULL_REQUEST_BRANCH" != "" ] || echo "--push")
         - python3 codalab_service.py start --services default monitor --version ${TRAVIS_BRANCH}
+        - pip install -r requirements.txt
```
epicfaace, line 32: Why do we need this line?

.travis.yml
```diff
@@ -29,13 +29,27 @@ jobs:
...
+        - tar -xzf geckodriver-v0.21.0-linux64.tar.gz -C geckodriver
+        - export PATH=$PATH:$PWD/geckodriver
       script:
         - python3 codalab_service.py test --version ${TRAVIS_BRANCH} default
           # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization)
         - docker restart codalab_worker_1 && python3 codalab_service.py test run --version ${TRAVIS_BRANCH}
         - docker stop codalab_worker_1
         - python3 codalab_service.py start --services worker --version ${TRAVIS_BRANCH} --shared-file-system
         - python3 codalab_service.py test --version ${TRAVIS_BRANCH} --shared-file-system basic run read write kill resources netcat netcurl workers
+        - python3 tests/ui/ui_tester.py --headless
```
epicfaace, line 43: How long do these tests run? If it takes very long, it may be worth running the ui tests in a parallel job to running the `test_cli.py` tests.

tests/ui/ui_tester.py
```diff
@@ -0,0 +1,173 @@
...
+        self._driver.close()
+
+        # Test Chrome
+        options = ChromeOptions()
+        add_headless(options)
+        self._driver = webdriver.Chrome(chrome_options=options)
+        self.test()
+        self._driver.close()
+
+    def login(self, username='codalab', password='codalab'):
```
epicfaace: I think we define the default username and password somewhere else in another file -- could we import those here?

tests/ui/ui_tester.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class WorksheetTest(UITester):
+    def __init__(self):
+        super().__init__('worksheet')
+
+    def test(self):
+        self.login()
+        self.wait_until_worksheet_loads()
+        self.click_link('Small Worksheet [cl_small_worksheet]')
```
epicfaace, line 159: Where in the code is the "Small Worksheet" created?

.travis.yml
teetone: I need Selenium and diffimg to be installed.

.travis.yml
teetone: It runs for about ~20 seconds.

.travis.yml
teetone: That's a good idea. I can put it in a separate script.

tests/ui/ui_tester.py
teetone: Do you know which file I can look into? Also, I left these as parameters we can pass in, just in case, we wanted to test creating a new user account.

tests/ui/ui_tester.py
teetone: It's in `create_sample_worksheet.py`. Because of the changes in https://github.com/codalab/codalab-worksheets/pull/2002, it will already be available in the Travis  environment.

tests/ui/ui_tester.py
epicfaace: It appears they are hardcoded here: https://github.com/codalab/codalab-worksheets/blob/5dfbacf6a544403cfb4f399a5b01657ba28b7870/.travis.yml#L21

Perhaps you could set the default to `os.getenv("CODALAB_USERNAME")` and `os.getenv("CODALAB_PASSWORD")`.

tests/ui/README.md
```diff
@@ -0,0 +1,64 @@
...
+[diffimg](https://pypi.org/project/diffimg/#history) for our screenshots comparison tool. Currently, our front end 
+automation only runs in Chrome and Firefox. We hope to support more browsers in the future. 
+
+### Setting up the tests on your local machine
+
+#### One-time setup
+
+1. If you haven't already, install the Python dependencies specified in `requirements.txt`, located in the root of the 
+repository, by running `pip install -r requirements.txt`.
+2. Next, install [ChromeDriver](https://chromedriver.chromium.org/getting-started) (for Chrome) and
```
epicfaace, line 13: Can we just ask them to run `./scripts/test-setup.py` instead?

tests/ui/README.md
```diff
@@ -0,0 +1,64 @@
...
+3. Start the UI tests by running `python tests/ui/ui_tester.py`. You can also pass in an additional argument 
+`--headless`, which will run the tests in headless mode. 
+
+### Maintaining the baseline images
+
+If the screenshot tests are failing, check the `diff` folder for the diff images that are generated when image 
+comparisons fail. If the output images is as expected, accept them as the new baselines by replacing the stale baseline 
+images, located in `baselines`, with the newly generated images in `out`. If the failures are not expected, look into 
+your code changes, as they may have cause the tests to fail.
+
```
epicfaace, line 35: Add a section about thresholds and how to change them if needed?

tests/ui/README.md
teetone: There are even easier ways to install them depending on your local machine. `test-setup.py` is mainly to get it to work on Travis. I guess I can mention both ways.

tests/ui/README.md
teetone: The threshold is already at an upper bound of 9%. I don't really want people changing it. I will add a comment in the code.

tests/ui/README.md
teetone: @epicfaace I actually updated the `test-setup.py` so local installation is a bit easier and still compatible with Travis. I updated the doc too.

tests/ui/README.md
```diff
@@ -11,8 +11,12 @@ automation only runs in Chrome and Firefox. We hope to support more browsers in
 1. If you haven't already, install the Python dependencies specified in `requirements.txt`, located in the root of the 
 repository, by running `pip install -r requirements.txt`.
 2. Next, install [ChromeDriver](https://chromedriver.chromium.org/getting-started) (for Chrome) and 
-[GeckoDriver](https://github.com/mozilla/geckodriver) (for Firefox). These drivers are the links between the selenium 
+[GeckoDriver](https://github.com/mozilla/geckodriver) (for Firefox). These drivers are the links between the Selenium 
 tests and their respective browsers.
+    1. Run `./scripts/test-setup.sh` and place the installed GeckoDriver on your path by running 
+    `export PATH=$PATH:/usr/bin/geckodriver`. This is not needed for ChromeDriver.
+    3. An alternative is to simply use [Homebrew](https://brew.sh/) and run `brew install geckodriver && brew cask
```
epicfaace:
```suggestion
    1. An alternative is to simply use [Homebrew](https://brew.sh/) and run `brew install geckodriver && brew cask 
```

teetone: Ah, thanks for catching that.

tests/ui/README.md
```diff
@@ -0,0 +1,68 @@
...
+### Setting up the tests on your local machine
+
+#### One-time setup
+
+1. If you haven't already, install the Python dependencies specified in `requirements.txt`, located in the root of the 
+repository, by running `pip install -r requirements.txt`.
+2. Next, install [ChromeDriver](https://chromedriver.chromium.org/getting-started) (for Chrome) and 
+[GeckoDriver](https://github.com/mozilla/geckodriver) (for Firefox). These drivers are the links between the Selenium 
+tests and their respective browsers.
+    1. Run `./scripts/test-setup.sh` and place the installed GeckoDriver on your path by running
```
epicfaace: You'd need to run it with `sudo` locally.
```suggestion
    1. Run `sudo ./scripts/test-setup.sh` and place the installed GeckoDriver on your path by running 
```

scripts/test-setup.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# For testing CodaLab in Chrome
+wget https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip
+unzip chromedriver_linux64.zip
+sudo mv chromedriver /usr/bin/chromedriver
+sudo chown root:root /usr/bin/chromedriver
+sudo chmod +x /usr/bin/chromedriver
+
```
epicfaace: Not super important, but it would be nice for setup if both zip files were deleted at the end.

tests/ui/README.md
```diff
@@ -0,0 +1,68 @@
...
+    1. Run `./scripts/test-setup.sh` and place the installed GeckoDriver on your path by running 
+    `export PATH=$PATH:/usr/bin/geckodriver`. This is not needed for ChromeDriver.
+    2. An alternative is to simply use [Homebrew](https://brew.sh/) and run `brew install geckodriver && brew cask 
+    install chromedriver`. 
+
+#### Running the tests locally
+
+1. From the root of the repository, start the CodaLab service by running `./codalab_service.py start -bd`.
+2. Populate your local machine with a comprehensive sample worksheet by running 
+`docker exec -it codalab_rest-server_1 /bin/bash -c "python3 scripts/create_sample_worksheet.py"`.
```
epicfaace, line 25: Instead of this, can we just ask them to run the command that runs the `worksheets` module in `test_cli`? This way, the local setup and travis setup will be more uniform.

tests/ui/README.md
```diff
@@ -0,0 +1,68 @@
...
+    `export PATH=$PATH:/usr/bin/geckodriver`. This is not needed for ChromeDriver.
+    2. An alternative is to simply use [Homebrew](https://brew.sh/) and run `brew install geckodriver && brew cask 
+    install chromedriver`. 
+
+#### Running the tests locally
+
+1. From the root of the repository, start the CodaLab service by running `./codalab_service.py start -bd`.
+2. Populate your local machine with a comprehensive sample worksheet by running 
+`docker exec -it codalab_rest-server_1 /bin/bash -c "python3 scripts/create_sample_worksheet.py"`. 
+3. Start the UI tests by running `python tests/ui/ui_tester.py`. You can also pass in an additional argument
```
epicfaace, line 26: I wasn't able to get this working locally on my computer (got issue - "Message: Unable to find a matching set of capabilities") but will need to debug that further. It works on travis anyway, which should be enough.

epicfaace (approved):

tests/ui/README.md
teetone: If I want to run and target a specific test, do I run `docker exec -it codalab_rest-server_1 /bin/bash -c "python3 test_cli.py worksheets"`? This seems to be failing for me on my local machine.

teetone: I think the current instruction on creating a sample worksheet is quick and should be good enough for now.

tests/ui/README.md
teetone: It's important that people set these test up on their development environment as well. Did `test-setup.sh` run successfully for you? Can you check if ChromeDriver and GeckoDriver are installed?

teetone: I tried again locally and it seems to work for me. I can merge in the code and have others try it out. Based on feedback, I can update the documentation on a follow-up PR, where I need to add more tests anyways.

PR #2031, opened 2020-02-24
Tag public workers

### LINKED ISSUE #2025: [FEATURE REQUEST] Tag for CodaLab public instances

I'd like a way to force my jobs to be run on the CodaLab public Azure instances. Right now, the only way to make this work is to shut down all the other existing workers.

you could do `--request-queue tag=public`, but I'm not sold on the name.

candicegjing:
Fixed #2025.

Currently, I only added logic to handle the use case for `tag=codalab-public`. When a user specifies `tag=public-codalab` when submitting a bundle, the bundle will be routed to CodaLab's public instances even if the user has private workers connected to CodaLab server.

In this PR, I moved the logic of filtering workers by request queue to the earliest point so that we only need to focus on deducting resources on valid workers instead of including those workers that might not be used at the end.     
The dispatching logic will do **exact matching** between **bundle.metadata.request_queue** with **worker's tag**. This means that there will no rerouting even if there are available workers out there for the current bundle owner.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -313,14 +313,17 @@ class BundleManager(object):
 
     def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_info_cache):
         """
-        Schedules STAGED bundles to run on the given workers. Always tries to schedule bundles to
-        run on workers that are owned by the owner of each bundle first. If there are no such qualified
-        private workers, uses CodaLab-owned workers, which have user ID root_user_id.
+        Schedule STAGED bundles to run on available workers based on the following logic:
+        1. If the bundle requests to run on a specific worker, tries to schedule the bundle
+           to run on a worker that has a tag exactly match with request_queue.
+        2. If the bundle doesn't request to run on a specific worker,
+          (1) try to schedule the bundle to run on a worker that belongs to the bundle's owner
+          (2) if there is no such qualified private worker, uses CodaLab-owned workers, which have user ID root_user_id.
         :param workers: a WorkerInfoAccessor object containing worker related information e.g. running uuid.
         :param staged_bundles_to_run: a list of tuples each contains a valid bundle and its bundle resources.
         :param user_info_cache: a dictionary mapping user id to user information.
         """
-        # Reorder the stage_bundles so that bundles which were requested to run on a personal worker
+        # Reorder the stage_bundles so that bundles which were requested to run on a specific worker
         # will be scheduled to run first
         staged_bundles_to_run.sort(
             key=lambda b: (b[0].metadata.request_queue is not None, b[0].metadata.request_queue),
@@ -332,37 +335,21 @@ class BundleManager(object):
 
         # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
-
-            def get_available_workers(user_id):
-                # Make a deepcopy of the workers list so the filtering and deducting don't modify the list
-                workers_list = copy.deepcopy(workers.user_owned_workers(user_id))
-                workers_list = self._deduct_worker_resources(workers_list, running_bundles_info)
-                workers_list = self._filter_and_sort_workers(workers_list, bundle, bundle_resources)
-                return workers_list
-
-            workers_list = None
-            if bundle.owner_id != self._model.root_user_id:
-                # First try private workers
-                workers_list = get_available_workers(bundle.owner_id)
-                # If there is no user_owned worker, try to schedule the current bundle to run on a CodaLab's public worker.
-                if len(workers_list) == 0:
-                    # Check if there is enough parallel run quota left for this user
-                    if (
-                        self._model.get_user_parallel_run_quota_left(
-                            bundle.owner_id, user_info_cache[bundle.owner_id]
-                        )
-                        <= 0
-                    ):
-                        logger.info(
-                            "User %s has no parallel run quota left, skipping job for now",
-                            bundle.owner_id,
-                        )
-                        # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
-                        continue
-            if not workers_list:
-                # Length is 0 (private user with no workers) or is None (root user)
-                workers_list = get_available_workers(self._model.root_user_id)
-
+            # Check if there is enough parallel run quota left for this user
+            if (
+                self._model.get_user_parallel_run_quota_left(
+                    bundle.owner_id, user_info_cache[bundle.owner_id]
+                )
+                <= 0
+            ):
+                # Don't include CodaLab-owned workers, as there is no parallel_run_quota left for this user.
+                codalab_owned_workers = []
+            else:
+                codalab_owned_workers = workers.user_owned_workers(self._model.root_user_id)
+            user_owned_workers = workers.user_owned_workers(bundle.owner_id)
+            workers_list = user_owned_workers + codalab_owned_workers
+            workers_list = self._deduct_worker_resources(workers_list, running_bundles_info)
+            workers_list = self._filter_and_sort_workers(workers_list, bundle, bundle_resources)
             # Try starting bundles on the workers that have enough computing resources
             for worker in workers_list:
                 if self._try_start_bundle(workers, worker, bundle, bundle_resources):
@@ -370,8 +357,9 @@ class BundleManager(object):
 
     def _deduct_worker_resources(self, workers_list, running_bundles_info):
         """
-        From each worker, subtract resources used by running bundles. Modifies the list.
+        From each worker, subtract resources used by running bundles.
         """
+        workers_list = copy.deepcopy(workers_list)
         for worker in workers_list:
             for uuid in worker['run_uuids']:
                 # Verify if the current bundle exists in both the worker table and the bundle table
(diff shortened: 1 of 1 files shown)


percyliang:
Can we just launch CodaLab workers with various tags and not change any of the scheduling logic?  That way, we can evolve the set of tags easily...e.g., to distinguish between cpus and gpus, etc.

candicegjing:
> Can we just launch CodaLab workers with various tags and not change any of the scheduling logic? That way, we can evolve the set of tags easily...e.g., to distinguish between cpus and gpus, etc.

Yes. I also thought about tagging our public instances. The thing is that in dispatching logic, there is some hidden logic that always runs private workers first no matter what. If we tag the public workers with a special tag, with the current scheduling logic, it still won't lead us to the correct result.

codalab/server/bundle_manager.py
```diff
@@ -310,9 +310,12 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
-        Schedules STAGED bundles to run on the given workers. Always tries to schedule bundles to
-        run on workers that are owned by the owner of each bundle first. If there are no such qualified
-        private workers, uses CodaLab-owned workers, which have user ID root_user_id.
+        Schedule STAGED bundles to run on available workers based on the following logic:
+        1. If the bundle requests to run on a specific worker, tries to schedule the bundle
+           to run on a worker that has a tag exactly match with bundles' request_queue.
+        2. If the bundle doesn't request to run on a specific worker,
+          (1) try to schedule the bundle to run on a private worker that is owned by the bundle's owner
+          (2) if there is no such qualified private worker, uses CodaLab-owned workers, which have user ID root_user_id.
         :param workers: a WorkerInfoAccessor object containing worker related information e.g. running uuid.
```
candicegjing: @percyliang and @bkgoksel Logic changes are reflected in the above comments from line 313 to line 318. No rerouting supported in the future dispatching logic anymore, meaning if there is no matched private worker can be found, the program won't try to reroute the bundle to run on any other available workers. The program will just do a **exact match** between **bundle.metadata.request_queue** and **worker's tag**

codalab/server/bundle_manager.py
```diff
@@ -333,36 +336,42 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
-                            bundle.owner_id,
-                        )
-                        # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
-                        continue
-            if not workers_list:
-                # Length is 0 (private user with no workers) or is None (root user)
-                workers_list = get_available_workers(self._model.root_user_id)
+                workers_list = filter_workers_by_request_queue(bundle.owner_id)
+
+            # If there is no user owned worker or the bundle is requested to run on CodaLab's public workers,
```
percyliang: Logic shouldn't make reference to a notion of public or not.

codalab/server/bundle_manager.py
```diff
@@ -333,36 +336,42 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
+                    logger.info(
+                        "User %s has no parallel run quota left, skipping job for now",
+                        bundle.owner_id,
+                    )
+                    # Don't start this bundle yet, as there is no parallel_run_quota left for this user.
+                    continue
+                # Try to matched workers in CodaLab-owned instances
+                workers_list = filter_workers_by_request_queue(self._model.root_user_id)
 
+            workers_list = self._deduct_worker_resources(workers_list, running_bundles_info)
```
percyliang, line 351: If there are private workers returned but they don't match the resource requirements, it seems like we don't have a chance to add back the codalab-owned workers?  In general, I find this logic difficult to follow and it would be nice if we simply 1) got all the workers (public and private), 2) filtered out the ones which don't satisfy the resource constraints, and 3) prioritized all the workers.

candicegjing: 1. If there aren't any private workers match with `request-queue` of the current bundle, it will fall into the condition check at [line 356](https://github.com/codalab/codalab-worksheets/pull/2031/files#diff-623d90f755b41b69c721e18042ffa858R356) and try to match those codalab-owned workers. Or I guess your concern is more on the case when a bundle didn't request for `request-queue`, but there are available workers for this user. However, if none of those private workers meet the resource requirements, we should reroute the job to a codalab-owned worker? 
2. As a side note, this rerouting logic in the current code was introduced very recently during [the hotfix/refactoring](https://github.com/codalab/codalab-worksheets/commit/ee5effc1e04815501f9bb41aeef615a92b9f3134#diff-623d90f755b41b69c721e18042ffa858) on `WorkerInfoAccessor`. If we look back a little bit earlier at [the scheduling logic](https://github.com/codalab/codalab-worksheets/blob/30e5da9da0401e8bb819adef33251845103e3160/codalab/server/bundle_manager.py#L332-L350), we didn't try to reroute the bundle to run on a public worker if there is no private worker could meet resource constraints. Let me know if we want to add it back.
3. I agree that the logic here is quite hard to follow. In the beginning, I also thought about filtering all the workers (public and private) first and do the other logic. But there is a parallel quota check when a bundle runs in public workers, which may require a way (e.g. special condition check) to know if the worker returns from the first filter is a public one. 

I guess I was trying to avoid this special condition check when I was changing this piece of logic. But if we think this sounds okay, I am happy to rewrite to this way.

percyliang: You can still do the parallel quota check after getting all the relevant workers first (since for each worker, you know whether it's public or not), which would happen in step 2).  You might lose a bit of performance by doing the 1) 2) 3) thing I suggested, but I think it's worth making the code easier to read so it's easy to reason about first, and then we can consider optimizations.  @bkgoksel can weigh in here too.

candicegjing: Sure. Sounds good. Will work on this~

codalab/server/bundle_manager.py
```diff
@@ -332,37 +335,32 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
+                    )
+                    <= 0
+                ):
+                    logger.info(
+                        "User %s has no parallel run quota left, skipping job for now",
+                        bundle.owner_id,
+                    )
+                    # Don't include CodaLab-owned workers, as there is no parallel_run_quota left for this user.
+                    codalab_owned_workers = []
+                workers_list = user_owned_workers + codalab_owned_workers
```
percyliang: Move this to right after 342.

codalab/server/bundle_manager.py
```diff
@@ -332,37 +335,32 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
 
         # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
+            codalab_owned_workers = copy.deepcopy(
+                workers.user_owned_workers(self._model.root_user_id)
+            )
+            user_owned_workers = copy.deepcopy(workers.user_owned_workers(bundle.owner_id))
+            if bundle.metadata.request_queue:
```
percyliang: Can't this logic go in the filter workers?  That would be cleaner, so all this function does is get the set of candidate workers?

codalab/server/bundle_manager.py
```diff
@@ -332,37 +335,32 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
 
         # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
+            codalab_owned_workers = copy.deepcopy(
```
percyliang: I think it'd be less error prone to not have these deepcopy, but make `deduct_worker_resources` do the deepcopy so that all functions behave immutably.

candicegjing: Yes, that's true. Updated.

codalab/server/bundle_manager.py
candicegjing: Moved back.

codalab/server/bundle_manager.py
```diff
@@ -332,37 +335,24 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
-                workers_list = get_available_workers(self._model.root_user_id)
-
+            # Check if there is enough parallel run quota left for this user
+            if (
+                self._model.get_user_parallel_run_quota_left(
+                    bundle.owner_id, user_info_cache[bundle.owner_id]
+                )
+                <= 0
+            ):
+                logger.info(
```
percyliang: This message doesn't make sense here because the job could still get run, delete it.

codalab/server/bundle_manager.py
```diff
@@ -372,6 +362,7 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
         """
         From each worker, subtract resources used by running bundles. Modifies the list.
```
percyliang: Update this comment - we don't modify the list anymore.

percyliang (approved):
This looks better now.  We probably want to test this logic fairly thoroughly on dev just to make sure we didn't screw something up.

codalab/server/bundle_manager.py
```diff
@@ -332,46 +335,31 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
             for worker in workers_list:
                 if self._try_start_bundle(workers, worker, bundle, bundle_resources):
                     break
 
     def _deduct_worker_resources(self, workers_list, running_bundles_info):
         """
-        From each worker, subtract resources used by running bundles. Modifies the list.
+        From each worker, subtract resources used by running bundles.
         """
+        workers_list = copy.deepcopy(workers_list)
```
epicfaace, line 362: Why do you no longer modify the list?

PR #2037, opened 2020-02-25
Remove uninformative NaNyNaNd in bundle view

### LINKED ISSUE #2036: Uninformative value NaNyNaNd in bundle view
When a bundle in a staged state, its `time` field is filled with value `NaNyNaNd` in the bundle view. It's uninformative and doesn't look good. It's better to fill the row with `0` or leave it empty?

candicegjing:
Fixed #2036 

This PR is to fix the displaying issue for the **time** row in the **bundle view** page.
Currently, when a bundle is in the staged state, it has the following bundle view page (**time=NaNyNaNd** ):
Screen Shot 2020-02-25 at 3 21 35 PM

After the fix, its bundle view page becomes:
Screen Shot 2020-02-25 at 3 22 00 PM

Not sure if this is the best way to fix it. If not, please let me know.

### DIFF
--- a/frontend/src/util/worksheet_utils.js
+++ b/frontend/src/util/worksheet_utils.js
@@ -7,7 +7,7 @@ export function renderDuration(s) {
     // s: number of seconds
     // Return a human-readable string.
     // Example: 100 => "1m40s", 10000 => "2h46m"
-    if (s === null) {
+    if (s == null) {
         return '<none>';
     }
 


frontend/src/util/worksheet_utils.js
```diff
@@ -7,6 +7,10 @@ export function renderDuration(s) {
     // s: number of seconds
     // Return a human-readable string.
     // Example: 100 => "1m40s", 10000 => "2h46m"
+    if (typeof s === 'undefined'){
```
percyliang: `s` should not be a string containing 'undefined'.  The fact that this is happening means that there's a bug elsewhere.

candicegjing: I double-checked the logic that generates this value. Here is the fact that I collected:
1. `time` field is not filled with a real value until the bundle moves to `RUNNING` stage. This means that the `time` field will not be assigned with real value when the bundle is in `STAGED` state.
2. When frontend code tries to display the `time` field from metadata, even though `metadata.time` variable is defined, but it still hasn't been assigned with real value which leads to `typeof s === "undefined"`.
3. When frontend tries to display the `time` field using [renderDuration function](https://github.com/codalab/codalab-worksheets/blob/master/frontend/src/util/worksheet_utils.js#L6-L30), since the value of `s` is undefined, variable `y` and `d` is `NaN`. At the end the function see [Line 30](https://github.com/codalab/codalab-worksheets/blob/master/frontend/src/util/worksheet_utils.js#L30), when it concatenates `Math.round(y) + 'y' + Math.round(d) + 'd';`, since both `y` and `d` are `NaN`, the final returning result becomes `NaNyNaNd`. 

Another option to fix this could be checking `isNaN(y)` and `isNaN(d)` before the last return statement. I think checking `s` at the beginning of this function might be better as the program doesn't need to go through the rest of the logic. Let me know if I missed anything.

frontend/src/util/worksheet_utils.js
```diff
@@ -7,6 +7,10 @@ export function renderDuration(s) {
     // s: number of seconds
     // Return a human-readable string.
     // Example: 100 => "1m40s", 10000 => "2h46m"
+    if (typeof s === 'undefined') {
+        return '';
+    }
+
     if (s === null) {
```
percyliang: These two cases shold be merged - `if (s == null)` should work.  Make the return `<none>`.

frontend/src/util/worksheet_utils.js
```diff
@@ -20,14 +16,17 @@ export function renderDuration(s) {
 
     s -= m * 60;
     var h = Math.floor(m / 60);
+
```
percyliang: Why add extra newlines?

percyliang (approved):

frontend/src/util/worksheet_utils.js
candicegjing: Removed those, they were added when I was debugging the logic.

PR #2040, opened 2020-02-27
Add queue in new run field

### LINKED ISSUE #2034: [FEATURE REQUEST] Allow to specify request_queue in "Edit and Rerun"
I sometimes need to re-run a job on a different queue when the original queue is temporarily unavailable. I couldn't do it on the web interface.

Allow us to specify the `request_queue` field in the web interface.

**Describe alternatives you've considered**
Use the web terminal with `mimic --request-queue` is also fine.  But I found that `u` is not working now, which in the past  is a shortcut to paste the selected bundle UUID to the terminal.

**Additional context**
We are using a internal CodaLab cluster hosted on AWS Batch.

yipenghe:
fix #2034 
Seems like would be handy for users, open to discuss where to place the queue field in the new run
Screenshot from 2020-02-26 21-19-41


percyliang:
All of our use cases are `--request-queue tag=foo`, so maybe best to show `tag=...` and just have people type in `foo`.  (Another option is to always assume `tag=X` whenever given `X`.)

In either case, we should add an info button.

I'd put the Queue at the very bottom.

yipenghe:
> All of our use cases are `--request-queue tag=foo`, so maybe best to show `tag=...` and just have people type in `foo`. (Another option is to always assume `tag=X` whenever given `X`.)
> 
> In either case, we should add an info button.
> 
> I'd put the Queue at the very bottom.

Updated the argument and tooltip

yipenghe:
Merge only after #2062

frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -434,6 +442,18 @@ class NewRun extends React.Component<{
                                     placeholder={`${kDefaultDockerCpu}`}
                                 />
                             </Grid>
+                            <ConfigLabel
+                                label="Queue"
+                                tooltip="Tag of the queue, this will add '--request-queue tag={input}' to the run"
```
epicfaace:
```suggestion
                                tooltip="Tag of the queue, this will add '--request-queue {input}' to the run"

```

frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -241,6 +243,13 @@ class NewRun extends React.Component<{
         if (cpu) args.push(`--request-cpus=${cpu}`);
         if (gpu) args.push(`--request-gpus=${gpu}`);
         if (docker) args.push(`--request-docker-image=${docker}`);
+        if (queue) {
+            let finalQueue = queue;
+            if (finalQueue.startsWith("tag=")) {
```
epicfaace: Why remove `tag=`? Given that we support both `--request-queue ...` and `--request-queue tag=...`

epicfaace (changes requested):

frontend/src/components/worksheets/NewRun/NewRun.jsx
yipenghe: Percy mentioned to strip it asap

epicfaace: Why? It seems simpler just to say "what you type in is verbatim passed as the argument to `--request-queue`" @percyliang

percyliang: I agree it's good to be consistent in both places, but I think we should deprecate `tag=...` though.  The rationale for stripping `tag=` as early as possible was so most of our code can just reason about the non-stripped version, but open to other suggestions.

yipenghe: I would go more for whatever is passed in is the tag if we support both. Unless we start rejecting  `tag=`, users who are used to it will continue to use that in command line as well and we won't be able to deprecate that

frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -241,6 +243,7 @@ class NewRun extends React.Component<{
         if (cpu) args.push(`--request-cpus=${cpu}`);
         if (gpu) args.push(`--request-gpus=${gpu}`);
         if (docker) args.push(`--request-docker-image=${docker}`);
+        if (queue) args.push(`--request-queue=${queue}`);
```
yipenghe, line 246: As discussed, since backend removes tag= anyway, we don't strip the tag= beforehand on frontend

percyliang (approved):

PR #2042, opened 2020-02-27
Bump marshmallow to 2.15.1

epicfaace:
Upgrading marshmallow was breaking our build due to [a change from 2.10.2 to 2.10.3](https://github.com/marshmallow-code/marshmallow/commit/d81cab413e231ec40123020f110a8c0af22163ed) that modified how integer fields with `as_string=True` are serialized.

This change caused an error when unpack_object was called, because in the following line, `obj_data` would no longer have the `id` attribute:

https://github.com/codalab/codalab-worksheets/blob/19f6cfa56e9ad1c2d4d2bd49eeba4ea7b89dba1a/codalab/client/json_api_client.py#L284

i.e., when printing out obj_data with Marshmallow 2.10.2, it would be equal to `{'type': 'worksheet-items', 'attributes': {'value': '', 'sort_key': None, 'type': 'bundle'}, 'relationships': {'worksheet': {'data': {'type': 'worksheets', 'id': '0xcd9def70173840beb122e0833e49356d'}}, 'subworksheet': {'data': None}, 'bundle': {'data': {'type': 'bundles', 'id': '0xb7f4ffe5e23945cca5ad02c87a076557'}}}, 'id': '58'}`

but in 2.10.3, it would be equal to `{'type': 'worksheet-items', 'attributes': {'type': 'bundle'}, 'relationships': {'worksheet': {'data': {'type': 'worksheets', 'id': '0xb091f526e109448798a54b5e098ed7c7'}}, 'bundle': {'data': {'type': 'bundles', 'id': '0x36a547c90f0742a7a6946dff07454ab4'}}}}`

Thus, this fix overrides `fields.Integer` to have the 2.10.2 behavior when serializing it. I'm not exactly sure why the 2.10.3 change caused the field to be dropped, though.


epicfaace:
TODO: file an issue with marshmallow about this

codalab/rest/schemas.py
```diff
@@ -16,6 +17,14 @@
 from codalab.objects.permission import parse_permission, permission_str
 
 
+class CompatibleInteger(fields.Integer):
+    def serialize(self, attr, obj, accessor=None):
+        """Overrides change done in https://github.com/marshmallow-code/marshmallow/compare/2.10.2...2.10.3#diff-c4881f63f8b05b8d3eadf046c3dc7912L641.
```
epicfaace:
```suggestion
        """Overrides change done from 2.10.2->2.10.3 in https://github.com/marshmallow-code/marshmallow/commit/d81cab413e231ec40123020f110a8c0af22163ed.
```

candicegjing (approved):
Tested locally, it works for me! 👍

Issue #1050, opened 2019-03-11
be able to upload directories


w0nche0l:
_From @codalab on December 27, 2015 10:10_

<a href="https://github.com/percyliang"><img src="https://avatars.githubusercontent.com/u/435384?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [percyliang](https://github.com/percyliang)**
_Tuesday Nov 17, 2015 at 15:26 GMT_
_Originally opened as https://github.com/codalab/codalab-competitions/issues/1406_

---

Do what DropBox or Google Drive do.

_Copied from original issue: codalab/codalab-worksheets#25_

w0nche0l:
_From @bkgoksel on August 3, 2018 23:13_

Isn't this supported now?

w0nche0l:
_From @percyliang on August 4, 2018 3:22_

From the CLI, but not the web interface.

Levilian:
One issue I faced is that when uploading folders the file structure will not be preserved but stored as a metadata attached to each individual file. I tried to search online how to rebuild the file directory from individual files but can't find any relevant resources.

PR #2043, opened 2020-02-27
Upload directory and multiple files, show upload progress bar, and keyboard-control upload menu

Levilian:
Implement #1050


percyliang:
Seems like we have a lot of files added unintentionally.  Probably best to recreate the branch to not have these in our history.

yipenghe:
Lol there still several unwanted files (#codalab_service.py# ...etc.

yipenghe:
There seems to be some very big changes, we should test it thoroughly on dev before merging to master

percyliang:
Are we sure all the original extraneous files are actually purged from the history?  I'd feel safer having a new branch...

Levilian:
> Are we sure all the original extraneous files are actually purged from the history? I'd feel safer having a new branch...

Yup! I rebased the branch and squashed the commits

codalab/lib/upload_manager.py
```diff
@@ -45,77 +46,100 @@ def upload_to_bundle_store(
         |simplify_archives|: whether to simplify unpacked archives so that if they
                              contain a single file, the final path is just that file,
                              not a directory containing that file.
-
+        |is_dir_upload|: whether the incoming sources are individual files that need to be rebuild
+        into a directory or merely individual files.
```
percyliang: Say more about what `sources` are in this case.

codalab/lib/upload_manager.py
```diff
@@ -45,77 +46,100 @@ def upload_to_bundle_store(
...
-                is_url, is_local_path, is_fileobj, filename = self._interpret_source(source)
-                source_output_path = os.path.join(bundle_path, filename)
-                if is_url:
-                    if git:
-                        source_output_path = file_util.strip_git_ext(source_output_path)
-                        file_util.git_clone(source, source_output_path)
-                    else:
-                        file_util.download_url(source, source_output_path)
-                        if unpack and self._can_unpack_file(source_output_path):
+        if is_dir_upload:
```
percyliang: I thought we went with the version where we zip up things on the client side, but we switched back to uploading individual files?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,46 +42,163 @@ class NewUpload extends React.Component<{
...
-        if (!file) {
+    uploadFile = (files) => {
+        if (!files) {
+            return;
+        }
+        let folderSize = 0;
+        for (const file of files) {
+            folderSize += file.size;
+        }
+        if (folderSize/1024/1024/1024 > 2) {
```
percyliang: Prefer multiplication on the right-hand side.

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,46 +42,163 @@ class NewUpload extends React.Component<{
...
+            if (after_sort_key) {
+                url += `&after_sort_key=${ after_sort_key }`;
+            }
+            $.ajax({
+                url,
+                data: JSON.stringify(createBundleData),
+                contentType: 'application/json',
+                type: 'POST',
+                success: (data, status, jqXHR) => {
+                    var bundleUuid = data.data[0].id;
```
percyliang: Use `let` or better yet `const`

frontend/src/components/worksheets/items/ActionButtons.js
```diff
@@ -44,18 +59,43 @@ class ActionButtons extends React.Component<{
...
+                            aria-label='Add New Upload'
+                            onClick={this.handleClick}
+                            disabled={!editPermission}
+                        >
                             <UploadIcon className={classes.buttonIcon} />
                             Upload
-                        </label>
-                    </Button>
+                        </Button>
+                        <Menu
```
yipenghe, line 85: It might be better to be a dropdown menu instead of covering upload

yipenghe (commented):
Using click works great!

Keyboard shortcut  (:0)doesn't seem to work, "a u" will trigger the menu to show, but "up" "down" keyboards will move both the focus on worksheet and the upload menu. Hitting "enter" also doesn't open the upload menu, but will trigger bundle detail / text editing in the worksheet area

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,46 +42,163 @@ class NewUpload extends React.Component<{
...
-            contentType: 'application/json',
-            type: 'POST',
-            success: (data, status, jqXHR) => {
-                var bundleUuid = data.data[0].id;
-                var reader = new FileReader();
-                reader.onload = () => {
-                    var arrayBuffer = reader.result,
-                        bytesArray = new Uint8Array(arrayBuffer);
+
+        var zip = new JSZip();
```
yipenghe: If we zipped here, why do we need to the is_dir_upload operations?

yipenghe:
I also like the percentage progress!

frontend/package.json
```diff
@@ -21,16 +21,19 @@
     "jquery-ui-bundle": "^1.12.1-migrate",
     "jquery.terminal": "^2.1.2",
     "jsonapi-datastore": "^0.4.0-beta",
+    "jszip": "^3.2.2",
     "lint-staged": "^8.2.1",
     "marked": "^0.6.2",
     "material-ui-chip-input": "^1.0.0-beta.16",
     "query-string": "^6.2.0",
     "react": "^16.8.6",
+    "react-circular-progressbar": "^2.0.3",
```
epicfaace, line 30: Can you update the title / description of this PR to make clear that this PR also adds the progress bar?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +42,171 @@ class NewUpload extends React.Component<{
...
-        if (!file) {
+    uploadFile = (files) => {
+        if (!files) {
+            return;
+        }
+        let folderSize = 0;
+        for (const file of files) {
+            folderSize += file.size;
+        }
+        if (folderSize > 2*1024*1024*1024) {
```
epicfaace: Please make this a constant -- it could be exported from codalab-worksheets/frontend/src/constants.js.

Similarly with the "2 GB" alert.

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +42,171 @@ class NewUpload extends React.Component<{
...
         if (!files.length) {
             return;
         }
-        this.uploadFile(files[0]);
+        this.uploadFolder(files);
     }
 
-    uploadFile = (file) => {
-        if (!file) {
+    uploadFile = (files) => {
```
epicfaace: Can you rename this method `uploadFiles`?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +42,171 @@ class NewUpload extends React.Component<{
...
+        if (!files) {
+            return;
+        }
+        let folderSize = 0;
```
epicfaace: You can instead do (looks nicer, also use const instead of let):

```
const folderSize = files.reduce((a, b) => a.size + b.size, 0);
```

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +42,171 @@ class NewUpload extends React.Component<{
...
+            alert('File size is large than 2 GB. Please upload your file(s) through CLI.');
             return;
         }
         const { worksheetUUID, after_sort_key } = this.props;
         const { name, description } = this.state;
-        const createBundleData = getDefaultBundleMetadata(name || file.name, description);
+
+        for (const file of files) {
+            const createBundleData = getDefaultBundleMetadata(name || file.name, description);
+            this.setState({
```
epicfaace: Right now, you're setting `uploading: true` before each file is uploaded and calling the success/failure callbacks (and `reloadWorksheet`) after each file is uploaded. Can you make it so that the former happens before the for loop (before _all_ files are uploaded) and the latter after _all_ files have been uploaded?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +42,171 @@ class NewUpload extends React.Component<{
...
-                    var url =
+
+        let zip = new JSZip();
+        [...files].map(file => {
+            zip.file(file.webkitRelativePath, file);
+        });
+
+        zip.generateAsync({type:"uint8array", compression: "DEFLATE"}).then((bytesArray) => {
+
+            $.ajax({
```
epicfaace, line 197: Can you unify this $.ajax call and the $.ajax call in `uploadFile` into a single function called by both functions?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -114,47 +232,62 @@ class NewUpload extends React.Component<{
...
+                />
+                { uploading && <CircularProgressbar
                         className={ classes.progress }
                         variant="determinate"
                         value={ percentComplete }
-                        size={ 80 }
+                        text={`${percentComplete}% uploaded`}
+                        styles={buildStyles({
+                            textSize: '12px',
+                            // pathColor: 'teal',
```
epicfaace: remove comments

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -164,11 +297,12 @@ class NewUpload extends React.Component<{
 
 const styles = (theme) => ({
     progress: {
+        "z-index": 1000,
```
epicfaace: Not sure if this would work -- you may have to do `zIndex: 1000`

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -470,6 +471,7 @@ class Worksheet extends React.Component {
             focusedBundleUuidList: focusedBundleUuidList,
             showNewRun: false,
             showNewText: false,
+            // uploadAnchor: null,
```
epicfaace: Why comment this out?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -730,7 +731,7 @@ class Worksheet extends React.Component {
                         if (this.state.focusIndex < 0) {
                             $('html, body').animate({ scrollTop: $(document).height() }, 'fast');
                         }
-                        document.querySelector('label[for=codalab-file-upload-input]').click();
+                        document.querySelector('Button[id=upload-button]').click();
```
```suggestion
                        document.querySelector('#upload-button').click();
```

frontend/src/components/worksheets/items/ActionButtons.js
```diff
@@ -44,18 +70,53 @@ class ActionButtons extends React.Component<{
...
-                        disabled={!editPermission}
-                    >
-                        <label className={classes.uploadLabel} for='codalab-file-upload-input'>
+                    <span>
+                        <Button
+                            size='small'
+                            color='inherit'
+                            id='upload-button'
+                            aria-label='Add New Upload'
+                            onClick={onShowNewUpload}
```
epicfaace: By adding the button to trigger uploads instead of the label, I think there's a possibility that you're undoing the fixes done in #1945. Can you make sure that the following things still work:
- uploads still work on chrome
- when I open the upload file dialog and exit it, then click "upload file" again, the dialog still shows up
- when I upload a file X, then click "upload file" again and select the same file X, the same file gets uploaded again

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1251,9 +1252,24 @@ class Worksheet extends React.Component {
         this.setState({ BulkBundleDialog: deleteWorksheetDialog });
     }
 
+    onShowNewUpload = (e) => {
```
epicfaace: Can you rename this to something like `showUploadMenu` (so that it doesn't sound too much like `onShowNewRun` or `onShowNewText`, which do very different things)? (and before #1945, we used to have a variable called `showNewUpload` as well)

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1251,9 +1252,24 @@ class Worksheet extends React.Component {
...
+        e.stopPropagation();
```
epicfaace: Why is this needed?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1251,9 +1252,24 @@ class Worksheet extends React.Component {
...
+        this.setState({ uploadAnchor: e.currentTarget });
+        let form = document.querySelector('Menu[id=upload-menu]');
```
epicfaace: You should do `#upload-menu`

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1251,9 +1252,24 @@ class Worksheet extends React.Component {
...
+        Mousetrap.unbind(['up', 'k']);
+        Mousetrap.unbind(['down', 'j']);
+        Mousetrap.unbind(['enter']);
+        Mousetrap.reset();
+        Mousetrap(form).bind(['enter'], function(e) {
+            e.preventDefault();
+            // console.log(e.target.firstElementChild.htmlFor);
```
epicfaace: remove comment

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1251,9 +1252,24 @@ class Worksheet extends React.Component {
...
+            document.querySelector('label[for=' + e.target.firstElementChild.htmlFor + ']').click();
+        });
```
epicfaace, line 1263: When the user opens the menu, can they press "esc" to exit out of it?

epicfaace, line 1263: When the user opens the menu and then exits out of it, does the normal functionality for "enter" / "up" / "down" / etc. still work?

epicfaace (changes requested):
Great work so far!

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
Levilian: https://developer.mozilla.org/en-US/docs/Web/API/FileList
`FileList API` is actually an object, not an array.

frontend/src/components/worksheets/Worksheet/Worksheet.js
Levilian: yes

Levilian: yes

frontend/src/util/ws_mousetrap_fork.js
```diff
@@ -1066,3 +1066,35 @@
...
     }
 })(typeof window !== 'undefined' ? window : null, typeof window !== 'undefined' ? document : null);
+
+/**
+ * adds a pause and unpause method to Mousetrap
+ * this allows you to enable or disable keyboard shortcuts
+ * without having to reset Mousetrap and rebind everything
+ */
+/* global Mousetrap:true */
+(function(Mousetrap) {
```
epicfaace, line 1076: Did you write this code yourself or was it automatically added somehow?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -114,47 +242,60 @@ class NewUpload extends React.Component<{
...
         this.setState({ percentComplete: 0, uploading: false });
     }
-    
 
     render() {
         const { classes } = this.props;
         const { percentComplete, uploading } = this.state;
 
         return (
+
```
epicfaace: Remove empty line

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +43,180 @@ class NewUpload extends React.Component<{
...
-        this.uploadFile(files[0]);
+        this.uploadFolder(files);
+    }
+
+    uploadFiles = (files) => {
+        if (!files) {
+            return;
+        }
+
+        if (!files) {
```
epicfaace: Remove this if statement

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -41,54 +43,180 @@ class NewUpload extends React.Component<{
...
-            success: (data, status, jqXHR) => {
-                var bundleUuid = data.data[0].id;
-                var reader = new FileReader();
-                reader.onload = () => {
-                    var arrayBuffer = reader.result,
-                        bytesArray = new Uint8Array(arrayBuffer);
-                    var url =
+
+        let zip = new JSZip();
+        [...files].map(file => {
```
epicfaace, line 191: Can you just use files instead of ...files?

epicfaace (changes requested):

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
Levilian: File API treats files as an object, so we need to flatten it.

frontend/src/util/ws_mousetrap_fork.js
Levilian: I added it so we could perform keyboard shortcuts albeit having the upload menu.

epicfaace: @Levilian it would be good if you moved this code to a separate file. Everything else in this file is vendor code from mousetrap, right?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
epicfaace: @Levilian -- this is important because they basically call the same endpoint. It's better to commit non-duplicative code, if possible, than leave things like this for future coders to refactor.

PR #2045, opened 2020-02-27
Fix bundle manager bug for bad queue argument

bkgoksel:
Small reference bug. If there's a run where `request_queue` is set but
the value is not `tag=<something>` the current code hits a buggy path.

This is a hotfix, but I think the whole `request_queue` semantics need
an update. As things as are `request_queue` can ONLY be used with
`--request_queue tag=<something>` and causes bugs if specified in any
other shape.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -694,7 +694,8 @@ class BundleManager(object):
         if tagm != None:
             worker_tag = tagm.group(1)
             matched_workers = [worker for worker in workers if worker['tag'] == worker_tag]
-        return matched_workers or []
+            return matched_workers
+        return []
 
     def _get_staged_bundles_to_run(self, workers, user_info_cache):
         """
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -690,11 +690,9 @@ class BundleManager(object):
         :param workers: a list of workers
         :return: a list of matched workers
         """
-        tagm = re.match('tag=(.+)', request_queue)
+        tag_match = re.match('tag=(.+)', request_queue)
         if tagm != None:
-            worker_tag = tagm.group(1)
-            matched_workers = [worker for worker in workers if worker['tag'] == worker_tag]
-            return matched_workers
+            return [worker for worker in workers if worker['tag'] == tag_match.group(1)]
         return []
 
     def _get_staged_bundles_to_run(self, workers, user_info_cache):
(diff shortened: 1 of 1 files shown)


candicegjing (commented):
Could you give an example on in which case this will break the logic?

codalab/server/bundle_manager.py
```diff
@@ -690,11 +690,10 @@ def _get_matched_workers(request_queue, workers):
         :param workers: a list of workers
         :return: a list of matched workers
         """
-        tagm = re.match('tag=(.+)', request_queue)
+        tag_match = re.match('tag=(.+)', request_queue)
         if tagm != None:
```
candicegjing: If we change the `tagm` to `tag_match`, should this be changed as well?

bkgoksel: oops, yes I

bkgoksel:
> Could you give an example on in which case this will break the logic?

`cl run date --request queue foo``

bkgoksel:
The problem with the old code was, if `tagm` is `None`, then `matched_workers` is not defined, but the return statement still refers to it. This function only gets called if `request_queue` is not `None`, so we only hit that situation if `request_queue` is not `None` but also is not of the form `tag=<something>`

candicegjing:
> The problem with the old code was, if `tagm` is `None`, then `matched_workers` is not defined, but the return statement still refers to it. This function only gets called if `request_queue` is not `None`, so we only hit that situation if `request_queue` is not `None` but also is not of the form `tag=<something>`

Yes. That's true! Thanks for catching this!

percyliang:
`request-queue` used to take non-tag arguments in an older worker, but we don't use that anymore and I agree it's confusing.  So I think we should accept `--request-queue foo` and `--request-queue tag=foo`, and internally just canonicalize to `foo` as soon as possible.

bkgoksel:
> `request-queue` used to take non-tag arguments in an older worker, but we don't use that anymore and I agree it's confusing. So I think we should accept `--request-queue foo` and `--request-queue tag=foo`, and internally just canonicalize to `foo` as soon as possible.

Is it OK if we merge this for now so this issue will be fixed in the next release and we can make a ticket for making `tag` redundant?

candicegjing (approved):

percyliang (commented):
Sounds good.

percyliang (approved):

yipenghe:
does this affect #2040 ?

PR #2048, opened 2020-02-28
Job assignment

### LINKED ISSUE #1953: Job assignment sometimes doesn't work well

candicegjing:
Fixed #1953 
Suppose we have the following `workers_list`:
```
workers_list = [{'worker_id': 6, 'gpus': 1, 'dep': 8, 'cpus':6, 'runs':[1]},
                {'worker_id': 5, 'gpus': 2, 'dep': 0, 'cpus':6, 'runs':[]},
                {'worker_id': 4, 'gpus': 1, 'dep': 8, 'cpus':6, 'runs':[1,2,3]},
                {'worker_id': 3, 'gpus': 1, 'dep': 2, 'cpus':6, 'runs':[1]},
                {'worker_id': 2, 'gpus': 1, 'dep': 4, 'cpus':6, 'runs':[1,2,3,4]},
                {'worker_id': 1, 'gpus': 0, 'dep': 5, 'cpus':8, 'runs':[1,2,3,4,5,6]},
                {'worker_id': 0, 'gpus': 0, 'dep': 7, 'cpus':8, 'runs':[]}]
```
When `bundle.metadata.request_gpus = 1`:
```
{'worker_id': 6, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1]}
{'worker_id': 4, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1, 2, 3]}
{'worker_id': 2, 'gpus': 1, 'dep': 4, 'cpus': 6, 'runs': [1, 2, 3, 4]}
{'worker_id': 3, 'gpus': 1, 'dep': 2, 'cpus': 6, 'runs': [1]}
{'worker_id': 5, 'gpus': 2, 'dep': 0, 'cpus': 6, 'runs': []}
```

When `bundle.metadata.request_gpus = 0`:
```
{'worker_id': 0, 'gpus': 0, 'dep': 7, 'cpus': 8, 'runs': []}
{'worker_id': 1, 'gpus': 0, 'dep': 5, 'cpus': 8, 'runs': [1, 2, 3, 4, 5, 6]}
{'worker_id': 6, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1]}
{'worker_id': 4, 'gpus': 1, 'dep': 8, 'cpus': 6, 'runs': [1, 2, 3]}
{'worker_id': 2, 'gpus': 1, 'dep': 4, 'cpus': 6, 'runs': [1, 2, 3, 4]}
{'worker_id': 3, 'gpus': 1, 'dep': 2, 'cpus': 6, 'runs': [1]}
{'worker_id': 5, 'gpus': 2, 'dep': 0, 'cpus': 6, 'runs': []}
```


codalab/server/bundle_manager.py
```diff
@@ -445,11 +439,19 @@ def get_sort_key(worker):
             else:
                 deps = set(worker['dependencies'])
                 num_available_deps = len(needed_deps & deps)
-            worker_id = worker['worker_id']
 
-            # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
-            gpu_priority = bundle_resources.gpus or not has_gpu[worker_id]
-            return (gpu_priority, num_available_deps, worker['cpus'], random.random())
+            # GPU priority is useful when sorting bundles that request for only CPUs. In this case, we want
+            # to rank CPU workers as high as possible and GPU works as low as possible, so that GPU workers
```
teetone: "GPU works" should be "GPU workers".

codalab/server/bundle_manager.py
```diff
@@ -445,11 +439,19 @@ def get_sort_key(worker):
             else:
                 deps = set(worker['dependencies'])
                 num_available_deps = len(needed_deps & deps)
-            worker_id = worker['worker_id']
 
-            # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
-            gpu_priority = bundle_resources.gpus or not has_gpu[worker_id]
```
teetone, line 451: I guess the previous implementation just checked if GPU's have been requested. With your new fix, we want bundles to use CPU workers as much possible, unless GPU's are requested. Am I understanding this correctly?

codalab/server/bundle_manager.py
```diff
@@ -445,11 +439,19 @@ def get_sort_key(worker):
...
-            return (gpu_priority, num_available_deps, worker['cpus'], random.random())
+            # GPU priority is useful when sorting bundles that request for only CPUs. In this case, we want
+            # to rank CPU workers as high as possible and GPU works as low as possible, so that GPU workers
+            # could be better utilized towards GPU oriented jobs.
+            gpu_priority = worker['gpus'] == 0
+
+            # If the bundle requested for GPUs, prioritize only those workers that have GPUs.
+            # If the bundle didn't request for GPUs (a CPU only job), prioritize all workers based criteria as follows:
+            # 1. prioritize workers that don't have GPUs first
+            # 2. number of available shared dependencies between bundle's and workers'.
```
teetone: Should this be "number of available shared dependencies between a bundle and the worker"?

codalab/server/bundle_manager.py
candicegjing: Yes! Updated.

codalab/server/bundle_manager.py
candicegjing: Yes, you are right! Basically there are two cases as follows:
1. bundle.request_gpus > 0
2. bundle.request_gpus == 0

In the first case, when a bundle requests for GPUs, since the logic already filters out CPU workers in [Line400-Line425](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L400-L425) before reaching to the sorting algorithm, adding logic to prioritize GPU workers in the highest priority takes no effect in the final result. 
`gpu_priority` would have a real impact on the final result in the second case when a bundle requests for CPUs only. In such a situation, the bundle can be scheduled to run on both GPU workers and CPU workers. However, since GPUs are considered as more precious computing resources than CPUs, we would want to utilize CPU workers as much as possible in the first place.

codalab/server/bundle_manager.py
candicegjing: Updated~

teetone (approved):

teetone:
LGTM!

codalab/server/bundle_manager.py
```diff
@@ -445,11 +439,19 @@ def get_sort_key(worker):
...
                 num_available_deps = len(needed_deps & deps)
-            worker_id = worker['worker_id']
 
-            # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
-            gpu_priority = bundle_resources.gpus or not has_gpu[worker_id]
-            return (gpu_priority, num_available_deps, worker['cpus'], random.random())
+            # GPU priority is useful when sorting bundles that request for only CPUs. In this case, we want
+            # to rank CPU workers as high as possible and GPU workers as low as possible, so that GPU workers
+            # could be better utilized towards GPU oriented jobs.
+            gpu_priority = worker['gpus'] == 0
```
percyliang: Generalize this to `-worker['gpus']` (want to favor workers with fewer GPUs).
Also prioritize workers with fewer CPUs by the same general logic (choose the worker with the fewest resources that still satisfies the resource requirements)?

codalab/server/bundle_manager.py
```diff
@@ -445,11 +439,19 @@ def get_sort_key(worker):
...
+
+            # If the bundle requested for GPUs, prioritize only those workers that have GPUs.
+            # If the bundle didn't request for GPUs (a CPU only job), prioritize all workers based criteria as follows:
+            # 1. prioritize workers that don't have GPUs first
```
percyliang: Be consistent - each of these bullet points should be in the form 'Prioritize workers that ...'.

percyliang (changes requested):

codalab/server/bundle_manager.py
candicegjing: Yes. I have added `worker['cpus']` right after the shared dependencies, so that it will prioritize the number of shared dependencies first and then `worker['cpus']`. We could also move its position higher, depending on how important we think those dependencies are. Let me know if we want to change the order.

codalab/server/bundle_manager.py
```diff
@@ -445,13 +439,23 @@ def get_sort_key(worker):
             else:
                 deps = set(worker['dependencies'])
                 num_available_deps = len(needed_deps & deps)
-            worker_id = worker['worker_id']
 
-            # if the bundle doesn't request GPUs (only request CPUs), prioritize workers that don't have GPUs
-            gpu_priority = bundle_resources.gpus or not has_gpu[worker_id]
-            return (gpu_priority, num_available_deps, worker['cpus'], random.random())
+            # If the bundle requested for GPUs, prioritize those GPU workers that have fewer GPUs.
```
percyliang: This should be ```Subject to the worker meeting the resource requirements of the bundle, we also want to:
1. Prioritize workers with fewer GPUs (including zero).
2. Prioritize workers that aready have more bundle dependencies.
3. Prioritize...
```

candicegjing:
Closing this PR as the commit history has been messed up. Please refer to the new PR #2056 for the job assignment changes.

PR #2049, opened 2020-03-01
Improve error handling for CLI

### LINKED ISSUE #1958: TypeError: expected string or bytes-like object
```
Error on request by matt.downey18(0xd72e93b8f2b448fc9318d136e89e555a):

POST /cli/command

Query params:

Form params:

JSON body:
{
"command": "cl run evaluate.py: 0xdcf10e1da32e45ce970bbb07fd4694eb",
"worksheet_uuid": "0x1fc53883bf2f432eb3bc0ae4164ee0dd"
}

Local variables:
bundle_spec=None
worksheet_uuid='0x1fc53883bf2f432eb3bc0ae4164ee0dd'
client=<codalab.client.json_api_client.JsonApiClient object at 0x7f678015a6a0>

rv = callback(*a, **ka)
File "/opt/codalab-worksheets/codalab/rest/cli.py", line 48, in post_worksheets_command
return general_command(query['worksheet_uuid'], query['command'])
File "/opt/codalab-worksheets/codalab/rest/cli.py", line 149, in general_command
structured_result = cli.do_command(args)
File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 839, in do_command
File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 833, in <lambda>
File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 1527, in do_run_command
targets = self.resolve_key_targets(client, worksheet_uuid, args.target_spec)
File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 585, in resolve_key_targets
client, worksheet_uuid, target_spec, allow_remote=False
File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 563, in resolve_target
File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 597, in resolve_bundle_uuid
if spec_util.UUID_REGEX.match(bundle_spec):
TypeError: expected string or bytes-like object
```

candicegjing:
Fixed #1958.

This PR covers two bad cases when a user submits a bundle through CLI:
1. `cl run : "..."`
2. `cl run scripts: "..."`

### DIFF
--- a/codalab/lib/cli_util.py
+++ b/codalab/lib/cli_util.py
@@ -62,7 +62,8 @@ def nested_dict_get(obj, *args, **kwargs):
 
 def parse_key_target(spec):
     """
-    Parses a keyed target spec into its key and the rest of the target spec
+    Parses a keyed target spec into its key and the rest of the target spec.
+    Raise UsageError when the value of the spec is empty.
     :param spec: a target spec in the form of
         [[<key>]:][<instance>::][<worksheet_spec>//]<bundle_spec>[/<subpath>]
     where <bundle_spec> is required and the rest are optional.
@@ -71,11 +72,19 @@ def parse_key_target(spec):
         - <key>: (<key> if present,
                     empty string if ':' in spec but no <key>,
                     None otherwise)
-        - <value> (where value is everyhing after a <key>: (or everything if no key specified)
+        - <value> (where value is everything after a <key>: (or everything if no key specified)
     """
-
     match = re.match(TARGET_KEY_REGEX, spec)
-    return match.groups()
+    key, value = match.groups()
+    # This check covers three usage errors:
+    # 1. both key and value are empty, e.g. "cl run : 'echo a'"
+    # 2. key is not empty, value is empty, e.g. "cl run a.txt: 'echo a'"
+    if value == '':
+        raise UsageError(
+            'target_spec (%s) in wrong format. Please provide a valid target_spec in the format of %s.'
+            % (spec, RUN_TARGET_SPEC_FORMAT)
+        )
+    return (key, value)
 
 
 def parse_target_spec(spec):
@@ -132,13 +141,6 @@ def desugar_command(orig_target_spec, command):
                     'key %s exists with multiple values: %s and %s' % (key, key2val[key], val)
                 )
         else:
-            if key is None:
-                raise UsageError(
-                    'target_spec is empty. Please provide a valid target_spec in the format of {}.'.format(
-                        RUN_TARGET_SPEC_FORMAT
-                    )
-                )
-
             key2val[key] = val
             target_spec.append(key + ':' + val)
         return key
--- a/tests/lib/cli_util_test.py
+++ b/tests/lib/cli_util_test.py
@@ -59,6 +59,10 @@ class CLIUtilTest(unittest.TestCase):
         for spec, expected_parse in cases:
             self.assertEqual(cli_util.parse_key_target(spec), expected_parse)
 
+        usage_error_cases = [':', 'a:', '']
+        for spec in usage_error_cases:
+            self.assertRaises(UsageError, lambda: cli_util.parse_key_target(spec))
+
     def test_parse_target_spec(self):
         cases = [
             (
@@ -107,3 +111,6 @@ class CLIUtilTest(unittest.TestCase):
         )
         self.assertRaises(UsageError, lambda: cli_util.desugar_command([], 'echo %a:b% %a:c%'))
         self.assertRaises(UsageError, lambda: cli_util.desugar_command([':b'], 'echo %b:c%'))
+        self.assertRaises(UsageError, lambda: cli_util.desugar_command(['b:'], 'echo %b:c%'))
+        self.assertRaises(UsageError, lambda: cli_util.desugar_command([':'], 'echo %b:c%'))
+        self.assertRaises(UsageError, lambda: cli_util.desugar_command([''], 'echo %b:c%'))


epicfaace (commented):
Looks good. Can you add a test in `test_cli.py` for this that tests the error messages? I believe we should be able to add CLI tests like these without much difficulty.

yipenghe (approved):
Make sure to address the test failure before merging into master

codalab/lib/cli_util.py
```diff
@@ -116,6 +116,14 @@ def desugar_command(orig_target_spec, command):
 
     def get(dep):  # Return the key
         key, val = parse_key_target(dep)
+        # This check covers two usage errors:
```
percyliang: Seems like this check should happen in the `parse_key_target` function, because there are many places where we need this check.  The error checking around the call sites of `parse_key_target` can probably be cleaned up / unified.

candicegjing: Double-checked our codebase and moved the value check to `parse_key_target` function. As for cleaning up the error checking around the call site of `parse_key_target` function, there are three places that use this:
1. [resolve_key_targets](https://github.com/codalab/codalab-worksheets/blob/a7a62e5848083f2f9c948bcd7969d8a8fba535f8/codalab/lib/bundle_cli.py#L578): the following two functions 2 & 3 call `resolve_key_targets` function:
    1. [do_run_command](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1532): there seems to be no further clean up could be done.
    1. [do_make_command](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1451): [line 1453](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1453) comes from when target_spec is an empty string. Then this seems to be a valid check as well.

Not sure if I understand correctly. If I missed anything, please do let me know.

tests/lib/cli_util_test.py
```diff
@@ -107,3 +107,5 @@ def test_desugar(self):
         )
         self.assertRaises(UsageError, lambda: cli_util.desugar_command([], 'echo %a:b% %a:c%'))
         self.assertRaises(UsageError, lambda: cli_util.desugar_command([':b'], 'echo %b:c%'))
+        self.assertRaises(UsageError, lambda: cli_util.desugar_command(['b:'], 'echo %b:c%'))
```
percyliang, line 114: Should add a test specifically for `parse_key_target`

codalab/lib/cli_util.py
```diff
@@ -71,11 +72,19 @@ def parse_key_target(spec):
...
+        - <value> (where value is everything after a <key>: (or everything if no key specified)
     """
-
     match = re.match(TARGET_KEY_REGEX, spec)
-    return match.groups()
+    key, value = match.groups()
+    # This check covers three usage errors:
+    # 1. both key and value are empty, e.g. ":"
+    # 2. key is not empty, value is empty, e.g. "<key>:"
+    if key is not None and value == '':
```
percyliang: Isn't the right logic simply just that `value == ''` regardless of what `key` is?

candicegjing: Yes, I originally thought the same, but once I saw these [lines](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1451-L1453), I was thinking maybe I should add a `None` value check. It seems that's the special case it looks for. But let me know if I misunderstood anything.

percyliang: How are those lines related to this check?  The current code would accept `key = None and value == ''`, which presumably can't happen anyway, but `value == ''` is simpler.  I'd just make sure we have tests for all the different cases.

candicegjing: From those [lines](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1451-L1453), it looks like it accepts `None` value and will convert it to an empty string when `None` value passed in (e.g. when passing an empty string `""` to `parse_key_target` function). If this is something won't happen, maybe I should delete the check in bundle_cli.py to make things more clear.

percyliang: It's worth following how `targets` gets used downstream...it would be nice not to have a `''` key in addition to `None`, but we shouldn't break anything.  I think there are three (valid cases):
1) `key:value`
2) `:value`
3) `value` (specific to make bundles)
We need to distinguish between 2) and 3).

candicegjing: I am really not sure if we can get rid of `None` key and `''` key check here. The following three functions call `parse_key_target` directly:
1. in [resolve_key_targets](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L571):https://github.com/codalab/codalab-worksheets/blob/fa36a533c25d9a24dea51bc92775f580c1a81155/codalab/lib/bundle_cli.py#L579-L584):
2. in [do_make_command](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1449) : https://github.com/codalab/codalab-worksheets/blob/fa36a533c25d9a24dea51bc92775f580c1a81155/codalab/lib/bundle_cli.py#L1453
It checks for the existence of `None` key for each target returned from item 1 (essentially, there will be at most 1 target having `None` key based on logic in item 1 ^).
3. in [desugar_command](https://github.com/codalab/codalab-worksheets/blob/fa36a533c25d9a24dea51bc92775f580c1a81155/codalab/lib/cli_util.py#L119-L125): https://github.com/codalab/codalab-worksheets/blob/fa36a533c25d9a24dea51bc92775f580c1a81155/codalab/lib/cli_util.py#L119-L125

It looks like each of them is parsing `''` key or `None` key to fulfill their own purpose and they don't seem to share any logic that can be combined. If I misunderstand our scope here, please do let me know.

percyliang: Yeah, I agree we can't merge them.  It seems like `key == ''` corresponds to 2) and `key is None` coresponds to 3).  But the original change of `value == ''` seems safe.

candicegjing: I see. I have updated the condition check.

tests/lib/cli_util_test.py
```diff
@@ -59,6 +59,10 @@ def test_parse_key_target(self):
         for spec, expected_parse in cases:
             self.assertEqual(cli_util.parse_key_target(spec), expected_parse)
 
+        usage_error_cases = [':', 'a:']
```
epicfaace: This seems different enough from the test cases before -- can you make this into a different function? For example, call it `test_parse_key_target_invalid`

codalab/lib/cli_util.py
```diff
@@ -71,11 +72,19 @@ def parse_key_target(spec):
         - <key>: (<key> if present,
                     empty string if ':' in spec but no <key>,
                     None otherwise)
-        - <value> (where value is everyhing after a <key>: (or everything if no key specified)
+        - <value> (where value is everything after a <key>: (or everything if no key specified)
     """
-
     match = re.match(TARGET_KEY_REGEX, spec)
-    return match.groups()
+    key, value = match.groups()
```
epicfaace, line 78: Can you clarify here in a comment which part is `key` and which part is `value` of the entire target spec,

` [[<key>]:][<instance>::][<worksheet_spec>//]<bundle_spec>[/<subpath>]`?

codalab/lib/cli_util.py
```diff
@@ -71,11 +72,19 @@ def parse_key_target(spec):
...
+    # This check covers three usage errors:
```
epicfaace, line 79: I don't think we need this comment -- it seems pretty clear that this is checking when `value` is empty.

codalab/lib/cli_util.py
```diff
@@ -62,7 +62,8 @@ def nested_dict_get(obj, *args, **kwargs):
 
 def parse_key_target(spec):
     """
-    Parses a keyed target spec into its key and the rest of the target spec
+    Parses a keyed target spec into its key and the rest of the target spec.
+    Raise UsageError when the value of the target spec is empty.
     :param spec: a target spec in the form of
         [[<key>]:][<instance>::][<worksheet_spec>//]<bundle_spec>[/<subpath>]
```
epicfaace, line 68: Should this be `[[<key>]:]` (RUN_TARGET_SPEC_FORMAT) or `[[<key>:]]` (MAKE_TARGET_SPEC_FORMAT)? What's the difference?

codalab/lib/cli_util.py
```diff
@@ -132,13 +141,6 @@ def get(dep):  # Return the key
                     'key %s exists with multiple values: %s and %s' % (key, key2val[key], val)
                 )
         else:
-            if key is None:
```
epicfaace, line 135: Why remove this check?

tests/lib/cli_util_test.py
epicfaace: Should we add `''` in the array too?

tests/lib/cli_util_test.py
```diff
@@ -59,6 +59,10 @@ def test_parse_key_target(self):
...
+        for spec in usage_error_cases:
```
epicfaace, line 63: I would just inline this:

```suggestion
        for spec in [':', 'a:']:
```

tests/lib/cli_util_test.py
candicegjing: Yes, I originally also thought about this. But it seems that all the tests in this file try to group their test cases (the succeeded ones + failed ones) together in this way. I just want to keep them consistent.

candicegjing: See this link: https://github.com/codalab/codalab-worksheets/pull/2049#discussion_r391375266

epicfaace: I don't think that comment changes the fact that we still want to throw an error when an empty string is passed to `parse_key_target`?

epicfaace: No problem (I personally think that grouping succeeding + failed test cases together is a bad practice though :wink: )

candicegjing: Never mind, I was thinking of something else. We should add an empty string to this test suite.

codalab/lib/cli_util.py
candicegjing: Since this logic covers some edge cases that weren't put into consideration before, I think it would be helpful to have this documented. (Updated with more details.)

tests/lib/cli_util_test.py
candicegjing: I was mainly trying to make the new test cases being consistent with the structure of the rest code in this test suite.

PR #2055, opened 2020-03-03
Error markdowns not edittable/deletable

### LINKED ISSUE #1727: [New Frontend] Errors shown in Worksheets should not be Editable. Just deletable.
**Describe the bug**
When getting an error on a worksheet, the error message markdown is editable itself. When editing, it will spawn another error message. 

**To Reproduce**
1. Go to  a worksheet.
2. Edit the source and type `% hi` (invalid).
3. See the error line pop on the worksheet.
4. Edit that line and see another error pop up below it.

**Expected behavior**
Error messages should just be deletable and not editable. We can also add a button that will take the user to the source of error to amend it.

**Screenshots**
unknown_directive_error_should_not_be_editable

yipenghe:
fix #1727 

Also adjusts the error message a bit:
Indicate the error is in source and correct the line index (source starts from index 1)

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -703,7 +703,9 @@ def interpret_items(schemas, raw_items, db_model=None):
             for item_index, bundle_info in bundle_infos:
                 if is_missing(bundle_info):
                     blocks.append(
-                        MarkupBlockSchema().load({'text': 'ERROR: cannot access bundle'}).data
+                        MarkupBlockSchema()
+                        .load({'text': 'ERROR: cannot access bundle', 'error': True})
+                        .data
                     )
                     continue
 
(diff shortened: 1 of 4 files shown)


codalab/lib/worksheet_util.py
```diff
@@ -992,7 +994,9 @@ def flush_worksheets():
             worksheet_infos[:] = []
             blocks.append(
                 MarkupBlockSchema()
-                .load({'text': 'Error on line %d: %s' % (raw_index, str(e))})
```
teetone, line 995: You have to update `create_sample_worksheet.py` as well because we rely on regex to match against this error message.

teetone (changes requested):

teetone (approved):

PR #2057, opened 2020-03-04
Rearrange CLI doc layout

candicegjing:
The CLI doc will look like this: https://github.com/candicegjing/codalab-worksheets/blob/master/docs/CLI-Reference.md#uinfo after rearranging the layout. Let me know how we think about this. I guess with this change, we can have an anchor link for each CLI command which allows us to reference the content easier.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -140,6 +140,9 @@ SERVER_COMMANDS = (
 )
 
 OTHER_COMMANDS = ('help', 'status', 'alias', 'config', 'logout')
+# Markdown headings
+HEADING_LEVEL_2 = '## '
+HEADING_LEVEL_3 = '### '
 
 
 class CodaLabArgumentParser(argparse.ArgumentParser):
@@ -258,7 +261,7 @@ class Commands(object):
         return register_command
 
     @classmethod
-    def help_text(cls, verbose):
+    def help_text(cls, verbose, markdown):
         def command_name(command):
             name = command
             aliases = cls.commands[command].aliases
@@ -307,6 +310,8 @@ class Commands(object):
                 )
 
             if verbose:
+                if markdown:
+                    name = HEADING_LEVEL_3 + name
                 return '%s%s:\n%s\n%s' % (
                     ' ' * indent,
                     name,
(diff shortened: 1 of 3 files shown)


percyliang:
This file is auto-generated, so you'll have to find an automatic way of formatting it.

candicegjing:
> This file is auto-generated, so you'll have to find an automatic way of formatting it.

@percyliang I added a new option `-d` to avoid overlapping with original `-v` option. When user types in `cl help -v -d`, the CLI will auto-generate the CLI documentation in the Markdown format. Please let me know how you think about this.

epicfaace (commented):
@candicegjing looks good. Why did you decide to add an additional option, `-d`? Why not just make this new format the default that shows up when using the `-v` option?

candicegjing:
> @candicegjing looks good. Why did you decide to add an additional option, `-d`? Why not just make this new format the default that shows up when using the `-v` option?

I wonder if anyone directly use this in command line to print out help information, would this markdown look a bit of unusual for user? Let me know how you think!

scripts/gen-cli-docs.py
```diff
@@ -25,7 +24,7 @@ def main():
 
     with open(os.path.join(args.docs, 'CLI-Reference.md'), 'w') as f:
         print(INTRODUCTION, file=f)
-        print(indent(Commands.help_text(True)), file=f)
+        print(indent(Commands.help_text(True, True)), file=f)
```
epicfaace:
```suggestion
        print(indent(Commands.help_text(verbose=True, docs=True)), file=f)
```

epicfaace (approved):

codalab/lib/bundle_cli.py
```diff
@@ -258,7 +262,7 @@ def register_command(function):
         return register_command
 
     @classmethod
-    def help_text(cls, verbose):
+    def help_text(cls, verbose, docs):
```
percyliang: rename `docs` to `for_docs` and add a comment saying what this is for.

codalab/lib/bundle_cli.py
```diff
@@ -857,21 +867,28 @@ def print_version(self):
             'Show usage information for commands.',
             '  help           : Show brief description for all commands.',
             '  help -v        : Show full usage information for all commands.',
+            '  help -v -d     : Show full usage information for all commands in the Markdown format.',
```
percyliang: "in Markdown format"

codalab/lib/bundle_cli.py
percyliang: Actually call this `markdown` since it's more accurate

codalab/lib/bundle_cli.py
```diff
@@ -857,21 +867,28 @@ def print_version(self):
...
             '  help <command> : Show full usage information for <command>.',
         ],
         arguments=(
             Commands.Argument('command', help='name of command to look up', nargs='?'),
             Commands.Argument(
                 '-v', '--verbose', action='store_true', help='Display all options of all commands.'
             ),
+            Commands.Argument(
+                '-d',
+                '--docs',
```
percyliang: `--markdown`

percyliang (changes requested):
This will be nice.  Can you include a screenshot of what this looks like and what the URL for the anchors look like?

codalab/lib/bundle_cli.py
epicfaace: So short form would probably be `-m` then

candicegjing:
> This will be nice. Can you include a screenshot of what this looks like and what the URL for the anchors look like?

This is a screenshot of what the CLI reference page will look like after the changes: 

Screen Shot 2020-03-22 at 4 16 57 PM

This page shows the full picture of how exactly the CLI reference page will look like : [The new CLI-Reference](https://github.com/candicegjing/codalab-worksheets/blob/master/docs/CLI-Reference.md).

The anchor will differ by command, e.g. 
1. `make` will have a web anchor: https://github.com/candicegjing/codalab-worksheets/blob/master/docs/CLI-Reference.md#make. 
1. `upload (up)` will have a web anchor: https://github.com/candicegjing/codalab-worksheets/blob/master/docs/CLI-Reference.md#upload-up. 

Let me know if there is anything else that needs to be updated.

scripts/gen-cli-docs.py
```diff
@@ -2,20 +2,19 @@
 Generate CLI docs.
 """
 import os
-import sys
 import argparse
-from codalab.bin import cl  # Important to put here to register the commands.
-from codalab.lib.bundle_cli import Commands
+from codalab.lib.bundle_cli import Commands, HEADING_LEVEL_1
```
percyliang: I feel like we can just use `#` here to avoid extra dependencies and make it more transparent what's going on...  If we wanted a library, it should be a separate file that both bundle CLI and this file imports, but that is probably overkill right now.

candicegjing: That's true. I have removed this additional import and its related code.

percyliang (approved):

PR #2062, opened 2020-03-04
Add support to specify request-queue without tag= prefix

### LINKED ISSUE #2061: Supporting request-queue for both formats: tag=<X> and X
When a bundle is requested to run on a specific worker, we currently use `--request-queue tag=<worker_X>` to indicate the worker that the bundle should run on. This ticket is to add support for another format: `--request-queue worker_X`.

candicegjing:
Fixed #2061.

This PR adds support to specify request-queue without `tag=` prefix

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -689,12 +689,12 @@ class BundleManager(object):
     def _get_matched_workers(request_queue, workers):
         """
         Get all of the workers that match with the name of the requested worker
-        :param request_queue: a tag that can be used to match workers
+        :param request_queue: a tag with the format "tag=worker_X" or "worker_X" that can be used to match workers
         :param workers: a list of workers
         :return: a list of matched workers
         """
-        tag_match = re.match('tag=(.+)', request_queue)
-        if tag_match != None:
+        tag_match = re.match('(?:tag=)?(.+)', request_queue)
+        if tag_match is not None:
             return [worker for worker in workers if worker['tag'] == tag_match.group(1)]
         return []
 
--- a/tests/server/bundle_manager_test.py
+++ b/tests/server/bundle_manager_test.py
@@ -84,7 +84,7 @@ class BundleManagerTest(unittest.TestCase):
                 'cpus': 6,
                 'gpus': 0,
                 'memory_bytes': 2 * 1000,
-                'tag': None,
+                'tag': 'worker_X',
                 'run_uuids': [],
                 'dependencies': [],
                 'shared_file_system': False,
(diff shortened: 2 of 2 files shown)


codalab/server/bundle_manager.py
```diff
@@ -695,8 +695,10 @@ def _get_matched_workers(request_queue, workers):
         """
         tag_match = re.match('tag=(.+)', request_queue)
         if tag_match != None:
-            return [worker for worker in workers if worker['tag'] == tag_match.group(1)]
-        return []
+            worker_tag = tag_match.group(1)
+            if worker_tag != None:
+                return [worker for worker in workers if worker['tag'] == worker_tag]
+        return [worker for worker in workers if worker['tag'] == request_queue]
```
bkgoksel: There's a simpler way to achieve this.
With regex, you can define optional, non-capturing groups so that the matched section of text doesn't change whether or not a prefix is present or not.

For this case, it would look like 
`re.match('(?:tag=)?(.+)', request_queue)`
* the `?:` in the parenthesis makes everything else inside not be captured in a match group, and the `?` right after makes it optional, so this whole regex means "if the user typed 'tag=<something>' match '<something>' otherwise match everything they wrote.

bkgoksel (changes requested):

codalab/server/bundle_manager.py
candicegjing: Yes, this is the cleanest way to handle it.

codalab/server/bundle_manager.py
```diff
@@ -689,14 +689,13 @@ def _check_resource_failure(
...
         :param workers: a list of workers
         :return: a list of matched workers
         """
-        tag_match = re.match('tag=(.+)', request_queue)
-        if tag_match != None:
-            return [worker for worker in workers if worker['tag'] == tag_match.group(1)]
-        return []
+        tag_match = re.match('(?:tag=)?(.+)', request_queue)
+        worker_tag = tag_match.group(1) if tag_match != None else None
+        return [worker for worker in workers if worker['tag'] == worker_tag]
```
bkgoksel: Does this list comprehension work when `worker_tag == None` ?

bkgoksel (commented):
one question

codalab/server/bundle_manager.py
candicegjing: When `worker_tag == None`, if there is no `worker['tag']` equals `None`, it will return an empty list. One thing I just realized is that what if there are workers with `tag==None`, then those workers will be selected in this logic, which might not be correct. I printed out the worker object as follows:
```
workers = [{'user_id': '0', 
'worker_id': 'Stanfords-MacBook-Pro.local', 
'tag': None, 
'cpus': 4, 
'gpus': 0, 
...,
'shared_file_system': False}]
``` 
It appears that a worker without a non-empty `tag` contains the value `None` (the default value when `--tag` isn't specified when starting a worker, otherwise, it will be a string), which might cause problems when returning matching workers. Updated the logic to cover this case.

yipenghe:
Does this affect #2040 ?

candicegjing:
> Does this affect #2040 ?

This change will cover both `--request-queue tag=foo` and `--request-queue foo`. So I think it won't affect changes in #2040. If not, please let me know

candicegjing:
@bkgoksel This is ready for another review~ Let me know if I need to change anything.

bkgoksel (approved):
awesome

PR #2063, opened 2020-03-05
Use "//"  comment for editor ctrl + / behavior

### LINKED ISSUE #2035: Comment out shortcut in worksheets mapped to wrong comment syntax?
**Describe the bug**
The usual comment out shortcut (command + /) wraps text with "" while the correct comment out syntax in worksheets is just "//".

**To Reproduce**
1. Go to any editable worksheet, click "edit source"
2. Press command + / on source line

**Expected behavior**
Text gets wrapped with // or  \ corresponds to comment.

yipenghe:
fix #2035 

When we hit "ctrl + /" in the source editor, it wraps the content with \ to indicate comment, however, the backend currently doesn't support this comment type, which is inconvenient. This is to support both "//" and "\" type comments


percyliang:
I'm hesitant to do this because this suggests we support multiline comments, which we don't.  I think we should make the editor not insert &lt;!-- --&gt;.

docs/REST-API-Reference.md
```diff
@@ -175,7 +175,7 @@ Name | Type
 
 Name | Type
 --- | ---
-`id` | Integer
+`id` | CompatibleInteger
```
yipenghe, line 178: These were auto generated changes by running pre-commit

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -980,6 +981,30 @@ class Worksheet extends React.Component {
                         this.discardChanges();
                     }.bind(this),
                 });
+                editor.commands.addCommand({
+                    name: 'comment',
+                    bindKey: { win: 'Ctrl+/', mac: 'cmd+/' },
+                    exec: function() {
+                        let selectionRange = editor.getSelectionRange();
+                        selectionRange.start.column = 0;
```
yipenghe: This moves the selection range to start at the beginning of the first line

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -980,6 +981,30 @@ class Worksheet extends React.Component {
...
+                        let content = editor.session.getTextRange(selectionRange);
+                        let splitLines = content.split('\n');
+                        let commentAll = true;
+                        if (splitLines[0].startsWith('//')) {
```
yipenghe: Our rule will be: if the first selected line is in commented status, we uncomment every line, else we comment every line

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -956,7 +956,9 @@ class Worksheet extends React.Component {
             editor.$blockScrolling = Infinity;
             editor.session.setUseWrapMode(false);
             editor.setShowPrintMargin(false);
-            editor.session.setMode('ace/mode/markdown');
+            editor.session.setMode('ace/mode/markdown', function() {
+                editor.session.$mode.blockComment = { start: '//', end: '' };
```
epicfaace, line 960: Adding a space makes it feel more natural and is more in line with what other IDEs do.

```suggestion
                editor.session.$mode.blockComment = { start: '// ', end: '' };
```

epicfaace (approved):

frontend/src/components/worksheets/Worksheet/Worksheet.js
yipenghe: While this is true, if someone manually adds "//" at the beginning, then using '// ' will make it "// //" instead of uncommenting it. If that makes sense

epicfaace: I see. I think we shouldn't use `blockComment` here. Can we use `lineCommentStart` instead?

See https://github.com/ajaxorg/ace/blob/9b5b63d1dc7c1b81b58d30c87d14b5905d030ca5/tool/tmlanguage.js#L655-L661

yipenghe: Hmmm, but if you select multiple line, it triggers blockComment I think. Moreover, the current behavior is to add \, not any of those that you highlighted

epicfaace: Ok, seems more complicated than I initially thought. The current "//" behavior is fine then.

Issue #1913, opened 2020-01-20
Many subsequent runs with large Docker images causes workers to go offline


teetone:
Creating many runs with large Docker images causes workers to go offline consistently. We need to investigate what's causing the workers to go offline.

To reproduce this issue use stress_test.py to create about 3000 bundles with large docker images (run `python stress_test.py --help` for more information on how to run the script).

teetone:
@percyliang, @candicegjing  Just a quick update, as of 3/9/2020, this is no longer an issue because of the LRU caching added in `docker_image_manager.py`. I stressed test `dev` with 3000 bundles with large Docker images, then 2000 bundles with large Docker images and the workers were able to finish running the jobs without any problems.

This was the state of the workers in beginning of the stress testing:

```
worker_id             cpus  gpus  memory  free_disk  last_checkin  tag  runs                                 shared_file_system
-------------------------------------------------------------------------------------------------------------------------------
vm-clws-dev-worker-0  3/4   0/0   13.7g   104g       3.0s ago           0x644c7c,0xa2e5e4,0x782082           False
vm-clws-dev-worker-1  3/4   0/0   13.7g   173g       4.0s ago           0x591bbd,0x3a75b2,0x7f822d           False
vm-clws-dev-worker-2  3/4   0/0   13.7g   173g       3.0s ago           0xfa7ed6,0xb8f65f,0x0ad921           False
vm-clws-dev-worker-3  3/4   0/0   13.7g   173g       2.0s ago           0x33a9eb,0x6a0048,0x1c2607           False
vm-clws-dev-worker-4  4/4   0/0   13.7g   173g       4.0s ago           0x80385d,0x170b46,0xdb12f9,0x486985  False
vm-clws-dev-worker-5  4/4   0/0   13.7g   173g       5.0s ago           0xafa39c,0x810bd0,0xd0fbea,0x9cd28d  False
```

This was the state of the workers, after the first round of the stress test:

```
worker_id             cpus  gpus  memory  free_disk  last_checkin  tag  runs      shared_file_system
----------------------------------------------------------------------------------------------------
vm-clws-dev-worker-0  1/4   0/0   13.7g   106g       2.0s ago           0x098208  False
vm-clws-dev-worker-1  0/4   0/0   13.7g   152g       1.0s ago                     False
vm-clws-dev-worker-2  0/4   0/0   13.7g   152g       1.0s ago                     False
[...]

PR #2076, opened 2020-03-10
Update stress test script.

teetone:
Updated stress test script for https://github.com/codalab/deployment/pull/16 and stress testing #1913 (many runs with large docker images).

- Allow users to pass in the instance (default to `localhost`) they want to run the stress tests against
- Added more large Docker images examples (20 total)
- Adjusted the heavy run's configurations (a bit lighter)
- Disable the disk write tests in the heavy run for now.


stress_test.py
```diff
@@ -267,15 +278,19 @@ def main():
...
         print('Setting the heavy configuration...')
         args.large_file_size_gb = 20
-        args.bundle_upload_count = 2000
-        args.create_worksheet_count = 2000
-        args.parallel_runs_count = 1000
-        args.large_docker_runs_count = 100
+        args.bundle_upload_count = 1000
+        args.create_worksheet_count = 1000
+        args.parallel_runs_count = 500
+        args.large_docker_runs_count = 2000
```
teetone, line 290: Changed `large_docker_runs_count` from 100 to 2000 because previously, we used to cycle through the list of 10 Docker example images 100 times (1000 runs total), but now instead we randomly pick an image for 2000 runs total. It makes the parameter name less ambiguous too.

stress_test.py
```diff
@@ -69,32 +70,47 @@ def delete(self):
...
+    This runner object holds the logic on how the stress tests are run based on the command line
+    arguments passed in.
 
     Args:
         cl: cl instance used to run CodaLab commands
         args: command line arguments
     """
 
-    # List of random, large docker images
+    # List of random, large Docker images in alphabetical order
```
epicfaace, line 81: Why specify "in alphabetical order"?

stress_test.py
epicfaace: @teetone What's the point of randomly picking a Docker image? It makes the test non-deterministic, which we might want to avoid. (if one of the docker images suddenly stops working / breaks, we would rather that that causes a deterministic failure of our tests than a non-deterministic failure)

stress_test.py
teetone: This is just for better organization. It's easy to which images are already in the list.

stress_test.py
teetone: @epicfaace In the previous version of the test, It will just cycle through the list of large Docker images. `docker_image_manager.py` uses an LRU cache to store the Docker images. With an LRU cache, it will simply kick out the images earlier in the list, dealing with the same sequence of image sizes over and over again. Now if we start picking images randomly, the LRU cache will have to deal with varying sizes of Docker images coming in, which better mimics real-world behavior.

To your second point, we select an image from a list of 20 images uniformly at random 2000 times. Each image in the list should be used multiple times. If a specific image is problematic, it will be caught by this stress test and fail. Also, our main concern and the point of the stress test is to see if our system can create a bunch of bundles with large Docker images and not if a certain image causes failures or not.

epicfaace (approved):

epicfaace (approved):

PR #2086, opened 2020-03-12
Async loading of worksheets by separating out search/wsearch

epicfaace:
Based on #1963. Async loading of worksheets (#1234).

Adds two parameters to the /rest/interpret endpoint:

`/rest/interpret?brief=1` is called first, renders everything but doesn't expand the `search` and `wsearch` directives.

The client can then call `/rest/interpret?directive=search 0x .limit=100` (URL encoded) for each `search` and `wsearch` directive, putting in the directive text in the `directive` parameter. I've added a semaphore to throttle these kinds of requests to be a max of 3 concurrent requests; thus, a worksheet with many `search`/`wsearch` directives will progressively resolve each of these directives from top to bottom.

Loading view (it's animated):

image

Time for the dashboard (first request loads the actual worksheet, additional requests are async requests):

image

Previous timeline for the single interpret request:

image


epicfaace:
@teetone I'm getting this test failure on CI -- do you know why it's happening?

```bash
12.92s$ python3 tests/ui/ui_tester.py --headless
  File "tests/ui/ui_tester.py", line 184, in <module>
    main()
  File "tests/ui/ui_tester.py", line 172, in main
    test.run()
  File "tests/ui/ui_tester.py", line 41, in run
    self.test()
  File "tests/ui/ui_tester.py", line 159, in test
    self.click_link('Small Worksheet [cl_small_worksheet]')
  File "tests/ui/ui_tester.py", line 62, in click_link
    link = self._driver.find_element_by_link_text(selector)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 428, in find_element_by_link_text
    return self.find_element(by=By.LINK_TEXT, value=link_text)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 978, in find_element
    'value': value})['value']
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.NoSuchElementException: Message: Unable to locate element: Small Worksheet [cl_small_worksheet]
The command "python3 tests/ui/ui_tester.py --headless" exited with 1.
```

codalab/lib/bundle_cli.py
```diff
@@ -3109,6 +3109,8 @@ def display_blocks(self, client, worksheet_info, interpreted_blocks):
                         '[Worksheet ' + self.simple_worksheet_str(worksheet_info) + ']',
                         file=self.stdout,
                     )
+            elif mode == BlockModes.placeholder_block:
+                print('[Placeholder]', block['directive'], file=self.stdout)
```
yipenghe, line 3153: Do we want to have the print?

teetone:
@epicfaace The error happens when it can't find the link `Small Worksheet [cl_small_worksheet]` on the dashboard worksheet. I'm not sure what elements show up when the async loading is done but I think you might need to wait on some other part of the page now.

codalab/rest/interpret.py
```diff
@@ -563,25 +577,13 @@ def resolve_items_into_infos(items):
...
-    resolve_items_into_infos on the returned raw_items.
-    """
-
-    (bundle_info, subworksheet_info, value_obj, item_type, id, sort_key) = raw_item
-
-    is_search = item_type == TYPE_DIRECTIVE and get_command(value_obj) == 'search'
-    is_wsearch = item_type == TYPE_DIRECTIVE and get_command(value_obj) == 'wsearch'
-
+    Perform a search query and return the resulting raw items.
+    Input: directive that is tokenized by formatting.string_to_tokens(), such as formatting.string_to_tokens("search 0x .limit=100")"""
```
yipenghe: break into two lines maybe?

codalab/rest/interpret.py
```diff
@@ -563,25 +577,13 @@ def resolve_items_into_infos(items):
...
+    command = get_command(value_obj)
+    is_search = command == 'search'
+    is_wsearch = command == 'wsearch'
     if is_search or is_wsearch:
         command = get_command(value_obj)
```
yipenghe: this could be removed if we now have line 584

codalab/rest/interpret.py
```diff
@@ -608,5 +610,25 @@ def expand_raw_item(raw_item):
                 raw_items.append(subworksheet_item(worksheet_info) + (None, None))
 
         return raw_items
+    else:
+        raise Exception("not a search query")
```
yipenghe: It seems a bit dangerous and confusing that this method throws but line 177 doesn't try to catch it. Would it be better to just return a empty array if we don't expect it to fail. If we want to enforce that the input must be a search query, better to document it in the function comment block

codalab/rest/interpret.py
```diff
@@ -608,5 +610,25 @@ def expand_raw_item(raw_item):
...
+    """
+    Raw items that include searches must be expanded into more raw items.
+    Input: Raw item.
+    Output: Array of raw items. If raw item does not need expanding,
+    this returns an 1-length array that contains original raw item,
+    otherwise it contains the search result. You do not need to call
+    resolve_items_into_infos on the returned raw_items.
+    """
+    (bundle_info, subworksheet_info, value_obj, item_type, id, sort_key) = raw_item
+    if item_type == TYPE_DIRECTIVE and get_command(value_obj) in ('search', 'wsearch'):
```
yipenghe, line 634: Here we check if a raw item is either 'search' or 'wsearch', but in line 177, we don't check the conditions. I think we should be consistent in a way either: calling perform_search_query requires checking get_command(value_obj) in ('search', 'wsearch')  or it's safe to call it even if the input is not a search object. If this makes sense

codalab/lib/bundle_cli.py
epicfaace: I added it for uniformity with the other print statements that exist for each block mode.

codalab/rest/interpret.py
epicfaace: I see what you mean, but in this case, we have to check if `get_command(value_obj) in ('search', 'wsearch')` because there are some directives that don't correspond to search queries. If we were to move the `get_command(value_obj) in ('search', 'wsearch')` logic into `perform_search_query`, then we would be calling `perform_search_query` even on non-search directives.

Currently, it's safe to call `perform_search_query` even if the input is not a search object -- it's just that we don't _want_ to call it (we want to instead just return `[raw_item]`) when the input is not a search object.

epicfaace:
Talked to @yipenghe about changes needed to get this working with focus index. We concluded:
- the only thing that needs to be changed is to add an empty placeholder item (it could look like an empty table) when search results return nothing. This is because, currently, search results that return nothing take up 0 items, but a placeholder block uses up one item -- so the focus issues are caused by this discrepancy when we have a placeholder block that takes up one item that then gets converted to a search result that takes up 0 items.
- we should look at refactoring raw_to_block and block_to_raw at another date, but this isn't necessary for this PR.

epicfaace:
Another issue -- which we missed yesterday -- is that you can't select or scroll between items in a table anymore.

epicfaace:
Latest updates added should resolve these issues:
- Add onAsyncItemLoad callback to update `this.state.ws.info.items` when a placeholder item has fully finished loading
- Show an item called "No results found" whenever no search results are found
- skip "No results found" when scrolling with the keyboard

epicfaace:
@yipenghe this should be ready for review now

yipenghe:
shift + r seems to break

epicfaace:
@yipenghe good catch. This was a really strange error caused by the fact that we were storing objects with functions in them in state (so when I tried to update state with the spread operator `...this.state.ws,`, those functions were gone). Had to do some refactoring -- can you review the latest changes and test it now?

yipenghe:
The code looks good and shift+r works well. However, I noticed that with the new refactored version, the frontend feels much laggy when scrolling, while the old one scrolls much smoother.

I'm not sure if it's only my device's problem, but to see what I mean, try:
1. git checkout async-search
2. ./codalab_service.py start -bd
3. switch between async-search and the master branch, and scroll on a large worksheet (dashboard possibly)

yipenghe (changes requested):
Another thing is now if we add a new run on a worksheet that has (w)search items, it will trigger the search boxes to be re-rendered as placeholder items and re-expand, whereas adding new runs in the current frontend is smoother. This is particularly obvious if we are adding new runs on large worksheets like dashboards

epicfaace:
I've fixed another issue in which too many requests were happening (like 20 instead of 3) because the `useEffect` function in `PlaceholderItem` was called too often whenever its parent re-rendered, rather than only on mount. This might help with the slowness.

epicfaace:
> Another thing is now if we add a new run on a worksheet that has (w)search items, it will trigger the search boxes to be re-rendered as placeholder items and re-expand, whereas adding new runs in the current frontend is smoother. This is particularly obvious if we are adding new runs on large worksheets like dashboards

@yipenghe I don't think we can stop the reloading of search results because adding a bundle may update those values. I've pushed a change that makes sure that the placeholder height is the same height as the original table upon worksheet update, though. Let me know if that works for you.

yipenghe:
> > Another thing is now if we add a new run on a worksheet that has (w)search items, it will trigger the search boxes to be re-rendered as placeholder items and re-expand, whereas adding new runs in the current frontend is smoother. This is particularly obvious if we are adding new runs on large worksheets like dashboards
> 
> @yipenghe I don't think we can stop the reloading of search results because adding a bundle may update those values. I've pushed a change that makes sure that the placeholder height is the same height as the original table upon worksheet update, though. Let me know if that works for you.

Is it possible to fix the search results until user refresh the page? That is, if a new bundle changes a search result, it's not shown until the next hard refresh. If it's too troublesome, we can do the current thing, since it's not obvious unless a user have a super big search table (which mostly is on dashboard and user can't change it anyway)

yipenghe (approved):
I think the code is good, after the request time change the laggyness goes away once the whole worksheet is loaded. Feel free to merge if you think fixing the search result is not ideal / makes code much more complicated

yipenghe:
Do you mind also pushing a commit to #2145 with the search logic changes?

epicfaace:
> Is it possible to fix the search results until user refresh the page?

Let's do this in a later issue -- this is a bigger change, because currently if you add a run / edit a cell, the entire worksheet reloads -- would like to keep that behavior in this PR and change it later if needed.

yipenghe:
> > Is it possible to fix the search results until user refresh the page?
> 
> Let's do this in a later issue -- this is a bigger change, because currently if you add a run / edit a cell, the entire worksheet reloads -- would like to keep that behavior in this PR and change it later if needed.

Sounds good, the bigger change can also speed up pasting, so we could work on it later together, good to merge

PR #2091, opened 2020-03-16
Travis fix

candicegjing:
Fixed #2090.

This PR fixed the issue in Travis build.

PR #2093, opened 2020-03-18
Support tagging CodaLab's public instances

candicegjing:
This is the changes done in the reverted PR #2031 .


epicfaace (commented):
Can we try to figure out what's causing the build failure? I would suggest dividing this PR into changes, then reverting one change at a time so you can see which change caused the build failure.

candicegjing:
> Can we try to figure out what's causing the build failure? I would suggest dividing this PR into changes, then reverting one change at a time so you can see which change caused the build failure.

The build failure is fixed in https://github.com/codalab/codalab-worksheets/pull/2091. It's merged into master yesterday. (Please do double-check.) Build should be good for now. This PR is not the root cause of the build failure.

PR #2104, opened 2020-03-23
move cpu and gpu resource check to the beginning of _transition_from_…

### LINKED ISSUE #2096: Incorrect number of available CPUs/GPUs in NLP cluster when using Slurm
When we login to NLP cluster and then request a certain number of computing resources (e.g. CPUs, GPUs, memory) via Slurm, e.g. `nlprun -g 1 -c 3 --memory 16g -m jagupard25`, this means that we are requesting for 1 GPU, 3 CPUs and 16 memory. However, we are seeing 40 CPUs in total reported when running workers command (see the second item in the list) 
Screen Shot 2020-03-20 at 4 48 27 PM
This could cause over assigning issues from the bundle manager to a private worker if the available CPUs/GPUs are incorrectly reported.

candicegjing:
Fixed #2096.

#2096 is caused by an outdated codalab package installed in NLP machine. While I was reading the code, I found that the resource check on CPU and GPU happens after the dependency downloading stage. I guess the order could be reversed so that the program can check for CPUs/GPUs first before it tries to download dependencies. Let me know if this is something we want to do.

### DIFF
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -272,7 +272,11 @@ class RunStateMachine(StateTransitioner):
         else:
             docker_network = self.docker_network_internal.name
 
+<<<<<<< HEAD
         # 3) Start container
+=======
+        # 4) Start container
+>>>>>>> 0f2c9526... move cpu and gpu resource check to the beginning of _transition_from_PREPARING
         try:
             container = docker_utils.start_bundle_container(
                 run_state.bundle_path,
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -164,6 +164,17 @@ class RunStateMachine(StateTransitioner):
         if run_state.is_killed:
             return run_state._replace(stage=RunStage.CLEANING_UP)
 
+        # Check CPU and GPU availability
+        try:
+            cpuset, gpuset = self.assign_cpu_and_gpu_sets_fn(
+                run_state.resources.cpus, run_state.resources.gpus
+            )
+        except Exception as e:
+            message = "Unexpectedly unable to assign enough resources: %s" % str(e)
+            logger.error(message)
+            logger.error(traceback.format_exc())
+            return run_state._replace(run_status=message)
+
         dependencies_ready = True
         status_messages = []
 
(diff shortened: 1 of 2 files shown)


codalab/worker/worker_run_state.py
```diff
@@ -262,16 +273,6 @@ def _transition_from_PREPARING(self, run_state):
         else:
             docker_network = self.docker_network_internal.name
 
-        try:
```
percyliang, line 265: The "3) Set up container" comment needs to be updated then, because it's not a good comment for just setting the `docker_network`.  Maybe just remove "3)" as a step.

codalab/worker/worker_run_state.py
```diff
@@ -164,6 +164,17 @@ def _transition_from_PREPARING(self, run_state):
         if run_state.is_killed:
             return run_state._replace(stage=RunStage.CLEANING_UP)
 
+        # Check CPU and GPU availability
+        try:
+            cpuset, gpuset = self.assign_cpu_and_gpu_sets_fn(
+                run_state.resources.cpus, run_state.resources.gpus
+            )
+        except Exception as e:
+            message = "Cannot assign enough resources: %s" % str(e)
```
percyliang: This really shouldn't happen - the bundle manager would have screwed up if we got here...  Make the message clear that this is unexpected.  Do we see this message in practice?

candicegjing: Yes, if everything goes well, it shouldn't happen. I guess this piece of logic would be helpful in the following two conditions:
1. bundle manager screws up at resource counting then leads to failures in assigning jobs to workers
1. there are bugs in cpu/gpu/memory reporting somewhere (e.g. worker side)

codalab/worker/worker_run_state.py
candicegjing: That's true. Removed.

codalab/worker/worker_run_state.py
percyliang: Can you update the `message`?  "... %s (is happening due to bugs or errors in the bundle manager assignment)"

percyliang (approved):

codalab/worker/worker_run_state.py
candicegjing: I should have updated this yesterday. One thing that I just realized is that this message will show up on the bundle page in the metadata field. Do we want to mention the reason on the Web UI? Or we log this message separately in the host machine (not pass it to `message`)?

percyliang: Ah, since it's going to show up in the UI, then we can maybe make it a bit more friendly, like "Unexpectedly unable to assign ..."

PR #2105, opened 2020-03-25
Change bundle dependency from dict to list

### LINKED ISSUE #2021: [USER ISSUE] Multiple aliases pointing to the same bundle spec are not supported
**Describe the bug**
When I set two bundle dependencies with the same bundle spec but different aliases, one of them is get dropped.

**Codalab Information**
- *Your username*: hfang
- *UUID of problem bundle (please don't delete)*: 0x218c2d796c9847a283817362459eb308
- *Error message received from Codalab (if applicable)*: N/A
 - *OS (if on CLI)*: N/A
 - *Codalab CLI Version (if on CLI)*:  N/A

**To Reproduce**
```
run foo:0xea12e8bbfa9d4e0788dfef6457df7cee/foo foo2:0xea12e8bbfa9d4e0788dfef6457df7cee --- "cat foo2"
```

**Expected behavior**
Both "foo" and "foo2" should be accessible and shown as dependencies.

**Screenshots**
N/A

**Additional context**
This is a sample job to illustrate the problem.
It happens for a complicated worksheet in practice, when sometimes the two dependencies can be different, but sometimes they point to the same bundle spec.

candicegjing:
Fixed #2021 

This PR will support multiple keys map to single value, e.g. `foo:<uuid_1>, foo2:<uuid_1>`. 
A real example could be as follows:
Screen Shot 2020-03-24 at 9 12 38 PM

### DIFF
--- a/codalab/worker/bundle_state.py
+++ b/codalab/worker/bundle_state.py
@@ -89,14 +89,14 @@ class BundleInfo(object):
                 ),
             )
             for dep in dependencies
-        ]  # type: List[Dependency]
+        ]  # type: List[(DependencyKey, Dependency)]
 
         self.location = location  # set if local filesystem
 
     @property
     def as_dict(self):
         dct = generic_to_dict(self)
-        dct['dependencies'] = [v._asdict() for k, v in dct['dependencies']]
+        dct['dependencies'] = [generic_to_dict(v) for k, v in dct['dependencies']]
         return dct
 
     def __str__(self):
--- a/codalab/worker/bundle_state.py
+++ b/codalab/worker/bundle_state.py
@@ -77,23 +77,26 @@ class BundleInfo(object):
         self.metadata = metadata
         self.args = args
         self.dependencies = [
-            Dependency(
-                parent_name=dep["parent_name"],
-                parent_path=dep["parent_path"],
-                parent_uuid=dep["parent_uuid"],
-                child_path=dep["child_path"],
-                child_uuid=dep["child_uuid"],
-                location=dep.get("location", None),
+            (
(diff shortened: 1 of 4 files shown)


codalab/worker/bundle_state.py
```diff
@@ -76,23 +76,27 @@ def __init__(
...
-            DependencyKey(dep["parent_uuid"], dep["parent_path"]): Dependency(
-                parent_name=dep["parent_name"],
-                parent_path=dep["parent_path"],
-                parent_uuid=dep["parent_uuid"],
-                child_path=dep["child_path"],
-                child_uuid=dep["child_uuid"],
-                location=dep.get("location", None),
+        self.dependencies = [
+            (
+                DependencyKey(dep["parent_uuid"], dep["parent_path"]),
```
percyliang, line 81: Don't need the key if it's a list because it's already in `Dependency`

percyliang (commented):
`self.dependencies` isn't used for dict lookup anywhere?  I would have assumed that the key was there for a reason...

candicegjing:
> `self.dependencies` isn't used for dict lookup anywhere? I would have assumed that the key was there for a reason...

I searched through our codebase and doubt that it has been used anywhere else other than this file. @bkgoksel Could you please verify in case I overlooked some details?

codalab/worker/bundle_state.py
candicegjing: That's true. I have updated the code.

percyliang (approved):
LGTM!

percyliang (approved):

candicegjing:
Close this PR as we decide to go with another solution which is documented at #2115

PR #2113, opened 2020-03-27
Adjust timezone information when calculating current container runnin…

### LINKED ISSUE #2112: [USER ISSUE] Bundles on jagupards have incorrect time while running
**Describe the bug**

Jobs that run on the jagupards (hosts in the Stanford NLP cluster) have incorrect timestamps while running—the durations are properly calculated when the jobs are finished. For instance, https://worksheets.codalab.org/bundles/0x477a3881d0184d4cbb212ad3c5c5b83f shows that the time is `-1y364d`. I guess there's some time subtraction going on based on what docker reports, and the time zone of the container might differ from codalab's server time?

**To Reproduce**
1. Run a bundle on a jagupard machine.

**Expected behavior**

Duration is properly calculated. 

Screen Shot 2020-03-26 at 1 26 00 PM

candicegjing:
Fixed #2112 

Based on documentation at [`datetime.datetime.replace`](https://docs.python.org/3/library/datetime.html#datetime.datetime.replace), it doesn't convert the date and time data to UTC timezone. In NLP cluster machine, since it's using Los Angeles time zone, the time zone calculation doesn't report the correct value. This PR is to fix this issue.

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -266,7 +266,7 @@ def get_container_running_time(container):
     end_time = (
         container.attrs['State']['FinishedAt']
         if container.attrs['State']['Status'] == 'exited'
-        else str(datetime.datetime.now().replace(tzinfo=tz.tzutc()))
+        else str(datetime.datetime.now(tz.tzutc()))
     )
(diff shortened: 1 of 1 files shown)


percyliang (approved):
Thanks!

PR #2115, opened 2020-03-27
Multi alias

### LINKED ISSUE #2021: [USER ISSUE] Multiple aliases pointing to the same bundle spec are not supported

candicegjing:
Fixed #2021

### DIFF
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -289,6 +289,10 @@ class RunStateMachine(StateTransitioner):
                 runtime=self.docker_runtime,
             )
             self.worker_docker_network.connect(container)
+        except docker_utils.DockerUserErrorException as e:
+            message = 'Cannot start Docker container: {}'.format(e)
+            logger.warning(message)
+            return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
         except Exception as e:
             message = 'Cannot start Docker container: {}'.format(e)
             logger.error(message)
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -10,7 +10,7 @@ import codalab.worker.docker_utils as docker_utils
 
 from codalab.lib.formatting import size_str, duration_str
 from codalab.worker.file_util import remove_path, get_path_size
-from codalab.worker.bundle_state import State
+from codalab.worker.bundle_state import State, DependencyKey
 from codalab.worker.fsm import DependencyStage, StateTransitioner
 from codalab.worker.worker_thread import ThreadDict
 
@@ -169,7 +169,8 @@ class RunStateMachine(StateTransitioner):
 
         if not self.shared_file_system:
             # No need to download dependencies if we're in the shared FS since they're already in our FS
-            for dep_key, dep in run_state.bundle.dependencies.items():
+            for dep in run_state.bundle.dependencies:
+                dep_key = DependencyKey(dep.parent_uuid, dep.parent_path)
                 dependency_state = self.dependency_manager.get(run_state.bundle.uuid, dep_key)
                 if dependency_state.stage == DependencyStage.DOWNLOADING:
                     status_messages.append(
@@ -233,7 +234,8 @@ class RunStateMachine(StateTransitioner):
         docker_dependencies_path = (
             '/' + run_state.bundle.uuid + ('_dependencies' if not self.shared_file_system else '')
         )
-        for dep_key, dep in run_state.bundle.dependencies.items():
+        for dep in run_state.bundle.dependencies:
+            dep_key = DependencyKey(dep.parent_uuid, dep.parent_path)
             full_child_path = os.path.normpath(os.path.join(run_state.bundle_path, dep.child_path))
             if not full_child_path.startswith(run_state.bundle_path):
                 # Dependencies should end up in their bundles (ie prevent using relative paths like ..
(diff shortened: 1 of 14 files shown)


codalab/worker/worker_run_state.py
```diff
@@ -439,7 +443,9 @@ def _transition_from_CLEANING_UP(self, run_state):
                     logger.error(traceback.format_exc())
                     time.sleep(1)
 
-        for dep_key, dep in run_state.bundle.dependencies.items():
+        for dep in run_state.bundle.dependencies:
+            dep_asdict = dep._asdict()
+            dep_key = DependencyKey(dep_asdict["parent_uuid"], dep_asdict["parent_path"])
```
percyliang: Why can't you access these fields directly from `dep`?

candicegjing: That's true! I have updated the code to access those fields directly from `dep`.

percyliang (approved):
LGTM!

percyliang:
Oh, would you mind adding a test to make sure this works?

candicegjing:
> Oh, would you mind adding a test to make sure this works?

There is a test suite created for `BundleInfo` object. I just reorganized it and added a bit more test cases to make the comparison clear, especially the part related to `dependencies`.

tests/worker/bundle_state_test.py
```diff
@@ -33,17 +33,66 @@ def test_bundle_info_serialization(self):
             ],
             'location': 'location',
         }
+
+    def tearDown(self) -> None:
+        del self.bundle_info_fields
+
+    def test_bundle_info_serialization(self):
```
percyliang, line 40: When I said adding a test, I meant not testing the serialization (which is not the main part of the logic), but rather the `cl` command in `test_cli.py` that has references the same bundle twice.  Basically a minimal test case that didn't work before but works now.

candicegjing: Oh, I see. I originally thought since we have altered the `dependencies` logic, it would be helpful to have tests that explicitly verify that.  But I agree that testing on `cl` command is more important. I have added two test cases in `test_cli.py`.

test_cli.py
```diff
@@ -1053,6 +1053,26 @@ def test(ctx):
         [cl, 'run', 'cat %%%s//%s%%/stdout' % (source_worksheet_full, name)], expected_exit_code=1
     )
 
+    # Multi-alias tests
+    multi_alias_uuid1 = _run_command(
+        [cl, 'run', 'foo:{}'.format(uuid), 'foo1:{}'.format(uuid), 'echo "two alias"']
+    )
+    wait(multi_alias_uuid1)
+    check_equals('ready', _run_command([cl, 'info', '-f', 'state', multi_alias_uuid1]))
```
percyliang: I think `wait` returns whether it failed or not, so you don't need a separate `cl info` call.

candicegjing: Yes. `Ready` is checked in `wait`. I removed all other similar cases as well in `test_cli.py`

test_cli.py
```diff
@@ -1045,14 +1042,33 @@ def test(ctx):
...
     check_equals(sugared_remote_uuid, _run_command([cl, 'search', sugared_remote_name, '-u']))
-    check_equals('ready', _run_command([cl, 'info', '-f', 'state', sugared_remote_uuid]))
     check_equals('hello', _run_command([cl, 'cat', sugared_remote_uuid + '/stdout']))
 
     # Explicitly fail when a remote instance name with : in it is supplied
     _run_command(
         [cl, 'run', 'cat %%%s//%s%%/stdout' % (source_worksheet_full, name)], expected_exit_code=1
     )
 
+    # Multi-alias tests
```
percyliang: Not clear what "multi-alias" means...explain: "Multiple keys pointing to the same bundle"

test_cli.py
```diff
@@ -1045,14 +1042,33 @@ def test(ctx):
...
+    check_equals('two aliases', _run_command([cl, 'cat', multi_alias_uuid1 + "/stdout"]))
+
+    multi_alias_uuid2 = _run_command(
+        [
+            cl,
+            'run',
+            'foo:{}'.format(uuid),
+            'foo1:{}'.format(uuid),
+            'foo2:{}'.format(uuid),
+            'echo "three aliases"',
```
percyliang, line 1062: It would be a nicer test if we `cat foo foo1 foo2` and make sure that they actually have the correct contents.  I think we don't need three aliases if we just have two.

candicegjing: Previously, two aliases worked well during my local testing. But three aliases one failed without the changes that I added later on. So I guess I'll just keep the three aliases test to be safe.

test_cli.py
candicegjing: Updated!

percyliang (approved):
Great, thanks!

PR #2117, opened 2020-03-29
Add tag_exclusive workers that only run bundles with matching tags

### LINKED ISSUE #2097: [FEATURE REQUEST] workers that _only_ run bundles with a matching tag

Currently, tagged codalab workers will run bundles without a `request_queue` set. Would it be possible to have some flag that forces a worker to *only* run bundles with a particular tag?

Rationale: Sometimes, I have a lot of bundles running that are untagged. I've already developed these bundles, and I'm sure they'll work—I'm just letting them run (think 50+ staged jobs).

In parallel, though, I'd often like to set up the next round of experiments, which sometimes means running stuff on codalab and seeing if it works. However, the development cycle is super slow since it's hard to get a worker—they're all busy running the untagged jobs waiting in the queue (even if I give the worker a special tag). It'd be nice if, in this case, these dev jobs were able to instantly land somewhere (assuming there isn't another one before it).

Not sure what a solution would look like here—maybe an extra flag on `cl-worker` that locks it to running bundles with a particular tag? If you use the flag with an untagged worker, it throws an error or something.

nelson-liu:
Fixes #2097 

This PR adds a new field to workers, `tag_exclusive` (not sold on the name, feel free to suggest alternatives).

When true, workers _only_ run bundles that match their tag. When false, existing behavior should be preserved (i.e., tagged workers can run bundles without a specified `request_queue`).

I'll be the first to admit that I'm pretty unfamiliar with the codebase, might want to review this one extra carefully 🧐


percyliang:
Thanks for being proactive about this!  It would be nice if we can avoid making schema changes, so let's discuss in #2097.

codalab/lib/bundle_cli.py
```diff
@@ -1052,6 +1052,7 @@ def do_workers_command(self, args):
             'tag',
             'runs',
             'shared_file_system',
+            'tag_exclusive',
```
percyliang, line 1066: Put this right after `tag` so that there's better conceptual grouping, same for the rest of the additions.

codalab/worker/main.py
```diff
@@ -120,6 +120,11 @@ def parse_args():
         action='store_true',
         help='To be used when the server and the worker share the bundle store on their filesystems.',
     )
+    parser.add_argument(
+        '--tag-exclusive',
+        action='store_true',
+        help='To be used when the worker should only run bundles that match its tag.',
```
percyliang: disambiguate pronoun: "that match the worker's tag"

codalab/server/bundle_manager.py
```diff
@@ -428,12 +437,14 @@ def get_sort_key(worker):
...
-            # 4. prioritize workers with fewer running jobs.
-            # 5. break ties randomly by a random seed.
+            # 1. prioritize workers that are tag-exclusive.
+            # 2. prioritize workers with fewer GPUs (including zero).
+            # 3. prioritize workers that have more bundle dependencies.
+            # 4. prioritize workers with fewer CPUs.
+            # 5. prioritize workers with fewer running jobs.
+            # 6. break ties randomly by a random seed.
             return (
+                -int(worker['tag_exclusive']),
```
percyliang: Can we do `not worker['tag_exclusive']`?

percyliang (approved):
LGTM - just a few minor nits.

One thing i was thinking was whether in the future we wanted a more powerful tag/matching system, where each worker / run had a set of tags, some of which were exclusive versus not.  Not sure if there's a pressing real use case for this, but perhaps worth keeping in mind.

percyliang (commented):
Another thing: It'd be good to write somewhere, perhaps as a comment block or in the documentation, a description of how one typically uses `--tag-exclusive`.

nelson-liu:
> One thing i was thinking was whether in the future we wanted a more powerful tag/matching system, where each worker / run had a set of tags, some of which were exclusive versus not. Not sure if there's a pressing real use case for this, but perhaps worth keeping in mind.

Not sure how the old slurm worker functioned, but a more full-featured tag matching system could be useful there—you could have a worker with tags for each of its slurm attributes (e.g., the gpu type), and bundles could request the combination of attributes they want (via `||` and `&&`).

alembic/versions/2020032904_add_tag_exclusive_workers_9d0f1ffb18e9.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '9d0f1ffb18e9'
+down_revision = '75d4288ae265'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('worker', sa.Column('tag_exclusive', sa.Boolean(), nullable=False, server_default='0'))
```
candicegjing: I found we have several places in our codebase have inconsistency between database definition and the real value that we are retrieving/setting. For example, we are setting the new column `tag_exclusive` with the default value `0` in the database. However, when we process this field later in either worker side or bundle-manager, we treat it as `True` or `False`. In the code logic, it doesn't seem to have a huge impact on the functionality. I just think it might be a bit more clear to make them consistent so that we don't need to guess around what the real value is.

codalab/lib/bundle_cli.py
nelson-liu: I assume you don't want me to change function signatures, and only change things where it's a cosmetic fix? Correct me if I'm wrong.

codalab/server/bundle_manager.py
```diff
@@ -388,6 +388,15 @@ def _filter_and_sort_workers(self, workers_list, bundle, bundle_resources):
...
         if bundle.metadata.request_queue:
             workers_list = self._get_matched_workers(bundle.metadata.request_queue, workers_list)
+        else:
+            # The bundle is untagged, so we want to keep workers that are not
+            # tag-exclusive or don't have a tag defined.
+            # (removing workers that are tag_exclusive and have tags defined).
+            workers_list = [
+                worker
+                for worker in workers_list
+                if not worker['tag_exclusive'] or not worker['tag']
```
candicegjing, line 398: I just want to understand the relationship between `tag` and `tag_exclusive`? My understanding is that there could be 4 possibilities as follows:

| | tag |  tag_exclusive |
| -- | ------------- | ----- |
|1 | null | True |
|2| tagged | True |
|3| null | False |
|4| tagged | False | 

`tag_exclusive` will take effect in the 2nd and the 4th case, right? Not sure if I understand it correctly, let me know

nelson-liu: Under the current implementation of tagging, yes. Note that the logic still applies in case (1)---the worker would take only untagged bundles. But this is the current behavior, anyway.

alembic/versions/2020032904_add_tag_exclusive_workers_9d0f1ffb18e9.py
nelson-liu: Do you want me to address this in this PR, or is this a separate issue?

candicegjing: Could we change the `server_default` value to be `True` or `False`? Not sure if this is the right syntax though.

codalab/server/bundle_manager.py
candicegjing: I guess I was thinking the case that a user has 2 private workers attached to the server. Both of them have their tag set. The 1st worker sets the `tag_exclusive` to be `True`, whereas the 2nd sets the `tag_exclusive` to be `False`. When this user submits a bundle without a `request-queue`, do we want to route this bundle to run on the 2nd worker? If we do, we might want to remove the logic `not worker['tag']`...

nelson-liu: Wouldn't the current logic route the bundle to the second worker? That was the intended role of `not tag_exclusive`

```
In [1]: worker = {"tag_exclusive": False, "tag": "sometag"}

In [2]: not worker['tag_exclusive'] or not worker['tag']
Out[2]: True
```

candicegjing: Yes. That's true. Never mind~ We still want to keep those untagged workers around in case the bundle aren't able to schedule to run on the 2nd private worker. Then it can still be routed to those untagged workers.

candicegjing (approved):
LGTM!

nelson-liu:
> Another thing: It'd be good to write somewhere, perhaps as a comment block or in the documentation, a description of how one typically uses --tag-exclusive.

I stuck this in the execution docs, unsure if it was the right place—feel free to let me know if it's not.

docs/Execution.md
```diff
@@ -135,3 +135,17 @@ Check the bundle's `stdout`, and you should see something similar to before:
...
 
 And that's all.  Congrats!
+
+## Frequently asked questions
+
+### Can I reserve a worker for a particular tag?
+
+Yes, you can use the `--tag-exclusive` flag of `cl worker` (along with a
+`--tag`) to prevent your worker from running untagged bundles, since
+`tag_exclusive` workers only run bundles that match their tag.
```
percyliang: maybe just "since tag-exclusive workers" (plain English)

docs/Execution.md
```diff
@@ -135,3 +135,17 @@ Check the bundle's `stdout`, and you should see something similar to before:
...
+
+To be concrete, a worker started with `cl worker --tag debug --tag-exclusive`
+will _only_ run bundles with `--request-queue tag=debug`. This is especially
```
percyliang: We're trying to deprecate `tag=`, so now you can just do `--request-queue debug` (former is still supported for now).

codalab/rest/workers.py
```diff
@@ -34,6 +34,7 @@ def checkin(worker_id):
         request.json.get("free_disk_bytes"),
         request.json["dependencies"],
         request.json.get("shared_file_system", False),
+        request.json.get("tag_exclusive", False),
```
percyliang, line 37: Hm, if we're going to have `tag_exclusive` after `shared_file_system` in some places, better to be consistent and put it after it everywhere, rather than having a mix.

docs/Execution.md
```diff
@@ -135,3 +135,17 @@ Check the bundle's `stdout`, and you should see something similar to before:
...
+
+Yes, you can use the `--tag-exclusive` flag of `cl worker` (along with a
+`--tag`) to prevent your worker from running untagged bundles, since
+tag-exclusive workers only run bundles that match their tag.
+
+To be concrete, a worker started with `cl worker --tag debug --tag-exclusive`
+will _only_ run bundles with `--request-queue debug`. This is especially
+useful for development, since it lets you set off a machine for fast-turnaround
+debugging of CodaLab bundles, even if you have many untagged jobs queued---just
+make sure to set `--request-queue tag=debug`.
```
percyliang: Remove `tag=`

PR #2118, opened 2020-03-29
Make cl workers show user-owned workers for non-root users

nelson-liu:
I think the cl workers output is _super_ helpful, but it's locked behind root restrictions. This PR makes the command output the individual user's workers, if the user is not root.

I couldn't find a test for this anywhere...maybe I'm missing something?

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1034,7 +1034,8 @@ class BundleCLI(object):
 
     @Commands.command(
         'workers',
-        help=['Display worker information of this CodaLab instance. Root user only.'],
+        help=['Display worker information of this CodaLab instance. '
+              'If not root user, only display information about your connected workers.'],
         arguments=(),
     )
     def do_workers_command(self, args):
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1034,8 +1034,10 @@ class BundleCLI(object):
 
     @Commands.command(
         'workers',
-        help=['Display worker information of this CodaLab instance. '
-              'If not root user, only display information about your connected workers.'],
+        help=[
+            'Display worker information of this CodaLab instance. '
+            'If not root user, only display information about your connected workers.'
+        ],
         arguments=(),
     )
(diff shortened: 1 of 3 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1034,7 +1034,10 @@ def auto_convert_type(value):
 
     @Commands.command(
         'workers',
-        help=['Display worker information of this CodaLab instance. Root user only.'],
+        help=[
+            'Display worker information of this CodaLab instance. '
+            'If not root user, only display information about your connected workers.'
```
percyliang: Change: "Display information about workers that you have connected to the CodaLab instance." (most people won't need to worry about the root user).

codalab/rest/workers.py
```diff
@@ -135,10 +135,10 @@ def start_bundle(worker_id, uuid):
 
 @get("/workers/info", name="workers_info", apply=AuthenticatedPlugin())
 def workers_info():
-    if request.user.user_id != local.model.root_user_id:
-        abort(http.client.UNAUTHORIZED, "User is not root user")
-
     data = local.worker_model.get_workers()
```
percyliang: rename `data` to `workers`

docs/CLI-Reference.md
```diff
@@ -449,7 +449,7 @@ Usage: `cl <command> <arguments>`
 
 ## Commands for managing server:
   ### workers:
-    Display worker information of this CodaLab instance. Root user only.
+    Display worker information of this CodaLab instance. If not root user, only display information about your connected workers.
```
percyliang: Use the description from above, but add: "For the root user, display all workers."

percyliang (changes requested):
Thanks, this is a good idea!  Just a few small cosmetic suggestions.

percyliang (approved):

docs/CLI-Reference.md
nelson-liu: Note that the pre-commit script will overwrite this change every time to match the output of `cl -h -m` (i.e., removing `For the root user, display all workers.`).

PR #2122, opened 2020-03-30
Readthedoc format

candicegjing:
This PR will fix the readthedoc displaying issue with the newly formatted CLI reference doc.

The new readthedoc will look like this:
Screen Shot 2020-03-30 at 2 17 24 PM

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -310,10 +310,12 @@ class Commands(object):
                 )
 
             if verbose:
+                initial_indent = ' ' * indent
                 if markdown:
                     name = HEADING_LEVEL_3 + name
+                    initial_indent = ''
                 return '%s%s:\n%s\n%s' % (
-                    ' ' * indent,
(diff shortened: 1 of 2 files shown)


percyliang (approved):

codalab/lib/bundle_cli.py
```diff
@@ -313,7 +313,8 @@ def render_args(arguments):
                 if markdown:
                     name = HEADING_LEVEL_3 + name
                 return '%s%s:\n%s\n%s' % (
-                    ' ' * indent,
+                    # This is to make github Markdown format compatible with the readthedoc theme.
```
teetone: Capitalize `GitHub` and should it be `Read the Docs`? https://docs.readthedocs.io/en/rel/theme.html

teetone (approved):

codalab/lib/bundle_cli.py
candicegjing: Updated!

PR #2125, opened 2020-04-01
Empty args

### LINKED ISSUE #2114: empty args (`cl run --- echo ""`) broken
`cl run 'echo ""'` works, but `cl run --- echo ""` doesn't.  Should make the latter work.

candicegjing:
Fixed #2114 

The fix in this PR will not preserve the original double quotes `""` as the command from the command line arguments. Instead, it will display single quotes `''` as a substitution as follows:
Screen Shot 2020-04-01 at 9 30 20 AM

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -28,6 +28,7 @@ import textwrap
 from collections import defaultdict
 from contextlib import closing
 from io import BytesIO
+from shlex import quote
 
 import argcomplete
 from argcomplete.completers import FilesCompleter, ChoicesCompleter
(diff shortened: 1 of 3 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -803,7 +804,9 @@ def collapse_bare_command(argv):
         """
         try:
             i = argv.index('---')
-            argv = argv[0:i] + [' '.join(argv[i + 1 :])]  # TODO: quote command properly
+            # Convert each element after '---' to a shell-escaped version of string.
+            argv[i + 1 :] = [quote(x) for x in argv[i + 1 :]]
```
percyliang: Don't mutate `argv`.

tests/lib/bundle_cli_test.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+from codalab.lib.bundle_cli import BundleCLI
+
+
+class BundleCliTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.bundle_cli = BundleCLI
+        self.argv = ['cl', 'run', '---', 'echo', '']
```
percyliang: Why is this part of `self`?  It's likely that we'll have multiple tests with different `argv`s.

codalab/lib/bundle_cli.py
candicegjing: Updated!

percyliang (approved):

PR #2127, opened 2020-04-01
Add new --memo option to cl mimic/run

candicegjing:
This is a draft PR that adds the new `--memo` option for `cl mimic/run`.  `--memo-not-failed` will be added later once the main structure is fixed.

Currently, I added two new parameters in the rest API to fetch bundles from the database:
1.`command`: a string that indicates the command of the current bundle.
2. `dependencies`: a list of uuids that indicates parents of the current bundle. 

`get_memoized_bundles()` function will fetch a list of matched bundles from the database and return it back to the caller. 

Right now, I am still trying to simplify the database query function in `get_memoized_bundles()`, coz it looks quite complicated and redundant. If anyone has any suggestions, please do let me know.


codalab/lib/bundle_util.py
```diff
@@ -261,8 +271,6 @@ def recurse(old_bundle_uuid):
 
             prelude_items = []  # The prelude that we're building up
             for item in worksheet_info['items']:
-                just_added = False
```
percyliang: Why delete these?

codalab/lib/bundle_cli.py
```diff
@@ -790,6 +790,7 @@ def get_provided_metadata(args):
             help='Operate on this worksheet (%s).' % WORKSHEET_SPEC_FORMAT,
             completer=WorksheetsCompleter,
         ),
+        Commands.Argument('-m', '--memo', help='Memoized runs', action='store_true'),
```
percyliang: Rename `--memoize` and have more descriptive help message.

codalab/lib/bundle_cli.py
```diff
@@ -1561,14 +1563,26 @@ def do_run_command(self, args):
...
-        self.wait(client, args, new_bundle['uuid'])
+        if args.memo:
+            memoized_bundles = client.fetch(
+                'bundles',
+                params={
+                    'command': args.command,
+                    'dependencies': [uuid for uuid, target in targets],
+                },
+            )
+            if len(memoized_bundles) > 0:
```
percyliang: If there are multiple ones, we should make sure to take the latest one.

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,58 @@ def get_bundle_uuids(self, conditions, max_results):
 
         return self._execute_query(query)
 
+    def get_memoized_bundles(self, user_id, command, dependencies):
+        # Select bundles that has the given command from the bundle table
```
percyliang: Add general docstring for method

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,58 @@ def get_bundle_uuids(self, conditions, max_results):
...
+        filter_on_command = (
+            select([cl_bundle.c.uuid])
+            .select_from(cl_bundle)
+            .where(and_(cl_bundle.c.command == command, cl_bundle.c.owner_id == user_id))
+            .alias("filter_on_command")
+        )
+        # Get a list of uuids that matches with the given dependencies from the cl_bundle table.
+        # This step is to standardize uuid for future comparison.
+        dep_clause = [
+            cl_bundle_dependency.c.parent_uuid.like('%' + uuid + '%') for uuid in dependencies
```
percyliang: Why is this not exact equality?

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,58 @@ def get_bundle_uuids(self, conditions, max_results):
...
+        build_dependency_query = (
+            select([cl_bundle.c.uuid]).select_from(cl_bundle).where(or_(*dep_clause))
+        )
+        dependency_full_list = self._execute_query(build_dependency_query)
+        # Ensure the concatenated bundles' uuid in ascending order
+        dependency_full_list.sort()
+        concat_dependencies = ','.join(dependency_full_list)
+        # Join between filter_on_command table and cl_bundle_dependency table
+        # to get bundles that match with both input command and dependencies
+        join = (
```
percyliang: Are we considering the key for each dependency as well?  `cl run a:foo b:bar cat a b` is different from `cl run a:bar b:foo cat a b`

candicegjing: Yes. That's a good catch! Currently, we don't have logic to handle this case. I'll add it.

codalab/model/bundle_model.py
candicegjing: The `uuid` passed in from the command line might be a partial string, e.g. 0x23fc3f. Then a string pattern matching operation could help us to find the exact bundle.

percyliang: When does this happen?  `uuid` should really always mean the full uuid.

candicegjing: This could happen in cases like this `cl run :0x23456 ...`. Then the `uuid` that gets passed into the function will be a partial one.

codalab/lib/bundle_util.py
candicegjing: `just_added` hasn't been used in any condition check. Looks redundant to me.

codalab/model/bundle_model.py
percyliang: I think in that case, the partial uuid should get resolved to a full uuid?

candicegjing: Yes, that's what we are trying to do here. Basically, I was trying to resolve the partial uuid to a full uuid so that we can do a comparison later on. But I open to discuss how we want to get this part resolved. 
Currently, I did two database queries in the `get_memoized_bundles` function. 
1. The first query is to resolve the partial uuid and build a standardized dependency string to be matched later on.
1. The second query is to do the real search on matched bundles. 

The 1st query can be saved by writing a customized function that does string matching between lists, e.g. comparing the input dependencies list with the dependency list constructed from the database query.
There might be some performance difference but it will depend on how many matched records we are getting back from database and how many existing records in the database. Please do correct me if I am wrong.

codalab/model/bundle_model.py
candicegjing: To resolve this use case, the solution I have is as follows:
1. pass `[a:uuid_a, b:uuid_b]` as a dependency list to `get_memoized_bundles` function.
1. resolve partial uuid in `uuid_a` and `uuid_b` and build a dependency string in ascending order.
1. query database between the `bundle` table and the `bundle_dependency` table based on **input command** and **the dependency list**, and build a dependency string in the same format as 2. Then select the uuids that have the same dependency string. 
1. Return the last uuid which indicates the most recent record from the result list. 

I think the current logic doesn't look really nice. I am still trying to come up with a better design in hoping to simplify this. But let me know if you have any suggestions.

codalab/lib/bundle_util.py
```diff
@@ -282,19 +295,20 @@ def recurse(old_bundle_uuid):
...
-                                data={
-                                    'type': worksheet_util.TYPE_BUNDLE,
-                                    'worksheet': JsonApiRelationship('worksheets', worksheet_uuid),
-                                    'bundle': JsonApiRelationship('bundles', new_bundle_uuid),
-                                },
-                                params={'uuid': worksheet_uuid},
-                            )
-                            new_bundle_uuids_added.add(new_bundle_uuid)
-                            just_added = True
+                            if new_bundle_uuid not in new_bundle_uuids_added:
```
candicegjing: This seems to be a bug in the original mimic code. I am seeing two bundles attached to the current worksheet when I was running `cl mimic <uuid>`. So I added logic here to check if the bundle has already been added.

codalab/model/bundle_model.py
percyliang: I thought the canonicalization happens earlier.  Are you sure at this point `uuid`s aren't just full uuids?

candicegjing: Yes, you are right. I have updated this part with the object that used to resolve the partial uuid issue.

codalab/lib/bundle_cli.py
candicegjing: Yes. The results are ordered by id, and we are going to return the last one if there is any.

codalab/lib/bundle_cli.py
```diff
@@ -790,6 +790,12 @@ def get_provided_metadata(args):
             help='Operate on this worksheet (%s).' % WORKSHEET_SPEC_FORMAT,
             completer=WorksheetsCompleter,
         ),
+        Commands.Argument(
+            '-m',
+            '--memoize',
+            help='Use a memoized bundle that matches with the given command and dependencies',
```
percyliang: "If a bundle with the same command and dependencies already exists, return it instead of creating a new one."

codalab/lib/bundle_cli.py
```diff
@@ -1547,6 +1553,12 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
             Commands.Argument(  # Internal for web FE positioned insert.
                 '-a', '--after_sort_key', help='Insert after this sort_key', completer=NullCompleter
             ),
+            Commands.Argument(
+                '-m',
+                '--memoize',
+                help='Use a memoized bundle that matches with the given command and dependencies',
```
percyliang: Same as above

codalab/lib/bundle_cli.py
```diff
@@ -1559,16 +1571,26 @@ def do_run_command(self, args):
...
-        new_bundle = client.create(
-            'bundles',
-            self.derive_bundle(RunBundle.BUNDLE_TYPE, args.command, targets, metadata),
-            params=params,
-        )
-
-        print(new_bundle['uuid'], file=self.stdout)
-        self.wait(client, args, new_bundle['uuid'])
+        if args.memoize:
+            dependencies = [key + ':' + bundle_target.bundle_uuid for key, bundle_target in targets]
```
percyliang: Is there a standard function we call on the key and uuid rather than having raw string concatenation?

codalab/lib/bundle_cli.py
```diff
@@ -1559,16 +1571,26 @@ def do_run_command(self, args):
...
+            memoized_bundles = client.fetch(
+                'bundles', params={'command': args.command, 'dependencies': dependencies}
+            )
+            if len(memoized_bundles) > 0:
+                print(memoized_bundles[-1]['uuid'], file=self.stdout)
+            else:
+                print("Cannot find any memoized bundle.", file=self.stdout)
```
percyliang: If you can't find it, create a new bundle as before.

codalab/lib/bundle_util.py
```diff
@@ -190,8 +190,23 @@ def recurse(old_bundle_uuid):
             new_info['metadata'] = new_metadata
             new_info['dependencies'] = new_dependencies
 
+            # Fetch the memoized bundle if the memoize option is set to be True
+            memoized_bundles = None
```
percyliang: Don't need this

codalab/lib/bundle_util.py
```diff
@@ -190,8 +190,23 @@ def recurse(old_bundle_uuid):
...
+            if memoize:
+                memoized_bundles = client.fetch(
+                    'bundles',
+                    params={
+                        'command': old_info['command'],
+                        'dependencies': [
+                            dep['child_path'] + ':' + dep['parent_uuid'] for dep in new_dependencies
```
percyliang: as before, avoid raw string concatenation with `:`, which is error prone

codalab/lib/bundle_util.py
percyliang: There sometimes could be multiple bundles on the same worksheet.  Can you provide the exact example / behavior?  In any case, I think we should make this change in a different PR.

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,85 @@ def get_bundle_uuids(self, conditions, max_results):
 
         return self._execute_query(query)
 
+    def get_memoized_bundles(self, user_id, command, dependencies):
+        '''
+        Get the most recently finished bundle that matches with input command and dependencies.
```
percyliang: Should be most recently created bundle (since we're sorting by id).

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,85 @@ def get_bundle_uuids(self, conditions, max_results):
...
+                    .select_from(cl_bundle_dependency)
+                    .where(cl_bundle_dependency.c.child_uuid.in_(uuids))
+                )
+                dependency_rows = connection.execute(query).fetchall()
+                dependency_uuids = set([row[0] for row in dependency_rows])
+                return [uuid for uuid in uuids if uuid not in dependency_uuids]
+
+            dep_clause = []
+            for dep in dependencies:
+                key, uuid = dep.split(':')
```
percyliang: bad to have raw manipulation, should use standard functions

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,85 @@ def get_bundle_uuids(self, conditions, max_results):
...
+        :param user_id: a string that specifies the current user id.
+        :param command: a string that defines the command that is used to search for memoized bundles in the database.
+        :param dependencies: a list of string in the structure of ["key1:uuid1","key2:uuid2"] defines
+                            dependencies that are used to search for memoized bundles in the database.
+        :return: a list of matched uuids.
+        '''
+        # Get a list of uuids that matches with the given dependencies from the cl_bundle table.
+        # This step is to standardize uuid for future comparison.
+        with self.engine.begin() as connection:
+            # Select bundles that has the given command from the bundle table.
```
percyliang: You should ideally do a single DB query that has the command + dependencies for efficiency.

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,85 @@ def get_bundle_uuids(self, conditions, max_results):
...
+                .select_from(cl_bundle)
+                .where(and_(cl_bundle.c.command == command, cl_bundle.c.owner_id == user_id))
+                # Ensure the order of the returning bundles will be in the order of they were created.
+                .order_by(cl_bundle.c.id)
+            )
+            rows = connection.execute(filter_on_command).fetchall()
+            if not rows:
+                return []
+            uuids = [row[0] for row in rows]
+            # When there is no dependency to be matched, the target memozied bundle
```
percyliang: Is there a way to simplify the logic below?

percyliang (changes requested):

codalab/lib/bundle_cli.py
candicegjing: I don't think we currently have a function that calls on the **key and uuid** together. Do you mean to pass the key and uuid **separately** as two parameters to function call?

codalab/lib/bundle_util.py
candicegjing: Sure. I'll create a separate PR. The behavior I saw is that when I ran `cl mimic <uuid>` on a worksheet, it created two bundles with the same uuid/content in that worksheet. I thought it should only create 1 bundle. But let me know if I misunderstood its functionality here.

codalab/model/bundle_model.py
candicegjing: I have been trying to get rid of this manipulation coz I do think it's not a good practice here. But the intention behind this is primarily trying to be able to search for a list of `(key, uuid)` pair that can match with the input command + dependencies in the database. If we take a look in the schema we have, `cl_bundle_dependency` has the following columns:
|index| column | 
|-- |-- |
| 1. | child_uuid | 
|2. | child_path |
|3.| parent_uuid|
|4.| parent_path|

`child_uuid` is the target bundle that we will be searching for. `key` will map to `child_path` and `uuid` will map to `parent_uuid`. Given that the relationship between `child_uuid` and `<child_path (key), parent_uuid (uuid)>` is `1-to-n` mapping, when we group by `child_uuid`, we need some ways to keep the records in the other two columns together. I have been searching for solutions to keep those records without doing any concatenation, but so far I haven't found anything yet. If you guys @teetone or @percyliang have any suggestions, please let me know. 
However, I think we can definitely avoid passing the raw string manipulation from CLI, add additional steps parsing the returning results from the database and then identify the right bundle to return.

codalab/model/bundle_model.py
candicegjing: My intention of `filter_on_command` first is to reduce the size of the records that will be searching on before the later join/selection operation. But it might not be significant in improving efficiency comparing to executing with one query all in once.

codalab/model/bundle_model.py
candicegjing: Did more tests and currently don't think we can combine those two conditions (with dependencies and without dependencies) into one together due to the following constraints:

1. When there is no dependency specified, we want to search for the bundles that ONLY exist in the `bundle` table, but NOT in the `bundle_dependency` table.
1. When there are dependencies specified, we want to do an **inner join** between the `bundle` table and the `bundle_dependency` table, so that we will be able to select those bundles that match with input command as well as input dependencies.

Essentially the above two conditions are different when reflected on SQL query, but I may not see the optimal solution, let me know if I miss anything here.

candicegjing: @percyliang Based on our discussion yesterday, I altered the database query logic to use join query on all dependencies. Changes are listed at https://github.com/codalab/codalab-worksheets/pull/2148. 
For the use case when there is no dependency provided, if we do want to include functionality to handle this, I don't think the logic can be combined with the one with dependencies. On the other hand, if we really don't want to handle this case, then I'll remove this piece of logic. 
I personally think providing the ability to return a memoized bundle without dependencies would be helpful especially in the case when the bundle might take a long time to finish. Suppose one user submitted a bundle that took a few hours to finish, after 1 month or so, he/she wanted to run it again, then they could really take advantage of using the memozie option. Let me know how you think!

candicegjing: At the same time, I am benchmarking those queries in staging using a production sized database just to make sure the performance of queries is good.

candicegjing: @percyliang I later realized that [union on joins](#2148) could be avoided and use 1 JOIN query a time. The reason is that the list of dependencies that passed into `get_memoized_bundle` is already deduped. As all dependencies are unique now, when we do GROUP BY `child_uuid`, we just need to aggregate the number of records per `child_uuid` and then compare it with the number of dependencies. If the numbers match, then we find a memoized bundle.

codalab/model/bundle_model.py
teetone: I think I understand the situation. When the `/bundles` endpoint is called, can we have the contract to be that a JSON list of `child_path` -`parent_uuid`  pairs has to be passed as the dependencies? Something like this:
```
x = '[{ "child_path":"some child", "parent_uuid": "0xblah"}, 
      { "child_path":"some child2", "parent_uuid": "0xblah2"}]'
```
Then, we can parse the JSON on the server-side using `json.loads`, so `json.loads(x)` should give you the list of dependencies you need. The other alternative (but I think less cleaner) solution would be to just pass in two separate lists.

codalab/rest/bundles.py
```diff
@@ -105,6 +105,8 @@ def _fetch_bundles():
     specs = query_get_list('specs')
     worksheet_uuid = request.query.get('worksheet')
     descendant_depth = query_get_type(int, 'depth', None)
+    command = query_get_type(str, 'command', '')
```
teetone, line 119: We should document the following parameters too.

codalab/model/bundle_model.py
candicegjing: I think this is a great idea! Let me change this.

codalab/rest/bundles.py
candicegjing: Done

PR #2129, opened 2020-04-02
Update bundle detail when detail was opened in created state

### LINKED ISSUE #2033: bundle details not refreshing for run bundles
Create a new run bundle.  Hit 'enter' to show bundle details.  If 'enter' is hit before the bundle status changes to staged, then the bundle status in the bundle details panel that shows up doesn't update.  If 'enter' is hit after the bundle status changes to staged, then it will update reliably.

yipenghe:
Fix #2033 

To test: go to bundleRow, always open set showDetail to be true all the time. Before the fix, the state in bundle detail will be stuck at "created". After the fix, it will be updated reliably


yipenghe:
@epicfaace Can you review?

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -63,9 +63,11 @@ class BundleDetail extends React.Component<
                 this.fetchBundleContents();
                 return;
             }
-            if (this.state.bundleInfo.bundle_type === 'run' && (this.state.bundleInfo.state === 'running' 
-                || this.state.bundleInfo.state === 'preparing' || this.state.bundleInfo.state === 'starting'
-                || this.state.bundleInfo.state === 'staged')){
+            if (this.state.bundleInfo.bundle_type === 'run' && (
+                    this.state.bundleInfo.state === 'created'
```
epicfaace: I think we should check for more bundle states -- in fact, use a regex `(uploading|created|staged|making|starting|preparing|running)` (comes from https://github.com/codalab/codalab-worksheets/blob/6e0d3e9faff013db8d77125329649dfec02511df/scripts/create_sample_worksheet.py#L38), in addition to `finalizing`

epicfaace (changes requested):

epicfaace (approved):

PR #2130, opened 2020-04-02
Add button to expand/shrink worksheet width

### LINKED ISSUE #1726: [New Frontend] Horizontal scrollbar doesn't reach the editor boundary
**Codalab Information**
- Browser: Chrome

**To Reproduce**
1. Make the browser window large.
2. Edit the source of a Worksheet.
3. See the scrollbar issue.

**Expected behavior**
Scrollbar should extend to the edge of the editor.

**Screenshots**
scroll_bar_edit_source

### LINKED ISSUE #2081: [USER ISSUE] Results table width is too narrow
**Describe the bug**
Results table takes only 1/3 on a wide screen.

image

**Codalab Information**
- *Your username*: ilylin
- *UUID of worksheet where problem occurred*: 0x922b658ce0c04ea98d149a5371afed98/
 - *OS (if on CLI)*:
 - *Browser (if on Web)*;

**To Reproduce**

**Expected behavior**
The results table should take more horizontal space

**Screenshots**

**Additional context**

yipenghe:
fix #2081 fix #1726 
Default to use 50% of the window size, and allow to expand to 80% of the window size. Utilizes local storage to remember the preference of the expand size, so user don't need to click on expand everytime they open a new tab. Using '+' for keyboard shortcut
Screenshot from 2020-04-02 03-09-34
Screenshot from 2020-04-02 03-09-39


percyliang:
Great.  Seems like the expand should essentially take up close to 100%.  For those wide tables, you really shouldn't waste any space.  Is there a way to pass the expansion mode in the URL too?

frontend/src/constants.js
```diff
@@ -2,3 +2,7 @@
 export const CODALAB_VERSION = '0.5.11';
 
 export const NAVBAR_HEIGHT = 60;
+
+// Worksheet width
+export const EXPAND_PERCENTAGE = '90%';
```
epicfaace: Can you rename these to `EXPANDED_WORKSHEET_WIDTH` and `DEFAULT_WORKSHEET_WIDTH`?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -581,6 +589,19 @@ class Worksheet extends React.Component {
     toggleGlossaryModal = () => {
         this.setState({ showGlossaryModal: !this.state.showGlossaryModal });
     };
+    toggleWorksheetSize = () => {
+        let newPercentage =
+            this.state.worksheetWidthPercentage === DEFAULT_PERCENTAGE
+                ? EXPAND_PERCENTAGE
+                : DEFAULT_PERCENTAGE;
+        window.localStorage.setItem('worksheetSize', newPercentage);
```
epicfaace: Can you refactor the localStorage worksheet size item name into a constant?

frontend/src/CodalabApp.js
```diff
@@ -101,7 +101,7 @@ function CodalabApp() {
                                 render={(props) => <Login {...props} auth={fakeAuth} />}
                             />
                             <PrivateRoute path='/account/profile' component={UserInfo} />
-                            <Route path='/worksheets/:uuid' component={Worksheet} />
+                            <Route path='/worksheets/:uuid/:mode?' component={Worksheet} />
```
epicfaace: I think it would be best to remove the URL change when expanding / shrinking, given that this is a global setting that is set in localStorage.

epicfaace (changes requested):
When I click the "expand" button, it takes a second or so before the worksheet actually expands. We should disable the button so that users do not accidentally click the button twice.

Same goes for the "+" keyboard shortcut -- otherwise it's easy for the user to accidentally expand twice and then go back to the normal size.

yipenghe:
> When I click the "expand" button, it takes a second or so before the worksheet actually expands. We should disable the button so that users do not accidentally click the button twice.
> 
> Same goes for the "+" keyboard shortcut -- otherwise it's easy for the user to accidentally expand twice and then go back to the normal size.

I think this is related to how large the worksheet is, for small worksheets it seems to pretty fast. I'm not sure how to disable only when the worksheet is trying to rerender the main area, I tried to add in the onclick event to set disabling value, but the batch state update will cause the button be disabled only along with the expansion update. I feel like it's OK since clicking twice doesn't do harm

epicfaace:
> I feel like it's OK since clicking twice doesn't do harm

I also got a faster computer so this isn't really an issue for me anymore haha

frontend/src/CodalabApp.js
epicfaace:
```suggestion
                            <Route path='/worksheets/:uuid' component={Worksheet} />
```

epicfaace (approved):

PR #2131, opened 2020-04-02
Fix mathjax loading race condition

### LINKED ISSUE #1851: MathJax seems to be not working consistently
Before  deploying the new frontend, the following worksheet with math symbols used to render properly. With the new frontend, it seems to be not working consistently.

https://worksheets.codalab.org/worksheets/0xd19352d4d7854ad484e4e04609c1671b

epicfaace:
fixes #1851

Previously, there was a race condition in which the mathjax script's onload function would not fire -- thus, the math would not be formatted in some cases.

This change instead creates the script and then adds the onload function before appending it to the DOM.

I tested this by refreshing the sample worksheet ~20 times and saw no issues with mathjax (while previously, doing so without the change would sometimes show non-formatted issues).


yipenghe (approved):
Probably would be good to comment what we are doing here

PR #2138, opened 2020-04-06
Prioritize bundle scheduling with request_priority

nelson-liu:
This PR changes the scheduling logic to use the bundle's `request_priority` metadata.

Before, bundles that were requested to run on specific machines (i.e., they provided a value for `request_queue`) were scheduled to run before all other bundles.

After this PR:

(1) all bundle sorting is done within-user. I made this change for the `request_queue` case as well because, if we tag public instances (https://github.com/codalab/codalab-worksheets/pull/2093), a user could jump the public queue by just specifying the right `request_queue`.
(2) The bundle ordering is determined first by the bundle's priority—higher priority bundles get scheduled first, regardless if they have a `request_queue` or not. Then, the normal logic applies (bundles with `request_queue` prioritized above those that don't).

Some thoughts:

I feel like the default value for `request_priority`, `None`, is a bit funky. Should I change this? There might be some backward-compatibility issues?

I'm also not entirely sure I'm happy with the way that the current code handles 0 priority and negative priority (currently, 0 priority or negative priority bundles have precedence over bundles with no priority specified).

Ideally, I'd like `None` priority to just be `0` priority, but I'm not sure what compatibility things I need to be worried about when making this change. Thoughts?

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -323,12 +323,34 @@ class BundleManager(object):
         :param staged_bundles_to_run: a list of tuples each contains a valid bundle and its bundle resources.
         :param user_info_cache: a dictionary mapping user id to user information.
         """
-        # Reorder the stage_bundles so that bundles which were requested to run on a specific worker
-        # will be scheduled to run first
-        staged_bundles_to_run.sort(
-            key=lambda b: (b[0].metadata.request_queue is not None, b[0].metadata.request_queue),
-            reverse=True,
-        )
+        # Build a dictionary which maps from user id to positions in the queue of the
+        # user's staged bundles.
+        user_queue_positions = {}
+        for queue_position, staged_bundle in enumerate(staged_bundles_to_run):
+            if staged_bundle[0].owner_id not in user_queue_positions:
+                user_queue_positions[staged_bundle[0].owner_id] = []
+            user_queue_positions[staged_bundle[0].owner_id].append(queue_position)
+
+        for user, queue_positions in user_queue_positions.items():
+            sorted_queue_positions = sorted(queue_positions)
+            # Get this user's staged bundles
+            user_staged_bundles = [
+                staged_bundles_to_run[queue_position] for queue_position in queue_positions
+            ]
+            # Sort the staged bundles for this user, according to
+            # (1) their priority and (2) whether it requested to run on a specific worker.
+            sorted_user_staged_bundles = sorted(
+                user_staged_bundles,
+                key=lambda b: (
+                    b[0].metadata.request_priority is not None,
+                    b[0].metadata.request_priority,
+                    b[0].metadata.request_queue is not None,
+                    b[0].metadata.request_queue,
+                ),
+                reverse=True,
+            )
+            for queue_position, bundle in zip(sorted_queue_positions, sorted_user_staged_bundles):
+                staged_bundles_to_run[queue_position] = bundle
 
         # Build a dictionary which maps from uuid to running bundle and bundle_resources
         running_bundles_info = self._get_running_bundles_info(workers, staged_bundles_to_run)
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -327,7 +327,13 @@ class BundleManager(object):
         :param user_info_cache: a dictionary mapping user id to user information.
         """
         # Build a dictionary which maps from user id to positions in the queue of the
-        # user's staged bundles.
+        # user's staged bundles. We use this to sort bundles within each user. For example,
+        # Suppose we have 4 staged bundles with the following attributes from 2 users:
+        # Users: [A, B, A, B, A]
+        # Bundle Priorities: [1, 2, 3, 1, 1]
+        # Bundle specified request_queue: [False, False, False, False, True]
+        # Original Bundle Order: [B1, B2, B3, B4, B5]
+        # Sorted bundle order: [B3, B2, B5, B4, B1]
         user_queue_positions = defaultdict(list)
         for queue_position, staged_bundle in enumerate(staged_bundles_to_run):
             user_queue_positions[staged_bundle[0].owner_id].append(queue_position)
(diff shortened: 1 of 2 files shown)


codalab/server/bundle_manager.py
```diff
@@ -323,12 +323,34 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
+        # Build a dictionary which maps from user id to positions in the queue of the
+        # user's staged bundles.
+        user_queue_positions = {}
+        for queue_position, staged_bundle in enumerate(staged_bundles_to_run):
+            if staged_bundle[0].owner_id not in user_queue_positions:
+                user_queue_positions[staged_bundle[0].owner_id] = []
+            user_queue_positions[staged_bundle[0].owner_id].append(queue_position)
+
+        for user, queue_positions in user_queue_positions.items():
+            sorted_queue_positions = sorted(queue_positions)
```
nelson-liu: `queue_positions` should already be sorted here, but I added this just to be defensive...

nelson-liu:
Also, should I / where should I add a test for this?

codalab/server/bundle_manager.py
```diff
@@ -314,21 +314,45 @@ def _bring_offline_stuck_running_bundles(self, workers):
     def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_info_cache):
         """
         Schedule STAGED bundles to run on available workers based on the following logic:
-        1. If the bundle requests to run on a specific worker, tries to schedule the bundle
-           to run on a worker that has a tag exactly match with request_queue.
+        1. For a given user, schedule the highest-priority bundles first, followed by bundles
+           that request to run on a specific worker.
+        2. If the bundle requests to run on a specific worker, schedule the bundle
```
percyliang, line 320: Numbering needs to be updated to 1, 2, 3.?

codalab/server/bundle_manager.py
```diff
@@ -314,21 +314,45 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
-        # will be scheduled to run first
-        staged_bundles_to_run.sort(
-            key=lambda b: (b[0].metadata.request_queue is not None, b[0].metadata.request_queue),
-            reverse=True,
-        )
+        # Build a dictionary which maps from user id to positions in the queue of the
+        # user's staged bundles.
+        user_queue_positions = {}
+        for queue_position, staged_bundle in enumerate(staged_bundles_to_run):
+            if staged_bundle[0].owner_id not in user_queue_positions:
```
percyliang: use `defaultdict(list)` so you don't need this extra `if`

codalab/server/bundle_manager.py
percyliang: Maybe change this to an `assert` (I guess that's being offensive)...

codalab/server/bundle_manager.py
```diff
@@ -314,21 +314,45 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
         :param staged_bundles_to_run: a list of tuples each contains a valid bundle and its bundle resources.
         :param user_info_cache: a dictionary mapping user id to user information.
         """
-        # Reorder the stage_bundles so that bundles which were requested to run on a specific worker
-        # will be scheduled to run first
-        staged_bundles_to_run.sort(
-            key=lambda b: (b[0].metadata.request_queue is not None, b[0].metadata.request_queue),
-            reverse=True,
-        )
+        # Build a dictionary which maps from user id to positions in the queue of the
```
percyliang, line 329: Would be helpful to document with a concrete example:
Users: A B A B
Priorities: 1 2 3 1
Original bundles order: B1 B2 B3 B4
New bundles order: B1 B4 B3 B2
or something to that effect

codalab/server/bundle_manager.py
```diff
@@ -314,21 +314,45 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
+            ]
+            # Sort the staged bundles for this user, according to
+            # (1) their priority and (2) whether it requested to run on a specific worker.
+            sorted_user_staged_bundles = sorted(
+                user_staged_bundles,
+                key=lambda b: (
+                    b[0].metadata.request_priority is not None,
+                    b[0].metadata.request_priority,
+                    b[0].metadata.request_queue is not None,
+                    b[0].metadata.request_queue,
```
percyliang, line 357: I know it was there before, but a bit weird that we're sorting by the `request_queue`...

codalab/server/bundle_manager.py
```diff
@@ -314,21 +314,45 @@ def _bring_offline_stuck_running_bundles(self, workers):
...
+            user_queue_positions[staged_bundle[0].owner_id].append(queue_position)
+
+        for user, queue_positions in user_queue_positions.items():
+            sorted_queue_positions = sorted(queue_positions)
+            # Get this user's staged bundles
+            user_staged_bundles = [
+                staged_bundles_to_run[queue_position] for queue_position in queue_positions
+            ]
+            # Sort the staged bundles for this user, according to
+            # (1) their priority and (2) whether it requested to run on a specific worker.
```
percyliang: Make it clearer which is better (larger is higher priority, specific worker gets higher priority).

percyliang (changes requested):
Looks great, thanks!  Just a few minor comments.

codalab/server/bundle_manager.py
nelson-liu: I think it's actually a great decision—if I have 2 jobs staged, and one has `request_queue` and the other doesn't, I want to run the job that's more constrained first, if possible?

candicegjing (approved):
Looks great to me! I think this is going to be very helpful for users to plan on arranging their jobs before/after submitting them.

candicegjing:
I just wonder if you guys @nelson-liu and @percyliang think it would be helpful to extract this functionality to be a single function? I guess it would be helpful for the following two cases:
1. We can have it easily tested
1. If in the future we want to add more flexibility to prioritize jobs, they can all go there and not touch other parts.

It doesn't have to be done in this PR but let me know.

percyliang:
The bundle manager really could use more tests in general.  Perhaps we should just get this in and then have a separate issue to figure out how to structure those tests.

percyliang (approved):
LGTM!

codalab/server/bundle_manager.py
percyliang: yes, the `request_queue is not None` makes sense, but I was talking about `request_queue`: sorting `bar` before `foo` seems arbitrary...

nelson-liu: Ah I see. I can open another PR after this one goes in to change that.

PR #2139, opened 2020-04-07
Don't schedule bundles by request_queue name

nelson-liu:
ref: https://github.com/codalab/codalab-worksheets/pull/2138#discussion_r404993533


percyliang (approved):
Thanks!

PR #2143, opened 2020-04-09
Cut bundle rows from worksheet

### LINKED ISSUE #2017: Cut (copy and remove source lines) bundle rows feature

yipenghe:
fix #2017  
Similar to copy but also removes the source line (if the corresponding source line is not a directive (meaning search/wsearch)


epicfaace (commented):
Should we have similar behavior to Excel? (in which when you click "Cut", the actual cell isn't deleted yet, but just has a dashed border -- and only when you click "Paste" will the original cell be deleted)

The advantage I see of this approach is that cutting / copying / pasting in Codalab is kind of slow so it might be better to not create a way for the user to accidentally cut a cell, and then it's deleted because they closed the page or something.

percyliang:
Can we make cut/copy/paste faster?  What's the bottleneck?  Probably we have to re-render the entire worksheet, but we can make that not the case (with some work).  Given that this is the primary way that we intend users to move bundles around, it's pretty important that these operations be fast.

yipenghe:
> Should we have similar behavior to Excel? (in which when you click "Cut", the actual cell isn't deleted yet, but just has a dashed border -- and only when you click "Paste" will the original cell be deleted)
> 
> The advantage I see of this approach is that cutting / copying / pasting in Codalab is kind of slow so it might be better to not create a way for the user to accidentally cut a cell, and then it's deleted because they closed the page or something.

I'm not sure that's feasible in our case, since you can hit paste on a different worksheet in a different tab or a different browser session

yipenghe:
> Can we make cut/copy/paste faster? What's the bottleneck? Probably we have to re-render the entire worksheet, but we can make that not the case (with some work). Given that this is the primary way that we intend users to move bundles around, it's pretty important that these operations be fast.

Copy/cut/paste should be no slower than editing the source directly, and now we don't require confirmation dialogs anymore, so it should be much convenient than using the markdown and I'm not sure how much it takes to make it faster than it currently is

yipenghe:
> Should we have similar behavior to Excel? (in which when you click "Cut", the actual cell isn't deleted yet, but just has a dashed border -- and only when you click "Paste" will the original cell be deleted)
> 
> The advantage I see of this approach is that cutting / copying / pasting in Codalab is kind of slow so it might be better to not create a way for the user to accidentally cut a cell, and then it's deleted because they closed the page or something.

I think we might instead add a confirmation dialog for the cut operation, showing the uuid[0:8], names etc, just like a delete operation to confirm they want to remove the lines. Since deleting the source line doesn't delete the actual bundles, they should still be able to recover them quickly from their dashboard by finding them in the floating search areas

frontend/src/components/worksheets/WorksheetDialogs.js
```diff
@@ -170,10 +169,6 @@ const styles = () => ({
         color: 'red',
         marginBottom: 20,
     },
-    copyDialog: {
```
yipenghe, line 173: not used anymore

codalab/rest/interpret.py
```diff
@@ -693,7 +693,7 @@ def resolve_items_into_infos(items):
         )
         if bundle_info is not None:
             bundle_info['sort_key'] = i['sort_key']
-            bundle_info['id'] = i['id']
+            bundle_info = dict(bundle_info, id = i['id'])
```
yipenghe, line 696: similar issue to sort keys

yipenghe, line 696: similar issue to sort keys

codalab/rest/interpret.py
```diff
@@ -693,7 +693,7 @@ def resolve_items_into_infos(items):
         )
         if bundle_info is not None:
             bundle_info['sort_key'] = i['sort_key']
-            bundle_info['id'] = i['id']
+            bundle_info = dict(bundle_info, id=i['id'])
```
yipenghe: similar issue to sort keys

epicfaace: Should we remove the earlier line and instead do:

```suggestion
            bundle_info = dict(bundle_info, id=i['id'], sort_key=i['sort_key'])
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -429,18 +431,45 @@ class Worksheet extends React.Component {
                     draggable: true,
                 });
             });
+            if (cmd_type === 'cut') {
+                // Remove the bundle lines
+                this.removeItemsFromSource(cutBundleIds);
```
epicfaace: I think it's a bit confusing that we create and populate cutBundleIds even if cmd_type is not equal to "cut". How about if we only populate validBundles, and then change this line to:

```suggestion
                this.removeItemsFromSource(validBundles.map(e => e.id));
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -429,18 +431,45 @@ class Worksheet extends React.Component {
...
+            }
         } else if (cmd_type === 'paste') {
             this.pasteBundlesToWorksheet();
         }
     };
 
+    removeItemsFromSource = (itemIds) => {
+        let worksheetUUID = this.state.ws.uuid;
+        // let after_sort_key;
```
epicfaace: remove this line

epicfaace (commented):
Looks good -- just a few questions.

Also, when I cut a bundle, the message says that I've "copied one bundle" -- perhaps it would be clearer to say "cut the bundle" instead.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -944,16 +973,42 @@ class Worksheet extends React.Component {
                 'keyup',
             );
         }
+        // paste after current focused cell
+        if (this.state.ws.info.edit_permission) {
```
epicfaace: Why did we need to add this code?

yipenghe:
> Looks good -- just a few questions.
> 
> Also, when I cut a bundle, the message says that I've "copied one bundle" -- perhaps it would be clearer to say "cut the bundle" instead.

Addressed

epicfaace (approved):

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -434,13 +440,35 @@ class Worksheet extends React.Component {
...
+    };
+
     confirmBundleRowAction = (code) => {
         if (
             !(
                 this.state.openDelete ||
                 this.state.openDetach ||
                 this.state.openKill ||
                 this.state.openCopy ||
+                this.state.openCut ||
```
epicfaace: What about these? I don't think openCut is ever set?

yipenghe: Ah good catch, forgot that we don't have cut dialogs anymore

PR #2145, opened 2020-04-09
Making info properties clear

### LINKED ISSUE #2144: make worksheet_info properties clearer
the names are confusing because "raw" refers to different things in different properties:
raw: refers to source lines in editor
raw_to_block, block_to_raw: "raw" refers to raw items (search directive lines have been expanded already)
Also the documentation for these mappings are lacking, which causes great confusion for people working on them

yipenghe:
Fix #2144 

The main goal of this PR is make several properties of worksheet_info more clearer to what they mean
- raw => source : this makes it consistent to what we call it in the frontend as in "edit source"

- raw_to_block, block_to_raw : since we change raw=>source, raw in these variables here will solely mean raw items that were interpreted from source lines

- documentation: I added quite a bit of documentation to what raw_to_block, block_to_raw mean, mainly because when I was working on the frontend, these mappings were very confusing, and were used to address several focus setting issues

- add raw_items_to_source_index: this is used for copy/paste/cut and will be used be schema items, adding it in this PR so later PRs can all use it 

- items: items should really be named "block_items" to avoid confusion, the only thing is I need to make sure ["items"] in other places of the codebase doesn't get affected


codalab/rest/interpret.py
```diff
@@ -563,10 +599,6 @@ def resolve_items_into_infos(items):
     return new_items
 
 
-def expand_raw_items(raw_items):
```
yipenghe: I removed this function because it doesn't really provide much help in understanding (the line is simple enough to know its expanding all raw_items), moreover we need to get the item_to_source mapping which requires two return values if we keep this function

codalab/lib/spec_util.py
```diff
@@ -11,9 +11,7 @@
 UUID_PREFIX_REGEX = re.compile('^0x[0-9a-f]{1,31}$')
 BEGIN_NAME_STR = '[a-zA-Z_]'
 NAME_STR = '[a-zA-Z_][a-zA-Z0-9_\.\-]*'
-NAME_PATTERN_STR = (
-    '[%\*a-zA-Z0-9_\.\-]+'
-)  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
+NAME_PATTERN_STR = '[%\*a-zA-Z0-9_\.\-]+'  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
```
yipenghe: black command did this

codalab/rest/interpret.py
```diff
@@ -130,7 +130,7 @@ def _interpret_genpath_table_contents():
         ]
     }
     ```
-
+
```
yipenghe: will remove

codalab/rest/interpret.py
```diff
@@ -158,6 +158,32 @@ def fetch_interpreted_worksheet(uuid):
     - resolve_interpreted_items: get more information about a worksheet.
     In the future, for large worksheets, might want to break this up so
     that we can render something basic.
+    Return:
```
yipenghe, line 161: Looking for suggestions on the documentation, the worksheet_info contains a lot of complexity so we really need a accurate and helpful documentation for later development

yipenghe:
This is kind of an important change (in my opinion) and I hope to make sure the change will actually make things clearer to what all the maps in worksheet_info mean and used for

codalab/rest/interpret.py
```diff
@@ -158,6 +158,32 @@ def fetch_interpreted_worksheet(uuid):
...
+    Return: 
+        worksheet_info dict{}:
+            key:[value_type] <description>
+            items:[list] resolved BLOCK items from raw_items, 
+                            bundles will be grouped into table block items, text items might be grouped into one markdown block
+                            TODO: This should really be renamed to block_items to avoid confusion, 
+                                    but too many places are using 'item'
+                                    and its not easy to tell if they all mean the same
+            source-list:[list] source lines
+            raw_to_block-list:[list] raw_item to its block index pair
```
epicfaace: If we've renamed "raw" to "source", then we might want to replace "raw_to_block" to "source_to_block", etc.

codalab/rest/interpret.py
```diff
@@ -158,6 +158,32 @@ def fetch_interpreted_worksheet(uuid):
...
     that we can render something basic.
+    Return: 
+        worksheet_info dict{}:
+            key:[value_type] <description>
+            items:[list] resolved BLOCK items from raw_items, 
+                            bundles will be grouped into table block items, text items might be grouped into one markdown block
+                            TODO: This should really be renamed to block_items to avoid confusion, 
+                                    but too many places are using 'item'
+                                    and its not easy to tell if they all mean the same
+            source-list:[list] source lines
```
percyliang: why `-list`?

codalab/rest/interpret.py
percyliang: Totally agree - we should probably make a Python object out of this (e.g., using `dataclass`)

codalab/rest/interpret.py
yipenghe: raw_to_block is actually raw items to block, not source

codalab/rest/interpret.py
percyliang: Probably should give broader context (either here or refer people to another place).
- source (raw worksheet_items in the DB, rendered as lines) [give example]
- blocks (semantic units that are rendered).

I wouldn't call it "block items" - I think items should always refer to raw items (because that's in the DB, which is annoying to change); just call them blocks everywhere.

percyliang (changes requested):
Definitely we should clean this up, thanks for doing this!  Some comments on how we can get to more consistent naming.

codalab/rest/interpret.py
yipenghe: by raw items, it means items rendered by the source but not grouped into block yet

codalab/rest/interpret.py
```diff
@@ -217,34 +254,36 @@ def fetch_interpreted_worksheet(uuid):
                         block['bundle_info'][j] = None
                 if not is_relevant_block:
                     interpreted_blocks['blocks'][i] = None
-
-    worksheet_info['items'] = resolve_interpreted_blocks(interpreted_blocks['blocks'])
+    # Grouped individual items into blocks
+    worksheet_info['blocks'] = resolve_interpreted_blocks(interpreted_blocks['blocks'])
```
yipenghe, line 282: After this line, the original worksheet_info['items'] are actually blocks

frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -281,8 +281,8 @@ class NewRun extends React.Component<{
         const { classes, ws } = this.props;
         this.shortcuts();
         let candidates: Bundle[] = [];
-        if(ws && ws.info && ws.info.items) {
-            ws.info.items.forEach((item) => {
+        if(ws && ws.info && ws.info.blocks) {
```
yipenghe, line 284: frontend code: just changing items=> blocks

codalab/rest/interpret.py
```diff
@@ -158,6 +158,34 @@ def fetch_interpreted_worksheet(uuid):
...
+        worksheet_info dict{}:
+            key:[value_type] <description>
+            blocks:[list] 
+                    Resolved worksheet blocks from raw_items.
+                    bundles will be grouped into table block items, 
+                    text items might be grouped into one markdown block etc.
+            source:[list] source lines
+            raw_to_block:[list] 
+                            Raw_items to its block index pair 
+                            for example, assume the first resolved block item is a bundle table that has 2 row,
```
teetone: Should it be "has 2 rows"?

teetone (approved):

codalab/rest/interpret.py
```diff
@@ -158,6 +158,34 @@ def fetch_interpreted_worksheet(uuid):
     - resolve_interpreted_items: get more information about a worksheet.
     In the future, for large worksheets, might want to break this up so
     that we can render something basic.
+    Return: 
+        worksheet_info dict{}:
+            key:[value_type] <description>
+            blocks:[list] 
+                    Resolved worksheet blocks from raw_items.
+                    bundles will be grouped into table block items,
```
armantajback: nit: revise casing through this block comment

yipenghe:
@percyliang  The merge is blocked by the outdated "request changes"

yipenghe:
@percyliang Can you re-review this / dismiss the request change?

percyliang (approved):
Great!

PR #2148, opened 2020-04-10
Memo union joins

candicegjing:
This draft PR simplifies the database query as much as I can. It basically alters the original SQL query to use join on all dependencies if there is any, and then union all those join queries to aggregate the final result in the end.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -677,8 +677,8 @@ class BundleModel(object):
 
         return self._execute_query(query)
 
-    def get_memoized_bundles(self, user_id, command, bundle_uuids, failed=False):
-        logger.info("command = {}, bundle_uuids = {}".format(command, bundle_uuids))
+    def get_memoized_bundles(self, user_id, command, dependencies):
+        logger.info("command = {}, bundle_uuids = {}".format(command, dependencies))
 
         filter_on_command = (
             select([cl_bundle.c.uuid])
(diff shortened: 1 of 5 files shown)


codalab/model/bundle_model.py
```diff
@@ -677,6 +677,90 @@ def get_bundle_uuids(self, conditions, max_results):
 
         return self._execute_query(query)
 
+    def get_memoized_bundles(self, user_id, command, dependencies):
+        '''
+        Get the most recently finished bundle that matches with input command and dependencies.
+        :param user_id: a string that specifies the current user id.
+        :param command: a string that defines the command that is used to search for memoized bundles in the database.
+        :param dependencies: a list of string in the structure of ["key1:uuid1","key2:uuid2"] defines
```
percyliang, line 685: "a list of strings of the form [...]"

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,90 @@ def get_bundle_uuids(self, conditions, max_results):
...
+                        )
+                        .select_from(
+                            cl_bundle.join(
+                                cl_bundle_dependency,
+                                cl_bundle.c.uuid == cl_bundle_dependency.c.child_uuid,
+                            )
+                        )
+                        .where(
+                            and_(
+                                cl_bundle.c.command == command,
```
percyliang, line 731: This should not be done for each dependency.

codalab/model/bundle_model.py
```diff
@@ -677,6 +677,90 @@ def get_bundle_uuids(self, conditions, max_results):
...
+        :param user_id: a string that specifies the current user id.
+        :param command: a string that defines the command that is used to search for memoized bundles in the database.
+        :param dependencies: a list of string in the structure of ["key1:uuid1","key2:uuid2"] defines
+                            dependencies that are used to search for memoized bundles in the database.
+        :return: a list of matched uuids.
+        '''
+
+        # When there is no dependency to be matched, the target memozied bundle
+        # should only exist in the bundle table but not in the bundle_dependency table.
+        if len(dependencies) == 0:
```
percyliang, line 692: I don't think you should need to special case the no dependencies case.

candicegjing: At this point, I wonder if we are actually talking about different things. I just want to make sure we are on the same page, could you please define the expected behavior when no dependency is passed to this function? 
In this PR, the behavior is defined as follows:
Suppose a user wants to run a command without any dependencies e.g. `date`, when he/she adds `--memoize` options in CLI (`cl run date --memoize`), if there is no previously created bundle, we return an empty list here. Otherwise, we want to return the most recently finished bundle that uses the same command `cl run date`.
In SQL, if we use the same JOIN query on the case when no dependency is specified as follows, we won't be able to retrieve any bundles even if there is one created before.
```
mysql> SELECT * FROM bundle JOIN bundle_dependency ON bundle.uuid = bundle_dependency.child_uuid WHERE bundle.command = 'date';
Empty set (0.00 sec)
```
Unless we do a **LEFT JOIN**:
```
mysql> SELECT uuid, child_uuid FROM bundle LEFT JOIN bundle_dependency ON bundle.uuid = bundle_dependency.child_uuid WHERE bundle.command = 'date';
+------------------------------------+------------+
| uuid                               | child_uuid |
+------------------------------------+------------+
| 0xe4cbf69320fa460ba151e18e4e0324b1 | NULL       |
| 0xb66da75ffb834050a3f8d5b8637fe3fb | NULL       |
| 0xfae685c553a9451eb22a4803b6e5da05 | NULL       |
 +------------------------------------+------------+
3 rows in set (0.00 sec)
```
But then this will be different than the **INNER JOIN** that we are currently using when user specified the dependencies. Let me know how you think.

Issue #2047, opened 2020-02-28
Set custom limits on memory available to worker


nelson-liu:
**Is your feature request related to a problem? Please describe.**

Right now, workers use `psutil.virtual_memory()` to get the memory limit available to jobs. However, there are cases where you'd want to set some custom memory limit—e.g., in Slurm, the controller might kill your worker process if the containers in aggregate exceed the memory limit [1].

[1] ok in practice this is not an issue, since slurm can't actually kill the docker images (a separate issue) and the docker runs aren't child processes / associated with the worker image. But, you'd probably still like to be courteous to your fellow cluster users.

**Describe the solution you'd like**
A flag like --cpuset or --gpuset or --max-work-dir-size that specifies the amount of memory to make available to the worker. Maybe something like `--max-memory`? Not sold on the name.

**Additional context**
This is like the RAM equivalent of #2046—overall, the goal is to ensure that the resources that slurm allocates are equivalent to the resources that a worker thinks it has.

PR #2151, opened 2020-04-13
set custom limits on memory through command line flag

yshen-google:
Issue #2047
Allow user to set custom memory amount through flag and taking the minimum of the custom memory and system total virtual memory.
Tested locally through flag control and memory amount was returned correctly.
@percyliang


codalab/worker/worker.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         commit_file,  # type: str
         cpuset,  # type: Set[str]
         gpuset,  # type: Set[str]
+        max_memory,  # type:Optinal[int]
```
nelson-liu: `s/Optinal/Optional/`

codalab/worker/worker.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         commit_file,  # type: str
         cpuset,  # type: Set[str]
         gpuset,  # type: Set[str]
+        max_memory,  # type:Optional[int]
```
candicegjing: Add a space between `:` and `Optional` and remove the square bracket around `int`.

candicegjing (commented):
Thanks for doing this! Other than the small formatting issue, everything else looks great to me.

codalab/worker/worker.py
yshen-google: Thanks for the review.
Space is added.
Regarding square bracket, it seems the other occurrences also use `Optional[{data type}]` style such as `dependency_manager,  # type: Optional[DependencyManager]` and `local_bundles_dir,  # type: Optional[str]`.

percyliang (approved):
Thanks!

codalab/worker/worker.py
candicegjing: Yes, you are right. Annotated arguments with a default value of None should use `[]`.

candicegjing (approved):
Thanks for doing this! Looks good to me!

Issue #1803, opened 2019-12-02
Provide more information about search results


yipenghe:
**Is your feature request related to a problem? Please describe.**

Currently if no result, the terminal doesn't says so, it also doesn't say it only lists 10 results by default

**Describe the solution you'd like**

Return no results when search/wsearch got 0 results, also say that only listing 10 results by default

PR #2160, opened 2020-04-15
Add more information to search results

yshen-google:
Fix to #1803

1. If there is no result, print "No results found".
2. Indicate that at most 10 results are returned even if there are more.

@percyliang


codalab/lib/bundle_cli.py
```diff
@@ -1924,14 +1926,21 @@ def do_search_command(self, args):
             params={'worksheet': worksheet_uuid, 'keywords': args.keywords, 'include': ['owner']},
         )
 
+        print('Note: at most 10 results will be shown.')
```
percyliang: Make the `10` filled in from a variable that governs the actual logic as opposed to hard-coded.

codalab/lib/bundle_cli.py
```diff
@@ -1924,14 +1926,21 @@ def do_search_command(self, args):
...
+
         # Print direct numeric result
         if 'meta' in bundles:
+            if len(bundles['meta']['result']) == 0:
```
percyliang: I don't think we need this here

codalab/worker/main.py
```diff
@@ -89,6 +89,13 @@ def parse_args():
         'the requested image will not be downloaded. '
         'The bundle depends on this image will fail accordingly.',
     )
+    parser.add_argument(
```
percyliang: This should not be part of this PR

codalab/lib/bundle_cli.py
yshen-google: Good point. I moved it to a const in bundle_model.py.

codalab/worker/main.py
yshen-google: Yeah... it was from the previous PR. Sorry that I didn't start this branch out of clean workspace.. Since the previous PR is merged to master, this should be resolved by now.

codalab/lib/bundle_cli.py
yshen-google: I see. I thought we also wanted to print that 10 results at most note when no aggregated information is returned. But I agree with you that it is needed here since it is aggregated result like count, etc. Done.

yshen-google:
Changes:

1. Move the 10 results at most note when len(bundles) == 0.
2. Move value 10 to a const variable in bundle_model.py.

PTAL

codalab/lib/bundle_cli.py
```diff
@@ -1931,7 +1933,10 @@ def do_search_command(self, args):
 
         # Print table
         if len(bundles) > 0:
+            print('Note: at most {} results will be shown.'.format(bundle_model.SEARCH_RESULTS_LIMIT))
```
percyliang: Only print this out when `len(bundles) == SEARCH_RESULTS_LIMIT`.  Otherwise, it's the full list.

percyliang: And in that case, print "Only {} results are shown.  Use .limit=N to show N."

nelson-liu:
Thoughts on sending the message to stderr ? I have some scripts downstream that interact with the stdout of `cl search`, and it'd be nice if existing functionality was preserved for them...

percyliang:
Related: the extra message should only be shown when `-u` is not specified (which only prints out the raw uuids).

codalab/lib/bundle_cli.py
yshen-google: Sounds good.

yshen-google:
> Thoughts on sending the message to stderr ? I have some scripts downstream that interact with the stdout of `cl search`, and it'd be nice if existing functionality was preserved for them...

Sounds good. Moved output to stderr.

yshen-google:
> Related: the extra message should only be shown when `-u` is not specified (which only prints out the raw uuids).

Done. This makes it ideal to have the print statement inside of print_bundle_info_list() since uuid arg is passed to that function.

yshen-google:
Changes so far:

1. Add changes to wls in addition of search.
2. Move outputs to stderr and only print extra messages when uuid flag is set to false.

yshen-google:
Not sure why travis-ci doesn't run at all. Filed #2191 to track.

yshen-google:
Closed this PR due to #2191

PR #2166, opened 2020-04-15
Fix CPU worker ranking problem

### LINKED ISSUE #2153: CPU job assignment issue
Currently, jobs that were requested to run on CPU workers are still being assigned to a GPU worker even with CPU workers were idle during the time of scheduling.
1. Without any dependencies specified
Screen Shot 2020-04-13 at 10 25 17 AM

1. With dependencies specified
Screen Shot 2020-04-13 at 10 26 05 AM

We want to avoid dispatching jobs that only require to run on CPU workers to GPU workers as much as possible.

candicegjing:
Fixed #2153 

In [filter_and_sort_workers](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L413), the computing resource values (CPU/GPU/Memory) in `workers_list` have already been altered and cannot reflect their original value. Then, in the sorting logic, when prioritizing GPU workers by `worker['gpus'] == 0`, it couldn't truly reflect if the worker has GPUs or not. 

There are two ways of changing adding this flag:
1. add an additional `has_gpus` flag in the `WorkerInfoAccessor` object
    1. Pros: each worker will come with this flag and easy to query. 
    1. Cons: alters the original worker object as it is retrieved from the database module (mis-match on length)
1. pass an additional parameter to indicate the original number of GPUs between functions in `bundle_manager.py`. 
    1. Pros: doesn't affect other parts of the logic, only scheduling part
    1. Cons: may not look clean.

### DIFF
--- a/codalab/server/worker_info_accessor.py
+++ b/codalab/server/worker_info_accessor.py
@@ -33,6 +33,9 @@ class WorkerInfoAccessor(object):
             for uuid in worker['run_uuids']:
                 self._uuid_to_worker[uuid] = worker
             self._user_id_to_workers[worker['user_id']].append(worker)
+            # 'gpus' field contains the number of free GPUs that comes with each worker. Adding an additional
+            # 'has_gpus' flag here to indicate if the current worker has GPUs or not.
+            worker['has_gpus'] = True if worker['gpus'] > 0 else False
 
     @refresh_cache
     def workers(self):
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -475,7 +475,7 @@ class BundleManager(object):
             # 6. break ties randomly by a random seed.
             return (
                 not worker['tag_exclusive'],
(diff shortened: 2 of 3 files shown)


codalab/server/worker_info_accessor.py
```diff
@@ -33,6 +33,8 @@ def _fetch_workers(self):
             for uuid in worker['run_uuids']:
                 self._uuid_to_worker[uuid] = worker
             self._user_id_to_workers[worker['user_id']].append(worker)
+            # Add an additional 'has_gpus' flag to indicate if the current worker has GPUs or not
+            worker['has_gpus'] = True if worker['gpus'] > 0 else False
```
percyliang, line 38: Add a comment that `gpus` is the number of free GPUs in general, so this isn't just trivial.

percyliang (approved):

teetone (approved):

PR #2167, opened 2020-04-15
Cleanup docker process if cl-worker gets terminated

### LINKED ISSUE #1908: docker doesn't respect slurm
When job gets pre-empted, because docker runs as root, the job doesn't actually get killed...  Might need to use singularity.

candicegjing:
Fixed #1908

This PR is to fix orphan processes that were left after kill the current `cl-worker` session in Slurm.
Currently, I am killing the container directly. Let me know if we want to have any other behaviors.

### DIFF
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -93,8 +93,6 @@ class Worker:
         self.last_checkin_successful = False
         self.last_time_ran = None  # type: Optional[bool]
 
-        self.containers = set()
-
         self.runs = {}  # type: Dict[str, RunState]
         self.init_docker_networks(docker_network_prefix)
         self.run_state_manager = RunStateMachine(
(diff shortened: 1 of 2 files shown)


epicfaace (approved):

codalab/worker/worker.py
```diff
@@ -209,6 +209,15 @@ def cleanup(self):
...
             logger.error("Cannot clear docker networks: %s", str(e))
 
+        # filter out existing runs
+        containers = [run.container for run in self.runs.values() if run.container_id is not None]
+        logger.info("Cleaning up bundle containers: {}".format(containers))
+        try:
+            for container in containers:
+                container.kill()
+        except docker.errors.APIError as e:
+            logger.info("Cannot clean up bundle containers: {}".format(str(e)))
```
teetone, line 219: This should be `logger.exception` or `logger.error`.

candicegjing:
Closing this PR now as there is another solution (#2169) to solve this issue.

Issue #2150, opened 2020-04-11
CodaLab workers could be more robust to (temporary) network disconnections


nelson-liu:
**Describe the bug**

I'm running several cl workers (think ~15), and when I logged in today, all my bundles were `worker_offline`. Checking the workers, I noticed that they all froze at `2020-04-11 05:06:16,786`.

Before, when this happens, I would have seen the "sleeping for 1 hour, help me" message. Now, even though there's no message, the process is still sleeping (I can't `Ctrl-C` out of it, need to go to htop and kill it and then restart). Ideally, it would reconnect to the server when it can and resume its checkins.

**To Reproduce**
Steps to reproduce the behavior :
1. Run a bunch of workers.
2. Wait for a network blip, or create one.

**Expected behavior**

The workers reconnect when they can.

percyliang:
That message means some internal error happened corresponding to some CodaLab bug that needs to be fixed, which is why we're not automatically reconnecting (otherwise, we'd likely not fix the actual problem).  There's no error message / stack trace before that in the docker logs?

nelson-liu:
Nope, there isn't—I expected there to be one too. Maybe something off with the logging level (the worker was run with `--verbose`)?

percyliang:
Can you reproduce this easily?  I thought the workers should be generally robust to disruptions in network...

nelson-liu:
This is fixed, I've seen them recover a few times now.

PR #2169, opened 2020-04-16
Clean up unfinished bundles

### LINKED ISSUE #1908: docker doesn't respect slurm

candicegjing:
Fixed #1908 

This PR provides another solution to #1908 . Basically, the solution in #2167 doesn't wait until all the status being uploaded to the bundle page, whereas the solution in this PR adds this functionality. The tradeoff is the waiting time from the user who kills the worker process. This PR increases the waiting time after a user sends the termination signal to cl-worker.

### DIFF
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -132,6 +132,12 @@ def parse_args():
         action='store_true',
         help='To be used when the worker should only run bundles that match the worker\'s tag.',
     )
+    parser.add_argument(
+        '--terminate',
+        action='store_true',
+        help='Terminate the worker and kill all the existing running bundles.',
+    )
+
     return parser.parse_args()
 
 
@@ -222,6 +228,7 @@ def main():
         args.tag_exclusive,
         docker_runtime=docker_runtime,
         docker_network_prefix=args.network_prefix,
+        terminate=args.terminate,
     )
 
     # Register a signal handler to ensure safe shutdown.
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -62,6 +62,8 @@ class Worker:
         tag_exclusive,  # type: bool
         docker_runtime=docker_utils.DEFAULT_RUNTIME,  # type: str
         docker_network_prefix='codalab_worker_network',  # type: str
+        # A flag indicating if all the existing running bundles will be killed along with the worker.
+        terminate=False,  # type: bool
     ):
         self.image_manager = image_manager
         self.dependency_manager = dependency_manager
@@ -90,6 +92,8 @@ class Worker:
         self.idle_seconds = idle_seconds
 
         self.stop = False
+        self.terminate = terminate
+
         self.last_checkin_successful = False
         self.last_time_ran = None  # type: Optional[bool]
 
@@ -180,6 +184,10 @@ class Worker:
                 self.save_state()
                 self.checkin()
                 self.save_state()
+                # Wait until existing bundles finished uploading their final status
+                if self.terminate:
+                    if self.terminate_containers() == 0:
+                        self.stop = True
                 if self.check_idle_stop():
                     self.stop = True
                     break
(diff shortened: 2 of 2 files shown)


codalab/worker/worker.py
```diff
@@ -191,6 +196,27 @@ def start(self):
...
 
+    def terminate_containers(self):
+        """
+        Clean up bundle containers by moving all the existing unfinished bundles to terminal states.
+
+        :returns: the number of unfinished bundles.
+        """
+        container_count = 0
+        try:
+            for bundle_uuid in self.runs.keys():
```
percyliang: Don't need `.keys()`

codalab/worker/worker.py
```diff
@@ -191,6 +196,27 @@ def start(self):
...
+                run_state = self.runs[bundle_uuid]
+                if run_state.stage != RunStage.FINISHED and run_state.container_id != None:
```
percyliang: `is not None`

codalab/worker/worker.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         self.idle_seconds = idle_seconds
 
         self.stop = False
+        self.terminating = False
```
percyliang: Document what the semantics of `terminating` and `stop` are, and what the flow is somewhere.

percyliang (approved):

nelson-liu:
Thoughts on having this as a flag to `cl-worker`? There are legitimate times where I need to restart a worker (e.g., network blips #2150) and I'd rather not lose all my runs as a result...

percyliang:
@nelson-liu : good point, let's make it a flag.  Now that I think about it, there are many times when we want to restart the worker and not kill its jobs (like on prod), so the default should be off.  Only when we run this inside slurm should the flag be on.

candicegjing:
@percyliang and @nelson-liu I have added a CLI argument to the worker, so that in the future when we want to kill running bundles along with worker, we just need to add a `--ternimate` flag when we start the worker. If we don't set the `--terminate` flag when starting the worker, the behavior will remain the same as it is right now, meaning killing the worker won't kill the running jobs. Let me know if this logic is desired.

codalab/worker/main.py
```diff
@@ -132,6 +132,12 @@ def parse_args():
         action='store_true',
         help='To be used when the worker should only run bundles that match the worker\'s tag.',
     )
+    parser.add_argument(
+        '--terminate',
```
percyliang, line 136: Make the flag more accurate: `--pass-down-termination`

codalab/worker/main.py
```diff
@@ -132,6 +132,12 @@ def parse_args():
...
+        action='store_true',
+        help='Terminate the worker and kill all the existing running bundles.',
```
percyliang, line 138: "If worker is terminated, kill all running bundles."

codalab/worker/worker.py
```diff
@@ -191,6 +199,24 @@ def start(self):
...
+        """
+        Clean up bundle containers by moving all the existing unfinished bundles to terminal states.
+
+        :returns: the number of unfinished bundles.
+        """
+        unfinished_bundles = []
+        for bundle_uuid in self.runs:
+            run_state = self.runs[bundle_uuid]
+            if run_state.stage != RunStage.FINISHED and run_state.container_id is not None:
+                run_state = run_state._replace(
```
percyliang, line 212: Should this be failed?  One could imagine this be restaged so that it gets tried again (somewhat risky if we get into infinite loops, but would be more ergonomic for the user in terms of being able to launch and forget).  @nelson-liu might have opinions.

nelson-liu: Restaging would be amazing, if it's easy to do here.

candicegjing: Since this PR already merged, let's add this to another PR. To clarify, do you guys want to move this bundle back to `staged` state, meaning it will be reassigned to a worker (even though it could be the same worker) from **bundle-manager**? Or do you guys just want to rerun it?

nelson-liu: i was thinking staged...it couldn't be the same worker unless the worker was rerun, right? Since the original worker was terminated.

candicegjing: hmm, I was thinking towards the case if the worker terminated earlier got reconnected again...

candicegjing: And yes, moving bundles from workers to `staged` state is something missing and I am working on. Then we can move those terminated bundles to the original `staged` state. Do you guys have other thoughts?

PR #2173, opened 2020-04-16
1287: refactor + fix copy test

### LINKED ISSUE #1287: fix copy test in test_cli.py
Currently, the CLI tests work only against one instance.  Allow starting two instances via ./codalab_service.py and passing one of the instances to test_cli.py so test copying bundles between two instances.

teetone:
Resolves #1287 

PR for refactoring to fix copy test. Created a test runner that now creates a second instance and runs the test. The second instance is passed into `test_cli`. Previous test logic is moved out of `codalab_service` to separate out the logic of starting CodaLab vs running tests on an instance of CodaLab.

Since we now create a second instance to test `Copy`, Travis builds now take an additional 3-4 minutes to complete.

Note: Copy test works now, except for the `cl cat` calls (see comment below).


candicegjing (commented):
So in the future, test_runner will include all the tests but codalab_service will still be responsible for building the services right?

test_cli.py
```diff
@@ -1,16 +1,13 @@
 # -*- coding: utf-8 -*-
 """
 Tests all the CLI functionality end-to-end.
-
```
percyliang: Why are we removing all these newlines?  Seems like it makes it harder to read.

test_cli.py
```diff
@@ -1058,17 +1006,14 @@ def test(ctx):
     dependent = _run_command([cl, 'run', ':%s' % special_name, 'cat %s/stdout' % special_name])
     wait(dependent)
     check_equals('hello', _run_command([cl, 'cat', dependent + '/stdout']))
-
```
percyliang: Can we leave all the newlines there?

test_runner.py
```diff
@@ -0,0 +1,132 @@
...
+
+import argparse
+import random
+import string
+import subprocess
+import sys
+
+
+class TestRunner(object):
+    _CODALAB_SERVICE_SCRIPT = 'codalab_service.py'
```
percyliang, line 18: Add docstring with overview of this file

test_runner.py
```diff
@@ -0,0 +1,132 @@
...
+            raise
+
+        return instance
+
+    def __init__(self, instance, tests):
+        self.instance = instance
+        self.temp_instance_name = 'temp-instance%s' % ''.join(
+            random.choice(string.digits) for _ in range(8)
+        )
+        self.temp_instance = TestRunner._create_temp_instance(self.temp_instance_name)
```
percyliang: We should only do this if `tests` require it...should be an annotation of the test.

percyliang (commented):
I think it's cleaner not having tests in `codalab_serivce.py`, but do we need `test_runner.py`?  I feel like this might be adding too many layers.  I think it'd be simpler if we just called `./codalab_service.py` twice to spin up two instances.  Then we just pass the two instances to `test_cli.py`.

test_cli.py
teetone: Not sure what happened here. I think when I copied the file over from another branch the spaces got removed. I fixed the spaces now.

test_runner.py
teetone: I added a check for this. I now only create a second instance if we are running `default`, `all` or if one of the test modules is `copy`.

teetone:
> I think it's cleaner not having tests in `codalab_serivce.py`, but do we need `test_runner.py`? I feel like this might be adding too many layers. I think it'd be simpler if we just called `./codalab_service.py` twice to spin up two instances. Then we just pass the two instances to `test_cli.py`.

@percyliang I think that would be cleaner and ideal, but the `test_runner.py` now has logic to see if a second instance is needed, find available ports that the second instance can use, clean up logic to shut down the second instance after testing and also runs front end tests. We can also add more test-related execution logic to `test_runner.py` in the future.

test_cli.py
```diff
@@ -1415,35 +1374,37 @@ def stress(
...
+
         # Upload to original worksheet, transfer to remote
         _run_command([cl, 'work', source_worksheet])
         uuid = _run_command([cl, 'upload', test_path('')])
         _run_command([cl, 'add', 'bundle', uuid, '--dest-worksheet', remote_worksheet])
-        check_agree([cl, 'info', '-f', 'data_hash,name', uuid])
-        check_agree([cl, 'cat', uuid])
+        compare_output([cl, 'info', '-f', 'data_hash,name', uuid])
+        # TODO: `cl cat` is not working even with the bundle available
+        # compare_output([cl, 'cat', uuid])
```
teetone: `Copy` test works now, but for some reason `cl cat` fails with no errors (i.e. < none >) and I can't figure it out. The original bundle and copied bundle both exist on the main instance and second instance respectively. @percyliang Is it okay if we create a separate, follow-up issue for this?

Failing with:
fix_cat

.travis.yml
```diff
@@ -34,13 +34,13 @@ jobs:
...
-        - docker restart codalab_worker_1 && python3 codalab_service.py test run --version ${TRAVIS_BRANCH}
+        - python3 test_runner.py --version ${TRAVIS_BRANCH} default
+        # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization)
+        - docker restart codalab_worker_1 && python3 test_runner.py --version ${TRAVIS_BRANCH} run
         - docker stop codalab_worker_1
-        - python3 codalab_service.py start --services worker --version ${TRAVIS_BRANCH} --shared-file-system
-        - python3 codalab_service.py test --version ${TRAVIS_BRANCH} --shared-file-system basic run read write kill resources netcat netcurl workers
-        - python3 tests/ui/ui_tester.py --headless
+        - python3 codalab_service.py start --services init worker --version ${TRAVIS_BRANCH} --shared-file-system
+        # Also run the frontend tests in this iteration of testing
```
teetone: I created a module `frontend` that `test_runner` recognizes and runs the front end tests.

codalab_service.py
```diff
@@ -667,7 +659,9 @@ def run_service_cmd(self, cmd, root=False, service='rest-server'):
 
     @staticmethod
     def wait(host, port, cmd):
-        return '/opt/wait-for-it.sh {}:{} -- {}'.format(host, port, cmd)
+        return '/opt/wait-for-it.sh --timeout={} {}:{} -- {}'.format(
```
teetone, line 662: I noticed that creating a new instance can take up to 4 minutes. I added a timeout here.

codalab_service.py
```diff
@@ -270,6 +271,20 @@ def has_callable_default(self):
         default=2048,
         help='Default memory (in MB) for each worker in the AWS Batch Worker Manager',
     ),
+    CodalabArg(
```
teetone, line 274: I added these environment variables because of a known Docker compose issue: https://github.com/docker/compose/issues/3927. Without these the build will fail intermittently.

codalab_service.py
```diff
@@ -697,6 +691,7 @@ def start_services(self):
                     ('cli/default_address', rest_url),
                     ('server/engine_url', mysql_url),
                     ('server/rest_host', '0.0.0.0'),
+                    ('server/rest_port', self.args.rest_port),
```
teetone, line 694: This took me a long time to figure out, but I noticed that the value of the `rest_port` parameter that is being passed in was never respected and was hardcoded with the default value of `2900` in `codalab_manager`.

docker/compose_files/docker-compose.yml
```diff
@@ -214,8 +234,12 @@ services:
       - ${CODALAB_HOME}:${CODALAB_HOME}
       - ${CODALAB_BUNDLE_MOUNT}:${CODALAB_BUNDLE_MOUNT}
       - ${CODALAB_MONITOR_DIR}:${CODALAB_MONITOR_DIR}
+    networks:
+      - rest-server
 
 networks:
+  rest-server:
```
teetone, line 241: I defined a new network called `rest-server`. Any service on this network can simply communicate with the rest server, including the rest server of a second instance.

test_runner.py
```diff
@@ -0,0 +1,160 @@
...
+            self._run_frontend_tests()
+
+        except subprocess.CalledProcessError as ex:
+            print('Exception while executing tests: %s' % ex.output)
+            success = False
+
+        self._cleanup()
+        return success
+
+    def _run_frontend_tests(self):
```
teetone, line 105: I run the Selenium tests here now too as long as the `frontend` module key is passed in.

docker/compose_files/docker-compose.yml
percyliang: What access does this allow precisely?  Before, the various docker containers (e.g., frontend) could still talk to the rest-server, no?

test_cli.py
```diff
@@ -1415,35 +1374,37 @@ def stress(
...
     """Test copying between instances."""
     source_worksheet = current_worksheet()
 
-    with temp_instance() as remote:
-        remote_worksheet = remote.home
-        _run_command([cl, 'work', remote_worksheet])
+    with remote_instance(ctx.second_instance) as remote:
 
-        def check_agree(command):
+        def compare_output(command):
```
percyliang: Maybe `compare_output_across_instances`?

test_cli.py
percyliang: Sure.

.travis.yml
```diff
@@ -34,13 +34,13 @@ jobs:
...
+        - python3 test_runner.py --version ${TRAVIS_BRANCH} frontend basic run read write kill resources netcat netcurl workers
```
percyliang, line 43: Why the long list of tests?  Is this for `--shared-file-system`?  If so, be explicit, but I don't think you need all of these.

percyliang (commented):
Overall looks good, just had a few questions.

docker/compose_files/docker-compose.yml
teetone: By default, all these containers could talk to each other without explicitly defining `networks` for each one, since a default network gets created for the app. 
I had to add `networks` with the same name for each container in order for all containers   to be able to talk to all other containers in a duplicate set for the second instance. I can try to find a simpler solution too.

teetone: I think once we define `networks` for one container or service, we override the default network.

test_cli.py
teetone: Yes, that's more explicit. I will update.

.travis.yml
teetone: I think this is to test shared file systems. This list of tests (aside from `frontend`)  was there before this change and I just copied it over. Should I maybe just run `basic`, `read`, `write`, `resources` and `workers`?

percyliang: I would have guessed that `basic` would be enough?  But maybe @bkgoksel has deeper context.

test_cli.py
teetone: Follow-up issue: https://github.com/codalab/codalab-worksheets/issues/2228

.travis.yml
teetone: I synced with @bkgoksel and it turns out these test modules somewhat tests shared file system code paths.

percyliang: Ok, then I guess we can leave them in for now and add a comment for why those are there.  In the future, we might want to figure out a more minimal set of tests.

teetone: I updated the comment and created a follow-up issue for it to track the work: https://github.com/codalab/codalab-worksheets/issues/2243

percyliang (approved):
Great to be able to test that logic now - thanks!

Issue #1123, opened 2019-04-03
Search bar to quickly switch between worksheets


bkgoksel:
Perhaps a dropdown on the navbar, or if we add a search bar, good to upweight a user's own worksheets on the ranked results.

percyliang:
This search bar can be universal where it can surface other things like bundles, users, and help.

Levilian:
Search bar functionality v0.1:

- [x] display results for `cl wsearch`
- [x] toggle between different personal worksheets
- [x] search other people's worksheets by name

Levilian:
Search bar functionality v0.2:

- [x] search multiple keywords at the same time
- [x] display advanced information - owner name, etc.
- [x] highlight the part of the title/name that matches keywords
- [x] make the search box bigger, more to the left
- [x] placeholder text: Search worksheets
- [x] keyboard shortcut to trigger (a + f)
- [x] rank .mine worksheets higher than others

PR #2181, opened 2020-04-17
Search bar for worksheets

Levilian:
Feature: #1123


codalab/rest/interpret.py
```diff
@@ -460,6 +464,23 @@ def set_error_data(block_index, message):
     return interpreted_blocks
 
 
+def interpret_wsearch(keywords):
+
+    """
+    Request to fetch a resource or resources.
+
+    :param resource_type: resource type as string
```
epicfaace: Can you update this doc?

codalab/rest/interpret.py
```diff
@@ -75,7 +75,11 @@ def _interpret_wsearch():
     }
```
epicfaace, line 81: Can you change this comment from "Returns worksheet items given a search query for worksheets" -- I don't think it returns worksheet items

frontend/package.json
```diff
@@ -44,6 +45,7 @@
     "reselect": "^4.0.0",
     "sass": "^1.23.3",
     "seamless-immutable": "^7.1.4",
+    "semantic-ui-react": "^0.88.2",
```
epicfaace, line 49: We're already using bootstrap -- can we not add the entire semantic-ui library? Could we instead have a component for a single search?

epicfaace (commented):
Can you add more space above the worksheet title?

on production:

image

this branch:

image

frontend/package.json
epicfaace: If we can also change the font to the default font (`"Roboto", "Helvetica Neue", sans-serif`) that would be great!

frontend/src/components/NavBar.js
```diff
@@ -107,10 +110,130 @@ class NavBar extends React.Component<{
...
+                    //             "title": "Brakus Group",
+                    //             "description": "Cloned interactive Graphic Interface",
+                    //         },
+                    //     ]
+                    // }
+                    const filteredResults = _.reduce(
+                        data.response,
+                        (memo, item) => {
+                            item.description = item.name.replace(re, '<mark>$&</mark>');
+                            item.title = item.title.replace(re, '<mark>$&</mark>');
```
epicfaace:
```suggestion
                            item.title = (item.title || "").replace(re, '<mark>$&</mark>');
```

I get this error because sometimes a worksheet's title will be equal to `null`.

image

frontend/src/components/NavBar.js
```diff
@@ -107,10 +110,130 @@ class NavBar extends React.Component<{
...
+
+    handleResultSelect = (e, { result }) => {
+        window.open('/worksheets/' + result.uuid, '_self');
+    };
+
+    initialState = { isLoading: false, results: [], value: '' };
+
+    resultRenderer = ({ title, description }) => [
+        <div key='content' className='content'>
+            {title && <div dangerouslySetInnerHTML={{ __html: title }} className='title'></div>}
```
epicfaace, line 143: You should always call `DOMPurify.sanitize(title)` before using `dangerouslySetInnerHTML`, or you have an XSS security risk (see https://github.com/cure53/DOMPurify -- and the same goes with `description`)

frontend/src/components/NavBar.js
```diff
@@ -107,10 +110,130 @@ class NavBar extends React.Component<{
...
+                    // {
+                    //     "name": "codalab",
+                    //     "results": [
+                    //         {
+                    //             "title": "Brakus Group",
+                    //             "description": "Cloned interactive Graphic Interface",
+                    //         },
+                    //     ]
+                    // }
+                    const filteredResults = _.reduce(
```
epicfaace: I think the code here would be clearer if you just used a for loop instead of `_.reduce`.

frontend/src/components/NavBar.js
```diff
@@ -136,8 +258,28 @@ class NavBar extends React.Component<{
...
+                                    results={results}
+                                    value={value}
+                                    showNoResults={true}
+                                    id='search-bar'
+                                />
+                            </div>
+                        )}
                         {!this.props.auth.isAuthenticated && (
                             <React.Fragment>
+                                <div className={classes.searchContainer}></div>
```
epicfaace:
```suggestion
                                <div className={classes.searchContainer} />
```

frontend/src/components/NavBar.js
```diff
@@ -122,7 +245,6 @@ class NavBar extends React.Component<{
             info: InfoIcon,
             warning: WarningIcon,
         }[this.state.snackbarVariant];
-
```
epicfaace: keep empty line

frontend/package.json
Levilian: Got it! Let me see how to only use a single component.

Levilian: Even if we are adding the entire library, we are only importing the relevant js files from the library. Does this still matter?

candicegjing (commented):
How should we search bundles in the search bar? Is there a pattern that we should follow? I just tried locally, but if I just search for `time`, it throws out an error. Maybe I didn't search it in the right way?
Screen Shot 2020-05-11 at 2 53 20 PM

Levilian:
> How should we search bundles in the search bar? Is there a pattern that we should follow? I just tried locally, but if I just search for `time`, it throws out an error. Maybe I didn't search it in the right way?
> Screen Shot 2020-05-11 at 2 53 20 PM

Are you on the latest branch? I'm sure this bug has been fixed a while ago...

frontend/src/components/NavBar.js
```diff
@@ -107,10 +111,158 @@ class NavBar extends React.Component<{
...
+                    */
+                    let filteredResults = {};
+                    for (let item of data.response) {
+                        // use DOMPurify to get rid of the XSS security risk
+                        item.description = DOMPurify.sanitize(
+                            item.name.replace(re, "<span id='highlight'>$&</span>"),
+                        );
+                        item.title = DOMPurify.sanitize(
+                            (item.title || '').replace(re, "<span id='highlight'>$&</span>"),
+                        );
```
Levilian, line 216: @candicegjing

candicegjing: Yes. I just pulled from master and it works now.

candicegjing (approved):
LGTM!

candicegjing:
@Levilian Just some thoughts on the search bar, it doesn't have to be implemented.
1. I was wondering if we could consider adding padding to the search bar? 
1. Shorten the length of the search bar?
1. When we are not signed in, the search bar doesn't show on the home page. When we are signed in, the search bar shows there. I just wonder if we want to make them consistent for the home page.

yipenghe (approved):
Screenshot from 2020-05-12 00-31-45
Great! Before merging, maybe address the followings:

The search seems to be taking too much space than it needs, I think it would better to limit its max width to be same as the search result menu and maybe place near the left/right side.

I think we should allow searching even when the user is not log in, as they can still use the wsearch command terminal anyway. This would be a great way for visitors to explore worksheets by searching.

Levilian:
> @Levilian Just some thoughts on the search bar, it doesn't have to be implemented.
> 
> 1. I was wondering if we could consider adding padding to the search bar?
> 2. Shorten the length of the search bar?
> 3. When we are not signed in, the search bar doesn't show on the home page. When we are signed in, the search bar shows there. I just wonder if we want to make them consistent for the home page.

Will do! Do we want to get this version in first since you'll deploy soon?

Levilian:
> Screenshot from 2020-05-12 00-31-45
> Great! Before merging, maybe address the followings:
> 
> The search seems to be taking too much space than it needs, I think it would better to limit its max width to be same as the search result menu and maybe place near the left/right side.
> 
> I think we should allow searching even when the user is not log in, as they can still use the wsearch command terminal anyway. This would be a great way for visitors to explore worksheets by searching.

I see. To implement the search, I called some local api and so I thought the requirement is to log in and verify the user identity. I don't know if the current implementation allows for searching without log-in. I'll test it.

candicegjing:
> > @Levilian Just some thoughts on the search bar, it doesn't have to be implemented.
> > 
> > 1. I was wondering if we could consider adding padding to the search bar?
> > 2. Shorten the length of the search bar?
> > 3. When we are not signed in, the search bar doesn't show on the home page. When we are signed in, the search bar shows there. I just wonder if we want to make them consistent for the home page.
> 
> Will do! Do we want to get this version in first since you'll deploy soon?

It doesn't have to be in this PR. We could make it in another PR.

candicegjing:
> > @Levilian Just some thoughts on the search bar, it doesn't have to be implemented.
> > 
> > 1. I was wondering if we could consider adding padding to the search bar?
> > 2. Shorten the length of the search bar?
> > 3. When we are not signed in, the search bar doesn't show on the home page. When we are signed in, the search bar shows there. I just wonder if we want to make them consistent for the home page.
> 
> Will do! Do we want to get this version in first since you'll deploy soon?

It doesn't have to be in this PR. We could make it in another PR.

frontend/package.json
epicfaace: @Levilian did you get a chance to look into this? including the entire semantic css may interfere with bootstrap css.

Levilian: @epicfaace So it seems there are two issues. One is adding the entire semantic-ui-react package. The other is the potential clash between semantic css and material css. 

The issue I run into with the first one is that the code not only depends on the behavior of Search js in semantic ui, but also the behavior of, say, Textbox as well. What is the true impact of importing only relevant js files from a sizable package? I thought only the size of the imported files determines the load time? 

The second issue - I was able to identify the relevant parts of the semantic ui css that would maintain the current look. Do I simply copy and paste those parts into our already bulky css? Or is the goal to scrape the semantic ui css and build new css?

PR #2182, opened 2020-04-17
Make search items respect schema headers; fix a focus issue

### LINKED ISSUE #2180: Missing remote column after v0.5.12 deployement
In [What's running on CodaLab?](https://worksheets.codalab.org/worksheets/0xa590fd1b68944a1a95c1c40c4931dc7b), we used to have a column that displays information about which worker the current bundle runs on. Now after `v0.5.12` deployment, this column disappeared as follows:

Screen Shot 2020-04-16 at 8 04 28 PM

This ticket is to investigate this issue and add that column back.

yipenghe:
fix #2180 
The placeholder item doesn't record schema header information, thus all search expanded tables now don't respect the schemas, this PR fixes it.

Also fixed a setFocus issue, where the calculated offset is 0 causes an infinite recursion

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -449,6 +449,9 @@ class Worksheet extends React.Component {
         if (item && (!item.props || !item.props.item)) {
             // Skip "no search results" items and scroll past them.
             const offset = index - this.state.focusIndex;
+            if (offset === 0) {
+                return;
+            }
             this.setFocus(index + offset, subIndex, shouldScroll);
             return;
         }
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -1101,8 +1101,12 @@ class Worksheet extends React.Component {
         { moveIndex = false, textDeleted = false } = {},
     ) => {
         let itemHeights = {};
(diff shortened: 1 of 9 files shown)


epicfaace:
I see the right header, but an empty table here

image

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -449,6 +449,9 @@ class Worksheet extends React.Component {
         if (item && (!item.props || !item.props.item)) {
             // Skip "no search results" items and scroll past them.
             const offset = index - this.state.focusIndex;
+            if (offset === 0) {
```
epicfaace, line 452: In what scenario did this give you an error?

epicfaace:
@yipenghe the initial solution you had didn't work because the table and its schemas need to be both available in `resolve_interpreted_items`; we can't dynamically combine them on the frontend as the rows need to be generated based on the headers on the backend `/interpret` endpoint. I've updated this PR to make it work though.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yipenghe: I was hitting on the last row of a search result table, clicked edit source and save

epicfaace:
Actually, there's still issues with this -- working on it

epicfaace (approved):

Issue #2094, opened 2020-03-19
Migrate CI build from Travis to GitHub Actions


candicegjing:
Move CI build from Travis to Git Action would save us around 10 mins (Travis build takes ~30 mins vs Action build takes ~20 mins)

candicegjing:
Sample test runs have been created under this repo: https://github.com/candicegjing/codalab-worksheets.
During retesting in early March.2020, we have started to notice the performance issue which is documented at [Python workflow runs longer now than in late Jan 2020](https://github.community/t5/GitHub-Actions/Python-workflow-runs-much-longer-now-than-in-Jan-27-2020/td-p/49428). Escalated this issue and currently it's under investigation. To track its progress, please refer to: https://github.com/actions/virtual-environments/issues/533

candicegjing:
Also tried to set up the CI build in Azure Pipeline, got a similar result: [Codala Azure Pipeline](https://dev.azure.com/jingge19/jingge19/_build/results?buildId=1&view=logs&j=19c563d8-3dcd-57d1-cd5e-9dd946b0a29b)
There seem to be other users that are facing the same situation: [Unusual-Minutes-Usage-and-Run-Durations](https://github.community/t5/GitHub-Actions/Unusual-Minutes-Usage-and-Run-Durations/m-p/50683)

epicfaace:
@candicegjing what's the status with this? It seems like you resolved the hanging build problem in https://github.com/actions/virtual-environments/issues/533?

PR #2185, opened 2020-04-19
Use GitHub Actions for CI

epicfaace:
for #2094

Based off of Jane's work in https://github.com/candicegjing/codalab-worksheets

Time: 40 minutes -> 15-20 minutes

Several improvements:
- Build images split into three parallel steps - rest-server, worker, frontend
- Tests split into around 10 parallel steps - each step runs about 4 tests, and there is a step which runs the UI tests.
- If a CLI test gives an error, it archives all Docker containers' logs so that they can be downloaded and inspected.
- If the UI test gives an error, it archives the screenshot images so that they can be downloaded and inspected.
- Additionally, the publish to pypi process has changed. Now, publishes to pypi happen on every *GitHub release* as opposed to every tag push. Effectively, this means that the release workflow has changed a bit:
  - Old workflow:
     - Wait 40 mins for Travis build on master to complete
     - Push a tag to master
     - Wait 40 mins for Travis build on master to complete, which also deploys to pypi
  - New workflow:
     - Wait for GitHub Actions build on master to complete
     - Create a new GitHub release
     - Wait for action to complete, which only deploys to pypi


epicfaace:
I've gotten this down to 15m 43s (https://github.com/codalab/codalab-worksheets/actions/runs/88264823). Perhaps with @teetone 's changes in #2173 we could speed it up though.

teetone:
@epicfaace  It won't be any faster with my changes in #2173. We decided to keep those shared file system test modules for now. I created a follow-up issue: https://github.com/codalab/codalab-worksheets/issues/2243.

epicfaace:
@percyliang this should be ready now for review. One issue is that when running `cl run` for the first time from a test, it always takes around 10 minutes for the run to get past the "staging" state (when running `cl wait`). I'm not sure if this happens on Travis, too, because Travis logs don't show timestamps.

If we fixed the above issue, we could probably reduce build time down to 10 minutes.

Nevertheless, I think the overall speed improvements from using GitHub Actions are so great that we should merge this change in and optimize further later.

monitor.py
```diff
@@ -329,7 +329,16 @@ def poll_online_workers():
                     'Uploaded file should contain the string BYTES_IN_MB, contents:\n' + cat_result,
                 )
             uuid = run_command(
-                ['cl', 'run', 'stress-test.pl:' + upload_uuid, 'perl stress-test.pl 5 10 10']
+                [
+                    'cl',
+                    'run',
+                    '--request-memory',
+                    '10m',
```
percyliang, line 336: Could we factor this out into a separate variable (e.g., `... + small_memory + ...` where `small_memory = ['--request-memory', '10m']`?

test_cli.py
```diff
@@ -135,10 +135,10 @@ def wait_until_state(uuid, expected_state, timeout_seconds=100):
                     uuid, expected_state, current_state
                 )
             )
-        time.sleep(0.5)
+        time.sleep(5)
```
percyliang: Doesn't this make things run a lot slower?

test_cli.py
```diff
@@ -207,7 +207,17 @@ def _run_command(
 ):
     # We skip using the cli directly if force_subprocess is set to true (which forces
     # us to use subprocess even for cl commands).
-    force_subprocess = not force_subprocess and args[0] == cl
+    if args[0] == cl:
+        force_subprocess = force_subprocess if args[0] == cl else True
+        # By default, request only 4m of memory and 1m of disk so that runs are faster.
```
percyliang: why does this make things faster?

test_cli.py
```diff
@@ -207,7 +207,17 @@ def _run_command(
...
+        if len(args) > 1 and args[1] == 'run' and '--request-memory' not in args:
+            args.insert(2, '4m')
+            args.insert(2, '--request-memory')
+        if len(args) > 1 and args[1] == 'run' and '--request-disk' not in args:
+            args.insert(2, '1m')
+            args.insert(2, '--request-disk')
```
percyliang: This is kind of gross...can we have `request_memory` and `request_disk` be parameters to this function so we always append the appropriate flags rather than doing this args hacking?

percyliang (commented):
It's exciting that this is basically working!  The main two comments are:
1) Can we refactor a bit so the memory flags aren't everywhere?
2) Can we figure out why there is that huge lag so we don't have to sleep as much?

test_cli.py
```diff
@@ -1113,16 +1118,9 @@ def test(ctx):
         ]
     )
     wait(remote_uuid)
-    # Since shared file system workers don't upload, exclude_patterns do not apply.
```
teetone, line 1115: @nelson-liu This test is always failing in GitHub Actions. It seems like `exclude_patterns` gets applied irregardless of whether a shared file system is used or not. Is this an acceptable behavior?

teetone:
@epicfaace GitHub Actions is passing consistently now. I ran it about 10 times and it's always passing. I also got FireFox to work again.

test_cli.py
nelson-liu: Oops, yes that is the intended behavior—the test is incorrect. I was initially trying to make the sharedfs case special, but decided against it (so, `exclude_patterns` are applied in all cases). I forgot to update the test, but things worked anyway because checking `"CODALAB_SHARED_FILE_SYSTEM"` actually doesn't do anything on travis (i.e., it always evaluates to false). Here's a diff i'd apply, sorry for the confusion!

```diff
diff --git a/test_cli.py b/test_cli.py
index 60a18ac8..7d3da5d3 100644
--- a/test_cli.py
+++ b/test_cli.py
@@ -1113,16 +1113,9 @@ def test(ctx):
         ]
     )
     wait(remote_uuid)
-    # Since shared file system workers don't upload, exclude_patterns do not apply.
-    # Verify that all files are kept if the worker is using a shared file system.
-    if os.environ.get("CODALAB_SHARED_FILE_SYSTEM"):
-        check_num_lines(
-            2 + 2 + 3, _run_command([cl, 'cat', remote_uuid])
-        )  # 2 header lines, 1 stdout file, 1 stderr file, 3 items at bundle target root
-    else:
-        check_num_lines(
-            2 + 2 + 1, _run_command([cl, 'cat', remote_uuid])
-        )  # 2 header lines, 1 stdout file, 1 stderr file, 1 item at bundle target root
+    check_num_lines(
+        2 + 2 + 1, _run_command([cl, 'cat', remote_uuid])
+    )  # 2 header lines, 1 stdout file, 1 stderr file, 1 item at bundle target root

     # Test multiple exclude_patterns
     remote_uuid = _run_command(
@@ -1136,16 +1129,9 @@ def test(ctx):
         ]
     )
     wait(remote_uuid)
-    # Since shared file system workers don't upload, exclude_patterns do not apply.
-    # Verify that all files are kept if the worker is using a shared file system.
-    if os.environ.get("CODALAB_SHARED_FILE_SYSTEM"):
-        check_num_lines(
-            2 + 2 + 3, _run_command([cl, 'cat', remote_uuid])
-        )  # 2 header lines, 1 stdout file, 1 stderr file, 3 items at bundle target root
-    else:
-        check_num_lines(
-            2 + 2 + 1, _run_command([cl, 'cat', remote_uuid])
-        )  # 2 header lines, 1 stdout file, 1 stderr file, 1 item at bundle target root
+    check_num_lines(
+        2 + 2 + 1, _run_command([cl, 'cat', remote_uuid])
+    )  # 2 header lines, 1 stdout file, 1 stderr file, 1 item at bundle target root


 @TestModule.register('read')
```

edit: oops, you already made the fixes, thanks for doing that!

teetone: @nelson-liu No problem!

epicfaace:
Let's merge this after the next release is complete.

test_cli.py
epicfaace: Otherwise, the entire run hangs and doesn't actually complete.

monitor.py
epicfaace: Not going to do that in this file, because it's only used once. Will refactor the way this works in `test_cli.py`, though.

test_cli.py
epicfaace: Done.

teetone (approved):

PR #2186, opened 2020-04-20
Move terminated bundle from worker back to staged state

### LINKED ISSUE #2183: Move terminated bundles back to staged state
Ref to the discussion in #2169

candicegjing:
Fixed #2183.

Added a new worker staged called **RESTAGED** (I am not entirely sure if this is a good name to use, please do feel free to suggest any.)

I have tested several rounds in my local machine, but still want to test more in the staging environment before this PR is ready so that I can make sure this will be working. But please feel free to suggest anything if you think it could make this PR better.  

### High-Level Overview
When a user triggers a termination signal from CLI, the worker will first check if the `pass-down-termination` signal is set to be `True`. 

#### Worker Side
   1. If not, the worker will stop processing future loops and save the existing runs as it's.
   1. If yes, the worker will process runs in the following order
         1. Collect all the unfinished bundles and move them to `RESTAGED` state 
         1. Wait until all the unfinished bundles are moved to `RESTAGED` state
               1. Stop the current run
               1. Clean up the docker container that associates with the run
               1. Clean up dependencies etc.
         1. Stop receiving runs from the server
         1. Save the worker state back to `worker-state.json` one last time before shutting down.
   
#### Server Side
1. When a bundle checks in with `RESTAGED` state, 
     1. Move the bundle state to `staged` state
     1. Remove the run entry from the worker_run table
     1. When starting any new bundle to the worker, check if it still exists.
     1. If the bundle got stuck in `STARTING` state for too long, restage it back to `staged` state. (This could happen as an edge case when the bundle-manager keeps trying to start bundle to run on a worker that is not previously terminated but terminated now. )


epicfaace (commented):
What exactly does reclaimed mean? Should we add a note to the documentation about reclaimed?

candicegjing:
> What exactly does reclaimed mean? Should we add a note to the documentation about reclaimed?

I just renamed it to be `RESTAGED` which I guess would be more clear.

epicfaace:
> I just renamed it to be `RESTAGED` which I guess would be more clear.

I'm still not sure what exactly this means. Why not keep using the "STAGED" status?

yipenghe:
What is the new stage supposed to indicate?

percyliang:
Let's not add a new state and just use `staged` if we can.

candicegjing:
> I'm still not sure what exactly this means. Why not keep using the "STAGED" status?

@epicfaace What's your understanding so far? It would be helpful if you can share your understanding and then we can identify the gap.

candicegjing:
> Let's not add a new state and just use `staged` if we can.

My intention to add a new state here is to make the logic and meaning clear. 
1. We are trying to move a bundle that has already been transferred to `staged` state back to `staged` state again --> `restaged` can help explain this and improve the readability in the same time.
1. `restage` is a term that has been used in the worker side code already.
1. I found it really confusing when I was reading/changing the code on the server-side as the `staged` state has been widely used there. 

So I personally prefer to use a state that can give the real meaning of what it represents. But if you guys don't think it's confusing to you, I'd happy to reuse the existing `staged` state.

epicfaace:
> @epicfaace What's your understanding so far? It would be helpful if you can share your understanding and then we can identify the gap.

Is "restaged" meant to be an intermediate stage in between being "staged" and going back to "staged" again? I think the name "restaged" is a bit confusing, because it sounds like it's already been "staged", when in reality, according to your description, "restaged" is an intermediate state that immediately precedes the bundle being "staged" again.

candicegjing:
> > @epicfaace What's your understanding so far? It would be helpful if you can share your understanding and then we can identify the gap.
> 
> Is "restaged" meant to be an intermediate stage in between being "staged" and going back to "staged" again? I think the name "restaged" is a bit confusing, because it sounds like it's already been "staged", when in reality, according to your description, "restaged" is an intermediate state that immediately precedes the bundle being "staged" again.

Sorry I feel I  should make the description clear. The bundle has already been moved from `staged` to `preparing`, or `running`, and need to be sent back to `staged` again. That's why I use `restaged`, as it's the second time to let the bundle in `staged` state in its life cycle.

yipenghe:
> > > @epicfaace What's your understanding so far? It would be helpful if you can share your understanding and then we can identify the gap.
> > 
> > 
> > Is "restaged" meant to be an intermediate stage in between being "staged" and going back to "staged" again? I think the name "restaged" is a bit confusing, because it sounds like it's already been "staged", when in reality, according to your description, "restaged" is an intermediate state that immediately precedes the bundle being "staged" again.
> 
> Sorry I feel I should make the description clear. The bundle has already been moved from `staged` to `preparing`, or `running`, and need to be sent back to `staged` again. That's why I use `restaged`, as it's the second time to let the bundle in `staged` state in its life cycle.

When does a bundle get restaged when it's already running? I thought most cases were that there aren't enough resources to run and got it to staged?

yipenghe:
> > > @epicfaace What's your understanding so far? It would be helpful if you can share your understanding and then we can identify the gap.
> > 
> > 
> > Is "restaged" meant to be an intermediate stage in between being "staged" and going back to "staged" again? I think the name "restaged" is a bit confusing, because it sounds like it's already been "staged", when in reality, according to your description, "restaged" is an intermediate state that immediately precedes the bundle being "staged" again.
> 
> Sorry I feel I should make the description clear. The bundle has already been moved from `staged` to `preparing`, or `running`, and need to be sent back to `staged` again. That's why I use `restaged`, as it's the second time to let the bundle in `staged` state in its life cycle.

Ah just saw the discussion in the other PR. Does restaged bundles get reruned?

candicegjing:
> > > > @epicfaace What's your understanding so far? It would be helpful if you can share your understanding and then we can identify the gap.
> > > 
> > > 
> > > Is "restaged" meant to be an intermediate stage in between being "staged" and going back to "staged" again? I think the name "restaged" is a bit confusing, because it sounds like it's already been "staged", when in reality, according to your description, "restaged" is an intermediate state that immediately precedes the bundle being "staged" again.
> > 
> > 
> > Sorry I feel I should make the description clear. The bundle has already been moved from `staged` to `preparing`, or `running`, and need to be sent back to `staged` again. That's why I use `restaged`, as it's the second time to let the bundle in `staged` state in its life cycle.
> 
> Ah just saw the discussion in the other PR. Does restaged bundles get reruned?

Yes, that's the goal. Eventually it will be reruned.

codalab/model/bundle_model.py
```diff
@@ -861,17 +867,23 @@ def transition_bundle_staged(self, bundle):
...
             if not row:
                 raise IntegrityError('Missing bundle with UUID %s' % bundle.uuid)
-            if row.state != State.STARTING:
-                # It is possible that this method is called on a bundle
-                # that has started running.
-                return False
 
-            update_message = {'state': State.STAGED, 'metadata': {'job_handle': None}}
-            self.update_bundle(bundle, update_message, connection)
+            # Reset all metadata fields to be None
```
percyliang: Can we programmatically get all the fields (where `generated=True`) from `run_bundle.py` and clear them?

codalab/model/bundle_model.py
```diff
@@ -1056,6 +1068,10 @@ def bundle_checkin(self, bundle, worker_run, user_id, worker_id):
             if not row:
                 return False
 
+            # Get staged bundle from worker checkin and move it to staged state
+            if worker_run.state == State.STAGED:
```
percyliang, line 1063: Why wasn't this needed here before?

codalab/worker/worker.py
```diff
@@ -238,10 +218,41 @@ def cleanup(self):
...
+
+    def process_terminate_signal(self):
+        # If received pass_down_termination signal from CLI to terminate the worker,
+        # wait until all the existing unfinished bundles are restaged, then stop the worker.
+        if self.terminate_and_restage:
+            if self.restage_bundles() == 0:
+                self.stop = True
+
+    def restage_bundles(self):
+        restaged_bundles = []
```
percyliang, line 253: Document the return type.

codalab/worker/worker.py
```diff
@@ -238,10 +218,41 @@ def cleanup(self):
...
+        terminal_stages = [RunStage.FINISHED, RunStage.STAGED]
+        for uuid in self.runs:
+            run_state = self.runs[uuid]
+            if run_state.stage not in terminal_stages:
+                self.restage(uuid)
+                restaged_bundles.append(uuid)
+        if len(restaged_bundles) == 0:
+            # reset the current runs to exclude bundles in terminal states: STAGED and FINISHED
+            self.runs = {
```
percyliang: Why wouldn't we do this in all cases?

codalab/worker/worker_run_state.py
```diff
@@ -475,6 +489,8 @@ def _transition_from_UPLOADING_RESULTS(self, run_state):
         If uploading and finished:
             Move to FINALIZING state
         """
+        if run_state.is_restaged:
+            return run_state._replace(stage=RunStage.STAGED)
```
percyliang: Do you need these two lines in so many places?  I would have thought you'd follow something analogous to `is_killed`.

codalab/model/bundle_model.py
candicegjing: What do you mean? I am not sure how it was done before. I am checking the `worker_run` state here to transition bundle from worker to `staged` state. Maybe I missed some context?

codalab/worker/worker.py
candicegjing: Do you mean why don't we do this in `process_runs` in the main loop?

codalab/worker/worker_run_state.py
candicegjing: Yes. That's what I originally suggested. We had a discussion with @nelson-liu two weeks ago to decide in which state a bundle can be moved from worker to `staged` state. @nelson-liu suggested allowing bundles in any states except for the final state to be able to move back to the `staged` state. And adding these two lines checking is to fulfill this requirement. But please let me know if this is not the case anymore.

codalab/worker/worker.py
candicegjing: Done

codalab/model/bundle_model.py
candicegjing: There is one issue of deleting all the fields where `generated=True`. The field `action` is marked with `generated=True`, but it's displayed in the `staged` bundle page.

percyliang: I think it's fine to clear it, since `action` is really specific to the particular execution of a bundle (e.g., kill).  And even if we wanted to preserve it, we could still programmatically get the fields ad just exclude that one.

candicegjing: Yeah, that's true. I updated this part to get all the fields with `generated=True` and exclude the `action` field for now. I guess keep the `action` field for `staged` bundle could be useful when it gets killed?

codalab/worker/worker.py
candicegjing: Looking at [your comment](https://github.com/codalab/codalab-worksheets/pull/2186/files#r421939322) now, I am wondering if you were asking "why do we exclude bundles in terminal stages (STAGED and FINISHED)", but not all other stages?

codalab/model/bundle_model.py
```diff
@@ -829,13 +829,19 @@ def transition_bundle_starting(self, bundle, user_id, worker_id):
...
         with self.engine.begin() as connection:
-            # Check that it still exists.
+            # Check if the requested bundle still exists.
             row = connection.execute(
                 cl_bundle.select().where(cl_bundle.c.id == bundle.id)
             ).fetchone()
             if not row:
                 # The user deleted the bundle.
                 return False
+            # Check if the designated worker still exists
```
percyliang: Is this necessary / related to this PR?

codalab/model/bundle_model.py
```diff
@@ -861,17 +867,19 @@ def transition_bundle_staged(self, bundle):
...
                 raise IntegrityError('Missing bundle with UUID %s' % bundle.uuid)
-            if row.state != State.STARTING:
-                # It is possible that this method is called on a bundle
-                # that has started running.
-                return False
 
-            update_message = {'state': State.STAGED, 'metadata': {'job_handle': None}}
-            self.update_bundle(bundle, update_message, connection)
+            # Reset all metadata fields that aren't input by user from RunBundle class to be None.
+            # Excluding the "action" field here as a staged bundle can be KILLED or REMOVED.
```
percyliang: Comment why `action` here - only one set by user?

codalab/worker/worker.py
```diff
@@ -238,10 +218,45 @@ def cleanup(self):
...
+        """
+        Restage bundles not in the final states [FINISHED and STAGED] from worker to server.
+        :return: the number of restaged bundles
+        """
+        restaged_bundles = []
+        terminal_stages = [RunStage.FINISHED, RunStage.STAGED]
+        for uuid in self.runs:
+            run_state = self.runs[uuid]
+            if run_state.stage not in terminal_stages:
+                self.restage(uuid)
```
percyliang: rename this `restage_bundle`

percyliang: Why do we have to restage `STAGED` bundles?

codalab/worker/worker_run_state.py
```diff
@@ -64,6 +64,13 @@ class RunStage(object):
     FINISHED = 'RUN_STAGE.FINISHED'
     WORKER_STATE_TO_SERVER_STATE[FINISHED] = State.READY
 
+    """
+    This stage will collect bundles in terminal states and 
+    sent them back to the server with STAGED state
+    """
+    STAGED = 'RUN_STAGE.STAGED'
```
percyliang: RESTAGED

codalab/model/bundle_model.py
candicegjing: The intention of adding this logic here is to handle the case when there are a large number of staged jobs waiting to be dispatched, the real dispatching logic will take longer than usual as follows https://github.com/codalab/codalab-worksheets/blob/dad99c952ecdc541cdc8b9a6f594ac2741c3086b/codalab/server/bundle_manager.py#L384-L408
However, the `WorkerInfoAccessor` object is fetched at the beginning of each scheduling loop. If there is any worker terminated in the middle of the dispatching loop,  the `WorkerInfoAccessor` object won't get refreshed. Then the worker terminated earlier will still present in the available `workers_list` and bundle manager will still try to assign jobs to this worker even though it's already terminated.

codalab/worker/worker.py
candicegjing: hmm, not sure what you mean here... I think we try to avoid restaging the `STAGED` bundles here.

codalab/model/bundle_model.py
candicegjing: I removed this one just now to reduce the complexity, coz this case can be handled by restaging bundles in the `starting` logic after timeout period (5 minutes).

candicegjing:
@percyliang I retested and rechecked those comments mentioned above. Please let me know if there is anything else needs to be modified.

codalab/model/bundle_model.py
```diff
@@ -861,17 +861,19 @@ def transition_bundle_staged(self, bundle):
...
                 raise IntegrityError('Missing bundle with UUID %s' % bundle.uuid)
-            if row.state != State.STARTING:
-                # It is possible that this method is called on a bundle
-                # that has started running.
-                return False
 
-            update_message = {'state': State.STAGED, 'metadata': {'job_handle': None}}
-            self.update_bundle(bundle, update_message, connection)
+            # Reset all metadata fields that aren't input by user from RunBundle class to be None.
+            # Excluding all the fields that can be set by users, e.g. the "action" field.
```
percyliang: "set by users, which for now is just the `action` field"

codalab/worker/worker.py
```diff
@@ -239,12 +217,50 @@ def cleanup(self):
         logger.info("Stopped Worker. Exiting")
 
     def signal(self):
-        # When the terminate flag is False, set the stop flag to stop running
-        # the worker without changing the status of existing running bundles.
-        if not self.terminate:
+        """
+        When the pass_down_termination flag is False, set the stop flag to stop running
+        the worker without changing the status of existing running bundles. Otherwise,
+        restage all bundles that are not in the terminal states [FINISHED, RESTAGED].
```
percyliang: more accurate to say "set a flag to restage..."

codalab/worker/worker.py
```diff
@@ -239,12 +217,50 @@ def cleanup(self):
...
+
+    def check_termination(self):
+        """
+        If received pass_down_termination signal from CLI to terminate the worker, wait until
+        all the existing unfinished bundles are restaged, reset runs, then stop the worker.
+        """
+        if self.terminate_and_restage:
+            if self.restage_bundles() == 0:
+                # Stop the worker
+                self.stop = True
```
percyliang: Should we rename `stop` to `terminate`?  "terminate" and "stop" are synonymous, right?  Just want to canonicalize terminology.

percyliang (approved):
Looks good!  Only some comments about comments and naming.

codalab/worker/worker.py
candicegjing: That's true. I just renamed `self.stop` to `self.terminate`.

codalab/model/bundle_model.py
candicegjing: Updated.

codalab/worker/worker.py
candicegjing: Updated.

PR #2189, opened 2020-04-21
Add flag to cl download to force overwriting of downloads

nelson-liu:
This is especially pertinent when you're doing `cl down <bundlename>`, and the contents of `<bundlename>` might be changing on the worksheet / you might be on a different worksheet...

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1277,6 +1277,12 @@ class BundleCLI(object):
                 help='Path to download bundle to.  By default, the bundle or subpath name in the current directory is used.',
             ),
             Commands.Argument(
+                '-f',
+                '--force',
+                action='store_true',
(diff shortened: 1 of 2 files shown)


nelson-liu:
An alternative solution would be to always overwrite when `--output-path` is provided (à la `wget`), but I think `--force` is a bit more explicit and intentional.

codalab/lib/bundle_cli.py
```diff
@@ -1276,6 +1276,12 @@ def do_upload_command(self, args):
                 '--output-path',
                 help='Path to download bundle to.  By default, the bundle or subpath name in the current directory is used.',
             ),
+            Commands.Argument(
+                '-f',
+                '--force',
```
percyliang, line 1281: Would be semantically more accurate to call this `--overwrite`, but alas, `-o` is taken...

percyliang (approved):
Cool, thanks!

PR #2190, opened 2020-04-21
Fix additem to subworksheets_block and table_block by giving them sort keys

### LINKED ISSUE #2227: Can't add items right after a bundle table, or after a bundle row, or after search results
Clicking on a bundle row and then clicking "add text" does not add anything after it.

epicfaace:
fixes #2227.

- Fix additem to subworksheets_block and table_block by giving them sort keys. Otherwise, we wouldn't be able to add items through the UI when selecting a subworksheets_block or table_block.
- When clicking "Edit Source" from one of the rows in a search results block, go to the line of the corresponding search directive.

- Also, allow adding items right after a bundle row. It looks like this:

image

However, the previous functionality only works if the bundle table is not the result of a search.
- if the bundle table is a result of a search, clicking on “add new text” adds text after the table
- if the bundle table is just a collection of bundles, clicking on “add new text” adds text after the selected bundle table entry


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1034,6 +1034,11 @@ class Worksheet extends React.Component {
                     rawIndex = this.state.ws.info.block_to_raw[focusIndexPair];
                 }
 
+                if (rawIndex === undefined) {
```
epicfaace, line 1061:
```suggestion
                // When clicking "Edit Source" from one of the rows in a search results block, go to the line of the corresponding search directive.
                if (rawIndex === undefined) {
```

yipenghe (approved):

yipenghe (changes requested):
Actually, we need sort keys for bundle rows as well, otherwise we wouldn't be able to insert within a table block. Currently the approach works if we only need to insert after a whole existing blcok

yipenghe (approved):
Seems like I don't need to sort key to complete the copy, and I can't use add-items unless all bundle rows have a sort key. The fix is still needed for adding text fix

frontend/src/util/worksheet_utils.js
```diff
@@ -302,41 +302,20 @@ export function createHandleRedirectFn(worksheetUuid) {
...
 }
 
-export function getMinMaxKeys(item) {
-    if (!item) {
-        return { minKey: null, maxKey: null };
-    }
-    let minKey = null;
-    let maxKey = null;
+// Return the sort key at index subFocusIndex, if subFocusIndex is defined.
+// Otherwise, return the largest sort_key.
```
yipenghe, line 306: The item here sounds more like a block (a bundle table block more than a bundle row)?

epicfaace: Sure, but we use the `item` terminology all throughout the frontend code base -- probably should refactor this in a separate PR.

PR #2193, opened 2020-04-22
use npm ci

### LINKED ISSUE #2172: Use `npm ci` instead of `npm install` on Dockerfile / Travis CI
This will give us more predictable builds as it will make sure that `package-lock.json` is completely updated.

Otherwise, when we install new dependencies, there is no way for the CI to ensure that `package-lock.json` is updated at the same time and checked in as well.

epicfaace:
fixes #2172. npm ci is faster -- and it also makes sure that our package-lock.json is actually updated whenever we install new dependencies. Turns out our package-lock.json was quite out of date (because it was usually only updated on the Docker volume and thus _not_ checked into source control), so I've added those changes here too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jobs:
       language: node_js
       node_js:
         - 13
-      install: cd frontend && npm install
+      install: cd frontend && npm ci
       script:
         - npm run check-ci
     - stage: check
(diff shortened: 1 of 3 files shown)


percyliang (approved):
Sounds great!  Is there any reason ever to use `install` or is `ci` just a strictly better version?

epicfaace:
> Sounds great! Is there any reason ever to use `install` or is `ci` just a strictly better version?

`install` should be used during development when adding a new package -- this will update `package-lock.json` to match the new dependencies.

`ci` _only_ uses `package-lock.json` to install dependencies (and gives an error if `package.json` and `package-lock.json` are out of sync), so it's meant to be used _after_ `package-lock.json` is updated. So it should always be used during CI.

PR #2199, opened 2020-04-22
Fix tailing cl wait output

### LINKED ISSUE #2195: cl wait -t option does not work
The `-t` option in `cl wait` does not work. Perhaps a remnant of the Python 2 ->3 migration?

```
epicfaace@Ashwins-MacBook-Pro codalab-worksheets % cl run "ping -c 1 google.com" --request-network
0xc6cdfd922bf646aabfa0b5538040a544
epicfaace@Ashwins-MacBook-Pro codalab-worksheets % cl wait 0xc6cdfd922bf646aabfa0b5538040a544 -t
  File "/Users/epicfaace/Library/Python/3.7/bin/cl", line 10, in <module>
    sys.exit(main())
  File "/Users/epicfaace/Library/Python/3.7/lib/python/site-packages/codalab/bin/cl.py", line 10, in main
    cli.do_command(sys.argv[1:])
  File "/Users/epicfaace/Library/Python/3.7/lib/python/site-packages/codalab/lib/bundle_cli.py", line 864, in do_command
  File "/Users/epicfaace/Library/Python/3.7/lib/python/site-packages/codalab/lib/bundle_cli.py", line 858, in <lambda>
  File "/Users/epicfaace/Library/Python/3.7/lib/python/site-packages/codalab/lib/bundle_cli.py", line 2439, in do_wait_command
    state = self.follow_targets(client, target.bundle_uuid, subpaths)
  File "/Users/epicfaace/Library/Python/3.7/lib/python/site-packages/codalab/lib/bundle_cli.py", line 2510, in follow_targets
    self.stdout.write(result)
TypeError: write() argument must be str, not bytes
```

candicegjing:
Fixed #2195 

Tested locally and I was able to see the output from a running bundle with this fix.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2516,7 +2516,7 @@ class BundleCLI(object):
                     if not result:
                         break
                     subpath_offset[i] += len(result)
-                    self.stdout.write(result)
+                    self.stdout.write(ensure_str(result))
                     if len(result) < READ_LENGTH:
                         # No more to read.
                         break


percyliang (approved):
Maybe we can add something to `test_cli.py`?

epicfaace (approved):

codalab/lib/bundle_cli.py
```diff
@@ -2516,7 +2516,7 @@ def follow_targets(self, client, bundle_uuid, subpaths, from_start=False):
                     if not result:
                         break
                     subpath_offset[i] += len(result)
-                    self.stdout.write(result)
+                    self.stdout.write(ensure_str(result))
```
epicfaace, line 2519: Is `result` only sometimes a string and sometimes binary?

if that's not the case, then we shouldn't use `ensure_str` and we should instead do the necessary encoding / decoding directly.

candicegjing: In this case, I can only reproduce `bytes` during the experiments that I was running (not sure if there will be other cases that will happen in the future). 
I agree that we should do necessary encoding or decoding directly. 
The only thing that makes me decide to use `ensure_str` function is that I learned it from a past task that there could be `UnicodeDecodeError` happening during the `decode` phase, and `ensure_str` contains the error handling when `UnicodeDecodeError` happens. I think we can reuse this piece of code even though it has one additional checking on if the current `result` is a string, instead of rewriting duplicate code to handle the exception. 
If we think we want to handle decoding directly, I am open to that.

Issue #2191, opened 2020-04-21
continuous-integration/travis-ci stuck on Expected — Waiting for status to be reported


yshen-google:
It seems continuous-integration/travis-ci is stuck on "Expected — Waiting for status to be reported" without any update. Shall we follow https://blog.travis-ci.com/2018-09-27-deprecating-github-commit-status-api-for-github-apps-managed-repositories and https://github.community/t5/GitHub-Actions/GitHub-Actions-Pull-requests-always-include-some-outdated-checks/m-p/31444#M794 to resolve this issue?

epicfaace:
This appears to be working for certain PRs, but not for others. Are you sure it's due to the problem linked above?

yshen-google:
Hi Ashwin, I am not completely sure it's due to the problem in the links, though they are the top 2 results I found. Yeah, I agree with you. Based on the observation, it seems some PRs have this problem but not others.

nelson-liu:
Seems like the PRs coming in from forks aren't being run on travis, while the PRs created from branches of `codalab/codalab-worksheets` are being run fine.

yshen-google:
Yup, were there any changes made to travis-ci? It seems in the past (not long time ago), travis-ci could still run on PRs originated from a forked repo.

nelson-liu:
can someone with write access to the repo go to https://travis-ci.org/github/codalab/codalab-worksheets/settings and make sure the "Build pushed pull requests" button is on?

<img width="1439" alt="Screen Shot 2020-04-22 at 6 42 52 PM" src="https://user-images.githubusercontent.com/7272031/80049971-3f173580-84c9-11ea-8641-a043f544e89f.png">

candicegjing:
I just updated the setting as follows:
<img width="935" alt="Screen Shot 2020-04-24 at 10 20 42 AM" src="https://user-images.githubusercontent.com/1483372/80239540-464b5a00-8615-11ea-89ca-2d9a654fbf76.png">
@yshen-google  and @nelson-liu Is this what we want?

percyliang:
No longer using travis.

PR #2200, opened 2020-04-23
Add more information on default limit number of search results

### LINKED ISSUE #1803: Provide more information about search results

yshen-google:
Due to #2191, it seems travis-ci is not run on PRs from forked repo. Thus, create this PR from branch on the current repo.

1. Print out the default limit on search results
2. Print out no results found if so


codalab/lib/bundle_cli.py
```diff
@@ -2040,6 +2046,14 @@ def get(i, info, col):
                 else:
                     return info.get(col, nested_dict_get(info, 'metadata', col))
 
+            if len(bundle_info_list) == bundle_model.SEARCH_RESULTS_LIMIT:
+                print(
+                    'Only {} results are shown. Use .limit=N to show N.'.format(
```
percyliang: "show the first N"

codalab/lib/bundle_cli.py
```diff
@@ -3184,18 +3198,28 @@ def do_wls_command(self, args):
             'worksheets',
             params={'keywords': args.keywords, 'include': ['owner', 'group_permissions']},
         )
+
         if args.uuid_only:
             for row in worksheet_dicts:
                 print(row['uuid'], file=self.stdout)
         else:
             if worksheet_dicts:
+                if len(worksheet_dicts) == bundle_model.SEARCH_RESULTS_LIMIT:
```
percyliang: Could you factor this block out to avoid code duplication?

codalab/lib/bundle_cli.py
yshen-google: Done.

codalab/lib/bundle_cli.py
yshen-google: Sounds good.

percyliang (approved):
Great, thanks!

candicegjing (approved):
Looks great to me!

PR #2201, opened 2020-04-23
Add to travis check steps to validate rest and cli docs are generated before pr submission

armantajback:
…


epicfaace:
@armantajback can you change this PR so it uses github actions instead?

epicfaace:
This PR is getting old so just closing this for now... we should definitely add something like this later though!

PR #2204, opened 2020-04-23
focus issue fixed

### LINKED ISSUE #2170: Focusing issue with the upload button
When you click on the upload button in the header, it opens up the upload menu. However, it auto focuses on the first menu item even if I'm not hovering on the menu. 
Screenshot from 2020-04-15 23-52-49
When my mouse hovers around the second upload item, it adds shades to the item, but the first menu still has a blue "selected" color, which is confusing

Screenshot from 2020-04-15 23-55-45

The ideal behaviour would be to not auto-select the first menu item after clicking

Levilian:
fix: #2170


yipenghe:
Currently if I don't use the keyboard, it still shows that the first menu item is selected. 

I think the ideal behaviour is if I first click upload button / click a + u to open, nothing is shown as selected. When I hover / hit down button, then the first button is shown as selected. Not sure if this was what we wanted? @percyliang @epicfaace 
Screenshot from 2020-04-29 14-48-55

yipenghe (approved):
Nice

PR #2218, opened 2020-04-29
Add apex to default-gpu docker image

robinjia:
I spent some time today trying to get [apex](https://github.com/NVIDIA/apex), an NVIDIA library for 16-bit floating point optimization in pytorch, into a docker image that ran on codalab. I got it working in my own docker image, and thought it could be useful to other people since apex is useful and not installable on top of the current default-gpu image. Summary of my changes:
- Base image is now `cuda:9.2-cudnn7-devel` rather than `cuda:9.0-cudnn7-runtime`. The most important change is to `-devel` which includes `nvcc` unlike the `-runtime` image. `nvcc` is needed to install apex (https://github.com/NVIDIA/apex/blob/master/examples/docker). 
- Bump to 9.2 is because recent versions of pytorch are only built with cuda 9.2 or 10+, and 10+ imposes restrictions on what hardware you can run on, making 9.2 the best choice. Installing apex requires pytorch to be built with the same version of cuda as you have installed. This is also why the pytorch installation command has been changed, to force 9.2 since 10.2 is the default. You could alternatively keep things at cuda 9.0 and use pytorch 1.1.0 with cudatoolkit=9.0.
- The changes to the order in which I install spacy is due to an unrelated issue having to do with how numpy no longer supports python 2.7.
- Finally, at the end I install apex. Note that I can only support 6.0, 6.1, 6.2, and 7.0 GPU architectures and not the latest Turing 7.5 because that requires cuda 10. For the NLP cluster this is only an issue for the RTX's


percyliang:
Cool!  Are there any backward incompatibilities (i.e., if we make this the default-gpu image, will anything stop working)?  Can you also quantify the benefits of Apex?  Is it mostly that you can now use larger models?

robinjia:
I don't believe there will be backwards compatibility issues. Yes apex lets you use larger models (basically you get 2x GPU RAM) and also larger batch sizes for any model to speed up training.

docker/dockerfiles/Dockerfile.default-gpu
```diff
@@ -92,6 +91,7 @@ RUN python3 -m pip install -U \
       scikit-image \
       nltk
 
+RUN pip2 install spacy
```
percyliang, line 94: Would you mind updating the cpu docker image to have the same lines?  Want to make sure the two images stay in sync for the prefix.

percyliang (approved):
Thanks!

Issue #2212, opened 2020-04-27
Add test cases to exercise wadd across instances


candicegjing:
Add test cases to exercise wadd across instances once we can start two CodaLab instances together for testing.

PR #2221, opened 2020-04-30
Fix wadd empty bundles across intances

### LINKED ISSUE #2219: Empty bundle got stuck in UPLOADING state when copy across instances
Empty bundle (bundle without any data but only metadata) got stuck in UPLOADING state when copy across instances using `cl wadd`. Refer: https://github.com/codalab/codalab-worksheets/issues/2210#issuecomment-620135221

Screen Shot 2020-04-29 at 11 07 19 AM

candicegjing:
Fixed #2219 

When coping over a bundle without contents (metadata only) across instances, the bundle will land in the destination worksheet with `UPLOADING` state without any further updates. This PR is to fix this issue.

Added an additional parameter `state_on_success` to indicate whether the current bundle has content or not. Then in the rest function, we check for this flag when we set the status of the current bundle. If this flag is detected to be not None, then we set the bundle state with the value it passed in. Otherwise, it will remain as it is, either `UPLOADING` or `CREATED`.

One thing I don't feel confident is that `create_bundles()` is to create multiple bundles at a time. Then `state_on_success` should tie to each bundle instead of all bundles, although this won't affect the functionality (right now we process one bundle a time, so this will still work but might not look reasonable to some degree).

If we don't handle the state update in `create_bundles()`, we might want to create a new API to process this. (There is an existing [update function](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/bundles.py#L305-L333). The `state` field is excluded from updating through the update function). I am open to suggestions.


codalab/lib/bundle_cli.py
```diff
@@ -1452,13 +1459,10 @@ def copy_bundle(
                 'worksheet': dest_worksheet_uuid,
                 'detached': not add_to_worksheet,
                 'wait_for_upload': True,
+                'state_on_success': None if has_contents else source_info['state'],
```
candicegjing: added an additional parameter `state_on_success` to indicate whether the current bundle has contents or not.

codalab/lib/bundle_cli.py
```diff
@@ -1469,14 +1473,6 @@ def copy_bundle(
         else:
             unpack = False
 
-        # Bundles stuck in non-final states such as 'running' should not keep
```
nelson-liu, line 1472: I'm a bit confused—what is the behavior of `wadd` across instances for bundles in non-final states, after this PR?

codalab/lib/bundle_cli.py
```diff
@@ -1444,6 +1444,13 @@ def copy_bundle(
 
         print("Copying %s..." % source_desc, file=self.stdout)
 
+        # If bundle contents don't exist, finish after just copying metadata
+        try:
```
percyliang, line 1448: Bundles that have been `rm -d` just have their `data_hash` missing, so I think it's semantically more accurate (and faster) to just check that (look in the `metadata`).

codalab/lib/bundle_cli.py
percyliang: What do we want the behavior to be?

candicegjing: @nelson-liu I think currently we don't copy bundles that are not in final states (READY or FAILED), see these lines [L1437-L1443](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L1437-L1443).

nelson-liu: oh oops, sorry for missing that! Yeah, I think that it's fine to just omit bundles that aren't in final states.

candicegjing: @nelson-liu no worries! I found the logic here a bit confusing as well when I was playing around. So I removed them. BTW, Bundles in `killed` state won't be copied over.

codalab/lib/bundle_cli.py
teetone: It seems like later we are just checking if `state_on_success` exists and if it does, we set the bundle's state to be the value of `state_on_success`. If that's the case, would it make more sense to rename `state_on_success` to something like `predetermined_state`?

codalab/rest/bundles.py
```diff
@@ -253,7 +259,9 @@ def _create_bundles():
         created_uuids.append(bundle_uuid)
         bundle_class = get_bundle_subclass(bundle['bundle_type'])
         bundle['owner_id'] = request.user.user_id
-        if issubclass(bundle_class, UploadedBundle) or query_get_bool('wait_for_upload', False):
+        if final_state:
+            bundle['state'] = final_state
+        elif issubclass(bundle_class, UploadedBundle) or query_get_bool('wait_for_upload', False):
             bundle['state'] = State.UPLOADING
```
teetone, line 258: From your PR description, would the bundle without any content hit this line and be stuck in uploading?

teetone:
I think we should add test coverage for this later after I get this PR merged in: https://github.com/codalab/codalab-worksheets/pull/2173

codalab/lib/bundle_cli.py
percyliang: +1 to `predetermined_state`

candicegjing: I was trying to follow the existing style to pass the bundle state information: https://github.com/codalab/codalab-worksheets/blob/02c9840744e5cfc88498f626d9f9fc888509f93f/codalab/lib/bundle_cli.py#L1491. But sure, we can change it to `predetermined_state` as they don't seem to be completely overlapped.

codalab/rest/bundles.py
candicegjing: That's the original issue we are trying to fix. Basically, for the future, when a user `wadd` bundle without any content, the logic will first hit `final_state` check and then set the bundle state to be whatever the final state is and skip setting it to be `UPLOADING`. This will make sure the bundle state changes to final states instead of getting stuck at `UPLOADING` state

candicegjing:
> I think we should add test coverage for this later after I get this PR merged in: #2173

Yes. I created a ticket to track this issue : https://github.com/codalab/codalab-worksheets/issues/2212

percyliang:
Stepping back, I'm not sure I understand something: why do bundles without data end up in the `uploading` state?  Is that a bug?  I would have expected the copied bundle to be in whatever state the source bundle was in regardless of whether there are contents.

Also, to confirm, we we're only copying over bundles in final states (`ready`, `failed`, `killed`)?

candicegjing:
> Stepping back, I'm not sure I understand something: why do bundles without data end up in the `uploading` state? Is that a bug? I would have expected the copied bundle to be in whatever state the source bundle was in regardless of whether there are contents.

1. Yes. Bundles without data end up in the uploading state **is a bug**. I guess it's originally intended to be reused for the bundle without contents and the bundle with contents. So when copy the metadata to another instance,  it makes perfect sense to set the bundle with `uploading` states. However, this `uploading` state never gets changed for the bundle without contents. That's what was discovered in the original ticket #2219 
1. Yes, we are only copying bundles in final states.

percyliang:
Could we follow the same code paths regardless of whether the bundle has data or not...It might be more consistent to have the same logic flip the state from `uploading` to `ready` in both cases rather than having separate logic to tweak the states.

candicegjing:
> Could we follow the same code paths regardless of whether the bundle has data or not...It might be more consistent to have the same logic flip the state from `uploading` to `ready` in both cases rather than having separate logic to tweak the states.

I had a similar thought at the beginning... Now, I guess sharing the same code path would be easier to read. I just updated the logic to not differentiate when uploading bundles with or without content.

codalab/lib/bundle_cli.py
```diff
@@ -1455,43 +1455,50 @@ def copy_bundle(
             },
         )
 
-        # If bundle contents don't exist, finish after just copying metadata
+        # Collect information about how server should unpack
+        filename = nested_dict_get(source_info, 'metadata', 'name')
```
percyliang: Is there a reason to move this up here?

codalab/lib/bundle_cli.py
```diff
@@ -1455,43 +1455,50 @@ def copy_bundle(
...
+                )
         else:
-            source_state = State.FAILED
-
-        # Send file over
-        progress = FileTransferProgress('Copied ', f=self.stderr)
-        source = source_client.fetch_contents_blob(BundleTarget(source_bundle_uuid, ''))
-        with closing(source), progress:
+            # Update the bundle state with final states
             dest_client.upload_contents_blob(
```
percyliang, line 1498: It looks like we're already calling `upload_contents_blob` above...what is this doing here?

candicegjing: I split them into two parts is because `source` is `None` when the bundle doesn't have any content. When bundle has content, `source` is a NOT `None` value, which can be closed. Another reason is that the parameters sent to both cases are different. e.g. `filename`, `unpack`, `simplify` are not needed when calling `upload_contents_blob`.

codalab/lib/bundle_cli.py
candicegjing: Oh, actually, this shouldn't be here. I might have moved it around during testing/debuging.

codalab/lib/bundle_cli.py
```diff
@@ -1455,43 +1455,50 @@ def copy_bundle(
             },
         )
 
-        # If bundle contents don't exist, finish after just copying metadata
+        # Fetch bundle contents from source client
```
percyliang: This is not fetching bundle contents, but just the metadata

codalab/lib/bundle_cli.py
```diff
@@ -1455,43 +1455,50 @@ def copy_bundle(
...
             unpack = True
         else:
             unpack = False
 
-        # Bundles stuck in non-final states such as 'running' should not keep
-        # that state at the destination server, and should instead just fallback
-        # to 'failed'
-        if source_info['state'] == State.READY:
-            source_state = State.READY
+        try:
```
percyliang, line 1475: Can we just check if `target_info is None` to avoid making a failing request?

percyliang (approved):
This is easier to follow now - thanks!  Could you add a test in `test_cli.py`?  I guess we need a test for `wadd` anyway...

codalab/lib/bundle_cli.py
candicegjing: Updated.

codalab/lib/bundle_cli.py
candicegjing: Updated.

candicegjing:
> This is easier to follow now - thanks! Could you add a test in `test_cli.py`? I guess we need a test for `wadd` anyway...

Yes. This will be tracked through https://github.com/codalab/codalab-worksheets/issues/2212.

PR #2231, opened 2020-04-30
Update default-cpu Dockerfile to match default-gpu

robinjia:
As per https://github.com/codalab/codalab-worksheets/pull/2218#discussion_r418121986


percyliang (approved):

docker/dockerfiles/Dockerfile.default-cpu
```diff
@@ -94,6 +93,7 @@ RUN python3 -m pip install -U \
       scikit-image \
       nltk
 
+RUN pip2 install spacy
```
percyliang, line 97: Maybe add a comment on why this is moved down.

PR #2236, opened 2020-05-01
fix-time-quota-calculation

### LINKED ISSUE #2229: Out of time quota seems to be miscalculated
On `cl upload`, I get the following error:

```Preparing upload archive...
UsageError: Out of time quota: 1y142d / 60d0h (845.6%)
tar: Write error```

This is preventing me from uploading anything new. I have a handful of bundles, but there is no way the total run time is over a year.

### PR DESCRIPTION
**Pull request by candicegjing**
Fixed #2229 

### Overview of the Root Cause
In our scheduling logic, the `acknowledge_recently_finished_bundles` which moves bundles from `FINALIZING` state to `FINISHED` state, happens after the dispatching function `_schedule_run_bundles_on_workers`:
 https://github.com/codalab/codalab-worksheets/blob/d09bcc962d14c69ccbff4233493e3101f90db48b/codalab/server/bundle_manager.py#L650-L675
Then when there are a significant number of jobs in `STAGED` state, `_schedule_run_bundles_on_workers` will take longer time than usual, and this increases the waiting time for jobs in `FINALIZING` state to be moved to `FINISHED` state. When a job stays longer in `FINALIZING` state, the job is still alive in a designated worker, then it keeps checking in to the rest-server with `FINALIZING` state. The server will keep calling `transition_bundle_finalizing` function, which contains the logic to increment the amount of time the user used, then the user's time quota gets added by `n` times (`n` will depend on waiting time and checkin frequency, so no one knows how many times exactly)
https://github.com/codalab/codalab-worksheets/blob/d09bcc962d14c69ccbff4233493e3101f90db48b/codalab/model/bundle_model.py#L967-L991

Currently, I moved the time quota update function to `transition_bundle_finished`, so this can ensure we add up the current time usage for one time only. This might not be the best solution. Ideally, moving a bundle from finalizing to the finished state shouldn't be blocked in this way. After all, there aren't too many things left in the `finalizing` state. (We might want to change the current state transition to allow non-blocking transition from finalizing to finished). But let me know if anyone has any suggestions.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -964,7 +964,7 @@ class BundleModel(object):
             self.update_bundle(bundle, bundle_update, connection)
         return True
 
-    def transition_bundle_finalizing(self, bundle, user_id, worker_run, connection):
+    def transition_bundle_finalizing(self, bundle, worker_run, connection):
         """
         Transitions bundle to FINALIZING state:
             Saves the failure message and exit code from the worker
@@ -1060,7 +1060,7 @@ class BundleModel(object):
                 self.transition_bundle_running(
                     bundle, worker_run, row, user_id, worker_id, connection
                 )
-                return self.transition_bundle_finalizing(bundle, user_id, worker_run, connection)
+                return self.transition_bundle_finalizing(bundle, worker_run, connection)
 
             if worker_run.state in [State.PREPARING, State.RUNNING]:
                 return self.transition_bundle_running(
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -984,10 +984,6 @@ class BundleModel(object):
         bundle_update = {'state': State.FINALIZING, 'metadata': metadata}
 
         self.update_bundle(bundle, bundle_update, connection)
-
-        if user_id == self.root_user_id:
-            self.increment_user_time_used(bundle.owner_id, getattr(bundle.metadata, 'time', 0))
-
         return True
 
     def transition_bundle_finished(self, bundle, bundle_location):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: codalab/model/bundle_model.py [hunk 1]
```diff
@@ -984,10 +984,6 @@ def transition_bundle_finalizing(self, bundle, user_id, worker_run, connection):
         bundle_update = {'state': State.FINALIZING, 'metadata': metadata}
 
         self.update_bundle(bundle, bundle_update, connection)
-
-        if user_id == self.root_user_id:
```
candicegjing, line 988: Not exactly sure why we compare the `user_id == self.root_user_id`? During my testing, I saw they were equal all the time. Maybe I missed some non-trivial details here?

percyliang: With user-owned workers, is it different?

codalab/model/bundle_model.py
```diff
@@ -1003,6 +999,9 @@ def transition_bundle_finished(self, bundle, bundle_location):
         if failure_message == 'Kill requested':
             state = State.KILLED
 
+        # Increment the amount of time used for the current user
+        self.increment_user_time_used(bundle.owner_id, metadata.get('time', 0))
```
percyliang, line 1003: Why is this moved here?

codalab/model/bundle_model.py
percyliang: I think this is here to only charge users for non user-owned workers.  This is important, probably should add a comment.

candicegjing: I tested using user-owned worker, but still always got them equal. If we want to avoid adding time spent on user-owned workers, we need to find another way.

candicegjing: Yes. Time used in user-owned workers should be excluded from this calculation. I have updated the code to increment time used only on public workers

percyliang (approved):
Thanks!

PR #2239, opened 2020-05-03
Prevent creating ._ files when zipping for upload

### LINKED ISSUE #2209: Get rid of ._ files when uploading from a Mac
Get rid of `._` files  when using the tar library with `COPYFILE_DISABLE=1`.

teetone:
Fixes #2209 

Prevent creating `._` files when zipping with the BSD tar library and always ignore `._` file  if there are any already existing.


codalab/worker/file_util.py
```diff
@@ -28,19 +41,24 @@ def tar_gzip_directory(
                       the directory structure are excluded.
     ignore_file: Name of the file where exclusion patterns are read from.
     """
-    # Always ignore entries specified by the ignore file (e.g. .gitignore)
     args = ['tar', 'czf', '-', '-C', directory_path]
+
+    # If the BSD tar library is being used, prepend COPYFILE_DISABLE=1 to prevent creating ._* files
+    if 'bsdtar' in get_tar_version_output():
+        args.insert(0, 'COPYFILE_DISABLE=1')
```
percyliang: Why not just include this always?

teetone: I originally always included it, but the command errors with GNU tar library.

codalab/worker/file_util.py
```diff
@@ -11,7 +11,20 @@
 from codalab.common import BINARY_PLACEHOLDER
 
 NONE_PLACEHOLDER = '<none>'
-GIT_PATTERN = '.git'
+
+# Patterns to always ignore when zipping up directories
+ALWAYS_IGNORE_PATTERNS = ['.git', '._*']
```
percyliang: Do we need to do something to ignore `__MACOSX`?

codalab/worker/file_util.py
percyliang: Also, if we're already ignoreing `._*` based on `ALWAYS_IGNORE_PATTERN`, then why do we need this?

codalab/worker/file_util.py
teetone: I actually don't get any `__MACOSX` folders when zipping, but I will add it to this list and write a test for it too.

codalab/worker/file_util.py
teetone: Initially I thought just adding it to the list will work, but it seems like `._` files get generated after it excludes files by patterns.

codalab/worker/file_util.py
teetone: For some reason, I can reproduce getting `._` files, but not the `__MACOSX` folder. I read some posts online and it seems like calling `bsdtar` with `--exclude __MACOSX`  should work.

codalab/worker/file_util.py
percyliang: In that case, we shouldn't include `._*` in the `ALWAYS_IGNORE_PATTERN`, right?

Hm, running `tar` (on Linux) with `COPYFILE_DISABLE=1` works for me - it just seems to be a no-op...

candicegjing (approved):
The change looks good to me. However, I wasn't able to reproduce those additional files `._` when I zip files on MacOS. Does this happen to a certain MacOS version?

teetone:
> The change looks good to me. However, I wasn't able to reproduce those additional files `._` when I zip files on MacOS. Does this happen to a certain MacOS version?

@candicegjing I'm not sure if this a problem with a specific MacOS version. I'm on Catalina, which version are you using by the way?

codalab/worker/file_util.py
teetone: I added `._*` just in case Mac users had those files lingering around. For some reason, it errored out on the Linux machine I was using. I will try it again.

candicegjing:
> @candicegjing I'm not sure if this a problem with a specific MacOS version. I'm on Catalina, which version are you using by the way?

I am also on Catalina (version 10.15.1). Interestingly, I don't see any of those files created when I zip files.

codalab/worker/file_util.py
```diff
@@ -28,19 +41,24 @@ def tar_gzip_directory(
                       the directory structure are excluded.
     ignore_file: Name of the file where exclusion patterns are read from.
     """
-    # Always ignore entries specified by the ignore file (e.g. .gitignore)
     args = ['tar', 'czf', '-', '-C', directory_path]
+
+    # If the BSD tar library is being used, append --disable-copy to prevent creating ._* files
+    if 'bsdtar' in get_tar_version_output():
+        args.append('--disable-copyfile')
```
teetone, line 48: It turns out the executable needs to be the first argument. I instead appended `--disable-copyfile` for BSD tar to get rid of the `._` files.

percyliang (approved):

PR #2240, opened 2020-05-04
Add padding for table standard cells

### LINKED ISSUE #1796: table widths are too narrow
https://worksheets.codalab.org/worksheets/0xa590fd1b68944a1a95c1c40c4931dc7b
Screenshot from 2019-12-01 20-03-06

In particular, the spacing between `command` and `state` is really small (more so than in the old UI).

yshen-google:
Fixes #1796

Add padding for table standard cells to avoid too little space between columns.

Before the change:
little_space_between_command_and_state

After the change:
after_change

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.scss
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.scss
@@ -204,9 +204,6 @@
     position: relative;
     /*margin-bottom: 20px;*/
   }
-  .worksheet_content .ws-item table > tbody > tr > td {
-    padding-right: 8px !important;
-  }
   .worksheet_content .ws-item .ws-checkbox {
     margin: 0;
     position: absolute;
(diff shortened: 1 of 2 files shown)


epicfaace (approved):
Thanks!

epicfaace (dismissed):
Actually, can you make this change in this line here? https://github.com/codalab/codalab-worksheets/blob/51bce7541be5ac827a2b16b9a4b46eb778f3c234/frontend/src/components/worksheets/items/TableItem/BundleRow.js#L385-L390

It appears that `rootNoPad` is used for bundle rows in two places, so making the change in the section above will fix the formatting issues for both those locations.

yshen-google:
> Actually, can you make this change in this line here?
> 
> https://github.com/codalab/codalab-worksheets/blob/51bce7541be5ac827a2b16b9a4b46eb778f3c234/frontend/src/components/worksheets/items/TableItem/BundleRow.js#L385-L390
> 
> It appears that `rootNoPad` is used for bundle rows in two places, so making the change in the section above will fix the formatting issues for both those locations.

Done. Looks like in this way, I need to adjust the padding to half (4px). With 5px, it's even too much:
too_much

yshen-google:
Could someone take a look at the changes post-comment? Thanks. @percyliang @epicfaace

yipenghe (approved):
Looks good to me

PR #2241, opened 2020-05-04
Fix broken cl edit function

### LINKED ISSUE #2237: [USER ISSUE] cl edit doesn't work
**Describe the bug**
When I try to use `cl edit` from the command line I get an exception

**Codalab Information**
- *Your username*: robinjia
- *UUID of worksheet where problem occurred*: 0x23acc0bd2aa044fd9300f73d3a5fbbb9
- *UUID of problem bundle (please don't delete)*: n/a

```
  File "/home/robinjia/.local/bin/cl", line 8, in <module>
    sys.exit(main())
  File "/home/robinjia/.local/lib/python3.6/site-packages/codalab/bin/cl.py", line 10, in main
    cli.do_command(sys.argv[1:])
  File "/home/robinjia/.local/lib/python3.6/site-packages/codalab/lib/bundle_cli.py", line 864, in do_command
  File "/home/robinjia/.local/lib/python3.6/site-packages/codalab/lib/bundle_cli.py", line 858, in <lambda>
  File "/home/robinjia/.local/lib/python3.6/site-packages/codalab/lib/bundle_cli.py", line 1708, in do_edit_command
    bundle_subclass, info['metadata']
  File "/home/robinjia/.local/lib/python3.6/site-packages/codalab/lib/metadata_util.py", line 73, in request_missing_metadata
    return parse_metadata_form(bundle_subclass, form_result)
  File "/home/robinjia/.local/lib/python3.6/site-packages/codalab/lib/metadata_util.py", line 85, in parse_metadata_form
    if line != '' and not line.startswith('//'):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

 - *OS (if on CLI)*: Ubuntu 18.04
 - *Codalab CLI Version (if on CLI)*: 0.5.13
 - *Browser (if on Web)*;

**To Reproduce**
1. `cl edit ^1` or it seems any bundle (mine opens vim)
2. Save and close

**Expected behavior**
Bundle info should update with no exception thrown

candicegjing:
Fixed #2237 

This PR will address the following two issues:
1. The original problem of `cl edit` reported in #2237 
1. Another problem is introduced from an old PR which added the functionality to delete `staged_status` using `None` value. Using `None` value to indicate deletion operation seems to be conflicting with existing features (e.g. `cl edit`). I added a new parameter to indicate the deletion operation for now. But I am open to any other suggestions.

### DIFF
--- a/codalab/lib/metadata_util.py
+++ b/codalab/lib/metadata_util.py
@@ -81,7 +81,7 @@ def parse_metadata_form(bundle_subclass, form_result):
     metadata_types = {spec.key: spec.type for spec in metadata_specs}
     result = {}
     for line in form_result:
-        line = line.strip().encode('UTF-8')
+        line = line.strip()
         if line != '' and not line.startswith('//'):
             if ':' not in line:
                 # TODO: don't delete everything; go back to the editor and show the error message
@@ -94,13 +94,14 @@ def parse_metadata_form(bundle_subclass, form_result):
             if metadata_key not in metadata_types:
                 raise UsageError('Unexpected metadata key: %s' % (metadata_key,))
             metadata_type = metadata_types[metadata_key]
+
             if metadata_type == list:
-                if any(unicode_util.contains_unicode(v) for v in result[metadata_key]):
+                remainders = remainder.split() if remainder else []
+                if any(unicode_util.contains_unicode(r) for r in remainders):
                     raise UsageError(
-                        'Metadata cannot contain unicode: %s = %s'
-                        % (metadata_key, result[metadata_key])
+                        'Metadata cannot contain unicode: %s = %s' % (metadata_key, remainder)
                     )
-                result[metadata_key] = remainder.split() if remainder else []
+                result[metadata_key] = remainders
             elif metadata_type == str:
                 if remainder is not None and unicode_util.contains_unicode(remainder):
                     raise UsageError(
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1089,7 +1089,7 @@ class BundleModel(object):
             self.do_multirow_insert(connection, cl_bundle_metadata, metadata_values)
             bundle.id = result.lastrowid
 
-    def update_bundle(self, bundle, update, connection=None):
+    def update_bundle(self, bundle, update, connection=None, delete=False):
         """
         For each key-value pair in the update dictionary, add or update key-value pair. Note
         that metadata keys not in the update dictionary are not affected in the update operation.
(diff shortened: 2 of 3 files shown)


codalab/lib/metadata_util.py
```diff
@@ -94,13 +94,14 @@ def parse_metadata_form(bundle_subclass, form_result):
...
-                if any(unicode_util.contains_unicode(v) for v in result[metadata_key]):
+                remainders = remainder.split() if remainder else []
+                if any(unicode_util.contains_unicode(r) for r in remainders):
                     raise UsageError(
-                        'Metadata cannot contain unicode: %s = %s'
-                        % (metadata_key, result[metadata_key])
+                        'Metadata cannot contain unicode: %s = %s' % (metadata_key, remainder)
                     )
-                result[metadata_key] = remainder.split() if remainder else []
+                result[metadata_key] = remainders
```
candicegjing, line 104: Not exactly sure if I have made the right changes here. From the context, it looks like we want to check for each element of the remainder to see if any of it contains Unicode character. But let me know if this is not the case.

codalab/server/bundle_manager.py
```diff
@@ -826,7 +826,9 @@ def _get_staged_bundles_to_run(self, workers, user_info_cache):
                     # Remove the uuid from self._bundles_without_matched_workers if a matched
                     # private worker is found in the system and update bundle's metadata
                     if bundle.uuid in self._bundles_without_matched_workers:
-                        self._model.update_bundle(bundle, {'metadata': {'staged_status': None}})
+                        self._model.update_bundle(
```
percyliang, line 851: Why is this related to `cl edit`?  Should be in a different PR?

candicegjing: `cl edit` doesn't work properly with the above string decoding fix. When a user types `cl edit <uuid>` to edit metadata fields in the pop-up window, if any of the fields are empty, CLI will just return a metadata key is missing as follows, as the following field has value `None`, which was treated as a deletion operation in update function.
```
stanford@Stanfords-MacBook-Pro codalab-worksheets % cl edit 0x075eaa877bf84ea68072995b19891474  
UsageError: Missing metadata key: allow_failed_dependencies
```

codalab/model/bundle_model.py
```diff
@@ -1106,7 +1106,8 @@ def update_bundle(self, bundle, update, connection=None):
         # Generate a list of metadata keys that will be deleted and udpate metadata key-value pair
         metadata_delete_keys = []
         for key, value in metadata_update.items():
-            if value is None:
+            # Delete the key,value pair when the delete flag is set to True and value is None
+            if delete and value is None:
```
percyliang, line 1112: We are changing the default behavior from `delete=True` to `delete=False`...are there any other calls to `update_bundle` that we need to update?

percyliang (approved):

codalab/model/bundle_model.py
candicegjing: No... `delete` operation is only introduced when we were trying to add the metadata field `staged_status`, as this field will disappear once the bundle successfully gets assigned to a worker. We didn't have this before, so no other functions or features are affected.

percyliang: The previous code is equivalent to `delete=True`, but now the default is `delete=False`.

PR #2246, opened 2020-05-06
Allow async loading of file contents

### LINKED ISSUE #2244: Async loading of file contents
This worksheet (https://worksheets.codalab.org/worksheets/0x55ed347e3beb4fa6971d3a226e48fc92) takes 15-16 seconds to load. Continuation of #1234.

See https://github.com/codalab/codalab-worksheets/pull/1963#pullrequestreview-355528703 for more context on the steps (this is step 4 mentioned in @percyliang 's comment)

epicfaace:
## Description
fixes #2244 

When `brief=1` is passed to the `/rest/interpret` endpoint, file contents are not resolved (i.e., `interpret_genpath_table_contents` is not called).

Then, unresolved columns look like this:

image

Async requests are then sent to `/rest/interpret/genpath-table-contents`, and the table contents are then resolved.

## Testing

1. Upload a folder with a file called `package.json` with the contents `{"version": 2}`
1. Add the following in the worksheet (replace 0xc30ee021930f4bb3a9671753d2a92094 with the bundle id)

```
% schema cub
% add uuid uuid '[0:8]'
% add name
% add output_value /package.json:version
% add time time duration
% add state
% add description

% display table cub
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
[dataset bundle]{0xc30ee021930f4bb3a9671753d2a92094}
```

1. The worksheet should now async load file contents.


codalab/rest/interpret.py
```diff
@@ -480,6 +497,8 @@ def interpret_genpath_table_contents(contents):
             # value can be either a string (already rendered) or a (bundle_uuid, genpath, post) triple
             if is_bundle_genpath_triple(value):
                 requests.append(value)
+    if dry_run:
+        return requests
```
percyliang: Hm, the fact that this function returns two types of things is non-ideal...Is there a way to avoid this?

epicfaace: We could make a separate function called `contains_bundle_genpath_triple`, but then we'd have to duplicate some logic -- do you think it's worth it to do that to avoid the problem you mentioned?

yipenghe: Maybe extract the requests generation out to a separate function, and call it in this function (replacing line 493-499), and call it directly where the dry_run is supposed to be true?

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -251,4 +254,46 @@ const styles = (theme) => ({
 
 const TableContainer = withStyles(styles)(_TableContainer);
 
-export default TableItem;
+async function fetchAsyncData({ contents }) {
```
yipenghe: Since this function is specific to the TableItem, it might be better to specify what data we are fetching in the name

yipenghe (approved):
The test description should be modified as upload a folder (not file) containing the package.json file. Also change the fetchAysncData to be more specific about what data might be better

PR #2247, opened 2020-05-06
Enable bundle selection on non-write access worksheets

yipenghe:
This will allow users to copy bundles directly from dashboard or other users' worksheets


yipenghe:
Opening another one based on master

PR #2250, opened 2020-05-07
Enable bundle selection for copy on all worksheets

### LINKED ISSUE #2158: should still show checkboxes even if don't have write access
Currently, if you're on a worksheet that you don't have write access to, checkboxes don't show up on the bundles.  But we still want to show the checkboxes because one might want to still copy bundles off of it.

yipenghe:
Fix #2158 , this will allow users to copy bundles from dashboard or other users' worksheets


epicfaace (approved):
Looks good!

PR #2257, opened 2020-05-11
Respect bundle priority by precomputing available workers during scheduling

### LINKED ISSUE #2256: Possible race condition in codalab bundle manager
**Describe the bug**

Sometimes, I see runs that don't have `request_priority` set landing on nodes, while almost-identical (metadata is the same, the only thing that differs is a single dependency) bundles that _have_ `request_priority` set are not run. I have a feeling that this might be due to a race condition in https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L367-L386 , but I'm not entirely sure.

Basically, I think the issue lies in the fact that worker resources for the current user / the public workers are updated on-demand when the bundle is about to be run. I could imagine a scenario where, in the middle of looping through all the staged run bundles (which are already sorted into the order that they should be run), a codalab worker frees up. So, the code thinks that the worker can take the bundle, and starts the bundle on the worker. However, there are bundles that were earlier in `staged_bundles_to_run` that have now being skipped over as a result.

**Expected behavior**
The ordering of the bundles as stored in `staged_bundles_to_run` is always respected.

nelson-liu:
Tries to fix #2256 . I don't really like having to manually keep track of / update `user_parallel_run_quota_left`, suggestions on how to do it better are welcome.

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -399,7 +399,7 @@ class BundleManager(object):
                     # decrement the parallel run quota left.
                     if worker["user_id"] == self._model.root_user_id:
                         user_parallel_run_quota_left[bundle.owner_id] -= 1
-                    # Update available worker resoures (this is an upper-bound),
+                    # Update available worker resoures. This is a lower-bound,
                     # since resources released by jobs that finish are not used until
                     # the next call to _schedule_run_bundles_on_workers.
                     worker['cpus'] -= bundle_resources.cpus
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -399,6 +399,12 @@ class BundleManager(object):
                     # decrement the parallel run quota left.
                     if worker["user_id"] == self._model.root_user_id:
                         user_parallel_run_quota_left[bundle.owner_id] -= 1
+                    # Update available worker resoures (this is an upper-bound),
+                    # since resources released by jobs that finish are not used until
+                    # the next call to _schedule_run_bundles_on_workers.
+                    worker['cpus'] -= bundle_resources.cpus
+                    worker['gpus'] -= bundle_resources.gpus
+                    worker['memory_bytes'] -= bundle_resources.memory
                     break
 
     def _deduct_worker_resources(self, workers_list, running_bundles_info):
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -363,26 +363,42 @@ class BundleManager(object):
         # Build a dictionary which maps from uuid to running bundle and bundle_resources
         running_bundles_info = self._get_running_bundles_info(workers, staged_bundles_to_run)
 
+        # We pre-compute the workers available to each user (and the codalab-owned workers),
+        # such that workers that come online or regain the necessary resources while we
+        # are attempting to run each staged bundle will respect the ordering of
+        # staged_bundles_to_run (i.e., they won't be used immediately, and will be instead
+        # assigned bundles on the next run of _run_iteration).
+        resource_deducted_user_owned_workers = {}
+        user_parallel_run_quota_left = {}
+        for user in user_queue_positions.keys():
+            resource_deducted_user_owned_workers[user] = self._deduct_worker_resources(
+                workers.user_owned_workers(user), running_bundles_info
+            )
+            user_parallel_run_quota_left[user] = self._model.get_user_parallel_run_quota_left(
+                user, user_info_cache[user]
+            )
+        resource_deducted_codalab_owned_workers = self._deduct_worker_resources(
+            workers.user_owned_workers(self._model.root_user_id), running_bundles_info
+        )
+
         # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
-            # Check if there is enough parallel run quota left for this user
-            if (
-                self._model.get_user_parallel_run_quota_left(
-                    bundle.owner_id, user_info_cache[bundle.owner_id]
+            if user_parallel_run_quota_left[bundle.owner_id] > 0:
+                workers_list = (
+                    resource_deducted_user_owned_workers[bundle.owner_id]
+                    + resource_deducted_codalab_owned_workers
                 )
-                <= 0
-            ):
-                # Don't include CodaLab-owned workers, as there is no parallel_run_quota left for this user.
-                codalab_owned_workers = []
             else:
-                codalab_owned_workers = workers.user_owned_workers(self._model.root_user_id)
-            user_owned_workers = workers.user_owned_workers(bundle.owner_id)
-            workers_list = user_owned_workers + codalab_owned_workers
-            workers_list = self._deduct_worker_resources(workers_list, running_bundles_info)
+                workers_list = resource_deducted_user_owned_workers[bundle.owner_id]
+
             workers_list = self._filter_and_sort_workers(workers_list, bundle, bundle_resources)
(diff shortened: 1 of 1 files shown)


candicegjing:
@nelson-liu Thanks for reporting and fixing this issue. I did some checks. Here is my understanding: the code fetches `workers` object from the database at the beginning of each scheduling cycle. https://github.com/codalab/codalab-worksheets/blob/7eecce209e6ab817e2476cb090d0e59311a0f045/codalab/server/bundle_manager.py#L663-L675
The `workers` object will then be updated within each scheduling cycle depending on the number of jobs that are successfully scheduled to run on available workers. If there is any worker gets freed up in the middle of the scheduling cycle, it needs to wait until the next scheduling cycle to be added to the available resource pool. So I am a little bit hesitant to conclude that this part of scheduling logic is the root cause of this issue. 
There are two possibilities that I could think of as of now:  
1. One possibility that I could think of is that this could happen if I increased your user parallel run quota in the middle of the scheduling cycle. Given that I don't have the full context when this happened, this might not be the real case. 
1. Another thing is that we don't know what jobs were running on those hosts when this happened. It could be that the resources that meet the requirement of **bundle_1** didn't get free up until the scheduling cycle hits **bundle_2** (bundle_1's priority is higher than bundle_2). 

I guess if we could have more information about what was happening during the issue, we could have more confidence in deciding the root cause.

nelson-liu:
> One possibility that I could think of is that this could happen if I increased your user parallel run quota in the middle of the scheduling cycle. Given that I don't have the full context of when this happened, this might not be the real case.

This is not the case. But isn't the available parallel run quota calculated dynamically with `self._model.get_user_parallel_run_quota_left`? So, if a public worker frees up during the iteration, then the result of `self._model.get_user_parallel_run_quota_left` could go from 0 to 1?

> Another thing is that we don't know what jobs were running on those hosts when this happened. It could be that the resources that meet the requirement of bundle_1 didn't get free up until the scheduling cycle hits bundle_2 (bundle_1's priority is higher than bundle_2).

I'm not sure how this is different than "If there is any worker gets freed up in the middle of the scheduling cycle, it needs to wait until the next scheduling cycle to be added to the available resource pool.". Maybe I'm misunderstanding, but these two statements seem to contradict each other...

candicegjing:
> I'm not sure how this is different than "If there is any worker gets freed up in the middle of the scheduling cycle, it needs to wait until the next scheduling cycle to be added to the available resource pool.". Maybe I'm misunderstanding, but these two statements seem to contradict each other...

1. Yes, the available parallel run quota is calculated dynamically for each bundle. But it's not directly related to the number of available workers. Basically freeing up one more worker doesn't necessarily increase your parallel run quota. 
1. Oh, I think I made a mistake here. I was brainstorming about several things here and I probably mixed them together. I guess what I wanted to say is that suppose bundle_1 should start earlier than bundle_2, there could be other reasons that lead to this situation, e.g. the connection between the server and an available worker wasn't established successfully or the worker went offline or something else. See this piece of logic:
https://github.com/codalab/codalab-worksheets/blob/7eecce209e6ab817e2476cb090d0e59311a0f045/codalab/server/bundle_manager.py#L494-L516
Even though **bundle_1** should start earlier than **bundle_2**, but it's not guaranteed to be like that.

nelson-liu:
> 1. Yes, the available parallel run quota is calculated dynamically for each bundle. But it's not directly related to the number of available workers. Basically freeing up one more worker doesn't necessarily increase your parallel run quota.

I agree with "freeing up one more worker doesn't necessarily increase your parallel run quota". But note that this function calculates the _available parallel quota *left*_. See https://github.com/codalab/codalab-worksheets/blob/9fd42142d2c3dedb1e91a1ee2fab8a75ac605d17/codalab/model/bundle_model.py#L2469-L2481

So, if a job ends on a public worker, the number should absolutely increment, right?

> I guess what I wanted to say is that suppose bundle_1 should start earlier than bundle_2, there could be other reasons that lead to this situation, e.g. the connection between the server and an available worker wasn't established successfully or the worker went offline or something else. See this piece of logic:

I see, yeah that's a case that's hard to avoid. In this case, there weren't any network issues, though...

candicegjing:
> So, if a job ends on a public worker, the number should absolutely increment, right?

That's correct. When a bundle lands in a public instance, parallel run quota will increase.

nelson-liu:
Right, so if it increases in the middle of the iteration, wouldn't that mean that it might run a bundle when another had a higher priority?

nelson-liu:
Also, I didn't notice this problem when I only had a few staged jobs...but it seems to happen fairly consistently now. Probably not coincidentally, I also launched a bunch of codalab bundles over the weekend (I have 1K+ staged bundles at this point), so I'd imagine that the loop that goes through all of the bundles takes quite awhile / there's a higher probability of a worker accepting a non-prioritized bundle.

candicegjing:
> Also, I didn't notice this problem when I only had a few staged jobs...but it seems to happen fairly consistently now. Probably not coincidentally, I also launched a bunch of codalab bundles over the weekend (I have 1K+ staged bundles at this point), so I'd imagine that the loop that goes through all of the bundles takes quite awhile / there's a higher probability of a worker accepting a non-prioritized bundle.

I realized that I need to revise one reply above which sounds misleading: when a bundle lands in a public worker, parallel run quota will **DECREASE**.

nelson-liu:
I think i'm a bit confused at this point—do you think this PR solves the issue or not?

codalab/server/bundle_manager.py
```diff
@@ -363,26 +363,42 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
+        for user in user_queue_positions.keys():
+            resource_deducted_user_owned_workers[user] = self._deduct_worker_resources(
+                workers.user_owned_workers(user), running_bundles_info
+            )
+            user_parallel_run_quota_left[user] = self._model.get_user_parallel_run_quota_left(
+                user, user_info_cache[user]
+            )
+        resource_deducted_codalab_owned_workers = self._deduct_worker_resources(
+            workers.user_owned_workers(self._model.root_user_id), running_bundles_info
+        )
```
candicegjing, line 382: I think the only issue here is that the function `_deduct_worker_resources` happens before the real dispatching logic. Then when it calculates the resources left on user_owned_worker or public_worker, it always operates on the same set of `run_uuids`. With more and more bundles successfully getting dispatched, this may lead to task over-assigning issue, as `run_uuids` now has more values than before.

nelson-liu: oops you're totally right—this needs to be updated in the loop as well, but only to deduct more resources that are taken.

nelson-liu: i pushed a rather manual fix to this issue. There's some code duplication between here and `_deduct_bundle_resources`, not sure if we can do anything cleaner.

candicegjing: Thanks very much! I think this will work. 
In terms of refactoring (it doesn't have to be done in this PR), I was wondering if we can do the following:
1. fetch users' parallel_run_quota once and store them in the dictionary `user_parallel_run_quota_left` that maps owner_id to the number of parallel_run_quota left
1. dispatch bundles as usual except that we use `user_parallel_run_quota_left[owner_id]` now instead of querying the database to check if public workers can be included
1. subtract `user_parallel_run_quota_left[owner_id]` by 1 for each successfully dispatched bundle.

I think this way we can avoid resource deduction in two places.

candicegjing (approved):
Thanks for fixing this issue!

PR #2262, opened 2020-05-12
Convert only empty string to a shell-escaped version of the string

### LINKED ISSUE #2261: Command submitted from rest cli end point failed due to shlex grouping
Command submitted from rest CLI endpoint can be interfered by `shlex` quote grouping. 
For instance, when a user submits a request from CLI endpoint passing a command:
```
run --- 'date;sleep 10'
```
Since the logic in rest CLI endpoint split command by whitespace:
https://github.com/codalab/codalab-worksheets/blob/962189207c363bc26c76b5d3a6d765d687d8c70e/codalab/rest/cli.py#L133-L136
The above example will be split into:
```
["'date;sleep'", '10']
```
Rather than 
```
['date;sleep', '10']
```

`["'date;sleep'", '10']` will cause failure as it's not a valid command.

candicegjing:
Fixed #2261 

Not sure if this is the optimal solution. But changing only the empty string to a shell-escaped version of string work for both use cases when user submit command from CLI and from rest cli endpoint: https://github.com/codalab/codalab-worksheets/blob/962189207c363bc26c76b5d3a6d765d687d8c70e/codalab/rest/cli.py#L18-L20

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -815,8 +815,8 @@ class BundleCLI(object):
         """
         try:
             i = argv.index('---')
-            # Convert the command after '---' to a shell-escaped version of the string.
-            shell_escaped_command = [quote(x) for x in argv[i + 1 :]]
+            # Convert empty string after '---' to a shell-escaped version of the string.
+            shell_escaped_command = [quote(x) if not x else x for x in argv[i + 1 :]]
             argv = argv[0:i] + [' '.join(shell_escaped_command)]
         except:
             pass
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -815,8 +814,8 @@ def collapse_bare_command(argv):
         """
         try:
             i = argv.index('---')
-            # Convert the command after '---' to a shell-escaped version of the string.
-            shell_escaped_command = [quote(x) for x in argv[i + 1 :]]
+            # Convert empty string after '---' to a shell-escaped version of the string.
+            shell_escaped_command = [shlex.quote(x) if not x else x for x in argv[i + 1 :]]
```
percyliang, line 818: Why are we not quoting anymore?  If I do `--- echo "Hello world"`, "Hello world" still needs to be quoted.

candicegjing: That's true

PR #2264, opened 2020-05-12
Making command parsing robust

### LINKED ISSUE #2261: Command submitted from rest cli end point failed due to shlex grouping

candicegjing:
Fixed #2261 

The changes made in this PR will fix the command string parsing issue discovered in #2261 when a user submits a command via rest API endpoint: https://github.com/codalab/codalab-worksheets/blob/962189207c363bc26c76b5d3a6d765d687d8c70e/codalab/rest/cli.py#L18-L20

Basically, the exiting command parsing from CLI is not compatible with parsing from rest api. This PR aims to fix this.
 
Things changed in this PR: 
1. in rest api call: I removed the `whitespace_split=True`.  There are two types of chars in command: punctuation chars and whitespace. Punctuation chars should be parsed in the first place and then whitespace. Based on [shlex.whitespace_split](https://docs.python.org/3.4/library/shlex.html#shlex.shlex.whitespace_split), when it sets to True, the command string will just be split by whitespace but not punctuation chars, which can be error-prone.
1. To get the list of default punctuation chars from shlex, I used a dummy command to initialize a shlex object. This is because of punctuation chars defined as a property to the class. If there are better ways to do this, please let me know.

Example of use cases from **rest api endpoint** :
1. `cl run --- date;sleep 10`
```
>>> command = 'cl run --- date;sleep 10'
>>> lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
>>> lexer.escapedquotes = '\'"'
>>> list(lexer)
['cl', 'run', '---', 'date', ';', 'sleep', '10']
```
2. `cl run --- date; sleep 10`
```
>>> command = 'cl run --- date; sleep 10'
>>> lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
>>> lexer.escapedquotes = '\'"'
>>> list(lexer)
['cl', 'run', '---', 'date', ';', 'sleep', '10']
```
3. `cl run --- python myscript.py`
```
>>> command = 'cl run --- python myscript.py'
>>> lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
>>> lexer.escapedquotes = '\'"'
>>> list(lexer)
['cl', 'run', '---', 'python', 'myscript.py']
```
**Usage from CLI will remain the same.**

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -29,7 +29,6 @@ import json
 from collections import defaultdict
 from contextlib import closing
 from io import BytesIO
-from shlex import quote
 
 import argcomplete
 from argcomplete.completers import FilesCompleter, ChoicesCompleter
@@ -156,6 +155,7 @@ HEADING_LEVEL_2 = '## '
 HEADING_LEVEL_3 = '### '
 
 NO_RESULTS_FOUND = 'No results found'
+COMMAND_PLACEHOLDER = '<cmd>'
 
 
 class CodaLabArgumentParser(argparse.ArgumentParser):
@@ -815,8 +815,15 @@ class BundleCLI(object):
         """
         try:
             i = argv.index('---')
-            # Convert the command after '---' to a shell-escaped version of the string.
-            shell_escaped_command = [quote(x) for x in argv[i + 1 :]]
+            # This is to get the default list of punctuation chars from shlex. When punctuation_chars = True,
+            # the class property punctuation_chars is set as '();<>|&'. Otherwise, an empty string will be returned.
+            punctuation_chars = list(
+                shlex.shlex(COMMAND_PLACEHOLDER, punctuation_chars=True).punctuation_chars
+            )
+            # Convert string that is not a punctuation char after '---' to a shell-escaped version of the string.
+            shell_escaped_command = [
+                shlex.quote(x) if x not in punctuation_chars else x for x in argv[i + 1 :]
+            ]
             argv = argv[0:i] + [' '.join(shell_escaped_command)]
         except:
             pass
--- a/codalab/rest/cli.py
+++ b/codalab/rest/cli.py
@@ -126,13 +126,12 @@ def general_command(worksheet_uuid, command):
     if isinstance(command, str):
         # shlex throws ValueError on incorrectly formatted commands
         try:
-            # see https://docs.python.org/2/library/shlex.html#shlex.shlex.escapedquotes
+            # see https://docs.python.org/3.4/library/shlex.html#shlex.shlex.escapedquotes
             # By default, the double quote can be escaped. By setting the
             # escapedquotes property, we are able to escape single quotes as well
             # examples: run '\''
-            lexer = shlex.shlex(command, posix=True)
+            lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
             lexer.escapedquotes = '\'"'
-            lexer.whitespace_split = True
             args = list(lexer)
         except ValueError as e:
             raise UsageError(str(e))
--- a/tests/lib/bundle_cli_test.py
+++ b/tests/lib/bundle_cli_test.py
@@ -26,3 +26,13 @@ class BundleCliTest(unittest.TestCase):
         expected_result = ['cl', 'run', "echo 'hello world!'"]
         actual_result = self.bundle_cli.collapse_bare_command(argv)
         self.assertEqual(actual_result, expected_result)
+
+    def test_collapse_bare_command_punctuation_char_semicolon(self):
+        argv = ['cl', 'run', '---', 'date', ';', 'sleep', '10']
+        expected_result = ['cl', 'run', 'date ; sleep 10']
+        actual_result = self.bundle_cli.collapse_bare_command(argv)
+        self.assertEqual(actual_result, expected_result)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/tests/lib/bundle_cli_test.py
+++ b/tests/lib/bundle_cli_test.py
@@ -32,7 +32,3 @@ class BundleCliTest(unittest.TestCase):
         expected_result = ['cl', 'run', 'date ; sleep 10']
         actual_result = self.bundle_cli.collapse_bare_command(argv)
         self.assertEqual(actual_result, expected_result)
-
-
-if __name__ == '__main__':
-    unittest.main()


codalab/rest/cli.py
```diff
@@ -126,13 +126,12 @@ def general_command(worksheet_uuid, command):
...
         try:
-            # see https://docs.python.org/2/library/shlex.html#shlex.shlex.escapedquotes
+            # see https://docs.python.org/3.4/library/shlex.html#shlex.shlex.escapedquotes
             # By default, the double quote can be escaped. By setting the
             # escapedquotes property, we are able to escape single quotes as well
             # examples: run '\''
-            lexer = shlex.shlex(command, posix=True)
+            lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
             lexer.escapedquotes = '\'"'
-            lexer.whitespace_split = True
```
candicegjing, line 135: I removed the whitespace_split=True here due to the following reason: there are two types of chars in command: punctuation chars (command separator) and whitespace. Punctuation chars should be parsed in the first place and then whitespace. Based on [shlex.whitespace_split](https://docs.python.org/3.4/library/shlex.html#shlex.shlex.whitespace_split), when it sets to True, the command string will just be split by whitespace but not punctuation chars, which can be error-prone.

codalab/lib/bundle_cli.py
```diff
@@ -815,8 +815,15 @@ def collapse_bare_command(argv):
         """
         try:
             i = argv.index('---')
-            # Convert the command after '---' to a shell-escaped version of the string.
-            shell_escaped_command = [quote(x) for x in argv[i + 1 :]]
+            # This is to get the default list of punctuation chars from shlex. When punctuation_chars = True,
+            # the class property punctuation_chars is set as '();<>|&'. Otherwise, an empty string will be returned.
+            punctuation_chars = list(
+                shlex.shlex(COMMAND_PLACEHOLDER, punctuation_chars=True).punctuation_chars
```
bkgoksel, line 821: Can't you initialize a `shlex.shlex` object with an empty string instead of a placeholder?

tests/lib/bundle_cli_test.py
```diff
@@ -26,3 +26,9 @@ def test_collapse_bare_command_non_empty_str_args_with_escaped_char(self):
         expected_result = ['cl', 'run', "echo 'hello world!'"]
         actual_result = self.bundle_cli.collapse_bare_command(argv)
         self.assertEqual(actual_result, expected_result)
+
+    def test_collapse_bare_command_punctuation_char_semicolon(self):
+        argv = ['cl', 'run', '---', 'date', ';', 'sleep', '10']
+        expected_result = ['cl', 'run', 'date ; sleep 10']
```
bkgoksel, line 32: Have you tested the alternatives
`cl run --- date; sleep 10` and `cl run --- date;sleep 10` ?

codalab/rest/cli.py
bkgoksel: what does this do if the command has other punctuation chars besides `;`? Will this split `python my_script.py` into `python`, `my_script`, `.`, `py`? Is that a problem?

bkgoksel (commented):
Added some questions since I'm not super familiar with how shlex works and how things feed into the rest of codalab code, and wanted to clarify a few things.

codalab/lib/bundle_cli.py
candicegjing: Yes. We can initialize `shlex.shlex` with an empty string. I was trying to keep it this placeholder variable consistent with other placeholder variables, e.g. https://github.com/codalab/codalab-worksheets/blob/962189207c363bc26c76b5d3a6d765d687d8c70e/codalab/common.py#L12
But we can certainly change it to an empty string.

codalab/rest/cli.py
candicegjing: This is the example that I should have added to the original description:
```
>>> import shlex
>>> command = 'cl run --- python myscript.py'
>>> lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
>>> lexer.escapedquotes = '\'"'
>>> list(lexer)
['cl', 'run', '---', 'python', 'myscript.py']
```

tests/lib/bundle_cli_test.py
candicegjing: This use case will be handled via rest api side. 
```
stanford@Stanfords-MacBook-Pro deployment % python3
Python 3.7.7 (default, Mar 10 2020, 15:43:33) 
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import shlex
>>> command = 'cl run --- date;sleep 10'
>>> lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
>>> lexer.escapedquotes = '\'"'
>>> list(lexer)
['cl', 'run', '---', 'date', ';', 'sleep', '10']
>>> command = 'cl run --- date; sleep 10'
>>> lexer = shlex.shlex(command, posix=True, punctuation_chars=True)
>>> lexer.escapedquotes = '\'"'
>>> list(lexer)
['cl', 'run', '---', 'date', ';', 'sleep', '10']
```

codalab/lib/bundle_cli.py
bkgoksel: It's a stylistic choice, whatever you prefer is fine. I was just curious.

codalab/rest/cli.py
bkgoksel: ah I see. Sounds good then.

bkgoksel (approved):
lgtm

codalab/lib/bundle_cli.py
percyliang: I think empty string would be better (add comment saying any string is fine) since we don't use the placeholder anywhere else.

codalab/lib/bundle_cli.py
```diff
@@ -815,8 +815,15 @@ def collapse_bare_command(argv):
...
+            )
+            # Convert string that is not a punctuation char after '---' to a shell-escaped version of the string.
+            shell_escaped_command = [
+                shlex.quote(x) if x not in punctuation_chars else x for x in argv[i + 1 :]
```
percyliang, line 825: why are we ignoring only punctuation characters by themselves?  Couldn't they show up as part of an argument (e.g., "hello;")?  I think I'm missing what problem we're trying to solve...  I was thinking that every `x` should just be quoted.

candicegjing: For example, we are trying to run `cl run --- date; sleep 10`, in `sys.argv`, the parameters will be changed to `argv = ['run', 'date', ';', 'sleep', '10']`. If we don't leave the punctuation characters out and `shlex.quote` on every single subcommand, we will get the new `argv` as follows:
```
>>> argv = ['run', 'date', ';', 'sleep', '10']
>>> shell_excaped_argv = [shlex.quote(x) for x in argv]
>>> shell_excaped_argv
['run', 'date', "';'", 'sleep', '10']
```
Note that there is an additional pair of quote wrapped around semicolon. Then when it gets executed on real system, we get the following result:
Screen Shot 2020-05-12 at 7 17 18 PM

Basically, shell won't be able to recognize semicolon as a command separator if it wrapped with single quotes.

percyliang: I think if you're using `---`, you should expect each subsequent argument to be treated as actual arguments in a command.  If you want a shell command (e.g., with `;`, `&`, etc.), you should just quote the entire command and not use `---`.  Otherwise I think the logic gets too complicated.  Is there a reason that this wouldn't work?

candicegjing: I think it's definitely better if users can properly quote their commands. This change primarily come up due to the issue @kerem discovered when he was using `cli/command` endpoint to submit shell command. It seems like if a user doesn't quote his/her command properly, whitespace splitting may end up grouping argument in the wrong way. 
If we want to keep the way as it is, maybe we could add some comments about properly quoting all the arguments?

percyliang: What's the exact use case that triggered a problem?  Is this from the web terminal?

candicegjing: If I am not wrong, he submitted his bundles using the rest api call (`cli/command`) in his code. @bkgoksel please correct me if I am wrong.

candicegjing: @percyliang I think the real issue here is that the command parsing rules from rest API (`cli/command`) endpoint is different from CLI. For instance,
1. In CLI, `cl run --- date; sleep 10` will be interpreted as `cl run --- date`. `sleep 10` will be ignored after the command separator (semicolon).
1. In rest API `cli/command` endpoint, when I submit a bundle from the web interface using the following patterns: 
    1. `cl run --- date; sleep 10` will be interpreted as `'date;' sleep 10`.
    1. `cl run --- date;sleep 10` (no whitespace after semicolon) will be interpreted as `'date;sleep' 10`
    1. `cl run --- 'date; sleep 10'` will be interpreted as `'date; sleep 10'`
    All of the above cases can't be executed correctly.

candicegjing:
Close this PR as @bkgoksel verified that this is not need from his side any more.

PR #2265, opened 2020-05-13
2212: Test that bundles are ready after performing wadd across instances

### LINKED ISSUE #2212: Add test cases to exercise wadd across instances

teetone:
Resolves #2212 

Modified existing `copy` test to ensure that bundles are in `ready` states after performing `wadd`  across instances.

### DIFF
--- a/test_cli.py
+++ b/test_cli.py
@@ -109,15 +109,17 @@ def get_info(uuid, key):
     return _run_command([cl, 'info', '-f', key, uuid])
 
 
-def wait_until_running(uuid, timeout_seconds=100):
+def wait_until_state(uuid, expected_state='running', timeout_seconds=100):
     start_time = time.time()
     while True:
-        if time.time() - start_time > 100:
+        if time.time() - start_time > timeout_seconds:
             raise AssertionError('timeout while waiting for %s to run' % uuid)
         state = get_info(uuid, 'state')
-        # Break when running or one of the final states
-        if state in {'running', 'ready', 'failed'}:
-            assert state == 'running', "waiting for 'running' state, but got '%s'" % state
+        # Break when running or one of the final states or expected_state
+        if state in {'running', 'ready', 'failed', expected_state}:
+            assert state == expected_state, "waiting for '()' state, but got '{}'".format(
+                expected_state, state
+            )
             return
         time.sleep(0.5)
 
@@ -125,7 +127,7 @@ def wait_until_running(uuid, timeout_seconds=100):
 def wait_for_contents(uuid, substring, timeout_seconds=100):
     start_time = time.time()
     while True:
-        if time.time() - start_time > 100:
+        if time.time() - start_time > timeout_seconds:
             raise AssertionError('timeout while waiting for %s to run' % uuid)
         try:
             out = _run_command([cl, 'cat', uuid])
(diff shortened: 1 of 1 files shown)


test_cli.py
```diff
@@ -109,23 +109,25 @@ def get_info(uuid, key):
...
     while True:
-        if time.time() - start_time > 100:
+        if time.time() - start_time > timeout_seconds:
             raise AssertionError('timeout while waiting for %s to run' % uuid)
         state = get_info(uuid, 'state')
-        # Break when running or one of the final states
-        if state in {'running', 'ready', 'failed'}:
-            assert state == 'running', "waiting for 'running' state, but got '%s'" % state
+        # Break when running or one of the final states or expected_state
+        if state in {'running', 'ready', 'failed', expected_state}:
```
percyliang: Add a comment to list of FINAL_STATES.  `killed` should be in here too.

Also, appending `expected_state` to the list of final states + running is weird semantically...because I would have expected to wait until `expected_state` was hit.

test_cli.py
```diff
@@ -1102,7 +1104,7 @@ def test(ctx):
             'ls dir; cat file; seq 1 10; touch done; while true; do sleep 60; done',
         ]
     )
-    wait_until_running(uuid)
+    wait_until_state(uuid)
```
percyliang: This call doesn't really make sense without `running` either in the function name or as an explicit argument.

teetone: I relied on the default value too much...

test_cli.py
teetone: I used the `FINAL_STATES` defined in `bundle_state.py`. I also separated the expected state into a separate check.

test_cli.py
```diff
@@ -109,23 +110,28 @@ def get_info(uuid, key):
     return _run_command([cl, 'info', '-f', key, uuid])
 
 
-def wait_until_running(uuid, timeout_seconds=100):
+def wait_until_state(uuid, expected_state, timeout_seconds=100):
     start_time = time.time()
     while True:
-        if time.time() - start_time > 100:
+        if time.time() - start_time > timeout_seconds:
```
percyliang, line 125: Can you add a docstring with the logic?  It's not obvious what role FINAL_STATES plays.

test_cli.py
```diff
@@ -1457,13 +1471,13 @@ def test(ctx):
...
     uuid = _run_command([cl, 'run', 'netcat-test.py:' + script_uuid, 'python netcat-test.py'])
-    wait_until_running(uuid)
+    wait_until_state(uuid, 'running')
     time.sleep(5)
     output = _run_command([cl, 'netcat', uuid, '5005', '---', 'hi patrick'])
     check_equals('No, this is dawg', output)
 
     uuid = _run_command([cl, 'run', 'netcat-test.py:' + script_uuid, 'python netcat-test.py'])
-    wait_until_running(uuid)
+    wait_until_state(uuid, 'running')
```
percyliang: Use `State.RUNNING` everywhere

percyliang (approved):

test_cli.py
teetone: I added a docstring and made the logic clearer by explicitly raising an `AssertionError` if the bundle is in a final state that is not the expected state.

PR #2268, opened 2020-05-14
Graphical editor for schemas in worksheet

### LINKED ISSUE #1767: Graphical editor for editing schema
1. mock design
2. implementation

yipenghe:
Resolve #1767 

Allows:
1. changing the fields (field, generated-path, post-processing) on the worksheet
2. move field rows up and down (addschema generated rows moves as a whole)
3. add new field rows after any existing ones

Overall:
Screenshot from 2020-07-29 14-52-37
includes an addschema directive:
Screenshot from 2020-07-29 14-53-21
Empty schema name or field name:
Screenshot from 2020-07-29 14-57-32
No edit permission:
Screenshot from 2020-07-29 14-54-29
Collapsed:
Screenshot from 2020-07-29 14-55-10


codalab/lib/bundle_cli.py
```diff
@@ -3222,6 +3222,9 @@ def display_blocks(self, client, worksheet_info, interpreted_blocks):
                     )
             elif mode == BlockModes.placeholder_block:
                 print('[Placeholder]', block['directive'], file=self.stdout)
+            elif mode == BlockModes.schema_block:
+                # print('[SchemaBlock]', file=self.stdout)
```
epicfaace: Should we keep this?
```suggestion
                print('[SchemaBlock]', file=self.stdout)
```

yipenghe: yeh, need to look into what's causing the test error first

yipenghe:
@teetone Can you help check the small worksheet test, I'm not sure how to add the correct expected lines, there is a mismatch problem between the expected lines and output

codalab/lib/worksheet_util.py
```diff
@@ -481,17 +482,18 @@ def format_metadata(metadata):
                 metadata[name] = apply_func(func, metadata[name])
 
 
-def canonicalize_schema_item(args):
+def canonicalize_schema_item(args, from_schema_name=None):
     """
     Users who type in schema items can specify a partial argument list.
     Return the canonicalize version (a triple).
+    from_schema_name: which schema this item belongs to, used to identify items added from addschema
```
epicfaace, line 490: Can we clarify here that we don't return a triple, but four items in a tuple?

codalab/lib/worksheet_util.py
```diff
@@ -481,17 +482,18 @@ def format_metadata(metadata):
                 metadata[name] = apply_func(func, metadata[name])
 
 
-def canonicalize_schema_item(args):
+def canonicalize_schema_item(args, from_schema_name=None):
```
epicfaace, line 486: Why do we need to return from_schema_name if that's what the user is already passing into this function itself?

epicfaace (commented):
I'm not able to save schemas. I click "edit", then type in the new value in the text box, but then when I click save, the schema just reverts to what it was previously.

codalab/lib/worksheet_util.py
yipenghe: Because the function returns a tuple which can't be later modified, so we have to inject the from_schema_name information at this time. T

yipenghe:
> I'm not able to save schemas. I click "edit", then type in the new value in the text box, but then when I click save, the schema just reverts to what it was previously.

What's the worksheet source you are testing with? I just tried and it saves properly

codalab/lib/worksheet_util.py
```diff
@@ -918,6 +922,29 @@ def flush_worksheets():
...
+                                'header': ["field", "generated-path", "post-processing"],
+                                'schema_name': current_schema_name,
+                                'field_rows': [
+                                    {
+                                        "field": field,
+                                        "generated-path": path,
+                                        "post-processing": post,
+                                        "from_schema_name": from_schema_name,
+                                    }
+                                    for field, path, post, from_schema_name in current_schema
```
epicfaace: For consistency

```suggestion
                                    for name, genpath, post, from_schema_name in current_schema
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,343 @@
...
+        let copiedRows = JSON.parse(JSON.stringify(rows));
+        copiedRows[idx][key] = e.target.value;
+        this.setState({ rows: [...copiedRows] });
+    };
+
+    moveFieldRow = (idx, direction) => () => {
+        if (!this.props.editPermission) return;
+        // -1 for moving up, 1 for moving down
+        const { rows } = this.state;
+        let copiedRows = JSON.parse(JSON.stringify(rows));
```
epicfaace: I think we can just do this here, because we're only modifying the order of the rows (not the properties of each element in `rows` itself)

```suggestion
        let copiedRows = [...rows];
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,343 @@
...
+        let curRow = this.state.rows;
+        curRow.splice(idx + 1, 0, newRow);
+        this.setState({ rows: curRow });
+    };
+
+    changeFieldValue = (idx, key) => (e) => {
+        if (!this.props.editPermission) return;
+        const { rows } = this.state;
+        let copiedRows = JSON.parse(JSON.stringify(rows));
+        copiedRows[idx][key] = e.target.value;
```
epicfaace: I think it's better to avoid `JSON.parse(JSON.stringify())`. Can we do this instead?

```suggestion
        let copiedRows = {...rows, [idx]: {...copiedRows[idx], [key]: e.target.value } };
```

In other cases, if this method gets too verbose (I don't think it's too verbose in this case), we could always use a library like [immutability-helper](https://github.com/kolodny/immutability-helper) to do the task.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,343 @@
...
+
+    addFieldRowAfter = (idx) => (e) => {
+        if (!this.props.editPermission) return;
+        const schemaItem = this.props.item;
+        const schemaHeaders = schemaItem.header;
+        let newRow = { from_schema_name: schemaItem.schema_name };
+        schemaHeaders.forEach((header) => {
+            newRow[header] = '';
+        });
+        let curRow = this.state.rows;
```
epicfaace:
```suggestion
        let curRows = [...this.state.rows];
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,343 @@
...
+import EditIcon from '@material-ui/icons/Edit';
+import ClearIcon from '@material-ui/icons/Clear';
+import classNames from 'classnames';
+
+class SchemaItem extends React.Component<{
+    worksheetUUID: string,
+    item: {},
+    reloadWorksheet: () => any,
+}> {
+    /** Constructor. */
```
epicfaace: remove this comment

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -545,6 +545,17 @@ class Worksheet extends React.Component {
         });
     };
 
+    updateSchemaItem = (rows, start_index, original_length) => {
+        // rows: list of string representing the new schema:
+        //      % schema example
+        //      % add uuid uuid [0:8]
+        // start_index: the index in source corresponding to % schema example
+        // original_length: how many fields the original schema occupied, one for the example (starting from add)
```
epicfaace: Not clear what original_length is. Would it be equal to 0 in the example above?

codalab/rest/worksheet_block_schemas.py
```diff
@@ -149,6 +151,19 @@ class BundleImageBlockSchema(BundleBlockSchema):
     width = fields.Integer()
 
 
+class SchemaBlockSchema(WorksheetBlockSchema):
+    """
+    Schema for user defined schemas in worksheets
```
epicfaace:
```suggestion
    Schema for user-defined schemas in worksheets
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,343 @@
...
+                newIndex += direction;
+            }
+        }
+        this.setState({ rows: copiedRows });
+    };
+
+    removeFieldRow = (idx) => () => {
+        if (!this.props.editPermission) return;
+        const { rows } = this.state;
+        let copiedRows = JSON.parse(JSON.stringify(rows));
```
epicfaace:
```suggestion
        let copiedRows = [...rows];
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    render() {
+        const { classes, editPermission, focused, item } = this.props;
+        const { editing, showSchemaDetail, rows } = this.state;
+        const schemaItem = item;
+        const schemaHeaders = schemaItem.header;
+        const schema_name = schemaItem.schema_name;
+        let headerHtml, bodyRowsHtml;
+        // console.log("props:",this.props.item.field_rows);
```
epicfaace: remove these console log statements

frontend/src/components/worksheets/Worksheet/Worksheet.js
yipenghe: Updated

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
yipenghe: Doesn't quite work, maybe because you need copiedRows to be first defined, updated to use [...rows] though

epicfaace: Aren't you essentially directly mutating state here now though? (because [...rows] only makes a shallow copy -- it doesn't copy every element in rows, and  copiedRows[idx][key] ends up modifying individual elements of the array)

yipenghe: I updated to use splice to first remove that reference and then add it back, so we don't directly modify the reference in state.rows, checked that modifying the copiedRows now doesn't modify the state.rows

percyliang:
Overall this is great!  Some comments:

1) Could we style 'Schema: foo' a bit better?  I think we want this to be as unobtrusive as possible...make make it gray and get rid of 'Schema: ' and replace it with an icon?

2) Could we maybe hide the (+) and trash buttons when we're not in edit mode?  It's not visually obvious that you're supposed to click on the pencil icon to edit first.

3) I find the (+) next to each row and above kind of counterintuitive because you're not adding to a row, you're adding between rows.  I think since you have the ability to move rows around, you can just have one (+) at the bottom or better yet, just have one empty row there at all times so people can just go and start typing (only in edit mode of course).

4) Can you link to the documentation on what generated-path and post-processing is?  You can provide a small (i) icon next to field, generated-path, post-processing to say what they are and give some examples.

5) I'd be nicer if you can drag and drop instead of having the move up and move down icons, but okay to leave this for later.

6) Could you please tool tips on each of the buttons to explain what they do?

7) The four buttons wrap for me (I'm in Firefox on Linux):
Screenshot from 2020-06-21 14-19-30

8) There are currently too many horizontal lines - I don't think you need both the table divider and the line for the textbox.

9) When you ctrl-enter to save the schema, the cursor on the worksheet disappears and can't be recovered.

yipenghe:
> Overall this is great! Some comments:
> 
> 1. Could we style 'Schema: foo' a bit better?  I think we want this to be as unobtrusive as possible...make make it gray and get rid of 'Schema: ' and replace it with an icon?
> 2. Could we maybe hide the (+) and trash buttons when we're not in edit mode?  It's not visually obvious that you're supposed to click on the pencil icon to edit first.
> 3. I find the (+) next to each row and above kind of counterintuitive because you're not adding to a row, you're adding between rows.  I think since you have the ability to move rows around, you can just have one (+) at the bottom or better yet, just have one empty row there at all times so people can just go and start typing (only in edit mode of course).
> 4. Can you link to the documentation on what generated-path and post-processing is?  You can provide a small (i) icon next to field, generated-path, post-processing to say what they are and give some examples.
> 5. I'd be nicer if you can drag and drop instead of having the move up and move down icons, but okay to leave this for later.
> 6. Could you please tool tips on each of the buttons to explain what they do?
> 7. The four buttons wrap for me (I'm in Firefox on Linux):
>    Screenshot from 2020-06-21 14-19-30
> 8. There are currently too many horizontal lines - I don't think you need both the table divider and the line for the textbox.
> 9. When you ctrl-enter to save the schema, the cursor on the worksheet disappears and can't be recovered.

For 3, while we can do that, it would add more trouble if we want to add a row in the middle of a schema with many fields. We could add a tooltip saying it's for adding a new field row, but removing it seems to me is sacrificing a lot of convenience (consider the large schemas used by semantic machines). It should be more obvious once a user know that the the three components of a row is fixed

For 7, I tested with both chrome and firefox on linux, but weirdly it's not wrapping (using the whole screen). If I shrink the browser window, it's will wrap at some point, but I feel like that's expected?

yipenghe:
> Overall this is great! Some comments:
> 
> 1. Could we style 'Schema: foo' a bit better?  I think we want this to be as unobtrusive as possible...make make it gray and get rid of 'Schema: ' and replace it with an icon?
> 2. Could we maybe hide the (+) and trash buttons when we're not in edit mode?  It's not visually obvious that you're supposed to click on the pencil icon to edit first.
> 3. I find the (+) next to each row and above kind of counterintuitive because you're not adding to a row, you're adding between rows.  I think since you have the ability to move rows around, you can just have one (+) at the bottom or better yet, just have one empty row there at all times so people can just go and start typing (only in edit mode of course).
> 4. Can you link to the documentation on what generated-path and post-processing is?  You can provide a small (i) icon next to field, generated-path, post-processing to say what they are and give some examples.
> 5. I'd be nicer if you can drag and drop instead of having the move up and move down icons, but okay to leave this for later.
> 6. Could you please tool tips on each of the buttons to explain what they do?
> 7. The four buttons wrap for me (I'm in Firefox on Linux):
>    Screenshot from 2020-06-21 14-19-30
> 8. There are currently too many horizontal lines - I don't think you need both the table divider and the line for the textbox.
> 9. When you ctrl-enter to save the schema, the cursor on the worksheet disappears and can't be recovered.

For 2, I made the edit pencil button more obvious that it is "active" when not in edit mode, I think it's better not to hide the icons

percyliang:
Thanks for the fixes.

1) Do we need an edit mode at all?  I find it a bit cumbersome when I'm quickly trying to edit the schema and get results.  Most of the time the schema will be hidden, so the user won't accidentally edit something, so you can think of opening it up as putting it in edit mode.  (You'd still have edit/discard).

2) The wrapping is still an issue for me (I don't think my resolution is that low, and there's plenty of whitespace)...could we enforce that the 4 icons don't wrap and let the other fields wrap?  The up and down arrows being on different lines is not good.

3) The check mark is not obvious that it's a save button - maybe a disk icon?  Or at least somehow make it more prominent like Edit.

4) It'd be nice when you hit (+), the keyboard focus was set to the new row.

We might want to make it possible to edit the schema name and add a new schema.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,405 @@
...
+                    <TableCell
+                        component='th'
+                        key={index}
+                        style={{ padding: '5', fontSize: '16px', maxWidth: '100' }}
+                    >
+                        {header}
+                        <Tooltip
+                            title={
+                                explanations[header] +
+                                'Click for more information and examples on schemas'
```
percyliang: End with period.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,405 @@
...
+
+    render() {
+        const { classes, editPermission, focused, item } = this.props;
+        const { showSchemaDetail, rows } = this.state;
+        const schemaItem = item;
+        const schemaHeaders = schemaItem.header;
+        const schema_name = schemaItem.schema_name;
+        let headerHtml, bodyRowsHtml;
+        const explanations = {
+            field: 'Name of the field. ',
```
percyliang: Let's say: "Column name that is displayed."

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,405 @@
...
+            'generalized-path':
+                "Either a field of bundle's metadata (like uuid, name) or a file path inside the bundle prefixed by / (like '/stdout').\n",
```
percyliang: "Either a bundle metadata field (e.g., uuid, name, time, state) or a file path inside the bundle (e.g., /stdout, /stats.json)."

Let's not end these definitions with whitespace, and add the whitespace later when we concatenate with the general suffix.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,405 @@
...
+            'post-processor':
+                'Optional, a function that transforms the string value result of generalized-path, for example uuid uuid [0:8] will keep the first 8 digits of the uuid. ',
```
percyliang: "How to render the value (e.g., %.3f renders 3 decimal points, [0:8] takes the first 8 characters, duration renders seconds, size renders bytes)."

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    render() {
+        const { classes, editPermission, focused, item } = this.props;
+        const { showSchemaDetail, rows } = this.state;
+        const schemaItem = item;
+        const schemaHeaders = schemaItem.header;
+        const schema_name = schemaItem.schema_name;
+        let headerHtml, bodyRowsHtml;
+        const explanations = {
```
percyliang, line 205: In general, "field", "generalized-path", "post-processor" aren't very intuitive, but that's the official names according to the code and documentation.  Maybe instead of having the gray placeholder text be '<none>', use 'display this column name', 'retrieve value from this path', 'how to render value', respectively.

percyliang (changes requested):
This is really nice.  Just a few more things:
1) If the textbox for the schema name is empty, it should show grayed placeholder text "Enter schema name".
2) If there have been no changes to the schema, the save and revert buttons should be grayed out (if this is easy to do).
3) It should not allow saving with an empty schema name, which wrecks the worksheet.
4) If I put 'uuid foo' into the field and save, then 'uuid' becomes the field and 'foo' becomes the generalized-path', which is weird.
5) When the schema block is collapsed, could we hide the schema name textbox.  In general, we want to make this unobstrusive as possible.
6) Don't allow multiple lines for any of the textboxes because it shouldn't be allowed and it messes up the worksheet.

yipenghe:
> This is really nice. Just a few more things:
> 
> 1. If the textbox for the schema name is empty, it should show grayed placeholder text "Enter schema name".
> 2. If there have been no changes to the schema, the save and revert buttons should be grayed out (if this is easy to do).
> 3. It should not allow saving with an empty schema name, which wrecks the worksheet.
> 4. If I put 'uuid foo' into the field and save, then 'uuid' becomes the field and 'foo' becomes the generalized-path', which is weird.
> 5. When the schema block is collapsed, could we hide the schema name textbox.  In general, we want to make this unobstrusive as possible.
> 6. Don't allow multiple lines for any of the textboxes because it shouldn't be allowed and it messes up the worksheet.

All addressed! Now no space, newlines are allowed in the textfields (name & headers)

yipenghe:
Updated screenshots on how it looks right now (as of July. 29th)

percyliang (approved):
Great, thanks!  One minor thing: we should allow spaces in 'field'.

yipenghe:
> Great, thanks! One minor thing: we should allow spaces in 'field'.

Is it field or post processor?

percyliang:
I mean the display name that shows up as column names.

Another thing I noticed: if you hit ctrl+enter on the schema name, it doesn't save.

yipenghe:
> I mean the display name that shows up as column names.
> 
> Another thing I noticed: if you hit ctrl+enter on the schema name, it doesn't save.

Oh I see, so we can use " " to quote the full name

Issue #1205, opened 2019-06-11
allow editing tags for worksheets (using same logic as for bundles)


percyliang:
Render tags in the worksheets info panel at the upper right part of the worksheet.

PR #2273, opened 2020-05-14
worksheet tag allowed editing

Levilian:
feature: #1205


frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -79,15 +79,17 @@ export default ({
...
                                         onClick={(ev) => {
                                             setAnchorEl(ev.currentTarget);
                                         }}
                                         className={classes.permissions}
-                                        style={{ float: 'right' }}
+                                        style={{ display: 'inline-block' }}
                                     >
                                         {renderPermissions(info)}
                                     </div>
+                                    &nbsp;
```
yipenghe: It probably will be better to use {" "} for spaces, ditto for line 82, 119 and 80

frontend/src/components/EditableField.js
```diff
@@ -169,15 +169,17 @@ export const EditableField = withStyles(efStyles)(EditableFieldBase);
 export class WorksheetEditableField extends React.Component<{
     uuid: string,
     fieldName: string,
+    dataType: string,
 }> {
     buildPayload(value) {
+        console.log(this.props.dataType);
```
yipenghe: should be removed

frontend/src/components/EditableField.js
```diff
@@ -169,15 +169,17 @@ export const EditableField = withStyles(efStyles)(EditableFieldBase);
...
         return {
             data: [
                 {
                     id: this.props.uuid,
                     type: 'worksheets',
                     attributes: {
-                        [this.props.fieldName]: value,
+                        [this.props.fieldName]: serializeFormat(value, this.props.dataType),
```
yipenghe, line 181: What is the serializeFormat trying to do here, it seems the value is still a string?

frontend/src/components/EditableField.js
Levilian: Oh nice catch

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
Levilian: You're right. It's more consistent. Another way is to change CSS instead of adding html like this.

frontend/src/components/EditableField.js
Levilian: The value is a string, but the dataType that we want to interpret the value is not.

yipenghe: Oh, so value is not a string, but we want to serialize it into a string?

yipenghe:
Did you forget the push the update commit?

frontend/src/components/EditableField.js
Levilian: Yes - It's a little confusing but this is what we did for `BundleEditableField`

Levilian:
> Did you forget the push the update commit?

Yes... Pushing it now

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -114,6 +114,17 @@ export default ({
                                             />
                                         </div>
                                     </Popover>
+                                    &nbsp;tags:&nbsp;
+                                    <div style={{ display: 'inline-block' }}>
+                                        <WorksheetEditableField
+                                            canEdit={canEdit}
+                                            dataType='list'
+                                            fieldName='tags'
+                                            value={info.tags.join(' | ')}
```
yipenghe: I think we wanted to remove the bar between the tags and treat space as the delimiter?

Levilian: true...

Levilian:
Worksheet tags
Screen Shot 2020-05-27 at 5 01 21 PM
Bundle tags
Screen Shot 2020-05-27 at 5 03 42 PM

yipenghe (approved):

PR #2275, opened 2020-05-14
Prioritize user's own worksheets during search

### LINKED ISSUE #2271: Prioritize user's own worksheets during search

Levilian:
Fix: #2271


Levilian:
When user codalab signed in:

Screen Shot 2020-05-21 at 1 08 29 AM


When user aaa signed in:

Screen Shot 2020-05-21 at 1 19 37 AM

epicfaace (commented):
Wouldn't it be simpler just to sort the `filteredResults` array so that it puts owner name at the beginning of the array? Then you wouldn't need the logic about creating the preRanking and combining to create finalResults.

Levilian:
> Wouldn't it be simpler just to sort the `filteredResults` array so that it puts owner name at the beginning of the array? Then you wouldn't need the logic about creating the preRanking and combining to create finalResults.

The issue is that .reduce() function does not guarantee the iteration order per the documentation - "iteration order is not guaranteed". Therefore we're only doing the sorting after the structuring of the dictionary is done.

epicfaace (approved):

PR #2277, opened 2020-05-14
search bar tweaks - font, usability

epicfaace:
- fix font of search bar text to Roboto
- make search bar have selected text when clicking on a result -- this makes it feel "smoother" for the user as there is usually 

Old:

image

New:

image

Now, if we click on, say, "my async text", the text box gets populated with that value right before the page navigates to the search result. This feels smoother -- and is how Google works.

image


yipenghe (approved):
Screenshot from 2020-05-20 18-25-30
Can you also make  the bottom shadow of the result menu extend when the result is long

Levilian:
> Screenshot from 2020-05-20 18-25-30
> Can you also make the bottom shadow of the result menu extend when the result is long

fixed
Screen Shot 2020-05-21 at 1 05 17 AM

Levilian (approved):
Looks good!

PR #2278, opened 2020-05-14
Slim down docker images

### LINKED ISSUE #2251: Slim down docker images
- We don't need to package node_modules in the frontend Docker container -- it just makes the container way bigger than it needs to be, as the dependencies are already included in the bundled JS. The container is nearly 500 MB -- see https://hub.docker.com/r/codalab/frontend/tags

- We can also slim down the backend images -- worker, server, and default-gpu/default-cpu (around 5 GB)

epicfaace:
Fixes #2251

## Overall size improvements
frontend: 524.45 MB -> 52.23 MB
server:  353.16 MB -> 232.53 MB
worker: 252.64 MB -> 158.86 MB

## Details
- Remove node_modules from frontend container by using multistage builds. Since we're using multistage builds anyway, I also took advantage of this to not run the production build of the React code when running the Docker container in development mode.
Size: 524.45 MB -> 341.23 MB
- Use node:10-slim instead of node:10 for base image for frontend
Size: 341.23 MB -> 53.23 MB
- slim down server and worker images by:
  - [deleting apt caches](https://dev.to/shriharshmishra/today-i-learned-a-simple-trick-to-slim-down-a-docker-image-ed4) - (Server - 353.16 MB -> 334.3 MB, Worker - 252.64 MB -> 233.78 MB)
  - using `--no-install-recommends` with `apt-get` - (Server - 334.3 MB -> 265.43 MB, Worker - 233.78 MB -> 171.86 MB)
  - using pip's `--no-cache-dir` - (Server - 265.43 MB -> 232.53 MB, Worker - 171.86 MB -> 158.86 MB)


docker/dockerfiles/Dockerfile.server
```diff
@@ -19,7 +21,8 @@ RUN apt-get update; apt-get install -y \
     python3-pip \
```
epicfaace, line 21:
```suggestion
    python3-pip \
    python3-setuptools \
```

docker/dockerfiles/Dockerfile.worker
```diff
@@ -12,12 +12,16 @@ MAINTAINER CodaLab Worksheets <codalab.worksheets@gmail.com>
...
-  python3.6-dev \
-  python3-pip;
+RUN apt-get update && \
+  apt-get install -y --no-install-recommends software-properties-common && \
+  add-apt-repository ppa:deadsnakes/ppa && \
+  apt-get update && \
+  apt-get install -y --no-install-recommends \
+    python3.6 \
+    python3.6-dev \
+    python3-pip && \
```
epicfaace:
```suggestion
    python3-pip \
    python3-setuptools && \
```

epicfaace:
@percyliang can you review this PR when you get a chance?

One drawback of using multiple stages on the frontend image is that it seems that it's not able to cache the first stage (the build of the production code) -- working on a way around this.

percyliang:
Can you comment on what the performance impact of both local development and GitHub Actions when you make incremental changes to the frontend?

epicfaace:
It's much faster to build a frontend image locally because we no longer run a production build with this PR.

Build time on CI is slower for frontend, unfortunately, because the non-final stages on Dockerfile.frontend aren't included in the cache when using `--cache-from`. (https://github.com/moby/moby/issues/34715) We would need to find away around this before being able to merge this PR.

yipenghe:
Are we still doing this?

epicfaace:
We need to follow these steps to properly enable caching of multistage builds: https://pythonspeed.com/articles/faster-multi-stage-builds/

Let's put this on the backlog for now, though.

PR #2279, opened 2020-05-18
Slurm worker manager

### LINKED ISSUE #1019: Slurm Worker
We want Codalab to be compatible with a university cluster running Slurm.
Immediately we want to let users of the Stanford NLP cluster use it to run their CodaLab jobs.

candicegjing:
Fixed #1019 .
I want to get some feedback for this PR in case I missed any details or use cases here.
### Overview
To submit a `slurm-batch` job, the user needs to run the following command to start the worker manager:
```
cl-worker-manager --server=http://localhost slurm-batch --nodelist=jagupard10
```
The Slurm batch worker manager will just run once without keep watching for future staged bundles. Hence, if the user wants to spawn a new Slurm worker, he/she has to run the same command another time. 

A successfully spawned worker job from `sc` will look like as follows:
```
stanford@Stanfords-MacBook-Pro codalab-worksheets % cl-worker-manager --server=http://localhost slurm-batch --nodelist=jagupard10
Namespace(cpus=1, dry_run=False, gpus=1, job_definition_name='codalab-slurm-worker', max_workers=10, memory_mb=2048, min_seconds_between_workers=60, min_workers=1, nodelist='jagupard10', once=True, partition='jag-standard', search=[], server='http://localhost', sleep_time=5, verbose=False, worker_idle_seconds=600, worker_manager_name='slurm-batch', worker_tag=None)
2020-05-20 17:45:39,693 Started worker manager.
2020-05-20 17:45:39,729 Staged bundles [state=staged]: 0xce9d53b2415e45c182a0693beed3847d 0xf311bf0cd69040cdb0e47cbb593ac3b7 0x4eb02dcc325244ff9b04c1931072bb2f 0x502fad7aabd343e99b89c71bf7485a71 0x0f91f28629bc4efcaeb753664a4aff20 0x2b7ffc86d4464c689b9f8d03179cb8c4 0x68a230ae604e4db2bb24cb1afbc3afa1 0x37f2595625794b6daace08b34930bff7 0x30c8ead85fda4358bac21973d8820e8a 0x090c800cdb594200818ffad8793aeef9
Get slurm jobs
2020-05-20 17:45:39,729 10 staged bundles (0 removed since last time), 0 worker jobs (min=1, max=10) (0 active, 0 pending)
2020-05-20 17:45:39,729 Want to launch a worker because we have 10 > 0 staged bundles
2020-05-20 17:45:39,729 Launch a worker because we are under the minimum (0 < 1)
2020-05-20 17:45:39,729 Starting a worker!
Slurm Batch Job Definition
#!/bin/bash

#SBATCH --cpus-per-task=3
#SBATCH --gres=gpu:1
#SBATCH --job-name=stanford-codalab-slurm-worker-3086758
#SBATCH --mem-per-cpu=2048
#SBATCH --nodelist=jagupard10
#SBATCH --ntasks-per-node=1
#SBATCH --open-mode=append
#SBATCH --output=stanford-codalab-slurm-worker-3086758.out
#SBATCH --partition=jag-standard
#SBATCH --time=10-0

srun --unbuffered cl-worker --server http://localhost --verbose --exit-when-idle --idle-seconds 600 --work-dir /Users/stanford/slurm-batch-scratch/4e524789af034496b1717e943ceed111 --id 4e524789af034496b1717e943ceed111 --network-prefix cl_worker_4e524789af034496b1717e943ceed111_network --pass-down-termination
2020-05-20 17:45:39,730 Saved the Slurm Batch Job Definition to /Users/stanford/slurm-batch-scratch/4e524789af034496b1717e943ceed111/stanford-codalab-slurm-worker-3086758.slurm
2020-05-20 17:45:39,733 Submitted job: sbatch stanford-codalab-slurm-worker-3086758.slurm
```

In the same time, I am adding more flexibility to Slurm job parameter.
Please feel free to suggest any use cases that I am missing here.


codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,207 @@
...
+            command.extend(['--tag', self.args.worker_tag])
+
+        slurm_args = self.create_slurm_args(self.args)
+        job_definition = self.create_job_definition(slurm_args=slurm_args, command=command)
+
+        # Not submit job to Slurm if dry run is specified
+        if self.args.dry_run:
+            return
+
+        batch_script = str(work_dir.joinpath(slurm_args['job-name'] + '.slurm'))
```
percyliang: Does this ever get cleaned up?

candicegjing: Hmm, not right now. I thought it would be nice to keep those configurations locally for future reference. But we can definitely clean it up if that's preferred.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,241 @@
...
+        # states here as we don't want to create more workers that we actually need.
+        return [WorkerJob(job) for job in running_jobs if job in self.submitted_jobs]
+
+    def start_worker_job(self):
+        """
+        Start a CodaLab Slurm worker that submits batch job to Slurm
+        """
+        worker_id = uuid.uuid4().hex
+        # Set user's local home directory as worker directory
+        work_dir = Path(str(Path.home()), "slurm-batch-scratch/{}".format(worker_id))
```
nelson-liu: I think this should be configurable somehow—if we leave this as-is for the Stanford NLP cluster, we'd be writing to NFS instead of the worker's local disk, which would be pretty slow (and possibly impact other users).

Something to keep in mind: if we all put them into a fixed path (say, `hostname/scr1`, like the default `nlp-cl-worker` script does, it's possible that that folder could become full and a bunch of jobs would fail. Personally, I use a script `get-scr-with-most-space` that returns the `hostname/scr*` with the most space, and set work-dir to `$(get-scr-with-most-space)/$USER-cl-worker`...not sure how that'd fit in here, but it'd be nice to spread around the usage somehow.

candicegjing: hmm, disk usage could be problematic here. I think spreading around disk usage will slow down the issue, but it will eventually happen at some point. I am wondering if `homename/scr*` is a shared disk space that can be used for all users when they are running the Slurm worker manager? Not sure if we can request some disk space to save the worker data? (But this will still go over the disk space) Setting a flag to limit the disk usage is another option, but users will still face the running out of the disk space issue. I think we probably want to choose an option to solve it to some extent to which we can minimize this problem here.

nelson-liu: Maybe we should allow workers to delete their `work-dir` upon exit?

candicegjing: Yeah, that sounds like a good idea. In addition to this, I am also wondering maybe we can pass the `work-dir` as an option to let users choose where they would want to put their data instead of hardcoded there?

nelson-liu: yup, that sounds great! Except it'd be a `work-dir-prefix` in this case, right?

codalab/worker_manager/slurm_batch_worker_manager.py
nelson-liu: +1 to cleaning up

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,240 @@
...
+    SRUN_COMMAND = 'srun'
+    SBATCH_COMMAND = 'sbatch'
+    SBATCH_PREFIX = '#SBATCH'
+    SQUEUE_COMMAND = 'squeue'
+    SBATCH_COMMAND_RETRUN_REGEX = re.compile(r'^Submitted batch job (\d+)$')
+
+    @staticmethod
+    def add_arguments_to_subparser(subparser):
+        subparser.add_argument(
+            '--job-definition-name',
```
nelson-liu: Can this get changed to "job name"? Job definitions are an AWS Batch-specific concept.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def create_slurm_args(self, args):
+        """
+        Convert command line arguments to Slurm
+        :param args: command line arguments
+        :return: a dictionary of Slurm arguments
+        """
+        slurm_args = {}
+        slurm_args['nodelist'] = self.args.nodelist
+        slurm_args['mem-per-cpu'] = self.args.memory_mb
```
nelson-liu: use `memory` instead?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,240 @@
...
+        slurm_args['partition'] = self.args.partition
+        slurm_args['gres'] = "gpu:" + str(self.args.gpus)
+        # job-name is unique
+        slurm_args['job-name'] = self.create_random_job_name(self.args.job_definition_name)
+        slurm_args['cpus-per-task'] = 3
```
nelson-liu: `self.cpus`?

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Done

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Do you mean `--mem` option or `self.args.memory_mb`? `memory_mb` is to match awsworkermanager.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Done

candicegjing:
@percyliang @nelson-liu This PR is tested on dev environment with the following cases:
1. Properly spawn the desired number of Slurm workers.
1. Jobs are able to run and finish once the allocated worker is ready.
1. Preemption on worker resources is working properly, including 
    1. jobs are able to send back to the `staged` state with `pass-down-termination` option specified
    1. local docker containers are stopped and removed as necessary  
1. Cancel a Slurm allocated worker resource is working properly, including 
    1. jobs can be stopped 
    1. local docker containers are stopped and removed as necessary


TODO
1. Testing of 1 job per worker will be done once the implementation of https://github.com/codalab/codalab-worksheets/issues/2287 is merged. 
1. Please feel free to suggest what else we need to add here.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+        subparser.add_argument(
+            '--dry-run',
+            action='store_true',
+            help='Print out Slurm batch job definition without submitting to Slurm',
+        )
+        subparser.add_argument(
+            '--user', type=str, default=getpass.getuser(), help='User to run the Batch jobs as'
+        )
+        subparser.add_argument(
+            '--password-file',
```
nelson-liu, line 71: can you make this optional and also check for the `CODALAB_USERNAME` and `CODALAB_PASSWORD` environment variables? Like how the AWSBatchWorkerManager does it: https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/aws_batch_worker_manager.py#L108-L109

candicegjing: Ah, I was looking for this!!! That would be great to have this from ENV. I am updating.

candicegjing: Also, this would require some manual setup to add both environment variables in your environment. Just something to remember, as they don't come as default.

nelson-liu: that's fine. I think you should make `--password-file` optional. If password-file is provided and the environment variables are set, default to the password-file. If password-file is not provided and the environment variables are set, use the environment variable. If the environment variables aren't set, and password-file isn't specified, raise an error?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+        print(job_definition)
+        return job_definition
+
+    def create_random_job_name(self, job_name):
+        """
+        Generate a random Slurm job name
+        :param job_name:
+        :return: slurm job name
+        """
+        return self.username + "-" + job_name + '-' + str(random.randint(0, 5000000))
```
nelson-liu: since we're importing uuid anyway, maybe replace `str(random.randint(0, 5000000))` with a uuid?

codalab/worker_manager/slurm_batch_worker_manager.py
nelson-liu: Oops, yes, I meant the `--mem` option instead of `mem-per-cpu`.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+    SQUEUE_COMMAND = 'squeue'
+    SBATCH_COMMAND_RETRUN_REGEX = re.compile(r'^Submitted batch job (\d+)$')
+
+    @staticmethod
+    def add_arguments_to_subparser(subparser):
+        subparser.add_argument(
+            '--job-name',
+            type=str,
+            default='codalab-slurm-worker',
+            help='Name for the job definitions that will be generated by this worker manager',
```
nelson-liu: Remove mention of job definitions here?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+        )
+        subparser.add_argument(
+            '--nodelist', type=str, default='', help='The worker node to run jobs in'
```
nelson-liu, line 41: Does the default of `''` work? i.e., if you set `slurm_args['nodelist'] = ''` below, is the `nodelist` all nodes, or is the nodelist no nodes? Maybe consider removing the default here, and, in `create_slurm_args`, only add the nodelist argument `if self.args.nodelist`?

candicegjing: Yes, I tried that. Default `--nodelist = ''` will just randomly give you a partition.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Yup, I think that works! I think I need to check the existence of `CODALAB_USERNAME` and `CODALAB_PASSWORD` early so that the newly spawned worker won't keep hanging there or failed inside the worker without giving any notice to the user.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        sbatch_args = [
+            '{} --{}={}'.format(self.SBATCH_PREFIX, key, slurm_args[key])
+            for key in sorted(slurm_args.keys())
+        ]
+        # Using the --unbuffered option with srun command will allow output
+        # appear in the output file as soon as it is produced.
+        srun_args = [self.SRUN_COMMAND, '--unbuffered'] + command
+        # job definition contains two sections: sbatch arguments and srun command
+        job_definition = '#!/bin/bash\n\n' + '\n'.join(sbatch_args) + '\n\n' + ' '.join(srun_args)
```
nelson-liu: `#!/usr/bin/env bash` as shebang?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def __init__(self, args):
+        super().__init__(args)
+        self.username = self.args.user
+        # A set of newly submitted job id to keep tracking worker status, as worker might not be created right away.
+        self.submitted_jobs = self.load_worker_jobs()
+
+    def load_worker_jobs(self):
+        """
+        Load worker jobs that are created using SlurmWorkerManager and
```
nelson-liu, line 92: not all slurm jobs are guaranteed to be created by the `slurmworkermanager`, right? I think you should also check the job name and make sure it matches the pattern created by `create_random_job_name`?

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Oh, actually, to make this more robust, I probably should do `which bash` first and them embed the returning value to this.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Yes, it's currently coded as matching the prefix `username + job_name`. Or maybe I can make it more explicit? Coz I check prefix username and see if job_name is in the string...

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def run_command(self, command):
+        proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        output, errors = proc.communicate()
+        if output:
+            logger.info("Executed command: {}".format(' '.join(command)))
+            return output.decode()
+        if errors:
+            print(
+                "Failed to executing {}: {}: {}".format(' '.join(command), errors, proc.returncode)
```
nelson-liu: `s/executing/execute/`

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Removed

codalab/worker_manager/slurm_batch_worker_manager.py
nelson-liu: Oh oops I was misled--you're right! The comment below with the example output made me think that you were just counting lines from `squeue` without looking at the name. Maybe update the comment, though

codalab/worker_manager/slurm_batch_worker_manager.py
nelson-liu: I think `#!/usr/bin/env` does the same thing as `which`—it returns the path to the executable that's first found in the $PATH.

candicegjing: Oh, sorry, yeah, you are right. I missed read the line. Updating...

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Oh, yes. The comments should be updated.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: That's a good idea. Coz we don't actually need those many random strings. Do we want to cut the first few digits e.g. 8 or we want to use the full hex string (32 digits), which might be quite long?

nelson-liu: Yup, let's shorten.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,266 @@
...
+        subparser.add_argument(
+            '--password-file',
+            type=str,
+            required=True,
+            help='Path to the file containing the username and '
+            'password for logging into the CodaLab worker '
+            'each on a separate line. If not specified, the '
+            'the worker will fail to start',
+        )
+        subparser.add_argument(
```
nelson-liu, line 78: Reminder to remove this once https://github.com/codalab/codalab-worksheets/pull/2312 is in

nelson-liu:
Hm, so I'm trying to run the `slurmworkermanager` and test it under some real-world use cases. There isn't a very satisfying way to set the `work-dir`, though.

Currently, in `nlp-cl-worker`, I have:

```
CMD="${CMD} --work-dir ${HOSTNAME}/$(get-scr-with-most-space)/$USER-cl-worker"
```

i.e., the work-dir is only set once we get into the host. There isn't a very good way of doing this at the moment. The best thing I can think of (disclaimer: entirely a hack) is to have a custom script `cl-worker` that comes before the codalab-installed `cl-worker`, where the custom script defines the `--work-dir` based off of the current host and otherwise is just a pass-through to the codalab-installed `cl-worker`. Ideally, we'd like to be able to populate these fields with arbitrary values from the host, but I can't think of a clean way to do this...thoughts?

candicegjing:
> disclaimer

1. For populating `$(get-scr-with-most-space)`, what are the potential directories that we are allowed to use?
1. Since this is going to be Slurm specific, we can add this functionality to this class.
1. Another note is that we need to be able to keep track of where the directory is. Otherwise, with more and more directories being created, we might completely lose control of where bundles are stored.

nelson-liu:
I think I figured out a workaround (for now). For each worker node, I made symbolic links from /scr/scrX -> /$(hostname -a)/scrX, and a separate link /scr/scr-with-most-space -> /$(hostname -a)/(scrX that has the most empty space) . So, for now, I'll just set the `work-dir` to the static path `/scr/scr-with-most-space` . let's see if it holds up, and then we can optimize if it doesn't.

> For populating $(get-scr-with-most-space), what are the potential directories that we are allowed to use?

`/$(hostname -a)/scr*`

> Since this is going to be Slurm specific, we can add this functionality to this class.

Hmm..it's technically SAIL-cluster specific. Not sure how I feel about adding something so specific to this class in particular.

> Another note is that we need to be able to keep track of where the directory is. Otherwise, with more and more directories being created, we might completely lose control of where bundles are stored.

I'm not sure what you mean, could you elaborate?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,266 @@
...
+            WorkerJob(active=True) if job in running_jobs else WorkerJob(active=False)
+            for job in self.submitted_jobs
+        ]
+
+    def start_worker_job(self):
+        """
+        Start a CodaLab Slurm worker that submits batch job to Slurm
+        """
+        worker_id = uuid.uuid4().hex
+        # Set up worker directory
```
nelson-liu: I think this should be omitted—the worker directory isn't necessarily guaranteed to exist on the host that's running the slurm worker.

nelson-liu: Sorry, I mean the line with `work_dir.mkdir(parents=True, exist_ok=True)`.

codalab/worker_manager/slurm_batch_worker_manager.py
nelson-liu: I think we should be saving the `batch_script` to a different directory than the `work-dir`, since the `work-dir` might not exist on the host running the `slurmworkermanager`. E.g., if we run the `slurmworkermanager` on sc, but we want to write to each worker's local disk (a path that doesn't exist on sc)

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Hmm, what do you mean? It's currenlty coded to create any missing directories as necessary. See this [pathlib.Path.mkdir](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir)

nelson-liu: Consider the following (actual) example: I want to run workers on the jagupard machines, which each have their own separate filesystem. I want the codalab workers to use the directory `/scr/scr-with-most-space` on each of the jagupard machines. So, I set that as the input work dir to the slurmworkermanager. However, I'm launching the slurmworkermanager from sc. It sees that I want to use `/scr/scr-with-most-space` as the work dir, so it tries to make `/scr/scr-with-most-space` on sc, not on the worker node. I don't have permissions to do this operation on sc, so it fails.

In general, I think the temporary directories we use to store sbatch scripts and other artifacts of the slurmworkermanager should be divorced from the directory that the worker, when it lands on a slurm node, runs its jobs in.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,274 @@
...
+            logger.error(
+                "Failed to create the Slurm work directory: {}. "
+                "Stop creating a new worker.".format(e)
+            )
+            return
+
+        # This needs to be a unique directory since Batch jobs may share a host
+        worker_network_prefix = 'cl_worker_{}_network'.format(worker_id)
+        # Codalab worker's work directory
+        worker_work_dir = Path(self.args.worker_work_dir_prefix, 'codalab-worker-scratch')
```
nelson-liu: Hm, this needs to be a unique directory, right? Else, if two workers are started on the same host, they'll share a working directory.

candicegjing: Added worker_id to differentiate.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+            '--job-name',
+            type=str,
+            default='codalab-slurm-worker',
+            help='Name for the job that will be generated by this worker manager',
+        )
+        subparser.add_argument(
+            '--nodelist', type=str, default='', help='The worker node to run jobs in'
+        )
+        subparser.add_argument(
+            '--partition', type=str, default='jag-standard', help='Name of batch job queue to use'
```
nelson-liu: Maybe this should be unset by default, it's a rather stanfordnlp-specific value.

nelson-liu: But you should flag the argument as `required=True`

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+        subparser.add_argument(
+            '--user', type=str, default=getpass.getuser(), help='User to run the Batch jobs as'
+        )
+        subparser.add_argument(
+            '--password-file',
+            type=str,
+            help='Path to the file containing the username and '
+            'password for logging into the CodaLab worker '
+            'each on a separate line. If not specified, the '
```
nelson-liu: update docstring to mention `CODALAB_USERNAME` and `CODALAB_PASSWORD`?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+                jobs_to_remove.add(job_id)
+        self.submitted_jobs = self.submitted_jobs - jobs_to_remove
+        logger.info("Submitted jobs: ".format(self.submitted_jobs))
+
+        # Get all the Slurm workers that are submitted by SlurmWorkerManager and owned by the current user.
+        # Returning result will be in the following format:
+        # JOBID:STATE (header won't be included with "--noheader" option)
+        # 1478828,PENDING
+        # 1478830,PENDING
+        jobs = self.run_command(
```
nelson-liu, line 138: You don't want to count all of the things returned by `squeue -u`, right? I think you want to match the name prefix, like how it's done in `load_worker_jobs`.

nelson-liu: Oh, I see that you later filter by checking if it's in `self.submitted_jobs`. let me know if i'm wrong about that.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        running_jobs = running_jobs.strip().split()
+
+        return [
+            WorkerJob(active=True) if job in running_jobs else WorkerJob(active=False)
+            for job in self.submitted_jobs
+        ]
+
+    def start_worker_job(self):
+        """
+        Start a CodaLab Slurm worker that submits batch job to Slurm
```
nelson-liu: nit, but the codalab slurm workeer doesn't submit a batch job to slurm. maybe `Start a CodaLab Slurm worker by submitting a batch job to slurm`?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+                "Failed to create the Slurm work directory: {}. "
+                "Stop creating a new worker.".format(e)
+            )
+            return
+
+        # This needs to be a unique directory since Batch jobs may share a host
+        worker_network_prefix = 'cl_worker_{}_network'.format(worker_id)
+        # Codalab worker's work directory
+        worker_work_dir = Path(
+            self.args.worker_work_dir_prefix, 'codalab-worker-scratch', worker_id
```
nelson-liu: Can you change `codalab-worker-scratch` to `codalab-slurm-worker-scratch`. Also, instead of just `worker_id`, can you do `username + worker_id` so it's easier to quickly figure out which ones belong to a particular user?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        ]
+        if self.args.worker_tag:
+            command.extend(['--tag', self.args.worker_tag])
+        if self.args.password_file:
+            command.extend(['--password-file', self.args.password_file])
+
+        slurm_args = self.create_slurm_args(worker_id)
+        job_definition = self.create_job_definition(slurm_args=slurm_args, command=command)
+
+        # Not submit job to Slurm if dry run is specified
```
nelson-liu: `s/Not/Do not`

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+    An enumeration of Slurm commands.
+    """
+    SRUN = 'srun'
+    SBATCH = 'sbatch'
+    SQUEUE = 'squeue'
+    SACCT = 'sacct'
+
+    # SBATCH configuration in bash script
+    SBATCH_PREFIX = '#SBATCH'
+    SBATCH_COMMAND_RETRUN_REGEX = re.compile(r'^Submitted batch job (\d+)$')
```
nelson-liu: `s/SBATCH_COMMAND_RETRUN_REGEX/SBATCH_COMMAND_RETURN_REGEX/`

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        if self.args.dry_run:
+            return
+
+        batch_script = str(slurm_work_dir.joinpath(slurm_args['job-name'] + '.slurm'))
+        self.save_job_definition(batch_script, job_definition)
+        job_id_str = self.run_command([self.SBATCH, batch_script])
+
+        match = re.match(self.SBATCH_COMMAND_RETRUN_REGEX, job_id_str)
+        if match is not None:
+            job_id = match.group(1)
```
nelson-liu: Can you add a logging message here? Maybe something along the lines of `Submitted batch job with job id <jobid>`

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        """
+        sbatch_args = [
+            '{} --{}={}'.format(self.SBATCH_PREFIX, key, slurm_args[key])
+            for key in sorted(slurm_args.keys())
+        ]
+
+        # Check the existence of environment variables CODALAB_USERNAME and
+        # CODALAB_PASSWORD when password_file is not given.
+        worker_authentication = (
+            '[ ! -z "${CODALAB_USERNAME}" ] && [ ! -z "${CODALAB_PASSWORD}" ] || exit;\n\n'
```
nelson-liu: Would it be possible to add some sort of logging message that explicitly states that the worker is exiting because there's no `--password-file` and `CODALAB_USERNAME` and `CODALAB_PASSWORD` are not set?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        self.submitted_jobs.add(job_id)
+
+    def run_command(self, command):
+        proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        output, errors = proc.communicate()
+        if output:
+            logger.info("Executed command: {}".format(' '.join(command)))
+            return output.decode()
+        if errors:
+            print("Failed to execute {}: {}: {}".format(' '.join(command), errors, proc.returncode))
```
nelson-liu: `logger` instead of print?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # Job definition contains two sections: sbatch arguments and srun command
+        job_definition = (
+            '#!/usr/bin/env bash\n\n'
+            + '\n'.join(sbatch_args)
+            + '\n\n'
+            + worker_authentication
+            + ' '.join(srun_args)
+        )
+        print("Slurm Batch Job Definition")
```
nelson-liu: logger instead of print? (also applies to line below)

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        print(job_definition)
+        return job_definition
+
+    def authenticate_codalab_account(self):
```
nelson-liu: seems like an unused function

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        slurm_args = {}
+        slurm_args['nodelist'] = self.args.nodelist
+        slurm_args['mem'] = self.args.memory_mb
+        slurm_args['partition'] = self.args.partition
+        slurm_args['gres'] = "gpu:" + str(self.args.gpus)
+        # job-name is unique
+        slurm_args['job-name'] = self.username + "-" + self.args.job_name + '-' + worker_id[:8]
+        slurm_args['cpus-per-task'] = str(self.args.cpus)
+        slurm_args['ntasks-per-node'] = 1
+        slurm_args['time'] = '10-0'
```
nelson-liu: Can we make this configurable, with a default of "10-0"?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+            for job in self.submitted_jobs
+        ]
+
+    def start_worker_job(self):
+        """
+        Start a CodaLab Slurm worker that submits batch job to Slurm
+        """
+        worker_id = uuid.uuid4().hex
+        # Set up the Slurm worker directory
+        slurm_work_dir = Path(self.args.slurm_work_dir, worker_id)
```
nelson-liu, line 204: Right now, the only thing being saved in slurm_work_dir is the slurm batch config, right? If that's the case, it seems unnecessary to create the extra folder `worker_id` in `slurm_work_dir`, i'd prefer for things to be just written directly to the directory.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        slurm_args['mem'] = self.args.memory_mb
+        slurm_args['partition'] = self.args.partition
+        slurm_args['gres'] = "gpu:" + str(self.args.gpus)
+        # job-name is unique
+        slurm_args['job-name'] = self.username + "-" + self.args.job_name + '-' + worker_id[:8]
+        slurm_args['cpus-per-task'] = str(self.args.cpus)
+        slurm_args['ntasks-per-node'] = 1
+        slurm_args['time'] = '10-0'
+        slurm_args['open-mode'] = 'append'
+        slurm_args['output'] = slurm_args['job-name'] + '.out'
```
nelson-liu: can we also save this in `slurm_work_dir`?

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: hm, I don't think we should mention `CODALAB_USERNAME` or `CODALAB_PASSWORD` here, as they are environment variables. Most importantly, they are not related to the password-file, right?

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: That's correct. It will only report jobs in `self.submitted_jobs` which already has a filter on the name prefix.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: hm, I am not sure about this. One reason is that most users might come from Stanford NLP, and I am not sure how many other places are using Slurm to be honestly. I am also wondering if we should change this to match with the current behavior of `nlprun` for an easier transition for users.

nelson-liu: I guess it depends overall on how stanfordnlp-specific the package should be —there are users outside of stanford, after all.

We can add a script on the cluster `nlp-cl-worker-manager`, just like how we have `nlp-cl-worker` right now, that sets reasonable stanford-specific defaults?

codalab/worker_manager/slurm_batch_worker_manager.py
nelson-liu: Well, the comment says that `If not specified, the the worker will fail to start` (also a typo that I just noticed). This isn't true, though—you can not specify a password-file and be fine, as long as `CODALAB_USERNAME` and `CODALAB_PASSWORD` are set.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: That's a good point. In fact, I am adding this logging information in the `run_command` function, so that we have more useful information logged.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Initially, I also thought about writing this in the logger. However, I found it doesn't look nice with the logging time in front of each line, as it breaks the bash scripts into lines. Besides, we already saved this file into local disk, add this to logger might be redundant?

nelson-liu: I thought the logger only adds a logging statement to the start of the first line? I think that regardless of whether it's print or logger, though, it's important to show this output—if I'm running the slurm worker manager, I don't want to have to be `cat`ting files to verify things are working, I'd rather just look at the workermanager output.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Yes, you are totally right! This should be updated as we just changed it to be an optional argument.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: yeah, we could do that and we can just preconfigure some of those settings in `nlp-cl-worker-manager`

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: It is currently generated in the worker manager output if you see the overview of this ticket as an example (just copied and pasted here): 
```
2020-05-20 17:45:39,729 Starting a worker!
Slurm Batch Job Definition
#!/bin/bash

#SBATCH --cpus-per-task=3
#SBATCH --gres=gpu:1
#SBATCH --job-name=stanford-codalab-slurm-worker-3086758
#SBATCH --mem-per-cpu=2048
#SBATCH --nodelist=jagupard10
#SBATCH --ntasks-per-node=1
#SBATCH --open-mode=append
#SBATCH --output=stanford-codalab-slurm-worker-3086758.out
#SBATCH --partition=jag-standard
#SBATCH --time=10-0

srun --unbuffered cl-worker --server http://localhost --verbose --exit-when-idle --idle-seconds 600 --work-dir /Users/stanford/slurm-batch-scratch/4e524789af034496b1717e943ceed111 --id 4e524789af034496b1717e943ceed111 --network-prefix cl_worker_4e524789af034496b1717e943ceed111_network --pass-down-termination
2020-05-20 17:45:39,730 Saved the Slurm....
``` 
I think what I was trying to say is that the format is a bit messy in my view, e.g. the first two lines got prefix ahead of them.
```
2020-06-01 09:55:44,590 Slurm Batch Job Definition
2020-06-01 09:55:44,590 #!/usr/bin/env bash

#SBATCH --cpus-per-task=1
#SBATCH --gres=gpu:1
#SBATCH --job-name=jingge19-codalab-slurm-worker-717845cc
#SBATCH --mem=2048
#SBATCH --nodelist=
#SBATCH --ntasks-per-node=1
#SBATCH --open-mode=append
#SBATCH --output=slurm-worker-scratch/717845cc05cb40f09a0e169cb1259d9b/jingge19-codalab-slurm-worker-717845cc.out
#SBATCH --partition=jag-standard
#SBATCH --time=10-0

srun --unbuffered cl-worker --server https://worksheets-dev.codalab.org --verbose --exit-when-idle --idle-seconds 600 --work-dir /sailhome/jingge19/slurm-codalab-worker-scratch/jingge19-717845cc05cb40f09a0e169cb1259d9b --id 717845cc05cb40f09a0e169cb1259d9b --network-prefix cl_worker_717845cc05cb40f09a0e169cb1259d9b_network --pass-down-termination --password-file /sailhome/jingge19/authentication
2020-06-01 09:55:44,592 Saved the Slurm....
```

But certainly, if you would prefer this way, I am happy to change it.

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Updated.

candicegjing:
@nelson-liu I updated the code. In addition to what you requested, I refactored a bit more just to make the functionality more robust and easy to read. Please take a look and let me know what else we need to modify here.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,356 @@
...
+        for job in jobs.strip().split():
+            job_id, job_name = job.split(',')
+            if job_name.startswith(self.username) and self.args.job_name in job_name:
+                submitted_jobs.add(job_id)
+        return submitted_jobs
+
+    def get_worker_jobs(self):
+        """
+        Return a list of workers in RUNNING and PENDING state.
+        The current Slurm version on NLP cluster is 17.11.13-2, which doesn't have rest api provided.
```
nelson-liu: Omit mention of NLP cluster? Just say something about how you do this for compatibility with older slurm setups. Should also probably have a note somewhere with the slurm version that this was developed / tested with, just for users' information

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -0,0 +1,356 @@
...
+        :param worker_id: a string of worker id in 32 digits.
+        :return: the command to run on
+        """
+        # This needs to be a unique directory since Batch jobs may share a host
+        worker_network_prefix = 'cl_worker_{}_network'.format(worker_id)
+        # Codalab worker's work directory
+        if self.args.worker_work_dir_prefix:
+            work_dir_prefix = Path(self.args.worker_work_dir_prefix)
+        else:
+            work_dir_prefix = Path.home()
```
nelson-liu: To match `cl worker`, should we make this `.` instead of `~`?

nelson-liu (approved):
LGTM, thanks for working on this!!

codalab/worker_manager/slurm_batch_worker_manager.py
candicegjing: Make sense.

Issue #2287, opened 2020-05-22
Add a flag to indicate number of jobs allowed to run on worker


candicegjing:
This is part of Slurm Worker Manager feature request. see #1019 

With this option, users can specify the number of jobs they wish to run on this worker. Once those jobs are finished, the worker will stop and exit.
With this option, in Slurm worker manager, we don't need to worry about matching tag between a bundle and a newly spawned worker, as even if we submit 10 jobs with the same tag since each worker can only run 1 job, then 10 jobs will eventually end up landing on 10 workers.

PR #2288, opened 2020-05-22
Add a flag to indicate the number of job allowed on worker. (design 1)

candicegjing:
This is a partial feature request #1019 . More details are documented at #2287.

### Overview
There are two potential implementations here:
1. Option 1 is to let the worker passively handle the job counting itself and not to bother the bundle-manager. This is implemented in this PR. 
    a. Pros: less code modification 
    b. Cons: some jobs may get stuck in the `starting` state for a while before moved back to the `staged` state. See the following link:
https://github.com/codalab/codalab-worksheets/blob/a65ad040aa9de2cb2fcdf3e09ec3fbcd3889fbc9/codalab/server/bundle_manager.py#L264-L271
1. Option 2 is to let the bundle-manager assign the number of jobs that the worker can take (there is no logic change in worker side expect the additional flag). This is implemented in #2289
    a. Pros: jobs won't get stuck in the `starting` state. (No frequent intermediate state transitions any more) 
    b. Cons: more code modification and a redundant extra column will be generated in the worker table. It's not quite useful in most of the cases...

### Question 
1. Do we consider reduce the timeout period from 5 minutes to something shorter for a smoother state transition?
1. Not exactly sure if the name of the new flag is acceptable. Please feel free to suggest.
1. Need to decide which option to use? Or maybe there are other solutions that are better than proposed ones.

### DIFF
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -113,13 +113,13 @@ def parse_args():
     parser.add_argument(
         '--exit-number-jobs',
         help='The worker will exit after this many jobs assigned and finished on this worker',
-        type=int,
+        type=check_positive_int,
         default=10000,
     )
     parser.add_argument(
         '--idle-seconds',
         help='Not running anything for this many seconds constitutes idle',
-        type=int,
+        type=check_positive_int,
         default=0,
     )
     parser.add_argument(
@@ -144,6 +144,18 @@ def parse_args():
     )
 
     return parser.parse_args()
+
+
+def check_positive_int(value):
+    """
+    Check if the CLI argument is a positive integer.
+    :param value: input value from CLI
+    :return:
+    """
+    value = int(value)
+    if value <= 0:
+        raise argparse.ArgumentTypeError('{} is an invalid positive int value'.format(value))
+    return value
 
 
 def connect_to_codalab_server(server, password_file):
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -119,7 +119,7 @@ def parse_args():
     parser.add_argument(
         '--idle-seconds',
         help='Not running anything for this many seconds constitutes idle',
-        type=check_positive_int,
+        type=int,
         default=0,
     )
     parser.add_argument(
@@ -150,7 +150,7 @@ def check_positive_int(value):
     """
     Check if the CLI argument is a positive integer.
     :param value: input value from CLI
-    :return:
+    :return: a positive integer value
     """
     value = int(value)
     if value <= 0:
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -111,10 +111,10 @@ def parse_args():
         help='If specified the worker quits if it finds itself with no jobs after a checkin',
     )
     parser.add_argument(
-        '--exit-number-jobs',
+        '--exit-after-num-runs',
         help='The worker will exit after this many jobs assigned and finished on this worker',
-        type=check_positive_int,
-        default=10000,
+        type=int,
+        default=sys.maxsize,
     )
     parser.add_argument(
         '--idle-seconds',
@@ -144,18 +144,6 @@ def parse_args():
     )
 
     return parser.parse_args()
-
-
-def check_positive_int(value):
-    """
-    Check if the CLI argument is a positive integer.
-    :param value: input value from CLI
-    :return: a positive integer value
-    """
-    value = int(value)
-    if value <= 0:
-        raise argparse.ArgumentTypeError('{} is an invalid positive int value'.format(value))
-    return value
 
 
 def connect_to_codalab_server(server, password_file):
@@ -239,7 +227,7 @@ def main():
         args.work_dir,
         local_bundles_dir,
         args.exit_when_idle,
-        args.exit_number_jobs,
+        args.exit_after_num_runs,
         args.idle_seconds,
         bundle_service,
         args.shared_file_system,
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -136,7 +136,12 @@ def parse_args():
         action='store_true',
         help='Terminate the worker and kill all the existing running bundles.',
     )
-
+    parser.add_argument(
+        '--exit-number-jobs',
+        help='Not running anything for this many jobs assigned to this worker',
+        type=int,
+        default=10000,
+    )
     return parser.parse_args()
 
 
@@ -221,6 +226,7 @@ def main():
         args.work_dir,
         local_bundles_dir,
         args.exit_when_idle,
+        args.exit_number_jobs,
         args.idle_seconds,
         bundle_service,
         args.shared_file_system,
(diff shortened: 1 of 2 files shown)


codalab/worker/main.py
```diff
@@ -140,6 +146,18 @@ def parse_args():
     return parser.parse_args()
 
 
+def check_positive_int(value):
```
candicegjing: This is to make sure the value from CLI is a positive value. Otherwise, the logic in [line 448](https://github.com/codalab/codalab-worksheets/pull/2288/files#diff-c8428f7094434e5ab0590e15a34df0a6R448), needs to be `self.exit_number_jobs<=0`

nelson-liu:
I think this approach is better than #2289, for the reasons you specified. Regarding the restaging issue: there's nothing preventing multiple jobs from being assigned to the worker and trying to run, right? i.e., say you have a worker launched with a run limit of 1 and it already has a job running. The bundlemanager is free to assign that worker another bundle, and might do so—it'd just time out after 5 * 60 seconds, and then go back to the bundlemanager to be restaged. However, is it possible that the bundlemanager will try to run the bundle on that same worker again?

codalab/worker/main.py
```diff
@@ -110,6 +110,12 @@ def parse_args():
         action='store_true',
         help='If specified the worker quits if it finds itself with no jobs after a checkin',
     )
+    parser.add_argument(
+        '--exit-number-jobs',
```
percyliang: In general, use `run` rather than `job`.  I'd suggest `--exit-after-num-runs`

codalab/worker/main.py
```diff
@@ -110,6 +110,12 @@ def parse_args():
...
+        help='The worker will exit after this many jobs assigned and finished on this worker',
+        type=check_positive_int,
+        default=10000,
```
percyliang: Make this `maxint`

codalab/worker/main.py
percyliang: I don't think we need this since I think it's very unlikely someone will pass in a negative value in here.  We don't check this elsewhere.

codalab/worker/worker.py
```diff
@@ -172,6 +174,15 @@ def check_idle_stop(self):
         is_idle = now - self.last_time_ran > self.idle_seconds
         return self.exit_when_idle and is_idle and self.last_checkin_successful
 
+    def check_job_number_stop(self):
```
percyliang: `check_num_runs_stop`

codalab/worker/worker.py
```diff
@@ -89,6 +90,7 @@ def __init__(
         self.shared_file_system = shared_file_system
 
         self.exit_when_idle = exit_when_idle
+        self.exit_number_jobs = exit_number_jobs
```
percyliang: I think it's clearer / safer to leave this as an immutable number and have another value `num_runs` which counts up.

percyliang (changes requested):

codalab/worker/main.py
candicegjing: Removed.

codalab/worker/main.py
candicegjing: Sure. Just updated!

codalab/worker/worker.py
candicegjing: Updated.

codalab/worker/worker.py
candicegjing: Updated.

candicegjing:
> I think this approach is better than #2289, for the reasons you specified. Regarding the restaging issue: there's nothing preventing multiple jobs from being assigned to the worker and trying to run, right? i.e., say you have a worker launched with a run limit of 1 and it already has a job running. The bundlemanager is free to assign that worker another bundle, and might do so—it'd just time out after 5 * 60 seconds, and then go back to the bundlemanager to be restaged. However, is it possible that the bundlemanager will try to run the bundle on that same worker again?

Yes. That's possible the bundle manager will try to run the bundle on the same worker again that just got assigned with a bundle earlier. So it's possible that we will see some frequent state transitions between the `starting` state and the `staged` state.

candicegjing:
> > I think this approach is better than #2289, for the reasons you specified. Regarding the restaging issue: there's nothing preventing multiple jobs from being assigned to the worker and trying to run, right? i.e., say you have a worker launched with a run limit of 1 and it already has a job running. The bundlemanager is free to assign that worker another bundle, and might do so—it'd just time out after 5 * 60 seconds, and then go back to the bundlemanager to be restaged. However, is it possible that the bundlemanager will try to run the bundle on that same worker again?
> 
> Yes. That's possible the bundle manager will try to run the bundle on the same worker again that just got assigned with a bundle earlier. So it's possible that we will see some frequent state transitions between the `starting` state and the `staged` state.

Another thing just comes into my mind is that once this is released, you would probably see more bundles not run in the original order you specified using `request-priority` than before, as the 5 minutes timeout is highly possible to change the execution order.

nelson-liu:
>Another thing just comes into my mind is that once this is released, you would probably see more bundles not run in the original order you specified using request-priority than before, as the 5 minutes timeout is highly possible to change the execution order.


Hmm, that's a good point...I do find the priority feature to be quite useful, personally speaking. For my particular use case, I think it should be ok—most of my jobs fill up an entire worker, so the bundle manager simply won't assign them.

This makes me think of another issue, though. suppose I have a machine with 4 cpus and 1 gpu, and it should only take 1 job. I run a job that requests 3 cpus and 1 gpu. If i run another job that takes 0 gpus, 1 cpu, will that job ever get run? Or will the bundle manager constantly keep assigning it to the occupied worker?

candicegjing:
> > Another thing just comes into my mind is that once this is released, you would probably see more bundles not run in the original order you specified using request-priority than before, as the 5 minutes timeout is highly possible to change the execution order.
> 
> Hmm, that's a good point...I do find the priority feature to be quite useful, personally speaking. For my particular use case, I think it should be ok—most of my jobs fill up an entire worker, so the bundle manager simply won't assign them.
> 
> This makes me think of another issue, though. suppose I have a machine with 4 cpus and 1 gpu, and it should only take 1 job. I run a job that requests 3 cpus and 1 gpu. If i run another job that takes 0 gpus, 1 cpu, will that job ever get run? Or will the bundle manager constantly keep assigning it to the occupied worker?

If the machine with 4 cpus and 1 gpu already took 1 job which requested for 3 cpu and 1 gpu, bundle manager will still try to assign the job with 0 gpu and 1 cpu over to this worker if that's the only worker you have and `reqeust-queue` is the same. So the status of this bundle will consistently switch back and forth between the `staged` state and the `starting` state. Unless you have another worker spawned, then this job will land on the new worker.

nelson-liu:
Thinking about this more, I think I actually would push for design #2. I think that in the current way that the codalab bundle manager is set up (with the bundle manager as the single source of truth of what bundle goes to run on which worker), it makes sense to handle this from the bundle-manager side. Design 1 only works as a side-effect of a failsafe that was put into place to ensure that bundles that never start would eventually run.

Sure, we have to add an extra column to the worker table, but I think that it'll be worth it.

percyliang:
That's a fair point - design #2 seems good then.

candicegjing:
Sure. I am closing this PR now. Will move to #2289

PR #2289, opened 2020-05-22
Add a CLI option to limit the number of jobs allowed to run on a worker 

candicegjing:
This is a partial feature request #1019 . More details are documented at #2287. 
This is an alternative implementation to #2288. Please refer to #2288 for more details.


percyliang:
I prefer design 1 because it doesn't involve schema changes and is isolated to the worker.

candicegjing:
@percyliang and @nelson-liu This is ready for another review. 
I later realized that to make this function work, design 1 and design 2 are both needed due to the fact that 
1. We don't want to over-assign bundles to a worker that has already accepted the number of bundles specified in `exit-after-num-runs`. This will avoid the following two cases:
      a. intermediate state transition between the `starting` state and the `preparing` state
      b. mess up the job priority
1. We want to stop the worker once it reaches its value of `exit-after-num-runs`

codalab/model/worker_model.py
```diff
@@ -188,13 +190,36 @@ def get_workers(self):
...
             }
             for row in worker_rows
         }
         for row in worker_run_rows:
             worker_dict[(row.user_id, row.worker_id)]['run_uuids'].append(row.run_uuid)
         return list(worker_dict.values())
 
+    def update_workers(self, user_id, worker_id, update):
+        """
+        Returns information about all the workers in the database. The return
```
nelson-liu: This doesn't just return information, but also updates the table if an update is provided, right? Might be worth updating the docstring in this case.

nelson-liu (commented):
LGTM, should we add this to the WorkerManager in this PR or in a separate PR?

codalab/model/worker_model.py
candicegjing: Yup. Updated!

PR #2292, opened 2020-05-23
Fix terminate loop (by Jane)

nelson-liu:
@candicegjing I realized that you never merged this in.
For context, without this patch, workers stop as soon as they're started if `--terminate` is true.


candicegjing:
Thanks @nelson-liu! I just realized that I added this fix into an old PR https://github.com/codalab/codalab-worksheets/pull/2186 which hasn't been merged yet. I should have split this into a smaller PR as you did here and get it merged earlier. Anyhow, thanks for this and I am going to merge this once the build is done.

nelson-liu:
oops, i didn't realize, sorry about that! thanks for taking care of this.

PR #2295, opened 2020-05-24
Mounting dependencies on paths specified by keys

### LINKED ISSUE #1115: allow mounting dependencies at not just single level
Currently, you can only do `cl run key:value ...`, where `key` is the name of the file/directory that appears in the current directory in the run.  Make it so that `key` can be hierarchical (e.g., `data/mnist`) to allow for a more flexible structure that matches what users do locally outside CodaLab.

Bonus: allow `key` to be '.', which mounts all the contents of a bundle (usually a git repo) into the current directory.  Only trickiness is that we need to identify all new outputs as the outputs of the run bundle.

The overall goal is to make it feel more like you're just operating in a directory.  If we allow `cl run` to take a list of possible output files (which can be nested like `data/mnist/postprocessed`), and allow multiple bundles to be keyed by '.', then we simulate a mutable file system with layers.

`cl run .:git-repo -n results1 ...`
`cl run .:git-repo .:results1 -n results2 ...`
...

This is certainly not the most ideal way to highlight the dependencies, but it makes it easy for people.

teetone:
Resolves #1115 

- Support `.`, which will mount all content on the top-level of the dependencies path
- Support custom paths via keys
- We currently don't support paths to be an ancestor of another path since we mount file-systems to be read-only


codalab/lib/bundle_cli.py
```diff
@@ -600,20 +600,40 @@ def resolve_key_targets(self, client, worksheet_uuid, target_specs):
         Helper: target_specs is a list of strings which are [<key>]:<target>
         Returns: [(key, worker.download_util.BundleTarget), ...]
         """
-        keys = set()
+
+        def is_ancestor(path1, path2):
+            # Checks if a path is an ancestor of the other path (e.g. foo and foo/bar)
```
percyliang: `"""Return whether `path1` is an ancestor of `path2`"""`

codalab/lib/bundle_cli.py
```diff
@@ -600,20 +600,40 @@ def resolve_key_targets(self, client, worksheet_uuid, target_specs):
...
+            i = 0
```
percyliang: I think we simplify the implementation with something like `path2.startswith(path1 + '/')`

codalab/worker/worker_run_state.py
```diff
@@ -161,6 +165,25 @@ def _transition_from_PREPARING(self, run_state):
             - Start the docker container
         4- If all is successful, move to RUNNING state
         """
+
+        def symlink(target, source):
+            try:
+                os.symlink(target, source)
+            except OSError as e:
+                # Override the existing symlink if one already exists
```
percyliang: Why are all these cases necessary now?

tests/files/dir3/dir1/f1
```diff
@@ -0,0 +1 @@
+first nested file
```
percyliang: Make sure these files have a trailing newline

codalab/worker/worker_run_state.py
```diff
@@ -264,10 +289,41 @@ def _transition_from_PREPARING(self, run_state):
...
+                # Process top-level content for the current dependency path
+                for content in os.listdir(dependency_path):
+                    content_child_path = os.path.normpath(
+                        os.path.join(run_state.bundle_path, content)
+                    )
+                    content_dependency_path = os.path.join(dependency_path, content)
+                    docker_dependency_path = os.path.join(docker_dependencies_path, content)
+
+                    # Set up a symlink from the content child path to the container mount location
+                    try:
```
percyliang: Could we try to refactor this a bit - the logic for symlinking/mounting could be a function, so that this part is just looping over the contents of `dependency_path` and calling this one function.

codalab/worker/worker_run_state.py
```diff
@@ -264,10 +289,41 @@ def _transition_from_PREPARING(self, run_state):
...
+                        symlink(docker_dependency_path, full_child_path)
+                    except OSError as e:
+                        return run_state._replace(
+                            stage=RunStage.CLEANING_UP, failure_message=str(e)
+                        )
+
+            if dep.child_path == '.':
+                # Process top-level content for the current dependency path
+                for content in os.listdir(dependency_path):
```
percyliang: I'd use `child` or `name` instead of `content`

codalab/worker/worker_run_state.py
```diff
@@ -446,6 +502,9 @@ def _transition_from_CLEANING_UP(self, run_state):
                     time.sleep(1)
 
         for dep in run_state.bundle.dependencies:
+            if dep.child_path == '.':
+                continue
+
             dep_key = DependencyKey(dep.parent_uuid, dep.parent_path)
             if not self.shared_file_system:  # No dependencies if shared fs worker
                 self.dependency_manager.release(run_state.bundle.uuid, dep_key)
```
percyliang, line 520: Why is this exempt for `.` dependencies?  Seems like it's still a dependency and the only thing that's affected is the mounting behavior.

percyliang (commented):
Logic looks good overall!  I'd try to make the code a bit more modular - the logic for attaching mounting a path within a dependency to some path versus how we interpret `.`.  Note that if we wanted to support having keys `foo` and `foo/bar`, then ideally we'd use the same logic for how we deal with `.`: just mount everything under `foo`.

codalab/worker/worker_run_state.py
```diff
@@ -264,10 +289,41 @@ def _transition_from_PREPARING(self, run_state):
...
                 )
-                os.symlink(docker_dependency_path, full_child_path)
-            # These are turned into docker volume bindings like:
-            #   dependency_path:docker_dependency_path:ro
-            docker_dependencies.append((dependency_path, docker_dependency_path))
+
+                if dep.child_path != '.':
+                    # A child path can be a nested path which we need to create directories for
+                    Path(full_child_path).mkdir(parents=True, exist_ok=True)
+                    try:
```
candicegjing: Just curious how we think about refactoring file operations from using `os` to `pathlib`?  `pathlib` seems to be more convenient and compact and easier to read. We might be able to remove a lot of extra code by using `pathlib`. If we think this is something worth doing, we could file a separate ticket?

codalab/lib/bundle_cli.py
teetone: Your suggestion was much simpler.

codalab/worker/worker_run_state.py
teetone: I prefer `pathlib` too. It provides more utility than `os` does.

codalab/worker/worker_run_state.py
teetone: After some more investigation, I found out that I wasn't cleaning up all the content that was mounted at the top. I fixed this in my recent commit.

codalab/worker/worker_run_state.py
teetone: I initially had this helper function because of the `FileExistsError` when creating symlinks for nested paths. I figured out that a better way to do this is to create all the directories up to the parent of the path and then create a symlink for the path itself.

teetone:
> Logic looks good overall! I'd try to make the code a bit more modular - the logic for attaching mounting a path within a dependency to some path versus how we interpret `.`. Note that if we wanted to support having keys `foo` and `foo/bar`, then ideally we'd use the same logic for how we deal with `.`: just mount everything under `foo`.

@percyliang I fixed the issue I was having with shared filesystems and added `test2` to  the shared filesystems run of our Travis build. I also did some refactoring and made the logic a bit more modular.

codalab/lib/bundle_cli.py
```diff
@@ -600,20 +600,33 @@ def resolve_key_targets(self, client, worksheet_uuid, target_specs):
         Helper: target_specs is a list of strings which are [<key>]:<target>
         Returns: [(key, worker.download_util.BundleTarget), ...]
         """
-        keys = set()
+
+        def is_ancestor(path1, path2):
```
percyliang: `is_ancestor_or_descendant` is probably clearer.  Note that if `path1 == path2`, we return false.  Is that the right behavior?

codalab/worker/worker_run_state.py
```diff
@@ -438,6 +480,13 @@ def _transition_from_CLEANING_UP(self, run_state):
             move to UPLOADING_RESULTS state
            Otherwise move to FINALIZING state
         """
+
+        def remove_path_helper(path):
```
percyliang: `remove_path_no_fail`?

codalab/worker/worker_run_state.py
```diff
@@ -457,15 +506,20 @@ def _transition_from_CLEANING_UP(self, run_state):
...
             if not self.shared_file_system:  # No dependencies if shared fs worker
+                dep_key = DependencyKey(dep.parent_uuid, dep.parent_path)
                 self.dependency_manager.release(run_state.bundle.uuid, dep_key)
 
-            child_path = os.path.join(run_state.bundle_path, dep.child_path)
-            try:
-                remove_path(child_path)
-            except Exception:
-                logger.error(traceback.format_exc())
+            if dep.child_path == RunStateMachine._CURRENT_DIRECTORY:
```
percyliang: Could we avoid having this logic duplicated here by having the first time when we are mounting things produce a list of things to remove (like `to_mount`)?  Then, we just need to iterate over that list here.

codalab/lib/bundle_cli.py
teetone: This should be okay because we check for duplicate keys before this check.

codalab/worker/worker_run_state.py
teetone: Yes, that was much cleaner. I updated this PR.

candicegjing (approved):

teetone:
@percyliang @candicegjing I modified the PR to support absolute paths. When a key is an absolute path, I ended up removing the leaf to clean up the dependency that is mounted.

tests/files/dir3/dir1/f1
```diff
@@ -0,0 +1,2 @@
...
```
percyliang, line 1: nit: remove extra newline

tests/files/dir3/f1
```diff
@@ -0,0 +1,2 @@
+first file in dir3
```
percyliang, line 1: nit: remove extra newline

percyliang (approved):
Great!

PR #2306, opened 2020-05-27
More resource allocation options for AWS Batch Worker Manager

nelson-liu:
We have heterogeneous AWS compute environments with multiples families of nodes (e.g., p2, g4dn, p3). Sometimes, it's useful to specify to specify the number of gpus your job needs because it makes resource allocation more efficient. E.g., if I want 1 GPU, but 100GB of memory, I'd rather my job land on a g4dn.8xlarge (128 GB RAM, 1 GPU), than a p2.8xlarge (8 GPUs, 488 GB of RAM).

This also adds a flag to control each worker's `--max-work-dir-size`

see: https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html (`resourceRequirements`)

### DIFF
--- a/codalab/worker_manager/aws_batch_worker_manager.py
+++ b/codalab/worker_manager/aws_batch_worker_manager.py
@@ -28,6 +28,9 @@ class AWSBatchWorkerManager(WorkerManager):
             '--cpus', type=int, default=1, help='Default number of CPUs for each worker'
         )
         subparser.add_argument(
+            '--gpus', type=int, default=0, help='Default number of GPUs to request for each worker'
+        )
+        subparser.add_argument(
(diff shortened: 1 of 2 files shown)


nelson-liu:
FWIW: I was able to successfully launch a worker on AWS batch after these changes / adding `--gpus 1`

percyliang (approved):
Great, thanks!

nelson-liu:
I also added a flag to control each worker's `--max-work-dir-size`, in case you wanted to take another look @percyliang

nelson-liu:
Hm, not sure why the last CI run failed when it was just a lint change. Maybe someone can rerun it?

PR #2310, opened 2020-05-27
Close editor before reloading after saving the text editor

### LINKED ISSUE #2301: markdown replicates
To reproduce, create a new worksheet, add a markdown block, add a run bundle.  The markdown block is inserted again at the end.
Screenshot from 2020-05-26 21-02-44

yipenghe:
fix #2301
<strike>The unbinding didn't work properly originally, the ctrl+enter can still invoke the last textbox we inserted, which is why we get duplicate markdowns</strike>
See comments below


epicfaace (commented):
Is it OK that now `Mousetrap.unbindGlobal(['ctrl+enter']);` will not be called if `this.saveText()`'s network call fails?

And to be clear -- was the issue previously that `Mousetrap.unbindGlobal(['ctrl+enter']);`was called too early (i.e. before the ajax request could complete)? how exactly does that cause duplicate markdown blocks?

yipenghe:
> Is it OK that now `Mousetrap.unbindGlobal(['ctrl+enter']);` will not be called if `this.saveText()`'s network call fails?
> 
> And to be clear -- was the issue previously that `Mousetrap.unbindGlobal(['ctrl+enter']);`was called too early (i.e. before the ajax request could complete)? how exactly does that cause duplicate markdown blocks?

Actually, after thinking more about it, it's because of the order of closeEditor and reload worksheet.
Originally, we first unbind, then the ajax call success and reloads, at this point, the textEditor was still open, so the reload again triggered the binding (so the binding was called twice in total), then we close the editor and the binding was still there. The original fix actually moves the unbind after the reload. (bind-unbind-bind)

I changed the order to first closeEditor and then reload, this way the reload will not rerender the editor box and therefore won't bind the ctrl+enter again, so we only do bind once.

epicfaace (approved):

PR #2312, opened 2020-05-27
Add --worker-work-dir-prefix to WorkerManager

nelson-liu:
Allows a user to specify where workers should write their working files.

ref: https://github.com/codalab/codalab-worksheets/pull/2279#discussion_r431369118

### DIFF
--- a/codalab/worker_manager/aws_batch_worker_manager.py
+++ b/codalab/worker_manager/aws_batch_worker_manager.py
@@ -67,9 +67,11 @@ class AWSBatchWorkerManager(WorkerManager):
         image = 'codalab/worker:' + os.environ.get('CODALAB_VERSION', 'latest')
         worker_id = uuid.uuid4().hex
         logger.debug('Starting worker %s with image %s', worker_id, image)
-        work_dir = '/tmp/cl_worker_{}_work_dir'.format(
-            worker_id
-        )  # This needs to be a unique directory since Batch jobs may share a host
+        work_dir_prefix = (
+            self.args.worker_work_dir_prefix if self.args.worker_work_dir_prefix else "/tmp/"
+        )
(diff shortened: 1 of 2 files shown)


codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -67,9 +67,11 @@ def start_worker_job(self):
         image = 'codalab/worker:' + os.environ.get('CODALAB_VERSION', 'latest')
         worker_id = uuid.uuid4().hex
         logger.debug('Starting worker %s with image %s', worker_id, image)
-        work_dir = '/tmp/cl_worker_{}_work_dir'.format(
-            worker_id
-        )  # This needs to be a unique directory since Batch jobs may share a host
+        work_dir_prefix = (
+            self.args.worker_work_dir_prefix if self.args.worker_work_dir_prefix else "/tmp/"
+        )
```
candicegjing, line 72: I am curious about the usage of `/tmp` directory on our cluster.  I did some random checks... Some nodes have only ~100M left, some others have ~20G. Will this cause any problem? As it seems to be a shared directory which can be store files from all users

nelson-liu: The AWS batch workers don't run on the cluster, so it's fine to use `tmp/` here. I've gotten yelled at before for writing too much to `/tmp/` on the cluster, so I wouldn't use that as the default for the slurm worker. This is also why I didn't just make the default `tmp/`, since I know that the default should probably be dependent on what type of worker is being run.

candicegjing: Sounds good!

candicegjing (approved):

PR #2314, opened 2020-05-27
Only show staged bundles from current user in WorkerManager

### LINKED ISSUE #2311: WorkerManager gets staged bundles not owned by the current user
Worker Manager fetches staged bundles not only owned by the current user, but also some other random users'. We should fetch staged bundles only owned by the current user.

nelson-liu:
Fixes https://github.com/codalab/codalab-worksheets/issues/2311


candicegjing:
LGTM!

PR #2317, opened 2020-05-28
Do not wadd bundles in non terminal states

### LINKED ISSUE #2294: Non-final bundles appear as invalid in dest worksheet after wadd
When using `wadd` across instances, some bundles appear as invalid:

Screen Shot 2020-05-22 at 5 09 32 PM

Examining more, it's exactly the bundles in nonfinal states.

Screen Shot 2020-05-22 at 5 10 01 PM

I guess this is because `wadd` saves all of the bundles in the source to the dest, even though not all of them eventually get copied:

https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L3345-L3352

Removing this chunk (and setting `add_to_worksheet=True` in `self.copy_bundle`) would fix this issue, but then the `replace` parameter wouldn't work...

candicegjing:
Fixed #2294 

This PR is to avoid copying bundles in non-terminal states in `wadd` operation.
#### Things to consider
1. Should we avoid copying bundles in the `failed` state? They seem to be not really useful though...
1. Should we add copying bundles in the `staged` state? Bundles in the `staged` state are similar to bundles without metadata. Once it gets copied over to the new instance, it will run after all its dependencies are ready. So would it be helpful for users to preserve their staged bundles?


test_cli.py
```diff
@@ -1434,6 +1434,10 @@ def compare_output_across_instances(command):
         _run_command([cl, 'rm', '-d', uuid])  # Keep only metadata
         _run_command([cl, 'add', 'bundle', uuid, '--dest-worksheet', remote_worksheet])
 
+        # Create at local, transfer to remote (non-terminal state bundle)
```
nelson-liu, line 1440: This comment confused me a bit—could you clarify that you're making the bundle locally, but you want to ensure that it _isn't_ transferred to remote (since it's non-terminal)?

codalab/lib/bundle_cli.py
```diff
@@ -3342,17 +3342,29 @@ def do_wadd_command(self, args):
...
+        valid_source_items = []
         # Save all items to the destination worksheet
         for item in source_items:
+            if item['type'] == worksheet_util.TYPE_BUNDLE:
+                if item['bundle']['state'] not in [State.READY, State.FAILED]:
+                    print(
+                        'Skip bundle {} because it has non-final state {}'.format(
+                            item['bundle']['id'], item['bundle']['state']
+                        ),
+                        file=self.stdout,
```
nelson-liu, line 3355: maybe this should be sent to stderr?

nelson-liu (commented):
Looks like the test is failing because `assert_bundles_ready(source_worksheet)` will never be true since `cl run date --request-gpus 100` will never move past staged?

LGTM though, modulo fixing the test.

codalab/lib/bundle_cli.py
```diff
@@ -3342,17 +3342,29 @@ def do_wadd_command(self, args):
             args.dest_worksheet_spec
         )
 
+        valid_source_items = []
         # Save all items to the destination worksheet
         for item in source_items:
+            if item['type'] == worksheet_util.TYPE_BUNDLE:
+                if item['bundle']['state'] not in [State.READY, State.FAILED]:
+                    print(
+                        'Skip bundle {} because it has non-final state {}'.format(
```
```suggestion
                        'Skipping bundle {} because it has non-final state {}'.format(
```

epicfaace (commented):
Looks good -- can you update documentation so that it's clear that wadd does not apply to bundles in non-terminal states?

test_cli.py
candicegjing: @nelson-liu I have updated the tests and added more comments. Let me know if you think this is clear.

candicegjing:
> Looks good -- can you update documentation so that it's clear that wadd does not apply to bundles in non-terminal states?

Added the documentation.

nelson-liu (approved):
LGTM, thanks!

codalab/lib/bundle_cli.py
```diff
@@ -3311,6 +3311,7 @@ def do_wrm_command(self, args):
         help=[
             'Append all the items of the source worksheet to the destination worksheet.',
             'Bundles that do not yet exist on the destination service will be copied over.',
+            'Bundles in non-terminal states (READY or FAILED) will not be copied over to destination worksheet.'
```
epicfaace: This will add a newline in the generated doc
```suggestion
            'Bundles in non-terminal states (READY or FAILED) will not be copied over to destination worksheet.',
```

epicfaace (approved):

PR #2321, opened 2020-05-28
Don't duplicate sort keys for the same bundle items

### LINKED ISSUE #2232: Sort keys should not be duplicated for the same bundle items
When we have bundles like

```
[]{a}
[]{a}
[]{b}
[]{a}
```

We get the following sort keys: `[132432, 132432, 132431, 132432]`

This is not good, because this makes it impossible to insert an item right after the first row in a table (even with #2190 merged in) -- instead, because all bundle rows referring to "a" have the same sort keys, it will be inserted at the end of the table.

To fix this, we should change the logic for `item_sort_key`:

https://github.com/codalab/codalab-worksheets/blob/5b9390b5ecdead2ef736c4e43dc94037f2aeb161/codalab/objects/worksheet.py#L20-L21

We should also do some sort of migration (either a one-time migration or a dynamic migration upon worksheet load) that would correct these duplicate sort keys for existing worksheets.

epicfaace:
Fixes #2232

The issue was that setting `bundle_info["key"]` was mutating the object, which was shared across multiple bundle items with the same uuid. This PR uses an immutable assignment instead.

It also does this for subworksheet_info (which is not relevant to #2232, but is good practice and consistent anyway).

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -921,11 +921,11 @@ def interpret_items(schemas, raw_items, db_model=None):
                 current_schema = None
 
             if item_type == TYPE_BUNDLE:
-                bundle_info["sort_key"] = sort_key
+                bundle_info = dict(bundle_info, sort_key=sort_key)
                 raw_to_block.append((len(blocks), len(bundle_infos)))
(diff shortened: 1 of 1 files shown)


yipenghe (approved):
Great!!

Issue #1246, opened 2019-06-24
Externalize CodaLab storage


percyliang:
Currently, CodaLab insists on "owning" the storage of all the bundles, and the storage is on the file system.  However, this is not sustainable.  Instead, we'd like the storage of bundles to be decentralized (just like compute is).  So people should be able to host their own bundles (say in s3 or blob storage).  CodaLab simply stores the metadata about that bundle (e.g., size and perhaps a prefix of the bundle), and then when runs depend on this bundle, the bundle can just be downloaded.  We do lose some guarantees on immutability because someone can just delete or modify their bundle.  If we really wanted to, we could compute a hash, or we can just live with the imperfection.

Parts:

- [x] #2322  - Implement --link functionality
- [ ] #2327 - Implement data connectors with Apache Beam

epicfaace:
Discussion from 4/2/2020
- alternative: Azure File Share (but it's not avilable for other browser)
- alternative: pixz
- advantages of pixz: standard library, has compression
- disadvantages of pixz: slightly slower because it uses chunks, when it lists files it cannot list size as well, not sure how extracting files from S3 will work
- conclusion: let's not use these alternatives, stick with the earlier design

nelson-liu:
Thanks for working on this, this would be a great addition! Would it be possible to get some more info on what this "earlier design" looks like?

epicfaace:
Sure. The initial doc where we outlined several options is here: https://docs.google.com/document/d/1E0U9xFopfZ0nc_xNG9w0UAZaGYtDRzZE0oCYn8-Ab_s/edit?usp=sharing

We settled on something similar to Option 2 -- using a tar file on S3, but also storing the metadata (all the header blocks of the tar file concatenated together) in another file that is stored on S3. 

This way, listing the contents of a directory just requires reading the metadata file. Additionally, downloading a single file from the archive just requires 1) reading the metadata to get the tar offset and size of the file, then 2) downloading the actual file itself, by passing in the tar offset and file size in the S3 `Range` header to only download the part of the tar file that we need. 

The point of this approach is to make it quicker / easier to list or download individual files, without necessitating the download of the entire .tar archive.

POC here: https://github.com/epicfaace/s3-tar

epicfaace:
We've had more discussions on this and an update is here (on the same doc): https://docs.google.com/document/d/1hfpfXBvoooIV6_cerKhnDu7X7t9o5hzGD7rtlTlDK_s/edit

[...]

PR #2326, opened 2020-05-28
Implement --link functionality for local filesystems and directories

### LINKED ISSUE #2322: Implement --link functionality
Part of #1246

Rules with shared file system:

- If shared file system and link: server sends worker just the path (/u/nlp/data)
- If shared file system and no link: server sends worker just the path (which must be in codalab bundle store /u/codalab/bundles/…)
- If not shared file system and link: server sends worker the contents (reading from the path)
- If not shared file system and not link: server sends worker the contents (reading from the path)

epicfaace:
Fixes #2322. Also added some TODO comments for places we'll have to change to remove filesystem-specific functionality to fully implement #1246.

## Usage

See the added documentation file for information on how to use `--link`.


codalab/bundles/uploaded_bundle.py
```diff
@@ -18,6 +18,17 @@ class UploadedBundle(NamedBundle):
     METADATA_SPECS.append(
         MetadataSpec('source_url', str, 'URL corresponding to the original source of this bundle.')
     )
+
+    METADATA_SPECS.append(
+        MetadataSpec('is_linked', bool, 'Whether bundle is linked.', generated=True, default=False)
```
percyliang: "Whether bundle is a link"

codalab/bundles/uploaded_bundle.py
```diff
@@ -18,6 +18,17 @@ class UploadedBundle(NamedBundle):
...
+    )
+    METADATA_SPECS.append(
+        MetadataSpec('link_url', str, 'Link URL of bundle.', generated=True)
```
percyliang: I don't think this should be `generated=True` since this is a user-specified field.

codalab/bundles/uploaded_bundle.py
```diff
@@ -18,6 +18,17 @@ class UploadedBundle(NamedBundle):
...
+    )
+    METADATA_SPECS.append(
+        MetadataSpec('link_format', str, 'Link format of bundle. Can be equal to'
+            '"raw" or "zip" (only "raw" is supported as of now).', generated=True)
```
percyliang: Probably remove `generated=True` here too.

codalab/lib/bundle_cli.py
```diff
@@ -1174,6 +1174,15 @@ def do_workers_command(self, args):
                 help='Name of file containing patterns matching files and directories to exclude from upload. '
                 'This option is currently only supported with the GNU tar library.',
             ),
+            Commands.Argument(
+                '-l',
+                '--link',
+                help='Makes the path the source of truth of the bundle, meaning that workers will retrieve the '
```
percyliang: I think it's more that the server retrieves from this bundle.  Where workers get their information is an orthogonal issue: by default we still go through the server, but when shared file system on, then a path (which happens to be the link path) is returned.

codalab/lib/bundle_cli.py
```diff
@@ -1201,8 +1210,21 @@ def do_upload_command(self, args):
...
         }
 
-        # Option 1: Upload contents string
-        if args.contents is not None:
+        # Bundles are not linked by default.
+        bundle_info['metadata']['is_linked'] = False
+
+        # Option 1: --link
+        if args.link:
+            if len(args.path) is not 1:
```
percyliang: `!=`

codalab/lib/bundle_cli.py
```diff
@@ -1201,8 +1210,21 @@ def do_upload_command(self, args):
...
+                raise UsageError("Only a single path can be uploaded when using --link.")
+            bundle_info['metadata']['link_url'] = args.path[0]
+            bundle_info['metadata']['is_linked'] = True
```
percyliang: We don't need this since it be inferred from `link_url`, right?

codalab/lib/download_manager.py
```diff
@@ -104,7 +104,8 @@ def _get_target_info_within_bundle(self, target, depth):
                 )
             )
         elif bundle_state != State.RUNNING:
-            bundle_path = self._bundle_store.get_bundle_location(target.bundle_uuid)
+            bundle = self._bundle_model.get_bundle(target.bundle_uuid)
```
percyliang: This introduces another DB call.  What are the performance implications?  We just need to get a single bit of information, don't need the full bundle technically...

percyliang (commented):
Mostly looks good!

codalab/lib/download_manager.py
epicfaace: I think it's okay. We already make another DB call with `get_bundle_state` earlier in this function, so we're only linearly increasing the runtime.

I'd rather prioritize keeping the changes simple and readable for now. In the future, though, we probably could optimize all calls to get_bundle_location for DB call speed, though. (using a db query similar to the one written in `BundleModel.get_bundle_names`)?

percyliang: `get_bundle` I think could be much heavier because it gets all the metadata too.  Can you profile to see what is the impact on performance?  Some people use CodaLab in a for loop, where the difference matters.  Either way, please make a comment that only the location is needed.

codalab/lib/bundle_cli.py
epicfaace: That's true (it could also be inferred from `link_format`), but I thought it would clearer to have a single field that we could use to check if a bundle is linked or not.

codalab/lib/download_manager.py
epicfaace: I think I'd rather just go ahead and optimize it rather than profile it 😉

codalab/lib/bundle_cli.py
epicfaace: Also, having a separate parameter called is_linked would make it easier to filter bundles -- for example, see this part of the code (which we would likely need to filter bundles that are linked out):

https://github.com/codalab/codalab-worksheets/blob/c9169eb207d5437dd8e98dd11f983269aa32daf8/codalab/model/bundle_model.py#L2502-L2504

codalab/lib/download_manager.py
epicfaace: It also appears we're calling `get_bundle` multiple times in a loop (for things such as resolving parent dependencies), so we should change this.

codalab/lib/bundle_cli.py
epicfaace: I'm just going to remove is_linked for now. Let's re-add it if we find the need for it later.

codalab/lib/bundle_cli.py
```diff
@@ -1174,6 +1174,15 @@ def do_workers_command(self, args):
                 help='Name of file containing patterns matching files and directories to exclude from upload. '
                 'This option is currently only supported with the GNU tar library.',
             ),
+            Commands.Argument(
+                '-l',
+                '--link',
+                help='Makes the path the source of truth of the bundle, meaning that the server will retrieve the '
+                'bundle directly from the specified path rather than downloading the bundle and storing its contents'
```
percyliang: "downloading" might be confusing since it's `cl upload` - just say rather than storing its contents

percyliang (approved):
Great!

codalab/lib/bundle_cli.py
```suggestion
                'bundle directly from the specified path rather than storing its contents'
```

codalab/server/bundle_manager.py
```diff
@@ -289,7 +296,7 @@ def _acknowledge_recently_finished_bundles(self, workers):
                         bundle.uuid, worker['worker_id']
                     )
                 )
-                bundle_location = self._bundle_store.get_bundle_location(bundle.uuid)
+                bundle_location = self._bundle_store.get_bundle_location(bundle=bundle)
```
epicfaace:
```suggestion
                bundle_location = self._bundle_store.get_bundle_location(bundle.uuid)
```

codalab/server/bundle_manager.py
```diff
@@ -518,7 +525,7 @@ def _try_start_bundle(self, workers, worker, bundle, bundle_resources):
             if worker['shared_file_system']:
                 # On a shared file system we create the path here to avoid NFS
                 # directory cache issues.
-                path = self._bundle_store.get_bundle_location(bundle.uuid)
+                path = self._bundle_store.get_bundle_location(bundle=bundle)
```
epicfaace:
```suggestion
                path = self._bundle_store.get_bundle_location(bundle.uuid)
```

codalab/server/bundle_manager.py
```diff
@@ -622,10 +629,10 @@ def _construct_run_message(self, shared_file_system, bundle, bundle_resources):
         message['type'] = 'run'
         message['bundle'] = bundle_util.bundle_to_bundle_info(self._model, bundle)
         if shared_file_system:
-            message['bundle']['location'] = self._bundle_store.get_bundle_location(bundle.uuid)
+            message['bundle']['location'] = self._bundle_store.get_bundle_location(bundle=bundle)
```
epicfaace:
```suggestion
            message['bundle']['location'] = self._bundle_store.get_bundle_location(bundle.uuid)
```

codalab/server/bundle_manager.py
```diff
@@ -622,10 +629,10 @@ def _construct_run_message(self, shared_file_system, bundle, bundle_resources):
...
             for dependency in message['bundle']['dependencies']:
                 dependency['location'] = self._bundle_store.get_bundle_location(
-                    dependency['parent_uuid']
+                    bundle_uuid=dependency['parent_uuid']
```
epicfaace:
```suggestion
                   dependency['parent_uuid']
```

codalab/rest/bundles.py
```diff
@@ -853,11 +857,17 @@ def delete_bundles(uuids, force, recursive, data_only, dry_run):
...
         # check first is needs to be deleted
-        bundle_location = local.bundle_store.get_bundle_location(uuid)
-        if os.path.lexists(bundle_location):
-            local.bundle_store.cleanup(uuid, dry_run)
+        bundle_link_url = bundle_link_urls.get(uuid)
+        if bundle_link_url:
+            # Don't physically delete linked bundles.
+            pass
+        else:
+            bundle_location = local.bundle_store.get_bundle_location(uuid=uuid)
```
```suggestion
            bundle_location = local.bundle_store.get_bundle_location(uuid)
```

epicfaace:
@percyliang I've added a mechanism for security by requiring the user to specify either `--link_allowed_paths` or `CODALAB_LINK_ALLOWED_PATHS` -- please re-review.

codalab/lib/bundle_cli.py
```diff
@@ -1189,6 +1189,17 @@ def do_workers_command(self, args):
...
                 'This option is currently only supported with the GNU tar library.',
             ),
+            Commands.Argument(
+                '-l',
+                '--link',
+                help='Makes the path the source of truth of the bundle, meaning that the server will retrieve the '
+                'bundle directly from the specified path rather than storing its contents'
+                'in its own bundle store. If this argument is used and a local file path is specified,'
+                'the path must match a glob expression specified by the CODALAB_LINK_ALLOWED_PATHS environment variable'
+                'or the --link_allowed_paths command-line parameter to codalab_service.py.',
```
percyliang: Don't include the allowed path comments because this is on the server side, and the user won't be able to control these things anyway.

test_cli.py
```diff
@@ -154,7 +154,12 @@ def wait_for_contents(uuid, substring, timeout_seconds=100):
 
 
 def wait(uuid, expected_exit_code=0):
-    _run_command([cl, 'wait', uuid], expected_exit_code)
+    try:
+        _run_command([cl, 'wait', uuid], expected_exit_code)
+    except AssertionError as e:
+        _run_command([cl, 'info', uuid])
```
percyliang, line 161: Why are we adding this?  What are the errors and is there a better way to handle them?

epicfaace: It's much easier to debug when the bundle fails on CI, because running `cl info` will then show the reason why the bundle has failed. The error is caused when the exit code is not equal to the expected exit code.

This change still causes the same error to be raised, but we just get the results of `cl info` printed to CI before that happens.

percyliang: What do you think of instead passing in a flag to `_run_command` that says if there's a failure to print out information about the bundle?  I'm guessing this type of output might be useful in many cases, not just here.

epicfaace: Interesting idea -- how would we get the bundle uuid though? I think just assuming that the third argument is equal to the bundle uuid may end up not being a good heuristic?

epicfaace:
@percyliang I added documentation -- please review

image

test_cli.py
percyliang: A uuid can be passed in as an optional argument.

docs/Externalization.md
```diff
@@ -0,0 +1,32 @@
...
+
+Compare this operation to a normal upload, in which `/var/data/a.txt` will be uploaded to the server and a copy of it will be stored in the bundle store:
+
+```bash
+cl upload /var/data/a.txt
+```
+
+### Allowed paths
+
+For security reasons, in order for the `--link` command to work with file paths, either the `CODALAB_LINK_ALLOWED_PATHS` environment variable or the `--link_allowed_paths` command-line parameter to `codalab_service.py` must be set on the server. This must be equal to a glob expression that specifies the allowed paths that `--link` can grab data from.
```
percyliang: say you also need `--shared-file-system`?

docs/Externalization.md
```diff
@@ -0,0 +1,32 @@
+# Storage Externalization
+
+## --link
```
percyliang: Make the structure clearer that the first half of the doc is for the user and the second half is for the administrator.  Some more context and motivation is probably required to fully appreciate this - not for the public instance, but for private instances of CodaLab on a shared file system...and this allows you to mutate the underlying files (useful for debugging code).

percyliang (approved):

docs/Externalization.md
epicfaace: Technically, `--link` works whether or not `--shared-file-system` is provided.

docs/Externalization.md
epicfaace: Updated. To be clear in the documentation, I said that `--link` is used for "private instances of CodaLab in which the server and users both have access to the same filesystem", because the usage of `--link` is orthogonal to the `--shared-file-system` tag (which refers to the server and workers sharing a filesystem).

percyliang (approved):
We should consider having the server check that the link is working (both for file paths and URLs).  I can see people uploading accidentally with link from their local laptop and wondering why it's not working.

Anyway, excited to have this feature!

epicfaace:
@teetone can you review this PR regarding security and `CODALAB_LINK_ALLOWED_PATHS`?

epicfaace:
@percyliang I've also added additional documentation to better clarify the interplay between `--link` and `--shared-file-system`:

image

percyliang:
Awesome!

test_cli.py
epicfaace: Let's do this in a separate PR.

epicfaace:
@percyliang ready for review again... there was some tricky logic involving creating the right mounts on docker-compose.yml that I'd like you to take a look at.

epicfaace:
@percyliang just a ping on this 🙂

codalab_service.py
```diff
@@ -293,6 +295,10 @@ def has_callable_default(self):
         default=SERVICE_REQUEST_TIMEOUT_SECONDS,
         help='Docker client timeout (in seconds)',
     ),
+    CodalabArg(
+        name='link_mounts',
+        help='Comma-separated list of directories that are mounted on the CodaLab server, allowing their contents to be used in the --link argument.',
```
percyliang: be more specific: "on the REST server"?

codalab_service.py
```diff
@@ -518,7 +524,26 @@ def __init__(self, args):
...
+            # so we can't just use regular interpolation with environment variables. Instead,
+            # we create a temporary file with the modified docker-compose.yml and use that file instead.
+            with open(os.path.join(self.compose_cwd, 'docker-compose.yml')) as f:
+                compose_options = yaml.load(f)
+            for mount_path in self.args.link_mounts.split(","):
+                mount_path = os.path.abspath(mount_path)
+                compose_options["x-codalab-server"]["volumes"].append(
+                    f"{mount_path}:/opt/codalab-worksheets-link-mounts{mount_path}"
+                )
+            with tempfile.NamedTemporaryFile(mode='w', suffix=".yml", delete=False) as f:
```
percyliang: Can we put this in `var_path` and name it something like `docker-compose-custom.yml`, so that it's easier to find / debug?

docs/Server-Setup.md
```diff
@@ -138,8 +138,10 @@ hour because lots of packages have to be installed):
 ## Testing
 
 Since tests run against an existing instance, make sure you update your instance first.
+You also need to set the CODALAB_LINK_MOUNTS environment variable so that the `link` tests
+can properly succeed.
 
-    ./codalab_service.py start -bd -s rest-server
+    CODALAB_LINK_MOUNTS=/tmp ./codalab_service.py start -bd -s rest-server
```
percyliang: Can we make this default to `var/codalab/link-mounts` or something?
1) /tmp might have stuff from other applications we don't want to expose by default
2) It'd be nicer to have a default so we don't have to worry about setting the environment variable to get sane behavior (e.g., for running tests).

test_cli.py
```diff
@@ -1195,6 +1201,51 @@ def test(ctx):
...
+    _run_command([cl, 'cat', uuid], 1)
+
+    # Upload file
+    # We create the file at /opt/codalab-worksheets-link-mounts/tmp because this test is running
+    # inside a Docker container (so the host directory /tmp is mounted at /opt/codalab-worksheets-link-mounts/tmp).
+
+    with tempfile.NamedTemporaryFile(
+        mode='w', dir='/opt/codalab-worksheets-link-mounts/tmp', suffix=".txt", delete=False
+    ) as f:
+        f.write("hello world!")
```
percyliang, line 1226: nit: can we end with a new line (use print)?

epicfaace: Wouldn't using print to a file make this far more complicated (because it requires redirecting stdout)? see https://stackabuse.com/writing-to-a-file-with-pythons-print-function/

codalab_service.py
epicfaace: Sounds good -- hopefully this won't cause issues with people concurrently running codalab_service.py, right? (as long as they specify different instance names, it should be fine, I guess)

docs/Server-Setup.md
epicfaace: Done

PR #2330, opened 2020-05-28
Add boto3 to requirements.txt

nelson-liu:
Otherwise, `pip install -e .` on a freshly cloned repo, followed by `cl-worker-manager -h` yields:
```
Traceback (most recent call last):
  File "/juice/scr/nfliu/miniconda3/envs/cl_dev/bin/cl-worker-manager", line 11, in <module>
    load_entry_point('codalab', 'console_scripts', 'cl-worker-manager')()
  File "/juice/scr/nfliu/miniconda3/envs/cl_dev/lib/python3.6/site-packages/pkg_resources/__init__.py"
, line 490, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/juice/scr/nfliu/miniconda3/envs/cl_dev/lib/python3.6/site-packages/pkg_resources/__init__.py"
, line 2853, in load_entry_point
    return ep.load()
  File "/juice/scr/nfliu/miniconda3/envs/cl_dev/lib/python3.6/site-packages/pkg_resources/__init__.py"
, line 2453, in load
    return self.resolve()
  File "/juice/scr/nfliu/miniconda3/envs/cl_dev/lib/python3.6/site-packages/pkg_resources/__init__.py"
, line 2459, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/juice/scr/nfliu/git/codalab-worksheets/codalab/worker_manager/main.py", line 7, in <module>
    from .aws_batch_worker_manager import AWSBatchWorkerManager
  File "/juice/scr/nfliu/git/codalab-worksheets/codalab/worker_manager/aws_batch_worker_manager.py", l
ine 2, in <module>
    import boto3
ModuleNotFoundError: No module named 'boto3'
```

PR #2331, opened 2020-05-28
Add nicer error message when running workermanager w/o boto3

nelson-liu:
Also removed 2 unused imports, sorry for polluting the diff.

### DIFF
--- a/codalab/worker_manager/aws_batch_worker_manager.py
+++ b/codalab/worker_manager/aws_batch_worker_manager.py
@@ -1,6 +1,9 @@
-import argparse
-import boto3
-import json
+try:
+    import boto3
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "Running the worker manager requires the boto3 module.\n"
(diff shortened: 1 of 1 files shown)


percyliang (approved):
Thanks!

candicegjing:
Thanks for making the changes! I am wondering if we would consider adding `boto3` as part a requirement when installing `pip3 install codalab`? Since the worker manager could have more users in the future, manually configuring the required packages might not be ideal. But let me know if I missed anything here.

nelson-liu:
@candicegjing percy had some reasoning in https://github.com/codalab/codalab-worksheets/pull/2330 , and I overall agree . It's a pretty heavyweight dependency for something that I would guess _most_ users who `pip install codalab` won't actually need.

candicegjing:
@nelson-liu I see. That makes sense then!

candicegjing:
@nelson-liu build is failing, can you black the format? Then I'll merge it.

nelson-liu:
oops, done

PR #2336, opened 2020-05-30
Refactor bundle manager

candicegjing:
Just to refactor and clean up unused functions/ merge duplicate logic.

1. create `user_parallel_run_quota_left` at the beginning of the main dispatching for-loop to track the number of parallel run quota left for each user.
1. In the for-loop:
     1. I just check if a user’s parallel run quota >= 0. If so, then this user can use `codalab_owned_worker`. If not, the `codalab_owned_worker` is not allowed to use.
     1. combine the `codalab_owned_worker` with the `user_owned_worker` as the list of workers that the current bundle can be assigned to as `workers_list`
     1. Deducting computing resources (cpu, gpu, memory) on this `workers_list`
     1. Filter out workers that don’t have enough resources for the run. 
     1. dispatch the bundle to run on a worker and deduct the `user_parallel_run_quota_left` if the assigned worker is a `codalab_owned_worker`
1. Removed refresh function in the `WorkerInfoAccessor`, as we want to keep the worker state immutable during each dispatching cycle.
1. _**Most importantly, this PR will fix the issue of omitting newly dispatched runs in the resource deduction phase of the current dispatching logic.**_


candicegjing:
@nelson-liu I can't at you from the reviewers list, so I just at you here. Since you worked on this last time, could you please review this PR?

codalab/server/bundle_manager.py
```diff
@@ -363,48 +363,37 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
             for worker in workers_list:
                 if self._try_start_bundle(workers, worker, bundle, bundle_resources):
                     # If we successfully started a bundle on a codalab-owned worker,
                     # decrement the parallel run quota left.
                     if worker["user_id"] == self._model.root_user_id:
                         user_parallel_run_quota_left[bundle.owner_id] -= 1
-                    # Update available worker resoures. This is a lower-bound,
-                    # since resources released by jobs that finish are not used until
-                    # the next call to _schedule_run_bundles_on_workers.
-                    worker['cpus'] -= bundle_resources.cpus
```
nelson-liu, line 406: Hm, why was this block removed? Won't workers be overscheduled as a result?

candicegjing: So this block of code can be refactored into `deduct_bundle_resources()` a few lines above directly. This won't cause overschedule as the resource will be deducted in the next loop when dispatching a new bundle.

candicegjing: Also I just updated the ticket description (which I should have done earlier) for clarification

teetone (approved):

nelson-liu:
Thinking more about it, I think we want to be keeping the `@refresh` in the WorkerInfoAccessor—it's used in more places than this function, after all?

candicegjing:
> Thinking more about it, I think we want to be keeping the `@refresh` in the WorkerInfoAccessor—it's used in more places than this function, after all?

What are the other places you mean? Could you point them out? 

The main reasons that I remove it are as follows: 
1. With the current logic of caching all workers (`resource_deducted_user_owned_workers` and `resource_deducted_codalab_owned_workers`), refreshing logic doesn't really have any real effect on dispatching logic.
1. I think we really just want to fetch the worker once for each scheduling loop to ensure that there aren't any changes in the middle of the dispatching logic (although this is conservative, I guess it's safer).
1. I think in our distributed system, there are just too many independent clock cycles that operate on their own which is hard to control sometimes. I want to minimize this as much as possible so that it'll be easier to maintain and reasoning in the future.

But let me know your thoughts on why you want to keep this.

nelson-liu:
> 1. With the current logic of caching all workers (resource_deducted_user_owned_workers and resource_deducted_codalab_owned_workers), refreshing logic doesn't really have any real effect on dispatching logic.

Yeah, I think unfortunately, we need the refreshing logic to have more effect on the dispatching logic. For instance, at the moment, we try to run bundles on workers that are no longer connected (see #2423) — in general, it feels like our knowledge about workers should be as up-to-date as possible when scheduling bundles.

> 2. I think we really just want to fetch the worker once for each scheduling loop to ensure that there aren't any changes in the middle of the dispatching logic (although this is conservative, I guess it's safer).

I agree. But I think part of the issue is that we actually don't want to be ignoring all changes, we only want to be ignoring changes that increase the resources we have (i.e., new workers, or new resources on workers). However, we still need to know about changes that decrease the resources we have (e.g., workers going offline).

Thinking more about it, the logic for removing workers doesn't rely on the refresh logic, so it could be ok to remove it...the case where a worker would decrease in resources is fairly unlikely (i.e., a user quits a worker and then immediately reconnects it with fewer resources, like 1 fewer GPU.)

nelson-liu:
Thinking more about this: I think this fix is good for the intended purpose (simplifying the scheduling logic, especially as it relates to respecting bundle priority).

However, doesn't it, in theory, revert the original motivation of refreshing the WorkerInfoAccessor cache ( https://github.com/codalab/codalab-worksheets/pull/1977 )? Refreshing seems necessary to prevent marking bundles as offline in the case where the bundle manager loop takes longer than one worker timeout cycle (which I've been running into in #2423 ).

In practice, this change is still probably ok, since the function that takes a long time (`self._schedule_run_bundles_on_workers`) is run at the very end, so by the time we reach the next iteration, things will be refreshed already (since we started a new BundleManager loop).

candicegjing:
> Yeah, I think unfortunately, we need the refreshing logic to have more effect on the dispatching logic. For instance, at the moment, we try to run bundles on workers that are no longer connected (see #2423) — in general, it feels like our knowledge about workers should be as up-to-date as possible when scheduling bundles.

I doubt about this, because `resource_deducted_user_owned_workers` and `resource_deducted_codalab_owned_workers` are not the original `WorkerInfoAccessor` object. They are new copies of the actual worker object. So refreshing on `WorkerInfoAccessor` object won't take any real effect on the dispatching logic.

> I agree. But I think part of the issue is that we actually don't want to be ignoring all changes, we only want to be ignoring changes that increase the resources we have (i.e., new workers, or new resources on workers). However, we still need to know about changes that decrease the resources we have (e.g., workers going offline).

We definitely don't want to ignore all changes. The issue here is that even with the refreshing logic, I don't think it can actually solve the problem from the root. I am trying to figure out a better way to handle this...

percyliang:
Need to think more about the invariances in the bundle manager, closing for now.

PR #2343, opened 2020-06-02
Fix logging of active workers in SlurmWorkerManager

nelson-liu:
this always prints `"(none)"` right now, because we're fetching both the run ID and the status, and then checking if the `runid,status` string is in `self.submitted_jobs`. It isn't (since the status is tacked on), so this removes the status from the output and consequently fixes the logging.


candicegjing (approved):
Thanks for fixing this!

PR #2344, opened 2020-06-02
Enable excluding patterns in run bundle uploaded output

### LINKED ISSUE #2338: Optionally prevent upload of files larger than a threshold in failing bundles
Sometimes, bundles generate massive amounts of intermediate output. For instance, training huggingface transformers squad models can generate 10+GB worth of cached features for the dataset. These are easily removed at the end of a successful run, but sometimes things go awry. The worker will then try to upload this absolutely massive bundle, and inevitably get stuck doing so—I have to go and manually kill it, clear the work-dir, and restart it to get it back up.

It'd be nice to have some way of saying, at cl run, "if this run fails, don't upload files that are individually larger than 2GB" or something to that effect.

nelson-liu:
Fixes #2338 

Adds a new bundle metadata field `upload_exclude_patterns`, which are applied when `tar`ing the file for upload.

~~I haven't actually tested this, but I'm assuming that `bundle_service_client#update_bundle_contents()` will update the paths in the bundle according to the `tar`'d archive that's uploaded, and the `data_size` will be computed accordingly?~~
update: added tests

### DIFF
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -32,6 +32,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_queue', str, 'Submit run to this job queue.', hide_when_anonymous=True, default=None))
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important).', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(MetadataSpec('upload_exclude_patterns', str, 'Exclude these file patterns when uploading run output.', default=None))
 
     METADATA_SPECS.append(MetadataSpec('actions', list, 'Actions (e.g., kill) that were performed on this run.', generated=True))
 
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -32,7 +32,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_queue', str, 'Submit run to this job queue.', hide_when_anonymous=True, default=None))
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important).', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
-    METADATA_SPECS.append(MetadataSpec('upload_exclude_patterns', str, 'Exclude these file patterns when uploading run output.', default=None))
+    METADATA_SPECS.append(MetadataSpec('exclude_patterns', str, 'Exclude these file patterns when uploading run output.', default=None))
 
     METADATA_SPECS.append(MetadataSpec('actions', list, 'Actions (e.g., kill) that were performed on this run.', generated=True))
 
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -32,7 +32,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_queue', str, 'Submit run to this job queue.', hide_when_anonymous=True, default=None))
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important).', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
-    METADATA_SPECS.append(MetadataSpec('exclude_patterns', list, 'Exclude these file patterns when uploading run output.', generated=True))
+    METADATA_SPECS.append(MetadataSpec('exclude_patterns', list, 'Exclude these file patterns when uploading run output.', default=[]))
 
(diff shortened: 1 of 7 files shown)


codalab/bundles/run_bundle.py
```diff
@@ -32,6 +32,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_queue', str, 'Submit run to this job queue.', hide_when_anonymous=True, default=None))
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important).', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(MetadataSpec('upload_exclude_patterns', str, 'Exclude these file patterns when uploading run output.', default=None))
```
percyliang: `upload` might be a bit confusing, because from a user's perspective, upload means `cl upload`.  Maybe just `exclude_patterns`?

percyliang (approved):
Looks pretty good to me!  Adding a test (in `test_cli.py`) would probably be a good idea.

candicegjing (approved):
LGTM! And I agree that adding a test will be helpful.

nelson-liu:
Hm, I actually realized that this might not work if you have multiple `--exclude-patterns`. In the upload CLI, the argparse argument has `nargs='*'`, so it can take multiple `--exclude-patterns`. How do i do the same for the run CLI?

Storing it as a list might be problematic, though, since the bundle metadata fields can only be strings...I guess I see two ways forward:

1. accept a list in the CLI, serialize it to a string, and then de-serialize when we're about to upload
2. accept a string in the cli (with some delimiter), and split on that delimiter when we're about to upload.

I think (1) is better to match with the functionality of `tar`, but I'm not sure how to make the CLI take in a list of strings (equivalent to `nargs='*'`) from just the run bundle metadata specs.

nelson-liu:
> Storing it as a list might be problematic, though, since the bundle metadata fields can only be strings

I was wrong about this. Lists are fine. I changed the metadatafield to a `list`, so things _should_ just work...

nelson-liu:
Ok, tests are passing locally for me now—hopefully the same on CI. I'd love for this to go in 0.5.14 , btw ( @yipenghe , also tagging the PR #2347 )

nelson-liu:
Ok, the tests that I added seem to pass. Looks like the ones failing are the tests that have always been flaky.

nelson-liu:
Hm, I realize that this doesn't handle the case where the worker and the server are on a shared filesystem (since no uploads happen, so it can't be excluded). ~~Should this flag be only for non-shared-filesystem workers? Or should we also remove the written bundle outputs for shared filesystem case?~~

I decided that the behavior of the shared filesystem case should match that of the non-shared-filesystem case, i.e., the bundle paths are deleted when the worker uses a shared filesystem.

nelson-liu:
Ok, all tests are passing here, modulo the flaky `search` test. This is ready for another look.

codalab/bundles/run_bundle.py
```diff
@@ -32,6 +32,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_queue', str, 'Submit run to this job queue.', hide_when_anonymous=True, default=None))
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important).', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(MetadataSpec('exclude_patterns', list, 'Exclude these file patterns when uploading run output.', default=[]))
```
percyliang: Maybe "Exclude these file patterns from being saved into the bundle contents"

codalab/worker/worker_run_state.py
```diff
@@ -471,11 +472,16 @@ def _transition_from_CLEANING_UP(self, run_state):
...
             return run_state._replace(
                 stage=RunStage.UPLOADING_RESULTS, run_status='Uploading results', container=None
             )
         else:
+            # No need to upload results since results are directly written to bundle store
+            # Delete any files that match the exclude_patterns .
+            for exclude_pattern in run_state.bundle.metadata["exclude_patterns"]:
+                full_pattern = os.path.join(run_state.bundle_path, exclude_pattern)
+                for file_path in glob.glob(full_pattern, recursive=True):
+                    remove_path(file_path)
```
percyliang: Is this secure?  what if `exclude_pattern` were `../../..`?

nelson-liu: You're right—I think a check to ensure that the filepath is within `run_state.bundle_path` should do the trick, right?

codalab/worker/file_util.py
```diff
@@ -315,3 +315,27 @@ def remove_path(path):
             shutil.rmtree(path)
         else:
             os.remove(path)
+
+
+def path_is_parent(parent_path, child_path):
+    """
+    Given a parent_path and a child_path, determine if the child path
+    is a strict subpath of the parent_path. In the case that the resolved
+    parent_path is equivalent to the resolved child_path, this funciton returns
```
percyliang: typo

percyliang (approved):
Thanks, LGTM!

PR #2345, opened 2020-06-02
Load jobs by partition in SlurmWorkerManager

nelson-liu:
This way, if you have two `slurmworkermanager` with otherwise identical arguments except for the partition (e.g., one is on `jag-lo`, another is on `jag-standard`), the worker can correctly load existing jobs for each.


candicegjing (approved):
LGTM! This would be helpful!

PR #2351, opened 2020-06-02
Fix Slurm GPU isolation

nelson-liu:
I'm running two workers in a `SlurmWorkerManager`, requesting 1 gpu each. However, `cl workers` reports:
```
$ cl workers
worker_id                         cpus  gpus  memory  free_disk  last_checkin  tag          runs      shared_file_system  tag_exclusive
---------------------------------------------------------------------------------------------------------------------------------------
9714103521d54e038af242d9b1708289  3/4   1/4   125g    801g       0.0s ago      nlpgpu       0xad402e  False               False
a3cb32180f904362a90014dbf84def6e  3/4   1/8   125g    845g       3.0s ago      nlpgpu       0x4e1914  False               False
```

Note that these workers are shown as having 4 and 8 gpus. More broadly, GPU isolation isn't working—we need to explicitly set the GPUSET when running the command. This PR provides a fix, and I've verified that it works.

@candicegjing can you look at this please? I think this is also quite important for 0.5.14 ( @yipenghe )


nelson-liu:
Would it be possible to get some eyes on this? I copied it straight from `/u/nlp/bin/nlp-cl-worker`, and the codalab slurm worker is quite broken without this change.

percyliang:
Looks reasonable.  Should this be something that's not specific to slurm worker manager but any worker?

nelson-liu:
Yeah, let's fold it into the worker. See https://github.com/codalab/codalab-worksheets/pull/2415

percyliang (approved):
Discussed offline, merge this in for now since it's not clear what the general solution should be.

nelson-liu:
just to link things together, the broader issue that requires a general solution is https://github.com/codalab/codalab-worksheets/issues/2661 .

PR #2354, opened 2020-06-03
Pass down max work dir size and delete workdir on exit to slurm workers

nelson-liu:
These were added to the AWS batch worker, but not the slurm worker


candicegjing (approved):
LGTM!

Issue #2327, opened 2020-05-28
Implement data connectors for --link using Apache Beam


epicfaace:
When using --link, we should be able to enter in urls starting with schemes such as gcs://, wasb://, s3://, etc.

For this, we can run Apache Beam pipelines as an abstraction to get the contents from the specified URL.

Part of #1246

percyliang:
Will the first version will allow us to use Azure Blob as a backend with credentials or just to allow public links?

epicfaace:
- Make sure Azure files are deleted upon bundle deletion
- Test with 10 GB uploads

epicfaace:
I'm going to close this issue in favor of issues that delineate clearer subtasks (such as #3188).

PR #2358, opened 2020-06-04
Drop-in Apache Beam FileSystem as a replacement for existing file open(), etc. methods

epicfaace:
For #2327

Merge this only after #2326 is merged


epicfaace:
Superseded by #2769.

PR #2366, opened 2020-06-07
Add CodaLabManagerState class, implement JSON and SQLite3 backends

### LINKED ISSUE #2361: Concurrency issues with state.json can lead to inadvertently changed states
**Describe the bug**
It seems like the session data for the CLI `state.json` is being modified outside of the current session, but I can't figure out where or by what. At a high level, I have a script that does `cl work <something>`, and then a bunch of `cl run` s. But while it's in the middle of those `cl run`s, somehow the `state.json` file gets modified and as a result, some of the runs fail because they can't find bundles that are supposed to be in the worksheet they're in (because the current worksheet for the session was mysteriously changed). As far as I've investigated, i think it's caused by something to do with authentication—whenever `state.json` is modified unintentionally, the token expiry is pushed back.

As an experiment last night, I deleted `state.json` and walked away from my computer. When I came back, `state.json` had been recreated, even though there were no running `cl` processes that I know of (this is on the NLP cluster, so the process could be on any of the head or worker nodes).

## More details:

I have a script that, broadly speaking, does a `cl work <some worksheet name>`, and then `cl run`s hundreds of bundles. This can take quite some time (~30 minutes). In my `cl run`s, I refer to bundles by name (not by uuid).

Sometimes, my `cl run`s fail because of a `NotFoundError`—CodaLab is trying to run the bundle, but it can't resolve the name -> uuid. If i ctrl-C / `cat ~/.codalab/state.json`, I see that the worksheet UUID assigned to my session has been modified! Throughout this whole process, I'm printing out the session ID—it doesn't change whenever I invoke `cl`, as expected (I've also tried setting `CODALAB_SESSION`, to no avail). Doing some extra logging, I noticed the following:

My script will `cl info` a bundle, and it works. Below, I describe the output of the logs.

I print out the state path it's reading from, this is the right state path
```
Reading state_path /sailhome/nfliu/.codalab/state.json
```

I print out the read state. This looks fine. I added an extra `hostname` field to the sessions value, since I suspect something is going on with other nodes sharing the same CODALAB_HOME...
```
State {'auth': {'http://codalab.stanford.edu': {'token_info': {'access_token': 'PGbYhAMOGZYKo2cmyUBPZ4bgqYt6Vf', 'expires_at': 1591337015.2242806, 'refresh_token': 'Ns4pFceba0HLz8MLVBCN7acqZW1ozW', 'scope': 'default', 'token_type': 'Bearer'}, 'username': 'nfliu'}}, 'last_check_version_datetime': '2020-06-05 05:03:35', 'sessions': {'23215': {'address': 'http://codalab.stanford.edu', 'hostname': 'scdt.stanford.edu', 'worksheet_uuid': '0x09d026dfee1646b0a8439587af6752a6'}}}
```

I print out the intermediate variables when `cl info` runs `parse_client_worksheet_uuid`. This all looks fine, and it returns the right worksheet UUID as expected by the state.
```
parse_client_worksheet_uuid#spec None
parse_client_worksheet_uuid#worksheet_util.CURRENT_WORKSHEET .
parse_client_worksheet_uuid#Empty spec, returning current worksheet
session#name 23215
{'address': 'http://codalab.stanford.edu', 'hostname': 'scdt.stanford.edu', 'worksheet_uuid': '0x09d026dfee1646b0a8439587af6752a6'}
```

I make sure to log the returned worksheet id, it's correct.
```
get_current_worksheet_uuid from session 0x09d026dfee1646b0a8439587af6752a6
get_current_worksheet_uuid 0x09d026dfee1646b0a8439587af6752a6
parse_client_worksheet_uuid#retval 0x09d026dfee1646b0a8439587af6752a6
```

However, immediately after, I try to do `cl info` again on a different bundle in the same worksheet, and things fail. It isn't able to find the bundle. The log output is sort of surprising:

It reads the state from the same file, but for some reason, the state has been modified. Now, it has dozens of entries. I'm not sure where they all came from. It's interesting to note that these ones *don't* have the hostname, so they aren't generated at least by my current session running `cl info`.

Also, note that the `expires_at` on the auth token changes. However, I added a logging statement to the `_cache_token` function that 

```
Reading state_path /sailhome/nfliu/.codalab/state.json
State {'auth': {'http://codalab.stanford.edu': {'token_info': {'access_token': '7vgrZa651Tum2qdSuypxlOHvWZtFfb', 'expires_at': 1591338738.799319, 'refresh_token': '18JuckW1gAUjAYDW2IJHoje2UZMw8J', 'scope': 'default', 'token_type': 'Bearer'}, 'username': 'nfliu'}, 'https://worksheets.codalab.org': {'token_info': {'access_token': 'y1W0g8x3phprj4cs0BODGrSHMMygkT', 'expires_at': 1591162591.4494832, 'refresh_token': 'FvGmYaONC9g9pWC9npOPhpDsLsxkEA', 'scope': 'default', 'token_type': 'Bearer'}, 'username': 'nfliu'}}, 'last_check_version_datetime': '2020-06-05 04:48:06', 'sessions': {'10098': {'address': 'https://worksheets.codalab.org', 'worksheet_uuid': ''}, ... <elided other sessions> ..., '23215': {'address': 'http://codalab.stanford.edu', 'worksheet_uuid': '0x6f70bd37400345c3933a17f413de87cc'}, ... <elided other sessions> ...}}
```

```
parse_client_worksheet_uuid#spec None
parse_client_worksheet_uuid#worksheet_util.CURRENT_WORKSHEET .
parse_client_worksheet_uuid#Empty spec, returning current worksheet
session#name 23215
{'address': 'http://codalab.stanford.edu', 'worksheet_uuid': '0x6f70bd37400345c3933a17f413de87cc'}
get_current_worksheet_uuid from session 0x6f70bd37400345c3933a17f413de87cc
get_current_worksheet_uuid 0x6f70bd37400345c3933a17f413de87cc
parse_client_worksheet_uuid#retval 0x6f70bd37400345c3933a17f413de87cc
NotFoundError: bundle spec run-rasor-synthetic_cloze_worddropout.50000-train-sampled_hps_d61e287707c6c2f3029a3da3721863c1 doesn't match any bundles
```

**To Reproduce**
I'm working on a minimal reproducible example, but I think it's somewhat machine/setup specific...

nelson-liu:
Fixes #2361 

Prior to this PR, the CodaLab manager state was a dictionary that was read from disk when the `CodaLabManager` is instantiated, and then rewritten to disk when important changes are made. This is unideal because:

(1) There is no protection against concurrent reads / writes.
(2) When the state is re-committed to disk, the _entire_ file is rewritten, rather than just incrementally changing the parts that matter.

These two can interact poorly when there are multiple `cl` processes sharing the same state file. For instance, there's nothing stopping 1 `cl` process from trying to read the state _right_ at the moment where another clears it out in the process of writing.

Rather than implement some complicated locking mechanism, I decided to just abstract out the state into a new `CodaLabManagerState` class, and add concrete implementations of a `CodaLabManagerJsonState` and a `CodaLabManagerSqlite3State`.
- The `CodaLabManagerJsonState` is functionally the same as what's currently in the codebase, and uses a JSON file on disk to store the state. 
- The `CodaLabManagerSqlite3State` implements the same interface, but uses a on-disk sqlite3 database to store the state. This is nice because we get incremental updates (i.e., you can change some rows of a particular table without destroying and rebuilding the entire db) and robustness to concurrency (it's single-writer, multiple-reader) [1].

This is all controlled by a new config key called `state_backend` (that lives in `config.json`). The default value when codalab is installed or the key is not specified is `json`. End users shouldn't see any changes in functionality unless they explicitly opt-in to the sqlite3 state backend by running `cl config state_backend sqlite3`.

I've been running this locally with the `sqlite3` backend for the past few days, and the issue in #2361 has gone away. I've also been able to run many concurrent cl processes in parallel with no issues (I set `$CODALAB_SESSION` to a random UUID for each of these, just to ensure there's no clashing, and it works fine).

To test this, I changed the Travis config to rerun all of the unit tests under each of the backends (`json` and `sqlite3`). There might be a way to do this that's better than copying and pasting the entire build stage, but I couldn't figure out a way to use environment matrices within a matrix `include` block.

[1]: [sqlite3 has a lot of optimizations to ensure that lock contention isn't that big of an issue](https://www.sqlite.org/lockingv3.html) . Besides, I doubt this will be a performance issue in the `CodaLabManager` case, since I don't think we necessarily expect a ton of concurrent read/writes on the state (thought I think the db would actually be fine if this were the case).


.travis.yml
```diff
@@ -18,7 +18,36 @@ jobs:
...
+        - ./scripts/test-setup.sh
+        - export PATH=$PATH:/usr/bin/geckodriver
+      script:
+        - python3 test_runner.py --version ${TRAVIS_BRANCH} default
+        # Make sure restarting worker doesn't cause any issues (ie in serialization/deserialization)
+        - docker restart codalab_worker_1 && python3 test_runner.py --version ${TRAVIS_BRANCH} run
+        - docker stop codalab_worker_1
+        - python3 codalab_service.py start --services init worker --version ${TRAVIS_BRANCH} --shared-file-system
+        # Run frontend tests + backend test modules that test shared file systems code paths
+        - python3 test_runner.py --version ${TRAVIS_BRANCH} frontend basic run read write kill resources netcat netcurl workers
```
epicfaace: Do we really need to run _all_ these tests again? What do you think a minimal set of tests would be (just to minimize additional CI build time)?

nelson-liu: yeah, I was being a bit conservative here just to ensure that things worked properly. In practice, when I was developing, I didn't see travis build times actually being that impacted. Everything runs in parallel regardless, and the time for each of the backends is basically identical. I know you've been working on getting things moved to Actions, do you think it would be more of a problem there?

> What do you think a minimal set of tests would be (just to minimize additional CI build time)?

I'm honestly not sure, I'm not too familiar with the test suite—your input would be valuable here. At various times when I was working on this PR, I had tests failing in each stage (frontend tests, shared fs tests, the regular tests), so they're all somewhat useful imo...

codalab/lib/codalab_manager_state.py
```diff
@@ -0,0 +1,201 @@
...
+            self.connection = sqlite3.connect(":memory:")
+        else:
+            # Read state database, creating if it doesn't exist.
+            self.connection = sqlite3.connect(self.state_path)
+        self.connection.row_factory = sqlite3.Row
+        # Create the necessary tables, if they don't exist
+        with self.connection:
+            c = self.connection.cursor()
+            c.execute(
+                'CREATE TABLE IF NOT EXISTS auth (server TEXT UNIQUE PRIMARY KEY, access_token TEXT, expires_at REAL, '
```
epicfaace, line 176: I'm worried sqlite is adding a lot of complexity (although I like the abstraction of `CodaLabManagerState`). How easy would it be to add a new column to state? (Would it be better to use an ORM or create migrations instead -- but that's even more complex, haha)

Are you sure we wouldn't be fine if we just added mutexes to make the process of reading from + writing to the JSON state file thread-safe? In that case, we could even create a new class, `CodaLabManagerJsonLockState`, that makes sure that the process of reading state & writing state is thread-safe.

Something like this:

```
acquire lock
read JSON state from disk
modify state and write to disk
release lock
```

nelson-liu: > How easy would it be to add a new column to state? (Would it be better to use an ORM or create migrations instead -- but that's even more complex, haha). 

Yeah, I thought about using an ORM / creating migrations, but I wanted to stay with something as simple as possible (this actually also motivated my use of sqlite, I'll mention more below). I don't think adding an extra column to the state is too difficult. The JSON state manager gets around this by just clobbering the existing state with whatever format it wants to write. We could do the same in the sqlite case, if we see that the existing table schema doesn't match what we expect. It's not ideal, but I think it gets around the issue fine without adding too much complexity.

Also, FWIW, blame seems to say that the state schema hasn't changed in several years. I don't want to say that it won't change in the future, but I historically it seems pretty infrequent—might be worth taking into consideration

> Are you sure we wouldn't be fine if we just added mutexes to make the process of reading from + writing to the JSON state file thread-safe?

No, I'm not sure :) I think that's a fine solution to the issue too. When I was first thinking of how to fix this issue, I was going to do something like this, but then I realized that sqlite has (1) basically already solved this problem, and (2) put a lot more effort into solving it robustly, so I opted for that instead. I thought that it'd be simpler to just import and use sqlite instead of building a lock system.

.travis.yml
percyliang: Is there a way to avoid all the config duplication with the two versions?

codalab/lib/codalab_manager_state.py
```diff
@@ -0,0 +1,201 @@
...
+        self._save_json_state()
+
+    def _save_json_state(self):
+        if self.temporary:
+            return
+        write_pretty_json(self.state, self.state_path)
+
+
+class CodaLabManagerSqlite3State(CodaLabManagerState):
+    def initialize_state(self):
```
percyliang, line 165: Could you provide an high-level overview of the database schema / what's in the state?

percyliang (commented):
Thanks for proactively doing this.

1) If we go with the sqlite route, I'd kind of prefer if we didn't have two backends.  Unlike other things like file system backends where there are tradeoffs, here sqlite seems like it just gives you better functionality.  You might argue that JSON is more interpretable, but we can use `cl` to print out the state if we really wanted to.

2) Taking a step back, when `state.json` was created, the `sessions` part of it was kind of done without much thought - the fact that they never get cleaned up is a known wart.  And that's the part of it that is causing problems...  We could just leave `state.json` alone and just move `sessions` into files (`~/.codalab/sessions/10202.json`).  If we go with one backend, this would be a smaller change than switching everything over to sqlite.  It resolves the concurrency issue for your use case I think (since you're mostly worried about clashes between `cl`'s on different sessions)?

3) I also wonder the following would work to get around your concurrency issues without any changes.  If you have a lot of concurrent `cl`s but on the same session, and instead of trying to change the current worksheet, you just pass `-w ...` explicitly to all the commands, then do all these problems go away because I don't think you'll be mutating `state.json` at all?

nelson-liu:
> If we go with the sqlite route, I'd kind of prefer if we didn't have two backends. Unlike other things like file system backends where there are tradeoffs, here sqlite seems like it just gives you better functionality. You might argue that JSON is more interpretable, but we can use cl to print out the state if we really wanted to.

One argument for keeping JSON around: `sqlite3` isn't necessarily available in all systems, and you can't just `pip install` it either. I'm pretty sure it's a python3 dependency at this point, and it comes installed with most sane distributions of python (e.g., conda provides sqlite3 by default in the environment). I just wasn't sure how you all would feel about making `sqlite3` an (implicit or explicit) dependency of codalab.

> It resolves the concurrency issue for your use case I think (since you're mostly worried about clashes between cl's on different sessions)?

Thread-safety within a process / session would be nice, as well.

> I also wonder the following would work to get around your concurrency issues without any changes. If you have a lot of concurrent cls but on the same session, and instead of trying to change the current worksheet, you just pass -w ... explicitly to all the commands, then do all these problems go away because I don't think you'll be mutating state.json at all?

I haven't tried this, but it seems like it'd work (as long as there's no need to read or write to `state.json`). I wasn't aware of `-w` before.

.travis.yml
nelson-liu: I tried looking for something, but I didn't see anything that would work without dramatically reworking the way the tests are set up (e.g., making a meta single test script that runs an appropriate check depending on an environment variable).

percyliang:
Just to follow up on this:
1) Does `-w` work for you?
2) Could we just use file locks on `state.json` to avoid messing up the state?
3) In the future, it seems like the right thing to do is to build a Python-based SDK (#2387) which I think would be faster (we discussed this briefly) and avoid this whole issue.

Just trying to see if there's a simple workaround now + the right long-term solution.

nelson-liu:
Thanks for checking back in!

> 1. Does -w work for you?

Sorry, I haven't tried this yet. I have a bunch of (semi-automatically-generated) scripts that actually do my `cl run` s for me, so it is somewhat nontrivial to go and change them. I tried it on a small scale, though, and it seems to work.

> 2. Could we just use file locks on state.json to avoid messing up the state?

I don't see any reason why this wouldn't work!

> 3. In the future, it seems like the right thing to do is to build a Python-based SDK (#2387) which I think would be faster (we discussed this briefly) and avoid this whole issue.

Are you thinking that such a python SDK store the state locally (in-memory)? Sounds like a good long-term solution to me, though.

> Just trying to see if there's a simple workaround now + the right long-term solution.

I must say that, for me, the easiest simple workaround has to just been to install this branch locally :) Totally understandable if we decide that we want to go for another solution, though, this is quite heavyweight (and I was mostly optimizing for what'd be easiest for me to implement, rather than what's best for the library).

epicfaace:
This PR has been here for a while... I'd probably be in favor of just not implementing this -- mainly, I don't know if the overhead of having to update a SQLite implementation whenever we change how CodaLab state works, and maintaining multiple implementations of state, is worth it at the moment. As discussed earlier, file locks may be a simpler way to solve this issue. If it's good with you, let's close this PR?

nelson-liu:
Oh oops, I missed this. I'll close this PR, I'm not committed to having this merged. I will probably continue to use it locally, though (and thus will keep periodically force-pushing to keep this branch up-to-date with master)

PR #2368, opened 2020-06-08
Replace sacct with scontrol, since sacct is only available on the slurm head node

nelson-liu:
The CodaLabSlurmWorkerManager works fine if you run it on squeue, but breaks if you try to run it on another other node (e.g., jagupard4). This is because it can't run `sacct`:

```
(base) nfliu at jagupard4 in ~
$ sacct -h
sacct: error: slurm_persist_conn_open_without_init: failed to open persistent connection to localhost:6819: Connection refused
sacct: error: slurmdbd: Sending PersistInit msg: Connection refused
sacct: error: Problem talking to the database: Connection refused
```

~~This replaces the `sacct` call with an equivalent `squeue` call, so you can run the SlurmWorkerManager on nodes besides `sc` (since `sc` can be flaky connection-wise).~~

The `squeue` call actually wasn't equivalent, since it doesn't show completed jobs. I replaced this with a call to `scontrol show jobid -d <jobid>`, and then use a regex to parse out the JobState.


candicegjing:
The intention that I use `sacct` here is that when I was testing `sbatch` command, I noticed that sometimes a job can throw an error before it got put into the Slurm manager queue which can be queried by `squeue`. Another thing to remember is that we can't use the Slurm worker manager on the actual worker nodes as there is no way for Slurm to re-request additional resources on top of the worker nodes.

nelson-liu:
> The intention that I use sacct here is that when I was testing sbatch command, I noticed that sometimes a job can throw an error before it got put into the Slurm manager queue which can be queried by squeue.

Can you elaborate a bit more here? I didn't see such an error, but I might also be misunderstanding...

> Another thing to remember is that we can't use the Slurm worker manager on the actual worker nodes as there is no way for Slurm to re-request additional resources on top of the worker nodes.

Hm, why not? It seems to be working just fine for me. I just want to be able to run it on a host that's not `sc`—I don't necessarily need to add more resources to an existing job or anything.

candicegjing:
> > The intention that I use sacct here is that when I was testing sbatch command, I noticed that sometimes a job can throw an error before it got put into the Slurm manager queue which can be queried by squeue.
> 
> Can you elaborate a bit more here? I didn't see such an error, but I might also be misunderstanding...
> 
I remembered seeing errors only in `sacct` after I submitted a job, but not in `squeue`. Let me see if I can reproduce this now, as it's a while ago.

> > Another thing to remember is that we can't use the Slurm worker manager on the actual worker nodes as there is no way for Slurm to re-request additional resources on top of the worker nodes.
> 
> Hm, why not? It seems to be working just fine for me. I just want to be able to run it on a host that's not `sc`—I don't necessarily need to add more resources to an existing job or anything.

To me, Slurm seems to use the master-slave mode. I tried several slave nodes e.g. jagupard[20, 24], but I am not able to re-request any additional resources. This seems to match with Slurm's architecture in general. But maybe I am wrong? What are the worker nodes you were able to re-request additional resources?

nelson-liu:
> To me, Slurm seems to use the master-slave mode. I tried several slave nodes e.g. jagupard[20, 24], but I am not able to re-request any additional resources. This seems to match with Slurm's architecture in general. But maybe I am wrong? What are the worker nodes you were able to re-request additional resources?

I don't want to re-request additional resources for the job. I want to be able to do `nlprun -g 0 -c 2 --memory 4g -p john`, and run the SlurmWorkerManager from the john machine instead of sc.

candicegjing:
> > > The intention that I use sacct here is that when I was testing sbatch command, I noticed that sometimes a job can throw an error before it got put into the Slurm manager queue which can be queried by squeue.
> > 
> > 
> > Can you elaborate a bit more here? I didn't see such an error, but I might also be misunderstanding...
> 
> I remembered seeing errors only in `sacct` after I submitted a job, but not in `squeue`. Let me see if I can reproduce this now, as it's a while ago.
> 
Finally I recalled the reason to use `sacct` now. `squeue` will return the job information while a job is still active in the Slurm scheduling queue. This means that once a job finished in final states, the job cannot be queried using `squeue`. Since SlurmWorkerManager has its own clock cycle to query the status of each job, by the time the SlurmWorkerManager requests for job status updates, the job may already finish in states like (COMPLETED, FAILED, etc) and doesn't present in the Slurm manager queue. When I was playing with Slurm, I found `sacct` can track a job's information in all steps. That's why I prefer to use `sacct` than `squeue`.

nelson-liu:
> When I was playing with Slurm, I found sacct can track a job's information in all steps. That's why I prefer to use sacct than squeue.

Yeah, I agree—I changed the PR to use `scontrol` instead of `squeue`, since, like you said, `squeue` doesn't show completed jobs.

candicegjing:
> > When I was playing with Slurm, I found sacct can track a job's information in all steps. That's why I prefer to use sacct than squeue.
> 
> Yeah, I agree—I changed the PR to use `scontrol` instead of `squeue`, since, like you said, `squeue` doesn't show completed jobs.

I see. What are the additional benefits that `scontrol` can bring to us comparing to `sacct` in filtering job status?

nelson-liu:
> I see. What are the additional benefits that scontrol can bring to us comparing to sacct in filtering job status?

scontrol works on hosts that aren't `sc.stanford.edu`

candicegjing:
> > To me, Slurm seems to use the master-slave mode. I tried several slave nodes e.g. jagupard[20, 24], but I am not able to re-request any additional resources. This seems to match with Slurm's architecture in general. But maybe I am wrong? What are the worker nodes you were able to re-request additional resources?
> 
> I don't want to re-request additional resources for the job. I want to be able to do `nlprun -g 0 -c 2 --memory 4g -p john`, and run the SlurmWorkerManager from the john machine instead of sc.

Maybe I didn't get to the point here... I guess my understanding is that the SlurmWorkerManager will automatically spawn resources and start codalab workers for users to use. Can you provide more details about your usage pattern of running SlurmWorkerManager after you request a resource by `nlprun -g 0 -c 2 --memory 4g -p john`?

nelson-liu:
> Can you provide more details about your usage pattern of running SlurmWorkerManager after you request a resource by nlprun -g 0 -c 2 --memory 4g -p john?

It's the same as `sc`. I just open a tmux session, and do `cl-worker-manager ...`. The sole benefit here is that I don't have to log in to `sc` to look at my workermanager, since sometimes (often) `sc` is hard to connect to / slow.

candicegjing (approved):

PR #2369, opened 2020-06-08
Make worker id match output directory and job name

nelson-liu:
Right now, when running the SlurmWorkerManager, it generates jobs with (for example):
- worker id: 53f9bcfd24ec4c648e1fafc6364b8d4b
- slurm job name: nfliu-codalab-slurm-worker-53f9bcfd
- output path: `/some/prefix/53f9bcfd24ec4c648e1fafc6364b8d4b` with files `nfliu-codalab-slurm-worker-53f9bcfd.out` and `nfliu-codalab-slurm-worker-53f9bcfd.slurm`.

This is a bit annoying because, given a run on codalab, you can't just necessarily look at the remote and know which slurm job it is. I've changed it such that it's easier to map between worker id <-> job name <-> output path, by making them all the same. So now (for the example above):

- worker id: nfliu-codalab-slurm-worker-53f9bcfd
- slurm job name: nfliu-codalab-slurm-worker-53f9bcfd
- output path: `some/prefix/nfliu-codalab-slurm-worker-53f9bcfd` with files `nfliu-codalab-slurm-worker-53f9bcfd.out` and `nfliu-codalab-slurm-worker-53f9bcfd.slurm`.

### DIFF
--- a/codalab/worker_manager/slurm_batch_worker_manager.py
+++ b/codalab/worker_manager/slurm_batch_worker_manager.py
@@ -171,7 +171,7 @@ class SlurmBatchWorkerManager(WorkerManager):
         """
         Start a CodaLab Slurm worker by submitting a batch job to Slurm
         """
-        worker_id = uuid.uuid4().hex
+        worker_id = self.username + "-" + self.args.job_name + '-' + uuid.uuid4().hex[:8]
 
         # Set up the Slurm worker directory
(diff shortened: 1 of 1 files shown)


candicegjing (approved):
Yeah, I had the same feeling to these as well. Setting them to be the same string sounds great to me!

PR #2372, opened 2020-06-08
Fix WorkerManager import from entrypoint

nelson-liu:
If you run `pip install .` in the source, and try to run `cl-worker-manager`, you get:

```
$ cl-worker-manager  -h
  File "/Users/nfliu/miniconda3/envs/cl_nfliu/bin/cl-worker-manager", line 5, in <module>
    from codalab.worker_manager.main import main
ModuleNotFoundError: No module named 'codalab.worker_manager'
```

This adds an __init_.py to make worker_manager a module.

(It works if you install in editable mode, weirdly enough)

### DIFF
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -438,13 +438,14 @@ class Worker:
         """
         Available disk space by bytes of this RunManager.
         """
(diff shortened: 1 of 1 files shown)


candicegjing (approved):
LGTM!

PR #2373, opened 2020-06-09
Add exit-after-num-runs=1 to slurm worker manager

candicegjing:
To complete #2289


nelson-liu:
Instead of setting this to 1 here, can you make this another argument to the general `WorkerManager` (since it's also useful in the `AWSBatchWorkerManager`), and then make the default CLI flag value 1?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -253,6 +253,8 @@ def setup_codalab_worker(self, worker_id):
             worker_network_prefix,
             # always set in the Slurm worker manager to ensure safe shutdown
             '--pass-down-termination',
+            '--exit-after-num-runs',
```
nelson-liu: instead of adding this here, can we do as below:

```
if self.args.exit_after_num_runs:
    command.extend(['--exit-after-num-runs', str(self.args.exit_after_num_runs)])
```

(same for the awsbatchworkermanager)

candicegjing: Based on the discussion [641011341](https://github.com/codalab/codalab-worksheets/pull/2373#issuecomment-641011341), we agree to set the CLI argument `exit-after-num-runs` with default value `1`. Then we don't need to check for its existence in both child class slurmbatchworkermanager and awsbatchworkermanager right? I have updated the help message if that's the one misleading.

candicegjing: Another thing is that if we consider making this argument optional without any default value, we probably need to know what are the use cases that users would prefer to use in most cases, e.g. run CodaLab workers with `exit-after-num-runs` set or not set. If users prefer to run CodaLab workers with `exit-after-num-runs` set, what value do they normally use? I think having a better understanding of these would be helpful to make those CLI arguments ergonomic (this probably should apply to other CLI arguments as well, as we do have many arguments now.)

nelson-liu: > Based on the discussion 641011341, we agree to set the CLI argument exit-after-num-runs with default value 1. Then we don't need to check for its existence in both child class slurmbatchworkermanager and awsbatchworkermanager right? I have updated the help message if that's the one misleading.

What if the user sets it to 0 / doesn't want it to apply (e.g., maybe they don't want it in the AWSBatchWorkerManager)? Checking if it exists is one way to facilitate that use.

candicegjing: I agree that this could potentially happen, but at a low chance. Basically, I am trying to follow the guideline which happened recently here: [discussion_r429495692](https://github.com/codalab/codalab-worksheets/pull/2288#discussion_r429495692). Basically, setting this variable to `0` could happen to all CLI arguments with the similar properties right? If we want to avoid this case, we probably would want to check all other variables as well for consistency.

nelson-liu:
The way that I see it, the list of arguments:

```
--server
--verbose
--exit-when-idle
--idle-seconds
--work-dir
--id
--network-prefix
```

all _must_ be specified. Thus, we don't have to check for their existence beforehand. 

However, the other remaining flag in that list, `--exit-after-num-runs`, isn't _required_ . A user can choose to disable this if they want, but the only way they'd do so is to set it to 0. However, they can't set it to 0 right now—it doesn't give the intended effect. Another fine alternative is to make the default None, and then set per-workermanager defaults, but that's a bit more involved I think.

So, I think `--exit-after-num-runs` is more like `--tag`, `--max-work-dir-size`, and `--worker-delete-work-dir-on-exit` in that they're not required to have a value, so it should be accordingly grouped with those flags as well.

candicegjing:
> However, the other remaining flag in that list, `--exit-after-num-runs`, isn't _required_ . A user can choose to disable this if they want, but the only way they'd do so is to set it to 0. However, they can't set it to 0 right now—it doesn't give the intended effect. Another fine alternative is to make the default None, and then set per-workermanager defaults, but that's a bit more involved I think.

I agree with you on most of the parts. However, I think we probably are not on the same page on one thing here. I just want to make sure our understanding of **"A user can choose to disable this if they want, but the only way they'd do so is to set it to 0"** are the same. When you say **"A user can choose to disable this**, do you mean to let the worker run as much as it can without any additional restriction?

nelson-liu:
> When you say "A user can choose to disable this, do you mean to let the worker run as much as it can without any additional restriction?

Yeah, in the current implementation, there's no way of specifying "just let me run this worker forever", right? I think there should be room for users to do that, if they so choose. A reasonable default is still "1", since that's what most people probably should do, but there also needs to be a way to turn it off completely.

candicegjing:
> > When you say "A user can choose to disable this, do you mean to let the worker run as much as it can without any additional restriction?
> 
> Yeah, in the current implementation, there's no way of specifying "just let me run this worker forever", right? I think there should be room for users to do that, if they so choose. A reasonable default is still "1", since that's what most people probably should do, but there also needs to be a way to turn it off completely.

Oh, I see why we can't reach a consensus now. If we want to **"just let me run this worker forever"**, setting `--exit-after-num-runs` to `0` won't help us achieve this goal. Setting `--exit-after-num-runs` to `0` means the worker will exit after no runs, which contradicts our original goal. On the worker side, this is set to use the maximum 32-bit integer value as default. If we really want to be more flexible for users to set this option, I would suggest not set it with a default value but use a wrapper like `nlprun` to set this to a value that users are most likely going to use.

nelson-liu:
>  If we want to "just let me run this worker forever", setting --exit-after-num-runs to 0 won't help us achieve this goal. Setting --exit-after-num-runs to 0 means the worker will exit after no runs, which contradicts our original goal.

Right, but what i'm saying is that it might be nice to be able to set `--worker-exit-after-num-runs` to a special sentinel value of 0 (or None, but that's hard to input into a CLI), which indicates "don't pass this value down". a default of None also works.

candicegjing:
> Right, but what i'm saying is that it might be nice to be able to set `--worker-exit-after-num-runs` to a special sentinel value of 0 (or None, but that's hard to input into a CLI), which indicates "don't pass this value down". a default of None also works.

Well, I think it's a bit hard to follow here. All the past discussions are focused on setting **`exit-after-num-runs` to 0 is the only way to let the worker run forever**, which is confusing in a way that users can pass a large enough value to `exit-after-num-runs` to make it work. I think it would be really helpful if we can be as precise as possible when discussing issues here. 

Another thing is that when we pass down values from the top-level worker manager to the actual CodaLab worker, it would be nice to keep the meaning of the value assigned to this argument in the worker manager the same as the CodaLab worker for easier reasoning. This is also what we have done for the rest of other worker CLI arguments now. Setting the value to `0` on the worker manager to let the worker run forever seems to be contradictory to the meaning of `0` in the actual worker. 

It sounds like we want to be able to let the worker manager has the flexibility to have worker accept any number of possible jobs now as opposed to the original design 1 worker runs 1 job[#issuecomment-632268004](https://github.com/codalab/codalab-worksheets/issues/1019#issuecomment-632268004), I would prefer to set its default value to `None` when users don't want to pass down this parameter at all (users don't really need to specify this value at all). Then users can have the maximum flexibility to use this argument, plus the code logic is consistent.

nelson-liu:
> Another thing is that when we pass down values from the top-level worker manager to the actual CodaLab worker, it would be nice to keep the meaning of the value assigned to this argument in the worker manager the same as the CodaLab worker for easier reasoning. This is also what we have done for the rest of other worker CLI arguments now. Setting the value to 0 on the worker manager to let the worker run forever seems to be contradictory to the meaning of 0 in the actual worker.


Yeah, this is a good point, let's leave it as-is then.

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -258,6 +258,8 @@ def setup_codalab_worker(self, worker_id):
             command.extend(['--tag', self.args.worker_tag])
         if self.args.password_file:
             command.extend(['--password-file', self.args.password_file])
+        if self.args.exit_after_num_runs and self.args.exit_after_num_runs > 0:
```
nelson-liu: I think this needs to be changed to `self.args.worker_exit_after_num_runs`

codalab/worker_manager/main.py
```diff
@@ -53,6 +53,11 @@ def main():
         default=1 * 60,
         type=int,
     )
+    parser.add_argument(
+        '--worker-exit-after-num-runs',
+        type=int,
+        help='The CodaLab worker will exit after this many of runs finished',
```
epicfaace:
```suggestion
        help='If specified, the CodaLab worker will exist after finishing this many runs.',
```

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -98,6 +98,8 @@ def start_worker_job(self):
             command.extend(['--max-work-dir-size', self.args.worker_max_work_dir_size])
         if self.args.worker_delete_work_dir_on_exit:
             command.extend(['--worker-delete-work-dir-on-exit'])
+        if self.args.exit_after_num_runs and self.args.exit_after_num_runs > 0:
```
yipenghe: Are exit_after_num_runs and worker_exit_after_num_runs the same? Ditto for slurm batch manager

nelson-liu (approved):

PR #2374, opened 2020-06-09
fix delete last bundle focus

### LINKED ISSUE #2363: deleting bundle makes cursor disappear
To reproduce, create a new worksheet.  Add a markdown block.  Add a run bundle.  Then delete the bundle (using keyboard shortcuts).  Then the cursor disappears (should be on the markdown block).  When try to cursor around, get:
Screenshot from 2020-06-06 22-39-46

yipenghe:
fix #2363

The current fix is to add a default focus when we encounter a out of bound case of setFocus. The root problem is that the original reload case doesn't consider deleting multiple bundles at a time. It doesn't seem trivial to fix at the time, so I would like to merge the temp fix for now.


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1342,28 +1342,31 @@ class Worksheet extends React.Component {
                         // the deleted bundle is the only item of the table
                         if (this.state.subFocusIndex === 0) {
                             // the deleted item is the last item of the worksheet
-                            if (items.length === focus + 1) {
+                            if (blocks.length === focus) {
```
yipenghe: This line fixes it

epicfaace (commented):
Sorry, I'm having some trouble reproducing this problem. How do you delete a bundle using keyboard shortcuts?

yuqijin:
Close it with PR #2963

PR #2381, opened 2020-06-10
Fix time displaying issue

### LINKED ISSUE #2380: Bad formatted metadata field time showing NaNyNaNd
Screen Shot 2020-06-09 at 5 05 14 PM

Metadata field **time** shows `NaNyNaNd` again due to changes in PR #2357

candicegjing:
Fixed #2380

This PR is to fix the value displaying at the `time` field
Screen Shot 2020-06-12 at 11 19 19 AM

### DIFF
--- a/frontend/src/util/worksheet_utils.js
+++ b/frontend/src/util/worksheet_utils.js
@@ -7,7 +7,7 @@ export function renderDuration(s) {
     // s: number of seconds
     // Return a human-readable string.
     // Example: 100 => "1m40s", 10000 => "2h46m"
-    if (s === null) {
+    if (s == null) {
         return '<none>';
     }
 
--- a/frontend/src/util/worksheet_utils.js
+++ b/frontend/src/util/worksheet_utils.js
@@ -7,6 +7,7 @@ export function renderDuration(s) {
     // s: number of seconds
     // Return a human-readable string.
     // Example: 100 => "1m40s", 10000 => "2h46m"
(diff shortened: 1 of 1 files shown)


yipenghe:
What exactly is s in this case, why do we need a loosely equal to check for the null? Is it actually a undefined rather than null (null == undefined is true, null === undefined is false)? If it's undefined we should check for undefined (=== undefined) instead of using ==

candicegjing:
@yipenghe Ref of the issue that fixed this a while ago: https://github.com/codalab/codalab-worksheets/pull/2037. Both `undefined` and `null` can happen.

yipenghe:
> @yipenghe Ref of the issue that fixed this a while ago: #2037. Both `undefined` and `null` can happen.

I would prefer using two conditions in this case (checking both === undefined and === null) so that we know both conditions can happen

yipenghe:
> @yipenghe Ref of the issue that fixed this a while ago: #2037. Both `undefined` and `null` can happen.

Or we should add a comment here that == is needed to check for both

candicegjing:
> > @yipenghe Ref of the issue that fixed this a while ago: #2037. Both `undefined` and `null` can happen.
> 
> Or we should add a comment here that == is needed to check for both

Yeah, sounds good! This can keep the logic as simple as possible. I am adding a comment here to indicate both situations can happen.

yipenghe (approved):

PR #2384, opened 2020-06-10
Revamp doc site, create a blog for announcements

epicfaace:
Also changed everything to use docusaurus instead of mkdocs. This should make our documentation site, hopefully, cleaner and more appealing. We can deploy this site automatically through GitHub Pages, so we wouldn't need to use mkdocs.

I chose docusaurus because it allows you to add custom pages and blog posts -- and it looks better than mkdocs.

When we add a new blog post, we can also send that announcement out as an email to our mailing list.

I've drafted part of an announcement, but this is still in progress -- if you could let me know of any major changes that should be added, let me know @candicegjing @Levilian @yipenghe @teetone 

Blog page (still in progress):

image

Docs site:

image

Home page:

image


yipenghe:
I think we should mention these two in frontend changes: Search bar and copy-cut-paste

docs-site/blog/2020-06-10-new-site.md
```diff
@@ -0,0 +1,24 @@
...
+tags: [introduction, release, blog]
+---
+
+We're pleased to announce a new documentation site, built with [Docusaurus](https://v2.docusaurus.io/). We'll also be sending out an occasional email digest with the updates that are posted to the blog here. If you're interested in being added to the list, please email <a href="mailto:codalab.worksheets@gmail.com">codalab.worksheets@gmail.com</a>.
+
+Here are some of the major new features that we have released within the past few months:
+
+## Frontend improvements
+
+- We've introduced a new frontend with drag and drop options. We also have a new UI for bundle detail layout ([#1543](https://github.com/codalab/codalab-worksheets/pull/1543))
```
yipenghe: I think the drag and drop should be removed. Do we need to mention the detail layout? It seems to be a long time ago

epicfaace:
We could also, alternatively, keep using mkdocs and use [mkdocs-material](https://squidfunk.github.io/mkdocs-material/) instead

epicfaace:
Let's keep using mkdocs and maybe use something like mkdocs-material in the future. It doesn't seem to be worth it to switch the entire docs platform / URLs at this point. Closing this PR for now.

PR #2386, opened 2020-06-11
Save expected line files to validate sample worksheet outputs

### LINKED ISSUE #2385: Sample worksheet test always pass when the small worksheet already exists
In create_sample_worksheet.py, the test_print relies on self._expected_lines, which only gets constructed when the sample worksheet is created. However, if we already run the test locally once, the creation will be skipped and _expected_lines will be empty.

Also _wait_for_bundles_to_finish needs the attribute self._valid_bundles to exist, which is also added only if we try to construct a new sample worksheet, this will make the test fail locally as well.
```
      self._wait_for_bundles_to_finish()
      output_lines = run_command([self._cl, 'print', self._worksheet_name]).split('\n')
      has_error = False
      for i in range(len(self._expected_lines)):
            if not re.match(self._expected_lines[i], output_lines[i]):
                has_error = True
                # Output mismatch message in red
                print(
                    '\033[91mMISMATCH! line: {} pattern: {} output: {}\033[0m'.format(
                        i + 1, self._expected_lines[i], output_lines[i]
                    )
                )
```

yipenghe:
fix #2385 , we should recreate the sample worksheet everytime we run the test, otherwise even if we changed the print output, the test locally will not recreate the sample worksheet. Also skipping the creation causes failures of the test as described in the PR


scripts/create_sample_worksheet.py
```diff
@@ -67,11 +67,11 @@ def create(self):
         ).split('\n')
         if re.match(SampleWorksheet._FULL_UUID_REGEX, worksheet_uuids[0]):
             print(
-                'There is already an existing {} with UUID {}. Skipping creating a sample worksheet...'.format(
+                'There is already an existing {} with UUID {}. Removing and creating a new sample worksheet...'.format(
                     self._worksheet_name, worksheet_uuids[0]
                 )
             )
-            return
+            run_command([self._cl, 'wrm', '--force', self._worksheet_name])
```
teetone: This logic was added so the same worksheet (including both small and large worksheet) does not accidentally get recreated multiple times, since it can even take some time to even create a small worksheet. I think a better fix here would be to refactor the logic, so we can skip creating a new worksheet if one exists, but still have the ability to validate its output.

yipenghe: one thing we could do is to save the expected lines somewhere

teetone: It seems like the build is failing now. 

I still think it will be cleaner if we refactor so that we don't need these extra files. We can just build the expected lines irregardless of whether or not a  worksheet was created.

yipenghe: That doesn't seem feasible: for valid bundles and valid worksheets, I guess you can hard code the expected lines to be self._entities_count, but what about search ",png", "tsv" "gif" bundles? I'm not sure where to find the exact numbers of expected lines to add for these results without actually running the search command

teetone:
Closing this for now, we can revisit in the future.

PR #2389, opened 2020-06-12
`uls` search for user stats

Levilian:
Feature: #427


epicfaace (commented):
Can you add some tests in the test_cli.py test suite?

codalab/lib/bundle_cli.py
```diff
@@ -674,7 +674,25 @@ def print_table(
         for row_dict in row_dicts:
             row = []
             for col in columns:
-                cell = row_dict.get(col)
+                # These fields will not be returned by the server if the
+                # authenticated user is not root
+                if col in ('last_login', 'email', 'time', 'disk', 'parallel_run_quota'):
```
percyliang: Not sure why this logic is needed.  We should ideally just print out whatever fields the server sends back.

codalab/lib/bundle_cli.py
```diff
@@ -3756,6 +3774,69 @@ def do_uedit_command(self, args):
...
+            Commands.Argument('-f', '--field', help='Print out these comma-separated fields.'),
+            Commands.Argument('-w', '--weekly_stats', help='Print out the summary report of weekly new and active users respectively.'),
+        ),
+    )
+    def do_uls_command(self, args):
+        """
+        Search for specific users.
+        """
+        client = self.manager.current_client()
+            #             user_id = client.fetch('user')['id']
```
percyliang: Remove commented code

codalab/lib/bundle_cli.py
```diff
@@ -3773,12 +3854,17 @@ def do_uinfo_command(self, args):
         Edit properties of users.
         """
         client = self.manager.current_client()
+        print(args.user_spec)
```
percyliang: Remove

codalab/model/bundle_model.py
```diff
@@ -2179,12 +2180,191 @@ def get_users(self, user_ids=None, usernames=None, check_active=True):
...
+            clauses.append(cl_user.c.last_login >= last_login)
 
         with self.engine.begin() as connection:
             rows = connection.execute(select([cl_user]).where(and_(*clauses))).fetchall()
 
         return [User(row) for row in rows]
 
+    def get_user_stats(self, keywords):
+        """
+        Get user stats, specifically used for weekly on-call.
```
percyliang: Remove 'weekly on-call'.  This is more general than that.

codalab/model/bundle_model.py
```diff
@@ -2179,12 +2180,191 @@ def get_users(self, user_ids=None, usernames=None, check_active=True):
...
+        - .active_before=<datetime>: return users whose last login is before the datetime
+        - .active_after=<datetime>: return users whose last login is after the datetime
+        - .joined_before=<datetime>: return users who joined before the datetime
+        - .joined_after=<datetime>: return users who joined after the datetime
+        - .offset=<int>: return bundles starting at this offset
+        - .limit=<int>: maximum number of bundles to return
+        - .count: just return the number of bundles
+        Keys are one of the following:
+        - User fields (e.g., name)
+        - Special fields (e.g., date_joined)
```
percyliang: What's the difference between these two types of fields?

codalab/model/bundle_model.py
```diff
@@ -2179,12 +2180,191 @@ def get_users(self, user_ids=None, usernames=None, check_active=True):
...
+            elif key == 'disk_quota':
+                clause = make_condition(key, cl_user.c.disk_quota, value)
+            elif key == 'disk_used':
+                clause = make_condition(key, cl_user.c.disk_used, value)
+            # Special fields
+            elif key in ('.joined_after', '.active_after', '.joined_before', '.active_before'):
+                if key == '.joined_after':
+                    clause = cl_user.c.date_joined >= value
+
+                if key == '.active_after':
```
percyliang: Use `elif` in this section

codalab/model/bundle_model.py
```diff
@@ -2179,12 +2180,191 @@ def get_users(self, user_ids=None, usernames=None, check_active=True):
...
+                    clause = cl_user.c.last_login >= value
+
+                if key == '.joined_before':
+                    clause = cl_user.c.date_joined <= value
+
+                if key == '.active_before':
+                    clause = cl_user.c.last_login <= value
+            elif key == '':  # Match any field
+                clause = []
+                clause.append(cl_user.c.id.like('%' + value + '%'))
```
percyliang: Doesn't really make sense to use the id, since that's internal

codalab/rest/users.py
```diff
@@ -143,6 +143,44 @@ def fetch_users():
     users = local.model.get_users(usernames=(usernames or None))
     return allowed_user_schema()(many=True).dump(users).data
 
+@get('/userstats')
```
percyliang: Oh, we already have a `/users` endpoint...where is `/users` used?  We should unify the two and call it `/users` since it seems like your new version is strictly more powerful?

percyliang (dismissed):
Cool.  +1 for tests.  Also, it'd be nice if we could refactor to avoid code duplication across search for bundles, worksheets, and users....

codalab/rest/users.py
Levilian: We rely on `/users` for fetching specific users by username and email directly, whereas `/userstats` fetches users by keywords in the form of username=codalab. Do we really want to combine the two?

codalab/model/bundle_model.py
```diff
@@ -2180,31 +2180,205 @@ def get_user(self, user_id=None, username=None, check_active=True):
...
+                clause = None
+                # Special functions
+                if key == '.offset':
+                    offset = int(value)
+                elif key == '.limit':
+                    limit = int(value)
+                elif key == '.format':
+                    format_func = value
+                # Bundle fields
+                elif key == 'id':
```
epicfaace: I think we can refactor many of these if / elif statements by just using something like `make_condition(key, getattr(cl_user.c, key), value)`

test_cli.py
```diff
@@ -923,6 +925,59 @@ def test_worksheet_tags(ctx):
...
+
+    # check .disk_used_more_than
+    check_contains(
+        ['test'], _run_command([cl, 'uls', '.disk_used_more_than=' + '10%', '-f', 'user_name'])
+    )
+
+    # check .time_used_more_than
+    check_contains(
+        ['test'], _run_command([cl, 'uls', '.time_used_more_than=' + '10%', '-f', 'user_name'])
+    )
```
epicfaace: Can you make sure that these tests test _all_ possible filters (such as disk_quota, etc.)?

codalab/lib/bundle_cli.py
```diff
@@ -662,12 +662,34 @@ def print_table(
         """
         Pretty-print a list of columns from each row in the given list of dicts.
         """
-        # Get the contents of the table
         rows = [columns]
+        # display restricted fields if the server returns those fields - which suggests the user is root
```
epicfaace, line 666: We should add a test for running this command with a root user and a non-root user to ensure that restricted fields are displayed only to root.

Also, can you document this feature (different formats for root and non-root users) in the CLI command documentation?

codalab/lib/bundle_cli.py
```diff
@@ -662,12 +662,34 @@ def print_table(
...
+        except KeyError:
+            pass
+
+        # Get the contents of the table
         for row_dict in row_dicts:
             row = []
             for col in columns:
-                cell = row_dict.get(col)
+                try:
+                    if col == 'time':
```
epicfaace, line 681: I don't know if we want to hardcode these columns in, because `print_table` is used in many other commands other than just `cl uls` and it may break backwards compatibility (and I'm not sure if all other tables will have time columns necessarily named "time")... what do you think @percyliang ?

Levilian: Would the try statement prevent backward incompatibility?

codalab/lib/bundle_cli.py
```diff
@@ -656,14 +656,67 @@ def resolve_worksheet_uuid(client, base_worksheet_uuid, worksheet_spec):
             )['uuid']
         return worksheet_uuid
 
+    def uls_print_table(
+        self, columns, row_dicts, user_defined=False,
+    ):
```
epicfaace, line 661: Can you add a docstring for this method?

codalab/lib/bundle_cli.py
```diff
@@ -3820,7 +3936,6 @@ def print_attribute(key, user, should_pretty_print):
             'affiliation',
             'url',
             'date_joined',
-            'last_login',
```
epicfaace: Did you mean to remove this (last_login)?

codalab/model/bundle_model.py
```diff
@@ -2184,31 +2164,195 @@ def get_user(self, user_id=None, username=None, check_active=True):
...
-        Get users.
+        Get user stats.
 
-        :param user_ids: user ids of users to fetch
+        Returns a bundle search result dict where:
+            result: list of bundle uuids matching search criteria in order
+                        specified for bundle searches
+                    single number value for aggregate searches(.count)
+            is_aggregate: True for aggregate searches, False otherwise
+        Each keyword is either:
```
epicfaace: We probably shouldn't duplicate the documentation, since this is already documented in the `uls` CLI command and it may be hard to keep both docstrings in sync. Can you maybe replace these details about keywords to something like "see the documentation for cl uls for information about keyword structure"?

codalab/model/bundle_model.py
```diff
@@ -2184,31 +2164,195 @@ def get_user(self, user_id=None, username=None, check_active=True):
...
         """
         clauses = []
+        offset = 0
+        format_func = None
+        limit = SEARCH_RESULTS_LIMIT
+        count = False
+        sort_key = [None]
+        aux_fields = []  # Fields (e.g., sorting) that we need to include in the query
+
+        # Number nested subqueries
```
epicfaace, line 2194: What does this mean? Could you elaborate in the comment?

codalab/rest/users.py
```diff
@@ -140,13 +140,44 @@ def fetch_users():
         filter[user_name]=name1,name2,...
         filter[email]=email1,email2,...
 
-    Fetches all users that match any of these usernames or emails.
+    Query parameters:
+
+    - `keywords`: Search keyword. May be provided multiples times for multiple
```
```suggestion
    - `keywords`: Search keyword. May be provided multiple times for multiple
```

codalab/rest/users.py
```diff
@@ -140,13 +140,44 @@ def fetch_users():
...
+    keywords = query_get_list('keywords')
+    if usernames == None and keywords == None:
+        abort(
+            http.client.BAD_REQUEST, "Request must include 'keywords' query parameter or usernames"
+        )
+
+    if request.user.user_id != local.model.root_user_id:
+        for key in keywords:
+            if any(
+                unaccessed_field in key for unaccessed_field in ['time', 'disk', 'active', 'email']
```
epicfaace: Why do we restrict only these fields? Did we already discuss which fields shouldn't be accessible? I'm wondering if other fields such as `parallel_run_quota` should also be inaccessible (alternatively, we could create a whitelist of allowed fields instead of a blacklist of disallowed fields)

cc @percyliang

scripts/create_sample_worksheet.py
```diff
@@ -158,7 +158,7 @@ def _create_dependencies(self):
             name = 'valid_private_worksheet_%s' % id
             title = 'Other Private Worksheet %s' % id
             uuid = self._create_tagged_worksheet(name, title)
-            run_command([self._cl, 'wperm', uuid, 'public', 'none'])
+            run_command([self._cl, 'wperm', uuid, 'public', 'n'])
```
epicfaace: Can we revert this?

epicfaace (commented):
Looks good! Left a couple of comments, mainly about documentation.

codalab/rest/schemas.py
```diff
@@ -271,6 +271,24 @@ class Meta:
...
 
+USER_ACCESSIBLE_FIELDS = (
+    'name',
+    'user_name',
+    'first_name',
+    'last_name',
+    'affiliation',
+    'url',
+    'disk_used',
+    'joined',
```
epicfaace: I don't think 'joined', 'size', and 'format' exist on the User model. Can you double check these values?

epicfaace (changes requested):

codalab/rest/schemas.py
Levilian: Yes - these are keywords that user can enter to complete the query. I will move the code to a different place to avoid this confusion.

epicfaace (approved):

PR #2391, opened 2020-06-12
Fix SlurmWorkerManager overlaunching

nelson-liu:
Merge #2368 before merging this

When a job gets pre-empted, Slurm puts it in to the `COMPLETING` stage before requeueing it.

Right now, since we remove jobs from the SlurmWorkerManager when it's in the `COMPLETING` state, we run the risk of over-launching jobs (since we think the jobs are done and should be removed, but they're actually be requeued).

Also adds some fixes to logging to make it more readable / usable.


candicegjing (approved):
LGTM! This is good to know!

PR #2412, opened 2020-06-15
Handle sort key is 0 condition on frontend and use correct sort key for new runs

### LINKED ISSUE #2396: when create bundle, cursor should be placed on it
When you create a run or upload a bundle, the cursor should be placed on it.  Currently, that doesn't happen (I think it is either on the one before or after).

### LINKED ISSUE #2401: inserting bundles in a table doesn't work
To reproduce, on a new worksheet, create two bundles, cursor to the first one and then type 'a r' to create a new run.  This run will actually be appended rather than inserted.

Before:
Screenshot from 2020-06-14 14-25-31
After trying to insert `echo bye`:
Screenshot from 2020-06-14 14-25-42

I've noticed that sometimes insertion works, but it's kind of unpredictable.

EDIT: I realized this bug comes up when there is an extra empty markdown line after the two bundles.  If there isn't, then there's no bug!

EDIT: sometimes I hit this bug when there's no new empty markdown line after the two bundles either.

yipenghe:
fix #2401 should also fix #2396  

To reproduce the problem, open an empty worksheet, add a new run, focus and add another new run. Focus on the first bundle, add new text or new run, it appends to the end

### DIFF
--- a/frontend/src/components/worksheets/NewRun/NewRun.jsx
+++ b/frontend/src/components/worksheets/NewRun/NewRun.jsx
@@ -234,7 +234,7 @@ class NewRun extends React.Component<{
 
         let args = ['run'];
 
-        if (after_sort_key) args.push(`-a ${ after_sort_key }`);
+        if (after_sort_key !== null || after_sort_key !== undefined) args.push(`-a ${ after_sort_key }`);
         if (name) args.push(`--name=${name}`);
         if (description) args.push(`--description=${description}`);
         if (tags) args.push(`--tags=${tags.map((tag) => `'${tag}'`).join(",")}`);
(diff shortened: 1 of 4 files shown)


epicfaace (approved):
The after_sort_key 0 check seems good. It might be cleaner just to do `if (after_sort_key || (after_sort_key === 0))` (just in case there are other falsey values that we want to account for)

What exactly is the difference between `this.props.after_sort_key` and `bundleInfo.sort_key`?

yipenghe:
> The after_sort_key 0 check seems good. It might be cleaner just to do `if (after_sort_key || (after_sort_key === 0))` (just in case there are other falsey values that we want to account for)
> 
> What exactly is the difference between `this.props.after_sort_key` and `bundleInfo.sort_key`?

Actually they should be the same without bug, the problem came from getAfterSoryKey function, which also didn't handle the zero case. I updated the getAfterSortKey and use this.props.after_sort_key instead for consistency since it seems other components use this

epicfaace (approved):

PR #2415, opened 2020-06-16
Run nvidia-smi directly on worker host to respect GPU isolation

nelson-liu:
Currently, workers parse their GPUSET with docker, which doesn't respect GPU isolation. I'm not sure what the original reason for doing this was / if there's anything I'm missing—any ideas?

Most of this new functionality is taken directly from https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/docker_utils.py#L106-L113 . I didn't think the overlap was large enough to warrant an extra function, especially since it's subprocess vs. `docker run`.


codalab/worker/main.py
```diff
@@ -298,8 +299,15 @@ def parse_gpuset_args(arg):
         return set()
 
     try:
-        all_gpus = docker_utils.get_nvidia_devices()  # Dict[GPU index: GPU UUID]
-    except docker_utils.DockerException:
+        nvidia_command = ['nvidia-smi', '--query-gpu=index,uuid', '--format=csv,noheader']
```
percyliang, line 316: Should add a comment that says we shouldn't trust docker (e.g., in slurm).  Is this a known docker bug?

nelson-liu: Yeah, I'll add a comment. Unclear if this is a bug or a feature, though—it's the same deal with cpuset, right? Docker just in general runs independent of most (any?) host-level restrictions.

percyliang (approved):

PR #2419, opened 2020-06-17
Fix accessing worker information from WorkerInfoAccessor

nelson-liu:
Before this diff:
```
  File "/opt/codalab-worksheets/codalab/server/bundle_manager.py", line 80, in run
    self._run_iteration()
  File "/opt/codalab-worksheets/codalab/server/bundle_manager.py", line 100, in _run_iteration
    self._schedule_run_bundles()
  File "/opt/codalab-worksheets/codalab/server/bundle_manager.py", line 714, in _schedule_run_bundles
    self._schedule_run_bundles_on_workers(workers, staged_bundles_to_run, user_info_cache)
  File "/opt/codalab-worksheets/codalab/server/bundle_manager.py", line 417, in _schedule_run_bundles_on_workers
    if worker['exit_after_num_runs'] < workers[worker['worker_id']]['exit_after_num_runs']:
TypeError: 'WorkerInfoAccessor' object is not subscriptable
```

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -414,7 +414,10 @@ class BundleManager(object):
         # before bundle manager's next scheduling loop
         for worker in workers_list:
             # Update workers that have "exit_after_num_runs" manually set from CLI.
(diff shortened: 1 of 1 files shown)


candicegjing (approved):
Good catch!

PR #2422, opened 2020-06-17
Don't fail WorkerManager if a network exception is encountered

nelson-liu:
Sometimes there are slight blips in connectivity to the codalab server which means that the following lines might fail:

https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/worker_manager.py#L89-L91

This makes it such that the workermanager can recover and keep operating, versus fatally exiting.

### DIFF
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -1,5 +1,6 @@
 import logging
 import time
+import traceback
 from codalab.lib.codalab_manager import CodaLabManager
 from codalab.worker.bundle_state import State
 from collections import namedtuple
@@ -75,7 +76,10 @@ class WorkerManager(object):
 
     def run_loop(self):
         while True:
-            self.run_one_iteration()
+            try:
+                self.run_one_iteration()
+            except Exception:
+                traceback.print_exc()
             if self.args.once:
                 break
             logger.debug('Sleeping {} seconds'.format(self.args.sleep_time))
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -85,6 +85,10 @@ class WorkerManager(object):
             try:
                 self.run_one_iteration()
             except (urllib.error.URLError, http.client.HTTPException, socket.error, NotFoundError):
+                # Sometimes, network errors occur when running the WorkerManager . These are often
+                # transient exceptions, and retrying the command would lead to success---as a result,
+                # we ignore these network-based exceptions (rather than fatally exiting from the
+                # WorkerManager )
                 traceback.print_exc()
             if self.args.once:
                 break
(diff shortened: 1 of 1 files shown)


codalab/worker_manager/worker_manager.py
```diff
@@ -75,7 +76,10 @@ def start_worker_job(self):
 
     def run_loop(self):
         while True:
-            self.run_one_iteration()
+            try:
+                self.run_one_iteration()
+            except Exception:
```
percyliang: What kind of exceptions happen?

I'm a bit worried that if we indiscriminately catch exceptions, we won't fix the root issues, and since this is in a loop, whatever failure get will just continue to happen.

nelson-liu: If you, say, upgrade the codalab instance, you'll briefly get a 404 or a ConnectionRefused error. Currently, this completely crashes the WorkerManager, while the worker itself is able to handle these blips. Maybe we should just try / except this one snippet (https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/worker_manager.py?rgh-link-date=2020-06-17T07%3A33%3A42Z#L89-L91), and return if it fails (so it can sleep and just retry?)

percyliang: I see.  Maybe we can pull out the factor out the logic used by the worker to connect and have everyone use that.

nelson-liu: Oh, it's not exactly the same logic except in the abstract sense—the worker / bundlemanager all just have:

```
while True / not exiting:
    try:
        do_some_action()
    except Exception:
        raise
    sleep()
```

see, e.g., https://github.com/codalab/codalab-worksheets/blob/ce690f5ab735b0fae50aecf2b00c3b2b527967aa/codalab/server/bundle_manager.py#L78-L84

The workermanager isn't doing any check-ins like the worker does. It just needs to fetch how many bundles the user currently has staged. If codalab is momentarily down, this fetch will fail and the entire program will exit. You're right that I should be able to just ignore the connection errors, though—I'll change this PR to do so.

codalab/worker_manager/worker_manager.py
```diff
@@ -75,7 +82,10 @@ def start_worker_job(self):
 
     def run_loop(self):
         while True:
-            self.run_one_iteration()
+            try:
+                self.run_one_iteration()
+            except (urllib.error.URLError, http.client.HTTPException, socket.error, NotFoundError):
```
percyliang, line 87: Add comment saying that this is to be robust against server disconnecting.

Also, is there a way to localize the try/except more or is this as far in you can push it?

percyliang (approved):
LGTM!

codalab/worker_manager/worker_manager.py
nelson-liu: I think this is as far as it can reasonably go while still maintaining good coverage.

PR #2424, opened 2020-06-18
Don't try to start bundles on offline / timed-out workers

### LINKED ISSUE #2423: BundleManager sometimes stalls for a very long time (~1 hour)
**Describe the bug**

Sometimes, the BundleManager appears unresponsive for a very long time. New runs are not staged or assigned to run on workers and dead workers are not cleaned up.

This is running on the latest master branch.

After investigating, I figured out that the BundleManager isn't stuck--it simply takes an inordinate amount of time to try to schedule each bundle. In `BundleManager._schedule_run_bundles_on_workers`, we iterate through the staged bundles and try to start them on each of the workers that have the resources to take the bundle (https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L386).

However, sometimes a worker goes offline in this process (this happens with the SlurmWorkerManager, when a worker gets pre-empted [1]). As a result, we have to iterate through *every bundle* and try to run it on a worker that's offline. This can take a large amount of time, upwards of an hour. This is bad because no bundles are staged for an hour+, which also means that new workers won't pick up jobs and might exit (if they have `--exit-on-idle` set to true).

[1]: I think there's an especially bad interaction with the `SlurmWorkerManager` and the restage-on-terminate feature—this issue always comes up when a job is pre-empted.

nelson-liu:
Fixes #2423 

We need to both `_cleanup_dead_workers` and re-fetch the workers, and then try to run bundles on _only_ the workers that are online.

Empirically, this doesn't seem to affect runtime ~~that much~~ at all, even with 40+ workers connected.

I also thought about fixing this issue by making workers `checkout` when they exit, but there's no guarantee of a clean termination (e.g., if a worker gets SIGKILLED).


candicegjing:
I like this idea to fix the dead worker problem! We definitely want to clean up dead workers by re-fetching workers when dispatching each bundle. 
The only thing that I was worried about is that re-fetching workers may not be done as we expected based on the current configuration of refreshing logic. The logic of refreshing workers is done every ~55 seconds. However, there is no guarantee that by the time we call the function `cleanup_dead_workers()`, `workers` will be re-fetched from the database. Given my benchmarking results, when we try to dispatch bundles to a dead worker, the time cost roughly stays between (0.3, 0.6). (Dispatching bundles to an online worker will be much more faster)  Suppose the workers get refreshed 30 seconds ago, when we call the function `cleanup_dead_workers()`, we still use the outdated worker information. If any worker gets disconnected from the server during this 30 seconds interval, we won't be able to remove them from the `workers` object. Hence, we cannot avoid dispatching bundles to the dead worker.  
Although re-fetching workers by the time we call the `cleanup_dead_workers()` function still cannot 100% guarantee us to have a sychronized worker state from bundle manager, but I think that's the best we can do for now.

nelson-liu:
> Suppose the workers get refreshed 30 seconds ago, when we call the function cleanup_dead_workers(), we still use the outdated worker information. If any worker gets disconnected from the server during this 30 seconds interval, we won't be able to remove them from the workers object. Hence, we cannot avoid dispatching bundles to the dead worker.

Let me know if I'm misunderstanding: yeah, we can't avoid dispatching bundles to dead workers wholesale. But, if it was refreshed 30 seconds earlier, it'd be re-refreshed 25 seconds later. At this point, we'd fetch from the database, clean up the dead workers, and then the bundles would be speedily dispatched to workers (or skipped, since there are no workers). So, the bundlemanager wouldn't stall any longer.

candicegjing:
> Let me know if I'm misunderstanding: yeah, we can't avoid dispatching bundles to dead workers wholesale. But, if it was refreshed 30 seconds earlier, it'd be re-refreshed 25 seconds later. At this point, we'd fetch from the database, clean up the dead workers, and then the bundles would be speedily dispatched to workers (or skipped, since there are no workers). So, the bundlemanager wouldn't stall any longer.

I see your point... I think there are 2 issues here. One is bundle manager stall issue, another is to keep the order of execution unchanged. Sure, after 25 seconds, the worker will be refreshed, and bundle manager won't stall anymore. But during these 25 seconds, there are bundles that will be dispatched to the dead workers, which mess up the order of execution. To me, it looks like we try to fix one problem, but another problem comes back.  I guess if we don't really care about the execution order for now, we can just go with this. If we want to have the order feature, we can fix it later?

nelson-liu:
> Sure, after 25 seconds, the worker will be refreshed, and bundle manager won't stall anymore. But during these 25 seconds, there are bundles that will be dispatched to the dead workers, which mess up the order of execution.

Ah, I see your point. This is why in this PR I make sure to explicitly only _remove_ workers that are dead. i.e., if a bundle is dispatched to a dead worker, and fails there are two options:

(1) there is another worker available to run it. In this case, the bundlemanager will try dispatching the bundle onto this worker.
(2) There is not another worker available to run it, so it's fine to skip it anyway.

Although we're refreshing the workers, the number of resources is monotonically decreasing in general. This is a rare case where, if a worker goes offline, but then comes back online with the same ID, that the filter wouldn't necessarily work—I can edit the PR to fix that and make the worker list cumulative.

Not sure if any of the above made sense, thoughts @candicegjing  ?

codalab/server/bundle_manager.py
```diff
@@ -412,7 +440,7 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
         # To avoid the potential race condition between bundle manager's dispatch frequency and
         # worker's checkin frequency, update the column "exit_after_num_runs" in worker table
         # before bundle manager's next scheduling loop
-        for worker in workers_list:
+        for worker in workers.workers():
```
nelson-liu: @candicegjing this is an unrelated fix, but I noticed that `workers_list` is essentially just the workers corresponding to the _last_ bundle, not all workers (as I think you intended it). Maybe you can check and verify this makes sense?

candicegjing: I think `workers_list` is what I want here because `workers_list` contains the minimum number of workers that would need to update their `exit_after_num_runs` value in the database (it's the list of eligible workers that could potentially run the current bundle).  Plus, `workers.workers()` doesn't reflect the changes of `exit_after_num_runs` for each worker, as we were calculating on a deep copy of the original object. But let me know if I misunderstood...

nelson-liu: > I think workers_list is what I want here because workers_list contains the minimum number of workers that would need to update their exit_after_num_runs value in the database (it's the list of eligible workers that could potentially run the current bundle)

oh, got it. I think you meant to indent this for loop by 4, then? Right now, it runs after all of the bundles are finished scheduling.

nelson-liu:
Ok, @percyliang / @teetone , here's a summary of what's going on here (and a recap of #2423 ):

#2423 tracks an issue where the BundleManager spends an inordinate amount of time in the `_schedule_run_bundles_on_workers` https://github.com/codalab/codalab-worksheets/blob/7426f4ca70a65a3f9e42926ea51cda7197b0d9ee/codalab/server/bundle_manager.py#L315 . After doing some digging, I figured out that the issue occurs because the worker information we have is stale—as a result, the BundleManager thinks that there are workers that can run jobs, when really they're offline. So, since the BundleManager thinks that the workers are available, it'll try to iteratively schedule each bundle on it (assuming the bundle meets resource requirements).

However, this scheduling is doomed to fail, since the worker is offline. When the `_try_start_bundle` function (https://github.com/codalab/codalab-worksheets/blob/7426f4ca70a65a3f9e42926ea51cda7197b0d9ee/codalab/server/bundle_manager.py#L398) fails, it sleeps for 0.3 seconds—as a result, each bundle takes at least 0.3 seconds to process. When we have lots of bundles (say, 10,000), this means that the BundleManager is stuck in this loop for 50 minutes. While it's in this loop, no new staged bundles can be run, no new workers will receive staged bundles, etc.

The fix I've applied here is to re-fetch the online workers before trying to schedule each bundle. The WorkerInfoAccessor will automatically refresh every 60 seconds, so by querying it before each bundle we schedule, a worker that goes offline will only be stale in the cache for a maximum of 60 seconds, a negligible amount. I store a running set of workers that have gone offline, in order to respect prioritization in the case where a worker is online, goes offline during bundle scheduling, and comes back online—we don't want the worker to be assigned bundles, since these bundles might have jumped the queue.

Let me know if that made sense / needs more clarification, happy to elaborate.

teetone:
@nelson-liu This makes sense to me.

teetone (approved):

PR #2425, opened 2020-06-18
test - use official python 3.6 images with debian buster instead of ubuntu

epicfaace:
This is an extension of #2278 for the server / worker images, but it's a bigger change than what's in that other PR.

Benefits of using the official python 3.6 images are -- reliability / security (official image is better than using the deadsnakes PPA) and speed (python is already installed on the base image).


epicfaace:
Actually, the built images with buster are actually much bigger (100 MB+ than ubuntu)

Issue #2395, opened 2020-06-14
keyboard shortcut for 'cut' not working


percyliang:
On Firefox and Chrome in Linux, 'a z' doesn't trigger 'Cut'.  Other keyboard shortcuts work, and cut works if you click on the button.

yipenghe:
can @epicfaace try it out, it's working for me on linux and windows on dev

percyliang:
If I replace `a z` with `a l` (or anything else) it works.  So weird.

Here's a suggestion that is both a workaround and perhaps a better solution overall: let's use `a d` for this, since with cut, I don't think we need detach anymore.  `a d` is also more ergonomic to type than `a z`.

yipenghe:
> If I replace `a z` with `a l` (or anything else) it works. So weird.
> 
> Here's a suggestion that is both a workaround and perhaps a better solution overall: let's use `a d` for this, since with cut, I don't think we need detach anymore. `a d` is also more ergonomic to type than `a z`.

Just to confirm, I will just remove the detach button and shortcut entirely and assign "a d" for cut

percyliang:
Yes.  Note: the keyboard shortcuts for this have been updated in #2409

PR #2427, opened 2020-06-18
Remove detach from the frontend

### LINKED ISSUE #2411: [Code cleanup] Deprecate detach on the frontend
Cut is better anyway

yipenghe:
fix #2411

### DIFF
--- a/frontend/src/components/worksheets/BundleBulkActionMenu.js
+++ b/frontend/src/components/worksheets/BundleBulkActionMenu.js
@@ -40,16 +40,6 @@ class BundleBulkActionMenu extends React.Component {
                 <Button
                     size='small'
                     color='inherit'
-                    aria-label='Detach'
-                    disabled={!this.props.editPermission}
-                    onClick={this.props.toggleCmdDialog('detach')}
-                >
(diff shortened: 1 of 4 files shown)


epicfaace (commented):
Looks good to me. Can you explain again the reason why we want to remove detach?

If someone wants to remove a bundle from the worksheet (but doesn't want to delete it), but _also_ doesn't intend to paste the bundle anywhere else, then I'm not sure if cut will still be intuitive for them?

yipenghe:
> Looks good to me. Can you explain again the reason why we want to remove detach?
> 
> If someone wants to remove a bundle from the worksheet (but doesn't want to delete it), but _also_ doesn't intend to paste the bundle anywhere else, then I'm not sure if cut will still be intuitive for them?

This was a follow up of #2395 , my thoughts on it is:
detach can only be used when there is one and only one reference of that bundle in the worksheet, so if we have:
[]{0x123}
[]{0x123}
we actually can't detach 0x123 in this case unless we edit source and remove one of them, otherwise it will report an error saying multiple reference, so i guess it isn't a useful operation itself
cc @percyliang

epicfaace (approved):

PR #2428, opened 2020-06-18
Remove select all for table

### LINKED ISSUE #2402: remove the checkbox for the table
While it's convenient for selecting/deselecting all the rows in the table, I think this is a rare operation and it adds more confusion and is distracting right now, so I think we should remove it.
Screenshot from 2020-06-14 14-30-33

yipenghe:
resolve #2402


epicfaace (changes requested):
Can you remove `this.handleSelectAllClick`, etc.?

epicfaace (approved):

PR #2433, opened 2020-06-18
Add a blog for announcements

epicfaace:
The blog now looks like this:

image


yipenghe:
Would it be better to place it at top maybe?

candicegjing (approved):
Looks great to me!

epicfaace:
> Would it be better to place it at top maybe?

I don't think so -- we'll probably have many more announcements and we wouldn't want to bloat the top with the list of all anouncements.

epicfaace:
Let's link to mailchimp directly instead.

PR #2449, opened 2020-06-19
Fix long test startup times by using python:3.6.10-slim-buster for default test runs

### LINKED ISSUE #2388: make first run on GitHub Actions not take 10 minutes

epicfaace:
Fixes #2388 by using python:3.6.10-slim-buster for default test runs.

Following @nelson-liu 's suggestion in https://github.com/codalab/codalab-worksheets/issues/2388#issuecomment-646436448.

This image is still around ~40 MB (as opposed to the 5 GB default-cpu image).

Time changes from ~20 mins -> 10 mins

### DIFF
--- a/scripts/create_sample_worksheet.py
+++ b/scripts/create_sample_worksheet.py
@@ -145,6 +145,8 @@ class SampleWorksheet:
                     'run',
                     '--request-memory',
                     '10m',
+                    '--request-docker-image',
+                    'python:3.6.10-slim-buster',
                     'echo codalab rules!',
                     '--tags=%s' % SampleWorksheet.TAG,
                 ]
@@ -162,6 +164,8 @@ class SampleWorksheet:
                     'run',
                     '--request-memory',
                     '10m',
+                    '--request-docker-image',
+                    'python:3.6.10-slim-buster',
                     'echo private run',
                     '--tags=%s' % SampleWorksheet.TAG,
                 ]
--- a/test_cli.py
+++ b/test_cli.py
@@ -1612,7 +1612,7 @@ def test(ctx):
 @TestModule.register('netcurl')
 def test(ctx):
     uuid = _run_command(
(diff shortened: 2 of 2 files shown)


percyliang:
Cool!

Going back, is there a way to cache the default-cpu image?  One small benefit of using that image is that it allows us to test our default images too.

epicfaace:
I'm not quite sure if that would help -- it appears that downloading from docker hub is actually faster than saving / caching images using github actions: https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei

Perhaps we should eventually use [docker-slim](https://github.com/docker-slim/docker-slim) to try to really make our default images smaller. For now, testing the default-cpu / default-gpu images on each build is probably not a priority, given that we don't publish images on each release (and when we do publish new default-cpu / default-gpu images, the codalab tests would only catch a regression after the fact because they use the `latest` version of these images). We could also just add a step where before we publish new docker codalab default-cpu / default-gpu images, we should test them (maybe by running the `docker` test in test_cli).

yipenghe:
Looks good, how long does it take to finish now

epicfaace:
@yipenghe the last build took 9 mins 31 s!

scripts/create_sample_worksheet.py
```diff
@@ -145,6 +145,8 @@ def _create_dependencies(self):
                     'run',
                     '--request-memory',
                     '10m',
+                    '--request-docker-image',
+                    'python:3.6.10-slim-buster',
```
epicfaace, line 149: This has some repetition -- we should refactor this to use something similar to `_run_command` in a separate PR.

yipenghe (approved):
Great!! Let's also add the test docker image before deployment step to the deployment repo

teetone:
@epicfaace Nice!!

PR #2451, opened 2020-06-19
Fix state transition in finalizing worker run state

candicegjing:
Fix state transition for restage in finalizing worker state.


epicfaace (commented):
What issue exactly is this fixing?

candicegjing:
> What issue exactly is this fixing?

This fix will correctly send bundle in the `finalizing` state back to the `staged` state when `pass-down-termination` option is specified for worker.

teetone (approved):

PR #2457, opened 2020-06-19
Fix github actions on forks by not calling --push

### LINKED ISSUE #2456: Github Actions CI fails on PRs from forks, since they can't access secrets
e.g., take a look at https://github.com/codalab/codalab-worksheets/pull/2424/checks?check_run_id=788984134 . It fails on the command:

```
subprocess.CalledProcessError: Command 'docker login --username  --password ' returned non-zero exit status 1.
```

I'm pretty sure the issue is that the secrets in the actions yml (https://github.com/codalab/codalab-worksheets/blob/master/.github/workflows/test.yml#L41-L43) aren't being shared with fork branches, since that'd be a security issue (a build could just print the secret, for instance)

epicfaace:
This conditional expression was in the old `.travis.yml`, but it isn't there in the new github actions workflow. This PR adds that expression so that `--push` is not called from forks.

Fixes #2456

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/setup-python@v1
         with:
           python-version: 3.6
-      - run: python3 codalab_service.py build --pull --version ${VERSION} --push -s ${SERVICE}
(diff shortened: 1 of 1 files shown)


nelson-liu (approved):
LGTM, thanks for the quick fix!

PR #2461, opened 2020-06-19
Actually fix builds from forks by building docker images when needed

epicfaace:
Actually fix builds from forks by building docker images when needed.

Note that this makes builds from forks take much longer on github actions because 1) the original docker image build steps can't be reused (since they can't be pushed / pulled to be reused in subsequent steps) and 2) each test suite job will end up building all images from scratch.

This will fix builds for now (and it'll probably still be faster than travis), but would appreciate some suggestions as to how to improve this workflow (@nelson-liu )

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Setup tests
         run: |
           sudo service mysql stop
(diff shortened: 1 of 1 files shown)


nelson-liu (approved):
thanks! 🤞

PR #2463, opened 2020-06-19
Share workers among group members

### LINKED ISSUE #2220: [FEATURE REQUEST] Sharing workers amongst user groups
I'd like to share a worker node (e.g. a computer at in a lab) among several users, however currently cl-worker is attached to a single user

A cl-worker is attached to a user group as opposed to a single user

For now, pick up bundles that you have `all` permissions for.

teetone:

- Allow users to specify a group when creating a worker
- Populate the group in the database on worker check-in
- When the worker cache is being updated, add all the workers the user owns + has access to

### Related issues

Resolves #2220 

### Checklist

* [x] I've added a stress test


codalab/model/bundle_model.py
```diff
@@ -1907,8 +1907,12 @@ def batch_get_user_in_group(self, **kwargs):
                 return []
         return [str_key_dict(row) for row in rows]
 
-    # Helper function: return list of group uuids that |user_id| is in.
-    def _get_user_groups(self, user_id):
+    def get_user_groups(self, user_id):
```
epicfaace, line 1920: Interesting that we already supported groups earlier. What were groups used for, if not for sharing workers?

codalab/server/worker_info_accessor.py
```diff
@@ -18,21 +19,32 @@ def wrapper(*args, **kwargs):
...
+    def __init__(self, model, worker_model, timeout_seconds):
         self._model = model
+        self._worker_model = worker_model
         self._timeout_seconds = timeout_seconds
         self._last_fetch = None
         self._fetch_workers()
 
     def _fetch_workers(self):
-        self._workers = {worker['worker_id']: worker for worker in self._model.get_workers()}
+        self._workers = {worker['worker_id']: worker for worker in self._worker_model.get_workers()}
```
epicfaace, line 38: What's the difference between `self._model.get_workers()` and `self._worker_model.get_workers()`?

codalab/server/worker_info_accessor.py
```diff
@@ -18,21 +19,32 @@ def wrapper(*args, **kwargs):
...
                 self._uuid_to_worker[uuid] = worker
-            self._user_id_to_workers[worker['user_id']].append(worker)
+
+            owner_id = worker['user_id']
+            self._user_id_to_workers[owner_id].append(worker)
+
+            # Add the worker to all the users of worker's group except the owner
+            memberships = self._model.batch_get_user_in_group(group_uuid=worker['group_uuid'])
+            for m in memberships:
+                if m['user_id'] != owner_id:
```
epicfaace, line 53: Probably could make _user_id_to_workers a defaultdict(set) so we wouldn't need this check, but that's a refactoring that's not needed in this PR

codalab/model/bundle_model.py
teetone: Groups are also used for worksheets and bundles.

codalab/server/worker_info_accessor.py
teetone: `_worker_model` has the logic to access to workers from the database, while `_model` is more general with access to other tables in the database. I needed both for this new feature.

codalab/objects/permission.py
```diff
@@ -80,24 +79,27 @@ def get_single_group(model, group_spec, search_fn):
 
 
 def _check_permissions(model, table, user, object_uuids, owner_ids, need_permission):
+    # Validate parameter values
```
epicfaace, line 82: Why was this moved earlier in the function?

docs/Execution.md
```diff
@@ -80,7 +80,13 @@ If the default workers are full or do not satisfy your needs, one of the advanta
 
 You should see that the run finished, and if you look at the `remote` metadata field, you should see your hostname.
 
-Note that only your runs will be run on your workers, so you don't have to worry about interference with other users.
+**Sharing a worker with a group** 
+
+You can also share your worker with other users within a group, by simply starting a worker with a valid group name specified:
```
epicfaace: Can you clarify what exactly "share your worker with other users within a group" entails? What additional permissions will those users within a group have?

test_cli.py
```diff
@@ -100,6 +101,52 @@ def current_user():
...
+    # Currently there isn't a method for creating a user with the CLI. Use CodaLabManager instead.
+    manager = CodaLabManager()
+    model = manager.model()
+
+    # Creates a user without going through the full sign-up process
+    model.add_user(
+        username, random_name(), '', '', password, '', user_id=username, is_verified=True
+    )
+    context.collect_user(username)
+    switch_user(username, password)
```
epicfaace: Can we remove this line, and then split out `create_user` and `switch_user` in the tests? It's not immediately clear from just `create_user` calls that this call also logs as to that particular user.

test_cli.py
```diff
@@ -364,9 +413,19 @@ def __exit__(self, exc_type, exc_value, tb):
...
+        # Delete all the extra users created during testing
+        model = self.manager.model()
+        for user in self.users:
+            model.delete_user(user)
+
         # Delete all groups (dedup first)
-        if len(self.groups) > 0:
-            _run_command([cl, 'grm'] + list(set(self.groups)))
+        for group in list(set(self.groups)):
+            _run_command([cl, 'grm', group])
```
epicfaace: Why change this? Wasn't the former (batch delete) more efficient?

test_cli.py
```diff
@@ -1918,7 +1980,50 @@ def test_workers(ctx):
...
+    add_user_to_group('group2', 'userB')
+
+    # Validate user's access to the newly created workers
+    check_workers('userA', ['worker2'])
+    check_workers('userB', ['worker1', 'worker2'])
+    check_workers('userC', [])
+    check_workers('userD', ['worker2'])
+
+    # Switch back to the root user for proper cleanup
+    switch_user('codalab')
```
epicfaace: Can we move this line to the `__enter__` method so it's called as a general startup function after all tests?

codalab/objects/permission.py
teetone: This doesn't affect functionality. I just think it's better to validate the parameter values at the beginning of the function.

test_cli.py
teetone: I might be wrong, but I don't think bulk delete is supported for `grm`.

Screen Shot 2020-09-12 at 1 16 59 AM

test_cli.py
teetone: That's cleaner. I updated.

docs/Execution.md
teetone: I updated the documentation. Hopefully, it's clearer.

epicfaace (approved):

Issue #2442, opened 2020-06-19
Runs fail because docker image (in codalab cache) cannot be found


nelson-liu:
**Describe the bug**

Sometimes, runs fail with error messages like:
```
Failed to download Docker image: Image nfliu/architecture-generalization-reading-comprehension-albert:latest was downloaded successfully, but it cannot be found locally due to unhandled error 404 Client Error: Not Found ("No such image: codalab-image-cache/last-used:1592516003.596136")
```

Looking at some worker logs, I see the following happening:

```
50607-2020-06-06 15:38:32,817 http://localhost:None "DELETE /v1.35/images/codalab-image-cache/last-used:1591433958.0564198?force=False&noprune=False HTTP/1.1" 404 78
50608:2020-06-06 15:38:32,818 Failed to download Docker image: Image nfliu/architecture-generalization-reading-comprehension-fusionnet:latest was downloaded successfully, but it cannot be found locally due to unhandled error 404 Client Error: Not Found ("No such image: codalab-image-cache/last-used:1591433958.0564198")
```

I haven't dug into the code, but it seems like images are getting pruned right before they have to get used, so when the worker tries to start a container on them, it can't.

PR #2467, opened 2020-06-19
Enable workers to exit on exceptions, instead of sleeping

### LINKED ISSUE #2441: Workers should have a flag that enable them to exit on exceptions

At the moment, when the worker runs into an exception, it'll sleep for an hour. https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L206-L211

This behavior isn't very good for workers that run on batch compute systems (e.g., Slurm or AWS Batch)—the worker will keep the job active, hogging resources, when often the error isn't something that can be recovered from by simply waiting (e.g. #2439 ). It'd be nice if there was a flag that enabled workers to terminate when they hit an exception, instead of sleeping in a loop forever.

nelson-liu:
Fixes #2441 

This is important especially when workers on running on batch compute, since some errors are irrecoverable (e.g., #2442 ). In this case, it's easier / more-efficient to just exit the worker and restart, but this isn't currently possible because workers cannot exit on Exceptions.

### DIFF
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -147,6 +147,11 @@ def parse_args():
         default=sys.maxsize,
         help='The worker quits after this many jobs assigned to this worker',
     )
+    parser.add_argument(
+        '--exit-on-exception',
+        action='store_true',
+        help="Exit the worker if it encounters an exception (rather than sleeping).",
+    )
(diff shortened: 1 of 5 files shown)


candicegjing (approved):
LGTM!

PR #2477, opened 2020-06-20
Remove the quotes around record item values

### LINKED ISSUE #2608: Remove quotes around record item

yipenghe:
resolve #2608


frontend/src/components/worksheets/items/RecordItem.js
```diff
@@ -62,19 +62,19 @@ class RecordItem extends React.Component {
...
-        var items = item.rows.map(function(item, index) {
-            var ref = 'row' + index;
+        let className = 'table table-record' + (focused ? ' focused' : '');
+        let bundleInfo = item.bundles_spec.bundle_infos[0];
+        let header = item.header;
+        let k = header[0];
+        let v = header[1];
+        let items = item.rows.map(function(item, index) {
+            let ref = 'row' + index;
+            let displayValue = JSON.stringify(item[v]); // stringify is needed to convert metadata objects
```
epicfaace, line 72: Why can't we just do `String(item[v])` instead of doing json stringify and then substringing it?

yipenghe: String() converts object to [object object] (literal) rather than giving the value

epicfaace: Right, but when the value is a string, we'd want to do `String(item[v])`, right?

yipenghe: item[v] is not guaranteed to be a string, which is why there was already a json.stringify before

epicfaace: When would item[v] be an object though? I'd think it would only be a string / int / etc.?

epicfaace (approved):

frontend/src/components/worksheets/items/RecordItem.js
yipenghe: I'm not quite sure but there was a issue regarding this and it turned out the value was an object

PR #2478, opened 2020-06-20
CI: Remove some parallelism

epicfaace:
Remove some unnecessary parallelism in github actions, so that we require less parallel workers per job.

The build still takes ~10 mins to complete, but it now requires only 16 parallel jobs instead of 22.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,8 +59,7 @@ jobs:
           - search read kill write mimic workers
           - resources
           - memoize
-          - copy netcat
-          - netcurl
+          - copy netcat netcurl
     steps:
       - name: Clear free space
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,7 +52,7 @@ jobs:
     needs: [build]
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     needs: [build]
     strategy:
       matrix:
-        test: [unittest gen-rest-docs gen-cli-docs gen-readthedocs basic status batch anonymous competition unicode rest1 upload1 upload2 upload3 upload4 download refs binary rm make worksheet_search worksheet_tags freeze detach perm search_time groups,search,run,run2,read,kill,write,mimic workers,resources,copy memoize,netcat,netcurl]
+        test: [unittest gen-rest-docs gen-cli-docs gen-readthedocs basic status batch anonymous competition unicode rest1 upload1 upload2 upload3 upload4 download refs binary rm make worksheet_search worksheet_tags freeze detach perm search_time groups,run,run2,search read kill write mimic workers,resources,memoize,copy netcat,netcurl]
```
teetone: Is it possible to add new lines in `yml` files? This line is a bit too long.

teetone (approved):

PR #2492, opened 2020-06-23
Backward compatibility

### LINKED ISSUE #2491: Ensure backward compatibility of worker run state
During the release of v0.5.15, we introduced a new field in RunState (a namedtuple object). In the deployment, since the old worker run state doesn't have this new field, we had some failures in starting workers with new logic. This ticket is to add this backward compatibility into our logic so that there won't be any manual work during future releases.

candicegjing:
Fixed #2491.

Added some logic to sync the old run state object with the new run state object. defaulting the fields with the value `None` for now, although it might not be the best default value to use. So please feel free to suggest.


nelson-liu:
> defaulting the fields with the value None for now, although it might not be the best default value to use. So please feel free to suggest.

Would it work to use the same default as https://github.com/codalab/codalab-worksheets/blob/ac11edc772c1a7aef0dc12e18a70ddfe149f90b1/codalab/worker/worker.py#L536-L553 ?

candicegjing:
> > defaulting the fields with the value None for now, although it might not be the best default value to use. So please feel free to suggest.
> 
> Would it work to use the same default as
> 
> https://github.com/codalab/codalab-worksheets/blob/ac11edc772c1a7aef0dc12e18a70ddfe149f90b1/codalab/worker/worker.py#L536-L553
> ?

I think that's a good idea! There seems to have [a `defaults` parameter](https://docs.python.org/3/library/collections.html#collections.namedtuple) that you can specify for `namedtuple` from the rightmost to left parameters. The above parameters when we initiate a RunState object could have been defined in `defaults`, but it might reduce the readability of code as of it's now. So I just defined the defaults of the new field in the RunState object and keep the rest of the code as it is right now.

nelson-liu (changes requested):
The `defaults` field of `namedtuple` isn't in python3.6 :( Would it work to just have a class? Or is it hard to do backwards-compatible serialization in that case?

Side note: If we weren't constrained to 3.6, I feel like `DataClasses` would be perfect here https://docs.python.org/3/library/dataclasses.html

candicegjing:
> The `defaults` field of `namedtuple` isn't in python3.6 :( Would it work to just have a class? Or is it hard to do backwards-compatible serialization in that case?

ah, you are exactly right! It's not in python3.6...... >.<  
Plus I think we probably need to stay on python3.6 for a while...
I guess using `None` as default for an older version of RunState would probably work, as it covers basically `False`, `0` and `None` in Python3. Plus, this is an old run, which doesn't really need those additional new fields at all. (although this is not a perfect solution, but I guess it works.)

codalab/worker/worker.py
```diff
@@ -168,6 +168,23 @@ def load_state(self):
...
+        """
+        for uuid, run_state in self.runs.items():
+            values = [getattr(run_state, name) for name in run_state._fields]
+            old_len = len(run_state._fields)
+            new_len = len(RunState._fields)
+            # When there are additional new fields detected, recreate the run_state object
+            # to include those new fields with default "None" specified from RunState object
+            if old_len < new_len:
+                for i in range(old_len, new_len):
+                    values.append(None)
```
teetone, line 197: Does this depend on the order of the field names of `RunState` defined in `worker_run_state.py`?

percyliang: Would be good to write this in a way that looks at the field names rather than assuming size and order are meaningful.

candicegjing: Sure. I originally thought these existing fields are needed to keep the current worker functional, and appending new fields at the end seems to be reasonable, so I compare the length and keep the order of the existing run state object. But looking for field names is definitely more robust in case we want to change the list in the future.

teetone (approved):

PR #2496, opened 2020-06-24
Switch docs site to mkdocs-material

epicfaace:
Switched the docs site to use the mkdocs-material theme.

I like this because:
- Looks better / more consistent with our main site, which is material-ui themed
- URLs for docs stay the same, as we're still using mkdocs
- We are more able to organize docs using tabs -- this provides a clear path for us to add things such as tutorials / guides without making the existing readthedocs site more cluttered
- Subsections are rendered as a separate navigation on the right instead of showing up on the left navigation, which again makes it feel less cluttered when you have large pages in the documentation

image


yipenghe:
Looks good to me, it's nice to have a more consistent style with the website, should we change?@percyliang

epicfaace:
@percyliang what do you think about the new mkdocs theme?

percyliang (approved):
Looks great!

PR #2499, opened 2020-06-24
Use docker BuildKit for building images

epicfaace:
Use [Docker BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/). Has a more interactive UI for development and is also faster to build images on CI.

BuildKit is only supported on Docker 18.09+, but with this change, people who are using older versions of Docker will still have the same experience as before.

For example, on the local development environment, we have a cleaner UI with time metrics:

image

CI also gives us time metrics now:

image

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -584,7 +584,7 @@ class CodalabServiceManager(object):
             success = True
         else:
             try:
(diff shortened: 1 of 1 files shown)


epicfaace:
Cache isn't working with buildkit, so let's hold off on this.

PR #2506, opened 2020-06-25
Don't limit WorkerManager staged bundle search by provided tags

nelson-liu:
It's nice to be able to specify tags for workers, but still leave them to run untagged bundles. The current behavior is that more workers won't be launched if there are staged bundles that match all other criteria except the tag.


nelson-liu:
in case the motivation for this fix isn't clear: in general, staged bundles should correspond to bundles that the workers launched by the workermanager can run.

candicegjing (approved):

PR #2508, opened 2020-06-25
Add --worker-pass-down-termination flag to WorkerManager

nelson-liu:
This enables use in the AWSBatchWorkerManager as well, useful for when spot instances get pre-empted.

### DIFF
--- a/codalab/worker_manager/aws_batch_worker_manager.py
+++ b/codalab/worker_manager/aws_batch_worker_manager.py
@@ -102,6 +102,8 @@ class AWSBatchWorkerManager(WorkerManager):
             command.extend(['--exit-after-num-runs', str(self.args.worker_exit_after_num_runs)])
         if self.args.worker_exit_on_exception:
             command.extend(['--exit-on-exception'])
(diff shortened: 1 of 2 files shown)


codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -102,6 +102,8 @@ def start_worker_job(self):
             command.extend(['--exit-after-num-runs', str(self.args.worker_exit_after_num_runs)])
         if self.args.worker_exit_on_exception:
             command.extend(['--exit-on-exception'])
+        if self.args.worker_pass_down_termination:
+            command.extend(['--pass-down-termination'])
```
percyliang, line 106: I know this is orthogonal to this PR, but seems like this block can be pulled out and shared across all worker managers?

percyliang (approved):

codalab/worker_manager/aws_batch_worker_manager.py
nelson-liu: Yeah, that's a good idea.

PR #2511, opened 2020-06-25
Set default value of exit_after_num_runs to maxint in rest API call

candicegjing:
This PR sets the default value of `exit_after_num_runs` to max int in rest call to avoid potential failures.

### DIFF
--- a/alembic/versions/2020060503_add_exit_after_number_of_runs__54e95ff2e718.py
+++ b/alembic/versions/2020060503_add_exit_after_number_of_runs__54e95ff2e718.py
@@ -12,7 +12,7 @@ down_revision = '9d0f1ffb18e9'
 
 from alembic import op
 import sqlalchemy as sa
-import sys
+
 
 def upgrade():
     op.add_column('worker', sa.Column('exit_after_num_runs', sa.Integer(), nullable=False, server_default='999999999'))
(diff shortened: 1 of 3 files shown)


nelson-liu (approved):

epicfaace (changes requested):
I don't think this will work -- see https://github.com/codalab/codalab-worksheets/pull/2403. Let's set it to 999999999 instead?

candicegjing:
> I don't think this will work -- see #2403. Let's set it to 999999999 instead?

I don't think these two are the same issue. #2403 is because the `server_default` needs to take a `string`. Here what I am changing is to ensure when a user uses an older version of the `codalab` package on client side, they can still run their `cl-worker` without a failure of missing column. 
Also just for reference: see comment: https://github.com/codalab/codalab-worksheets/pull/2288#discussion_r429599614 and worker side is defined as follows:
https://github.com/codalab/codalab-worksheets/blob/37d948d6be5068ebb5d60b54ee544fd90b7330d0/codalab/worker/main.py#L144-L149

candicegjing:
@epicfaace If we really think maxint is too large, I think we should make all three places consistent.

nelson-liu:
+1 to making all 3 consistent . 999999999 seems fine to me

nelson-liu (approved):

PR #2518, opened 2020-06-27
[Don't update branch, don't squash merge] Merge staging -> master

epicfaace:
Staging branch.

Make sure you merge this branch with a merge commit so that the v0.5.16 tag remains intact.


teetone:
@epicfaace I don't think we should merge `master` back into `staging`. I think we should keep them separate because we would have to retest all the new changes in master, making the deployment to `prod` slow.

scripts/travis-deploy.sh
```diff
@@ -13,7 +13,7 @@ tag=$1
 # by an external user. Then we shouldn't push the docker image to docker hub.
 PUSH_FLAG=$([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
 
-python3 codalab_service.py build --version $tag --pull $PUSH_FLAG
+python3 codalab_service.py build --version v$tag --pull $PUSH_FLAG
```
epicfaace, line 16: Our old Travis build step, for each release X.X.X, pushes a docker image with tag “X.X.X” and a docker image with tag “vX.X.X”

Our GitHub Actions workflow only pushes images with tag “vX.X.X” (this change makes that happen).

We only need “vX.X.X” images for deployment, though — the “X.X.X” images appear to be unused.

teetone:
@epicfaace  Is merging staging back into master still necessary?

teetone (approved):

epicfaace:
@teetone I think so? Otherwise, we may force push staging later and might still want the v0.5.16 tag to still point to a commit on master.

PR #2520, opened 2020-06-29
Add new run, text, upload above the first item when no focus

### LINKED ISSUE #2400: 'g g' should go to the very top
Currently, it goes to the first item, but it should go beyond that to focus index `-1`, because that's what happens when you repeatedly move up (e.g., hit 'k' repeatedly) for consistency.

### LINKED ISSUE #2482: can't text/upload/run on a newly loaded worksheet
To reproduce, create a new worksheet with one bundle on it.  Refresh the page.  Then click 'text', 'upload', or 'run'.  Nothing happens.  You have to select the last row for it to work.

This is because the focus is at -1 when the page loads.

In the past, we have appended new blocks to the worksheet, in concordance with what happens on the CLI.  However, this doesn't mesh well with the intuition that the cursor is right before the first item on the page, so the actually new blocks should be prepended to the worksheet.  Having a reverse chronological research log isn't itself a bad idea, but the problem is that the whole CodaLab model of always referring to the *last* bundle gets broken.

Perhaps a safe thing to do is only allow text/run/upload if the focus is not -1 on a non-empty worksheet, and always put a 'Go to end' button.

Or we can just have an as-is mentality and just prepend when focus is -1.

yipenghe:
This allows user to add items above the first item and reduce the need to edit in source, which aligns with our frontend goal (help user not use source as much as possible)
fix #2482 
also fix #2400 since it was a one line change


frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -197,30 +196,25 @@ class WorksheetItemList extends React.Component {
...
             var worksheet_items = [];
             info.blocks.forEach(
                 function(item, index) {
                     const focused = index === this.props.focusIndex;
 
-                    // focusedForButtons determines whether clicking on Cell/Upload/Run will
+                    // focused determines whether clicking on Cell/Upload/Run will
                     // apply to this cell. If nothing is focused (focusIndex = -1),
-                    // append to the end by default.
-                    const focusedForButtons =
```
yipenghe, line 210: This originally was used for appending to the end, since we are now prepending to the top of worksheet, no need for this anymore

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -124,14 +123,14 @@ class WorksheetItemList extends React.Component {
     }
 
     capture_keys() {
-        // Move focus to the top
+        // Move focus to the top, above the first item of worksheet
```
yipenghe, line 126: This resolves #2400

epicfaace (approved):

PR #2540, opened 2020-07-01
Create pull_request_template.md

epicfaace:
Add a PR template -- this will make sure PRs are created in a consistent format with screenshots, etc.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,18 @@
+### Reasons for making this change
+
+<!-- Add a reason for making this change here. -->
+
+### Related issues
+
+<!-- Add a reference to issues resolved, if applicable (for example, "fixes #1"). -->
+
+### Screenshots
+
+<!-- Add screenshots, if necessary -->
+
+### Checklist
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Screenshots
+
+<!-- Add screenshots, if necessary -->
+
+### Checklist
+
+* [ ] I've added a screenshot of the changes, if this is a frontend change
+* [ ] I've added and/or updated tests, if this is a backend change
+* [ ] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
```
teetone, line 17: I'm not sure if we need to explicitly add this, since the build will fail if the style check fails.

teetone (approved):

.github/pull_request_template.md
epicfaace: I don't think we have checks yet to make sure that the latest CLI / REST documentation is generated, though.

PR #2549, opened 2020-07-02
Fix indentation when updating workers' exit_after_num_runs in BundleManager

nelson-liu:
Right now, the BundleManager only updates `exit_after_num_runs` for workers that are eligible to run the _last_ bundle , since `workers_list` is shadowed in the loop and recreated. I'm pretty sure this is not the intended behavior / a bug.

I'm pretty sure we want to be running this after every bundle dispatch instead, ref: https://github.com/codalab/codalab-worksheets/pull/2424#discussion_r447828225

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -381,7 +381,6 @@ class BundleManager(object):
             workers.user_owned_workers(self._model.root_user_id), running_bundles_info
         )
 
-        workers_list = []
         # Dispatch bundles
         for bundle, bundle_resources in staged_bundles_to_run:
             if user_parallel_run_quota_left[bundle.owner_id] > 0:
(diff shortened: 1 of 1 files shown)


codalab/server/bundle_manager.py
```diff
@@ -409,20 +408,20 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
...
-            if (
-                worker['exit_after_num_runs']
-                < workers._workers[worker['worker_id']]['exit_after_num_runs']
-            ):
-                self._worker_model.update_workers(
-                    worker["user_id"],
-                    worker['worker_id'],
-                    {'exit_after_num_runs': worker['exit_after_num_runs']},
-                )
+            # To avoid the potential race condition between bundle manager's dispatch frequency and
```
teetone, line 411: @nelson-liu If I'm not mistaken, I think the old indentation was correct. The new indentation will update the worker table in the database for every staged bundle. Workers with a value of 0 for `exit_after_num_runs` will get filtered out in line 395.

nelson-liu: Hmm, how was this code working before, then? as-is, the value of `workers_list` depends on what the workers for the _last_ bundle are (i.e., it's bundle-order dependent). So, you could have different results depending on the bundle order, which doesn't seem ideal...thoughts?

nelson-liu:
Closing this since things seem to be working ok, but we should revisit as part of rethinking the bundlemanager #2555

PR #2557, opened 2020-07-02
Fix resolution of aggregate search directives

### LINKED ISSUE #2205: Aggregate search queries are rendered twice on the dashboard
Some contents are rendered twice on the dashboard:
https://worksheets.codalab.org/worksheets/0xb843335f02a9443995db2013fbe2e7f4/
Screenshot from 2020-04-23 21-56-18
Maybe this has something to do with the recent changes for async loading? @epicfaace

epicfaace:
Fixes #2205.

When the worksheet source looks like this,

image

This PR fixes it so that the result isn't duplicated -- it instead resolves to two markdown blocks (and the search result is read-only):

image


epicfaace:
Need to change this so that it renders the search result on a new block, not as part of the topmost block.

yipenghe:
I think we want it to be on the same block
for example:
a
b

would appear as one block, unless you add # a to mark it as a title

epicfaace:
> I think we want it to be on the same block

I thought the previous behavior was for it to be on different blocks if the block below it is rendering a search result -- at least that's what @percyliang had mentioned when we last discussed this issue. I don't recall noticing earlier, though, though it'd be easiest at this point to just keep them having separate blocks.

epicfaace:
@yipenghe this is ready for another review -- I made it so that the second markdown block is now read-only.

codalab/rest/interpret.py
```diff
@@ -260,16 +261,22 @@ def fetch_interpreted_worksheet(uuid):
...
         # Only expand the search item corresponding to the given directive.
         # Used in async loading to only load a single table.
-        item_idx = 0
+        new_items = []
         for i, item in enumerate(worksheet_info['items']):
             (bundle_info, subworksheet_info, value_obj, item_type, id, sort_key) = item
             if directive == formatting.tokens_to_string(value_obj):
                 search_results = perform_search_query(value_obj)
-                item_idx = i
+                new_items.append(item)
```
yipenghe: I'm kind of confused by the name new_items, what does it actually include? Maybe change to a clearer name or add a comment saying what's inside?

yipenghe (approved):

PR #2560, opened 2020-07-03
Don't filter workermanager to current user's staged bundles if user is codalab

nelson-liu:


codalab/worker_manager/worker_manager.py
```diff
@@ -104,7 +104,12 @@ def run_loop(self):
 
     def run_one_iteration(self):
         # Get staged bundles for the current user.
-        keywords = ['state=' + State.STAGED] + [".mine"] + self.args.search
+        keywords = ['state=' + State.STAGED] + self.args.search
+        # If the current user is "codalab", don't filter by .mine because the workers owned
+        # by "codalab" can be shared by all users. But, for all other users, we only
+        # want to see their staged bundles.
+        if os.environ.get('CODALAB_USERNAME') != "codalab":
```
percyliang, line 112: Is this the same username that the `cl-worker` launched uses?

nelson-liu: yes, for both the aws and slurm worker managers.

percyliang (approved):
Maybe add your comment that the principle is that you want to get all the bundles that can be run by this user.

PR #2568, opened 2020-07-07
Remove test cleanup on CI

epicfaace:

Remove test cleanup on CI, to make it faster. We don't need to delete all the created bundles if we're running on GitHub Actions.

GHA always sets the `CI` environment variable to true: https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables


test_cli.py
```diff
@@ -334,33 +334,6 @@ def __exit__(self, exc_type, exc_value, tb):
         os.environ.clear()
         os.environ.update(self.original_environ)
 
-        _run_command([cl, 'work', self.original_worksheet])
```
teetone, line 337: If we get rid of this, we populate the instances that we run the tests on without cleaning it up. Does this mean we will only use GHA to run tests going forward? I'm fine with this too, since the GHA builds are fast, but at the same time, I'm not sure if we want to lose the flexibility on running on any instance.

percyliang: I'd like tests to work locally too.  Can we make it so that you can clear any old testing state with one command?

epicfaace: Got it -- I've instead changed it so that it skips cleanup only in CI environments.

teetone (approved):

PR #2576, opened 2020-07-08
Revert "Bump query-string from 6.2.0 to 6.13.1 in /frontend"

epicfaace:
Reverts codalab/codalab-worksheets#2574

Accidentally merged this

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -14293,12 +14293,11 @@
       "dev": true
     },
     "query-string": {
-      "version": "6.13.1",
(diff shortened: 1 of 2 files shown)


yipenghe (approved):
What does this upgrade break though

epicfaace:
@yipenghe I'm not quite sure. It seems that all our tests have suddenly started timing out -- even this PR doesn't fix it though, so it might not even have to do with the query-string upgrade.

It's probably something to do with GHA?

PR #2587, opened 2020-07-09
Prune unused networks before creating new ones

nelson-liu:

Docker's default local bridge network only supports 30 different networks. If we try to make more, we get the error: "ERROR: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network"

This PR fixes this case by just pruning old unused networks before making new ones.


codalab/worker/worker.py
```diff
@@ -141,6 +141,13 @@ def create_or_get_network(name, internal, verbose):
                     logger.debug('Network %s already exists, reusing', name)
                 return self.docker.networks.list(names=[name])[0]
 
+        # Docker's default local bridge network only supports 30 different networks
+        # (each one of them uniquely identifiable by their name), so we prune old,
+        # unused docker networks, or network creation might fail. We only prune docker networks
+        # older than 1h, to avoid interfering with any newly-created (but still unused) networks
+        # that might have been created by other workers.
+        self.docker.networks.prune(filters={"until": "1h"})
```
epicfaace, line 149: Why 1 hour? How long does a newly-created (but still unused) network created by a worker usually stay unused until it is used?

nelson-liu: the 1h is arbitrary and also super conservative—i'd imagine it's usually ~5m max, but I figured that pruning with any frequency at all would fix the issue.

epicfaace (approved):

nelson-liu:
thanks for taking a look @epicfaace !

PR #2588, opened 2020-07-09
Give text item correct sort key

### LINKED ISSUE #2421: insertion of bundle happens in wrong place
To reproduce, create a new worksheet, add ```# a
b```
Then upload a bundle.  The bundle gets inserted in between rather than at the end.
Screenshot from 2020-06-17 00-14-30

yipenghe:

Fix bug
### Related issues

fix #2421 

The cause:
Each row of markup are individual items and have their own sort keys but consecutive markup rows are rendered as one item on the frontend (think bundle and bundle tables). However unlike bundle tables, where each bundle rows can be focused separately, markdown blocks has only one item and thus the subFocusIndex is always 0. getAfterSortKey returns the sort key corresponding to subFocusIndex when there is one defined, but in markup block, we should return the largest sort key, which corresponds to the last line's sort key.

Also fixed a bug: Edit a markup block and make changes, then press 'esc'. Then hit "ctrl + enter" to save, it will still save the changes


epicfaace (commented):
I'm able to verify that the second bug is fixed, but I'm not able to reproduce the first bug on worksheets.codalab.org. How exactly did you end up reproducing it?

yipenghe:
> I'm able to verify that the second bug is fixed, but I'm not able to reproduce the first bug on worksheets.codalab.org. How exactly did you end up reproducing it?

In source:
1: # a
2: b
save, it should save as one markup block
then try inserting after the one markup block

epicfaace (approved):

PR #2600, opened 2020-07-15
2547: Support protected mode

### LINKED ISSUE #2547: protected mode for CodaLab instances
By default, anyone (without an account) can (i) sign up for an account and (ii) view all public-readable bundles/worksheets.  Sometimes, we want CodaLab instances which are on the public Internet so that AWS workers can connect to them but are effectively locked down (in protected mode).  In this mode, only users with accounts can log in.

If you're not logged in, everything returns access denied except for the login page (including all REST endpoints, so that the CLI won't work either unless you log in).  

Anyone can also still sign up, but any signup produces an email containing a link that an admin has to approve (in addition to the usual email verification of the user).

teetone:

Allows CodaLab instances to be on the public internet, but in protected mode.

### Related issues

Resolves #2547 

### Screenshots

Screen Shot 2020-07-18 at 12 13 54 AM

### Checklist

  Follow up: #2611


codalab/server/authenticated_plugin.py
```diff
@@ -1,52 +1,87 @@
...
+import os
 
-def user_is_authenticated():
-    return hasattr(request, 'user') and request.user is not None and request.user is not PUBLIC_USER
 
+class AuthenticationPlugin(ABC):
+    _NOT_AUTHENTICATED_ERROR = 'User is not authenticated'
+    _NOT_VERIFIED_ERROR = 'User is not verified'
+    _ACCESS_DENIED_ERROR = (
+        'User is not given access to this CodaLab instance. Please contact the admin for access.'
```
epicfaace:
```suggestion
        'User has not been given access to this CodaLab instance. Please contact the admin for access.'
```

codalab/server/authenticated_plugin.py
```diff
@@ -59,16 +94,44 @@ class AuthenticatedPlugin(object):
...
+        def wrapper(*args, **kwargs):
+            self.check_user_authenticated()
+            return callback(*args, **kwargs)
+
+        return wrapper
+
+
+class AuthenticatedProtectedPlugin(AuthenticationPlugin):
+    """
+    Fails the request if the user is not authenticated. It doesn't matter if the instance is protected or not.
```
epicfaace: I think this should clarify the comment more:

```suggestion
    Fails the request if the user is not authenticated. If the instance is protected, fails the request if the user has not been granted access to the instance.
```

codalab/server/authenticated_plugin.py
```diff
@@ -1,52 +1,87 @@
...
-            if not user_is_authenticated():
+            if not self.user_is_authenticated():
                 request.user = PUBLIC_USER
             return callback(*args, **kwargs)
 
         return wrapper
 
 
-class AuthenticatedPlugin(object):
+class AuthenticatedPlugin(AuthenticationPlugin):
```
epicfaace: Can we remove `AuthenticatedPlugin` altogether? It seems that we should be fine to change the remaining endpoints where this plugin is used (such as /account/reset, etc.) to use `AuthenticatedProtectedPlugin` instead.

codalab/server/authenticated_plugin.py
```diff
@@ -1,52 +1,87 @@
...
 
 from codalab.lib.server_util import redirect_with_query
 from codalab.objects.user import PUBLIC_USER
 
+import os
 
-def user_is_authenticated():
-    return hasattr(request, 'user') and request.user is not None and request.user is not PUBLIC_USER
 
+class AuthenticationPlugin(ABC):
```
epicfaace: Can we rename this to `AuthPlugin`? This will make it clearer and less easy to confuse with `AuthenticatedPlugin`.

codalab/server/authenticated_plugin.py
```diff
@@ -1,52 +1,87 @@
...
+    _NOT_AUTHENTICATED_ERROR = 'User is not authenticated'
```
epicfaace:
```suggestion
    _NOT_AUTHENTICATED_ERROR = 'User is not authenticated.'
```

codalab/server/authenticated_plugin.py
```diff
@@ -1,52 +1,87 @@
...
+    _NOT_VERIFIED_ERROR = 'User is not verified'
```
epicfaace:
```suggestion
    _NOT_VERIFIED_ERROR = 'User is not verified.'
```

codalab/lib/bundle_cli.py
```diff
@@ -3725,12 +3725,25 @@ def do_chown_command(self, args):
             Commands.Argument(
                 '-d', '--disk-quota', help='Total amount of disk allowed (e.g., 3, 3k, 3m, 3g, 3t)'
             ),
+            Commands.Argument(
```
epicfaace, line 3728: Can you add a page / guide in the documentation about:
- what protected mode is
- how to set up a server with protected mode
- what commands on the CLI need to be run to grant / remove access

codalab/lib/bundle_cli.py
```diff
@@ -3725,12 +3725,25 @@ def do_chown_command(self, args):
...
+                help='Remove user\'s access when the CodaLab instance is in protected mode',
+            ),
         ),
     )
     def do_uedit_command(self, args):
         """
         Edit properties of users.
         """
+        if args.grant_access and args.remove_access:
+            raise UsageError('Can\'t grant and remove access for a user.')
```
```suggestion
            raise UsageError('Can\'t both grant and remove access for a user.')
```

codalab/rest/users.py
```diff
@@ -157,13 +163,25 @@ def update_users():
...
+    users = AdminUserSchema(strict=True, many=True).load(request.json, partial=True).data
 
     if len(users) != 1:
-        abort(http.client.BAD_REQUEST, "Users can only be updated on at a time.")
+        abort(http.client.BAD_REQUEST, "Users can only be updated one at a time.")
+
+    if 'has_access' in users[0]:
+        # has_access can only be updated in protected mode
+        if os.environ.get('CODALAB_PROTECTED_MODE') != 'True':
+            abort(http.client.BAD_REQUEST, "This CodaLab instance is not in protected mode")
```
```suggestion
            abort(http.client.BAD_REQUEST, "This CodaLab instance is not in protected mode.")
```

codalab/rest/users.py
```diff
@@ -157,13 +163,25 @@ def update_users():
...
+
+        # Only verified users can be given access to a protected instance
+        if not local.model.is_verified(users[0]['user_id']):
+            abort(
+                http.client.BAD_REQUEST,
+                "User has to be verified first in order to be granted access",
```
```suggestion
                "User has to be verified first in order to be granted access.",
```

codalab/server/authenticated_plugin.py
```diff
@@ -59,16 +94,44 @@ class AuthenticatedPlugin(object):
...
+            if self.is_protected_mode():
+                self.check_has_access()
+            return callback(*args, **kwargs)
+
+        return wrapper
 
+
+class ProtectedPlugin(AuthenticationPlugin):
+    """
+    Checks if the user is authenticated and has access to the instance only if the instance is in protected mode.
```
epicfaace: It was a bit hard to immediately understand what the classes were doing by just reading the comments.

Could we structure all our comments for these classes like this:

```
In non-protected mode, grants anonymous access.

In protected mode, checks if the user is authenticated and has access to the instance.
```

codalab/rest/bundles.py
```diff
@@ -35,14 +35,14 @@
...
-from codalab.server.authenticated_plugin import AuthenticatedPlugin
+from codalab.server.authenticated_plugin import AuthenticatedProtectedPlugin, ProtectedPlugin
 from codalab.worker.bundle_state import State
 from codalab.worker.download_util import BundleTarget
 
 logger = logging.getLogger(__name__)
 
 
-@get('/bundles/<uuid:re:%s>' % spec_util.UUID_STR)
+@get('/bundles/<uuid:re:%s>' % spec_util.UUID_STR, apply=ProtectedPlugin())
```
epicfaace, line 45: What do you think if we remodel this around a "mixin"-type approach? This would mean that we could create the following classes that can then be combined (`apply` takes an array):

```
AuthenticatedPlugin # checks to see if users are authenticated (regardless of protected mode)
ProtectedPlugin # if protected mode is set, checks to see if users have been granted access
```

This means that our current classes would map to the following:

`apply=ProtectedPlugin()` -> `apply=ProtectedPlugin()`
`apply=AuthenticatedPlugin()` -> `apply=AuthenticatedPlugin()`
`apply=AuthenticatedProtectedPlugin()` -> `apply=[AuthenticatedPlugin(), ProtectedPlugin()]`

Really, this change would just involve removing `AuthenticatedProtectedPlugin()`, but I think it makes it easier to see what is going on and lets you compose more plugins in the future.

.github/workflows/test.yml
```diff
@@ -176,6 +176,53 @@ jobs:
...
           path: /tmp/logs
 
+  test_backend_protected_mode:
+    name: Test backend - protected mode
+    runs-on: ubuntu-latest
+    needs: [build]
+    strategy:
+      matrix:
+        test:
+          - basic status batch anonymous competition unicode rest1 upload1 download refs binary rm make worksheet_search worksheet_tags freeze detach perm search_time groups
```
epicfaace, line 187: Can we just run `protected_mode` and maybe one other test here? Do we really need to run all tests?

codalab/server/authenticated_plugin.py
teetone: I think we still need this to give users access to manage their accounts and view account info.

.github/workflows/test.yml
teetone: All these tests seem to tests different functionalities. I removed a few, but I would like to keep the rest for now for peace of mind.

codalab/lib/bundle_cli.py
teetone: I added a section to `CLI-Reference.md`. Can you take a look at it?

codalab/rest/bundles.py
teetone: I think that's a good idea too, but I don't foresee adding another plugin in the near future. Would it be okay if I keep it as is for now and maybe we can try this approach if we need to add more plugins?

epicfaace (approved):

PR #2606, opened 2020-07-15
Run stress tests on CI

epicfaace:

The stress tests can be triggered manually with `workflow_dispatch` (see https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).

Also, clean up stress test files from other failed runs.


.github/workflows/stress.yml
```diff
@@ -0,0 +1,21 @@
+name: Stress Test
+on:
+  workflow_dispatch:
+  push:
```
epicfaace: Keeping the `push:` key in for testing purposes -- will remove this right before merging.

The reason we need this is that `workflow_dispatch` events can only be triggered manually from the `master` branch.

epicfaace:
@teetone let's just get a non-heavy version in to start with? We can always iterate on it later.

stress_test.py
```diff
@@ -287,13 +289,13 @@ def main():
...
-        args.create_worksheet_count = 1000
+        args.bundle_upload_count = 500
+        args.create_worksheet_count = 500
         args.parallel_runs_count = 500
-        args.large_docker_runs_count = 2000
-        args.test_infinite_memory = True
-        args.test_infinite_gpu = True
-        args.infinite_gpu_runs_count = 100
+        args.large_docker_runs_count = 200
+        args.test_infinite_memory = False
```
teetone, line 305: @epicfaace I updated the numbers and added `--heavy`. I will bring back the infinite runs in the near future.

teetone:
@epicfaace  I also added tests that create runs that request a gpu.

teetone (approved):

Issue #2619, opened 2020-07-16
Clean up canEdit and editMode variable names and functions 


yipenghe:
These two names are kind of confusing in Worksheet.js, they both represent variable names and function names and it's not clear that editMode means opening the source 

- [x] cleanup canEdit
- [x] cleanup editMode

PR #2620, opened 2020-07-16
remove canEdit in worksheet.js

yipenghe:

canEdit had two reference in Worksheet.js, including a variable name and a function name.
The variable is actually not used anywhere, and the function collides with the editPermission variable in rendering.
There is also a canEdit field in editableFieldBase which is used for text.
The PR removes the canEdit variable in Worksheet.JS and renames canEdit function to hasEditPermission to remove redundancy and make it clearer what the function is doing.
The canEdit field in now only kept in editableFieldBase

### Related issues

part of #2619

### DIFF
--- a/frontend/src/components/FileBrowser/FileBrowser.js
+++ b/frontend/src/components/FileBrowser/FileBrowser.js
@@ -18,7 +18,6 @@ export class FileBrowser extends React.Component<
         buildPayload: (string) => {},
         method: string,
         url: string,
-        canEdit?: boolean,
         onChange?: () => void,
     },
     {
@@ -34,7 +33,6 @@ export class FileBrowser extends React.Component<
     /** Prop default values. */
     static defaultProps = {
         method: 'POST',
-        canEdit: true,
     };
 
     constructor(props) {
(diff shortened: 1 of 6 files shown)


frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -34,7 +33,6 @@ export class FileBrowser extends React.Component<
     /** Prop default values. */
     static defaultProps = {
         method: 'POST',
-        canEdit: true,
```
yipenghe, line 37: unused in this component

frontend/src/components/worksheets/items/WorksheetItem.js
```diff
@@ -77,7 +77,6 @@ class WorksheetItem extends React.Component {
 
         var self = this;
         var tableClassName = this.props.focused ? 'table focused' : 'table';
-        var canEdit = this.props.canEdit;
```
yipenghe, line 80: No item actually uses this props field

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -155,7 +154,6 @@ class TableItem extends React.Component<{
                     bundleInfo={bundleInfo}
                     uuid={bundleInfo.uuid}
                     headerItems={headerItems}
-                    canEdit={canEdit}
```
yipenghe, line 158: bundleRow  does not use canEdit

epicfaace (approved):

PR #2636, opened 2020-07-20
Update REST API Reference

nelson-liu:
(this change came up when running ./pre-commit.sh )


nelson-liu:
hm i wonder what's going on here...

epicfaace (approved):
Strange, but let's just merge anyway and file as a bug. It appears to be "correct" (at least predictable) -- the same thing happened in #2647.

PR #2649, opened 2020-07-23
Revert "Run nvidia-smi directly on worker host to respect GPU isolation"

epicfaace:
Reverts codalab/codalab-worksheets#2415

This is causing the GPUs to all show up as "0/0" on prod. This is because the `codalab_worker_1` docker image doesn't have `nvidia-smi` installed, so the command fails.

### DIFF
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -10,7 +10,6 @@ import logging
 import signal
 import socket
 import stat
-import subprocess
 import sys
 import psutil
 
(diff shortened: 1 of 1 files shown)


yipenghe:
The PR was merged around July 1st, how come the failure happens now though

yipenghe (approved):

yipenghe:
Ok it seems like none of the jobs that requested GPUs actually ran since the commit went to prod...

yipenghe:
Maybe we should go for the installing nvidia-smi solution

nelson-liu:
alternatively: we can try to run nvidia-smi on the host. if it works, great. if not, try the old way (download the cuda image, run nvidia-smi).

nelson-liu:
> Maybe we should go for the installing nvidia-smi solution

yeah, i'd be in favor of this too. Should be straightforward—instead of doing FROM ubuntu16.04, use a cuda container on ubuntu 16.04

epicfaace:
Closing this in favor of #2650.

epicfaace:
Actually, let's merge this. #2650 didn't fix the issue (we actually have to run the entire worker Docker container under the nvidia runtime -- `--worker=nvidia`), so this merits a deeper discussion on how to better fix this.

PR #2670, opened 2020-07-30
Add editing table schema functionality

### LINKED ISSUE #2436: Allow to style table with schema

yipenghe:
resolve #2436 

### Screenshots
Added the same icon as schema item to the header row
Screenshot from 2020-07-29 18-25-00

Clicking on it will show the following:
Screenshot from 2020-07-29 18-25-56

I added the backend support for both table and record mode, but only finished the frontend part for table. The record part needs some more design


codalab/lib/worksheet_util.py
```diff
@@ -678,10 +678,11 @@ def genpath_to_target(bundle_info, genpath):
         # strip off the leading / from genpath to create a subpath in the target.
         return (bundle_info['uuid'], genpath[1:])
 
-    def flush_bundles():
+    def flush_bundles(bundle_block_start_index):
```
percyliang, line 684: Maybe more descriptive to call it `display_mode_start_index`

percyliang (commented):
Could we make the textbox for editing the schema name thinner?

Actually, what about just having a dropdown box to select the schema?  Since they shouldn't be specifying a schema that's not defined anyway.

yipenghe:
> Could we make the textbox for editing the schema name thinner?
> 
> Actually, what about just having a dropdown box to select the schema? Since they shouldn't be specifying a schema that's not defined anyway.

That requires remembering what schemas are before each tables. It also won't be easy for a user to switch the orders of the schema unless we add left and right shift buttons for each dropdown box for that usage. Also textbox makes deleting multiple unwanted schemas quite easy. I guess if we really want dropdown instead, the best way would be to improve it in latter iterations and get this textbox version out for usage.

percyliang (approved):

codalab/lib/worksheet_util.py
yipenghe: It points to the first []{0x123} for bundle blocks even if there is a % display above since default schema doesn't require such line

PR #2672, opened 2020-07-30
2360: Support cl run --interactive

### LINKED ISSUE #2360: support `cl run --interactive`
Currently, you have to create a run bundle, and then either wait an indeterminate amount of time for the job to run or spin up your own worker.

Another problem is that you can’t just poke around in the shell interactively, attach a debugger, etc.

With this change, you will be able to run

```cl run … --interactive```

Assuming shared file system and `--link`, this will create a run bundle immediately in an `interacting` state, and launch a Docker container with the appropriate dependencies mounted, runs the command that you want and then drops you into a shell (in the Docker container) where you can continue to poke around.  (Note this is a different flow because usually the worker launches Docker; the client generally does not.  So we might have to refactor some things so that the client can also run docker.  There was a local mode that was created earlier which just launched docker but didn't create a bundle, but that was deprecated.)

Bonus: you can press ‘up’ to access the last command.

When you exit the interactive session, an editor pops up with the list of all the commands that were run in the interactive session.  User can delete any extraneous commands and do any editing (e.g., remove errors or diagnostic commands like ls).  When the editing is done, the official command for the bundle that actually gets created.

If one performs `cl mimic` on this bundle, then the bundle will run for real on the worker, and this will verify that the quick and dirty interactive construction was legitimate.

### PR DESCRIPTION
**Pull request by teetone**
### cl run --interactive Workflow

1. The user starts an interactive session by running `cl run --interactive` with the desired parameters and dependencies.
2. Create a Docker container with the specified image on the client machine for the interactive session.
3. Mount dependencies as read-only.
4. Set working directory to some arbitrary path: `/<session uuid>`
5. The user will interact with this container and try out different commands. Once satisfied the user will exit the bash session. All the commands the user tried out will be stored at path `/root/.bash_history` in the container.
6. Copy `.bash_history` out to the host machine.
7. Open an editor and allow the user to select and edit commands for the official run.
8. Stop and remove the interactive session container.
9. Create a run using the final command. From this point, it’s just the same workflow as running `cl run`.

### Related issues

Resolves #2360 

### Other changes

I also removed `cl docker` and the documentation for that command.

### Checklist

### CODE REVIEW

File: codalab/lib/interactive_session.py [hunk 1]
```diff
@@ -0,0 +1,232 @@
...
+           in the container.
+        6. Copy `.bash_history` out to the host machine.
+        7. Open an editor and allow the user to select and edit commands for the official run.
+        8. Return the final command.
+        9. Stop and remove the interactive session container.
+    """
+
+    _BASH_HISTORY_CONTAINER_PATH = '/root/.bash_history'
+    _CHOOSE_COMMAND_INSTRUCTIONS = (
+        "\n\n@@\n"
```
percyliang: I think it's more natural to use `# `, which corresponds to shell comments.

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,232 @@
...
+        os.system('vi %s' % path)
+
+        # Extract out the final commands minus the instructions
+        commands = []
+        for line in open(path).read().splitlines():
+            command = line.lstrip().rstrip()
+            if command and not command.startswith(InteractiveSession._INSTRUCTIONS_DELIMITER):
+                commands.append(command)
+
+        final_command = '&&'.join(commands)
```
percyliang: You can actually have multiline using `\n`; might be better to be consistent with what happens in the frontend when you have a multiline command.

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # Write out the commands to choose from and the instructions out to a file
+        path = os.path.join(self._bundle_path, 'edit_commands.txt')
+        with open(path, 'w') as f:
+            for command in candidate_commands:
+                f.write(command)
+            f.write(InteractiveSession._CHOOSE_COMMAND_INSTRUCTIONS)
+
+        # Use vi to allow users to choose commands
+        os.system('vi %s' % path)
```
percyliang: use `editor_util.py`

codalab/lib/bundle_cli.py
```diff
@@ -1621,10 +1622,32 @@ def do_run_command(self, args):
...
+            bundle_uuids = [bundle_target.bundle_uuid for _, bundle_target in targets]
+            bundles_locations = client.get_bundles_locations(bundle_uuids)
+
+            docker_image = metadata.get('request_docker_image', None)
+            if not docker_image:
+                # If a Docker image is not specified, use the default CPU worker image for the interactive session
+                docker_image = self.manager.config['workers']['default_cpu_image']
+
+            # Start an interactive session to allow users to figure out the command to run
+            session = InteractiveSession(docker_image, self.manager, targets, bundles_locations)
```
percyliang: Could you pass the `args.command` into the InteractiveSession and put it in `.bash_history` so the command is available to the user?

percyliang: If `--interactive` is set, then can we make the command optional?  It's a bit awkward to have to type `""`.

codalab/lib/bundle_cli.py
```diff
@@ -1612,6 +1608,11 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
                 help='If a bundle with the same command and dependencies already exists, return it instead of creating a new one.',
                 action='store_true',
             ),
+            Commands.Argument(
+                '--interactive',
```
percyliang, line 1631: Add `-i` short flag

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,232 @@
...
+        self._stderr = stderr
+
+    def start(self):
+        self._bundle_path = os.path.join(
+            self._manager.codalab_home, 'local_bundles', self._session_uuid
+        )
+        os.makedirs(self._bundle_path)
+
+        run_command = self.get_docker_run_command()
+        print('\nStarting an interactive session...', file=self._stdout)
```
percyliang: Could we put all of this under a `--verbose` flag, which is off by default?

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,232 @@
...
+        print('%s\n' % run_command, file=self._stdout)
+        print('=' * 150, file=self._stdout)
+        print('Session UUID: ', self._session_uuid, file=self._stdout)
```
percyliang: Get rid of extra space after `:` for all these print statements.

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,232 @@
...
+
+        self._container = self._start_session(run_command)
+        return self._construct_final_command()
+
+    def get_docker_run_command(self):
+        def get_docker_path(sub_path):
+            return os.path.sep + os.path.join(self._session_uuid, sub_path)
+
+        # Use a dict to keep track of volumes to mount. The key is the path on Docker and the value is the local path.
+        volumes = {}
```
percyliang, line 119: Put a comment about synchronizing this with the worker code that does this.

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,232 @@
...
+
+        final_command = '&&'.join(commands)
+        print('\nFinal constructed command:\n%s' % final_command, file=self._stdout)
+        return final_command
+
+    def _get_bash_history(self):
+        # Copies out .bash_history from the container to bundle_path
+        path = os.path.join(self._bundle_path, '.bash_history')
+        f = open(path, 'wb')
+        stream, _ = self._container.get_archive(InteractiveSession._BASH_HISTORY_CONTAINER_PATH)
```
percyliang, line 216: Can we use `copy` from `file_util.py`?

percyliang (dismissed):
Cool!  Just a few small-ish things.  We should get this in and iterate.

But we should think about
1) Refactoring to reduce redundancy with the worker.
2) Allowing things to work when the client does not share a file system with the master.  In that case, we have to use a download manager to download the bundles locally and maintain a bundle cache.  In this case, the client looks more like a worker.  This is also useful when you've run something in CodaLab and you want to rerun/tweak it and it fails and we want to debug.

codalab/lib/bundle_cli.py
teetone: I actually couldn't find a way around this. Since 0 or more targets can be passed in and we have `target_spec` as the positional argument before `command`, not passing in a command will wrongly treat the last target as the command.

codalab/lib/interactive_session.py
teetone: `copy` doesn't seem to work with the data stream that is being returned by `get_archive`.

codalab/lib/bundle_cli.py
```diff
@@ -1612,6 +1608,12 @@ def derive_bundle(self, bundle_type, command, targets, metadata):
                 help='If a bundle with the same command and dependencies already exists, return it instead of creating a new one.',
                 action='store_true',
             ),
+            Commands.Argument(
+                '-i',
+                '--interactive',
+                help='Beta feature - Start an interactive session to construct your run command.',
```
epicfaace, line 1632:
```suggestion
                help='Start an interactive session to construct your run command.',
```

I think we should remove this. We have "beta feature" in a few other commands, which are really already released and supported by us.

epicfaace (commented):
- Can you add a documentation page with sample usage? This is especially important because we have no direct CLI tests (so it's harder to infer CLI usage from just looking at existing tests).
- It seems strange that I can't just run `cl run --interactive` but have to include my own command, `cl run --interactive echo`. I think the former would also be intuitive if I want to use a shell to run all my commands.

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,258 @@
...
+    session = InteractiveSession(
+        args.docker_image, manager=CodaLabManager(), dependencies=[], bundle_locations={}
+    )
+    session.start()
+    session.cleanup()
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Runs the specified CodaLab stress tests against the specified CodaLab instance (defaults to localhost).'
```
epicfaace: This should be changed

codalab/lib/interactive_session.py
```diff
@@ -0,0 +1,258 @@
...
+def main():
+    # Example usage of InteractiveSession
+    session = InteractiveSession(
+        args.docker_image, manager=CodaLabManager(), dependencies=[], bundle_locations={}
+    )
+    session.start()
+    session.cleanup()
+
+
+if __name__ == '__main__':
```
epicfaace: Is this even used (calling this file directly)? We should probably document its usage somewhere or remove it.

codalab/rest/bundles.py
```diff
@@ -377,6 +378,21 @@ def _set_bundle_permissions():
...
+def _fetch_locations():
+    """
+    Fetch locations of bundles.
+
+    Query parameters:
+    - `uuids`: List of bundle UUID's to get the locations for
+    """
+    bundle_uuids = query_get_list('uuids')
+    uuids_to_locations = {
+        uuid: local.bundle_store.get_bundle_location(uuid) for uuid in bundle_uuids
```
epicfaace: Once the link PR gets in, this will need to be changed to something like 

```python
bundle_link_urls = local.model.get_bundle_metadata(bundle_uuids, "link_url")
uuids_to_locations = {
        uuid: bundle_link_urls.get("uuid") or local.bundle_store.get_bundle_location(uuid) for uuid in bundle_uuids
}
```

epicfaace (commented):
Instead of using `os.system` (which can have security issues because input isn't sanitized), can we instead pass an array of args to `subprocess.check_output` directly? In some cases, you're already constructing this array, only to join it into a string later. See https://security.openstack.org/guidelines/dg_use-subprocess-securely.html

codalab/lib/bundle_cli.py
teetone: Calling the other features "beta" might be incorrect, but this is really still a beta feature.

codalab/lib/interactive_session.py
teetone: I got rid of the whole main function. I left it there for testing purposes, but it's no longer needed.

teetone:
@epicfaace I actually couldn't find a way around adding an empty command. See my comment here: https://github.com/codalab/codalab-worksheets/pull/2672#discussion_r463920276

teetone:
> Instead of using `os.system` (which can have security issues because input isn't sanitized), can we instead pass an array of args to `subprocess.check_output` directly? In some cases, you're already constructing this array, only to join it into a string later. See https://security.openstack.org/guidelines/dg_use-subprocess-securely.html

@epicfaace I'm not sure what the security vulnerability is here. There are two instances of `os.system` in `interactive_session.py` and I am running `os.system` on commands that we construct and have full control over. In the link you provided, it gives examples of creating functions that take any input. 

> * Can you add a documentation page with sample usage? This is especially important because we have no direct CLI tests (so it's harder to infer CLI usage from just looking at existing tests).
> * It seems strange that I can't just run `cl run --interactive` but have to include my own command, `cl run --interactive echo`. I think the former would also be intuitive if I want to use a shell to run all my commands.

I actually couldn't find a way around adding an empty command. See my comment here: #2672 (comment)

epicfaace:
> There are two instances of `os.system` in `interactive_session.py` and I am running `os.system` on commands that we construct and have full control over. In the link you provided, it gives examples of creating functions that take any input.

That's correct, though [the link](https://security.openstack.org/guidelines/dg_use-subprocess-securely.html) says:

> - Even if _your_ code is not vulnerable, the next person who maintains can easily introduce a vulnerability.

It's therefore best practice to avoid os.system unless there's absolutely no way not to.

epicfaace (approved):
Looks good. I'd prefer if we can remove usage of `os.system`, but if we want to get this merged quickly, it's also fine to just merge now and handle that later (since we have a couple of other uses of `os.system` in the codebase anyway)

teetone:
> Looks good. I'd prefer if we can remove usage of `os.system`, but if we want to get this merged quickly, it's also fine to just merge now and handle that later (since we have a couple of other uses of `os.system` in the codebase anyway)

Okay, we can sync up on `os.system`. I will merge this PR in for now.

Issue #2080, opened 2020-03-11
add type annotations to Python code


percyliang:
Start with annotating the types of the most common functions (important interfaces like `bundle_model.py`).

In some places, we use tuples to represent worksheet items and worksheet_info is a complex dictionary.  We should consider using `TypedDict` for that.  These are probably places where annotations would help a lot.

candicegjing:
These are returned when running `mypy` to type check `bundle_model.py`:
```
stanford@Stanfords-MacBook-Pro codalab-worksheets % mypy codalab/model/bundle_model.py 
codalab/rest/worksheet_block_schemas.py:6: error: Skipping analyzing 'bottle': found module but no type hints or library stubs
codalab/rest/worksheet_block_schemas.py:7: error: Skipping analyzing 'marshmallow': found module but no type hints or library stubs
codalab/rest/worksheet_block_schemas.py:8: error: Skipping analyzing 'marshmallow_jsonapi': found module but no type hints or library stubs
codalab/worker/bundle_state.py:65: error: Name 'Dict' is not defined
codalab/worker/bundle_state.py:65: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Dict")
codalab/worker/bundle_state.py:65: error: Name 'Any' is not defined
codalab/worker/bundle_state.py:65: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Any")
codalab/worker/bundle_state.py:66: error: Name 'List' is not defined
codalab/worker/bundle_state.py:66: note: Did you forget to import it from "typing"? (Suggestion: "from typing import List")
codalab/worker/bundle_state.py:66: error: Name 'Dict' is not defined
codalab/worker/bundle_state.py:66: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Dict")
codalab/worker/bundle_state.py:67: error: Name 'Any' is not defined
codalab/worker/bundle_state.py:67: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Any")
codalab/worker/bundle_state.py:68: error: Name 'Optional' is not defined
codalab/worker/bundle_state.py:68: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Optional")
codalab/worker/bundle_state.py:79: error: Name 'Dict' is not defined
codalab/worker/bundle_state.py:79: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Dict")
[...]

PR #2674, opened 2020-07-30
Resolve found module but no type hints or library stubs

yiboliu:

This commit resolves all issues annotated as 'found module but no type hints or library stubs' by adding `# type: ignore`. Ref: https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
### Related issues
https://github.com/codalab/codalab-worksheets/issues/2080

### Checklist


epicfaace (commented):
Could we instead use the approach in which we specify ignored libraries in a mypy config file and use the `ignore_missing_imports` option (see https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-type-hints-for-third-party-library)?

I think that will be easier for us to keep track of to keep all these rules in a single file (such as eslint ignored rules) rather than having them set in comments whenever the libraries are imported.

yiboliu:
> Could we instead use the approach in which we specify ignored libraries in a mypy config file and use the `ignore_missing_imports` option (see https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-type-hints-for-third-party-library)?
> 
> I think that will be easier for us to keep track of to keep all these rules in a single file (such as eslint ignored rules) rather than having them set in comments whenever the libraries are imported.

To add to a config file, do we have a default one? Or I can create one myself? https://mypy.readthedocs.io/en/latest/config_file.html#config-file

percyliang:
Could you add a call in the `pre-commit.sh` script?  Are there no warnings now?  If so, we should also add it to CI.

yiboliu:
> Could you add a call in the `pre-commit.sh` script? Are there no warnings now? If so, we should also add it to CI.

No warnings about `found module but no type hints or library stubs`. I'll fix other issues in following commits.

pre-commit.sh
```diff
@@ -19,3 +19,6 @@ venv/bin/mkdocs build || { rm -rf venv; exit 1; }  # Outputs to `site`
 # Fix Python and JavaScript style (mutates code!)
 venv/bin/black codalab scripts *.py || exit
 npm run --prefix frontend format
+
+#Check about mypy warnings
```
percyliang: nit: space between '#' and 'Check'

pre-commit.sh
```diff
@@ -19,3 +19,6 @@ venv/bin/mkdocs build || { rm -rf venv; exit 1; }  # Outputs to `site`
 # Fix Python and JavaScript style (mutates code!)
 venv/bin/black codalab scripts *.py || exit
 npm run --prefix frontend format
+
+# Check about mypy warnings
+mypy codalab/model/bundle_model.py
```
percyliang: Make sure lines end with newline

teetone:
Will the other classes mentioned in the issue https://github.com/codalab/codalab-worksheets/issues/2080 be handled in another pull request?

teetone (approved):

yiboliu:
> Will the other classes mentioned in the issue #2080 be handled in another pull request?

Actually I used `mypy .` to identify problems, so all classes are resolved.

PR #2677, opened 2020-08-03
Add type annotations with mypy

### LINKED ISSUE #2080: add type annotations to Python code

yiboliu:
1. Resolve Exception must be derived from BaseException
2. Resolve Need type annotation for '*'
3. Resolve Cannot find implementation or library stub for module named '*'
4. Add the requirements.dev.txt into installation list
5. Further comments will be added if there are more to commit here

### Related issues
Resolves #2080 

### Checklist


codalab/worker/rest_client.py
```diff
@@ -27,7 +28,7 @@ class RestClient(object):
     """
     A dictionary of additional headers to send along with HTTP requests.
     """
-    _extra_headers = {}
+    _extra_headers: Dict[Any, Any] = {}
```
percyliang: Can the types be more specific?  At least the key is a string, no?

requirements-server.txt
```diff
@@ -8,7 +8,7 @@
 -r requirements.dev.txt
 
 # For worker manager
-boto3==1.9.228
+boto3~=1.14.33
```
percyliang: Are these part of this PR?

percyliang (commented):
Can you write a PR description saying more about what the scope of this PR is?  Why these particular additions and these files?  It doesn't seem like it's complete, so does this correspond to some conceptual subset of work?

requirements-server.txt
yiboliu: Yes, it's upgrading a boto3 version. It's suggested by pycharm.

codalab/worker/rest_client.py
yiboliu: sure, I can do it.

requirements-server.txt
```diff
@@ -17,5 +17,5 @@ nose==1.3.7
 # Server-specific
 alembic==1.0.0
 gunicorn==20.0.4
-oauthlib==2.1.0
+oauthlib~=3.1.0
```
teetone: I think this PR should just be about mypy related changes. Also,  we tried to upgrade `oauthlib` to `v3.1.0` and it broke our authentication. See this recent commit here: https://github.com/codalab/codalab-worksheets/commit/1f2c0ea6fadc0eaf5e86bc3d5b407c146bf348d4

requirements.txt
```diff
@@ -13,3 +13,5 @@ fusepy==2.0.4
 python-dateutil==2.8.1
 diffimg==0.2.3
 selenium==3.141.0
+
+requests~=2.24.0
```
teetone: Can you explain what this library is for?

codalab/lib/bundle_fuse.py
```diff
@@ -8,6 +8,8 @@
 from collections import OrderedDict
 
 from contextlib import closing
+from typing import Dict, Any
```
teetone: Is `Any` being used in this file?

test_cli.py
```diff
@@ -386,7 +388,7 @@ class TestModule(object):
     a decorator to register new modules and a class method to run modules by name.
     """
 
-    modules = OrderedDict()
+    modules: Dict[str, Any] = OrderedDict()
```
teetone: We should try to avoid `Any` as much as we can. I think `Any` is the same as not specifying a type. Should it be `TestModule` instead?

requirements.txt
yiboliu: There are a few files importing this library, like codalab/rest/interpret.py, codalab/worker/docker_utils.py, etc

requirements-server.txt
yiboliu: yeah, I can roll it back.

codalab/lib/bundle_fuse.py
yiboliu: I removed it and usages.

test_cli.py
yiboliu: Are we using python 3.7 or newer? If not we can't refer to the class within itself. I worked out another way.

percyliang: You can use a string: `"TestModule"`

yiboliu: Yeah, I did it in my latest update.

requirements-server.txt
epicfaace: Can you remove the `~=` in all dependencies? We want to pin dependencies to exact versions using `==` instead.

Also, are these dependency upgrades required for the mypy checks to work? If not, we should do them in a separate PR (dependabot might have already caught a few of them).

epicfaace (changes requested):

teetone (approved):

epicfaace (approved):

PR #2706, opened 2020-08-05
Flexible tag search query for worker manager

bkgoksel:

#2564 broke the aws batch worker manager use case at SM because we use tag exclusive workers, and submit runs with `request_queue=tag=<tag-name>` (as that was the only supported usage pattern by the most recent codalab version at the time we set the infrastructure). After #2564  jobs submitted with `--request-queue=tag=<tag>` will not be found by worker managers and will get stuck in staged state forever. 

We might want to move to submitting our jobs with `--request-queue=<tag>` but it'll take time (for all the different processes that submit jobs to pick up that change) and for the time being this is the only way we can resume using codalab (short of forking again)

### Checklist

* [X] I've added and/or updated tests, if this is a backend change
* [X] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script


bkgoksel:
I have no idea why running `./pre-commit.sh` modified these other files 🤷 ? happy to revert changes to those before merging.

yipenghe:
> I have no idea why running `./pre-commit.sh` modified these other files 🤷 ? happy to revert changes to those before merging.

Someone forgot to run it for their PR, it's good

teetone (approved):

bkgoksel:
I'm not sure who all would need to see this in codalab team so not gonna merge it right away but please feel free to do so if it's ok to merge @teetone and thanks for reviewing!

codalab/worker_manager/worker_manager.py
```diff
@@ -113,7 +113,7 @@ def run_one_iteration(self):
         if os.environ.get('CODALAB_USERNAME') != "codalab":
             keywords += [".mine"]
         if self.args.worker_tag_exclusive and self.args.worker_tag:
-            keywords += ["request_queue=" + self.args.worker_tag]
+            keywords += ["request_queue=%s,tag=%s" % (self.args.worker_tag, self.args.worker_tag)]
```
percyliang, line 121: Could you leave a comment here for why we need both and also say that in the future, we just need `request_queue`?

bkgoksel: sure. also worth noting that there's a corresponding piece of code in bundle manager that treats `tag=` as an optional but matching prefix there. it's probably best to keep these two consistent ie. if this is removed then so should that bit.

bkgoksel: done

percyliang: Just realized that this is matching <worker_tag> or tag=<worker_tag>, which is the natural way to read this expression and makes more sense.  Could you add a quick comment?

percyliang (approved):

codalab/worker_manager/worker_manager.py
bkgoksel: oh, sure

PR #2707, opened 2020-08-05
Add requests to requirements.txt

yiboliu:
Add a required package requests

### Checklist


epicfaace (approved):
Thanks!

PR #2710, opened 2020-08-06
Make sure run container names are random

### LINKED ISSUE #2111: Prod issue - Worker failing with conflict in container name
```Traceback (most recent call last):
  File "/opt/codalab/worker/docker_utils.py", line 33, in wrapper
    return f(*args, **kwargs)
  File "/opt/codalab/worker/docker_utils.py", line 165, in start_bundle_container
    stdin_open=tty,
  File "/usr/local/lib/python3.6/dist-packages/docker/models/containers.py", line 785, in run
    detach=detach, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/docker/models/containers.py", line 843, in create
    resp = self.client.api.create_container(**create_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/docker/api/container.py", line 427, in create_container
    return self.create_container_from_config(config, name)
  File "/usr/local/lib/python3.6/dist-packages/docker/api/container.py", line 438, in create_container_from_config
    return self._result(res, True)
  File "/usr/local/lib/python3.6/dist-packages/docker/api/client.py", line 262, in _result
    self._raise_for_status(response)
  File "/usr/local/lib/python3.6/dist-packages/docker/api/client.py", line 258, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/usr/local/lib/python3.6/dist-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.APIError: 409 Client Error: Conflict ("Conflict. The container name "/codalab_run_0x05179aa8558d4ba0b3377e6d358b9602" is already in use by container "995fe505626bbd1eff4f30f0a5d2141fbad585ea4934c93b5a664a3d21d577b0". You have to remove (or rename) that container to be able to reuse that name.")

  File "/opt/codalab/worker/worker_run_state.py", line 287, in _transition_from_PREPARING
    runtime=self.docker_runtime,
  File "/opt/codalab/worker/docker_utils.py", line 41, in wrapper
codalab.worker.docker_utils.DockerException: Unable to start Docker container: 409 Client Error: Conflict ("Conflict. The container name "/codalab_run_0x05179aa8558d4ba0b3377e6d358b9602" is already in use by container "995fe505626bbd1eff4f30f0a5d2141fbad585ea4934c93b5a664a3d21d577b0". You have to remove (or rename) that container to be able to reuse that name.")```

epicfaace:

Fixes #2111. That issue happens whenever there's a problem with starting the Docker container, it isn't removed, and a new container with the same name (for the same run) is attempted to start. This wasn't a problem until #1437 was merged, because before #1437 was merged, all run containers would have been assigned random names.

This PR keeps the UUID in the container name, but adds some randomness to the names so that names cannot clash.

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -13,6 +13,7 @@ from dateutil import parser, tz
 import datetime
 import re
 import requests
+import secrets
 
 
 MIN_API_VERSION = '1.17'
@@ -172,7 +173,7 @@ def start_bundle_container(
         environment['NVIDIA_VISIBLE_DEVICES'] = ','.join(gpuset) if gpuset else 'all'
 
     # Name the container with the UUID for readability
-    container_name = 'codalab_run_%s' % uuid
+    container_name = f"codalab_run_{uuid}_{secrets.token_hex(5)}"
     container = client.containers.run(
         image=docker_image,
         command=docker_command,


nelson-liu:
do you think this is treating the symptom, and not the disease? It feels like instead of working around this issue (by making names random), we should be figuring out why we're trying to launch containers for the same bundle / why containers are failing to start? These sort of issues seem hard to diagnose without this (blatant) error

epicfaace:
Doesn't #2111 actually cause the entire worker to go down though? In that case, it might be better to let it silently fail with this PR (and potentially just log if something is wrong).

I don't recall if that's the case, though -- if the error isn't that serious and is just a message, we can leave the behavior in as a useful warning.

nelson-liu:
> Doesn't #2111 actually cause the entire worker to go down though? In that case, it might be better to let it silently fail with this PR (and potentially just log if something is wrong).


yeah, it does. Well, the worker doesn't go down, it's just stuck in a loop of sleeps. The issue is that even if you re-run with a different name, the container likely won't start, right (since it didn't start the first time, so it seems likely that it won't start if you try to rerun it with a different name)?

nelson-liu:
Thinking about this again because I ran into a case where a Slurm-running worker would always land on the same node, and always fail because it can't start the container due to a duplicate name (the first error was due to a flaky GPU issue). Is it worth cleaning up the `Created` containers if we attempt to run, and then fail?

nelson-liu:
(opened a pr at https://github.com/codalab/codalab-worksheets/pull/2766)

Issue #2712, opened 2020-08-06
Remove usage of string refs


epicfaace:
String refs in React are deprecated / buggy and shouldn't be used.

PR #2713, opened 2020-08-06
Refactor scrolling on worksheets ui

### LINKED ISSUE #2685: worksheet page doesn't scroll automatically anymore when use keyboard shortcuts
On a long worksheet, when you scroll up and down using arrow keys or k/j, the worksheet used to follow along, but now it doesn't anymore.

epicfaace:
Refactor scrolling on worksheets ui.
- Rather than string refs, I directly get elements through their IDs. This is better than using string refs, which are deprecated and can be buggy.
- Don't use keepPosInView -- use [scrollIntoView()](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) instead, which is simpler and faster.

Fixes #2685, related to #2712.

### DIFF
--- a/frontend/src/components/worksheets/items/WorksheetItem.js
+++ b/frontend/src/components/worksheets/items/WorksheetItem.js
@@ -79,14 +79,15 @@ class WorksheetItem extends React.Component {
         var tableClassName = this.props.focused ? 'table focused' : 'table';
         var items = this._getItems();
 
+        const { focusIndex } = this.props;
+
         var body_rows_html = items.map(function(row_item, row_index) {
-            var row_ref = 'row' + row_index;
             var row_focused = self.props.focused && row_index === self.props.subFocusIndex;
             var url = '/worksheets/' + row_item.uuid;
             return (
                 <TableWorksheetRow
                     key={row_index}
-                    ref={row_ref}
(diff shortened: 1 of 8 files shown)


yipenghe:
Seems to work nice when we release the key after pressing, but if we hold the key down, the behaviour is weird

frontend/src/components/worksheets/items/WorksheetItem.js
```diff
@@ -140,7 +141,7 @@ class TableWorksheetRow extends React.Component {
         var item = this.props.item;
         var className = /*'type-worksheet' + */ this.props.focused ? ' focused' : '';
         return (
-            <tr className={className}>
+            <tr className={className} id={this.props.id}>
```
michaeldu1, line 144: can pull out id with other props since you do it already

michaeldu1 (approved):

michaeldu1:
I also cannot reproduce the behavior Yipeng mentioned. Tested on local branch and code lgtm!

frontend/src/components/worksheets/items/WorksheetItem.js
epicfaace: Not in the `render()` function, though. will just leave it as is

PR #2726, opened 2020-08-09
Resolve Invalid type: ClassVar nested inside other type

yiboliu:
Resolve the `error: Invalid type: ClassVar nested inside other type`
Resolve Decorated property not supported
Resolve argument incompabilities
Resolve mypy complaining functions defined earlier: since mypy is unhappy with functions defined with same signature and return type, I had to change function name. 
Resolve Incompatible types in assignment due to different assigned type
Resolve Incompatible types in assignment due to different type defined in base class
Left problem of ` error: Self argument missing for a non-static method (or an invalid type for self)`, which is still open in mypy discussion forum: https://github.com/python/mypy/issues/7778

### Related issues
#2080 

### Checklist

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -212,17 +212,17 @@ class CodaLabManager(object):
 
         return config
 
-    @property
+    @property  # type: ignore
     @cached
     def config_path(self):
         return os.getenv('CODALAB_CONFIG', os.path.join(self.codalab_home, 'config.json'))
 
-    @property
+    @property  # type: ignore
     @cached
     def state_path(self):
         return os.getenv('CODALAB_STATE', os.path.join(self.codalab_home, 'state.json'))
 
-    @property
+    @property  # type: ignore
     @cached
     def codalab_home(self):
         from codalab.lib import path_util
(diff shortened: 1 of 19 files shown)


codalab/server/worker_info_accessor.py
```diff
@@ -1,13 +1,20 @@
...
 
 class WorkerInfoAccessor(object):
     """
     Helps with accessing the list of workers returned by the worker model.
     """
 
-    def refresh_cache(f):
+    def refresh_cache(
+        f: Union[
+            Callable[['WorkerInfoAccessor'], Any],
```
teetone: Can we do something about `Any` here?

codalab/lib/codalab_manager.py
```diff
@@ -212,17 +212,17 @@ def init_config(self, dry_run=False):
 
         return config
 
-    @property
+    @property  # type: ignore
```
teetone, line 215: Why is `ignore` needed here?

yiboliu: this is the suggested way to resolve warnings of Decorator type not supported

codalab/server/worker_info_accessor.py
yiboliu: Sometimes it returns a list, sometimes a boolean and sometimes even nothing. I feel like making it as Any fits the purpose of this decorator better, because it's used to pre-process parameters. I can change the parameters to str, as it's the only type showed up.

yiboliu: I can change the return type to what we have for now to see if it passes tests.

teetone (approved):

PR #2732, opened 2020-08-09
cleanup pre-commit.sh

percyliang:

On a fresh checkout of master, `./pre-commit.sh` was failing.  I removed the `--check` flag, and it works now.

EDIT: added `--check` back in (turns out I was using the wrong npm version).  So this PR basically just cleans up the `pre-commit.sh` script.

### DIFF
--- a/codalab/lib/spec_util.py
+++ b/codalab/lib/spec_util.py
@@ -11,7 +11,9 @@ UUID_REGEX = re.compile('^' + UUID_STR + '$')
 UUID_PREFIX_REGEX = re.compile('^0x[0-9a-f]{1,31}$')
 BEGIN_NAME_STR = '[a-zA-Z_]'
 NAME_STR = '[a-zA-Z_][a-zA-Z0-9_\.\-]*'
-NAME_PATTERN_STR = '[%\*a-zA-Z0-9_\.\-]+'  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
+NAME_PATTERN_STR = (
+    '[%\*a-zA-Z0-9_\.\-]+'
+)  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
 NAME_REGEX = re.compile('^' + NAME_STR + '$')  # Names (exact match)
 NAME_PATTERN_REGEX = re.compile('^(' + NAME_PATTERN_STR + ')$')  # Name pattern (loose match)
 NAME_PATTERN_HISTORY_REGEX = re.compile('^(' + NAME_PATTERN_STR + ')\^([0-9]*)$')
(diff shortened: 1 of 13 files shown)


codalab/lib/spec_util.py
```diff
@@ -11,7 +11,9 @@
 UUID_PREFIX_REGEX = re.compile('^0x[0-9a-f]{1,31}$')
 BEGIN_NAME_STR = '[a-zA-Z_]'
 NAME_STR = '[a-zA-Z_][a-zA-Z0-9_\.\-]*'
-NAME_PATTERN_STR = '[%\*a-zA-Z0-9_\.\-]+'  # Allow % for matching wildcard (SQL syntax), and * (regular expressions)
+NAME_PATTERN_STR = (
```
percyliang: re-format

frontend/package.json
```diff
@@ -58,7 +58,7 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "check-ci": "CI=true eslint src && prettier --config ./.prettierrc --check 'src/**/*.js'",
-    "format": "CI=true eslint src --fix && prettier --config ./.prettierrc --check 'src/**/*.js' --write"
+    "format": "CI=true eslint src --fix && prettier --config ./.prettierrc 'src/**/*.js' --write"
```
percyliang: key change

rc
```diff
@@ -31,3 +31,6 @@ if [ ! -e "$codalabdir/venv/bin/register-python-argcomplete" ]; then
 else
     eval "$($codalabdir/venv/bin/register-python-argcomplete cl)"
 fi
+
+# Used the version of `cl` that's installed into `venv` by default.
+. venv/bin/activate
```
percyliang, line 36: so if you `. rc` and run `pre-commit.sh`, then you can use the latest version for quick testing.

pre-commit.sh
```diff
@@ -18,5 +18,5 @@ venv/bin/mkdocs build || { rm -rf venv; exit 1; }  # Outputs to `site`
 # Note: run `venv/bin/mkdocs serve` for a live preview
 
 # Fix Python and JavaScript style (mutates code!)
-venv/bin/black codalab scripts *.py || exit
-npm run --prefix frontend format
+venv/bin/black codalab scripts *.py || exit 1
+npm run --prefix frontend format || exit 1
```
percyliang: Drive-by cleanup

epicfaace: Can we do `set -e` at the beginning of the script and remove the `exit 1`'s everywhere?

frontend/package.json
epicfaace: Why remove `--check`?

pre-commit.sh
percyliang: good idea, done

frontend/package.json
percyliang: Reverted...turned out I had the wrong version of prettier.

epicfaace (approved):

PR #2733, opened 2020-08-09
Create an indicator for when the cursor is "above" all bundles

### LINKED ISSUE #2406: indicator for when the cursor is "above" all bundles
When you first load a worksheet, when you click out and when you move up beyond the first bundle, the cursor is in an invisible position, which makes it hard to reason about.  Consider adding the blue cursor bar (or a compressed version of it) to an empty space above the first bundle to make the state more intuitive.

michaeldu1:

Previously unintuitive if there is no current focused item

### Related issues
fixes #2406 

### Screenshots

Screen Shot 2020-08-09 at 5 22 11 PM
Screen Shot 2020-08-09 at 5 22 17 PM

### Checklist


percyliang:
The indicator should be a block that's above all the blocks and is not specific to a table.

epicfaace (changes requested):
- The indicator should appear only once, at the top of the worksheet.
- The indicator should only be visible when no item is focused.

image

michaeldu1:
Screen Shot 2020-08-23 at 10 42 28 PM

epicfaace (changes requested):
It seems that the title of the worksheet moves down a bit when I focus and then unfocus from the dummy header. Can you fix the margins so that this is not the case?

See:

image

image



Also, potentially it would look better if we highlighted this grey space in blue instead (instead of adding space to the top of the white page)?

image

epicfaace:
New screenshot:

image

epicfaace (approved):

PR #2734, opened 2020-08-10
Make SlurmBatchWorkerManager use per-user scratch spaces

### LINKED ISSUE #2689: SlurmWorkerManager shared output path should be all-writeable by default
https://github.com/codalab/codalab-worksheets/blob/2de9c561bfd9b2c1a39c8773fd9e3766bb96b311/codalab/worker_manager/slurm_batch_worker_manager.py#L241

Else, other users that run the workermanager get:

```
PermissionError: [Errno 13] Permission denied: '/scr/scr-with-most-space/slurm-codalab-worker-scratch/user-codalab-slurm-worker-a2b76ac7'
srun: error: jagupard21: task 0: Exited with exit code 1
```

nelson-liu:
Fixes #2689

### DIFF
--- a/codalab/worker_manager/slurm_batch_worker_manager.py
+++ b/codalab/worker_manager/slurm_batch_worker_manager.py
@@ -244,7 +244,9 @@ class SlurmBatchWorkerManager(WorkerManager):
         else:
             work_dir_prefix = Path()
 
-        worker_work_dir = work_dir_prefix.joinpath(Path('slurm-codalab-worker-scratch', worker_id))
+        worker_work_dir = work_dir_prefix.joinpath(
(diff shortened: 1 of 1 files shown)


percyliang (approved):
Should we rename to make it clear that this is a directory containing worker directories, not an actual worker directory?

nelson-liu:
yeah, that's a good idea—I'm struggling to come up with a nice name, though...`<user>-slurm-codalab-worker-scratches` doesn't seem right, and every word I am thinking of to replace `scratches` could still lead to confusion over whether this is an actual worker directory, or a directory containing worker directories.

percyliang:
What about `<user>-SlurmBatchWorkerManager-scratch`?

PR #2735, opened 2020-08-10
Validate new worksheet name on frontend

### LINKED ISSUE #2408: validate worksheet name when creating workshet
On frontend, should error when people try to enter an invalid worksheet name during creation (e.g., include a space).  Right now, we get a silent failure.
Screenshot from 2020-06-14 15-39-34

michaeldu1:

Previously silent error on front end 

### Related issues

fixes #2408 

### Checklist


frontend/src/components/NavBar.js
```diff
@@ -92,6 +92,15 @@ class NavBar extends React.Component<{
 
     createNewWorksheet() {
         this.resetDialog();
+        if (this.state.newWorksheetName.includes(' ')) {
```
percyliang: Can you make this logic match the backend (there's a regex)?  Put a comment to that file to keep it consistent?

frontend/src/components/NavBar.js
```diff
@@ -37,6 +37,7 @@ import { executeCommand } from '../util/cli_utils';
 import DOMPurify from 'dompurify';
 
 const kDefaultWorksheetName = 'unnamed';
+const NAME_STR_REGEX = new RegExp('^[a-zA-Z_][a-zA-Z0-9_.-]*$'); // matches regex on the backend in spec_util.py
```
epicfaace:
```suggestion
const NAME_STR_REGEX = /^[a-zA-Z_][a-zA-Z0-9_.-]*$/; // matches regex on the backend in spec_util.py
```

frontend/src/components/NavBar.js
```diff
@@ -37,6 +37,7 @@ import { executeCommand } from '../util/cli_utils';
 import DOMPurify from 'dompurify';
 
 const kDefaultWorksheetName = 'unnamed';
+const NAME_STR_REGEX = /^[a-zA-Z_][a-zA-Z0-9_.-]*$/; // matches regex on the backend in spec_util.py
```
epicfaace: Can you rename this to NAME_REGEX, and also extract this regex into constants.js? It's also declared / used in NewWorksheet.js, so we want to only define it in one place.

frontend/src/components/NavBar.js
```diff
@@ -92,6 +93,15 @@ class NavBar extends React.Component<{
 
     createNewWorksheet() {
         this.resetDialog();
+        if (!NAME_STR_REGEX.test(this.state.newWorksheetName)) {
+            this.setState({
+                snackbarShow: true,
+                snackbarMessage: `Worksheet name with name ${this.state.newWorksheetName} may not contain spaces`,
```
epicfaace: We should make the message the same as on the backend: https://github.com/codalab/codalab-worksheets/blob/69ee12fa74c0699cfe9d3f998179ef85959a5eef/codalab/lib/spec_util.py#L58

epicfaace (changes requested):
Can you actually check the logic here with what's already in NewWorksheet.js (https://github.com/codalab/codalab-worksheets/blob/ba9ed2f032f4627b14915454daa3ef6d161c4408/frontend/src/components/worksheets/NewWorksheet.js#L59)? I wonder if that logic just needs some fixing instead of having to add new logic in NavBar.js.

epicfaace:
> Can you actually check the logic here with what's already in NewWorksheet.js (
> 
> https://github.com/codalab/codalab-worksheets/blob/ba9ed2f032f4627b14915454daa3ef6d161c4408/frontend/src/components/worksheets/NewWorksheet.js#L59
> 
> )? I wonder if that logic just needs some fixing instead of having to add new logic in NavBar.js.


@michaeldu1 do you know if that logic I linked to in this comment is relevant at all to solving this issue? Wouldn't want there to be duplicate code for checking names if one isn't used.

michaeldu1:
@epicfaace I don't see NewWorksheet.js used anywhere in the repo -- is it possible that this an old component?

epicfaace:
> @epicfaace I don't see NewWorksheet.js used anywhere in the repo -- is it possible that this an old component?

Yeah, that's possible. I was thinking we could remove it if we didn't need to use it; however, let's just merge this PR for now.

epicfaace (approved):

PR #2740, opened 2020-08-10
Worker telemetry with Sentry

### LINKED ISSUE #2720: more graceful way of dealing with worker failures
Currently, the worker sleeps for 1 hour and begs for help upon failure, which is good impetus for us to figure out what's wrong, but this means bundles go offline even if the error is transient.  We don't want to just exit and restage the bundle immediately since that might result in an infinite loop.

As a compromise, let's change the behavior to: (i) sleep 5 minutes, (ii) send an email to the CodaLab team so that we actually take a look, and (iii) exit so that the bundle gets restaged.

nelson-liu:

Fixes https://github.com/codalab/codalab-worksheets/issues/2720

This PR adds support for doing some telemetry with [Sentry](https://sentry.io/). When I sat down to fix the issue above, I realized that (1) email is a bit annoying, since you have to configure this per-server and possibly even per-user (i.e., we can't just use codalab sendgrid here, and different servers might have different postfix configurations / use different ports, etc etc), and (2) people have already built robust tooling for this sort of stuff (e.g., Sentry).

So, I decided to explore it a bit and it turns out to be more featureful and easy to use than I expected. In short, Sentry will:

- Log any unhandled exceptions that cause the application to crash
- Log handled exceptions in the `Worker` before we sleep for an hour or exit.

One primary concern I had was to make it **exceedingly clear** when user data is getting logged and sent to remote servers by Sentry (and that this is False by default everywhere). When Sentry is initialized, a glaring error message is printed:

```
2020-08-09 19:31:45,947 @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
2020-08-09 19:31:45,948 @       ATTENTION: Reporting exceptions to the CodaLab team via Sentry.      @
2020-08-09 19:31:45,948 @  This will log personally-identifying data (e.g., username, instance URL). @
2020-08-09 19:31:45,948 @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
```

### DIFF
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -14,7 +14,7 @@ import sys
 import psutil
 
 from codalab.lib.formatting import parse_size
-from codalab.lib.telemetry_util import initialize_sentry, load_sentry_data
+from codalab.lib.telemetry_util import initialize_sentry, load_sentry_data, using_sentry
 from .bundle_service_client import BundleServiceClient, BundleAuthException
 from . import docker_utils
 from .worker import Worker
@@ -200,16 +200,16 @@ def main():
         format='%(asctime)s %(message)s', level=(logging.DEBUG if args.verbose else logging.INFO)
     )
 
-    # Initialize sentry logging (this is a null-op if
-    # the CODALAB_SENTRY_INGEST_URL env var is not set).
-    initialize_sentry()
+    # Initialize sentry logging
+    if using_sentry():
+        initialize_sentry()
 
     # This quits if connection unsuccessful
     bundle_service = connect_to_codalab_server(args.server, args.password_file)
 
-    # Load some data into sentry (this is a null-op if
-    # the CODALAB_SENTRY_INGEST_URL env var is not set).
-    load_sentry_data(username=bundle_service._username, **vars(args))
+    # Load some data into sentry
+    if using_sentry():
+        load_sentry_data(username=bundle_service._username, **vars(args))
 
     if args.shared_file_system:
         # No need to store bundles locally if filesystems are shared
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -14,6 +14,7 @@ import sys
 import psutil
 
 from codalab.lib.formatting import parse_size
+from codalab.lib.telemetry_util import initialize_sentry, load_sentry_data
 from .bundle_service_client import BundleServiceClient, BundleAuthException
 from . import docker_utils
 from .worker import Worker
@@ -193,12 +194,23 @@ def connect_to_codalab_server(server, password_file):
 
 def main():
     args = parse_args()
-    # This quits if connection unsuccessful
-    bundle_service = connect_to_codalab_server(args.server, args.password_file)
+
     # Configure logging
     logging.basicConfig(
         format='%(asctime)s %(message)s', level=(logging.DEBUG if args.verbose else logging.INFO)
     )
+
+    # Initialize sentry logging (this is a null-op if
+    # the CODALAB_SENTRY_INGEST_URL env var is not set).
+    initialize_sentry()
+
+    # This quits if connection unsuccessful
+    bundle_service = connect_to_codalab_server(args.server, args.password_file)
+
+    # Load some data into sentry (this is a null-op if
+    # the CODALAB_SENTRY_INGEST_URL env var is not set).
+    load_sentry_data(username=bundle_service._username, **vars(args))
+
     if args.shared_file_system:
         # No need to store bundles locally if filesystems are shared
         local_bundles_dir = None
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -13,7 +13,7 @@ from typing import Optional, Set, Dict
 import psutil
 
 import docker
-from codalab.lib.telemetry_util import capture_exception, initialize_sentry, load_sentry_data
+from codalab.lib.telemetry_util import capture_exception, using_sentry
 import codalab.worker.docker_utils as docker_utils
 import requests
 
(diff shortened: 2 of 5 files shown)


epicfaace (commented):
Looks great! Where have you created the Sentry account?

I also think it would be best if we made the sentry ingest URL (and enabling Sentry in general) a configurable option to the CodaLab server. This way, if someone starts their own CodaLab server, they won't be putting data into our Sentry logs by default.

nelson-liu:
> Looks great! Where have you created the Sentry account?

I made a sentry team, which can add individual members, etc. If you dm me your email (or any email), I can add you.

> I also think it would be best if we made the sentry ingest URL (and enabling Sentry in general) a configurable option to the CodaLab server. This way, if someone starts their own CodaLab server, they won't be putting data into our Sentry logs by default.

Yup, great idea, and we also shouldn't force people hosting their own servers to use it.

nelson-liu:
> I also think it would be best if we made the sentry ingest URL (and enabling Sentry in general) a configurable option to the CodaLab server. This way, if someone starts their own CodaLab server, they won't be putting data into our Sentry logs by default.


This is done, but can you check if I did it properly [`749fd60` (#2740)](https://github.com/codalab/codalab-worksheets/pull/2740/commits/749fd6037c9a8231fca1bce241aff46c8f913ff7) ? I added a new arg to codalab_service that would take an ingest URL, but i'm not sure how that arg is exposed to the server itself—is it through an (automatically created) environment variable? or would we have to set the environment variable when we deploy?

nelson-liu:
(i also regenerated the ingest url, since it was publicly exposed)

teetone (approved):

codalab/worker/worker.py
```diff
@@ -109,6 +110,11 @@ def __init__(
         self.pass_down_termination = pass_down_termination
         self.exit_on_exception = exit_on_exception
 
+        # Initialize sentry logging (this is a null-op if
+        # the CODALAB_SENTRY_INGEST_URL env var is not set).
+        initialize_sentry()
```
percyliang: Why do you need this both here and in `main.py`?

codalab/worker/worker.py
```diff
@@ -259,13 +265,18 @@ def start(self):
...
+                capture_exception()
                 traceback.print_exc()
                 if self.exit_on_exception:
-                    logger.error('Encountered exception, terminating the worker...')
+                    logger.error(
+                        'Encountered exception, terminating the worker after sleeping for 5 minutes...'
+                    )
                     self.terminate = True
+                    # Sleep for 5 minutes
+                    time.sleep(5 * 60)
```
percyliang, line 272: Why sleep 5 min?

codalab/worker/main.py
```diff
@@ -193,12 +194,23 @@ def connect_to_codalab_server(server, password_file):
...
-    bundle_service = connect_to_codalab_server(args.server, args.password_file)
+
     # Configure logging
     logging.basicConfig(
         format='%(asctime)s %(message)s', level=(logging.DEBUG if args.verbose else logging.INFO)
     )
+
+    # Initialize sentry logging (this is a null-op if
+    # the CODALAB_SENTRY_INGEST_URL env var is not set).
+    initialize_sentry()
```
percyliang: Seems like it would be more explicit if it were something like:
```if use_sentry(): initialize_sentry()```
where `use_sentry` is something from sentry_util.py` that just does the simple check.
(Same for `load_sentry_data` below.)
That way you don't need the detailed comment and to maintain it.

codalab/worker/worker.py
nelson-liu: Do people ever instantiate workers without using `main.py`? Was trying to cover that case

codalab/worker/worker.py
nelson-liu: https://github.com/codalab/codalab-worksheets/issues/2720#issue-674801797 , i thought your rationale was to mitigate infinite loops

codalab/worker/main.py
nelson-liu: good idea, i'll change it to use that.

nelson-liu:
How do people feel about storing this in an environment variable? I noticed that the codalab email credentials are actually stored in config.json...should we do the same here, or is environment variable preferred?

codalab/worker/worker.py
nelson-liu: (and i wanted it in `main.py` in case there were any exceptions between when args are parsed and when the `Worker` is constructed)

percyliang: I don't think people should be instantiating workers directly?  But I think that `initialize_sentry()` probably should just be called as early as possible in the appropriate top-level main script (e.g., rest server).

codalab/worker/worker.py
percyliang: Hm, I was thinking the 5 minutes would replace the sleeping for 1 hour logic.

nelson-liu: even in the case where workers don't exit on exception (this isn't the default)?

percyliang:
We used to have everything in `config.json`, but with `codalab_service.py`, things are increasingly moving towards environment variables, though things haven't been quite migrated yet.  I think `config.json` should only be for client stuff going forward, since for server deploys, we use `codalab_service.py` and environment variables.  Now, `cl-worker` is client-side, but presumably only the workers that are started by codalab_service.py are going to want to use sentry?

nelson-liu:
> Now, cl-worker is client-side, but presumably only the workers that are started by codalab_service.py are going to want to use sentry?

I definitely plan on using it with any workers I launch. Might also be good to turn on for workers launched by `nlp-cl-worker-manager`.

nelson-liu:
(to be clear, i think it'd fine if the client-side `cl-worker` looks at environment variables or at config.json—I have no real opinion here, so I'm curious if those with more context on the project have a preference)

percyliang:
Hm, maybe leaving it as an environment variable is fine.

PR #2742, opened 2020-08-11
Enable arbitrary field modification for `cl edit`

### LINKED ISSUE #2217: Enable editing any bundle field through the CLI

Right now, `cl edit` on the CLI only allows you to edit the name, description, and tags. However, on the web interface, you can change arbitrary fields on bundles (e.g., `request_memory`, `request_network`, etc). It'd be nice to be able to do this on the CLI as well.

The ability to do `cl edit <insert any field here> <insert your value here>`

yshen-google:

As requested by #2217 we would like to have an option to modify arbitrary bundle field within `cl edit`. Tested locally and works fine.

### Related issues

fixes #2217 

### Checklist


percyliang (changes requested):
Can we remove the changes not relevant to this PR?  Could you please add a test?

What happens when you set an invalid field?  Can you check that the field name and value are valid and provide an error message?

yshen-google:
> Can we remove the changes not relevant to this PR? Could you please add a test?
> 
> What happens when you set an invalid field? Can you check that the field name and value are valid and provide an error message?

I tested locally in the case of invalid field name and value, it will raise UsageError as expected with proper error message. I also added unit test coverage for the valid and invalid cases.
Also, there are some changes automatically done by the linter. If you think these shouldn't be included in this PR, I can remove them manually.

nelson-liu:
> Also, there are some changes automatically done by the linter. If you think these shouldn't be included in this PR, I can remove them manually.


which linter? The only python linter we use is `black`, which wouldn't make those changes

yshen-google:
> > Also, there are some changes automatically done by the linter. If you think these shouldn't be included in this PR, I can remove them manually.
> 
> which linter? The only python linter we use is `black`, which wouldn't make those changes

I think `Pylint`? I used some other linter at the beginning, but later realized we use `black`. I did reformat with `black` again also, but it didn't revert those changes... I guess these changes are acceptable by both linters. So I can only revert them manually if needed.

yshen-google:
> > Also, there are some changes automatically done by the linter. If you think these shouldn't be included in this PR, I can remove them manually.
> 
> which linter? The only python linter we use is `black`, which wouldn't make those changes

May also be autopep8

yshen-google:
Fixed all failing tests. @percyliang @nelson-liu PTAL.

nelson-liu:
> So I can only revert them manually if needed.

would you mind doing so? It'd be nice to have changes only relevant to this PR. Sorry for the inconvenience!

yshen-google:
> > So I can only revert them manually if needed.
> 
> would you mind doing so? It'd be nice to have changes only relevant to this PR. Sorry for the inconvenience!

Sure, no worries. The lint changes are removed. PTAL @nelson-liu

test_cli.py
```diff
@@ -2171,6 +2172,30 @@ def test(ctx):
     check_equals(_run_command([cl, 'uinfo', user_name, '-f', 'has_access']), 'True')
 
 
+@TestModule.register('edit_arbitrary_field')
```
nelson-liu: Thoughts on making this into a general edit test while we're at it?

epicfaace: Yes, can you rename it to `edit` instead? We want our test suites in test_cli.py to be named not too specifically -- it's better to be general.

epicfaace (changes requested):

test_cli.py
```diff
@@ -2171,6 +2172,30 @@ def test(ctx):
...
+def test(ctx):
+    uuid = _run_command([cl, 'run', 'echo hello'], request_memory='10m')
+    check_equals('10m', get_info(uuid, 'request_memory'))
+    _run_command([cl, 'edit', uuid, '--field', 'request_memory', '12m'])
+    check_equals('12m', get_info(uuid, 'request_memory'))
+
+    # invalid field name
+    try:
+        _run_command([cl, 'edit', uuid, '-f', 'invalid_field', 'value'], expected_exit_code=1)
+    except UsageError:
```
epicfaace: You should throw an error if a `UsageError` is _not_ thrown.

epicfaace (changes requested):

test_cli.py
yshen-google: Done.

yshen-google:
@nelson-liu @epicfaace @percyliang Requested changes have been done. PTAL.

nelson-liu (commented):
LGTM! I thought there was a more idiomatic way of testing for specific exceptions with `_run_command`— @epicfaace would probably know more than i do .

epicfaace (approved):
This seems good. @nelson-liu , I think the idiomatic way would be just not to check for the exception type at all.

Actually, I think that would be fine and make the code a bit cleaner -- @yshen-google , could you do that? That means that your test would only need to have a line like:

```
_run_command([cl, 'edit', uuid, '-f', 'invalid_field', 'value'], expected_exit_code=1)
```

and you wouldn't need to add the try / except clauses.

yshen-google:
> This seems good. @nelson-liu , I think the idiomatic way would be just not to check for the exception type at all.
> 
> Actually, I think that would be fine and make the code a bit cleaner -- @yshen-google , could you do that? That means that your test would only need to have a line like:
> 
> ```
> _run_command([cl, 'edit', uuid, '-f', 'invalid_field', 'value'], expected_exit_code=1)
> ```
> 
> and you wouldn't need to add the try / except clauses.

Done. My original intention was to explicitly catch UsageError so that we know a proper exception is raised. But I am fine with this too.

percyliang (approved):

Issue #2714, opened 2020-08-06
Rethink testing infrastructure


epicfaace:
- REST API tests
- Unit test code coverage check in GHA

- multiple users/permissions

percyliang:
Should first start by writing up a document taking stock of our current tests (unit tests, test cli, frontend selenium tests) and describing the various types of things we need to test.  Tests are currently hard to write because we haven't invested in the proper scaffolding to make testing ergonomic.  Things which are particularly challenging: multiple users/permissions, live running workers, etc.

PR #2750, opened 2020-08-13
Reorganize and scaffold testing infrastructure

epicfaace:
For #2714.

- Reorganize Python tests to be in the `tests` directory.
- Add scaffolding for frontend unit tests that mock network calls; this should be a lot less expensive than using selenium.
- Add scaffolding for REST API tests. They use [webtest](https://docs.pylonsproject.org/projects/webtest/en/latest/), which can test a WSGI server without having to actually run a HTTP server, so should be significantly faster than our current CLI tests.
- Add code coverage checks for both the frontend unit tests and REST API tests. Right now, the thresholds are set to a low value (10 - 20%), but we should increase it as we get more coverage.
- Outline testing philosophy / guidelines in the "code overview" doc. Basically, we should always add unit tests with any new change we have.

See this doc (https://github.com/codalab/codalab-worksheets/blob/tests/docs/Code-Overview.md#tests) for more details on each test type and the testing philosophy / guidelines that I've proposed.


epicfaace:
@teetone CI is passing and this PR is ready now -- can you review now?

epicfaace:
just bumping this @percyliang and/or @teetone and/or @yiboliu

teetone (approved):

PR #2759, opened 2020-08-16
Some more conversions of logger.error to logger.warning

nelson-liu:
These aren't really errors, but more warnings for the user. I'm mostly making this change because sentry records every `logger.error`, and these are rather innocuous statements.


codalab/worker/worker.py
```diff
@@ -275,7 +275,7 @@ def start(self):
                     # We sleep in 5-second increments to check
                     # if the worker needs to terminate (say, if it's received
                     # a SIGTERM signal).
-                    logger.error('Sleeping for 1 hour due to exception...please help me!')
+                    logger.warning('Sleeping for 1 hour due to exception...please help me!')
```
epicfaace, line 278: Is this an innocuous message?

nelson-liu: i think so, because we've already printed the `Exception` traceback just before it (`traceback.print_exc()`). The message is just conveying that the worker is sleeping for an hour.

epicfaace: I'd imagine that we'd want to record this error in sentry, though. Would this mean that Sentry isn't recording this though? or would the previous `traceback.print_exc()` function also get logged in Sentry?

nelson-liu: the issue is that logging this message isn't actually informative of what's happening. The current behavior with sentry (which i've actually really enjoyed using, fwiw) means that an exception gets logged 3 times:
- The actual exception is logged. This is the most important, since it also provides the stack trace / code context / what the values of local variables were, which is enormously useful for debugging
- The traceback is logged as text (I think from `traceback.print_exc()`).
- A message `Sleeping for 1 hour due to exception...please help me!` / `'Encountered exception, terminating the worker after sleeping for 5 minutes...'` is logged, which is the least informative of the bunch.

epicfaace (approved):

PR #2762, opened 2020-08-16
switch shift-enter and enter in New Run logic change

### LINKED ISSUE #2392: switch shift-enter and enter in New Run
In the worksheets New Run interface, enter creates a new line for a command, and ctrl-enter creates the run.  Should make enter create the run and shift-enter create a line line like in slack to make it more intuitive and the common case easier.

michaeldu1:

https://github.com/codalab/codalab-worksheets/issues/2392

fixes #2392 

### Screenshots

Tested locally

### Checklist


epicfaace (changes requested):
Please update the keyboard shortcuts modal to have the new keyboard shortcuts.

frontend/src/components/worksheets/InformationModal/InformationModal.js
```diff
@@ -45,13 +45,13 @@ const KeyboardShortcutModal = ({ show, toggle }) => (
                     </tr>
                     <tr>
                         <td>
-                            <kbd>enter</kbd>
+                            <kbd>shift+enter</kbd>
```
epicfaace: This isn't correct. enter is still a keyboard shortcut for "start editing markdown".

You need to add a new entry for "shift+enter" that means "Add a new line in markdown".

epicfaace (changes requested):

frontend/src/components/worksheets/InformationModal/InformationModal.js
```diff
@@ -45,15 +45,15 @@ const KeyboardShortcutModal = ({ show, toggle }) => (
...
```
epicfaace: @michaeldu1 oops, I didn't realize that this PR was about switching keyboard shortcuts in the *new run* textbox, not the regular markdown textbox. In that case, we don't need to update the keyboard shortcuts for markdown.

epicfaace (approved):

Issue #2566, opened 2020-07-06
Github Actions - check for CLI / REST documentation generation


epicfaace:
We don't have checks yet to make sure that the latest CLI / REST documentation is generated.

percyliang:
Functionally, this is `./pre-commit.sh` and making sure there are no changes?

epicfaace:
Yes. We can just run:

```
./pre-commit.sh
git diff --exit-code
```

yiboliu:
PR closed

PR #2764, opened 2020-08-17
Add pre-commit.sh to the test.yml

### LINKED ISSUE #2176: Add precommit to GitHub Actions
- Add flag (--check?) to doc generating script to fail if file isn’t up to date
- Run pre-commit in GitHub Actions
- Everyone just remember to run pre-commit.sh

yiboliu:
This is to resolve #2176 by adding pre-commit.sh check to test.yml in workflows.

Add running pre-commit.sh before commit.  

### Related issues
#2176
#2566 

### Checklist


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+
+    # Runs a single command using the runners shell
+    - name: Run a one-line script
+      run: ./pre-commit.sh
```
percyliang: Pre-commit mutates the code, so need to check that it didn't change.  This also seems redundant with some of the checks already there...should we organize a bit better (cc @epicfaace)?

epicfaace:
@yiboliu , issue #2176 is not about adding `pre-commit.yml`. We don't want to run precommit checks automatically upon every commit, actually, because the checks can take a long time to run.

That issue is just about adding a new check in the GitHub Actions `format` job, in which we want to run `./pre-commit.sh` and make sure that no files have been changed. Sorry if that wasn't clear!

epicfaace:
@yiboliu can you please rename this PR and change the description (it might be better actually to just create a separate PR)? Additionally, you may need to run ./pre-commit.sh first to fix the formatting for CI.

pre-commit.sh
```diff
@@ -5,8 +5,10 @@
 # Script fails immediately when any command fails.
 set -e
 
+python -m pip install --user virtualenv
+
```
epicfaace: `python` might refer to Python 2 for many people's systems -- could you keep the original virtualenv setup on this file?

epicfaace (changes requested):
Looks like the precommit change is failing. Can you run pre-commit.sh and then check in the changes so that the build passes?

.github/workflows/test.yml
```diff
@@ -30,8 +30,8 @@ jobs:
           key: pip-${{ hashFiles('requirements.dev.txt') }}
           restore-keys: |
             pip-
+      - run: ./pre-commit.sh && git diff --exit-code
       - run: pip install -r requirements.dev.txt
```
epicfaace: I think you can remove this line now.

epicfaace (changes requested):

epicfaace (approved):

Issue #2745, opened 2020-08-12
gpuworkers were down


yiboliu:
**Describe the bug**
```Traceback (most recent call last):
  File "/opt/codalab/worker/dependency_manager.py", line 169, in loop
    self._save_state()
  File "/opt/codalab/worker/dependency_manager.py", line 87, in _save_state
    self._state_committer.commit({'dependencies': self._dependencies, 'paths': self._paths})
  File "/opt/codalab/worker/state_committer.py", line 29, in commit
    with tempfile.NamedTemporaryFile() as f:
  File "/usr/lib/python3.6/tempfile.py", line 690, in NamedTemporaryFile
    (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
  File "/usr/lib/python3.6/tempfile.py", line 401, in _mkstemp_inner
    fd = _os.open(file, flags, 0o600)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpes4wfts8'
```
**Codalab Information**
- *Your username*: yiboliu
 - *OS (if on CLI)*: Linux

PR #2765, opened 2020-08-17
Resolve potential file not found error

yiboliu:
Sometimes temp file created could be deleted and thus a FileNotFoundError occurs and the gpu worker is down. This commit applies a change suggested by [discussion forum](https://bugs.python.org/issue29573)

### Related issues
#2745

### Checklist

### DIFF
--- a/codalab/worker/state_committer.py
+++ b/codalab/worker/state_committer.py
@@ -1,6 +1,11 @@
+import logging
+import os
 import tempfile
 import shutil
 from . import pyjson
+
+
+logger = logging.getLogger(__name__)
 
 
 class BaseStateCommitter(object):
@@ -15,6 +20,7 @@ class BaseStateCommitter(object):
 
 class JsonStateCommitter(BaseStateCommitter):
     def __init__(self, json_path):
+        self.temp_file = None
         self._state_file = json_path
 
     def load(self, default=None):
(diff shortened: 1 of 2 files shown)


codalab/worker/state_committer.py
```diff
@@ -26,7 +28,14 @@ def load(self, default=None):
 
     def commit(self, state):
         """ Write out the state in JSON format to a temporary file and rename it into place """
-        with tempfile.NamedTemporaryFile() as f:
-            f.write(pyjson.dumps(state).encode())
-            f.flush()
-            shutil.copyfile(f.name, self._state_file)
+        with tempfile.NamedTemporaryFile(delete=False) as f:
```
teetone, line 35: Since we are setting `delete=False`, can you add a test that ensures that the temp file eventually gets deleted?

codalab/worker/state_committer.py
```diff
@@ -26,7 +28,14 @@ def load(self, default=None):
...
+            try:
+                f.write(pyjson.dumps(state).encode())
+                f.flush()
+                shutil.copyfile(f.name, self._state_file)
+            finally:
+                try:
+                    os.unlink(f.name)
+                except FileNotFoundError:
+                    logging.info("Problem occurred in creating temp file {}".format(f.name))
```
teetone: You actually need to get the logger with `logger = logging.getLogger(__name__)` and then use the logger to log `logger.info("something")`.

teetone: We are catching an error with `os.unlink`. I think the log message should mention that instead.

teetone: It should also be `logger.error` instead of `logger.info`.

codalab/worker/state_committer.py
```diff
@@ -26,7 +28,14 @@ def load(self, default=None):
...
+                    raise
```
teetone: Do we need to raise an error here?

teetone (changes requested):

teetone:
@yiboliu Just to follow up on our meeting earlier, can you add a link to the discussion forum that suggested this solution and a comment in the code explaining why `delete=False` is needed?

yiboliu:
> @yiboliu Just to follow up on our meeting earlier, can you add a link to the discussion forum that suggested this solution and a comment in the code explaining why `delete=False` is needed?

It's already there. If you click the `discussion forum` the website will be directed.

codalab/worker/state_committer.py
yiboliu: So after I used logger.error, we shouldn't need to raise an error anymore.

codalab/worker/state_committer.py
yiboliu: Added. Let's see if it works

teetone (approved):

teetone:
> > @yiboliu Just to follow up on our meeting earlier, can you add a link to the discussion forum that suggested this solution and a comment in the code explaining why `delete=False` is needed?
> 
> It's already there. If you click the `discussion forum` the website will be directed.

Can you also add a comment in the code?

PR #2766, opened 2020-08-18
Remove "CREATED" docker containers when they've failed to start

nelson-liu:

Currently, when we try to run a container, and then fail, the worker will try again and then fail because the desired container name already exists ( https://github.com/codalab/codalab-worksheets/pull/2710 ). This is a rather uninformative error message.

~~Instead, I propose that we should just remove the docker container after it exits. This is the equivalent of doing `docker run --rm <something>`. So, if it passes, things are automatically cleaned up (and we can remove that part of the worker logic). If it fails, a more-informative exception that actually signals the issue will be raised.~~

This doesn't work, since we need to inspect docker containers after they exit, so we actually can't remove them (since we need them around). Instead, let's just remove them if we are unsuccessful in starting them.

### Related issues

https://github.com/codalab/codalab-worksheets/pull/2710


nelson-liu:
(The `auto_remove` argument was added in 2017, so CodaLab probably predates its existence, and the logic for removing containers after the runs have finished was probably just copy and pasted and maintained during a few refactors)

nelson-liu:
@epicfaace I realized now that this doesn't work, since we need to inspect docker containers after they exit (e.g., to get their exit code), so we actually can't remove them (since we need them around). Instead, let's just remove them if we are unsuccessful in starting them.

epicfaace:
> (The `auto_remove` argument was added in 2017, so CodaLab probably predates its existence, and the logic for removing containers after the runs have finished was probably just copy and pasted and maintained during a few refactors)

Why can't we just use the `auto_remove` argument to `client.containers.run` instead of the logic that was added in this PR?

nelson-liu:
There's some logic that needs to inspect the exit codes of containers to propagate errors. If the containers are auto-removed, we can't access the exit codes, and that part of the logic breaks.

see, e.g., https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L399-L407

nelson-liu:
I'd still love to get this in, maybe @teetone / @epicfaace can take a look?

teetone (approved):

PR #2767, opened 2020-08-19
Set timeouts for urlopen calls

nelson-liu:
It's possible that `urlopen` hangs indefinitely, so we add a timeout to prevent this case (since exceptions are usually caught anyway)

### DIFF
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -10,7 +10,7 @@ import http.client
 # http://semver.org/
 CODALAB_VERSION = '0.5.21'
 BINARY_PLACEHOLDER = '<binary>'
-URLOPEN_TIMEOUT = 30
+URLOPEN_TIMEOUT_SECONDS = 30
 
 
 class IntegrityError(ValueError):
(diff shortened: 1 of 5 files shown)


codalab/lib/file_util.py
```diff
@@ -62,7 +62,7 @@ def download_url(source_url, target_path, print_status=False):
     """
     Download the file at |source_url| and write it to |target_path|.
     """
-    in_file = urllib.request.urlopen(source_url)
+    in_file = urllib.request.urlopen(source_url, timeout=30)
```
percyliang: Could we make this a constant that's used consistently throughout?

nelson-liu:
i was talking to michael xie, and he saw that all of his workers went offline due to the power outage—however, his workers were running on the main (azure) instance, so ostensibly only the workers lost connectivity but not the server...I didn't think that the workers would freeze in that setting as well, but it seems like they do.

nelson-liu:
Ok, this should be ready for another look

codalab/common.py
```diff
@@ -10,6 +10,7 @@
 # http://semver.org/
 CODALAB_VERSION = '0.5.21'
 BINARY_PLACEHOLDER = '<binary>'
+URLOPEN_TIMEOUT = 30
```
teetone: Should we rename this to `URLOPEN_TIMEOUT_SECONDS`?

teetone (approved):

PR #2769, opened 2020-08-20
Use Azure Blob Storage for file uploads

epicfaace:
Diff: https://github.com/codalab/codalab-worksheets/compare/azurite..beam-uploads

For #2327.

Running `cl upload -a` will make that bundle be uploaded to Azure Blob Storage. The Blob Storage path will be stored in azfs://storageclwsdev0/bundles.

image

Done:
- uploading
- basic downloading
- support directories
- support using these bundles as dependencies

We don't support permissions yet; we can do that in a future PR.


codalab/lib/download_manager.py
```diff
@@ -303,6 +307,9 @@ def netcat(self, uuid, port, message):
         return bytestring
 
     def _is_available_locally(self, target):
+        file_path = self._get_target_path(target)
+        if file_path.startswith("azfs://"):
```
percyliang, line 311: Maybe we should rename `_is_available_locally` since AzureFS is definitely not local (is available externally or directly)?

codalab/lib/download_manager.py
```diff
@@ -321,7 +328,7 @@ def _get_target_path(self, target):
         bundle_link_url = self._bundle_model.get_bundle_metadata(
             [target.bundle_uuid], "link_url"
         ).get(target.bundle_uuid)
-        if bundle_link_url:
+        if bundle_link_url and not bundle_link_url.startswith("azfs://"):
```
percyliang, line 331: Would be to refactor the check for `azfs://` in one place.  Also, let's think through what this will look like once we have `s3://`.  Probably makes sense to have a broader function that captures all these cases.

codalab/lib/path_util.py
```diff
@@ -298,10 +299,33 @@ def rename(old_path, new_path):
     subprocess.call(['mv', old_path, new_path])
 
 
+def parse_azure_url(url):
```
percyliang, line 302: Let's get in the habit of type annotating everything.  Should we make a dataclass instead of using a tuple?  Eventually, I think we should migrate all our tuples over to dataclasses, which will make the code easier to understand.

codalab/lib/path_util.py
```diff
@@ -298,10 +299,33 @@ def rename(old_path, new_path):
...
+    # file in directory: "azfs://storageclwsdev0/bundles/uuid/contents.zip/file1"
```
percyliang, line 303: Write a nicer docstring.

codalab/lib/upload_manager.py
```diff
@@ -4,6 +4,9 @@
 
 from codalab.common import UsageError
 from codalab.lib import crypt_util, file_util, path_util, zip_util
+from codalab.lib.beam.filesystems import FileSystems
+
+# TODO: fix methods here.
```
percyliang, line 9: What's this?  explain more?

codalab/lib/upload_manager.py
```diff
@@ -58,17 +61,18 @@ def upload_to_bundle_store(
...
         )
         bundle_link_url = getattr(bundle.metadata, "link_url", None)
-        if bundle_link_url:
-            # Don't do anything for linked bundles.
-            return
-        bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
+        bundle_path = bundle_link_url or self._bundle_store.get_bundle_location(bundle.uuid)
         try:
-            path_util.make_directory(bundle_path)
+            if not bundle_path.startswith("azfs://"):
```
percyliang, line 66: same comment as before about refactoring azfs

codalab/lib/zip_util.py
```diff
@@ -88,6 +90,7 @@ def unpack(ext, source, dest_path):
                 source = temp_path
                 delete_source = True
 
+            # TODO(Ashwin): preserve permissions with -X, if needed.
```
percyliang: Why not preserve the permissions now?

codalab/lib/zip_util.py
```diff
@@ -148,19 +152,34 @@ def resolve_source(source):
...
-                ignore_file=ignore_file,
-            )
-            return {
-                'fileobj': archived,
-                'filename': filename + '.tar.gz',
-                'filesize': None,
-                'should_unpack': True,
-                'should_simplify': False,
-            }
+            if use_azure_blob_beta:
```
percyliang, line 138: Put a comment saying that we eventually want to unify this.

codalab/model/bundle_model.py
```diff
@@ -1047,15 +1048,21 @@ def update_disk_metadata(self, bundle, bundle_location, enforce_disk_quota=False
...
         Updates the database rows for the bundle and user with the new disk use
         """
-        dirs_and_files = None
-        if os.path.isdir(bundle_location):
-            dirs_and_files = path_util.recursive_ls(bundle_location)
+        if bundle_location.startswith("azfs://"):
+            # TODO(Ashwin): add a FileSystems.size() method to Apache Beam to make this less verbose.
+            size = FileSystems.get_filesystem(bundle_location).size(bundle_location)
+            # TODO(Ashwin): use an actual hash instead of the size in the future.
+            data_hash = size
```
percyliang, line 1055: I don't think we will have a hash in the future since it would require reading everything, so just set it to None?

codalab/rest/bundles.py
```diff
@@ -736,7 +746,28 @@ def _update_bundle_contents_blob(uuid):
...
             sources = [(filename, request['wsgi.input'])]
-        if sources:
+
+        if use_azure_blob_beta:
+            local.model.update_bundle(
+                bundle,
+                {
+                    'metadata': {
+                        # TODO(Ashwin): don't hardcode storage account name.
+                        'link_url': f"azfs://storageclwsdev0/bundles/{bundle.uuid}{'/contents.zip' if filename.endswith('.zip') else '/contents'}",
```
percyliang, line 757: What's our general solution here?  We should talk about how we want to name these for dev and prod?  I guess we should also document how these are setup in the deployment repo.

percyliang, line 757: I thought we were just going to zip up everything for consistency and to simplify the logic?

codalab/server/bundle_manager.py
```diff
@@ -129,13 +137,6 @@ def _stage_bundles(self):
                 bundles_to_fail.append((bundle, str(e)))
                 continue
 
-            missing_uuids = parent_uuids - all_parent_uuids
```
percyliang: Why is this moved up?  In general, can you add a few more comments to your PRs?

codalab/worker/download_util.py
```diff
@@ -131,3 +145,84 @@ def _compute_target_info(path, depth):
     if result is None:
         raise PathException()
     return result
+
+
+def _compute_target_info_beam(path, depth):
+    # TODO (Ashwin): handle depth.
+    bundle_uuid, zip_path, zip_subpath = parse_azure_url(path)
+    if zip_subpath == "/images/":
```
percyliang: What's special about `/images/`?  Please put more comments, including adding a docstring.

codalab_service.py
```diff
@@ -26,7 +26,16 @@
 import yaml
 import tempfile
 
-DEFAULT_SERVICES = ['mysql', 'nginx', 'frontend', 'rest-server', 'bundle-manager', 'worker', 'init']
+DEFAULT_SERVICES = [
+    'mysql',
+    'nginx',
+    'frontend',
+    'azurite',
```
percyliang: Is this only for testing though?  If so, then don't need this by default.

codalab/lib/bundle_cli.py
```diff
@@ -1195,6 +1195,13 @@ def do_workers_command(self, args):
                 action='store_true',
                 default=False,
             ),
+            Commands.Argument(
+                '-a',
+                '--use_azure_blob_beta',
```
percyliang: I think our convention is to use `-` rather than `_` for all command-line args.

codalab/lib/beam/blobstoragefilesystem.py
```diff
@@ -0,0 +1,301 @@
+#
```
percyliang, line 1: Is this copied verbatim from somewhere?  If so, we should say where it's copied from, why we can't use it as a library, and any changes we're making.

tests/unit/lib/path_util_test.py
```diff
@@ -114,3 +114,32 @@ def mkdir_with_other_failure(path):
...
+            "azfs://storageclwsdev0/bundles/uuid/contents.zip"
+        )
+        self.assertEqual(bundle_uuid, "uuid")
+        self.assertEqual(zip_path, "azfs://storageclwsdev0/bundles/uuid/contents.zip")
+        self.assertEqual(zip_subpath, "")
+
+    def test_directory_with_subpath(self):
+        """Parse a URL referring to a subpath within a zipped directory."""
+        bundle_uuid, zip_path, zip_subpath = path_util.parse_azure_url(
+            "azfs://storageclwsdev0/bundles/uuid/contents.zip/a/b.txt"
```
percyliang, line 141: Refactor so that `storageclwsdev0` should only show up once, otherwise we'll have to change it in a lot of places.

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,234 @@
+import unittest
```
percyliang: These bundle manager tests shouldn't be in this PR, right?

percyliang (changes requested):
Exciting to see the Azure Blob storage come together!  I think my main comment is that we could refactor things a bit more out to set the stage for more general externalization.  Right now, Azure is hardcoded in a lot of places, and also, you kind of have to track the logic of how things are stored in a lot of places.  Could we consolidate this coupling behind a nice interface somehow?

codalab/lib/beam/filesystems.py
```diff
@@ -0,0 +1,24 @@
+import os
+from apache_beam.io.filesystems import FileSystems
+from .blobstoragefilesystem import BlobStorageFileSystem
+from azure.storage.blob import BlobServiceClient
+from azure.core.exceptions import ResourceExistsError
+
+"""
+Modifies Beam to add support for Azure Blob Storage filesystems.
+
+Based on work from https://github.com/apache/beam/pull/12581. Once that PR is added into Beam,
```
epicfaace, line 9: @percyliang see this regarding where those two files (the copied verbatim files -- blobstoragefilesystem.py and blobstorageio.py) came from. Although that PR has been merged into beam, a new version hasn't yet been released with those changes, so we may just have to keep those two files in for now.

tests/unit/server/bundle_manager/__init__.py
epicfaace: Forgot to mention this in the PR, but this PR is based on #2795, as I used some of the common utilities from that PR for testing here. It would be good to merge this once #2795 is merged.

codalab/server/bundle_manager.py
epicfaace: See my comment about this PR being based on #2795 (this was a change done in #2795)

codalab/rest/bundles.py
epicfaace: We need to pass the storage container name in as environment variables (will update to do that).

codalab/model/bundle_model.py
epicfaace: Why not just keep it at `size` (which is what I think we discussed when originally designing --link)? I can remove the TODO comment if that's final, though.

codalab/lib/upload_manager.py
epicfaace: Some of my comments / debugging statements are outdated 🙂 -- I'll update / remove them.

codalab/rest/bundles.py
epicfaace: No, I think we decided against that -- we need to use `/contents` for single files, because otherwise we may not be able to distinguish between a bundle with a directory with a single file vs. a bundle with just a single file.

codalab/bundles/named_bundle.py
```diff
@@ -24,6 +24,9 @@ class NamedBundle(Bundle):
         MetadataSpec('created', int, 'Time when this bundle was created.', generated=True, formatting='date'),
         MetadataSpec('data_size', int, 'Size of this bundle (in bytes).', generated=True, formatting='size'),
         MetadataSpec('failure_message', str, 'Error message if this run bundle failed.', generated=True,),
+        MetadataSpec('link_url', str, 'Link URL of bundle.', optional=True),
```
epicfaace, line 27: I needed to add these here so that we can have linked bundles that are non-UploadedBundles (we will need to do this anyway once run bundles upload their results to Azure Blob Storage).

codalab/lib/bundle_cli.py
epicfaace: Done

codalab/worker/download_util.py
epicfaace: That was also for testing -- updated now.

tests/unit/lib/path_util_test.py
epicfaace: Will do this for the general code, though for the tests themselves, it should be okay to use `storageclwsdev0` for now as we're just checking for parsing.

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -60,7 +60,7 @@
 }
 
 
-class BaseBundleManagerTest(unittest.TestCase):
+class TestBase:
```
epicfaace, line 66: This class has effectively become a "test base class" generic test base that's used for non-BundleManager tests, so it should eventually be moved to a separate file. However, I'm holding off on doing that until #2795 is merged, in order to minimize merge issues / conflicts.

PR #2775, opened 2020-08-21
Avoid conflicting worker resources

### LINKED ISSUE #2746: Workers getting assigned more bundles than they can run
**Describe the bug**
As of today, on the SM deployment running v0.5.20 we've started having workers that are getting assigned jobs whose total requested CPU numbers are more than what's available on the worker. In general we have workers that have N CPUs and jobs also request N CPUs. Today we started seeing workers with `2N/N` or `3N/N` CPU usage on the `cl workers` output command, and the runs have been showing the `Unexpectedly unable to assign enough resources to bundle 0x123456567897812e: Requested more CPUs (N) than available (0 currently out of N on the machine)` error message.

**Codalab Information**
- SM internal deployment, running AWS worker manager with, everything is on v0.5.20

**To Reproduce**
Unclear

**Expected behavior**
Workers should not get bundles assigned to them beyond their capacity.

teetone:

When we precompute the available worker resources, we call `_deduct_worker_resources` twice to get the list of workers that the user owns and a list of CodaLab public workers. In `_deduct_worker_resources`, a deep copy of the worker list (with resources deducted) is returned back to the caller.

deduct_called_twice

When the root account of an instance runs an experiment, two identical lists of workers are precomputed, but only the resources of one of them gets modified as they are deep copies of each other (not by reference), which will eventually cause the over-assigning of bundles to workers.

This issue is reproducible on our `dev` environment with the stress test I've added in this PR, only if the stress tests are ran with the root account. This issue is not reproducible if any other user account is used. I think the reason why we don't see this issue in `prod` is because we don't usually use the root account to create runs.

The fix in this PR is to simply remove the deep copy logic in `_deduct_worker_resources`. I am not sure why the `deepcopy` was added, but it doesn't seem necessary. If it is necessary, I can add it back and use another solution, which is to aggregate and take the minimum between the two lists. With the change in this PR, the issue is no longer reproducible.

### Related issues

Potentially resolves #2746 

### Checklist

* [x] I've added a stress test


nelson-liu:
> I am not sure why the deepcopy was added, but it doesn't seem necessary. 

This is the diff that added it: https://github.com/codalab/codalab-worksheets/commit/85d2e61a7fbe3d228cf942bc7bb3ef27b93b8f71#diff-623d90f755b41b69c721e18042ffa858L337 . Agreed that it isn't necessary, since now we explicitly want to modify the list.

nelson-liu (approved):
Thanks for figuring this out, the precomputation logic is quite confusing...

percyliang:
My recollection was that deepcopy was actually moved from somewhere else, so I wouldn't just remove it without tracking that down.  In general, the goal was to make the code more functional.  All the mutation gets really confusing, so it'd be nice to start moving the code in that direction.

bkgoksel:
Yep there used to be some other bug due to the same lists being modified around, that's why we added that deepcopy a long time ago. However this code has changed quite a bit since, so I can't point out what exactly was going on. Wouldn't another solution to this be to just skip the entire `user_owned_workers` code block if the user is the root user?

teetone:
I kept the `deepcopy` for now and added a check for the root user instead.

PR #2795, opened 2020-08-31
BundleManager tests

### LINKED ISSUE #2794: Add BundleManager tests

epicfaace:

Fixes #2794. This PR builds off of #2750.

The approach I take is to:
- Use a sqlite in-memory database. This is better than mocking out all database calls, because then the tests become less brittle and more maintainable as the implementation details change.
- Added tests for the four main methods run in `_run_iteration`.
- Fixed a bug in the BundleManager staging logic in which if parent bundles are missing, an "insufficient permissions" error comes up instead of a "missing parent bundles" error.


codalab/model/sqlite_model.py
```diff
@@ -0,0 +1,20 @@
+"""
+SQLite is a subclass of BundleModel that stores metadata on a SQLite
```
percyliang: We used to have an SQLite option, but got rid if it because it didn't support all the things we needed (like multirow insert, IIRC)?  Is that going to be a problem here?

epicfaace: I think we actually got rid of it because it was just unused code (see #1333). This PR adds the SQLite support back in in a more simplified way than the way it was initially implemented, though (see #1351).

I think we can continue to use SQLite, and if we get to problems with it we can consider modifying it. The advantages of being able to create and manipulate a fully in-memory database before each test case are very good, though.

codalab/server/bundle_manager.py
```diff
@@ -513,6 +517,7 @@ def _filter_and_sort_workers(self, workers_list, bundle, bundle_resources):
         workers_list = [
             worker for worker in workers_list if worker['cpus'] >= bundle_resources.cpus
         ]
+        print("cpus", bundle_resources.cpus)
```
epicfaace:
```suggestion
```

nelson-liu:
can you elaborate on `Fixed a bug in the BundleManager logic in which if a user has requested no CPUs and one GPU for a bundle, it treats this request as "1 CPU and 1 GPU".` What's the intended behavior?

epicfaace:
I would have thought that a request with 0 CPUs and 1 GPU should be treated as "0 CPUs and 1 GPU" -- this way, a CPU won't be requested / deducted if it's not requested by the job. However, CodaLab would treat this request as "1 CPU and 1 GPU" even though the user didn't request a CPU.

codalab/server/bundle_manager.py
```diff
@@ -118,6 +118,13 @@ def _stage_bundles(self):
         for bundle in bundles:
             parent_uuids = set(dep.parent_uuid for dep in bundle.dependencies)
 
+            missing_uuids = parent_uuids - all_parent_uuids
```
percyliang, line 125: Why was this moved?

codalab/server/bundle_manager.py
```diff
@@ -176,14 +176,18 @@ def _make_bundles(self):
...
             self._model.update_bundle(bundle, {'state': State.MAKING})
             with self._make_uuids_lock:
                 self._make_uuids.add(bundle.uuid)
             # Making a bundle could take time, so do the work in a separate
             # thread to ensure quick scheduling.
-            threading.Thread(target=BundleManager._make_bundle, args=[self, bundle]).start()
+            t = threading.Thread(target=BundleManager._make_bundle, args=[self, bundle])
+            threads.append(t)
+            t.start()
+        return threads
```
percyliang, line 194: Add type annotation to this function (let's just get in the habit of doing that)

codalab/server/bundle_manager.py
```diff
@@ -604,10 +608,11 @@ def _try_start_bundle(self, workers, worker, bundle, bundle_resources):
...
         """
         Compute the CPU limit used for scheduling the run.
-        The default of 1 is for backwards compatibility for
-        runs from before when we added client-side defaults
+        The default of 1 (if no GPUs specified)
+        is for backwards compatibility for runs from before
+        when we added client-side defaults
         """
-        if not bundle.metadata.request_cpus:
+        if not bundle.metadata.request_cpus and not bundle.metadata.request_gpus:
```
percyliang: was something breaking?

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    "failure_message": "Kill requested",
+    "request_disk": "",
+    "request_gpus": 0,
+    "remote": "vm-clws-prod-worker-3",
+    "exclude_patterns": [],
+}
+
+BASE_METADATA_MAKE_BUNDLE = {
+    "description": "",
+    "name": "run-python",
```
percyliang: make bundles are unlikely to be called `run-python`

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    "created": 1495784349,
+    "failure_message": "Kill requested",
```
percyliang: Why is there a failure message here?  Seems odd to be the default for all these bundles

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    Base class for BundleManager tests with a CodaLab Manager hitting a real, in-memory database.
+    """
+
+    def setUp(self):
+        self.codalab_manager = CodaLabManager()
+        self.codalab_manager.config['server']['class'] = 'SQLiteModel'
+        self.bundle_manager = BundleManager(self.codalab_manager)
+        self.user_id = generate_uuid()
+        self.bundle_manager._model.add_user(
```
percyliang, line 76: It'd be nice if the user creation were refactored into a testing utilities for creating all the assets that we need

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+            "username",
+            "email@email.com",
+            "first name",
```
percyliang: Put actual names here to make it more realistic / less confusing what these strings are

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,272 @@
...
+        self.bundle_manager._model.save_bundle(bundle)
+
+        self.bundle_manager._make_bundles()
+
+        self.assertTrue(self.bundle_manager._is_making_bundles())
+        bundle = self.bundle_manager._model.get_bundle(bundle.uuid)
+        self.assertEqual(bundle.state, State.MAKING)
+
+    def test_bundle_no_dependencies(self):
+        bundle = MakeBundle.construct(
```
percyliang: Can you put docstrings on all these tests to summarize what's being tested?

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,272 @@
...
+        self.assertTrue(self.bundle_manager._is_making_bundles())
+        for t in threads:
+            t.join()
+        self.assertFalse(self.bundle_manager._is_making_bundles())
+
+        bundle = self.bundle_manager._model.get_bundle(bundle.uuid)
+        self.assertEqual(bundle.state, State.READY)
+
+    def test_single_dependency(self):
+        parent = RunBundle.construct(
```
percyliang: Would be good to refactor these into general utilities for constructing bundles

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,272 @@
...
+        parent2 = RunBundle.construct(
+            targets=[],
+            command='',
+            metadata=BASE_METADATA,
+            owner_id=self.user_id,
+            uuid=generate_uuid(),
+            state=State.READY,
+        )
+        self.bundle_manager._model.save_bundle(parent2)
+        with open(self.codalab_manager.bundle_store().get_bundle_location(parent2.uuid), "w+") as f:
```
percyliang: avoid code duplication here; can create a general helper function for creating K dependencies

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,272 @@
...
+        self.assertTrue(self.bundle_manager._is_making_bundles())
+        for t in threads:
+            t.join()
+        self.assertFalse(self.bundle_manager._is_making_bundles())
+
+        bundle = self.bundle_manager._model.get_bundle(bundle.uuid)
+
+        self.assertEqual(bundle.state, State.READY)
+
+        with open(
```
percyliang: Add a comment for each code block

percyliang (commented):
It's great that we're finally getting tests for the bundle manager!  Two comments:
1) It'd be nice to add more high-level comments about what is being tested so it's easier to read at a glance.
2) I think we can refactor a lot of this code into general utilities for building up bundles/users, etc. that is independent of the bundle manager.

nelson-liu:
> I would have thought that a request with 0 CPUs and 1 GPU should be treated as "0 CPUs and 1 GPU" -- this way, a CPU won't be requested / deducted if it's not requested by the job. 

Maybe I'm missing something, but don't we always need at least 1 CPU to run the container (or at least a nonzero CPU value)?

codalab/server/bundle_manager.py
epicfaace: There was a bug in the BundleManager staging logic in which if parent bundles are missing, an "insufficient permissions" error comes up instead of a "missing parent bundles" error.

I came across this when writing one of the tests for this function, and that test now checks for the right error message.

epicfaace:
> Maybe I'm missing something, but don't we always need at least 1 CPU to run the container (or at least a nonzero CPU value)?

You could be right. In that case, I'll revert to the previous behavior.

codalab/server/bundle_manager.py
epicfaace: See @nelson-liu 's comment -- I wasn't aware that this was intended functionality; I'm going to revert this change.

tests/unit/server/bundle_manager/__init__.py
epicfaace: For now, I'm refactoring this into `bundle_manager/__init__.py`, as it's only shared across other BundleManager tests. When we add tests for other classes, we can decide which methods are common across those classes and BundleManager, and then move those to a common testing utils file.

tests/unit/server/bundle_manager/make_bundles_test.py
epicfaace: Done

tests/unit/server/bundle_manager/make_bundles_test.py
epicfaace: Done

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+                    "child_path": "src",
+                }
+            )
+        ]
+        return bundle, parent
+
+    def create_bundle_two_deps(self):
+        parent1 = self.create_run_bundle(state=State.READY)
+        with open(self.codalab_manager.bundle_store().get_bundle_location(parent1.uuid), "w+") as f:
+            f.write("hello world 1")
```
percyliang: Can we refactor and push adding contents into `create_run_bundle` to eliminate duplicate code?

tests/unit/server/bundle_manager/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+        bundle.dependencies = [
+            Dependency(
+                {
+                    "parent_uuid": parent1.uuid,
+                    "parent_path": "",
+                    "child_uuid": bundle.uuid,
+                    "child_path": "src1",
+                }
+            ),
```
percyliang, line 205: Could we wrap this in a function so we don't have raw strings?  Ideally this should be a dataclass...

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class BundleManagerMakeBundlesTest(BaseBundleManagerTest):
+    def test_no_bundles(self):
+        """With no bundles available, nothing should happen."""
+        self.bundle_manager._make_bundles()
+        self.assertFalse(self.bundle_manager._is_making_bundles())
+
+    def test_restage_stuck_bundle(self):
+        """Bundles stuck in "MAKING" should be restaged and go back to the "MAKING" state."""
```
percyliang, line 35: When do bundles get stuck in MAKING?

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,150 @@
...
+        for t in threads:
+            t.join()
+        self.assertFalse(self.bundle_manager._is_making_bundles())
+
+        bundle = self.bundle_manager._model.get_bundle(bundle.uuid)
+
+        self.assertEqual(bundle.state, State.READY)
+
+        with self.read_bundle(bundle, "src") as f:
+            self.assertEqual(f.read(), "hello world")
```
percyliang: Could we have all these strings defined in one place?

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertFalse(self.bundle_manager._is_making_bundles())
+
+        bundle = self.bundle_manager._model.get_bundle(bundle.uuid)
+
+        self.assertEqual(bundle.state, State.READY)
+
+        with self.read_bundle(bundle, "src1") as f:
+            self.assertEqual(f.read(), "hello world 1")
+        with self.read_bundle(bundle, "src2") as f:
+            self.assertEqual(f.read(), "hello world 2")
```
percyliang: same as above

tests/unit/server/bundle_manager/make_bundles_test.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+        ]
+        self.save_bundle(parent)
+        self.save_bundle(bundle)
+
+        threads = self.bundle_manager._make_bundles()
+        self.assertTrue(self.bundle_manager._is_making_bundles())
+        for t in threads:
+            t.join()
+        self.assertFalse(self.bundle_manager._is_making_bundles())
```
percyliang: Could we refactor this piece that shows up earlier?

tests/unit/server/bundle_manager_test.py
```diff
@@ -149,6 +153,23 @@ def get_sample_workers_list(self):
...
+                'max_request_memory': '100m',
+                'min_request_memory': '10m',
+                'max_request_disk': '10m',
+            }
+        }
+        self.bundle_manager = BundleManager(self.codalab_manager)
+        self.assertEqual(self.bundle_manager._max_request_time, 50)
+        self.assertEqual(self.bundle_manager._max_request_memory, 100 * 1024 * 1024)
+        self.assertEqual(self.bundle_manager._min_request_memory, 10 * 1024 * 1024)
+        self.assertEqual(self.bundle_manager._max_request_disk, 10 * 1024 * 1024)
```
percyliang, line 171: Don't we already have this above?

percyliang (commented):
Overall looks good!  The refactoring makes the tests a lot easier to read now.  I think there are some places where we can refactor even more, so that the code for each test is closer to the high-level semantics of what is to be tested (think about how you would describe it in English).

tests/unit/server/bundle_manager/make_bundles_test.py
epicfaace: See this comment:

https://github.com/codalab/codalab-worksheets/blob/d50731b304eac68a39b5dbe1548a8bb6752c823f/codalab/server/bundle_manager.py#L175-L178

tests/unit/server/bundle_manager/__init__.py
epicfaace: I'd rather not -- this `open(...)` code is only called within the reused / helper functions three times, not in actual test cases. Additionally, I can't think of a good, general way to add contents into create_run_bundle that would also take into account, say, creating a bundle directory. Finally, `create_run_bundle` doesn't itself save bundles to the database, so it might be a bit confusing if it constructs a bundle but then also persists data to the filesystem.

percyliang: What about creating another function `create_run_bundle_with_contents` or something?  It's three times now, but it seems likely that the number will grow as we add more tests.

tests/unit/server/bundle_manager/__init__.py
epicfaace: `Dependency` is an ORMObject (https://github.com/codalab/codalab-worksheets/blob/a2306fac3d4a6cfbb2d4172f19a6cf9f0c37cbf1/codalab/model/orm_object.py#L44), and I just followed the way its constructor works, so if we wanted to change the way it's constructed, we could probably do that, though I think that's beyond the scope of this PR.

tests/unit/server/bundle_manager/make_bundles_test.py
epicfaace: Done

tests/unit/server/bundle_manager_test.py
epicfaace: I just moved the old / legacy bundle manager tests we already had previously to this file -- don't want to refactor that, beyond the scope of this PR. Didn't want to remove them because we might as well keep them.

tests/unit/server/bundle_manager/make_bundles_test.py
epicfaace: Can you explain more about what you mean by this? I could define strings such as `HELLO_WORLD` and `HELLO_WORLD_1` or `SAMPLE_CONTENTS` and `SAMPLE_CONTENTS_1`, but I don't think that really adds any clarity than just directly adding strings such as `"hello world"`.

percyliang (approved):

tests/unit/server/bundle_manager/make_bundles_test.py
epicfaace: Done

tests/unit/server/bundle_manager/__init__.py
epicfaace: Done

PR #2796, opened 2020-08-31
Add type annotations

yiboliu:
Add type annotations to all functions in python code files. 

### Related issues
#2080 

### Checklist

### DIFF
--- a/codalab/bin/server.py
+++ b/codalab/bin/server.py
@@ -13,18 +13,18 @@ from codalab.lib.codalab_manager import CodaLabManager
 class ClFileWatcherEventHandler(FileSystemEventHandler):
     SERVER_PROCESS = None
 
-    def __init__(self, argv):
+    def __init__(self, argv) -> None:
         super(ClFileWatcherEventHandler, self).__init__()
         self.argv = argv
         self.restart()
 
-    def restart(self):
+    def restart(self) -> None:
         if self.SERVER_PROCESS:
             self.SERVER_PROCESS.kill()
 
         self.SERVER_PROCESS = subprocess.Popen(self.argv)
 
-    def on_any_event(self, event):
+    def on_any_event(self, event) -> None:
         extensions_to_watch = ('.js', '.py', '.html', '.css', '.tpl')
         file_extension = os.path.splitext(event.src_path)[1]
 
@@ -33,7 +33,7 @@ class ClFileWatcherEventHandler(FileSystemEventHandler):
             self.restart()
 
 
-def run_server_with_watch():
+def run_server_with_watch() -> None:
     modified_argv = list(sys.argv)
     modified_argv[0] = 'cl'
     modified_argv.remove('--watch')
@@ -51,7 +51,7 @@ def run_server_with_watch():
     observer.join()
 
 
-def main():
+def main() -> None:
     parser = argparse.ArgumentParser()
     parser.add_argument('--watch', help='Restart the server on code changes.', action='store_true')
     parser.add_argument(
--- a/codalab/bundles/__init__.py
+++ b/codalab/bundles/__init__.py
@@ -1,3 +1,5 @@
+from typing import Tuple
+
 from codalab.bundles.dataset_bundle import DatasetBundle
 from codalab.bundles.make_bundle import MakeBundle
 from codalab.bundles.program_bundle import ProgramBundle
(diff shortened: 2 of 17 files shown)


codalab/bin/server.py
```diff
@@ -13,18 +13,18 @@
...
-    def restart(self):
+    def restart(self) -> None:
         if self.SERVER_PROCESS:
             self.SERVER_PROCESS.kill()
 
         self.SERVER_PROCESS = subprocess.Popen(self.argv)
 
-    def on_any_event(self, event):
+    def on_any_event(self, event) -> None:
         extensions_to_watch = ('.js', '.py', '.html', '.css', '.tpl')
```
teetone, line 28: Should we type annotate variables too?

codalab/bundles/__init__.py
```diff
@@ -6,7 +8,7 @@
 from codalab.bundles.private_bundle import PrivateBundle
 
 
-BUNDLE_SUBCLASSES = (DatasetBundle, MakeBundle, ProgramBundle, RunBundle, PrivateBundle)
+BUNDLE_SUBCLASSES: Tuple = (DatasetBundle, MakeBundle, ProgramBundle, RunBundle, PrivateBundle)
```
teetone, line 11: Looks like the `get_bundle_subclass` function is not type annotated.

codalab/bundles/named_bundle.py
```diff
@@ -34,7 +34,7 @@ def construct(cls, row):
         row['metadata'] = dict(row['metadata'], created=int(time.time()))
         return cls(row)
 
-    def validate(self):
+    def validate(self) -> None:
```
teetone, line 37: `__repr__` and `construct` methods are missing the return type annotation.

codalab/client/json_api_client.py
```diff
@@ -559,7 +561,7 @@ def delete_relationship(
         )
 
     @wrap_exception('Unable to update authenticated user')
-    def update_authenticated_user(self, data, params=None):
+    def update_authenticated_user(self, data, params=None) -> Dict:
```
teetone, line 564: Can we specify the types for the key and value?

codalab/lib/worksheet_util.py
```diff
@@ -74,19 +76,19 @@
...
 
 
-def bundle_item(x):
-    return (x, None, '', TYPE_BUNDLE)  # TODO: replace '' with None when tables.py schema is updated
+def bundle_item(x) -> Tuple:
+    return x, None, '', TYPE_BUNDLE  # TODO: replace '' with None when tables.py schema is updated
 
 
-def subworksheet_item(x):
+def subworksheet_item(x) -> Tuple:
```
teetone, line 91: Can we specify the types for the Tuple elements?

codalab/lib/worksheet_util.py
```diff
@@ -264,13 +266,13 @@ def request_lines(worksheet_info):
     return form_result
 
 
-def parse_worksheet_form(form_result, model, user, worksheet_uuid):
+def parse_worksheet_form(form_result, model, user, worksheet_uuid) -> List:
```
teetone, line 269: Same with the `List`

codalab/lib/worksheet_util.py
```diff
@@ -483,27 +487,27 @@ def format_metadata(metadata):
...
         raise UsageError('Invalid number of arguments: %s' % (args,))
 
 
-def canonicalize_schema_items(items):
+def canonicalize_schema_items(items) -> List:
     return [canonicalize_schema_item(item) for item in items]
 
 
-def apply_func(func, arg):
+def apply_func(func, arg) -> Union[str, Dict, Tuple[Optional[Dict], ...]]:
```
teetone, line 510: When is `...` needed?

codalab/rest/interpret.py
```diff
@@ -52,7 +54,7 @@
 
 
 @post('/interpret/search', apply=ProtectedPlugin())
-def _interpret_search():
+def _interpret_search() -> worksheet:
```
teetone, line 57: Is this the correct type?

teetone (changes requested):

codalab/client/json_api_client.py
yiboliu: My plan is to go back to this after the general annotations are done.

teetone: It feels inconsistent. For some methods, it looks like you type annotated the data structure elements of the return types already (e.g. https://github.com/codalab/codalab-worksheets/pull/2796/files#diff-ec08b8c99e59740ca250510b62677643R516), but for some you didn't.

yiboliu: It was done in the previous stage, which was resolving errors.

codalab/bin/server.py
teetone: @yiboliu It might be too much to type annotate the function body. What do you think about the type annotating the function parameters in this PR, so the function definition is type annotated?

PR #2809, opened 2020-09-03
Fix public worker check

teetone:

If a public worker was not specified via `CODALAB_PUBLIC_WORKERS`, the previous logic would incorrectly always treat it as if there was one worker configured.

I also added a check to ensure the monitor runs start only if there are public workers.

### Related issues

Resolves https://github.com/codalab/deployment/issues/56.

### Checklist


monitor.py
```diff
@@ -258,10 +269,9 @@ def check_disk_space(paths):
 
 
 def poll_online_workers():
+    public_workers = get_public_workers()
     if len(public_workers) == 0:
-        error_logs(
-            'worker check failed', 'Missing value for environment variable CODALAB_PUBLIC_WORKERS.'
-        )
+        log("Environment variable CODALAB_PUBLIC_WORKERS was not configured.")
```
percyliang: Nit: Can we change this simply to CODALAB_PUBLIC_WORKERS is empty?  It might have been configured to be empty, technically.

percyliang (approved):
Thanks!

PR #2818, opened 2020-09-09
Update Interactive Session edit commands instructions

### LINKED ISSUE #2715: cl run --interactive part 2
Follow up of #2360 

Per @robinjia's feedback:
- Let the user know that deleting all the commands will not create a bundle
- Find an easy way to quit without saving

teetone:

Per Robin's feedback:

- Let the user know that deleting all the commands will not create a bundle
- Find an easy way to quit without creating a bundle

### Related issues

Resolves #2715 

### Checklist

### DIFF
--- a/codalab/lib/interactive_session.py
+++ b/codalab/lib/interactive_session.py
@@ -36,13 +36,9 @@ class InteractiveSession:
         "\n\n#\n"
         "# Choose the commands to use for cl run:\n"
         "#\n"
-        "# 1. Delete the lines of any unwanted commands. Removing all commands will not create a bundle.\n"
+        "# 1. Delete the lines of any unwanted commands. If you don't want to create the bundle, just remove all commands.\n"
         "# 2. Add any additional commands on a new line above these instructions.\n"
         "#\n"
-        "# To quit without creating a bundle, run:\n"
-        "#\n"
-        "#\t:%d | wq\n"
-        "#"
     )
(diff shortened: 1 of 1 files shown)


codalab/lib/interactive_session.py
```diff
@@ -36,9 +36,13 @@ class InteractiveSession:
         "\n\n#\n"
         "# Choose the commands to use for cl run:\n"
         "#\n"
-        "# 1. Delete the lines of any unwanted commands.\n"
+        "# 1. Delete the lines of any unwanted commands. Removing all commands will not create a bundle.\n"
         "# 2. Add any additional commands on a new line above these instructions.\n"
         "#\n"
+        "# To quit without creating a bundle, run:\n"
```
percyliang: I'd get rid of this since it's vi specific and and is subsumed by the first comment above (removing all commands).

codalab/lib/interactive_session.py
```diff
@@ -36,9 +36,13 @@ class InteractiveSession:
         "\n\n#\n"
         "# Choose the commands to use for cl run:\n"
         "#\n"
-        "# 1. Delete the lines of any unwanted commands.\n"
+        "# 1. Delete the lines of any unwanted commands. Removing all commands will not create a bundle.\n"
```
percyliang: Rephrase: "If you don't want to create the bundle, just remove all commands."

percyliang (approved):

PR #2835, opened 2020-09-14
Provide sticky header for result tables

### LINKED ISSUE #2082: [FEATURE REQUEST] Make it possible to lock column names when scrolling the results table
It's hard to understand which number corresponds to what when you have to scroll up and down to see the column names. 

Lock the column names row when scrolling

**Describe alternatives you've considered**
Any way to see which column is this without scrolling all the way up. Tooltip is a possibility but not ideal

**Additional context**

yuqijin:

Enable users to understand which number corresponds to without scrolling up and down to see the column names.
Lock the column names row when scrolling by using the 'sticky' position.

### Related issues

Address #2082 

### Screenshots

2082-sticky-header-result-table

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/TableItem.js
+++ b/frontend/src/components/worksheets/items/TableItem/TableItem.js
@@ -140,7 +140,21 @@ class TableItem extends React.Component<{
                     onMouseLeave={(e) => this.setState({ hovered: false })}
                     component='th'
                     key={index}
-                    style={index === 0 ? { paddingLeft: editPermission ? '30px' : '70px' } : {}}
+                    style={
+                        index === 0
+                            ? {
+                                  paddingLeft: editPermission ? '30px' : '70px',
+                                  position: 'sticky',
+                                  top: 0,
+                                  backgroundColor: '#F8F9FA',
+                                  zIndex: 1,
+                              }
+                            : {
+                                  position: 'sticky',
+                                  top: 0,
+                                  backgroundColor: '#F8F9FA',
+                              }
+                    }
                 >
                     {editPermission && index === 0 && (
                         <Tooltip title={'Change the schemas of this table'}>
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -140,7 +140,21 @@ class TableItem extends React.Component<{
...
+                        index === 0
+                            ? {
+                                  paddingLeft: editPermission ? '30px' : '70px',
+                                  position: 'sticky',
+                                  top: 0,
+                                  backgroundColor: '#F8F9FA',
+                                  zIndex: 1,
+                              }
+                            : {
+                                  position: 'sticky',
```
epicfaace, line 153: Why are we making all the other headers (even those at indices > 0) sticky as well?

yuqijin: I think the user requires to lock all the columns, such as uuid, name, epochs ...? So here if we only lock the header for whose index = 0, only the first column, i.e. the <uuid> column will be locked, and it will look like this:
Screen Shot 2020-09-14 at 6 06 00 PM. So that's why I make all the headers sticky.

epicfaace: Ah, I see index is for columns, not rows.

Could we just set zIndex to 1 for all the columns so that we can remove the conditional to simplify the code?

yuqijin: But besides the zIndex, we also need the conditional for deciding the `paddingLeft`? And it cannot be removed? (Cuz only the first column needs the left padding.)

epicfaace: Got it, makes sense, so we can't remove the conditional.

Why do we need to conditionally set the zIndex though?

yuqijin: I just set zIndex for all the columns. I previously conditionally set the Index because we only need to uplift the zIndex for the 1st column to make sure the 1st column covers the checkbox when scrolling. But I guess it would be good if we keep the zIndex for all the headers the same.

epicfaace (approved):

PR #2836, opened 2020-09-15
Fix stress test path

epicfaace:
Fix the stress test file path (which was changed in #2750) so that stress tests can work again.


yuqijin (approved):
Looks good!

PR #2844, opened 2020-09-15
Fix flake8 errors in codalab/

yiboliu:
Add fixing of errors in codalab/

### Related issues
#2660 

### Checklist


codalab/bundles/uploaded_bundle.py
```diff
@@ -26,8 +26,9 @@ class UploadedBundle(NamedBundle):
     )
     METADATA_SPECS.append(
         MetadataSpec('link_format', str, 'Link format of bundle. Can be equal to'
-            '"raw" or "zip" (only "raw" is supported as of now).', optional=True)
+                                         '"raw" or "zip" (only "raw" is supported as of now).', optional=True)
```
teetone, line 29: Nit: Is this spacing too much? Maybe one less tab?

codalab/lib/bundle_util.py
```diff
@@ -2,7 +2,10 @@
 import json
 
 from codalab.bundles import get_bundle_subclass
-from codalab.client.json_api_client import JsonApiClient, JsonApiRelationship
+from codalab.client.json_api_client import (
+    JsonApiClient,  # NOQA - JsonApiClient used in comments
```
teetone: Can you clarify what this means?

codalab/worker/bundle_service_client.py
```diff
@@ -97,12 +99,14 @@ def _authorize(self):
             'Content-Type': 'application/x-www-form-urlencoded',
             'X-Requested-With': 'XMLHttpRequest',
         }
-        request = urllib.request.Request(
+        the_request = urllib.request.Request(
```
teetone: Can we rename this variable? Maybe `request_to_send`?

setup.cfg
```diff
@@ -34,3 +34,22 @@ ignore_missing_imports = True
...
+# E731 - do not assign a lambda expression, use a def, doesn't apply to our code
+
+# F401 - module imported but not used
+
+# W503 - line break before binary operator, conflicting with black
+# W605 - invalid escape sequence, modifying will cause test failures
+
+ignore = E203,E712,E731,W503
+per-file-ignores =
+  #  JsonApiClient used in comments, black doesn't inline comments in that line
```
teetone: Can you clarify what this means?

setup.cfg
```diff
@@ -34,3 +34,22 @@ ignore_missing_imports = True
...
+  codalab/lib/bundle_util.py: F401,F841
+  #  Commented explicitly that we want to keep the imports.
+  codalab/server/rest_server.py: F401
```
teetone, line 51: Can you update the comment to `# Keep the unused imports as they are used to route the rest service`?

setup.cfg
```diff
@@ -34,3 +34,22 @@ ignore_missing_imports = True
 [flake8]
 max-line-length = 200
 exclude = venv/*,var/*
+
+# E203 - White space before ':', conflicting with black
+# E712 - comparison to False should be 'if cond is False:' or 'if not cond:', doesn't apply to our code
```
teetone: Why are we ignoring `E712`? It seems important to have.

https://www.flake8rules.com/rules/E712.html

setup.cfg
```diff
@@ -34,3 +34,22 @@ ignore_missing_imports = True
...
+# E731 - do not assign a lambda expression, use a def, doesn't apply to our code
+
+# F401 - module imported but not used
+
+# W503 - line break before binary operator, conflicting with black
```
teetone: nit: update lines 44 and 38 with `(conflicts with black)`

setup.cfg
```diff
@@ -34,3 +34,22 @@ ignore_missing_imports = True
 [flake8]
 max-line-length = 200
 exclude = venv/*,var/*
+
+# E203 - White space before ':', conflicting with black
+# E712 - comparison to False should be 'if cond is False:' or 'if not cond:', doesn't apply to our code
+# E731 - do not assign a lambda expression, use a def, doesn't apply to our code
```
teetone: Can you omit `doesn't apply to our code`?

codalab/bundles/uploaded_bundle.py
yiboliu: oh, this indentation is suggested by pycharm, to be at the same place of the beginning of former part.

setup.cfg
```diff
@@ -34,3 +34,22 @@ ignore_missing_imports = True
 [flake8]
 max-line-length = 200
 exclude = venv/*,var/*
+
```
teetone, line 37: Thanks for adding the error code comments. Can we organize the comments a bit better? Maybe something like this:

```
# Ignore completely:
# <flake8 code> - <description>
# <flake8 code> - <description>
#
# Ignore on a per file basis:
# <flake8 code> - <description>
# <flake8 code> - <description>
```

teetone (changes requested):

codalab/lib/bundle_util.py
yiboliu: This means asking flake8 to ignore errors in this line. It's complaining about JsonApiClient is not used, but it's actually used in comments so I suppress it.

setup.cfg
yiboliu: The inline above comments was split into multiple lines then purpose of suppressing flake8 errors isn't served. So I do it here again. But would like to leave the inline comments there so other people can be aware of.

setup.cfg
yiboliu: This suggestion doesn't work for our code. Changing them could cause errors.

setup.cfg
teetone: Can you share the links to the lines that are problematic? I will take a look.

I think it's better to ignore a line or two within the file instead of ignoring a rule for the whole file.

setup.cfg
yiboliu: done

setup.cfg
yiboliu: done

setup.cfg
yiboliu: done

setup.cfg
yiboliu: done

codalab/lib/bundle_util.py
teetone: According to the flake8 [documentation](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html), should it be `# noqa: <error code>`?

codalab/worker/bundle_service_client.py
yiboliu: done

setup.cfg
yiboliu: The module is imported in this [line](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_util.py#L5), and it's used [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_util.py#L56).

yiboliu: The problem is if I add the inline comments to ignore it, then the black will split my comments to the following lines and that will make it won't work.

codalab/lib/bundle_util.py
yiboliu: Another thing I can try would be to just leave the comment as `# noqa: F401` but can't add the description/explanation on the same line, have to do it in another line. How does that sound?

teetone: Yeah I think you can just add a comment before.

setup.cfg
teetone: Yeah we shouldn't ignore a rule for an entire file for just that one line.

setup.cfg
teetone: I think that's a problem with our code then and it should be fixed.

yiboliu:
```
>>> None is False
False
>>> None == False
False
>>> not None
True
>>> 0 == False
True
>>> 0 is False
False
```
This is the case that the suggestion doesn't work well.

codalab/lib/crypt_util.py
```diff
@@ -186,8 +186,8 @@ def pbkdf2(password, salt, iterations, dklen=0, digest=None):
             dklen = hlen
         if dklen > (2 ** 32 - 1) * hlen:
             raise OverflowError('dklen too big')
-        l = -(-dklen // hlen)
-        r = dklen - (l - 1) * hlen
+        remainder = -(-dklen // hlen)
```
teetone: `//` is floor division, so it doesn't return the remainder.

codalab/lib/worksheet_util.py
```diff
@@ -64,11 +64,15 @@
 WORKSHEET_ITEM_TYPES = (TYPE_MARKUP, TYPE_DIRECTIVE, TYPE_BUNDLE, TYPE_WORKSHEET)
 
 
-BUNDLE_REGEX = re.compile('^\s*(\[(.*)\])?\s*\{([^{]*)\}\s*$')
-SUBWORKSHEET_REGEX = re.compile('^\s*(\[(.*)\])?\s*\{\{(.*)\}\}\s*$')
+BUNDLE_REGEX = re.compile(
+    '^\s*(\\[(.*)\\])?\s*\\{([^{]*)\\}\s*$'  # NOQA - modifying escapes could create end-to-end test failure
```
teetone: I think this is wrong formatting according to flake8. See my comment [here](https://github.com/codalab/codalab-worksheets/pull/2844#discussion_r495536795).

codalab/model/bundle_model.py
```diff
@@ -848,7 +848,10 @@ def transition_bundle_starting(self, bundle, user_id, worker_id):
             # Check if the designated worker is going to be terminated soon
             row = connection.execute(
                 cl_worker.select().where(
-                    and_(cl_worker.c.worker_id == worker_id, cl_worker.c.is_terminating == False)
+                    and_(
+                        cl_worker.c.worker_id == worker_id,
+                        cl_worker.c.is_terminating == False,  # NOQA
```
teetone: Same with this one. There is no error code and comment explaining why we are ignoring the rule.

codalab/rest/interpret.py
```diff
@@ -68,7 +62,7 @@ def _interpret_search():
     }
     ```
     """
-    return interpret_search(request.json)
```
teetone: Why is this being modified?

codalab/worker/docker_utils.py
```diff
@@ -26,13 +26,13 @@
 
 # This specific error happens when a user specifies an image with an incompatible version of CUDA
 NVIDIA_MOUNT_ERROR_REGEX = (
-    '[\s\S]*OCI runtime create failed[\s\S]*stderr:[\s\S]*nvidia-container-cli: '
-    'mount error: file creation failed:[\s\S]*nvidia-smi'
-)
+    '[\s\S]*OCI runtime create failed[\s\S]*stderr:[\s\S]*nvidia-container-cli: '  # NOQA
+    'mount error: file creation failed:[\s\S]*nvidia-smi'  # NOQA
+)  # Modifying escapes here could cause unit test failure.
```
teetone: Is this causing failures in `docker_utils_test.py`? I think we should fix this. It seems a bit odd that we are modifying  `MEMORY_LIMIT_ERROR_REGEX`, but not `NVIDIA_MOUNT_ERROR_REGEX`.

setup.cfg
```diff
@@ -34,7 +34,18 @@ ignore_missing_imports = True
...
+# Ignore completely:
+# E203 - White space before ':', (conflicts with black)
+# E731 - do not assign a lambda expression, use a def
+#
```
teetone, line 41: nit: can we get rid of this extra line here and on line 43?

codalab/rest/interpret.py
yiboliu: I can't find any method with name interpret_search, and both PyCharm and Flake8 are complaining.

codalab/worker/docker_utils.py
yiboliu: Not really. It's always the end-to-end test is failing. I checked all the log files in the failing PR. The only message I got is `400 Client Error: Bad Request ("Unknown runtime specified nvidia")`, which matches none of these regexes.

teetone: Which unit test is failing?

yiboliu: Oh it's not unit test. It's end-to-end(frontend) test. Like what I ran here: https://github.com/codalab/codalab-worksheets/actions/runs/281351629

codalab/lib/worksheet_util.py
```diff
@@ -547,7 +547,7 @@ def apply_func(func, arg):
                 t = tokens[2].replace(esc_slash, '/')
                 arg = re.sub(s, t, arg)
             elif f.startswith('['):  # substring
-                m = re.match('\[(.*):(.*)\]', f)
+                m = re.match('\\[(.*):(.*)\\]', f)
```
teetone: Can you undo this change for now?

teetone (approved):

PR #2866, opened 2020-09-21
Remove componentWillMount and componentWillReceiveProps in FileBrowser

### LINKED ISSUE #2616: Remove usage of componentWillMount
We use componentWillMount only in two spots of our code -- we should remove this & replace with a better solution, since componentWillMount is deprecated in React (and will eventually be removed from React).

yuqijin:

componentWillMount is deprecated in React, and we only used componentWillMount in FileBrowser. I moved the original code in `componentWillMount` to `componentDidMount` and `constructor`.

### Related issues

fix #2616 

### Screenshots

2616-willMount
Everything works well.

### Checklist

### DIFF
--- a/frontend/src/components/FileBrowser/FileBrowser.js
+++ b/frontend/src/components/FileBrowser/FileBrowser.js
@@ -50,16 +50,18 @@ export class FileBrowser extends React.Component<
         }
     }
 
-    componentWillReceiveProps(nextProps) {
-        if (nextProps.isRunBundleUIVisible === false && this.state.isVisible) {
-            this.setState({ isVisible: false });
-            this.getDOMNode()
-                .getElementsByClassName('file-browser-arrow')[0]
-                .click();
+    static getDerivedStateFromProps(nextProps, prevState) {
+        if (nextProps.isRunBundleUIVisible === false && prevState.isVisible) {
+            return { isVisible: false };
         }
     }
 
     componentDidUpdate(prevProps, prevState) {
+        if (this.props.isRunBundleUIVisible === false && prevState.isVisible) {
+            this.getDOMNode()
+                .getElementsByClassName('file-browser-arrow')[0]
+                .click();
+        }
         if (prevProps.uuid !== this.props.uuid) {
             // Reset and fire off an asynchronous fetch for new data
             this.setState({ fileBrowserData: {} });
--- a/frontend/src/components/FileBrowser/FileBrowser.js
+++ b/frontend/src/components/FileBrowser/FileBrowser.js
@@ -40,8 +40,14 @@ export class FileBrowser extends React.Component<
         this.state = {
             currentWorkingDirectory: '',
             fileBrowserData: {},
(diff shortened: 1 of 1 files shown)


frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -470,6 +469,7 @@ export class FileBrowserLite extends React.Component<
         this.state = {
             currentWorkingDirectory: '',
             fileBrowserData: {},
+            isVisible: !this.props.startCollapsed,
```
yipenghe: This state doesn't seem to be in use anymore, let's remove it

frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -40,10 +40,16 @@ export class FileBrowser extends React.Component<
...
         };
     }
 
+    componentDidMount() {
+        if (!this.props.startCollapsed) {
+            this.updateFileBrowser('');
+        }
+    }
+
     componentWillReceiveProps(nextProps) {
```
yipenghe: This life cycle function (https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) is actually going to be deprecated as well, maybe it would be good to refactor and remove this function as well in this PR

frontend/src/components/FileBrowser/FileBrowser.js
yuqijin: Removed the `isvisible` in FileBrowserLite

frontend/src/components/FileBrowser/FileBrowser.js
yuqijin: Removed commponentWillReceiveProps.

frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -40,34 +40,35 @@ export class FileBrowser extends React.Component<
...
-            this.getDOMNode()
-                .getElementsByClassName('file-browser-arrow')[0]
-                .click();
+    componentDidMount() {
+        if (!this.props.startCollapsed) {
+            this.updateFileBrowser('');
+        }
+    }
+
+    static getDerivedStateFromProps(nextProps, prevState) {
```
yipenghe, line 53: I'm assuming you referred to this article for the replacement? https://medium.com/hackernoon/replacing-componentwillreceiveprops-with-getderivedstatefromprops-c3956f7ce607
just adding comment here in case in the future someone needs context

yipenghe (approved):

yipenghe:
#m lgtm

frontend/src/components/FileBrowser/FileBrowser.js
yuqijin: Actually I referred to the official doc: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html 
But thanks for putting the reference here!

PR #2867, opened 2020-09-21
Fix the bug of upload not releasing keyboard shortcuts

### LINKED ISSUE #2364: upload doesn't release keyboard shortcuts
On a worksheet, hit 'a u' to trigger upload, select a file.  After the upload hitting 'enter' on the upload bundle triggers another upload popup rather than bundle details.

yuqijin:

On a worksheet, hit 'a u' to trigger upload, select by press 'enter', but two upload menus will appear consecutively. 
The bug results from that `stopPropagation` allows other handlers on the same element to be executed, so another hidden upload menu will also be clicked when we press 'enter' later, and thus two menus appear consecutively, so I use `stopImmediatePropagation` instead, and it prevents every event from running.

### Related issues

fix #2364 

### Screenshots

2364-upload
Works normally now.

### Checklist


yipenghe:
The bug seems to still appear when using click to open the upload menu and use keyboard to select upload file and enter for confirming

yuqijin:
> The bug seems to still appear when using click to open the upload menu and use keyboard to select upload file and enter for confirming

Fix the bug in this workflow by adding `preventDefault()` to the `enter` button.

yipenghe (approved):

yipenghe:
Looks good to me!

PR #2877, opened 2020-09-22
Fix bug of 'g g' not scrolling to top

### LINKED ISSUE #2874: gg should scroll to the top
Hitting 'g g' currently sets the cursor to the right place (at the top), but it doesn't scroll there anymore, which is bad.

yuqijin:

fix gg shortcut

### Related issues

fix #2874 

### Screenshots

2874-gg

### Checklist

### DIFF
--- a/frontend/src/__tests__/__snapshots__/Worksheet.test.js.snap
+++ b/frontend/src/__tests__/__snapshots__/Worksheet.test.js.snap
@@ -542,6 +542,7 @@ Object {
             >
               <div
                 class="Worksheet-worksheetDummyHeader-4"
+                id="worksheet_header"
               />
               <div
                 class="Worksheet-worksheetInner-3"
@@ -1304,6 +1305,7 @@ Object {
           >
             <div
               class="Worksheet-worksheetDummyHeader-4"
+              id="worksheet_header"
             />
             <div
               class="Worksheet-worksheetInner-3"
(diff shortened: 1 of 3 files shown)


frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -135,7 +135,7 @@ class WorksheetItemList extends React.Component {
                 $('body')
                     .stop(true)
                     .animate({ scrollTop: 0 }, 'fast');
-                this.props.setFocus(-1, 0);
+                this.props.setFocus(0, 0);
```
percyliang: We actually want the cursor to be at the "-1" position though, before any bundles.

yipenghe (commented):
The setFocus should be -1 as described in the comment near line 130. The root cause is here:
https://github.com/codalab/codalab-worksheets/blob/master/frontend/src/components/worksheets/Worksheet/Worksheet.js#L666

The node here will be null (you can console.log the node) when we set the focus above the -1 item, since there is no item with id `codalab-worksheet-item--1`. A possible fix will to manually give the -1 item the id  codalab-worksheet-item--1 or some other special id, and find the item using that id when the passed in index argument is -1

frontend/src/__tests__/__snapshots__/Worksheet.test.js.snap
```diff
@@ -542,6 +542,7 @@ Object {
             >
               <div
                 class="Worksheet-worksheetDummyHeader-4"
+                id="worksheet_header"
```
yipenghe, line 545: let's use worksheet_dummy_header for consistency

yipenghe (approved):
Works well!

PR #2879, opened 2020-09-23
Better style of bundle table header

### LINKED ISSUE #2397: style the bundle table headers better
The tables on the worksheet have headers which are too light and too tall, which makes it hard to read.
Screenshot from 2020-06-14 14-04-56
Make the header slightly darker and the same height as regular rows.

yuqijin:

Make the header in black color and the same height as regular rows.

### Related issues

fix #2397 

### Screenshots

2397-table-header-style

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/TableItem.js
+++ b/frontend/src/components/worksheets/items/TableItem/TableItem.js
@@ -148,12 +148,20 @@ class TableItem extends React.Component<{
                                   top: 0,
                                   backgroundColor: '#F8F9FA',
                                   zIndex: 1,
+                                  color: '#000000',
+                                  paddingBottom: 0,
+                                  paddingTop: 0,
+                                  height: 26,
                               }
                             : {
                                   position: 'sticky',
                                   top: 0,
                                   backgroundColor: '#F8F9FA',
                                   zIndex: 1,
+                                  color: '#000000',
+                                  paddingBottom: 0,
+                                  paddingTop: 0,
+                                  height: 26,
                               }
                     }
                 >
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -148,20 +148,28 @@ class TableItem extends React.Component<{
                                   top: 0,
```
yipenghe, line 148: Since the two conditions are identical except for paddingLeft, maybe consider doing something similar to https://github.com/codalab/codalab-worksheets/blob/master/frontend/src/components/worksheets/items/TableItem/BundleRow.js#L192-L197, so we won't need to update two places in the future

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -148,20 +148,28 @@ class TableItem extends React.Component<{
...
                                   backgroundColor: '#F8F9FA',
                                   zIndex: 1,
+                                  color: '#000000',
```
yipenghe: Since the two conditions are identical except for paddingLeft, maybe consider doing something similar to https://github.com/codalab/codalab-worksheets/blob/master/frontend/src/components/worksheets/items/TableItem/BundleRow.js#L192-L197, so we won't need to update two places in the future

yuqijin: Take your advice and refactor the code!

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -140,28 +140,26 @@ class TableItem extends React.Component<{
...
-                                  top: 0,
-                                  backgroundColor: '#F8F9FA',
-                                  zIndex: 1,
-                              }
-                            : {
-                                  position: 'sticky',
-                                  top: 0,
-                                  backgroundColor: '#F8F9FA',
-                                  zIndex: 1,
+                                  paddingLeft: editPermission ? '30px' : '40px',
```
yipenghe, line 151: We can also make this a class and do something like [classes.firstColumnHeader]: index === 0 and not use style here at all

yipenghe (approved):
<strike>Feel free to decide whether to adopt my suggestion above</strike> Never mind, looks good to me

PR #2889, opened 2020-09-24
Optionally enable sentry on AWS batch workermanager-launched workers

nelson-liu:
This just exports the right environment variable in the container, if we're using sentry on the host that's running the WorkerManager.


codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -171,6 +173,11 @@ def start_worker_job(self):
                 {'sourceVolume': 'shared_dir', 'containerPath': bundle_mount, 'readOnly': False}
             )
 
+        if using_sentry:
```
teetone: Should this be `using_sentry()`?

nelson-liu: yup, great catch

teetone (approved):

PR #2892, opened 2020-09-24
Add urlopen with retry

teetone:

Added a retry mechanism for `urlopen` to help avoid failures from a single request attempt.

### Related issues

Resolves #2892 

### Checklist


epicfaace (commented):
This looks good -- what do you think of the alternative of just replacing calls to urlopen to `requests` (which uses retries automatically)?

nelson-liu:
> This looks good -- what do you think of the alternative of just replacing calls to urlopen to requests (which uses retries automatically)?

+1 for just using requests

teetone:
> This looks good -- what do you think of the alternative of just replacing calls to urlopen to `requests` (which uses retries automatically)?

Correct me if I'm wrong, but would that require replacing `urllib` with `Requests` and rewriting `rest_client.py` and it's subclasses?

epicfaace:
Yeah, I guess that might be beyond the scope of this PR. We could shelve that as another issue and continue ahead with this approach for now.

yiboliu (approved):
LGTM

codalab/common.py
```diff
@@ -116,3 +120,16 @@ def ensure_str(response):
         return response.decode()
     except UnicodeDecodeError:
         return BINARY_PLACEHOLDER
+
+
+@retry(urllib.error.URLError, tries=2, delay=1, backoff=2)
+def urlopen_with_retry(request: urllib.request.Request, timeout: int = URLOPEN_TIMEOUT_SECONDS):
+    """
+    Makes a request using urlopen with a timeout of 60 seconds and retries on failures.
```
epicfaace:
```suggestion
    Makes a request using urlopen with a timeout of URLOPEN_TIMEOUT_SECONDS seconds and retries on failures.
```

epicfaace (approved):

PR #2894, opened 2020-09-25
Put the hostname in the remote: field for workermanager-launched jobs

nelson-liu:
Right now, the workermanager launches jobs that have a remote: field of something like `nfliu-codalab-slurm-worker-443cb25d`. This is annoying if you want to figure out the hostname of the node that the job ran on.

This changes the `id` field to `{shorthostname}-{workerid}`, so the remote field would be something like `jagupard12-nfliu-codalab-slurm-worker-443cb25d`

### DIFF
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -94,7 +94,7 @@ class WorkerManager(object):
             '--work-dir',
             work_dir,
             '--id',
-            worker_id,
+            f'$(hostname -s)-{worker_id}',
             '--network-prefix',
             'cl_worker_{}_network'.format(worker_id),
         ]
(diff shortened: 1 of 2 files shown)


nelson-liu:
Ok, I just tested this and verified that it works properly.

Here's a bundle https://codalab.stanford.edu/bundles/0x73638f3dab90415a98657eae35fe9e2e , the remote is `john1-nfliu-codalab-slurm-worker-e4489f45`. Similarly, cl workers shows me:
```
john1-nfliu-codalab-slurm-worker-e4489f45                0/6   0/0   125g    13.7g      4.0s ago             debug                  False               False          999999999            False
```

teetone (approved):

PR #2897, opened 2020-09-25
Make AWSBatchWorkerManager boto3 warning stricter and more robust

nelson-liu:
A few changes here:

1. Since boto3 is imported at the top of the file, you can't import the AWSBatchWorkerManager unless you can import boto3. This is bad, because `codalab/worker_manager/main.py` imports `AWSBatchWorkerManager` simply to access `AWSBatchWorkerManager.DESCRIPTION`. As a result, you can't run the `slurm-batch` worker manager unless you have boto3, since the `AWSBatchWorkerManager` is imported along with it. So, I made the boto3 import lazy
2. Previously, the warning had a hardcoded boto3 version number. This changes it to be a bit more robust, since it will now read the dependency from `requirements-server.txt`


codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -60,6 +69,17 @@ def add_arguments_to_subparser(subparser):
 
     def __init__(self, args):
         super().__init__(args)
+        # We import this lazily, so a user doesn't have to install boto3 unless
+        # they absolutely want to run the AWS worker manager (versus if it's incidentally)
```
epicfaace:
```suggestion
        # they absolutely want to run the AWS worker manager, versus if it's incidentally
```

codalab/worker_manager/aws_batch_worker_manager.py
```diff
@@ -60,6 +69,17 @@ def add_arguments_to_subparser(subparser):
...
+        # imported by other code (e.g., to access AWSBatchWorkerManager.DESCRIPTION , as done
+        # in codalab/worker_manager/main.py ).
+        try:
+            import boto3
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "Running the AWS worker manager requires the boto3 module.\n"
+                f"Please run: pip install {get_boto3_pip_version()}"
```
epicfaace: I feel like using `get_boto3_pip_version` is adding a lot of complexity, and I don't think we have tests that catch this (so, if we rename requirements-server.txt, the AWSBatchWorkerManager will silently fail). Moreover, it appears that boto3 hasn't ever released any breaking changes. Could we just say "Please run: pip install boto3" instead?

nelson-liu: good idea, i'll make this change

epicfaace (approved):

epicfaace (approved):

PR #2904, opened 2020-09-27
Make worker timeout and check-in frequency configurable

### LINKED ISSUE #2891: Make worker check-in frequency and worker timeout configurable
**Describe the bug**
- Add a delay for worker check-in, which can be specified when creating a CodaLab worker
- Make `WORKER_TIMEOUT` in `bundle_manager.py` configurable

teetone:

- Added `--bundle_manager_worker_timeout_seconds` to `codalab_service`
- Added `--checkin-frequency-seconds` to workers
- Added `--worker-checkin-frequency-seconds` to worker managers

### Related issues

Resolves #2891 

### Checklist


codalab/server/bundle_manager.py
```diff
@@ -42,7 +41,7 @@ class BundleManager(object):
     Assigns run bundles to workers and makes make bundles.
     """
 
-    def __init__(self, codalab_manager):
+    def __init__(self, codalab_manager, worker_timeout_seconds):
```
epicfaace: Can we make this argument optional? It would make it a bit easier for test setup, so we aren't forced to specify `worker_timeout_seconds`.

teetone: I set it to the previous default of 60 seconds.

epicfaace (approved):

PR #2907, opened 2020-09-29
Remove scrollbar in table

### LINKED ISSUE #2873: scrolling using keyboard shortcuts is
When you hit 'j' or 'k' to scroll down/up on a long worksheet, the worksheet scrolls with each key press, which leads to jerky behavior.  The desired behavior (as it was before) is to scroll only when the selected bundle cursor is near the edge and scroll a more significant amount so that scrolling happens less frequently.

### LINKED ISSUE #2875: tables shouldn't have scrollbars
There should only be one scrollbar for the worksheet, but now for some reason, tables have scroll bars too, which makes it very confusing.
Screenshot from 2020-09-22 01-17-28

yuqijin:

Remove the scrollbar in the table. The table cannot be scrolled anymore.

### Related issues

fixes #2875 ; reopens #2082

### Screenshots
No scroll bar: 
Screen Shot 2020-10-03 at 3 01 55 PM

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/TableItem.js
+++ b/frontend/src/components/worksheets/items/TableItem/TableItem.js
@@ -145,15 +145,7 @@ class TableItem extends React.Component<{
                             [classes.tableHeader]: true,
                         }),
                     }}
-                    style={
-                        index === 0
-                            ? {
-                                  paddingLeft: editPermission ? '30px' : '40px',
-                                  paddingBottom: 0,
-                                  paddingTop: 0,
-                              }
-                            : {}
-                    }
+                    style={index === 0 ? { paddingLeft: editPermission ? '30px' : '70px' } : {}}
                 >
                     {editPermission && index === 0 && (
                         <Tooltip title={'Change the schemas of this table'}>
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -328,6 +333,11 @@ class _TableContainer extends React.Component {
 const styles = (theme) => ({
     tableContainer: {
         position: 'relative',
+        '&::-webkit-scrollbar': {
```
epicfaace: Why do this instead of `overflow: hidden`?

yuqijin: I just switched the method! Thanks for your advice.

epicfaace: Is there a chance we might cut off text in the table, this way? Or is that not going to be an issue?

yuqijin: I think you can scroll left and right to see the text if it is overflow. In the gif you can see no text is cut off.

yuqijin: Screen Shot 2020-09-30 at 2 07 52 PM

yuqijin: It is proved that the `overflow` method will cut off the text. So I switched back to the specific webkit-scrollbar method.

epicfaace (approved):

PR #2911, opened 2020-09-30
Fix 'j  k' keyboard shortcut for scroll up and down

### LINKED ISSUE #2873: scrolling using keyboard shortcuts is

yuqijin:
To provide the desired behavior of scrolling only when the selected bundle cursor is near the edge and scroll a more significant amount so that scrolling happens less frequently. So now, first check whether an element is visible or not, then scroll.

Ref: https://stackoverflow.com/questions/487073/how-to-check-if-element-is-visible-after-scrolling

### Related issues

fixes: #2873 

### Screenshots

Homepage:
2873-homescroll

Regular Worksheet:

2873-tablescroll

### Checklist


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -598,6 +598,33 @@ class Worksheet extends React.Component {
...
+        // Make the judgement only if the shouldScroll parameter has the default value of `True`
+        // Logic: Only scroll when the target element is completely not on the screen
+        if (shouldScroll) {
+            const element = subIndex
+                ? $(`#codalab-worksheet-item-${index}-subitem-${subIndex}`)
+                : $(`#codalab-worksheet-item-${index}`);
+
+            function isOnScreen(element) {
+                if (element.offset() === undefined) return false;
+                var elementOffsetTop = element.offset().top;
```
yipenghe: let's use `let` instead of `var`

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -598,6 +598,33 @@ class Worksheet extends React.Component {
             return;
         }
         const item = this.refs.list.refs['item' + index];
+
+        // Make the judgement only if the shouldScroll parameter has the default value of `True`
+        // Logic: Only scroll when the target element is completely not on the screen
```
epicfaace: Can you mention a bit more in the comment about what purpose this serves? (i.e., making sure that the screen doesn't scroll when you normally press j / k, until the target element is completely not on the screen)

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -598,6 +598,33 @@ class Worksheet extends React.Component {
...
+        if (shouldScroll) {
+            const element = subIndex
+                ? $(`#codalab-worksheet-item-${index}-subitem-${subIndex}`)
+                : $(`#codalab-worksheet-item-${index}`);
+
+            function isOnScreen(element) {
+                if (element.offset() === undefined) return false;
+                var elementOffsetTop = element.offset().top;
+                // For result table which should first scroll itself when a row inside it is not visible anymore.
+                // The max height for the result table is 500.
```
epicfaace: Is there a way to factor out "500" as a constant, so we can easily change it later without having to go through this code?

yipenghe: I guess (<= 262 ||   >= 262 + MAX_TABLE_HEIGHT) would do (line 614 below)

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Have made the modification!

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Changed to `let`. Thanks!

epicfaace (approved):

PR #2912, opened 2020-09-30
Fix frontend npm build warnings

### LINKED ISSUE #1094: Fix frontend npm build warnings

yuqijin:

Fix all existing frontend npm build warnings.
Warnings include:
- unused vars
- direct mutation of state
- shadowing of global property `undefined`
- unsafe references in Function declared in a loop 
- fake links that should really just be buttons

### Related issues

fixes #1094 

### Screenshots

No more npm warning when running `pre-commit.sh`:

1094-var

Use `<button>` to replace `<a>`:

1094-button

### Checklist

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+    "extends": "react-app",
+    "rules": {
+    }
+}
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,5 +1,0 @@
-{
-    "extends": "react-app",
-    "rules": {
-    }
-}
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,9 +1,5 @@
 {
     "extends": "react-app",
     "rules": {
-        "no-loop-func": "off",
-        "jsx-a11y/anchor-is-valid": "off",
-        "no-shadow-restricted-names": "off",
-        "react/no-direct-mutation-state": "off"
     }
 }
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,7 +1,6 @@
 {
     "extends": "react-app",
     "rules": {
-        "no-unused-vars": "off",
         "no-loop-func": "off",
         "jsx-a11y/anchor-is-valid": "off",
         "no-shadow-restricted-names": "off",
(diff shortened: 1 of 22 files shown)


frontend/.eslintrc.json
```diff
@@ -1,10 +1,5 @@
 {
     "extends": "react-app",
     "rules": {
```
epicfaace: I think you could probably remove this file entirely, since it's no longer overriding anything anymore.

yuqijin: I guess maybe we could not remove the configuration file, since it might cause the error: 
Screen Shot 2020-10-01 at 4 08 30 PM

epicfaace: Oh, ok, got it

epicfaace (approved):
Great work!

frontend/src/util/ws_mousetrap_fork.js
```diff
@@ -20,7 +20,7 @@
  * @version 1.6.3
  * @url craig.is/killing/mice
  */
-(function(window, document, undefined) {
+(function(window, document) {
```
epicfaace: I think this change broke something. Now, if I use j / k to scroll down a worksheet list (subworksheets block) and press enter on one of them, I get this error:

image


It's better to just leave this file as-is and then just include ws_mousetrap_fork in the eslintrc so it's ignored by eslint, since this is external code, anyway.

epicfaace (changes requested):

frontend/src/util/ws_mousetrap_fork.js
yuqijin: Thanks for pointing it out. Just updated it.

epicfaace (approved):

frontend/src/util/ws_mousetrap_fork.js
epicfaace: Actually, this error appears to be happening on master, and is unrelated to these changes, so will make a new issue for it (#2919).

PR #2941, opened 2020-10-06
Better style of the schema blocks

### LINKED ISSUE #2776: pile of schemas rendering is awkward
Now that we render schema definitions (` % schema ...`), they kind of pile up at the beginning of a worksheet, which is ugly:
Screenshot from 2020-08-23 21-17-50

Options:
1) Reduce the amount of vertical space taken by each block.  We'd probably have to make this really thin with no spacing between schema blocks. 
2) Have an explicit verbose option...this might lead to complexity and be confusing.

yuqijin:

Provide a better style for a pile of schemas rendering

### Related issues

fixes #2776 

### Screenshots

Previous:
2776-before

Now:
2776-after

### Checklist


percyliang:
Thanks, this looks better, but I think it might not be enough to be non-intrusive.  I wonder if making it a thin bar with a tooltip might be good, though that still might look ugly.  Clearly the least ugly thing would be to hide it and have a mode to show it, but that introduces other complications...  What do you think?

yuqijin:
> Thanks, this looks better, but I think it might not be enough to be non-intrusive. I wonder if making it a thin bar with a tooltip might be good, though that still might look ugly. Clearly the least ugly thing would be to hide it and have a mode to show it, but that introduces other complications... What do you think?

Oh, I see. I am worried whether it will be a bit confusing for the user if we used a thin bar with a tooltip. So maybe I could try the option2 to implement some verbose mode to display the schemas. (Though now I am not quite sure about the complexity.)

percyliang:
Agree the bar might look mysterious/confusing, though maybe something can show up on mouse over?  Or the icon were just smaller?  I think we can mock out a few options against the actual worksheets (look at the worksheets in Gallery and check that they seem reasonable)?  On second thought, maybe we should hold off on the verbose option since we already have a lot of bugs in the frontend, and so I'd be reluctant to add even more complexity.

yuqijin:
> Agree the bar might look mysterious/confusing, though maybe something can show up on mouse over? Or the icon were just smaller? I think we can mock out a few options against the actual worksheets (look at the worksheets in Gallery and check that they seem reasonable)? On second thought, maybe we should hold off on the verbose option since we already have a lot of bugs in the frontend, and so I'd be reluctant to add even more complexity.

I see! Then I think I could start with the thin bar option (I think one of the reasons why the piles of schemas look ugly is there is too much space on the right side of the schema button, so maybe a thin bar will look better.

yuqijin:
@percyliang Hi Percy, I have tried several options for different styles of the schema button, and here are them:

Thin bar w/ icon: It is intuitive to the user,  but may look a bit squeezing
icon

Thin bar w/ arrow: It only contains the arrow.
arrow

Thin bar only: Maybe it is too abstract to the user. 
none

Do you think any of them is good enough? Thanks!

percyliang:
Thanks for the different options.  Maybe we can just keep the first one, but left justify the icons?  So the only difference between what we have right now is that we make the bar longer.  And then making things take less vertical space.

yuqijin:
> Thanks for the different options. Maybe we can just keep the first one, but left justify the icons? So the only difference between what we have right now is that we make the bar longer. And then making things take less vertical space.

Thank you for your advice! I left justify the icons.
2776-2nd

yuqijin:
Current style: 
icon

epicfaace (approved):

epicfaace:
@yuqijin can you fix merge conflicts?

yuqijin:
> @yuqijin can you fix merge conflicts?

The conflict makes the button style changes a lot. I am still working on it.

Issue #2665, opened 2020-07-28
Test "cl work" command in test_cli.py


epicfaace:
We need to add tests involving running `cl work [domainname]::` in test_cli.py. Right now, we only catch errors in this command in the stress tests, not in the actual CLI tests.

yiboliu:
Add a test module to try running `cl work [domainname]::` see if it succeeds. Need more info about what domainname needs to be. starting with `localhost`

yiboliu:
localhost is finished. If more domain name checkings are needed, I'll reopen the issue.

PR #2942, opened 2020-10-07
test-cli

yiboliu:
Add tests for `cl work [domainname]::`

### Related issues
#2665 

### Checklist


tests/cli/test_cli.py
```diff
@@ -25,7 +25,7 @@
 from codalab.worker.download_util import BundleTarget
 from codalab.worker.bundle_state import State
 from scripts.create_sample_worksheet import SampleWorksheet
-from scripts.test_util import Colorizer, run_command
+from scripts.test_util import Colorizer, run_command, cleanup
```
teetone: Is this an unused import?

teetone (approved):

tests/cli/test_cli.py
yiboliu: oh yes, I'll remove it. Thanks

tests/cli/test_cli.py
```diff
@@ -560,6 +560,12 @@ def run(cls, tests, instance, second_instance):
 ############################################################
 
 
+@TestModule.register('unittest')
```
epicfaace, line 563: @yiboliu you have to give the TestModule.register a different name, I think, for it to actually be run in the tests.

PR #2946, opened 2020-10-09
Azure Batch Worker Manager

### LINKED ISSUE #1237: Support Azure batch workers

teetone:

- Added the new Azure Batch-based Worker Manager
- Filter out bundles based on the resources specified when creating a Worker Manager
- Add the ability to start the Azure Batch Worker Manager via Docker Compose

### Related issues

Resolves #1237 

### Checklist


codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Create a job using the pool
+        job_id = 'azure-{}-{}'.format(self.args.job_name, worker_id)
+        job = batch.models.JobAddParameter(
+            id=job_id,
+            display_name=self.args.job_name,
+            pool_info=batch.models.PoolInformation(pool_id=self.args.pool_id),
+            common_environment_settings=[
+                batch.models.EnvironmentSetting(
+                    name='CODALAB_USERNAME', value=os.environ.get('CODALAB_USERNAME')
```
epicfaace: Will these variables always be set?

teetone: For our use case now, it will always be set.

codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class AzureBatchWorkerManager(WorkerManager):
+    NAME: str = 'azure-batch'
+    DESCRIPTION: str = 'Worker manager for submitting jobs to Azure Batch'
+
+    @staticmethod
+    def add_arguments_to_subparser(subparser: ArgumentParser) -> None:
+        subparser.add_argument(
+            '--account-name', type=str, help='Azure Batch account name',
```
nelson-liu: do these need to be `required`?

codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -0,0 +1,152 @@
...
+
+        task_container_run_options: List[str] = [
+            '--cpus %d' % self.args.cpus,
+            '--memory %dM' % self.args.memory_mb,
+            '--volume /var/run/docker.sock:/var/run/docker.sock',
+            '--volume %s:%s' % (work_dir, work_dir),
+            '--user %s' % self.args.user,
+        ]
+
+        if os.environ.get('CODALAB_USERNAME') and os.environ.get('CODALAB_PASSWORD'):
```
nelson-liu, line 106: If this `if` block doesn't run, the worker won't be able to connect, right? Can we make it error if these environment variables aren't set?
(The AWSBatchWorkerManager errors because job definition creation fails in the case that these variables are None, but it could probably be more explicit there as well)

codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -0,0 +1,152 @@
...
+                '--volume shared_dir:%s' % os.environ.get('CODALAB_BUNDLE_MOUNT')
+            )
+
+        # Configure Sentry
+        if using_sentry():
+            task_container_run_options.append(
+                '--env CODALAB_SENTRY_INGEST_URL=%s' % CODALAB_SENTRY_INGEST
+            )
+
+        command_line: str = "/bin/sh -c '{}'".format(' '.join(command))
```
nelson-liu: /bin/bash for parity with AWSWorkerManager?

codalab/worker_manager/worker_manager.py
```diff
@@ -247,3 +252,19 @@ def run_one_iteration(self):
             logger.info('Starting a worker!')
             self.start_worker_job()
             self.last_worker_start_time = time.time()
+
+    def filter_bundles(
```
nelson-liu: This is a much needed change---thanks! Is it worth logging the bundles that get filtered out? I could see a scenario where someone launches a worker manager that doesn't have sufficient resources for a staged bundle, and the user is confused why the worker manager isn't launching jobs.

codalab/worker_manager/worker_manager.py
```diff
@@ -247,3 +252,19 @@ def run_one_iteration(self):
...
+        self, bundles: List[Dict[str, Dict[str, Union[int, str]]]]
+    ) -> List[Dict[str, Dict[str, Union[int, str]]]]:
```
epicfaace: Can we define a type alias for `List[Dict[str, Dict[str, Union[int, str]]]]` just to make this a bit cleaner?

codalab_service.py
```diff
@@ -256,8 +262,8 @@ def has_callable_default(self):
     # Worker manager
     CodalabArg(
         name='worker_manager_type',
-        help='Type of worker manager (e.g., aws-batch, azure-batch, slurm); only aws-batch supported right now',
-        default='aws-batch',
+        help='Type of worker manager (azure-batch or aws-batch)',
+        default='azure-batch',
```
epicfaace, line 266:
```suggestion
        default='aws-batch',
```

Better to keep it `aws-batch` by default for backwards compatibility?

codalab_service.py
```diff
@@ -321,9 +336,20 @@ def has_callable_default(self):
 for worker_manager_type in ['cpu', 'gpu']:
```
epicfaace, line 342: Why do we have different args for the default # of CPUs / memory based on the worker manager type?

teetone: Users can use different types of VMs based on different types of worker manager. In our case (the CodaLab Team), we use a pool of VM's that have 6 CPU's and 55GB of memory for each VM for the `gpu` WorkerManager and another pool of VM's that have 4 CPU's and 13GB of memory for each VM for the `cpu` WorkerManager.

codalab_service.py
teetone: I don't think we need to maintain backwards compatibility. Before the code changes in this PR, starting AWS Worker Managers via `codalab_service` was broken.

codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -0,0 +1,157 @@
...
+        OutputFileUploadCondition,
+        OutputFileUploadOptions,
+        TaskAddParameter,
+        TaskCounts,
+        TaskContainerSettings,
+    )
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "Running the worker manager requires the azure-batch module.\n"
+        "Please run: pip install azure-batch==9.0.0"
```
nelson-liu: do we need the version pin?

codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -0,0 +1,157 @@
...
+    )
+
+import logging
+import os
+import uuid
+from argparse import ArgumentParser
```
nelson-liu, line 23: alphabetical order? not sure if the linter complains about this.

codalab_service.py
```diff
@@ -321,9 +342,20 @@ def has_callable_default(self):
...
+            type=int,
+            default=1,
+            help='Default number of CPUs for each worker started by the {} worker manager'.format(
+                worker_manager_type
+            ),
+        ),
+        CodalabArg(
+            name='worker_manager_{}_default_memory_mb'.format(worker_manager_type),
+            type=int,
+            default=1024,
```
nelson-liu: this default doesn't match the 2048 above---is that intentional?

nelson-liu (approved):

codalab/worker_manager/azure_batch_worker_manager.py
teetone: It looks like they release a new version every 6 months. I don't think it needs to be pinned. I will remove the version for now.

codalab/worker_manager/azure_batch_worker_manager.py
teetone: I don't think Black complains about this. I actually wasn't sure what the best practice for import statements was, so I followed the style here: https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement

codalab_service.py
teetone: I arbitrarily set the value. I will update to be consistent with the worker managers.

PR #2955, opened 2020-10-09
Set up Azurite and Azure connection

epicfaace:
- Install Azure Blob Storage requirements in requirements.txt
- Add an "azurite" service to codalab_service.py
- Add an initialize-azurite script to it, which automatically creates the "bundles" container.
- A user can run `python3 codalab_service.py default azurite` to run the azurite service as well.


teetone (approved):
Can you add documentation for this new service?

Issue #2316, opened 2020-05-27
Show better error message when authentication failed 


candicegjing:
```
(py-3.6.8) jingge19@sc:~/codalab-worksheets$ cl-worker-manager --server=https://worksheets.codalab.org/ slurm-batch --nodelist=jagupard10
2020-05-27 10:31:07,174 Started worker manager.
Requesting access at https://worksheets.codalab.org/
Username: <username>
Password: 
Traceback (most recent call last):
  File "/sailhome/jingge19/codalab-worksheets/codalab/client/json_api_client.py", line 16, in wrapper
    return f(*args, **kwargs)
  File "/sailhome/jingge19/codalab-worksheets/codalab/client/json_api_client.py", line 415, in fetch
    query_params=self._pack_params(params),
  File "/sailhome/jingge19/codalab-worksheets/codalab/worker/rest_client.py", line 62, in _make_request
    access_token = self._get_access_token()
  File "/sailhome/jingge19/codalab-worksheets/codalab/lib/codalab_manager.py", line 426, in <lambda>
    lambda: self._authenticate(address, auth_handler),
  File "/sailhome/jingge19/codalab-worksheets/codalab/lib/codalab_manager.py", line 496, in _authenticate
    token_info = auth_handler.generate_token('credentials', username, password)
  File "/sailhome/jingge19/codalab-worksheets/codalab/server/auth.py", line 42, in generate_token
    {'grant_type': 'password', 'username': username, 'password': key}
  File "/sailhome/jingge19/codalab-worksheets/codalab/server/auth.py", line 61, in _make_token_request
    response = urllib.request.urlopen(request)
  File "/u/nlp/anaconda/main/anaconda3/envs/py-3.6.8/lib/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/u/nlp/anaconda/main/anaconda3/envs/py-3.6.8/lib/python3.6/urllib/request.py", line 532, in open
    response = meth(req, response)
  File "/u/nlp/anaconda/main/anaconda3/envs/py-3.6.8/lib/python3.6/urllib/request.py", line 642, in http_response
    'http', request, response, code, msg, hdrs)
  File "/u/nlp/anaconda/main/anaconda3/envs/py-3.6.8/lib/python3.6/urllib/request.py", line 570, in error
    return self._call_chain(*args)
[...]

PR #2957, opened 2020-10-11
Improve the UI message for PermissionError from log in

yiboliu:
Suppress extra stack traces for errors occurring when login credentials are incorrect

### Related issues
#2316 

### Screenshots

image
afer commit 4:
image

### Checklist

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -38,7 +38,7 @@ import time
 from distutils.util import strtobool
 
 from codalab.client.json_api_client import JsonApiClient
-from codalab.common import CODALAB_VERSION, PermissionError, UsageError
+from codalab.common import CODALAB_VERSION, UsageError, LoginPermissionError
 from codalab.lib.bundle_store import MultiDiskBundleStore
 from codalab.lib.crypt_util import get_random_string
 from codalab.lib.download_manager import DownloadManager
@@ -49,16 +49,6 @@ from codalab.lib import formatting
 from codalab.model.worker_model import WorkerModel
 
 
-def exception_handler(exctype, value, traceback, debug_hook=sys.excepthook):
-    # All your trace are belong to us!
-    # your format
-    if exctype == PermissionError:
-        print("%s: %s" % (exctype.__name__, value))
-    else:
-        debug_hook(exctype, value, traceback)
-
-
-sys.excepthook = exception_handler
 MAIN_BUNDLE_SERVICE = 'https://worksheets.codalab.org'
 
 
@@ -511,7 +501,7 @@ class CodaLabManager(object):
 
         token_info = auth_handler.generate_token('credentials', username, password)
         if token_info is None:
-            raise PermissionError("Invalid username or password.")
+            raise LoginPermissionError("Invalid username or password.")
         return _cache_token(token_info, username)
 
     def get_current_worksheet_uuid(self):
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -49,6 +49,16 @@ from codalab.lib import formatting
 from codalab.model.worker_model import WorkerModel
 
 
+def exception_handler(exctype, value, traceback, debug_hook=sys.excepthook):
+    # All your trace are belong to us!
+    # your format
+    if exctype == PermissionError:
+        print("%s: %s" % (exctype.__name__, value))
+    else:
+        debug_hook(exctype, value, traceback)
+
+
+sys.excepthook = exception_handler
 MAIN_BUNDLE_SERVICE = 'https://worksheets.codalab.org'
 
 
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -67,5 +67,3 @@ class RestOAuthHandler(object):
             if e.code == 401:
                 return None
             raise e
-        except urllib.error.URLError:
-            print("Your login info is incorrect. Please try again.")
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -68,4 +68,4 @@ class RestOAuthHandler(object):
                 return None
             raise e
         except urllib.error.URLError:
-            print("Your login info is incorrect. Please double check")
+            print("Your login info is incorrect. Please try again.")
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -6,7 +6,7 @@ import json
 import urllib.request
 import urllib.parse
 import urllib.error
-from codalab.common import URLOPEN_TIMEOUT_SECONDS
+from codalab.common import URLOPEN_TIMEOUT_SECONDS, LoginPermissionError
 
 
 # TODO(sckoo): clean up auth logic across:
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -66,4 +66,6 @@ class RestOAuthHandler(object):
         except urllib.error.HTTPError as e:
             if e.code == 401:
                 return None
-            raise
+            raise e
+        except urllib.error.URLError:
+            print("Your login info is incorrect. Please double check")
--- a/codalab/worker_manager/main.py
+++ b/codalab/worker_manager/main.py
@@ -4,6 +4,8 @@ Main entry point for the worker managers.
 
 import argparse
 import logging
+
+from codalab.common import LoginPermissionError
 from .aws_batch_worker_manager import AWSBatchWorkerManager
 from .azure_batch_worker_manager import AzureBatchWorkerManager
 from .slurm_batch_worker_manager import SlurmBatchWorkerManager
@@ -130,7 +132,10 @@ def main():
         )
 
     manager = worker_manager_types[args.worker_manager_name](args)
-    manager.run_loop()
+    try:
+        manager.run_loop()
+    except LoginPermissionError:
+        print('Login credentials are incorrect. Please double check')
 
 
 if __name__ == '__main__':
--- a/codalab/worker_manager/main.py
+++ b/codalab/worker_manager/main.py
@@ -5,7 +5,6 @@ Main entry point for the worker managers.
 import argparse
 import logging
 
-from codalab.common import LoginPermissionError
 from .aws_batch_worker_manager import AWSBatchWorkerManager
 from .azure_batch_worker_manager import AzureBatchWorkerManager
 from .slurm_batch_worker_manager import SlurmBatchWorkerManager
@@ -132,10 +131,7 @@ def main():
         )
 
     manager = worker_manager_types[args.worker_manager_name](args)
-    try:
-        manager.run_loop()
-    except LoginPermissionError:
-        print('All attempts have failed. Please double check your login credentials.')
+    manager.run_loop()
 
 
 if __name__ == '__main__':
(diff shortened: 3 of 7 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -49,6 +49,16 @@
 from codalab.model.worker_model import WorkerModel
 
 
+def exception_handler(exctype, value, traceback, debug_hook=sys.excepthook):
```
percyliang: Why put this in `codalab_manager.py`?

percyliang: Also, this seems like a potentially nice way to handle exceptions, but I wonder if it might be confusing / dangerous, since it has a global effect on the entire codebase?  Also, since we don't handle our other error messages through this mechanism...

yiboliu: This is not a global change. I put it here because we only want to handle the PermissionError in codalab_manager. I tested that PermissionError raised other places will be displayed in the conventional way.

epicfaace: I think it might be nicer to instead just catch PermissionError in cl.py, just like what we do with `KeyboardInterrupt`: https://github.com/codalab/codalab-worksheets/blob/5b83914cf4e73d70a9fe39ce62513c13222c9e1d/codalab/bin/cl.py#L7-L13

yiboliu: > I think it might be nicer to instead just catch PermissionError in cl.py, just like what we do with `KeyboardInterrupt`:
> 
> https://github.com/codalab/codalab-worksheets/blob/5b83914cf4e73d70a9fe39ce62513c13222c9e1d/codalab/bin/cl.py#L7-L13

Do we want to handle all PermissionErrors in this way? I noticed that we have multiple places raising PermissionError.

epicfaace: Hmm, in that case maybe it would be best to create a subclass of PermissionError called `LoginPermissionError` and then only catch `LoginPermissionError`.

yiboliu: The stacktrace doesn't go thru cl.py:
```
  File "/Users/yibol/codalab-worksheets/venv/bin/cl-worker-manager", line 33, in <module>
    sys.exit(load_entry_point('codalab', 'console_scripts', 'cl-worker-manager')())
  File "/Users/yibol/codalab-worksheets/codalab/worker_manager/main.py", line 121, in main
    manager.run_loop()
  File "/Users/yibol/codalab-worksheets/codalab/worker_manager/worker_manager.py", line 125, in run_loop
    self.run_one_iteration()
  File "/Users/yibol/codalab-worksheets/codalab/worker_manager/worker_manager.py", line 161, in run_one_iteration
    'bundles', params={'worksheet': None, 'keywords': keywords, 'include': ['owner']}
  File "/Users/yibol/codalab-worksheets/codalab/client/json_api_client.py", line 18, in wrapper
    return f(*args, **kwargs)
  File "/Users/yibol/codalab-worksheets/codalab/client/json_api_client.py", line 417, in fetch
    query_params=self._pack_params(params),
  File "/Users/yibol/codalab-worksheets/codalab/worker/rest_client.py", line 66, in _make_request
    access_token = self._get_access_token()
  File "/Users/yibol/codalab-worksheets/codalab/lib/codalab_manager.py", line 432, in <lambda>
    lambda: self._authenticate(address, auth_handler),
  File "/Users/yibol/codalab-worksheets/codalab/lib/codalab_manager.py", line 504, in _authenticate
    raise LoginPermissionError("Invalid username or password.")
codalab.common.LoginPermissionError: Invalid username or password.
```
Instead I did this in `codalab-worksheets/codalab/worker_manager/main.py`

yiboliu: Screenshot updated.

codalab/server/auth.py
```diff
@@ -66,4 +66,6 @@ def _make_token_request(self, data):
         except urllib.error.HTTPError as e:
             if e.code == 401:
                 return None
-            raise
+            raise e
+        except urllib.error.URLError:
+            print("Your login info is incorrect. Please double check")
```
teetone: In your latest screenshot, it looks like both error messages show up:

"Your login info is incorrect. Please double check"
'Login credentials are incorrect. Please double check'

yiboliu: Yeah, that's because they come from two separate processes. I can definitely remove one of them, but then we won't know where exactly are errors thrown. How do you think?

teetone: We should only show one of these errors.

codalab/worker_manager/worker_manager.py
```diff
@@ -124,19 +125,24 @@ def run_loop(self):
...
                 traceback.print_exc()
+            except (urllib.error.URLError, JsonApiException, ssl.SSLError) as e:
+                if "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed" in str(e):
+                    print("Your login info is incorrect. Please try again.")
+                else:
+                    traceback.print_exc()
+
             if self.args.once:
                 break
+            print('Sleeping {} seconds'.format(self.args.sleep_time))
```
teetone: We should probably omit this. This message is being shown to the user.

codalab/worker_manager/worker_manager.py
```diff
@@ -124,19 +125,24 @@ def run_loop(self):
...
                 NotFoundError,
-                JsonApiException,
             ):
                 # Sometimes, network errors occur when running the WorkerManager . These are often
                 # transient exceptions, and retrying the command would lead to success---as a result,
                 # we ignore these network-based exceptions (rather than fatally exiting from the
                 # WorkerManager )
                 traceback.print_exc()
+            except (urllib.error.URLError, JsonApiException, ssl.SSLError) as e:
+                if "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed" in str(e):
```
teetone: Is there a cleaner way to check if the user entered incorrect login info? Is it possible to check the type of exception instead?

codalab/worker_manager/main.py
```diff
@@ -118,7 +120,10 @@ def main():
         )
 
     manager = worker_manager_types[args.worker_manager_name](args)
-    manager.run_loop()
+    try:
+        manager.run_loop()
+    except LoginPermissionError:
+        print('All attempts have failed. Please double check your login credentials.')
```
teetone: I think you mentioned that it will infinitely prompt the user to retry. Can you clarify for what case it will hit this logic?

teetone (changes requested):
Can you add a test for this change maybe in `test_cli`?

codalab/worker_manager/worker_manager.py
yiboliu: The reason I do this is that in the case if there are other places throwing these types of errors, we don't want to handle in the same way, as the causes could be different.

codalab/worker_manager/main.py
yiboliu: Currently it's not hitting this point. The reason I didn't delete it is that if we are going to limit the number of retries, or give the user an option to quit, then we can use this message. If we are not going to do that, I'll delete it.

yiboliu: If we are not going to do that, I'll delete it.

codalab/worker_manager/worker_manager.py
teetone: I'm not sure what you mean by same way. Can you just check the exception type instead or is that not possible?

tests/cli/test_cli.py
```diff
@@ -2366,6 +2366,22 @@ def test_nonexistent(ctx):
...
+    password = os.getenv("CODALAB_PASSWORD")
+    del os.environ["CODALAB_USERNAME"]
+    del os.environ["CODALAB_PASSWORD"]
+    _run_command([cl, 'logout'])
+    os.environ["CODALAB_USERNAME"] = username
+    os.environ["CODALAB_PASSWORD"] = "wrongpassword"
+    result = _run_command([cl_worker_manager, '--server=https://worksheets.codalab.org/', 'slurm-batch', '--partition', 'foo'], 1)
+    print("result is " + str(result))
+    sys.stdout.flush()
+    os.environ["CODALAB_USERNAME"] = username
```
teetone, line 2392: Isn't this already handled inside the cleanup logic?

codalab/worker_manager/worker_manager.py
yiboliu: It is possible but I mean, if there are somewhere else throw the same type of error due to some other reason, do we still want to show this message? It might be not accurate.

tests/cli/test_cli.py
yiboliu: I'm following the example in `test_auth`

tests/cli/test_cli.py
```diff
@@ -2403,10 +2429,17 @@ def test_nonexistent(ctx):
         choices=list(TestModule.modules.keys()) + ['all', 'default'],
         help='Tests to run from: {%(choices)s}',
     )
+    parser.add_argument(
+        '--cl-worker-manager',
```
epicfaace, line 2433: Why did you add this argument? Seems like it's not used.

tests/cli/test_cli.py
```diff
@@ -2367,6 +2367,32 @@ def test_nonexistent(ctx):
...
+        [
+            cl_worker_manager,
+            '--server=https://worksheets.codalab.org/',
+            'slurm-batch',
+            '--partition',
+            'foo',
+        ],
+        1,
+    )
+    print("result is " + str(result))
```
epicfaace, line 2390: Should you be trying to check something is equal such as using `check_equals` instead of just printing the result?

tests/cli/test_cli.py
epicfaace: Why is the test for this PR specifically targeting `cl-worker-manager`? it might be simpler to just add a new test to `test_auth` with the regular `cl` command.

codalab/server/auth.py
```diff
@@ -6,7 +6,7 @@
 import urllib.request
 import urllib.parse
 import urllib.error
-from codalab.common import URLOPEN_TIMEOUT_SECONDS
+from codalab.common import URLOPEN_TIMEOUT_SECONDS, LoginPermissionError
```
epicfaace, line 9: rm import, not used

codalab/worker_manager/worker_manager.py
```diff
@@ -151,17 +152,20 @@ def run_loop(self):
...
+                urllib.error.URLError,
                 JsonApiException,
+                ssl.SSLError,
             ):
                 # Sometimes, network errors occur when running the WorkerManager . These are often
                 # transient exceptions, and retrying the command would lead to success---as a result,
                 # we ignore these network-based exceptions (rather than fatally exiting from the
                 # WorkerManager )
                 traceback.print_exc()
+            except NotFoundError:
```
epicfaace, line 166: Bit confused why you don't check for LoginPermissionError -- what's the difference between that error and NotFoundError in this context?

tests/cli/test_cli.py
epicfaace: Also, we need to have a test that checks to ensure that "Your login info is incorrect. Please try again." is printed to the console when the user enters incorrect credentials.

PR #2961, opened 2020-10-14
Regularize the column header name

yuqijin:

Make sure the column header does not contain the truncate symbols.

### Screenshots

Before:
Screen Shot 2020-10-14 at 2 24 55 PM

After:
Screen Shot 2020-10-14 at 2 38 39 PM

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/items/TableItem/TableItem.js
+++ b/frontend/src/components/worksheets/items/TableItem/TableItem.js
@@ -134,6 +134,8 @@ class TableItem extends React.Component<{
         let bundleInfos = item.bundles_spec.bundle_infos;
         let headerItems = item.header;
         let headerHtml = headerItems.map((item, index) => {
+            item = item === 'uuid[0:8]' ? 'uuid' : item;
+            item = item === 'summary[0:1024]' ? 'summary' : item;
             return (
                 <TableCell
                     onMouseEnter={(e) => this.setState({ hovered: true })}


frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -134,6 +134,8 @@ class TableItem extends React.Component<{
         let bundleInfos = item.bundles_spec.bundle_infos;
         let headerItems = item.header;
         let headerHtml = headerItems.map((item, index) => {
+            item = item === 'uuid[0:8]' ? 'uuid' : item;
+            item = item === 'summary[0:1024]' ? 'summary' : item;
```
percyliang, line 138: This wasn't happening before...would be not good to hard code this.  It seems like we should be displaying the field name, not how it's computed?

yuqijin: I found the table column is also `uuid[0:8]` in a PR submitted half a year ago [2190](https://github.com/codalab/codalab-worksheets/pull/2190). And it seems like the field name is also`uuid[0:8]`, so that's why we display it: https://github.com/codalab/codalab-worksheets/blob/34863d39ae696c343d6addc41a47c0a87c533271/codalab/lib/worksheet_util.py#L577.

percyliang: Ah, I see, the `uuid[0:8]` was there all along.  The `summary[0:1024]` is new because of another change, which is the part that looks a bit ugly.  But maybe we can just leave things as is and close this PR in the interest of being clear and explicit.

yuqijin: Got it!

PR #2963, opened 2020-10-15
Fix frontend focus issue

### LINKED ISSUE #2947: Frontend Focus Issue
- [x] When delete bundles or text, make sure that the cursor goes to the next item 

  - When deleting the last item in the table, 
    - If the table is the last item in the worksheet, go to the previous item in the table
    - Else, go to the next item in the worksheet
  - When deleting the non-consecutive items, logic is the same as deleting the last item in the deleting list
- [x] When insert schema, the cursor should be on the schema items
- [x] If have a search directive, putting the command in the wrong place
- [x] If no focused item exists, when adding schema, the focus should be moved to the schema editor

yuqijin:

There are bugs existing for the focus issue when deleting/inserting items.
The previous code does not support deleting multiple bundles, which is undesirable.

### Related issues

fixes #2947

### Screenshots
Delete multiple/non-consecutive bundles:
delete-multi

Delete single (last) item:
delete-single

Insert items by editing source:
insert-source

Uploading files:
insert-ui

Paste bundles:
paste

Search directive:
search

### Checklist


codalab/lib/worksheet_util.py
```diff
@@ -1008,6 +1008,7 @@ def flush_worksheets():
                     raw_to_block.append((len(blocks) - 1, 0))
             elif item_type == TYPE_DIRECTIVE:
                 command = get_command(value_obj)
+                raw_to_block_has_append = False
```
epicfaace: What does `raw_to_block_has_append` mean? I wonder if we should name it to something more clear...

yuqijin: It means whether the `raw_to_block` var has executed the `append` operation. So that we can decide whether we should do the  `append` operation or not later. If you think it is not clear, do you have any advice on it?

yuqijin: How about `raw_to_block_need_append`?

codalab/lib/worksheet_util.py
```diff
@@ -1019,6 +1020,8 @@ def flush_worksheets():
                     current_schema_ids.append(item_id)
                     current_schema_name = name
                     schemas[name] = current_schema = []
+                    raw_to_block.append((len(blocks) - 1 + len(current_schema_ids), 0))
```
yipenghe, line 1024: Need a comment on what this calculation means and why we need it, and maybe we can name the raw_to_block_has_append to appended_\<what the calculation represents\>

yuqijin: Changed the variable name and added comments.

yipenghe: I think appended_schema_blocks_index would be better cc @epicfaace

codalab/lib/worksheet_util.py
```diff
@@ -1055,7 +1059,8 @@ def flush_worksheets():
                 else:
                     raise UsageError("unknown directive `%s`" % command)
 
-                raw_to_block.append(None)
+                if not appended_raw_to_block:
```
yipenghe: why don't we need to append None if there weren't any schema blocks? (I also can't recall why we appended None before), if you have context can you add a comment here as well why the append None is needed?

codalab/lib/worksheet_util.py
yuqijin: I changed the name to `appended_schema_blocks_index`

codalab/lib/worksheet_util.py
yuqijin: Based on my understanding, it is because we need a placeholder for other cmds.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1410,7 +1442,11 @@ class Worksheet extends React.Component {
                             // When deleting text, we want the focus to stay at the same index,
                             // unless it is the last item in the worksheet, at which point the
                             // focus goes to the last item in the worksheet.
-                            this.setFocus(items.length === focus ? items.length - 1 : focus, 'end');
+                            this.setFocus(items.length === focus ? items.length - 1 : focus, 0);
+                        }
+                        if (searchDeleted) {
```
yipenghe: I'm not quite sure what searchDeleted is supposed to mean, from line 318, it seems to only be checking whether the command was 'rm', why is not called "fromDeleteCommand". Also,the comment says the condition is " no bundle deleted",but  wouldn't it still reach here even if some bundle was deleted?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -676,19 +692,11 @@ class Worksheet extends React.Component {
...
-        ) {
-            console.log('Out of bounds');
-            return; // Out of bounds (note index = -1 is okay)
-        }
+
         let focusedBundleUuidList = [];
         if (index !== -1) {
             // index !== -1 means something is selected.
-            // focusedBundleUuidList is a list of uuids of all bundles after the selected bundle (itself included)
+            // focusedBundleUuidList is a list of ids of all items after the selected bundle (itself included)
```
yipenghe: Are bundles' ids the same as their uuid? If so, better comment here as well

yipenghe (commented):
Overall looks good, just some questions

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: I renamed `searchDeleted` to `searchDirectiveDeleted`, so the naming pattern is the same as `textDeleted`, which represents which items we aim to delete. 
We will execute this if condition only when we delete some bundles through the search directive, since the number of bundles does not change.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Thanks for your advice, I have included it in the comments.

yipenghe (approved):
Great. Maybe just the variable name change we discussed offline

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Have switched the var name to `fromDeleteCommand`

PR #2965, opened 2020-10-16
Resolve following flake8 violations

yiboliu:
More flake8 violations introduced after previous commit. So resolve them here

#2660 
### Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,5 +50,5 @@ ignore = E203,E731,W503,W605
 per-file-ignores =
   #  Keep the unused imports as they are used to route the rest service
   codalab/server/rest_server.py: F401
-  #  Keep the previous format of documentation
+  #  Modifying those texts in accordance with flake8 will change how the generated markdown look like
   ./scripts/gen-rest-docs.py: E501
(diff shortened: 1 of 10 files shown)


setup.cfg
```diff
@@ -44,8 +44,14 @@ exclude = venv/*,var/*
...
 # Ignore on a per file basis:
 # F401 - module imported but not used
+# E501 - line too long, exceeds max-line-length
+# F821 - undefined name
 
 ignore = E203,E731,W503,W605
 per-file-ignores =
   #  Keep the unused imports as they are used to route the rest service
   codalab/server/rest_server.py: F401
+  #  Keep the previous format of documentation
```
teetone: This file isn't documentation. It's a script.

setup.cfg
```diff
@@ -44,8 +44,14 @@ exclude = venv/*,var/*
...
+  ./scripts/gen-rest-docs.py: E501
+  #  Keep the original function name here
+  ./codalab/rest/interpret.py: F821
```
teetone: Which line is this failing? If it's just one or two instance, can you ignore it in the file?

setup.cfg
yiboliu: I know, but I think the text here will be written to the markdown. That's why I worded like this

teetone (approved):

PR #2968, opened 2020-10-16
Copy over BlobStorageFileSystem code from Apache Beam

epicfaace:
Copy over BlobStorageFileSystem code from Apache Beam. This will be used in future PRs to actually connect to Azure Blob Storage.


teetone (approved):
If you haven't already, can you create an issue that tracks removing these files later?

PR #2969, opened 2020-10-16
[Blob Storage] Set up abstraction for parsing linked URLs

epicfaace:
Set up an abstraction for parsing linked URLs, especially URLs that refer to a particular file that is inside a `.zip` file. Uses the LinkedBundlePath dataclass to store this information; right now, this is only used for Azure Blob Storage, but can be generalized to other storage systems.

Dataclasses aren't supported in Python 3.6 so I've just added a pip package for it, but we can remove that package once we upgrade the minimum version of Python required.


codalab/lib/path_util.py
```diff
@@ -298,6 +301,75 @@ def rename(old_path, new_path):
...
+
+    storage_type: StorageType
+    bundle_path: str
+    is_zip: bool
+    uses_beam: bool
+    zip_subpath: str
+    bundle_uuid: str
+
+
+def parse_linked_bundle_url(url):
```
teetone, line 336: Would it be cleaner if we just made this logic be part of the constructor of `LinkedBundlePath` that would take in the url?

codalab/lib/path_util.py
```diff
@@ -298,6 +301,75 @@ def rename(old_path, new_path):
...
+
+        Returns a LinkedBundlePath instance to encode this information.
+    """
+    if url.startswith("azfs://"):
+        uses_beam = True
+        storage_type = StorageType.AZURE_BLOB_STORAGE
+        url = url[len("azfs://") :]
+        storage_account, container, bundle_uuid, contents_file, *remainder = url.split("/", 4)
+        bundle_path = f"azfs://{storage_account}/{container}/{bundle_uuid}/{contents_file}"
+        if contents_file.endswith(".zip"):
```
teetone: I think this can just be rewritten as:

```
is_zip = contents_file.endswith(".zip")
zip_subpath = remainder[0] if is_zip and len(remainder) else None
```

teetone (approved):

codalab/lib/path_util.py
epicfaace: I think it's a bit complicated to override constructors for frozen dataclasses; can't think of a nice way to make this work that way.

PR #2970, opened 2020-10-16
Azure Blob Storage - Set up upload / download with .zip files

### LINKED ISSUE #3188: Azure Blob Storage - set up basic upload with .zip files
Set up basic upload option that uses .zip files on Azure Blob Storage for storage.

This upload will be used when the user uses the `-a` option when running `cl upload`.

Once a bundle is uploaded, it should be able to be download / previewed regularly, just like any other bundle.

epicfaace:
Update upload / download logic to work with .zip-file-based structures (such as that involved in Blob Storage files). Fixes #3188.

I add an Azure Blob Storage-backed upload/download test to ensure that uploading and downloading files works properly (https://github.com/codalab/codalab-worksheets/pull/2970/files#diff-d9f4f7687e004cda7812561dce93eb41bacda5b8042e470cecaacbb2546f45d1). For this, I create a `MockBlobStorageFileSystem` that uses a file-based backend (but which behaves like Azure), which is used by the upload / download unit tests in place of actually using Azure Blob Storage for speed.

Note: this code does not yet preserve permissions for files in Azure Blob Storage. Will do that in a future PR.


epicfaace:
Going to merge this into #3198.

PR #2973, opened 2020-10-16
Revert "`uls` search for user stats"

epicfaace:
Because CI is failing -- reverts codalab/codalab-worksheets#2389

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -137,7 +137,7 @@ GROUP_AND_PERMISSION_COMMANDS = (
     'chown',
 )
 
-USER_COMMANDS = ('uinfo', 'uedit', 'ufarewell', 'uls')
+USER_COMMANDS = ('uinfo', 'uedit', 'ufarewell')
 
 SERVER_COMMANDS = (
     'workers',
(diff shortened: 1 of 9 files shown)


Levilian (commented):
These are the only things we need to remove to pass the format test. 
`./tests/cli/test_cli.py:29:1: F401 'codalab.objects.user.User' imported but unused
./tests/cli/test_cli.py:30:1: F401 'codalab.model.tables.NOTIFICATIONS_IMPORTANT' imported but unused
./tests/cli/test_cli.py:2363:1: F811 redefinition of unused 'test' from line 1018
./codalab/rest/users.py:178:18: E711 comparison to None should be 'if cond is None:'`

Do we want to do it here or revert and then merge?

epicfaace:
> These are the only things we need to remove to pass the format test.
> `./tests/cli/test_cli.py:29:1: F401 'codalab.objects.user.User' imported but unused ./tests/cli/test_cli.py:30:1: F401 'codalab.model.tables.NOTIFICATIONS_IMPORTANT' imported but unused ./tests/cli/test_cli.py:2363:1: F811 redefinition of unused 'test' from line 1018 ./codalab/rest/users.py:178:18: E711 comparison to None should be 'if cond is None:'`
> 
> Do we want to do it here or revert and then merge?

@Levilian seems like #2975 fixed it, so we should be good!

PR #2976, opened 2020-10-17
Better handle EC2 spot interruptions in AWS Batch WorkerManager

nelson-liu:

When EC2 spot instances are preempted, they're given 2 minutes to clean up. These cleanup signals are sent via an AWS web endpoint (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#spot-instance-termination-notices). Then, the container terminates.

Right now, we don't catch these interruption notices, so jobs running on AWS workers will go into `worker_offline` state instead of being restaged. This PR adds a script to (1) catch the EC2 preemption notices and (2) send a SIGTERM to all CodaLab workers when it sees one. When the CodaLab worker gets a SIGTERM, it should restage its jobs (if applicable) / clean up and prepare for a graceful exit.

### DIFF
--- a/scripts/detect-ec2-spot-preemption.sh
+++ b/scripts/detect-ec2-spot-preemption.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+while true
+do
+    # This IP address comes from:
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#spot-instance-termination-notices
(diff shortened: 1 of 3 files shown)


scripts/detect-ec2-spot-preemption.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+while true
```
percyliang, line 3: Could you document where this script comes from / where the IP address, etc. comes from?

nelson-liu:
superseded by #2977

PR #2977, opened 2020-10-17
Better handle EC2 spot interruptions in AWS Batch WorkerManager

nelson-liu:
(Continuation of #2976 , opening from a branch on codalab/codalab-worksheets so the tag is pushed to docker hub)

When EC2 spot instances are preempted, they're given 2 minutes to clean up. These cleanup signals are sent via an AWS web endpoint (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#spot-instance-termination-notices). Then, the container terminates.

Right now, we don't catch these interruption notices, so jobs running on AWS workers will go into `worker_offline` state instead of being restaged. This PR adds a script to (1) catch the EC2 preemption notices and (2) send a SIGTERM to all CodaLab workers when it sees one. When the CodaLab worker gets a SIGTERM, it should restage its jobs (if applicable) / clean up and prepare for a graceful exit.

### DIFF
--- a/scripts/detect-ec2-spot-preemption.sh
+++ b/scripts/detect-ec2-spot-preemption.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+while true
+do
+    # This IP address comes from:
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#spot-instance-termination-notices
+    # It's a special endpoint set up by AWS whereby AWS instances can view metadata about themselves.
+    # One such piece of metadata is the termination time, which is only set when the spot instance is to be
+    # pre-empted (you get a 404 otherwise).
+    # This script was partially taken from https://stackoverflow.com/q/32613600/14089059 .
+    if [ -z $(curl -Is http://169.254.169.254/latest/meta-data/spot/termination-time | head -1 | grep 404 | cut -d \  -f 2) ]
+    then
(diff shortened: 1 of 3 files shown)


scripts/detect-ec2-spot-preemption.sh
```diff
@@ -0,0 +1,21 @@
...
+
+while true
+do
+    # This IP address comes from:
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#spot-instance-termination-notices
+    # It's a special endpoint set up by AWS whereby AWS instances can view metadata about themselves.
+    # One such piece of metadata is the termination time, which is only set when the spot instance is to be
+    # pre-empted (you get a 404 otherwise).
+    # This script was partially taken from https://stackoverflow.com/q/32613600/14089059 .
+    if [ -z $(curl -Is http://169.254.169.254/latest/meta-data/spot/termination-time | head -1 | grep 404 | cut -d \  -f 2) ]
```
epicfaace: Could we use the instance-action endpoint instead? It might make the code a bit cleaner and seems to be the recommended way to do it:

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#spot-instance-termination-notices

nelson-liu:
I tested this branch as-is on some live AWS Btach jobs before ashwin's comment, and it seemed to work (according to the Batch logs, that is)

epicfaace (approved):

PR #2979, opened 2020-10-21
Provide a keyboard shortcut to exit schema editor 

### LINKED ISSUE #2951: Keyboard Shortcut Issue
- [x] Sometimes ‘k’ skips over a bundle, though ‘up’ works; sometimes happens after pasting in a bundle

- [x] Editing schema - ‘enter’ shouldn’t add a space
updated: 
The reason why this happens is we manually replace `space`  with `_` when editing the schema name since `space`  is not allowed in a legal schema name. [code](https://github.com/codalab/codalab-worksheets/blob/c9197db059a7c2146be643ad9fc293bfd4cdb5cd/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js#L148)
- [x] Be able to get out of editing schema using keyboard shortcuts
- [x] hitting 'enter' on a list of worksheets on a worksheets does nothing #2939

yuqijin:

We can use `ctrl+enter` to exit from the schema editor and save the changes. However, previously, the workflow is when the changes are saved, no reminder/dialog is provided, and the editor remains open, which is confusing to the user. 

Option1: **Dialog**
1. User press `ctrl+enter` to exit from the schema editor and save the changes
2. A dialog of "Schema Saved" is shown on the screen for 1.5s and then automatically hidden
3. The schema editor closed automatically.

 Option2: **Popover**
1. User press `ctrl+enter` to exit from the schema editor and save the changes
2. A popover containing "Schema Saved" is shown on the screen for 1.5s and then automatically hidden
3. The schema editor closed automatically.
### Related issues

fixes part of #2951 

### Screenshots
 **Previous**:
before

 **Dialog**:
after

 **Popover**:
popover

### Checklist


frontend/src/components/worksheets/WorksheetDialogs.js
```diff
@@ -199,6 +199,20 @@ class WorksheetDialogs extends React.Component {
                         </DialogContentText>
                     </DialogContent>
                 </Dialog>
+                {/* Schema saved message dialog */}
+                <Dialog
```
epicfaace: Perhaps we should use the `Snackbar` component, which we use elsewhere for these kinds of messages?

epicfaace (approved):

PR #2991, opened 2020-10-26
Keep all other fields intact and clear only the password field when the user fails in the sign up process

### LINKED ISSUE #2404: when signup for account, all fields are cleared when passwords don't match
Don't clear fields since making users re-type everything is pretty annoying, especially when they're trying to sign up.

yuqijin:

Enable the users not to re-type everything when they're trying to sign up. Clear the password field since it contains secure information.

### Related issues

fixes #2404 

### Screenshots

ezgif com-gif-maker

### Checklist

### DIFF
--- a/frontend/src/components/SignUp.js
+++ b/frontend/src/components/SignUp.js
@@ -43,6 +43,28 @@ export class SignUp extends React.Component {
         );
     };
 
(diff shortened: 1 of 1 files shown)


yipenghe (approved):
Nice!

PR #2994, opened 2020-10-26
Show a dialog for schema deletion

### LINKED ISSUE #2917: Delete schema - show a dialog instead of inline buttons
Right now, when you delete a schema, it just shows two inline buttons.

image

This doesn't feel very consistent with the rest of the deletion experience. Instead, we should show a dialog box, just like what we do for deleting a worksheet:

image

yuqijin:

Instead of showing inline buttons, provide a dialog for schema deletion, so that it is more consistent with other delete actions.
Inline buttons:
94928056-155b3500-0491-11eb-9508-70e97beace51

### Related issues

fixes #2917 

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
+++ b/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
@@ -38,7 +38,6 @@ class SchemaItem extends React.Component<{
             rows: [...this.props.item.field_rows],
             curSchemaName: this.props.item.schema_name,
             newAddedRow: -1,
-            confirmingDeletion: false,
         };
     }
 
@@ -50,7 +49,6 @@ class SchemaItem extends React.Component<{
         }
         if (save) {
             this.saveSchema();
-            this.setState({ confirmingDeletion: false });
         }
     };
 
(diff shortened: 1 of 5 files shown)


yipenghe:
From the recording, it seems that after you cancel the dialog, you can't click the delete button gain

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -317,6 +314,7 @@ class SchemaItem extends React.Component<{
                                         return;
                                     }
                                     this.setState({ confirmingDeletion: true });
+                                    this.props.setDeleteItemCallback(this.deleteThisSchema);
```
yipenghe, line 314: I think you need to set confirmingDeletion to false in the deleteThisSchema function, so the button will be re-enabled

yipenghe (changes requested):

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
yuqijin: Thanks for pointing out the bug! But actually, I think we can delete the `confirmingDeletion` state variable since when we use dialog, all other buttons on the page will be automatically disabled.

yipenghe: Ah makes sense, great!

yipenghe (approved):

PR #2995, opened 2020-10-28
Fix bug of insertion after pasting bundles; Fix bug for `Add Schema` action button

### LINKED ISSUE #2949: When focusing on a bundle, Adding schema doesn’t work
**Describe the bug**
When adding schema at the end of a worksheet, the new schema will replace the old one, which is a bug.
image

yuqijin:

**Fix bug of insertion after pasting bundles**
Previously, if we 

1. create a new worksheet
2. paste bundles
3. click `Add Text` or `Add schema` button

The new items will be added to the wrong place. Ther bug is due to when we paste bundles to a worksheet, the sort keys for the bundles are `None`, so when we want to add things after them, the new items will always be added to the top of the worksheet.
**Solution**
Assign a sort key for the pasting bundles.

 **Fix bug for `Add Schema` action button**
Previously, when we focus on a bundle, we cannot add new schemas, and that's the reason why the `Add Schema` button does not work sometimes.  
**Solution**
Enable the user to add schemas when focusing on a bundle.

### Related issues

fixes #2949 

### Screenshots
**Fix bug of insertion after pasting bundles**
paste

 **Fix bug for `Add Schema` action button**
schema

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -471,6 +471,10 @@ class Worksheet extends React.Component {
         let actualData = { items };
         if (after_sort_key) {
             actualData['after_sort_key'] = after_sort_key;
+        } else {
+            // If no location for the insertion is specified,
+            // insert the new item at the top of the worksheet in default.
+            actualData['after_sort_key'] = -1;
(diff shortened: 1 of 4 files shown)


yipenghe (approved):
Looks good

Issue #2510, opened 2020-06-25
codalab-service doesn't work


percyliang:
`pip install codalab` then run `codalab-service start`.

```
$ codalab-service start
Traceback (most recent call last):
  File "/home/pliang/.local/bin/codalab-service", line 7, in <module>
    from codalab_service import main
ModuleNotFoundError: No module named 'codalab_service'
```

percyliang:
Motivation: this would allow people to start local CodaLab instances which would make it easier to try out experiments very fast locally.

yiboliu:
Currently running the command locally produces this stacktrace:
```Traceback (most recent call last):
  File "/Users/yibol/codalab-worksheets/venv/bin/codalab-service", line 33, in <module>
    sys.exit(load_entry_point('codalab', 'console_scripts', 'codalab-service')())
  File "/Users/yibol/codalab-worksheets/codalab_service.py", line 98, in main
    service_manager.execute()
  File "/Users/yibol/codalab-worksheets/codalab_service.py", line 578, in execute
    self.start_services()
  File "/Users/yibol/codalab-worksheets/codalab_service.py", line 758, in start_services
    'if [ $(alembic current | wc -l) -gt 0 ]; then echo upgrade; alembic upgrade head; fi'
  File "/Users/yibol/codalab-worksheets/codalab_service.py", line 698, in run_service_cmd
    ' bash -c \'%s\'' % cmd
  File "/Users/yibol/codalab-worksheets/codalab_service.py", line 677, in _run_compose_cmd
    raise Exception('Command exited with code {}'.format(popen.returncode))
Exception: Command exited with code 255
```

yiboliu:
If I reinstall with codalab-yibol and then run with `codalab-service start`, I get error similar to #2800 
```Traceback (most recent call last):
  File "/Users/yibol/codalab-worksheets/env/bin/codalab-service", line 33, in <module>
    sys.exit(load_entry_point('codalab', 'console_scripts', 'codalab-service')())
  File "/Users/yibol/codalab-worksheets/env/bin/codalab-service", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/Users/yibol/codalab-worksheets/env/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 558, in distribution
    return Distribution.from_name(distribution_name)
[...]

PR #2997, opened 2020-10-29
Fix "codalab-service start" command

yiboliu:
When users do a `pip install codalab`, they are not able to start the service by `codalab-service start`. I'm adding the codalab_service.py and associated docker files to the installed files, so that the codalab service can start successfully.

### Related issues
#2510

### Checklist


epicfaace:
@yiboliu can you update this PR description to explain what the problem exactly was and how your solution fixes it?

epicfaace (commented):
Can you update test_cli.py to ensure that `codalab-service` works and doesn't give an error? Like with `cl-competitiond` (https://github.com/codalab/codalab-worksheets/blob/service2/tests/cli/test_cli.py#L1977)

epicfaace:
Can you also update the PR title to be more descriptive (as it will become the commit message when merged)?

tests/cli/test_cli.py
```diff
@@ -1984,6 +1984,11 @@ def test_competition(ctx):
         os.remove(out_file)
 
 
+@TestModule.register('start')
+def test_competition(ctx):
```
```suggestion
def test_start(ctx):
```

epicfaace (approved):

Issue #2760, opened 2020-08-16
TypeError: expected string or bytes-like object


teetone:
**Describe the bug**
Seeing the following stack trace on the prod server:

```
>> timeout 60s cl rm   (exit code 1, time 8.62s [limit: 15s,60s]; avg 8.00s; max 8.62s)
Traceback (most recent call last):
  File "/usr/local/bin/cl", line 33, in 
    sys.exit(load_entry_point('codalab', 'console_scripts', 'cl')())
  File "/opt/codalab-worksheets/codalab/bin/cl.py", line 10, in main
    cli.do_command(sys.argv[1:])
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 885, in do_command
    structured_result = command_fn()
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 879, in 
    command_fn = lambda: args.function(self, args)
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 1904, in do_rm_command
    bundle_uuids = self.target_specs_to_bundle_uuids(client, worksheet_uuid, args.bundle_spec)
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 547, in target_specs_to_bundle_uuids
    for spec in target_specs
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 547, in 
    for spec in target_specs
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 588, in resolve_target
    bundle_uuid = BundleCLI.resolve_bundle_uuid(client, worksheet_uuid, bundle_spec)
  File "/opt/codalab-worksheets/codalab/lib/bundle_cli.py", line 635, in resolve_bundle_uuid
    if spec_util.UUID_REGEX.match(bundle_spec):
TypeError: expected string or bytes-like object
```

I see an issue was previously filed for the same error: https://github.com/codalab/codalab-worksheets/issues/2349 but it seems like it was never resolved.

PR #3000, opened 2020-10-30
Resolve crash of an null-valued bundle spec

yiboliu:
When the bundle_spec returned as None, there will be a crash of regex matching invalid type. Since bundle_spec is required, I'm checking if it's missing - if yes, then throw an UsageError

### Related issues
#2760 

### Screenshots
image

### Checklist

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -563,7 +563,7 @@ class BundleCLI(object):
         instance, worksheet_spec, bundle_spec, subpath = parse_target_spec(target_spec)
 
         if bundle_spec is None:
-            raise UsageError('bundle spec is not missing')
+            raise UsageError('bundle spec is missing')
 
         if instance is not None:
             if self.headless:
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -563,7 +563,7 @@ class BundleCLI(object):
         instance, worksheet_spec, bundle_spec, subpath = parse_target_spec(target_spec)
 
         if bundle_spec is None:
-            raise UsageError('bundle spec is missing')
+            raise UsageError('Bundle spec is missing')
 
         if instance is not None:
             if self.headless:
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -562,6 +562,9 @@ class BundleCLI(object):
         """
         instance, worksheet_spec, bundle_spec, subpath = parse_target_spec(target_spec)
 
+        if bundle_spec is None:
+            raise UsageError('bundle spec is not missing')
+
         if instance is not None:
             if self.headless:
                 raise UsageError('Cannot use alias on web CLI')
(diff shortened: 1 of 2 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -562,6 +562,9 @@ def resolve_target(
         """
         instance, worksheet_spec, bundle_spec, subpath = parse_target_spec(target_spec)
 
+        if bundle_spec is None:
+            raise UsageError('bundle spec is not missing')
```
teetone: Don't you mean `Bundle spec is missing`?

teetone (commented):
Can you add a test for this in `test_cli.py`?

codalab/lib/bundle_cli.py
yiboliu: oh yeah. Thanks for pointing out.

tests/cli/test_cli.py
```diff
@@ -2366,6 +2366,12 @@ def test_nonexistent(ctx):
     _run_command([cl, 'work', 'nonexistent::'], expected_exit_code=1)
 
 
+@TestModule.register('rm_empty')
```
teetone: This test is probably not running right now. You have to register the module in `test.yml`. Can you just add this test to the existing 'rm' module?

codalab/lib/bundle_cli.py
```diff
@@ -562,6 +562,9 @@ def resolve_target(
         """
         instance, worksheet_spec, bundle_spec, subpath = parse_target_spec(target_spec)
 
+        if bundle_spec is None:
+            raise UsageError('bundle spec is missing')
```
teetone: Capitalize `Bundle`

teetone (changes requested):

tests/cli/test_cli.py
```diff
@@ -2366,6 +2366,12 @@ def test_nonexistent(ctx):
...
+def test_rm_empty(ctx):
+    result = _run_command([cl, 'rm', '""'], expected_exit_code=1)
```
teetone: Can this just be an empty string '' instead of '""'?

tests/cli/test_cli.py
```diff
@@ -896,6 +896,8 @@ def test_rm(ctx):
     uuid = _run_command([cl, 'upload', test_path('a.txt')])
     _run_command([cl, 'add', 'bundle', uuid])  # Duplicate
     _run_command([cl, 'rm', uuid])  # Can delete even though it exists twice on the same worksheet
+    result = _run_command([cl, 'rm', ''], expected_exit_code=1)  # Empty parameter should give an Usage error
+    check_equals(result, '')
```
teetone: Is this necessary to check?

teetone (approved):

tests/cli/test_cli.py
yiboliu: I thought I could check the info printed to the console this way, which turned out to be not true.

PR #3008, opened 2020-11-02
Switch INT id PK to BIGINT id PK, retaining auto-increment

nelson-liu:
# Reasons for making this change

codalab.stanford.edu is broken because the primary key on the bundle_metadata table exceeds the INT range. This is what I see server-side when I try to upload:

```
sqlalchemy.exc.IntegrityError: (MySQLdb._exceptions.IntegrityError) (1062, "Duplicate entry '2147483647' for key 'PRIMARY'")
[SQL: INSERT INTO bundle_metadata (bundle_uuid, metadata_key, metadata_value) VALUES (%s, %s, %s), (%s, %s, %s), (%s, %s, %s), (%s, %s, %s), (%s, %s, %s), (%s, %s, %s), (%s, %s, %s)]
[parameters: ('0x582ac54ba0894c69a800bc982cbdb1cf', 'name', 'squadv1.1format..iqa3-train.json', '0x582ac54ba0894c69a800bc982cbdb1cf', 'description', '', '0x582ac54ba0894c69a800bc982cbdb1cf', 'created', '1604347113', '0x582ac54ba0894c69a800bc982cbdb1cf', 'license', '', '0x582ac54ba0894c69a800bc982cbdb1cf', 'source_url', '', '0x582ac54ba0894c69a800bc982cbdb1cf', 'link_url', '', '0x582ac54ba0894c69a800bc982cbdb1cf', 'link_format', '')]
(Background on this error at: http://sqlalche.me/e/13/gkpj)
```

@teetone and I were talking about this, and it's apparently possible to blow past the INT range on an `id` field without having MAXINT records because of things like `REPLACE`, which re-create a new row (thus incrementing the id). I have been doing a lot of `cl edit`s, so this seems like the culprit.

To fix this, I changed INT to BIGINT across all `id` PK fields.

After this change:
```
mysql> describe bundle_metadata;
+----------------+-------------+------+-----+---------+----------------+
| Field          | Type        | Null | Key | Default | Extra          |
+----------------+-------------+------+-----+---------+----------------+
| id             | bigint(20)  | NO   | PRI | NULL    | auto_increment |
| bundle_uuid    | varchar(63) | NO   | MUL | NULL    |                |
| metadata_key   | varchar(63) | NO   | MUL | NULL    |                |
| metadata_value | text        | NO   |     | NULL    |                |
+----------------+-------------+------+-----+---------+----------------+
```

The `BigInteger().with_variant(Integer, "sqlite")` is necessary because we want to use `BigInteger`, but sqlite doesn't allow `BigInteger` auto-incrementing columns. So, we use INTEGER in this case instead.

### DIFF
--- a/alembic/versions/2020110220_change_id_primary_keys_from__fcb22a612d2a.py
+++ b/alembic/versions/2020110220_change_id_primary_keys_from__fcb22a612d2a.py
@@ -1,0 +1,54 @@
+"""Change id primary keys from Integer to BigInteger, retaining auto-increment
+
+Revision ID: fcb22a612d2a
+Revises: 9d3ab0b9d304
+Create Date: 2020-11-02 20:57:08.361251
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = 'fcb22a612d2a'
+down_revision = '9d3ab0b9d304'
+
+
+def upgrade():
+    op.alter_column("bundle", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.Integer(), nullable=False)
+    op.alter_column("bundle_metadata", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.Integer(), nullable=False)
+    op.alter_column("bundle_dependency", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.Integer(), nullable=False)
+    op.alter_column("worksheet", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.Integer(), nullable=False)
+    op.alter_column("worksheet_item", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.Integer(), nullable=False)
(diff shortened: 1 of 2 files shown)


bkgoksel (approved):
LGTM! AFAIK the main culprit for this is the `last_checkin` (or whatever it's called).
Basically for each bundle, every time a worker checks in for it (every few seconds), a new table entry is made. You can imagine how quickly that would go thru the ids.

Honestly just moving that field to be a core `bundle` table schema field rather than bundle metadata would probably fix 99% of this issue.

Another hacky/quick fix is to insert everything from the existing bundle metadata table to a new temp table, then rename the tables such that the new table becomes the bundle metadata table. That defragments the ids. This is what we did at SM and it kept us going for many months without having to do a bigger fix.

nelson-liu:
> Basically for each bundle, every time a worker checks in for it (every few seconds), a new table entry is made. You can imagine how quickly that would go thru the ids.

I totally forgot about this. We've also been running a very large amount of workers recently (100+ish) and operations can get pretty slow; makes sense that there's a bunch of DB writing happening that is slowing stuff down.

> Another hacky/quick fix is to insert everything from the existing bundle metadata table to a new temp table, then rename the tables such that the new table becomes the bundle metadata table. That defragments the ids. This is what we did at SM and it kept us going for many months without having to do a bigger fix.

I just did this---thanks for the tip!

teetone (approved):

Issue #2282, opened 2020-05-20
Publish codalab package to pypi with wheels, not eggs


epicfaace:
Wheels are now considered the standard for publishing Python packages. See https://pythonwheels.com/ and https://packaging.python.org/discussions/wheel-vs-egg/

PR #3014, opened 2020-11-04
Wheel

yiboliu:
Use wheel as format of uploading to PyPi. 

### Related issues
#2282 

### Checklist

### DIFF
--- a/scripts/upload-to-pypi.sh
+++ b/scripts/upload-to-pypi.sh
@@ -11,6 +11,6 @@ VER=$1
 set -e
 
 echo "==> Packaging codalab"
-python3 setup.py sdist
+python3 setup.py bdist_wheel
 echo "==> Uploading codalab"
-twine upload dist/codalab-$VER.tar.gz
+twine upload dist/codalab-$VER-py3-none-any.whl
--- a/scripts/upload-to-pypi.sh
+++ b/scripts/upload-to-pypi.sh
@@ -11,6 +11,7 @@ VER=$1
 set -e
 
 echo "==> Packaging codalab"
-python3 setup.py bdist_wheel
+python3 setup.py bdist_wheel sdist
 echo "==> Uploading codalab"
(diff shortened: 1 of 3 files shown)


nelson-liu (approved):

scripts/upload-to-pypi.sh
```diff
@@ -11,6 +11,6 @@ VER=$1
 set -e
 
 echo "==> Packaging codalab"
-python3 setup.py sdist
+python3 setup.py bdist_wheel
```
epicfaace: Can you ensure we publish _both_ eggs and wheels? It's good to publish source distributions (eggs) as well, see this guide: https://packaging.python.org/guides/distributing-packages-using-setuptools/#packaging-your-project

epicfaace (changes requested):

epicfaace (commented):
Can you also double check to ensure that `LICENSE.txt` is included in the built wheel? See https://wheel.readthedocs.io/en/latest/user_guide.html#including-license-files-in-the-generated-wheel-file

scripts/upload-to-pypi.sh
yiboliu: Sure. I just pass both argument to it.

yiboliu:
> Can you also double check to ensure that `LICENSE.txt` is included in the built wheel? See https://wheel.readthedocs.io/en/latest/user_guide.html#including-license-files-in-the-generated-wheel-file

image

epicfaace (approved):

setup.cfg
```diff
@@ -64,3 +64,5 @@ ignore_missing_imports = True
 
 [mypy-urllib3.util.retry]
 ignore_missing_imports = True
+[metadata]
```
epicfaace: We actually don't need this part; it seems that files matching `LICEN[CS]E*` are automatically included: https://wheel.readthedocs.io/en/latest/user_guide.html#including-license-files-in-the-generated-wheel-file

PR #3012, opened 2020-11-04
Stream file when gzip compressing

nelson-liu:
Attempting to fix #3011 --- instead of `read()`ing the entire file into memory, make a `GzipFile` and have it write to a `BytesIO` stream.

PR #3015, opened 2020-11-04
Add stress test for bundle in memory

### LINKED ISSUE #3013: Stress test having a large bundle in memory
Add a stress test that has a large bundle as a dependency. This is to catch regressions such as https://github.com/codalab/codalab-worksheets/issues/3011.

teetone:

We need this stress test to test out https://github.com/codalab/codalab-worksheets/pull/3012

Resolves #3013 

### Checklist

### DIFF
--- a/tests/stress/stress_test.py
+++ b/tests/stress/stress_test.py
@@ -160,6 +160,8 @@ class StressTestRunner:
 
     def _test_large_bundle(self):
         self._set_worksheet('large_bundles')
+        # Set this to larger than the max memory on the system to test that data is being
+        # streamed when the large bundle is being used as a dependency.
         large_file = TestFile('large_file', self._args.large_file_size_gb * 1000)
         dependency_uuid = self._run_bundle([self._cl, 'upload', large_file.name()])
         large_file.delete()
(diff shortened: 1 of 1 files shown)


nelson-liu (approved):

tests/stress/stress_test.py
```diff
@@ -161,8 +161,16 @@ def _heartbeat(self):
     def _test_large_bundle(self):
         self._set_worksheet('large_bundles')
         large_file = TestFile('large_file', self._args.large_file_size_gb * 1000)
-        self._run_bundle([self._cl, 'upload', large_file.name()])
+        dependency_uuid = self._run_bundle([self._cl, 'upload', large_file.name()])
```
percyliang, line 166: Can you document that this needs to be larger than the max memory on the system to test that things are being streamed?

percyliang: Also, one way to make tests run faster is to impose artificial memory limits on the process.

tests/stress/stress_test.py
```diff
@@ -303,7 +311,7 @@ def main():
 
     if args.heavy:
         print('Setting the heavy configuration...')
-        args.large_file_size_gb = 10
+        args.large_file_size_gb = 32
```
nelson-liu, line 316: could set this to `max(10, gigabytes of ram on the machine)`

teetone: I'm not sure if it's easy to get the memory of the machine programmatically since this script runs remotely.

PR #3016, opened 2020-11-04
Fix the bug of "Save" Button for worksheet not working

### LINKED ISSUE #3007: Save button does not work when the user edits the source for a worksheet
**Describe the bug**
image

**To Reproduce**
1. Click on 'edit source'
2. Make some changes
4. Click on 'save'

yuqijin:

The save button did not work. The reason is `setState` is an async function. We should first ensure the frontend state has already been updated properly before we make requests to the backend.

### Related issues

fixes #3007 

### Screenshots

ezgif com-gif-maker

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -1120,30 +1120,29 @@ class Worksheet extends React.Component {
             if (this.hasEditPermission()) {
                 var editor = ace.edit('worksheet-editor');
                 if (saveChanges) {
-                    // Use callback function to ensure the worksheet info will not be sent to the backend until the frontend state has finished updating
-                    this.setState(
-                        {
-                            ws: {
-                                ...this.state.ws,
-                                info: {
-                                    ...this.state.ws.info,
-                                    source: editor.getValue().split('\n'),
-                                },
+                    // Use promise here to ensure the worksheet info will not be sent to the backend until the frontend state has finished updating
+                    const setAsyncState = (newState) =>
+                        new Promise((resolve) => this.setState(newState, resolve));
+                    setAsyncState({
+                        ws: {
+                            ...this.state.ws,
+                            info: {
+                                ...this.state.ws.info,
+                                source: editor.getValue().split('\n'),
                             },
                         },
-                        () => {
-                            var rawIndex = editor.getCursorPosition().row;
-                            this.setState({
-                                inSourceEditMode: false,
-                                editorEnabled: false,
-                            }); // Needs to be after getting the raw contents
-                            if (saveChanges) {
-                                this.saveAndUpdateWorksheet(saveChanges, rawIndex);
-                            } else {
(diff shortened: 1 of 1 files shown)


epicfaace (approved):

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1120,25 +1120,30 @@ class Worksheet extends React.Component {
...
                 var editor = ace.edit('worksheet-editor');
                 if (saveChanges) {
-                    this.setState({
-                        ws: {
-                            ...this.state.ws,
-                            info: {
-                                ...this.state.ws.info,
-                                source: editor.getValue().split('\n'),
+                    // Use callback function to ensure the worksheet info will not be sent to the backend until the frontend state has finished updating
+                    this.setState(
```
epicfaace, line 1125: Actually, can we wrap this in a promise and then use `await` instead? Given that we want to refactor callbacks into promises anyway

epicfaace (changes requested):

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Actually for setState, I do not think we should wrap it into a promise and use await since it already provides a callback to take actions when setState is completed

epicfaace: What's the drawback of wrapping it in a promise and using await? It seems like it would make the code look at a bit cleaner, related to @Levilian 's work of changing callbacks -> promises?

something like this:

```
await new Promise(e => this.setState({...}, e));
```

If there's a drawback in doing it this way, we can keep using the callbacks, though.

yuqijin: I updated the PR to use `promise` instead since you mentioned in the offline discussion that we have a code style convention of using `promise` instead.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1121,26 +1121,30 @@ class Worksheet extends React.Component {
             if (this.hasEditPermission()) {
                 var editor = ace.edit('worksheet-editor');
                 if (saveChanges) {
-                    this.setState({
+                    // Use promise here to ensure the worksheet info will not be sent to the backend until the frontend state has finished updating
+                    const setAsyncState = (newState) =>
+                        new Promise((resolve) => this.setState(newState, resolve));
+                    setAsyncState({
```
epicfaace: We could just do this instead (using `async` / `await` instead of callbacks):

```
await new Promise(e => this.setState({
                        ws: {
                            ...this.state.ws,
                            info: {
                                ...this.state.ws.info,
                                source: editor.getValue().split('\n'),
                            },
                        },
                    }, e));
```

epicfaace (approved):

epicfaace (approved):

PR #2985, opened 2020-10-23
Resolve async upload files issue

Levilian:
Fix #2950

PR #3018, opened 2020-11-05
Scroll to newly opened Schema Editor/ New Run; Show entire Bundle Detail/ New Run 

### LINKED ISSUE #2079: show entire bundle details / new run on screen when expanded
When you use the keyboard and go to a bundle and then click enter, often the bundle details goes off the screen and you have to scroll with the mouse to see it.  This is also the case when you hit 'a r' to create a new run - the command might not even be visible....  So it'd be nice to scroll so that the details box fits on the screen.

At the same time, I'm a bit wary of thrashing around too much, especially on some corner case, so I think we should proceed with care.

### LINKED ISSUE #2988: Scroll to the newly opened schema editor
**Describe the bug**
When the user pressed the <Add Schema> UI button to open the schema editor, the page should scroll to the newly opened schema editor. Otherwise, it is confusing.

**Screenshots**
ezgif com-gif-maker (1)

### LINKED ISSUE #3066: New run appears at the top of the page when focusing on other items
image

**To Reproduce**
1. focus on a table item
2. use `a r` to create a new run
3. new run appears at the top of the page

**Expected behavior**
new run appears under the focused item

yuqijin:
There still exist several scrolling issues and bugs:

1. If the user upload items to an empty worksheet, then add other items, the newly added items will always be put on the top of the page. It is because no `sort_key` is assigned to the items uploaded before. Also, when the user uploads several files, sometimes the files will share the same `sort_key`, which might be caused by some racing conditions in the backend. The PR #2985 should have fixed this.
top-0

2. When the user add `new_run` or `schema`, the page should scroll to the newly opened items.

3. The entire bundle detail should be shown on the screen when being opened for the first time. 
detail-0

### Related issues

fixes #2079 
fixes #2988 
fixes #3066 

### Screenshots
**Add items to an empty worksheet**
top-1

**Scroll to `new_run`**
ewrun

**Scroll to `schema`**
schema

**Show entire bundle detail**
detail-1

### Checklist


frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -216,6 +216,7 @@ class BundleDetail extends React.Component<
 
         return (
             <ConfigurationPanel
+                ref={(node) => this.scrollToNewlyOpenedDetail(node)}
```
epicfaace, line 220: Instead of calling the `scrollToNewlyOpenedDetail` function directly in the `ref` function, it might be clearer to instead just store the ref and call `scrollToNewlyOpenedDetail` in `componentDidMount` (so it's clear that it's only called when the bundle detail is opened).

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -88,9 +96,7 @@ class NewUpload extends React.Component<{
             const createBundleData = getDefaultBundleMetadata(name || file.name, description);
             index += 1;
             let url = `/rest/bundles?worksheet=${ worksheetUUID }`;
-            if (after_sort_key) {
-                url += `&after_sort_key=${ after_sort_key }`;
-            }
+            url = this._setAfterSortKey(after_sort_key,url);
```
epicfaace: I think it might be clearer if we just remove the `_setAfterSortKey` function, and just do this:

```
url += `&after_sort_key=${ after_sort_key || -1 }`;
```

Also, is there a chance `after_sort_key` could be equal to 0? If so, we should replace the above by `isNaN(after_sort_key) ? -1: after_sort_key`

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -208,6 +208,30 @@ class SchemaItem extends React.Component<{
         }
     }
 
+    componentDidMount() {
+        if (this.props.create) {
```
epicfaace, line 228: Can you add a comment to this function explaining what this code is meant for?

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -456,6 +480,20 @@ class SchemaItem extends React.Component<{
                     if (this.props.create) return;
                     this.props.setFocus(this.props.focusIndex, 0);
                 }}
+                id={
+                    document.getElementById(
```
epicfaace: It's probably worth refactoring this into a separate function, since we repeat it above in `componentDidMount`, and it's >10 lines of code.

frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -317,16 +310,7 @@ class BundleRow extends Component {
...
-                                    const node = document.getElementById(
-                                        'codalab-worksheet-item-' +
-                                            this.state.focusIndex +
-                                            'subitem-' +
-                                            this.props.rowIndex +
-                                            '-detail',
-                                    );
-                                    node.scrollIntoView({ block: 'center' });
-                                }}
+                                onOpen={() => {}}
```
epicfaace, line 313: Why remove this?

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
yuqijin: it did not work if you just store the ref and call `scrollToNewlyOpenedDetail` in `componentDidMount`. It cannot be ensured that the ref is created before `componentDidMount`. I have already tried but failed.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
yuqijin: done

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
yuqijin: They are different actually. One is for setting id, the other is for getting id.

frontend/src/components/worksheets/items/TableItem/BundleRow.js
yuqijin: it was created by me for the bundle detail scrolling stuff. Since I switched a method now, I removed the old code

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
yuqijin: done

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
epicfaace: Right, but you could refactor this into a method like `getId()` and then replace the code in the other place with something like this:

```
const node = document.getElementById(this.getId());`
```

yuqijin: Refactored.

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
epicfaace: Ok, did you try creating the ref in the `constructor()` perhaps, instead?

If that doesn't work, please just add a comment here explaining that the ref is created only once, and that this is the only way to properly create the ref before `componentDidMount()`.

yuqijin: Yes, it did not work. 
I have updated the comment.

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -184,9 +182,7 @@ class NewUpload extends React.Component<{
             uploading: true,
         });
         let url = `/rest/bundles?worksheet=${ worksheetUUID }`;
-        if (after_sort_key) {
-            url += `&after_sort_key=${ after_sort_key }`;
-        }
+        url = this._setAfterSortKey(after_sort_key,url);
```
epicfaace: Did you mean to change this too? 

```suggestion
        url += `&after_sort_key=${ isNaN(after_sort_key) ? -1: after_sort_key }`;

```

epicfaace (approved):

PR #3022, opened 2020-11-05
Fix alembic migration by explicitly setting autoincrement to True

nelson-liu:

If you pull a fresh copy of codalab from master and run it locally, the database correctly has auto-increment. However, upgrading from an existing db removes the autoincrement. This seems like a bug on the alembic side (perhaps related to https://github.com/sqlalchemy/alembic/issues/413), but regardless, we have to deal with it and ensure that the DB schema when upgrading matches the DB schema when starting fresh from master.

I tested this by:

1. `git checkout f304d903` (before primary key change)
2. build server (just run `codalab_service start -bd`) , run some bundles
3. inspect the DB:

```
mysql> describe bundle_metadata;
+----------------+-------------+------+-----+---------+----------------+
| Field          | Type        | Null | Key | Default | Extra          |
+----------------+-------------+------+-----+---------+----------------+
| id             | int(11)     | NO   | PRI | NULL    | auto_increment |
| bundle_uuid    | varchar(63) | NO   | MUL | NULL    |                |
| metadata_key   | varchar(63) | NO   | MUL | NULL    |                |
| metadata_value | text        | NO   |     | NULL    |                |
+----------------+-------------+------+-----+---------+----------------+
4 rows in set (0.00 sec)

mysql> SELECT `AUTO_INCREMENT` FROM  INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'codalab_bundles' AND   TABLE_NAME   = 'bundle_metadata';
+----------------+
| AUTO_INCREMENT |
+----------------+
|            125 |
+----------------+
1 row in set (0.01 sec)
```

4. `git checkout 8e58f510` (this commit)
5. Rebuild server (just run `codalab_service start -bd`)
6. Inspect the db, ensure that int -> bigint, autoincrement remains, and the value of autoincrement remains constant:

```
mysql> describe bundle_metadata;
+----------------+-------------+------+-----+---------+----------------+
| Field          | Type        | Null | Key | Default | Extra          |
+----------------+-------------+------+-----+---------+----------------+
| id             | bigint(20)  | NO   | PRI | NULL    | auto_increment |
| bundle_uuid    | varchar(63) | NO   | MUL | NULL    |                |
| metadata_key   | varchar(63) | NO   | MUL | NULL    |                |
| metadata_value | text        | NO   |     | NULL    |                |
+----------------+-------------+------+-----+---------+----------------+
4 rows in set (0.00 sec)

mysql> SELECT `AUTO_INCREMENT` FROM  INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'codalab_bundles' AND   TABLE_NAME   = 'bundle_metadata';
+----------------+
| AUTO_INCREMENT |
+----------------+
|            125 |
+----------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/alembic/versions/2020110510_explicitly_set_autoincrement__4bac4855e710.py
+++ b/alembic/versions/2020110510_explicitly_set_autoincrement__4bac4855e710.py
@@ -1,0 +1,54 @@
+"""Explicitly set autoincrement to True
+
+Revision ID: 4bac4855e710
+Revises: fcb22a612d2a
+Create Date: 2020-11-05 10:27:43.989896
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '4bac4855e710'
+down_revision = 'fcb22a612d2a'
+
+
+def upgrade():
+    op.alter_column("bundle", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.BigInteger().with_variant(sa.Integer, "sqlite"), nullable=False, autoincrement=True)
+    op.alter_column("bundle_metadata", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.BigInteger().with_variant(sa.Integer, "sqlite"), nullable=False, autoincrement=True)
+    op.alter_column("bundle_dependency", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.BigInteger().with_variant(sa.Integer, "sqlite"), nullable=False, autoincrement=True)
+    op.alter_column("worksheet", 'id', type_=sa.BigInteger().with_variant(sa.Integer, "sqlite"), existing_type=sa.BigInteger().with_variant(sa.Integer, "sqlite"), nullable=False, autoincrement=True)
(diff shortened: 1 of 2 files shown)


nelson-liu:
I made a large set of changes in 6e3b7e1, because I noticed that the upgrade wasn't working if we were already on the bad revision (and had gotten there via an upgrade). It doesn't work in this case because the DB schema has the id as `bigint` type, but not auto-incrementing. Thus, whenever something tries to write to the DB, it's continuously (over)written to the row with id 0. Having a row with id 0 actually means that you can't just straightforwardly turn on auto-incrementing, since auto-incrementing tries to write to row 0. So, it tries to move id 0 to id 1, but there might already be a row with id 1, and thus things fail. See https://stackoverflow.com/a/6469632/14089059

To handle this case, I executed raw SQL to move rows with id 0 to the end of the table (the next highest ID), and then ALTERed the columns to be auto-incrementing. It's documented in a bit more detail in the code itself.

I'm a bit concerned about concurrent writes while we migrate (e.g., if we move the 0 row somewhere else, but the 0 row is re-written to befor we can do the migration). When upgrading, would be good to fully `codalab_service.py stop` the instance and then `codalab_service.py start`.

yiboliu (approved):

PR #3023, opened 2020-11-06
Consider the height of Worksheet Header and Nav Bar when deciding whether an element is on the screen or not

yuqijin:
### Screenshots

ezgif com-gif-maker

### Checklist


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -673,7 +673,10 @@ class Worksheet extends React.Component {
                 let elementHeight = element.height();
                 let screenScrollTop = $(window).scrollTop();
                 let screenHeight = $(window).height();
-                let scrollIsAboveElement = elementOffsetTop + elementHeight - screenScrollTop >= 0;
+                // 170 is the sum of height of WorksheetHeader and NavBar.
+                // Since they block the user's view, we should take them into account when calculating whether the item is on the screen or not.
+                let scrollIsAboveElement =
+                    elementOffsetTop + elementHeight - screenScrollTop - 170 >= 0;
```
epicfaace: Can we declare 170 in constants.js here? https://github.com/codalab/codalab-worksheets/blob/01b17c4499c0746af43433e922ff05cd860dd212/frontend/src/constants.js#L7

yuqijin: Thanks for the advice. updated.

frontend/src/constants.js
```diff
@@ -5,6 +5,7 @@ export const CODALAB_VERSION = '0.5.26';
 export const NAME_REGEX = /^[a-zA-Z_][a-zA-Z0-9_.-]*$/i;
 
 export const NAVBAR_HEIGHT = 60;
+export const HEADER_HEIGHT = 170;
```
yipenghe, line 8: So I guess the header height is hard coded 170 for the header component? Make sure to update there to use this constant as well so things don't go out of sync

yipenghe (approved):

PR #3025, opened 2020-11-06
changed schema column names

### LINKED ISSUE #2918: Remove placeholder text for schema editor
The placeholder text appears to be cut off on my screen.

image

We can probably just remove them, since the tooltips next to the "?" buttons on the header have more information about what each of these fields are, anyway.

Levilian:
Fix #2918


yiboliu (commented):
Probably you need to update the test

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -237,7 +236,7 @@ class SchemaItem extends React.Component<{
                         key={index}
                         style={{ padding: '5', fontSize: '16px', maxWidth: '100' }}
                     >
-                        {header}
+                        {headerText[header]}
```
epicfaace, line 239: I think we should still keep saying "generalized-path" and "post-processor", as that is the same terminology we use in the docs: https://codalab-worksheets.readthedocs.io/en/latest/Worksheet-Markdown/#schemas

Levilian: So we'll simply remove the placeholder text but keep the column titles?

Levilian: Actually this issue was brought up last week during the meeting and Percy said it's fine to have two different sets of terminology - this is for clarity's sake to display a more human-readable name on the front end.

yuqijin (approved):
looks good to me

PR #3026, opened 2020-11-06
Fix race condition in the Stress Test

teetone:

There is a race condition where the dependency can be deleted before the bundle finishes. This fix waits for the bundle to finish before cleanup.

### Checklist

### DIFF
--- a/tests/stress/stress_test.py
+++ b/tests/stress/stress_test.py
@@ -165,14 +165,16 @@ class StressTestRunner:
         large_file = TestFile('large_file', self._args.large_file_size_gb * 1000)
         dependency_uuid = self._run_bundle([self._cl, 'upload', large_file.name()])
         large_file.delete()
-        self._run_bundle(
+        uuid = self._run_bundle(
             [
                 self._cl,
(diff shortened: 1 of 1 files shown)


tests/stress/stress_test.py
```diff
@@ -165,14 +165,16 @@ def _test_large_bundle(self):
...
         dependency_uuid = self._run_bundle([self._cl, 'upload', large_file.name()])
         large_file.delete()
-        self._run_bundle(
+        uuid = self._run_bundle(
             [
                 self._cl,
                 'run',
-                'wc -c large_dependency',
                 'large_dependency:{}'.format(dependency_uuid),
+                'wc -c large_dependency',
```
nelson-liu, line 173: just curious, is the order change significant here?

nelson-liu (approved):

yiboliu (approved):

tests/stress/stress_test.py
teetone: I think it should be dependencies first.

PR #3027, opened 2020-11-06
Increase timeout when downloading dependencies

teetone:

When a worker downloads a large dependency, it will exceed the read timeout. This PR increases the default timeout and adds a separate timeout for downloading the bundle contents.

### Checklist


yiboliu (approved):

yiboliu:
As long as our customers are happy with this significant increase in timeout

nelson-liu:
this seems fine---there used to not even be a timeout, so i think just having one is the right thing.

nelson-liu (approved):

nelson-liu:
hm looks like 1d63f7a3c6c6656fab9b01912bf6cd9d06c0a0bf introduces some format violations that need to be fixed before this can be merged.

codalab/worker/bundle_service_client.py
```diff
@@ -179,5 +179,6 @@ def get_bundle_contents(self, uuid, path):
             '/bundles/' + uuid + '/contents/blob/' + path,
             headers={'Accept-Encoding': 'gzip'},
             return_response=True,
+            timeout_seconds=10 * 60,
```
epicfaace: Should this be set to `URLOPEN_TIMEOUT_SECONDS`?

nelson-liu: I think the intention was to make the timeout a bit more lenient for `/bundles/contents/blob/`

nelson-liu: Does this timeout mean that the download has to take less than `10*60` seconds?

epicfaace: Perhaps we set it to `URLOPEN_TIMEOUT_SECONDS * 2` instead then?

teetone: Updated

PR #3030, opened 2020-11-06
Fix CI

epicfaace:
- Fix failing master branch by fixing flake8 errors
- Ensure that the "CI" job is never skipped
- Fix mergify conditions so that branches are never automerged unless the "All CI tasks complete" check succeeds


yuqijin (approved):
looks good to me

PR #3031, opened 2020-11-06
uls - add default if no argument passed

### LINKED ISSUE #2972: Add a sane default for cl uls
When typing "cl uls", it should list all users instead of just providing an error message.

Also, the error message isn't very instructive -- maybe we can reword to give an example:

```
usage: cl uls [-h] [-f FIELD] keywords [keywords ...]
cl uls: error: the following arguments are required: keywords
```

Levilian:
Fix #2972


epicfaace (commented):
Can you add a test to ensure that the default is shown when no argument is passed?

codalab/lib/bundle_cli.py
```diff
@@ -171,7 +171,7 @@ def print_help(self, out_file=None):
     def error(self, message):
         # Adapted from original:
         # https://hg.python.org/cpython/file/2.7/Lib/argparse.py
-        if len(sys.argv) == 1 or (len(sys.argv) == 2 and sys.argv[1] in ['--help', '-h']):
+        if len(sys.argv) == 2 and sys.argv[1] in ['--help', '-h']:
```
epicfaace: I don't think we need this change?

codalab/lib/bundle_cli.py
```diff
@@ -3853,7 +3852,10 @@ def do_uls_command(self, args):
         Search for specific users.
         """
         client = self.manager.current_client()
-        users = client.fetch('users', params={'keywords': args.keywords})
+        if args.keywords is None:
+            users = client.fetch('users', params={'keywords': ''})
+        else:
+            users = client.fetch('users', params={'keywords': args.keywords})
```
epicfaace: This would be cleaner than the if statement:

```
users = client.fetch('users', params={'keywords': args.keywords or ''})
```

tests/cli/test_cli.py
```diff
@@ -1060,6 +1060,11 @@ def test(ctx):
         '3', _run_command([cl, 'uls', '.time_used_less_than=' + '1%', '-f', 'parallel_run_quota'])
     )
 
+    # check no argument passed
```
epicfaace: Be sure to specify _what_ we're expecting when no argument is passed.

```suggestion
    # List all users if no argument is passed
```

epicfaace (changes requested):

codalab/lib/bundle_cli.py
```diff
@@ -171,7 +171,7 @@ def print_help(self, out_file=None):
     def error(self, message):
         # Adapted from original:
         # https://hg.python.org/cpython/file/2.7/Lib/argparse.py
-        if len(sys.argv) == 1 or (len(sys.argv) == 2 and sys.argv[1] in ['--help', '-h']):
+        if len(sys.argv) == 1 or len(sys.argv) == 2 and sys.argv[1] in ['--help', '-h']:
```
epicfaace: I think you meant to switch this back to what it was before:

```suggestion
        if len(sys.argv) == 1 or (len(sys.argv) == 2 and sys.argv[1] in ['--help', '-h']):
```

codalab/lib/bundle_cli.py
```diff
@@ -3853,7 +3852,7 @@ def do_uls_command(self, args):
         Search for specific users.
         """
         client = self.manager.current_client()
-        users = client.fetch('users', params={'keywords': args.keywords})
+        users = client.fetch('users', params={'keywords': args.keywords or ''})
```
epicfaace, line 3861: Can you add a comment that this code makes `cl uls` with no arguments send an empty query?

epicfaace (approved):

epicfaace:
@Levilian I noticed a flaky test that caused CI to fail on this PR -- fixing it in #3089.

PR #3038, opened 2020-11-07
Fix warning of "Length of raw_to_block does not match length of raw_items"

### LINKED ISSUE #3036: Prod logs WARNINGs
We get many errors: "WARNING: Length of raw_to_block does not match length of raw_items"

yuqijin:

For `search` directives, when updating `raw_to_block`, the `raw_items` should be updated as well.

### Related issues

fixes #3036 

### Screenshots
No warnings

image

### Checklist


codalab/lib/worksheet_util.py
```diff
@@ -1054,12 +1054,12 @@ def flush_worksheets():
                         )
                         .data
                     )
-
+                    appended_directive_blocks_index = True
                     raw_to_block.append((len(blocks) - 1, 0))
                 else:
                     raise UsageError("unknown directive `%s`" % command)
                 # Add a placeholder for other cmds
```
epicfaace: Sorry, I don't recall -- what's the point of this `None` placeholder again? Can you expand this comment so it's clear to future readers of the code what it's used for?

yuqijin: Comment updated. Previously, when the item type is `directive`, `None` will always be added to `raw_to_block`, which is buggy since, for `addschema` / `search` cmd, some substantial items will be added to `raw_items`, so we need to update `raw_to_block` as well.

codalab/lib/worksheet_util.py
```diff
@@ -1054,12 +1054,12 @@ def flush_worksheets():
...
                         .data
                     )
-
+                    appended_directive_blocks_index = True
                     raw_to_block.append((len(blocks) - 1, 0))
                 else:
                     raise UsageError("unknown directive `%s`" % command)
-                # Add a placeholder for other cmds
-                if not appended_schema_blocks_index:
+                # Add a placeholder for other commands of a directive item since they do not represent substantial items
```
epicfaace:
```suggestion
                # Add an empty item placeholder for other directives, since they do not represent substantial items
```

epicfaace (approved):

Issue #3040, opened 2020-11-07
Switch from urllib to requests


nelson-liu:
urllib is antiquated, overly verbose, and often comes with poor defaults---I think it'd be a good move to switch to `requests`.

(I thought @teetone made an issue for this, but I can't find it...)

PR #3041, opened 2020-11-07
Replace urllib with requests in _upload_with_chunked_encoding

nelson-liu:

I was looking at this function recently and it was kind of inscrutable---replaced urllib with requests here, since it's much more readable and perhaps more robust as well. 

A small piece of #3040 .

### DIFF
--- a/tests/unit/worker/file_util_test.py
+++ b/tests/unit/worker/file_util_test.py
@@ -8,6 +8,7 @@ from codalab.worker.file_util import (
     get_file_size,
     gzip_bytestring,
     remove_path,
+    stream_chunks_from_fileobj,
     tar_gzip_directory,
     un_gzip_stream,
     un_bz2_file,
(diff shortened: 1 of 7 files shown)


tests/unit/worker/file_util_test.py
```diff
@@ -58,6 +59,15 @@ def test_bz2_file(self):
     def test_gzip_bytestring(self):
         self.assertEqual(un_gzip_bytestring(gzip_bytestring(b'contents')), b'contents')
 
+    def test_stream_chunks_from_fileobj(self):
```
epicfaace, line 62: Would it be possible to add a unit test for _upload_with_chunked_encoding?

epicfaace:
I'm not going to work on #3124 for the time being, so let's just keep this PR open for now.

epicfaace:
Actually, I'm working on #3124.

PR #3045, opened 2020-11-09
Make gzip_file nonblocking / return faster

nelson-liu:

Despite the fix in #3012 , `gzip_file` takes quite a bit of time to return---this is in contrast to the previous implementation (that used a subprocess out to `gzip -c -n`). This is a big problem for large files (e.g., dependencies over 3G in size), since the request might time out before this function returns.

This is a fix to make it return instantly by explicitly setting up a buffer for it to write compressed data to / for a client to read compressed data from.

I also used the `BytesBuffer` to improve `un_gzip_stream`, since right now it does a bunch of expensive string concatenation (and thus, string copying).


epicfaace (commented):
This looks good, but at this point, do you think it might be cleaner to use something like `zlib.compressobj` instead? https://www.programcreek.com/python/example/1511/zlib.compressobj

The docs for this method say, "Returns a compression object, to be used for compressing data streams that won’t fit into memory at once." https://docs.python.org/3/library/zlib.html#zlib.compressobj

nelson-liu:
Maybe I'm getting this wrong, but I thought the `Compression` object is simply used to compress chunks? e.g., if you look at `un_gzip_stream` https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/file_util.py#L242-L289 , it uses `decompressobj` but still has to write a bunch of boilerplate to wrap the compressed data into a file-like.

In other words, with `compressobj`, we'd still need to wrap it around a file-like i think. So, it's more of a replacement for `gzip.GzipFile` in the diff, versus the new `GzipStream` class; feels like a lateral move in complexity to me. Let me know if I'm misunderstanding, though.

epicfaace (approved):
Yeah, that makes sense. Thanks!

PR #3049, opened 2020-11-10
Add flag to exit workermanager after X seconds have elapsed since the last launched worker

nelson-liu:

We don't want people to run their workermanagers forever if they aren't using them---this pollutes `sc`, and prevents them from getting the latest version of the workermanager. This flag optionally exits the worker manager after X seconds have elapsed since the last worker was launched.


percyliang (commented):
Could we have it just restart itself so it picks up new updates so people can just forget about it?

PR #3050, opened 2020-11-10
Add option to restart WorkerManager processes after X seconds

nelson-liu:

Continuation of #3049 .

To test (doesn't have to be on the cluster):
```
cl-worker-manager --restart-after-seconds 10 slurm-batch --partition jag-lo
```

You should see a `Restarting...` message followed by the `Started worker manager` string after ~10 seconds

### DIFF
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -2,7 +2,9 @@ from collections import namedtuple
 import http
 import logging
 import os
+import psutil
 import socket
+import sys
 import time
 import traceback
 import urllib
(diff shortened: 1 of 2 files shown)


codalab/worker_manager/worker_manager.py
```diff
@@ -21,6 +23,20 @@
 WorkerJob = namedtuple('WorkerJob', ['active'])
 
 
+def restart():
+    """
+    Restarts the current program, cleaning up file objects and descriptors
+    """
+    try:
+        p = psutil.Process(os.getpid())
+        for handler in p.open_files() + p.connections():
```
percyliang, line 32: What things are open that we need to cleanup?

percyliang (approved):
LGTM!

codalab/worker_manager/worker_manager.py
nelson-liu: As of now, there shouldn't be any I think...but I added this because I figured it couldn't hurt / might future-proof things a bit. Also, even if this block completely fails, the restarting should still work fine.

Issue #2993, opened 2020-10-26
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'


teetone:
**Describe the bug**
Request `POST /account/login` failed with the following stack trace:

```
POST /account/login

Query params:

Form params:
username='webmaster'
password='<censored>'

JSON body:
<none>

Local variables:
params={'error': 'Login/password did not match.', 'next': None}
redirect_uri=None

Traceback (most recent call last):
File "/opt/codalab-worksheets/codalab/server/rest_server.py", line 106, in wrapper
return callback(*args, **kwargs)
File "/usr/local/bin/bottle.py", line 1748, in wrapper
rv = callback(*a, **ka)
File "/opt/codalab-worksheets/codalab/server/json_api_plugin.py", line 18, in wrapper
result = callback(*args, **kwargs)
File "/opt/codalab-worksheets/codalab/rest/account.py", line 46, in do_login
error_uri, {"error": "Login/password did not match.", "next": success_uri}
File "/opt/codalab-worksheets/codalab/lib/server_util.py", line 173, in redirect_with_query
return redirect(redirect_uri + '?' + urllib.parse.urlencode(params))
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

**Codalab Information**
 - *Codalab CLI Version (if on CLI)*: 0.5.25

**Expected behavior**

We should fail gracefully.

yiboliu:
Not happening anymore. My suggestion is to give a default value to the `redirect_uri` when it's None in the #3054 . Feel free to comment your thoughts.

PR #3054, opened 2020-11-10
Handle the login failure in a graceful way

yiboliu:
When the login credentials don't match and error_uri is None, the program is failing in a way that involves type matching, whereas it should indicate explicitly that error_uri is None. I give the error_uri a default value, when it's None to handle it. 

### Related issues
#2993 

### Checklist

### DIFF
--- a/codalab/rest/account.py
+++ b/codalab/rest/account.py
@@ -42,6 +42,8 @@ def do_login():
 
     user = local.model.get_user(username=username)
     if not (user and user.check_password(password)):
+        if error_uri is None:
+            error_uri = '/'
         return redirect_with_query(
             error_uri, {"error": "Login/password did not match.", "next": success_uri}
         )


epicfaace (commented):
What does the failure message actually look like? Can we add a test in test_cli.py to ensure it looks like what we expect?

yiboliu:
> What does the failure message actually look like? Can we add a test in test_cli.py to ensure it looks like what we expect?

It just says `Local variables:
params={'error': 'Login/password did not match.', 'next': None}
redirect_uri=None`. When `redirect_uri` is None, there will be a type mismatching error in string concatenation. So what I'm doing here is to give `redirect_uri` a default string value.

codalab/rest/account.py
```diff
@@ -42,6 +42,8 @@ def do_login():
 
     user = local.model.get_user(username=username)
     if not (user and user.check_password(password)):
+        if error_uri is None:
```
teetone, line 45: Do we know for what case `error_uri` will be `None`?

Issue #3058, opened 2020-11-12
Version conflict of urllib3


yiboliu:
**Describe the bug**
```Traceback (most recent call last):
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/bin/cl", line 33, in <module>
    sys.exit(load_entry_point('***', 'console_scripts', 'cl')())
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/bin/cl", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 105, in load
    module = import_module(match.group('module'))
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/runner/work/***-worksheets/***-worksheets/***/bin/cl.py", line 3, in <module>
    from ***.lib.bundle_cli import BundleCLI
  File "/Users/runner/work/***-worksheets/***-worksheets/***/lib/bundle_cli.py", line 38, in <module>
    import ***.model.bundle_model as bundle_model
  File "/Users/runner/work/***-worksheets/***-worksheets/***/model/bundle_model.py", line 54, in <module>
    from ***.rest.util import get_group_info
  File "/Users/runner/work/***-worksheets/***-worksheets/***/rest/util.py", line 11, in <module>
    from ***.lib import bundle_util
  File "/Users/runner/work/***-worksheets/***-worksheets/***/lib/bundle_util.py", line 7, in <module>
    from ***.client.json_api_client import JsonApiClient, JsonApiRelationship  # NOQA: F401
  File "/Users/runner/work/***-worksheets/***-worksheets/***/client/json_api_client.py", line 10, in <module>
    from ***.worker.rest_client import RestClient, RestClientException
[...]

PR #3060, opened 2020-11-12
Add pip check to test.yml

yiboliu:
Check if there's any broken version requirements

### Related issues
#3058 

### Checklist

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -9,9 +9,15 @@ if ! [ -e venv ]; then
   python3 -m pip install virtualenv
   python3 -m virtualenv -p python3 venv
 fi
-venv/bin/pip install -r requirements-server.txt
-venv/bin/pip install -r requirements.docs.txt
-venv/bin/pip install -r requirements.dev.txt
+venv/bin/pip install -r requirements-server.txt --no-cache
+venv/bin/pip install -r requirements.docs.txt --no-cache
+venv/bin/pip install -r requirements.dev.txt --no-cache
+
+venv/bin/pip check || let val=$?
+if [ $val -ne 0 ] ; then
+  echo "Broken version requirements detected";
+  # Should I exit with 1 here?
+fi
 
 venv/bin/pip install -e .
 
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -13,7 +13,7 @@ venv/bin/pip install -r requirements-server.txt --no-cache
 venv/bin/pip install -r requirements.docs.txt --no-cache
 venv/bin/pip install -r requirements.dev.txt --no-cache
 
-venv/bin/pip check 
+venv/bin/pip check
 
 venv/bin/pip install -e .
 
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -13,11 +13,7 @@ venv/bin/pip install -r requirements-server.txt --no-cache
 venv/bin/pip install -r requirements.docs.txt --no-cache
 venv/bin/pip install -r requirements.dev.txt --no-cache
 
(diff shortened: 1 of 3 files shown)


nelson-liu:
hm, this should be failing, but it isn't, because i think it's running outside of the virtualenv that is set up by `pre-commit.sh`. Maybe we should add pip check to `pre-commit.sh` instead?

Also, might be good to do the `pip install`s in `pre-commit.sh` with `--no-cache`

pre-commit.sh
```diff
@@ -9,9 +9,15 @@ if ! [ -e venv ]; then
...
   python3 -m virtualenv -p python3 venv
 fi
-venv/bin/pip install -r requirements-server.txt
-venv/bin/pip install -r requirements.docs.txt
-venv/bin/pip install -r requirements.dev.txt
+venv/bin/pip install -r requirements-server.txt --no-cache
+venv/bin/pip install -r requirements.docs.txt --no-cache
+venv/bin/pip install -r requirements.dev.txt --no-cache
+
+venv/bin/pip check || let val=$?
```
nelson-liu: you don't need the `||` or anything below since the script runs with `set -e`. When `pip check` exits with 1, the whole script will.

nelson-liu:
Looks like this still fails because there's a conflict in the package version for mock, which we pin and apache beam also pins.

pre-commit.sh
```diff
@@ -9,9 +9,11 @@ if ! [ -e venv ]; then
...
   python3 -m virtualenv -p python3 venv
 fi
-venv/bin/pip install -r requirements-server.txt
-venv/bin/pip install -r requirements.docs.txt
-venv/bin/pip install -r requirements.dev.txt
+venv/bin/pip install -r requirements-server.txt --no-cache
+venv/bin/pip install -r requirements.docs.txt --no-cache
+venv/bin/pip install -r requirements.dev.txt --no-cache
+
+venv/bin/pip check
```
epicfaace: Should we run `pip check` after packages are installed with `pip install -e .`?

pre-commit.sh
```diff
@@ -9,12 +9,14 @@ if ! [ -e venv ]; then
   python3 -m pip install virtualenv
   python3 -m virtualenv -p python3 venv
 fi
-venv/bin/pip install -r requirements-server.txt
-venv/bin/pip install -r requirements.docs.txt
-venv/bin/pip install -r requirements.dev.txt
+venv/bin/pip install -r requirements-server.txt --no-cache
```
epicfaace: why did you add --no-cache? isn’t it faster / easier if we don’t have that argument in?

requirements-server.txt
```diff
@@ -11,7 +11,7 @@ boto3==1.15.7
 azure-batch==9.0.0
 
 # For testing
-mock==3.0.5
+mock==2.0.0
```
epicfaace, line 14: Why do we need to downgrade mock? Was this causing dependency conflict issues?

yiboliu:
```
apache-beam 2.25.0 requires mock<3.0.0,>=1.0.1, but you'll have mock 3.0.5 which is incompatible.
```

pre-commit.sh
nelson-liu: I think that without `--no-cache`, you might inadvertently install an older version of the package if it happens to be on your system. Might also need `-U` on these installs, so the venv automatically gets updated if the pins are relaxed..

yiboliu: We have version specifications right? Each module is specified to a particular version.

epicfaace (approved):

PR #3062, opened 2020-11-13
refactor certain promises into async/await

Levilian:
#2987


frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -174,9 +174,8 @@ class BundleDetail extends React.Component<
                         }
                     }.bind(this),
                 );
-                $.when.apply($, fetchRequests).then(() => {
-                    this.setState(stateUpdate);
-                });
+                $.when.apply($, fetchRequests);
```
epicfaace:
```suggestion
                await $.when.apply($, fetchRequests);

```

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -193,68 +193,67 @@ class NewUpload extends React.Component<{
...
-                                    jqHXR.responseText,
-                                    'refresh and try again.',
-                                ),
-                            );
-                            this.props.onUploadFinish();
+            });
+    
+        } catch (error) {
+            this.clearProgress();
+            alert(createAlertText(url, error.responseText));
```
epicfaace, line 227: Shouldn't the `'refresh and try again.'` and `onUploadFinish` be called here? I think you might have switched the two `catch` statements.

frontend/src/components/worksheets/WorksheetNameSearch.js
```diff
@@ -19,8 +19,9 @@ export default class extends React.Component {
         }
 
         try {
-            const response = await fetch(`/rest/worksheets?specs=${name}`).then((e) => e.json());
-            const uuid = response.data[0].id;
+            const response = await fetch(`/rest/worksheets?specs=${name}`);
```
epicfaace: I think in this case, we don't necessarily need to add the additional `await` here. Personally, `.then(e => e.json())` seems simpler / more readable than having to declare a new variable called `json`.

frontend/src/serviceWorker.js
```diff
@@ -29,7 +29,7 @@ export function register(config) {
             return;
         }
 
-        window.addEventListener('load', () => {
+        window.addEventListener('load', async () => {
```
epicfaace: I think this file is generated by create-react-app, so we shouldn't have to modify it -- let's leave it as is.

frontend/src/util/cli_utils.js
```diff
@@ -24,17 +24,15 @@ import $ from 'jquery';
...
-    })
-    .always(function() {
-        terminal.resume();
-        if (!isEnabled) {
-            terminal.disable();
-        }
-        self.props.reloadWorksheet();
-    });
+    } finally {
+    terminal.resume();
```
epicfaace: add indent here

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
Levilian: No - here we're simply creating a bundle. Sending the data comes after the successful creation of the bundle.

frontend/src/components/Bundle/Bundle.js
```diff
@@ -148,9 +148,8 @@ class Bundle extends React.Component<
                             }
                         }.bind(this),
                     );
-                    $.when.apply($, fetchRequests).then(() => {
```
epicfaace, line 151: I'm actually a bit confused about what this logic is actually doing. Why is it calling `$.when` with `fetchRequests` as the argument twice?

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -174,9 +174,8 @@ class BundleDetail extends React.Component<
                         }
                     }.bind(this),
                 );
-                $.when.apply($, fetchRequests).then(() => {
```
epicfaace, line 177: Similar question here

frontend/src/components/Bundle/Bundle.js
Levilian: https://medium.com/sungthecoder/making-multiple-ajax-calls-and-deciphering-when-apply-array-b35d1b4b1f50
So here we're doing multiple AJAX calls and only go on to the then command if all of them are resolved.

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
Levilian: https://medium.com/sungthecoder/making-multiple-ajax-calls-and-deciphering-when-apply-array-b35d1b4b1f50
So here we're doing multiple AJAX calls and only go on to the then command if all of them are resolved.

frontend/src/components/Bundle/Bundle.js
epicfaace: Ah, I see. I think it's clearer to instead just do `Promise.all(...fetchRequests)` as that is native to JS and doesn't require jQuery.

I'm still confused as to why we also have `$.when(fetchRequests)` on line 154, though? That's what I meant about "calling $.when with fetchRequests as the argument twice".

Feel free to ping me on Slack so we can discuss this further.

epicfaace (approved):

Issue #542, opened 2016-07-29
Support time zones consistently and comprehensively


kashizui:
Currently, timestamps are created inconsistently, using `datetime.datetime.utcnow()` in some places and `time.time()` in others. We need to
- [ ] standardize how we create timestamps throughout CodaLab (e.g. always use UTC when we store timestamps)
- [ ] localize timestamps to the user's time zone when displaying them in the CLI and on the web interface

kdexd:
Hi @kashizui:
I did some quick path finding and discovered all occurences of `time` module and it's methods. For one trial and error iteration, I have replaced all the storage methods and comparisons to `datetime.datetime` format, with UTC. `time` module is now only used to sleep.

Any further changes on current patch will be made within the lines of this patch.
- Patch: https://github.com/codalab/codalab-cli/compare/develop...karandesai-96:fix-issue-542
- Build: https://circleci.com/gh/karandesai-96/codalab-cli/37

Please have a look and let me know whether it can be worked upon or it isn't feasible. Thanks :smile:

kashizui:
This looks good so far, thanks!

kdexd:
That's great, although it will required some changes to inculcate this new way. `datetime.datetime` objects can be easily stored in _SQL_ database, but they are not _JSON_ serializable. Although `datetime.datetime` provides a method to convert them into a string using `isoformat()`. Many popular _APIs_ provide timestamp info through isoformat strings, github _API_ being one of them. We will have to write a small wrapper function to construct a `datetime` object out of _ISO_ format string but that is super easy. I am letting it stay for now, and we will discuss more on this once my existing PRs are suitable to merge.

kashizui:
That's a good point, these changes will probably fail as they are now.

We use `marshmallow` to de/serialize objects in the REST API, and it supports DateTime fields. We will have to look at the rest of the system to figure out how to deal with each case individually - we may want to convert the datetimes into integers when necessary.

PR #3073, opened 2020-11-18
Add "date_created" and "date_last_modified" to worksheet 

### LINKED ISSUE #3067: Add `date` property to a worksheet
Currently, a worksheet does not have the property of "created date". 
We want to have this property for sorting the worksheets on the new landing page and show it to the user as well.

code ptr: https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/worksheets.py

yuqijin:

Need to record "created date" and "last modified date" for each worksheet.
**Backend** 
1. Use the `DateTime` type to store the two fields. All UTC time.
2. When a worksheet is inserted into the database, both `created date` and `last modified date` will be set to the current timestamp.
3. `created date` will never change once it has been set.
4. `last modified date` will be updated when the worksheet metadata changes or its worksheet_items changes.
5. For previous worksheets which do not have "created date" and "last modified date" recorded (Mysql does not store information for each row), follow the same logic as `user`'s `date_joined` and `last_login`
- created date: stored as "0000-00-00 00:00:00", since the field can't be NULL.
- last modified date: stored as NULL. 

**Frontend**
1. Show the last modified date in the Worksheet Header, under the Worksheet title.
2. Convert the UTC time to the local time zone.

### Related issues

fixes #3067 
related: #542 

### Screenshots
**Database**
Screen Shot 2020-11-18 at 2 51 15 AM

**Frontend**

ezgif com-gif-maker

### Checklist


codalab/model/tables.py
```diff
@@ -104,6 +104,12 @@
         'frozen', DateTime, nullable=True
     ),  # When the worksheet was frozen (forever immutable) if it is.
     Column('is_anonymous', Boolean, nullable=False, default=False),
+    Column(
+        'date_created', DateTime, nullable=False
+    ),  # When the worksheet was created; Set to zero date if the worksheet created before v0.5.31; In default will be set to current_timestamp
```
epicfaace:
```suggestion
    ),  # When the worksheet was created; Set to zero date if the worksheet created before v0.5.31; Set to current timestamp by default

File: codalab/model/tables.py [hunk 2]
```diff
@@ -104,6 +104,12 @@
...
+    Column(
+        'date_last_modified', DateTime
+    ),  # When the worksheet was last modified; Set to null if the worksheet created before v0.5.31; In default will be set to current_timestamp
```
epicfaace:
```suggestion
    ),  # When the worksheet was last modified; Set to null if the worksheet created before v0.5.31; Set to current_timestamp by default
```

codalab/model/tables.py
epicfaace: I'm not sure if zero date would be a good default here, because zero date in Unix time is 1970-01-01T00:00:00Z. Do you think perhaps we should just make it nullable and set earlier worksheets' date created to null? Or is it better to keep it at zero date?

alembic/versions/2020111806_add_created_date_and_last__0602bcb323f4.py
```diff
@@ -0,0 +1,24 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '0602bcb323f4'
+down_revision = '4bac4855e710'
+
+
+def upgrade():
+    op.add_column('worksheet', sa.Column('date_created', sa.DateTime(), nullable=False))
```
epicfaace, line 18: Shouldn't we include the default value (zero date) somewhere in the migration? Otherwise, I'm not sure where the default could be specified.

codalab/model/bundle_model.py
```diff
@@ -1518,13 +1518,16 @@ def new_worksheet(self, worksheet):
         """
         Save the given (empty) worksheet to the database. On success, set its id.
         """
+        now = datetime.datetime.utcnow()
```
epicfaace, line 1521: This creates a naive datetime. Perhaps it would be better to create a timezone-aware datetime in UTC using `datetime.now()` (this would involve setting the `timezone` attribute in the sqlalchemy `DateTime` class to UTC)?

codalab/model/bundle_model.py
```diff
@@ -1689,6 +1701,10 @@ def update_worksheet_metadata(self, worksheet, info):
             worksheet.owner_id = info['owner_id']
         if 'title' in info:
             info['title'] = self.encode_str(info['title'])
+        # Always update worksheet's last modified date to current timestamp(UTC)
```
```suggestion
        # Always update worksheet's last modified date to current timestamp (UTC)
```

codalab/model/bundle_model.py
```diff
@@ -1676,6 +1680,14 @@ def update_worksheet_items(self, worksheet_uuid, last_item_id, length, new_items
             if result.rowcount < length:
                 raise UsageError('Worksheet %s was updated concurrently!' % (worksheet_uuid,))
             self.do_multirow_insert(connection, cl_worksheet_item, new_item_values)
+        self.update_worksheet_last_modified_date(worksheet_uuid)
+
+    def update_worksheet_last_modified_date(self, worksheet_id):
+        """
+        Update worksheet's last modified date to now.
```
epicfaace, line 1687: Can you add a comment here about how calling `update_worksheet_metadata` with an empty argument essentially is equivalent to updating the last modified date?

codalab/model/orm_object.py
```diff
@@ -38,6 +38,7 @@ class ORMObject(object):
         Tuple[str, str, str, str, str, str],
         Tuple[str, str, str, str],
         Tuple[str, str, str, str, str, str, str],
+        Tuple[str, str, str, str, str, str, str, str],
```
epicfaace: Could we just replace this entire thing with `Tuple[str, ...]` which works for all variable-length tuples (see https://github.com/python/mypy/issues/184#issuecomment-70385687)? The current way we type this property doesn't seem to be really readable / useful and just makes it harder to update columns.

codalab/model/bundle_model.py
yuqijin: Based on the discussion in the meeting, we'd prefer to store UTC time in DB, and convert it to local time in the frontend.

codalab/model/tables.py
yuqijin: I set both date's default value to be null.

alembic/versions/2020111806_add_created_date_and_last__0602bcb323f4.py
yuqijin: I set the default to be null so we do not need to specify it here. :)

codalab/model/bundle_model.py
yuqijin: Thanks for the advice. Added.

codalab/model/orm_object.py
yuqijin: Thanks for the suggestion. I updated the type.

codalab/model/bundle_model.py
epicfaace: Sounds good

codalab/rest/schemas.py
```diff
@@ -124,6 +124,8 @@ class WorksheetSchema(Schema):
     title = fields.String()
     frozen = fields.DateTime(allow_none=True)
     is_anonymous = fields.Bool()
+    date_created = fields.DateTime()
```
epicfaace: Should we do this too?:

```suggestion
    date_created = fields.DateTime(allow_none=True)
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -214,6 +214,21 @@ class Worksheet extends React.Component {
         return wsItemsIdDict;
     }
 
+    convertDatetimeToLocalDatetime = (date) => {
```
epicfaace: Let's use the native `Date` functions for this instead.

Can you see if you can get the backend to return the date in ISO 8601 format (like "2020-11-18T06:43:05Z")?

Then, we can just do this:
```
date = new Date("2020-11-18T06:43:05Z");
return date.toString();
```

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -67,6 +68,13 @@ export default ({
                                 onChange={() => reloadWorksheet()}
                                 allowASCII={true}
                             />
+                            {/*Hide invalid last modified date*/}
+                            {info && info['date_last_modified'] ? (
+                                <div style={{ fontWeight: 'normal', fontSize: 'small' }}>
+                                    Last Modified Date:{' '}
+                                    {convertDatetimeToLocalDatetime(info['date_last_modified'])}
```
epicfaace: If we follow the suggestion I mentioned above, we could maybe even get rid of `convertDatetimeToLocalDatetime` altogether, and just replace by inlining it: `new Date(...).toString()`

epicfaace (changes requested):

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: I think the difference is the current method does the format processing in the frontend while your purposed method does the processing in the backend? Why do we want to do the processing in the backend?

codalab/rest/schemas.py
yuqijin: good catch

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Change applied.

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
yuqijin: Change applied

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1351,6 +1353,14 @@ class Worksheet extends React.Component {
         return count;
     }
 
+    handleDateFormat(date) {
+        // Append 'Z' to indicate the time is UTC time.
```
```suggestion
        // Append 'Z' to convert the time to ISO format (in UTC).
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1351,6 +1353,14 @@ class Worksheet extends React.Component {
...
+        if (date && date[date.length - 1] !== 'Z') {
```
epicfaace: Is there a chance the date will sometimes have Z and sometimes not Z at the end? If not, we can just remove this if statement and always append "Z".

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -67,6 +67,13 @@ export default ({
                                 onChange={() => reloadWorksheet()}
                                 allowASCII={true}
                             />
+                            {/*Hide invalid last modified date*/}
```
epicfaace:
```suggestion
                            {/* Hide invalid last modified date */}
```

epicfaace: When would last modified be invalid?

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Sometimes the date might be undefined. So I keep the check for `if(date)` here.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: change applied

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
yuqijin: change applied

yuqijin: the date might be an undefined one

epicfaace (commented):
@yuqijin this looks good! One more thing -- can you add some tests in `test_cli.py`? I'd imagine you could probably create a test in which you create a worksheet, then check its date modified / date created, then modify that worksheet, then ensure that date_modified changed and date_created didn't change.

tests/unit/rest/worksheets_test.py
```diff
@@ -28,3 +30,10 @@ def test_create(self):
                 }
             ],
         )
+        # Verify the date_created and date_last_modified field has been set to the current time for a new created worksheet
+        worksheet_info = self.app.get('/rest/interpret/worksheet/' + worksheet_id).json
```
epicfaace, line 34: Can you use freezegun to manually set the time (see this for an example: https://github.com/codalab/codalab-worksheets/blob/d403edc937ff598d930bbc13b53cb3922af9c4c2/tests/unit/rest/bundles_test.py#L5)

This way, you wouldn't have to use assertAlmostEqual -- you can just use assertEqual (and then the test would be less flaky since it doesn't have to take one second).

epicfaace (commented):
Thanks for the test -- one more change.

epicfaace (approved):

tests/unit/rest/worksheets_test.py
epicfaace: Actually, this isn't super important -- we could maybe just do this in a later PR.

PR #3074, opened 2020-11-20
Suppress printing the stacktrace of login error

yiboliu:
Suppress extra stack traces for errors occurring when login credentials are incorrect

### Related issues
#2316

### Screenshots
image

### Checklist

### DIFF
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -6,7 +6,7 @@ import json
 import urllib.request
 import urllib.parse
 import urllib.error
-from codalab.common import URLOPEN_TIMEOUT_SECONDS
+from codalab.common import URLOPEN_TIMEOUT_SECONDS, LoginPermissionError
 
 
 # TODO(sckoo): clean up auth logic across:
@@ -66,5 +66,6 @@ class RestOAuthHandler(object):
         except urllib.error.HTTPError as e:
             if e.code == 401:
                 return None
-            print("Invalid username or password.")
-            raise
+            if e.code == 404:
+                raise LoginPermissionError("Invalid username or password.")
+            raise e
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -66,4 +66,5 @@ class RestOAuthHandler(object):
         except urllib.error.HTTPError as e:
             if e.code == 401:
                 return None
+            print("Invalid username or password.")
             raise
--- a/codalab/server/auth.py
+++ b/codalab/server/auth.py
@@ -66,6 +66,6 @@ class RestOAuthHandler(object):
         except urllib.error.HTTPError as e:
             if e.code == 401:
                 return None
-            if e.code == 404:
+            elif e.code == 404:
                 raise LoginPermissionError("Invalid username or password.")
             raise e
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -147,27 +147,24 @@ class WorkerManager(object):
         return command
 
     def run_loop(self):
-        while True:
-            try:
-                self.run_one_iteration()
-            except (
-                urllib.error.URLError,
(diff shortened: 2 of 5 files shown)


epicfaace:
@yiboliu I'm a bit confused as to the difference between this PR and #3054. Could you please consolidate?

yiboliu:
> @yiboliu I'm a bit confused as to the difference between this PR and #3054. Could you please consolidate?

Oh, it's actually two different problems. This one is solving the problem occurred if people fail to enter correct login info for cl-worker-manager, to print the correct message to user in cli. #3054 is purely for backend handling when login info is incorrect. Sometimes the error_uri is None and my solution is to give it a default value.

tests/cli/test_cli.py
```diff
@@ -2389,6 +2389,29 @@ def test_nonexistent(ctx):
     _run_command([cl, 'work', 'nonexistent::'], expected_exit_code=1)
 
 
+@TestModule.register('worker_manager')
+def test_false_login(ctx):
```
teetone: Can we rename this test to `test_incorrect_login`?

codalab/worker_manager/worker_manager.py
```diff
@@ -147,25 +147,24 @@ def build_command(self, worker_id: str, work_dir: str) -> List[str]:
...
-                # Sometimes, network errors occur when running the WorkerManager . These are often
-                # transient exceptions, and retrying the command would lead to success---as a result,
-                # we ignore these network-based exceptions (rather than fatally exiting from the
-                # WorkerManager )
-                traceback.print_exc()
-            if self.args.once:
-                break
-            logger.debug('Sleeping {} seconds'.format(self.args.sleep_time))
-            time.sleep(self.args.sleep_time)
+        try:
```
teetone: If you get rid of the `while True` loop here, it will only one iteration. Is this intended? That doesn't seem right.

codalab/server/auth.py
```diff
@@ -66,4 +66,6 @@ def _make_token_request(self, data):
         except urllib.error.HTTPError as e:
             if e.code == 401:
                 return None
-            raise
+            if e.code == 404:
```
teetone: Should it be `elif` here?

teetone (changes requested):

codalab/server/auth.py
yiboliu: logically it should be no difference right?

codalab/worker_manager/worker_manager.py
yiboliu: That one iteration could still produce errors and we need to handle them. What's the problem?

teetone: The worker manager loops infinitely checking if it should spawn a worker. By getting rid of the `while True` loop it will only do this once.

codalab/server/auth.py
teetone: Yes logically it makes no difference, but since the two cases are mutually exclusive, it makes it clearer that only one of them will execute.

codalab/worker_manager/worker_manager.py
teetone: Just to be more clear, logging into CodaLab is not the only functionality of `run_one_iteration()`. I recommend taking a look at what the method does.

yiboliu: Gotcha, I put back the loop and added some additional changes

percyliang: I think this came from a discussion where if login fails, then you just exit without trying indefinitely, so it's easier to test the code.

yiboliu: Yes I remember it. So the most recent code only allows login attempt once and maintain other functionalities of `run_one_iteration()`

teetone (approved):

teetone (approved):

PR #3077, opened 2020-11-20
Fix flakiness of upload-download test

epicfaace:
Sometimes, the contents can vary in order; this PR changes the logic in the test to check only for contents without depending on the order in which the contents are retrieved.


yuqijin (approved):

yuqijin (changes requested):
The test still failed

yuqijin (approved):

Issue #3003, opened 2020-10-30
View CPU / memory / network load stats for bundles


epicfaace:
This is useful so you can tell if you need to scale up a bundle.

Some stats that might be useful:
- Max CPU utilization
- Min CPU utilization
- Average CPU utilization
- Min / max / average memory used

Maybe we could even show graphs in the future.

Made a [doc](https://docs.google.com/document/d/1qvgttYm26aPyOVcoRO6_3n3rts0UB3NZIS5TG5nTnJk/edit) to track the solution evolves.

PR #3081, opened 2020-12-01
Get the stats data from docker

yiboliu:
Add new metadata entries, cpu_usage and memory_limit, to the metadata of each running bundle.
 

### Related issues
#3003 

### Screenshots
image

### Checklist


codalab/bundles/run_bundle.py
```diff
@@ -29,8 +29,10 @@ class RunBundle(DerivedBundle):
                                                                     'codalab/default-cpu:latest) we wish to use.', completer=DockerImagesCompleter, hide_when_anonymous=True, default=None))
     METADATA_SPECS.append(MetadataSpec('request_time', str, 'Amount of time (e.g., 3, 3m, 3h, 3d) allowed for this run. Defaults to user time quota left.', formatting='duration', default=None))
     METADATA_SPECS.append(MetadataSpec('request_memory', str, 'Amount of memory (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run.', formatting='size', default='2g'))
+    METADATA_SPECS.append(MetadataSpec('memory_usage_out_of_limit', str, 'Portion of memory used by this container (e.g., 93.31MiB / 13.67GiB, 8.035MiB / 13.67GiB)', completer=DockerStatsCompleter))
```
epicfaace:
```suggestion
    METADATA_SPECS.append(MetadataSpec('memory_usage_out_of_limit', str, 'Portion of memory used by this container (e.g., 93.31MiB / 13.67GiB)', completer=DockerStatsCompleter))
```

Also, will it show "MiB" / "GiB" or "MB" / "GB"?

codalab/bundles/run_bundle.py
```diff
@@ -29,8 +29,10 @@ class RunBundle(DerivedBundle):
...
     METADATA_SPECS.append(MetadataSpec('request_disk', str, 'Amount of disk space (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run. Defaults to user disk quota left.', formatting='size', default=None))
     METADATA_SPECS.append(MetadataSpec('request_cpus', int, 'Number of CPUs allowed for this run.', default=1))
+    METADATA_SPECS.append(MetadataSpec('cpu_usage', str, 'Portion of CPU used by this container (e.g., 0.28%)', completer=DockerStatsCompleter))
```
epicfaace: Is this the maximum percentage of CPU used by this container? Or the percentage used at the current / latest time? (same question for memory, too)

codalab/bundles/run_bundle.py
epicfaace: For consistency with `cpu_usage`:

```suggestion
    METADATA_SPECS.append(MetadataSpec('memory_usage', str, 'Portion of memory used by this container (e.g., 93.31MiB / 13.67GiB, 8.035MiB / 13.67GiB)', completer=DockerStatsCompleter))
```

codalab/bundles/run_bundle.py
yiboliu: So these are all for the current percentage of CPU and memory used by this container. If you specify the flag `stream` to be true, then it will constantly output the value. I think we can probably hold the values in a list and plot it out or keep updating the max/avg values.

codalab/bundles/run_bundle.py
yiboliu: "MiB" / "GiB". If "MB" / "GB" is preferred, I can make a parser to convert the format.

percyliang: I wouldn't use a processed, string representation here.  We already have `memory` and `memory_max` fields.  I'd add `memory_total_available` for the total amount of memory

These should be `generated=True`, don't add a completer - see how other fields are done.

codalab/lib/completers.py
```diff
@@ -213,6 +213,18 @@ def __call__(self, prefix, action=None, parser=None, parsed_args=None):
             warn('Error: {}'.format(ex))
 
 
+class DockerStatsCompleter(CodaLabCompleter):
```
percyliang: Confused why there's a completer...these are for autocomplete on the CLI, but is irrelevant for reporting information back from docker since the user is not specifying them.

codalab/bundles/run_bundle.py
epicfaace: I think the preferred format would be similar to request_memory; i.e., formatted as (3, 3k, 3m, 3g, 3t). I think we already have a helper function that performs this format.

codalab/bundles/run_bundle.py
percyliang: Can we make these float rather than str?

codalab/lib/completers.py
yiboliu: Oh I didn't know that. I just found `DockerImagesCompleter` is performing something similar, so referred to this example.

codalab/bundles/run_bundle.py
yiboliu: Definitely.

codalab/bundles/run_bundle.py
```diff
@@ -31,6 +31,8 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_memory', str, 'Amount of memory (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run.', formatting='size', default='2g'))
     METADATA_SPECS.append(MetadataSpec('request_disk', str, 'Amount of disk space (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run. Defaults to user disk quota left.', formatting='size', default=None))
     METADATA_SPECS.append(MetadataSpec('request_cpus', int, 'Number of CPUs allowed for this run.', default=1))
+    METADATA_SPECS.append(MetadataSpec('cpu_usage', str, 'Portion of CPU used by this container (e.g., 0.28)', generated=True))
+    METADATA_SPECS.append(MetadataSpec('memory_usage', str, 'Portion of Memory used by this container (e.g., 0.16', generated=True))
```
teetone: Missing a paren at the end

codalab/bundles/run_bundle.py
```diff
@@ -31,6 +31,8 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_memory', str, 'Amount of memory (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run.', formatting='size', default='2g'))
     METADATA_SPECS.append(MetadataSpec('request_disk', str, 'Amount of disk space (e.g., 3, 3k, 3m, 3g, 3t) allowed for this run. Defaults to user disk quota left.', formatting='size', default=None))
     METADATA_SPECS.append(MetadataSpec('request_cpus', int, 'Number of CPUs allowed for this run.', default=1))
+    METADATA_SPECS.append(MetadataSpec('cpu_usage', str, 'Portion of CPU used by this container (e.g., 0.28)', generated=True))
```
teetone: Should this be a percentage instead?

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,33 @@ def get_container_stats(container):
...
+        cpu_used = int(container_stats['cpu_stats']['cpu_usage']['total_usage'])
+        total_cpu = int(container_stats['cpu_stats']['system_cpu_usage'])
+        cpu_usage = str(float(cpu_used / total_cpu))
+        memory_usage = str(
+            float(
+                container_stats['memory_stats']['usage'] / container_stats['memory_stats']['limit']
+            )
+        )
+        return cpu_usage, memory_usage
+    except KeyError:
```
teetone, line 301: When would a KeyError occur?

codalab/worker/worker_run_state.py
```diff
@@ -407,6 +409,17 @@ def check_and_report_finished(run_state):
             )
 
         def check_resource_utilization(run_state):
+
```
teetone: Nit: extra line

codalab/worker/worker_run_state.py
```diff
@@ -407,6 +409,17 @@ def check_and_report_finished(run_state):
...
+            try:
+                cpu_usage, memory_usage = docker_utils.get_container_stats_on_mac(
```
teetone: It's not clear to me what on mac means here. Does this command only work on a Mac?

teetone (commented):
Looking at your screenshot, do you think we should round the numbers?

yiboliu:
> Looking at your screenshot, do you think we should round the numbers?

You mean like rounding it to percentage representation?

codalab/bundles/run_bundle.py
yiboliu: I am ok with either way. If you think percentage represents better, I can definitely do it.

codalab/worker/docker_utils.py
yiboliu: Oh I commented at the other place but forgot here. It's when the bundle that's being queried has finished/exited, some keys will be missing.

codalab/worker/worker_run_state.py
yiboliu: The reason I name it is the existing way that queries stats of docker containers does not work on Mac. Ref: https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/docker_utils.py#L228. I created this method so that I can develop on my machine.

codalab/worker/docker_utils.py
teetone: Do you know why some keys would be missing? Is it a bug?

teetone:
> > Looking at your screenshot, do you think we should round the numbers?
> 
> You mean like rounding it to percentage representation?

I was suggesting rounding it to show less digits.

codalab/worker/docker_utils.py
yiboliu: It's not a bug. It's what docker apis return when called upon a finished/exited container.

codalab/worker/docker_utils.py
```diff
@@ -261,6 +265,39 @@ def get_container_stats(container):
...
+
+
+@wrap_exception('Unable to check Docker API for container')
+def get_container_stats_helper_by_name(container_name: str) -> Tuple[str, str]:
+    try:
+        container_stats: dict = client.containers.get(container_name).stats(stream=False)
+        cpu_used: int = int(container_stats['cpu_stats']['cpu_usage']['total_usage'])
+        total_cpu: int = int(container_stats['cpu_stats']['system_cpu_usage'])
+        cpu_usage: str = str(round(float(cpu_used / total_cpu) * 100, 3)) + " %"
+        memory_usage: str = (
```
teetone: It might be cleaner to use string format. Can you also add a comment of an example before and after the formatting and rounding?

teetone (approved):
Looks good, super useful feature

codalab/worker/docker_utils.py
```diff
@@ -279,25 +279,24 @@ def get_container_stats_helper_by_name(container_name: str) -> Tuple[str, str]:
...
-            + " %"
         )
         return cpu_usage, memory_usage
     except docker.errors.NotFound:
         raise
 
 
+# Without this processing of the float number, the output will look like "0.123456789";
+# With it, the output will look like "12.346 %"
+def float_to_percentage(float_number: float) -> str:
```
percyliang: Improve doc string with `"""Takes a ..."""` to describe what this function does.

codalab/bundles/run_bundle.py
percyliang: These should be floats.

codalab/bundles/run_bundle.py
```diff
@@ -31,6 +31,8 @@ class RunBundle(DerivedBundle):
...
+    METADATA_SPECS.append(MetadataSpec('memory_usage', str, 'Portion of Memory used by this container (e.g., 0.16)', generated=True))
```
percyliang: float

percyliang: Also move these after the `request_...` fields...

codalab/worker/docker_utils.py
```diff
@@ -261,6 +265,38 @@ def get_container_stats(container):
     return stats
 
 
+@wrap_exception('Unable to check Docker API for container')
+def get_container_stats_on_mac(container: docker.models.containers.Container):
```
percyliang: Why `_mac`?  Seems very specific.

codalab/worker/worker_run_state.py
```diff
@@ -406,7 +408,17 @@ def check_and_report_finished(run_state):
...
-        def check_resource_utilization(run_state):
+        def check_resource_utilization(run_state: RunState):
+            try:
+                cpu_usage, memory_usage = docker_utils.get_container_stats_on_mac(
+                    run_state.container
+                )
+                run_state = run_state._replace(cpu_usage=cpu_usage)
+                run_state = run_state._replace(memory_usage=memory_usage)
+            except KeyError:
+                # If the bundle has exited, then some of the keys are missing, so we just dismiss it.
```
percyliang: Where does this `KeyError` come from?

codalab/bundles/run_bundle.py
yiboliu: the output is in the format of percentage for more straightforward.

codalab/worker/worker_run_state.py
yiboliu: It comes from `docker_utils.get_container_stats_on_mac`. If the bundle has exited, for example, the `container_stats['memory_stats']` is an empty dict, and some other keys are also missing.

codalab/worker/docker_utils.py
yiboliu: The existing method to retrieve container stats doesn't work on mac, so I created this method to develop on my machine. I'm experimenting to see if I can get the values from the existing method.

codalab/worker/worker_run_state.py
percyliang: Can't you handle this exception inside `get_container_stats_on_mac` then?

codalab/bundles/run_bundle.py
```diff
@@ -36,6 +36,11 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important). Negative '
                                                                 'priority bundles are queued behind bundles with no specified priority.', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(
```
epicfaace, line 39: These should probably be floats instead of a string

epicfaace (changes requested):

codalab/bundles/run_bundle.py
yiboliu: the output is in the format of percentage for more straightforward. like `12.346 %`

codalab/bundles/run_bundle.py
```diff
@@ -36,6 +36,11 @@ class RunBundle(DerivedBundle):
...
+        MetadataSpec('cpu_usage', str, 'Portion of CPU used by this container (e.g., 28 %)', generated=True))
+    METADATA_SPECS.append(
+        MetadataSpec('memory_usage', str, 'Portion of Memory used by this container (e.g., 16 %)', generated=True))
```
epicfaace: Store memory_limit instead

codalab/bundles/run_bundle.py
```diff
@@ -36,6 +36,11 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important). Negative '
                                                                 'priority bundles are queued behind bundles with no specified priority.', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(
+        MetadataSpec('cpu_usage', float, 'Portion of CPU used by this container (e.g., 0.24)', generated=True))
+    METADATA_SPECS.append(
+        MetadataSpec('memory_limit', int, 'Limit of Memory available to this container (e.g., 2085326848)', generated=True))
```
epicfaace, line 42:
```suggestion
        MetadataSpec('memory_limit', int, 'Limit of Memory, in bytes, available to this container (e.g., 2085326848)', generated=True))
```

codalab/worker/docker_utils.py
```diff
@@ -226,6 +229,7 @@ def get_bundle_container_volume_binds(bundle_path, docker_bundle_path, dependenc
 def get_container_stats(container):
     # We don't use the stats API since it doesn't seem to be reliable, and
     # is definitely slow. This doesn't work on Mac.
+    # possible ref: https://scoutapm.com/blog/monitoring-docker-containers-from-scratch
```
epicfaace: I don't think we need this?

yiboliu: I can delete it

codalab/worker/docker_utils.py
```diff
@@ -261,6 +265,55 @@ def get_container_stats(container):
...
+        system_delta: int = container_stats['cpu_stats']['system_cpu_usage'] - container_stats[
+            'precpu_stats'
+        ]['system_cpu_usage']
+        if system_delta > 0 and cpu_delta > 0:
+            cpu_usage: float = float(cpu_delta / system_delta) * float(
+                len(container_stats['cpu_stats']['cpu_usage']['percpu_usage'])
+            )
+            return cpu_usage
+        return 0.0
+    except KeyError:
```
epicfaace, line 301: When is a `KeyError` raised?

codalab/worker/docker_utils.py
```diff
@@ -261,6 +265,55 @@ def get_container_stats(container):
...
+        memory_limit: int = get_memory_limit(container_stats)
+
+        return cpu_usage, memory_limit
+    except docker.errors.NotFound:
+        raise
+
+
+# The ref of the way of calculation:
+# https://www.jcham.com/2016/02/09/calculating-cpu-percent-and-memory-percentage-for-containers/
+def get_cpu_usage(container_stats: dict) -> float:
```
epicfaace, line 282: Add a docstring here describing what's being calculated.

We should also link, in addition to the above link, to this link (which shows how the actual Docker client uses this same code): https://github.com/moby/moby/blob/131e2bf12b2e1b3ee31b628a501f96bbb901f479/api/client/stats.go#L309

codalab/worker/docker_utils.py
```diff
@@ -261,6 +265,55 @@ def get_container_stats(container):
     return stats
 
 
+@wrap_exception('Unable to check Docker API for container')
+def get_container_stats_on_mac(container: docker.models.containers.Container):
```
epicfaace: Let's rename this to `get_container_stats_with_docker_stats`

epicfaace: Can you add docstrings for each of these methods?

codalab/worker/docker_utils.py
```diff
@@ -261,6 +264,55 @@ def get_container_stats(container):
...
+    if container_exists(container):
+        return get_container_stats_helper_by_name(container.name)
+    else:
+        return 0.0, 0
+
+
+@wrap_exception('Unable to check Docker API for container')
+def get_container_stats_helper_by_name(container_name: str) -> Tuple[float, int]:
```
epicfaace: Let's combine this function with `get_container_stats_on_mac`.

epicfaace (commented):
Let's add some tests for this functionality. Maybe you could run a bundle in test_cli.py, then check the metadata using `cl info` to ensure that cpu_usage and memory_limit fields have been populated

codalab/worker/docker_utils.py
yiboliu: Sometimes when a container isn't fully ready and will give incomplete results of stats. That's when a KeyError occur

yiboliu: If the bundle is not fully ready or has exited, then the stats returned is missing some keys.

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
     return stats
 
 
+@wrap_exception('Unable to check Docker API for container')
+def get_container_stats_with_docker_stats(container: docker.models.containers.Container):
+    """Takes a container, returns the cpu_usage and memory_usage"""
```
epicfaace:
```suggestion
    """Returns the cpu usage and memory limit of a container using the Docker Stats API."""
```

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
...
+        except docker.errors.NotFound:
+            raise
+    else:
+        return 0.0, 0
+
+
+# The ref of the way of calculation:
+# https://www.jcham.com/2016/02/09/calculating-cpu-percent-and-memory-percentage-for-containers/
+# How docker client calcualtes it:
+# https://github.com/moby/moby/blob/131e2bf12b2e1b3ee31b628a501f96bbb901f479/api/client/stats.go#L309
```
epicfaace: Can you move these comments to the docstring below?

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
...
+def get_cpu_usage(container_stats: dict) -> float:
+    """Takes a dictionary of container stats returned by docker stats, returns cpu usage"""
```
epicfaace: C
```suggestion
    """Calculates CPU usage from container stats returned from the Docker Stats API."""
```

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
...
+            'precpu_stats'
+        ]['system_cpu_usage']
+        if system_delta > 0 and cpu_delta > 0:
+            cpu_usage: float = float(cpu_delta / system_delta) * float(
+                len(container_stats['cpu_stats']['cpu_usage']['percpu_usage'])
+            )
+            return cpu_usage
+        return 0.0
+    except KeyError:
+        # If the bundle is not fully ready or has exited, then the stats returned is missing some keys.
```
epicfaace:
```suggestion
        # The stats returned may be missing some keys if the bundle is not fully ready or has exited.
```

tests/cli/test_cli.py
```diff
@@ -2440,6 +2440,23 @@ def test_wopen(ctx):
     _run_command([cl, 'wopen', 'nonexistent'], expected_exit_code=1)
 
 
+@TestModule.register('metadata')
+def test_metadata(ctx):
```
epicfaace: Can you add this to `test_run` instead?

tests/cli/test_cli.py
```diff
@@ -2440,6 +2440,23 @@ def test_wopen(ctx):
...
+            cl,
+            "run",
+            "'for i in {1..60}; do sleep 1; done'",
+            "--request-memory",
+            "15m",
+            "--request-docker-image",
+            "python:3.6.10-buster",
+        ]
+    )
+    _run_command([cl, 'info', uuid, '-f', 'cpu_usage'])
```
epicfaace: Can you use the `get_info` function here instead?

epicfaace: We also need to check that the _result_ of this command is something sensible (i.e., something greater than 0)

tests/cli/test_cli.py
```diff
@@ -1357,6 +1357,21 @@ def test_run(ctx):
         2 + 2 + 1, _run_command([cl, 'cat', remote_uuid])
     )  # 2 header lines, 1 stdout file, 1 stderr file, 1 item at bundle target root
 
+    # test metadata
```
```suggestion
    # Test that cpu_usage and memory_limit are properly populated
```

tests/cli/test_cli.py
```diff
@@ -1357,6 +1357,21 @@ def test_run(ctx):
...
+    uuid = _run_command(
+        [
+            cl,
+            "run",
+            "'for i in {1..60}; do sleep 1; done'",
+            "--request-memory",
+            "15m",
+            "--request-docker-image",
```
epicfaace: I think you can remove `--request-docker-image`. It's automatically appended when you run `_run_command`.

tests/cli/test_cli.py
```diff
@@ -1357,6 +1357,21 @@ def test_run(ctx):
...
+            "python:3.6.10-buster",
+        ]
+    )
+    check_not_equals(get_info(uuid, 'cpu_usage'), '0.0')
+    check_not_equals(get_info(uuid, 'memory_limit'), '0')
```
epicfaace: Can we instead check to ensure that when we parse it as an int, it's greater than 0?

`assert int(get_info(uuid, 'memory_limit')) > 0`

This way, we'll ensure the returned value is correct (as opposed to something else like `None`)

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
...
+        ]['system_cpu_usage']
+        if system_delta > 0 and cpu_delta > 0:
+            cpu_usage: float = float(cpu_delta / system_delta) * float(
+                len(container_stats['cpu_stats']['cpu_usage']['percpu_usage'])
+            )
+            return cpu_usage
+        return 0.0
+    except KeyError:
+        # The stats returned may be missing some keys if the bundle is not fully ready or has exited.
+        # We can just skip for now and wait for calling next time.
```
epicfaace:
```suggestion
        # We can just skip for now and wait until this function is called the next time.
```

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
...
+            raise
+    else:
+        return 0.0, 0
+
+
+def get_cpu_usage(container_stats: dict) -> float:
+    """Calculates CPU usage from container stats returned from the Docker Stats API.
+       The ref of the way of calculation:
+       https://www.jcham.com/2016/02/09/calculating-cpu-percent-and-memory-percentage-for-containers/
+       How docker client calcualtes it:
```
epicfaace:
```suggestion
       That method is also based on how the docker client calculates it:
```

codalab/worker/docker_utils.py
```diff
@@ -261,6 +262,57 @@ def get_container_stats(container):
...
+            return cpu_usage, memory_limit
+        except docker.errors.NotFound:
+            raise
+    else:
+        return 0.0, 0
+
+
+def get_cpu_usage(container_stats: dict) -> float:
+    """Calculates CPU usage from container stats returned from the Docker Stats API.
+       The ref of the way of calculation:
```
epicfaace:
```suggestion
       The way of calculation comes from here:
```

epicfaace (approved):

PR #3094, opened 2020-12-04
Properly upload archives from a URL with a query string

### LINKED ISSUE #3090: [USER ISSUE] Bundle with URLs with a query string do not have file type recognized
A bundle with a URL with a query string (such as https://www.dropbox.com/s/i8x1p0im11rmur9/iwildcam_data.tar.gz?dl=1) does not have its extension recognized as ".tar.gz". Thus, the bundle is just treated as "binary" and not extracted, as it's supposed to do.

See https://worksheets.codalab.org/bundles/0x2a18d6f000de48debe40b0e05c07b418

epicfaace:
Fixes #3090.

I ended up adding some unit tests for zip_util (initially I was trying to fix the problem in zip_util, until I saw where the actual problem is), but the tests shouldn't hurt to merge anyway.


epicfaace:
@teetone could you review when you get a chance?

teetone (approved):

PR #3105, opened 2020-12-10
Return x-codalab-target-size header

### LINKED ISSUE #3102: Include Content-Length in /blob bundle download endpoint
Original question posted in #codalab Slack channel:

> For these very large files, it'd be nice to have a progress bar when downloading. Currently, our progress bar shows how much has been downloaded, but we can't access the total size, so we can't show a percentage.
> If we do:
> 
> ```
> resp = urllib.request.urlopen(url)
> length = resp.getheader('content-length')
> ```
> 
> length will be None, as "content-length" is not provided in the header. Do you know if there's a solution to this, apart from hardcoding the sizes?

epicfaace:
Return x-codalab-target-size header -- returns the target size when downloading a bundle or a particular file from a bundle.

Fixes #3102.

Also added unit test coverage for file uploading / downloading through the REST API endpoints.

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -722,10 +722,7 @@ def _update_bundle_contents_blob(uuid):
     check_bundles_have_all_permission(local.model, request.user, [uuid])
     bundle = local.model.get_bundle(uuid)
     if bundle.state in State.FINAL_STATES:
-        abort(
-            http.client.FORBIDDEN,
-            f'Contents cannot be modified, bundle already finalized, in state {State.FINAL_STATES}.',
-        )
+        abort(http.client.FORBIDDEN, 'Contents cannot be modified, bundle already finalized.')
 
     # Get and validate query parameters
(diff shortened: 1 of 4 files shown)


codalab/rest/bundles.py
```diff
@@ -678,6 +680,7 @@ def _fetch_bundle_contents_blob(uuid, path=''):
     else:
         response.set_header('Content-Disposition', 'attachment; filename="%s"' % filename)
     response.set_header('Target-Type', target_info['type'])
+    response.set_header('X-Codalab-Target-Size', target_info['size'])
```
percyliang, line 687: Should we put a comment that this is the uncompressed size, whereas the file will be compressed?

percyliang: Oh, this is Target-Size, so I guess it's fine.  But how does the client know how to interpret this header?

epicfaace: I added a note to the docstring in this method describing this fact of X-CodaLab-Target-Size.

teetone (approved):

PR #3113, opened 2020-12-13
Add cl open command for opening bundles in browser

nelson-liu:

This adds a new command `cl open` to open bundles in the browser from the cli. This is useful because when debugging bundles, I often have a workflow like:
- run some `cl search` / have some bundle name i want to investigate
- `cl info` on the short uuid from `cl search` / the bundle name
- get the full uuid from the `cl info` output
- copy and paste this into my browser

Instead, now I can do:
- `cl open <bundle ref>`

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2882,7 +2882,7 @@ class BundleCLI(object):
             'bundles', params={'specs': args.bundle_spec, 'worksheet': worksheet_uuid},
         )
 
-        for i, info in enumerate(bundles):
+        for info in bundles:
             webbrowser.open(self.bundle_url(info['id']))
 
         # Headless client should fire OpenBundle UI action
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2861,10 +2861,10 @@ class BundleCLI(object):
     @Commands.command(
         'open',
         aliases=('o',),
-        help='Open a bundle detail page in a local web browser.',
+        help='Open bundle(s) detail page(s) in a local web browser.',
         arguments=(
             Commands.Argument(
-                'bundle_spec', help=BUNDLE_SPEC_FORMAT, nargs='*', completer=BundlesCompleter
+                'bundle_spec', help=BUNDLE_SPEC_FORMAT, nargs='+', completer=BundlesCompleter
             ),
             Commands.Argument(
                 '-w',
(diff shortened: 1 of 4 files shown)


nelson-liu:
I tested things locally, but not sure how to write a unit test for this.

epicfaace (commented):
Maybe you can just add something in test_cli.py that ensures running `cl open` on a bundle fails with a nonexistent bundle, but succeeds with a regular bundle id?

codalab/lib/bundle_cli.py
```diff
@@ -2856,6 +2858,40 @@ def do_write_command(self, args):
         )
         print(target.bundle_uuid, file=self.stdout)
 
+    @Commands.command(
+        'open',
+        aliases=('o',),
+        help='Open a bundle detail page in a local web browser.',
```
percyliang: Does this actually open multiple bundles that match the bundle spec?

nelson-liu: yes, or you can specify multiple specs (e.g., `cl open ^1 ^2`). i'll update the docstring.

percyliang:
Just thinking through the worksheets case and how to make the UX consistent.  Right now, `cl work` prints out the URL which you can just click on to launch in the browser, which seems fine.  If `cl info` printed out the URL of the bundle, would that be good enough?  That would reduce the number of commands that need to be supported, but it wouldn't be as discoverabale.

The other direction to support `cl open` and `cl wopen`...

nelson-liu:
Yeah, I thought about adding another link to cl info, but I think ultimately I (personally) like the ux of having an explicit command to open. This is because:

1. To open the current worksheet, I'd rather "cl wopen"  than "cl st" and click the URL
2. Accessing the url in "cl info" requires me to go visually parse the output, which can be especially annoying if I'm working on a laptop screen since I'd have to scroll and find it.
3. cl open seems better for scripting. You can now do things like "find all failing bundles on this worksheet and open them in my browser" (so I can inspect the stdout and see why they failed)

So, I think I like open / wopen because (1) I'm allergic to using the mouse and vastly prefer the keyboard and (2) it opens up some new scripting possibilities.

codalab/lib/bundle_cli.py
```diff
@@ -2856,6 +2858,40 @@ def do_write_command(self, args):
...
+    )
+    def do_open_command(self, args):
+        args.bundle_spec = spec_util.expand_specs(args.bundle_spec)
+        client, worksheet_uuid = self.parse_client_worksheet_uuid(args.worksheet_spec)
+
+        bundles = client.fetch(
+            'bundles', params={'specs': args.bundle_spec, 'worksheet': worksheet_uuid},
+        )
+
+        for i, info in enumerate(bundles):
```
percyliang: You can do `for info in bundles`, right?

percyliang (approved):

codalab/lib/bundle_cli.py
nelson-liu: yup, thanks. surprised flake8 didn't catch this.

PR #3115, opened 2020-12-13
Make timeout configurable with environment variable

EntilZha:

I've been running into issues with timeouts and noticed a prior change that increased it. This helped, but didn't eliminate the issue. Instead of increasing it, it would be convenient for it to be user configurable through an environment variable.

### Checklist

* [ x] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script

### DIFF
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -11,11 +11,7 @@ import http.client
 # http://semver.org/
 CODALAB_VERSION = '0.5.32'
 BINARY_PLACEHOLDER = '<binary>'
-ENV_TIMEOUT = os.environ.get('CODALAB_URLOPEN_TIMEOUT_SECONDS')
-if ENV_TIMEOUT is None:
-    URLOPEN_TIMEOUT_SECONDS = 5 * 60
-else:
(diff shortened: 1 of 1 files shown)


codalab/common.py
```diff
@@ -4,13 +4,18 @@
...
 """
+import os
 import http.client
 
 # Increment this on master when ready to cut a release.
 # http://semver.org/
 CODALAB_VERSION = '0.5.32'
 BINARY_PLACEHOLDER = '<binary>'
-URLOPEN_TIMEOUT_SECONDS = 5 * 60
+ENV_TIMEOUT = os.environ.get('CODALAB_URLOPEN_TIMEOUT_SECONDS')
```
nelson-liu: how about doing `URLOPEN_TIMEOUT_SECONDS = int(os.environ.get('CODALAB_URLOPEN_TIMEOUT_SECONDS', 5 * 60))`?

EntilZha: Done

nelson-liu (approved):
Thanks for the PR!

PR #3122, opened 2020-12-17
Fix target sizes for directories

### LINKED ISSUE #3120: Target sizes for directories are 4096, instead of the size of the entire directory
The target size of directories should be reported as the size of the entire directory, not just 4096. This value is exposed to the user in the X-Codalab-Target-Size header in the bundle download endpoint.

image

epicfaace:
Fixes #3120. This is technically a backwards incompatible change (cc @percyliang ), but I think it makes sense to report the target sizes for directories as equal to the size of their contents instead of always equal to 4096.

### DIFF
--- a/codalab/worker/download_util.py
+++ b/codalab/worker/download_util.py
@@ -124,6 +124,7 @@ def _compute_target_info(path, depth):
         result['type'] = 'file'
     elif os.path.isdir(path):
         result['type'] = 'directory'
+        # Retrieve the size of the directory and all its contents
         result['size'] = get_size(path)
         if depth > 0:
             result['contents'] = [
--- a/codalab/worker/download_util.py
+++ b/codalab/worker/download_util.py
@@ -1,4 +1,5 @@
 import os
+from codalab.lib.path_util import get_size
 
 
 class PathException(Exception):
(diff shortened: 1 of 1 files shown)


codalab/worker/download_util.py
```diff
@@ -123,6 +124,8 @@ def _compute_target_info(path, depth):
         result['type'] = 'file'
     elif os.path.isdir(path):
         result['type'] = 'directory'
+        # Retrieve the size of the directory and all its contents
+        result['size'] = get_size(path)
```
percyliang, line 128: This could be really slow, which is why we didn't put that in...

percyliang: For getting the size for download, could we just use data_size for the top-level bundle and individual files?

epicfaace: Yes, we could do that, though that would mean that getting the target info of a directory within a bundle would still return 4096.

percyliang: That's fine or we can make it return `None` or something to not be confusing.

epicfaace:
Let's not do this approach -- instead, we'll just return data_size for the target size in the future.

Issue #3082, opened 2020-12-01
[USER ISSUE] Trying to upload large folder fails.


henrikmarklund:
**Describe the bug**
I'm trying to upload a folder of size  86G. I run cl upload -n 'name' /path/to/data/. This fails after 64G with  json_api_client.JsonApiException.

**Codalab Information**
- *Your username*: henrikmarklund
- *UUID of worksheet where problem occurred*: 0xb44731cc8e8a4265a20146c3887b6b90
- *UUID of problem bundle (please don't delete)*: 0x877ce404aa6a43adac7a95a4a8ee8c51
- *Error message received from Codalab (if applicable)*:
cl upload -n 'iwildcam' /iris/u/marklund/data/iwildcam/data
Preparing upload archive...
Uploading data.tar.gz (0x877ce404aa6a43adac7a95a4a8ee8c51) to https://worksheets.codalab.org
Sent 64189.78MiB [7.88MiB/sec]
Traceback (most recent call last):
  File "/iris/u/marklund/miniconda3/lib/python3.7/site-packages/codalab/client/json_api_client.py", line 18, in wrapper
    return f(*args, **kwargs)
  File "/iris/u/marklund/miniconda3/lib/python3.7/site-packages/codalab/client/json_api_client.py", line 660, in upload_contents_blob
    progress_callback=progress_callback,
  File "/iris/u/marklund/miniconda3/lib/python3.7/site-packages/codalab/worker/rest_client.py", line 153, in _upload_with_chunked_encoding
    conn.send(b'%X\r\n%s\r\n' % (len(to_send), to_send))
  File "/iris/u/marklund/miniconda3/lib/python3.7/http/client.py", line 987, in send  
    self.sock.sendall(data)
  File "/iris/u/marklund/miniconda3/lib/python3.7/ssl.py", line 1034, in sendall
    v = self.send(byte_view[count:])
  File "/iris/u/marklund/miniconda3/lib/python3.7/ssl.py", line 1003, in send
    return self._sslobj.write(data)
BrokenPipeError: [Errno 32] Broken pipe

Traceback (most recent call last):
  File "/iris/u/marklund/miniconda3/bin/cl", line 8, in <module>
    sys.exit(main())
  File "/iris/u/marklund/miniconda3/lib/python3.7/site-packages/codalab/bin/cl.py", line 10, in main
    cli.do_command(sys.argv[1:])
[...]

PR #3123, opened 2020-12-18
Add stress tests within test_cli.py for uploading large files

epicfaace:

Adding tests to debug issues with uploading large files and preventing future regressions. I'm adding these tests directly in test_cli.py; that way, they won't be dependent on worksheets-dev.codalab.org as the stress tests are, so they'll hopefully be isolated, run more often, be less flaky, and run faster (as large files no longer have to be transferred across the network, since it's just accessing `localhost`).

### Related issues

#3082

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,6 +90,7 @@ jobs:
           - copy netcat netcurl
           - edit
           - open wopen
+          - stress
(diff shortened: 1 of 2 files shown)


epicfaace:
Let's just use regular stress tests for this purpose, for now.

PR #3124, opened 2020-12-20
Replace urllib with requests in _upload_with_chunked_encoding (2)

epicfaace:
Extending on #3041.


nelson-liu:
thanks @epicfaace , sorry I haven't had a chance to finish up the original PR.

nelson-liu:
hmm looks like tests are failing, but i can't tell if it's related to this PR.

epicfaace:
@nelson-liu I think we shouldn't actually replace urllib with requests here, because of the nature of the fix in #3285. There's no way to encode the behavior implemented in #3285 (keep timing out, sending null bytes) when we're using requests.

There is a major improvement / refactor we can do to the code through #3286, though.

PR #3131, opened 2020-12-21
Revert gzip changes

epicfaace:
One possible issue for the slowness could be that when I changed gzip_file earlier, I switched it from running gzip in a new process to just using the same existing process.

https://github.com/codalab/codalab-worksheets/pull/2952/files#diff-2b2d1ef00e8805335562a7967af36158e9140f6c4a5039f3462517795e4d67f6L109-L112

Previously, the codalab rest server may have run gzip in separate processes (which can then utilize multiple cores), but now, all that work is done in a single process.

If this is the issue, we could try either of the following:
- revert the gzip changes, so it uses subprocess to run gzip instead
- increase the number of processes for the server

This PR does the former.

### DIFF
--- a/codalab/worker/file_util.py
+++ b/codalab/worker/file_util.py
@@ -211,31 +211,37 @@ def gzip_file(file_path):
     """
     Returns a file-like object containing the gzipped version of the given file.
     """
-    BUFFER_SIZE = 100 * 1024 * 1024  # Zip in chunks of 100MB
-
-    class GzipStream:
-        def __init__(self, fileobj):
-            self.__input = fileobj
-            self.__buffer = BytesBuffer()
-            self.__gzip = gzip.GzipFile(None, mode='wb', fileobj=self.__buffer)
-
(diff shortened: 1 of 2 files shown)


percyliang:
This seems like a likely culprit (it does feel like some sort of undesired blocking is happening).  Could we test the performance implications of this?

percyliang (approved):

codalab/worker/file_util.py
```diff
@@ -211,31 +211,12 @@ def gzip_file(file_path):
...
-                if not s:
-                    self.__gzip.close()
-                    break
-                self.__gzip.write(s)
-            return self.__buffer.read(size)
-
-        def close(self):
-            self.__input.close()
-
+    args = ['gzip', '-c', '-n', file_path]
```
percyliang, line 216: Could you document that it's right now important for gzip to run in a separate process, otherwise things grind to a halt?

PR #3135, opened 2020-12-27
Resolve uploading files failure

### LINKED ISSUE #3106: Can't upload .tar.gz files through the web UI
Steps to reproduce:
- Click "Upload file"
- Upload a .tar.gz file
- I get the following errors:

image

Also, the request to https://worksheets-dev.codalab.org/rest/bundles/0x1b56e9c9267b4983ba73f8e5ec5f1471/contents/blob/?finalize=1&filename=_0x64d435b1adff..2bea2d713ae2464.tar.gz&unpack=1 fails because the payload isn't correct.

image

This is probably a frontend issue, as the request payload should **not** be equal to `zip error: Invalid command arguments (long option 'disable-copyfile' not supported)`.

Levilian:
Fix #3106

### DIFF
--- a/frontend/src/components/worksheets/NewUpload/NewUpload.jsx
+++ b/frontend/src/components/worksheets/NewUpload/NewUpload.jsx
@@ -110,7 +110,7 @@ class NewUpload extends React.Component<{
                         );
                         this.props.onUploadFinish();
                         reject(error);
-                    },
+                    }.bind(this),
                 });
 
             };


epicfaace:
Thanks -- is the second issue mentioned in that issue resolved as well / are you still seeing it? Here's the screenshot for that error again:

image

Levilian:
Screen Shot 2020-12-28 at 1 29 56 AM
Yes @epicfaace the second issue should have been fixed with the change.

epicfaace (approved):
Cool, works for me!

PR #3136, opened 2020-12-27
Use gevent workers for better performance; add performance test

epicfaace:
## Motivation

@percyliang I did a basic performance test for #3131, and it was slightly faster to switch to use gzip subprocesses. However, when I tried adding a test to test_cli.py for performance testing (in this PR), it turned out that whether or not we use gzip in a subprocess, it's really slow and the server times out / goes down.

I dug deeper and it appears that gevent workers are better suited for I/O-bounded applications, like ours, than gthread, which we're currently using. See this for more info: https://medium.com/building-the-system/gunicorn-3-means-of-concurrency-efbb547674b7

## This PR

Thus, this PR does two things:
- Revert https://github.com/codalab/codalab-worksheets/commit/561da0ee8a7925aca4d251e4aac3c06fa4d91a48, so that we use gevent with 10 processes instead of using gthread with a single process / 50 threads.
- Create a performance test in test_cli.py. This test uploads and downloads files / directories at 10x concurrency, then periodically runs "cl workers" and times how long that command takes.

## Performance test results

I ran the performance test using varying combinations of worker types, and whether we use the gzip subprocess or stick with the existing code (gzipstream). The times taken should start very small, but gradually increase as the server gets more load with the concurrent uploads / downloads.

**gthread, gzipstream (what we currently have in master):**
`cl workers` times out, server went down

```
>> cl workers
 (exit code test-cli exception, expected 0 BAD)
...
codalab.client.json_api_client.JsonApiException: Unable to create bundles: timed out
```

**gthread, gzip subprocess (what #3131 would have done):**
`cl workers` times out, server went down. Same error message as above

**gevent with 10 workers, gzipstream:**
All times (s) taken for `cl workers`: [0.7127954959869385, 0.6075131893157959, 0.7263393402099609, 7.872021913528442, 8.671122312545776, 7.271690130233765, 0.5290284156799316, 0.4893498420715332, 2.7468934059143066, 7.838416576385498, 8.023988962173462, 6.165715456008911]. Average time: 4.304572920004527

**gevent with 10 workers, gzip subprocess:**
All times (s) taken for `cl workers`: [0.7731211185455322, 0.6943981647491455, 0.37362003326416016, 5.811000347137451, 9.235201120376587, 8.597550868988037, 3.903759002685547, 0.5644106864929199, 0.6485815048217773, 3.2893669605255127, 9.11350131034851, 9.951630115509033, 8.249640464782715, 0.6637413501739502]. Average time: 4.419251646314349

It appears that when we use gevent, there's not a very big difference between using / not using the gzip subprocess (and perhaps it's slightly worse to use the gzip processes), so this PR doesn't add the gzip subprocess back in (unlike #3131).

Finally, I also tried using gthread with 10 workers (we currently use only 1 worker), but it still timed out in the performance test. This leads me to believe that we should not use gthread at all, as even scaling the number of workers doesn't help with the concurrent requests in the performance test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,6 +90,7 @@ jobs:
           - copy netcat netcurl
           - edit
           - open wopen
+          - performance
     steps:
       - name: Clear free space
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,7 +80,7 @@ jobs:
     strategy:
       matrix:
         test:
-          - unittest gen-rest-docs gen-cli-docs gen-readthedocs basic auth status batch anonymous competition unicode rest1 upload1 upload2 upload3 upload4 download refs binary rm make worksheet_search worksheet_tags freeze detach perm search_time groups
+          # - unittest gen-rest-docs gen-cli-docs gen-readthedocs basic auth status batch anonymous competition unicode rest1 upload1 upload2 upload3 upload4 download refs binary rm make worksheet_search worksheet_tags freeze detach perm search_time groups
           - worker_manager
           - run
           - run2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,7 +80,7 @@ jobs:
     strategy:
       matrix:
         test:
-          # - unittest gen-rest-docs gen-cli-docs gen-readthedocs basic auth status batch anonymous competition unicode rest1 upload1 upload2 upload3 upload4 download refs binary rm make worksheet_search worksheet_tags freeze detach perm search_time groups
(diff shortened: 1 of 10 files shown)


epicfaace:
Hi @klopyrev , do you remember why we originally switched from gevent to gthread in https://github.com/codalab/codalab-worksheets/commit/561da0ee8a7925aca4d251e4aac3c06fa4d91a48? Want to make sure we understand the original motivation for this change before we revert it.

klopyrev:
I have an explanation for the change in the original pull request:
https://github.com/codalab/codalab-worksheets/pull/377

epicfaace:
Thanks @klopyrev ! It appears that this change might be a bit more involved, then, so let's not implement for now.

PR #3142, opened 2020-12-30
Enable customizing cl search and cl ls output columns

nelson-liu:

Right now, `cl search` and `cl ls` display a fixed set of columns (although all the information is fetched). This enables the user to pick what fields they want to show up.

I have a use case where I need to do a `cl search` and examine the `request_priority` of each search result. Right now, I have to do:

1. `cl search <critera>`
2. iterate through search results, running `cl info -f` on each search result

This blows up the # of network roundtrips considerably, slowing things down. Now, I can just do:

1. `cl search <critera> -f uuid,name,request_priority`

and I can get something like:

```
uuid      name                 request_priority
-----------------------------------------------
0x58947c  glove.840B.300d
0xa60e66  select_ensemble
0xdf051b  run-predictions      0
0xfe18a1  predictions-albert
0x005f26  run-predictions      0
0x8cb148  predictions-electra
0x27b373  run-predictions      0
0xfa3087  predictions-electra
0xb1f4ea  src
0x42180e  electra_model
```

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -155,6 +155,15 @@ HEADING_LEVEL_2 = '## '
 HEADING_LEVEL_3 = '### '
 
 NO_RESULTS_FOUND = 'No results found'
+DEFAULT_BUNDLE_INFO_LIST_FIELDS = (
+    'uuid',
+    'name',
+    'summary',
+    'owner',
+    'created',
+    'data_size',
+    'state',
+)
 
 
 class CodaLabArgumentParser(argparse.ArgumentParser):
@@ -2042,6 +2051,13 @@ class BundleCLI(object):
         arguments=(
             Commands.Argument('keywords', help='Keywords to search for.', nargs='+'),
             Commands.Argument(
+                '-f',
+                '--field',
+                type=str,
+                default="uuid,name,summary,owner,created,data_size,state",
+                help='Print out these comma-separated fields in the results table',
+            ),
+            Commands.Argument(
                 '-a',
                 '--append',
                 help='Append these bundles to the current worksheet.',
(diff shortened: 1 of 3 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -2041,6 +2050,13 @@ def do_rm_command(self, args):
         ],
         arguments=(
             Commands.Argument('keywords', help='Keywords to search for.', nargs='+'),
+            Commands.Argument(
+                '-f',
+                '--field',
+                type=str,
+                default="uuid,name,summary,owner,created,data_size,state",
```
percyliang: Can we do `','.join(DEFAULT_BUNDLE_INFO_LIST_FIELDS)`?

codalab/lib/bundle_cli.py
```diff
@@ -2118,6 +2136,13 @@ def create_reference_map(self, info_type, info_list):
         name='ls',
         help='List bundles in a worksheet.',
         arguments=(
+            Commands.Argument(
+                '-f',
+                '--field',
+                type=str,
+                default="uuid,name,summary,owner,created,data_size,state",
```
percyliang: same comment as above

percyliang (approved):
Nice!

codalab/lib/bundle_cli.py
nelson-liu:
```suggestion
                default=','.join(DEFAULT_BUNDLE_INFO_LIST_FIELDS),
```

codalab/lib/bundle_cli.py
nelson-liu:
```suggestion
                default=','.join(DEFAULT_BUNDLE_INFO_LIST_FIELDS),
```

PR #3145, opened 2021-01-03
Read state from state.json to persist sessions

### LINKED ISSUE #3078: CLI frequently forgets which worksheet I am on
**Describe the bug**
After not working for a few minutes, codalab forgets the result of the previous "cl work nlp::" command and tries to connect to the public site.

**Codalab Information**
- *Your username*:   AngledLuffa
 - *OS (if on CLI)*:
this is on the sc server on the stanford cluster
 - *Codalab CLI Version (if on CLI)*: 0.5.29

**To Reproduce**
1. Do codalab stuff
2. Wait a little while (much less than an hour)
3. have to run "cl work nlp::" again

**Expected behavior**
Would prefer it to stay on the same worksheet

teetone:

This PR fixes the issues where users get logged out abruptly or the CLI forgets what worksheet the user was working on. These issues happen when a user has more than one session running with multiple instances of the client each with a `CodaLabManager` on a single machine for a particular CodaLab instance. That is why only our Stanford users run into these issues, as they usually have multiple screen sessions running worker managers for https://codalab.stanford.edu.

Before this fix, a `CodaLabManager` would read from `state.json` once on initialization. For any operations that alter the state, it would make changes to its own copy stored in memory and override whatever was already written out to `state.json`. This causes problems as any old `CodaLabManager` could wipe out any new session information written by a newer `CodaLabManager`.

The fix is to simply always read the state from `state.json`.

### Related issues

Resolves #3078 

### Checklist

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -304,7 +304,8 @@ class CodaLabManager(object):
         """
         Return the current session.
         """
-        sessions = self.state['sessions']
+        state = self.get_state()
+        sessions = state['sessions']
         name = self.session_name()
         if name not in sessions:
             # New session: set the address and worksheet uuid to the default (main if not specified)
@@ -328,7 +329,7 @@ class CodaLabManager(object):
     @cached
     def model(self):
         """
-        Return a model.  Called by the server.
+        Return a model. Called by the server.
         """
         model_class = self.config['server']['class']
         model = None
@@ -453,7 +454,8 @@ class CodaLabManager(object):
         :param auth_handler: AuthHandler through which to authenticate
         :return: access token
         """
-        auth = self.state['auth'].get(cache_key, {})
+        state = self.get_state()
+        auth = state['auth'].get(cache_key, {})
 
         def _cache_token(token_info, username=None):
             '''
@@ -474,7 +476,7 @@ class CodaLabManager(object):
             token_info = auth['token_info']
             expires_at = token_info.get('expires_at', 0.0)
 
-            # If token is not nearing expiration, just return it.
+            # If token is not nearing expiration (more than ten minutes left), just return it.
             if expires_at >= (time.time() + 10 * 60):
                 return token_info['access_token']
 
@@ -486,7 +488,7 @@ class CodaLabManager(object):
                 return _cache_token(token_info)
 
         # If we get here, a valid token is not already available.
-        auth = self.state['auth'][cache_key] = {}
+        auth = state['auth'][cache_key] = {}
 
         username = os.environ.get('CODALAB_USERNAME')
         password = os.environ.get('CODALAB_PASSWORD')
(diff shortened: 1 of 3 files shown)


percyliang:
Ah! Good find.  Just checking - are there any performance ramifications with reading/writing `state.json` more frequently?  Also, should we worry about race conditions (I remember @nelson-liu ran into some issues here)? 

It seems like in the case where you have a screen session (and therefore one CODALAB_SESSION) with a worker manager, the codalab worker manager shouldn't even have a codalab session so it shouldn't bother you...

teetone:
> Ah! Good find. Just checking - are there any performance ramifications with reading/writing `state.json` more frequently? Also, should we worry about race conditions (I remember @nelson-liu ran into some issues here)?

The client would need to read from the json file a lot more frequently, but I think the cost is minimal. I will sync up with Nelson to learn more about the possible race conditions.

> It seems like in the case where you have a screen session (and therefore one CODALAB_SESSION) with a worker manager, the codalab worker manager shouldn't even have a codalab session so it shouldn't bother you...

Looking at the code, it looks like the worker manager needs an instance of `CodaLabManager` and a client to search for bundles.

percyliang:
We could have a flag where the `CodaLabManager` doesn't create a session which is set when it's invoked by the worker manager?

teetone:
> We could have a flag where the `CodaLabManager` doesn't create a session which is set when it's invoked by the worker manager?

Okay, I see, that makes sense. I think it also stores the token information in the state after authenticating or refreshing a session, but the worker manager's `CodaLabManager` can just update the state in memory without having to write to the json file. I will test this out.

nelson-liu:
> Also, should we worry about race conditions (I remember @nelson-liu ran into some issues here)?

I think the issue with race conditions is as pertinent as it was before, and this change doesn't really make it any worse (although it's still a problem). I actually have still been maintaining a personal branch with the sqlite3 state manager, just because I need to be able to handle lots of parallel codalab processes at once (though I think I'm perhaps the only person who uses codalab in this way, so I think it's fine if it's not in master)

codalab/lib/codalab_manager.py
```diff
@@ -555,15 +558,22 @@ def check_version(self, server_version):
...
+        if address in state['auth']:
+            del state['auth'][address]
             self.save_state()
 
     def save_config(self):
         if self.temporary:
             return
         write_pretty_json(self.config, self.config_path)
 
+    def get_state(self):
```
epicfaace, line 571: This isn't absolutely necessary for this PR, but it might be cleaner (and maybe be good as a future improvement to improve readability / reduce mistakes) to just make `this.state` a property, then define getters and setters for it as shown here: https://docs.python.org/3/library/functions.html#property

tests/unit/lib/codalab_manager_test.py
```diff
@@ -0,0 +1,102 @@
...
+    def test_get_state_for_temp_codalab_manager(self):
+        manager: CodaLabManager = CodaLabManager(temporary=True)
+        self.assertEqual(manager.get_state(), {'auth': {}, 'sessions': {}})
+        manager.save_state()
+        self.assertFalse(
+            os.path.exists(manager.state_path),
+            msg='Assert that the current state is not written out to state_path for a temporary CodaLabManager',
+        )
+
+    def test_state_persistence(self):
```
epicfaace, line 32: Can you add docstrings for both tests briefly explaining what they're testing? (they'd be a good addition to the existing inline comments)

epicfaace:
Closing this in favor of #3151. @nelson-liu would you be able to make a separate issue explaining more about the race conditions (so it can be addressed in a future PR)?

Issue #3064, opened 2020-11-13
codalab-service doesn't have a test


yiboliu:
**Describe the bug**
codalab-service doesn't have a test in cli to test if it can start successfully. Need to add one. 

**Codalab Information**
- *Your username*: yiboliu

**Expected behavior**
Should have a test

PR #3147, opened 2021-01-05
Create a test for codalab service start

yiboliu:
Add a test for codalab service start

### Related issues
#3064 

### Checklist

### DIFF
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -153,7 +153,7 @@ CODALAB_ARGUMENTS = [
     CodalabArg(
         name='version',
         help='Version of CodaLab (usually the branch name)',
-        default=get_default_version(),
+        default="master",
         flag='-v',
     ),
     CodalabArg(
--- a/codalab_service.py
+++ b/codalab_service.py
@@ -150,12 +150,7 @@ class CodalabArg(object):
 
 CODALAB_ARGUMENTS = [
     # Basic settings
-    CodalabArg(
-        name='version',
-        help='Version of CodaLab (usually the branch name)',
-        default="master",
-        flag='-v',
-    ),
+    CodalabArg(name='version', help='Version of CodaLab (usually the branch name)', flag='-v',),
     CodalabArg(
         name='instance_name',
         help='Instance name (prefixed to Docker containers)',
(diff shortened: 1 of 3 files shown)


codalab_service.py
```diff
@@ -153,7 +153,7 @@ def has_callable_default(self):
     CodalabArg(
         name='version',
         help='Version of CodaLab (usually the branch name)',
```
teetone: Why is this being hardcoded?

codalab_service.py
```diff
@@ -153,7 +153,7 @@ def has_callable_default(self):
...
-        default=get_default_version(),
+        default="master",
```
epicfaace: @teetone the problem was that `get_default_version()` gives an error when run on the tests, because it ends up calling git, but the docker containers don't have git installed. However, I don't think this change is what we want.

@yiboliu -- To avoid calling `get_default_version()` all the time, let's instead do:

```suggestion
        default=None
```

And in the code, later, we can do:

```
args.version = args.version or get_default_version()
```

Finally, in test_cli.py, we can do:

```
  _run_command(['codalab-service', '--help', '--version', 'master'], expected_exit_code=0)
```

epicfaace (changes requested):

codalab_service.py
yiboliu: @teetone @epicfaace oh, I actually planned to just hard-code it here and refine it later in another PR, since the work seems not related to this topic. As we discussed last time, I planned to rename the `version` to `branch` to avoid the confusion with the version variable in setup.py and keep using this `get_default_version()` at that time. What's the scenario that we need the user to manually configure the branch name? I think the correct value of the branch can always get retrieved from the function right?

epicfaace: I don't think we want to hard-code the version at all, though. Even if it's just temporary, we shouldn't merge a PR to master unless we're okay with releasing it.

The scenario for allowing the user to manually configure the branch name is that if they want to run a particular tag / version of codalab, they can specify it. For example, instead of having to run `git checkout v0.5.2`, to run version 0.5.2, you can just specify `--version v0.5.2` to run it that way.

yiboliu: Ok, then that makes sense to me.

epicfaace (approved):

PR #3151, opened 2021-01-08
Use a temporary CodaLabManager for WorkerManagers

### LINKED ISSUE #3078: CLI frequently forgets which worksheet I am on

teetone:

The changes in this PR is an alternative way to fix the sessions issue. A temporary `CodaLabManager` maintains its state in memory and never writes to `state.json`. Therefore, a WorkerManager in a different screen session wouldn't be able to erase sessions information stored in `state.json`.

### Related issues

Related PR: https://github.com/codalab/codalab-worksheets/pull/3145
Resolves https://github.com/codalab/codalab-worksheets/issues/3078

### Checklist

### DIFF
--- a/codalab/worker_manager/worker_manager.py
+++ b/codalab/worker_manager/worker_manager.py
@@ -89,7 +89,7 @@ class WorkerManager(object):
 
     def __init__(self, args):
         self.args = args
-        self.codalab_manager = CodaLabManager()
+        self.codalab_manager = CodaLabManager(args.temp_session)
         self.codalab_client = self.codalab_manager.client(args.server)
         self.staged_uuids = []
         self.worker_manager_start_time = time.time()
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -121,7 +121,7 @@ class CodaLabManager(object):
     temporary: don't use config files
     '''
 
-    def __init__(self, temporary=False, config=None, clients=None):
+    def __init__(self, temporary=False, config={}, clients={}):
         self.cache = {}
         self.temporary = temporary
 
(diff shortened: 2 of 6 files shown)


epicfaace:
Are you sure this solves #3078 in all cases? If a user has multiple sessions open (without a session created by a WorkerManager), I think this issue would still happen, right?

teetone:
> Are you sure this solves #3078 in all cases? If a user has multiple sessions open (without a session created by a WorkerManager), I think this issue would still happen, right?

We synced about this in our meeting, but this will only fix the worker manager problem.

codalab/worker_manager/worker_manager.py
```diff
@@ -89,7 +89,7 @@ def add_arguments_to_subparser(subparser: ArgumentParser) -> None:
 
     def __init__(self, args):
         self.args = args
-        self.codalab_manager = CodaLabManager()
+        self.codalab_manager = CodaLabManager(args.temp_session)
```
epicfaace, line 92: For clarity

```suggestion
        self.codalab_manager = CodaLabManager(temporary=args.temp_session)
```

epicfaace (approved):

teetone:
I will wait for @nelson-liu's approval too just in case.

nelson-liu (approved):

PR #3152, opened 2021-01-08
Remove "last modified date" from frontend

epicfaace:
Sorry, I think I missed this in my review of #3073. Right now, the "last modified date" shows up prominently under the title:

image

I don't think it should be in such a prominent place, given that the title bar is also sticky (so we're taking up additional screen space from the user), and the user would likely not have to refer to the "last modified date" very much often. We should probably rethink where to add it on the frontend -- maybe we add a button at the top right saying "worksheet info" that shows this information -- but I don't think it belongs below the title. I'm just reverting this change in this PR, but we should consider where to add it later in the future.

What do you think, @yuqijin ?


epicfaace:
Replaced by #3156

PR #3157, opened 2021-01-09
Add messages to stress test

yiboliu:
Add messages to stress test to show parts of the test has finished

### Checklist

### DIFF
--- a/tests/stress/stress_test.py
+++ b/tests/stress/stress_test.py
@@ -117,26 +117,37 @@ class StressTestRunner:
         print('Running stress tests...')
         self._start_heartbeat()
         self._test_large_bundle()
+        print('_test_large_bundle finished')
         self.cleanup()
         self._test_many_gpu_runs()
+        print('_test_many_gpu_runs finished')
         self.cleanup()
         self._test_multiple_cpus_runs_count()
+        print('_test_multiple_cpus_runs_count finished')
         self.cleanup()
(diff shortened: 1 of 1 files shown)


teetone:
This change is not necessary. You can just check the status page: https://worksheets-dev.codalab.org/worksheets/0x0174eb8eb3ef43cd9394dff419982e46

epicfaace:
It might be useful to show all messages on the stress test output so it's self-contained (because the stress test status page can change once another stress test is run / bundles are manually stopped) -- what do you think @teetone ?

teetone:
> It might be useful to show all messages on the stress test output so it's self-contained (because the stress test status page can change once another stress test is run / bundles are manually stopped) -- what do you think @teetone ?

I rarely check the standard logs just because the output is very large and the status page shows the current test that is running. I only check the output when the script has stopped running and there is an error message. I'm fine with this change too though.

epicfaace (approved):

epicfaace (changes requested):
Actually, can you make all these messages print out in this function (https://github.com/codalab/codalab-worksheets/blob/425e33d59a738888a500ba3c4da62a8f951b9cef/tests/stress/stress_test.py#L114)?

This will make it easier to keep track of messages, rather than putting the messages directly in the function definitions themselves.

epicfaace (approved):

PR #3159, opened 2021-01-09
Replace a bad image tag

yiboliu:
jenkins:latest doesn't work anymore. Even though it exists [here](https://hub.docker.com/_/jenkins), `docker pull jenkins:latest` or `docker pull alpine` works neither locally nor on the dev server. I replaced it with a specific version tag according to this [article](https://vsupalov.com/docker-latest-tag/). 

### Checklist

### DIFF
--- a/tests/stress/stress_test.py
+++ b/tests/stress/stress_test.py
@@ -85,7 +85,7 @@ class StressTestRunner:
         'couchbase:latest',
         'golang:latest',
         'iwane/numpy-matplotlib',
-        'jenkins:jenkins',
+        'jenkins/jenkins',
         'larger/rdp:dev',
         'maven:latest',
         'mysql:latest',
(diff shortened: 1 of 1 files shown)


nelson-liu:
The Docker image registry says that the `jenkins` image is deprecated, and that you should use https://hub.docker.com/r/jenkins/jenkins instead (`jenkins/jenkins`) ---maybe we should switch?

tests/stress/stress_test.py
```diff
@@ -85,7 +85,7 @@ class StressTestRunner:
         'couchbase:latest',
         'golang:latest',
         'iwane/numpy-matplotlib',
-        'jenkins:latest',
+        'jenkins:jenkins',
```
teetone: Should it be `jenkins/jenkins:latest`?

teetone (changes requested):

tests/stress/stress_test.py
yiboliu: the default tag is `latest`. So it will automatically use `latest`

teetone: No, I meant it should be `jenkins/jenkins`, right?

teetone (approved):

PR #3160, opened 2021-01-10
Use a temporary CodaLabManager for unit tests

epicfaace:

When I was reviewing #3151, I realized that the transient CodaLabManager created before unit tests should also be set to `temporary=True`. This should make tests a bit faster, and it should also make it easier to parallelize tests in the future (plus it's best for unit tests to have no side effects).

### DIFF
--- a/tests/unit/server/bundle_manager/__init__.py
+++ b/tests/unit/server/bundle_manager/__init__.py
@@ -68,7 +68,7 @@ class TestBase:
     """
 
     def setUp(self):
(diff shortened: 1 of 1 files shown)


teetone (approved):
It looks like the test pipeline is failing, but the change looks good me.

epicfaace:
Actually, going to close this. Realized that a temporary CodaLabManager doesn't set the proper default settings (such as DB URL) that are actually needed in the tests.

PR #3167, opened 2021-01-11
Fix flaky BundlesTest

### LINKED ISSUE #3166: BundlesTest is flaky
Sometimes, the last_updated attribute is added for some reason, but we don't always expect the last_updated attribute to exist.

image

epicfaace:

Fixes #3166. I'm not sure why last_updated is only sometimes present, but for now, I'm making the test remove that field before comparing equality, so that the test stops nondeterministically failing.


epicfaace:
@yiboliu can you review when you get a chance?

yiboliu (approved):

PR #3169, opened 2021-01-11
Fix log message when filtering out bundles in the worker manager code

teetone:

Log the memory of the workers when filtering out a bundle in `worker_manager.py`.


nelson-liu (approved):

nelson-liu:
oops, sorry for not catching this earlier

PR #3172, opened 2021-01-12
Remove last modified date

### LINKED ISSUE #3156: Format date modified to be more human-friendly
Use https://github.com/hustcc/timeago.js

yuqijin:
### Related issues

fixes #3156 

### Screenshots

before:
Screen Shot 2021-01-11 at 9 10 46 PM

after:
Screen Shot 2021-01-11 at 9 11 38 PM

### Checklist

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -16647,6 +16647,11 @@
       "integrity": "sha512-eHY7nBftgThBqOyHGVN+l8gF0BucP09fMo0oO/Lb0w1OF80dJv+lDVpXG60WMQvkcxAkNybKsrEIE3ZtKGmPrA==",
       "dev": true
     },
+    "timeago.js": {
+      "version": "4.0.2",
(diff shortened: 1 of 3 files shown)


percyliang:
This increases the amount of vertical space dedicated to header info, which is non-ideal.  Is there a way to not do that?  I think it might be more useful for the dashboard.

yuqijin:
> This increases the amount of vertical space dedicated to header info, which is non-ideal.  Is there a way to not do that?  I think it might be more useful for the dashboard.

I see. How about displaying the date in the old format for the worksheets, and display the date in the new format on the dashboard?

yuqijin:
Remove the last_modified_date from regular worksheets to avoid extra vertical space in the header.
Screen Shot 2021-01-12 at 1 09 27 AM

Will add it for the worksheets displayed on the dashboard in a human-friendly format instead.

epicfaace:
Okay, that sounds good. I'll merge this PR for now so that we don't show the last modified date on the header in prod.

epicfaace (approved):

PR #3178, opened 2021-01-13
Disable the terminal when focusing on the search bar

### LINKED ISSUE #2367: Worksheet search bar text can be mirrored in console
1. Open the codalab web CLI
2. type some text, don't press enter
3. go to the search bar, type some text. That text will be appended to the cli command string.

yuqijin:

Previously, the search bar text will be automatically mirrored to the open terminal. It is because the terminal is still in an activated state when we move focus to the search bar.
Also, I removed the unused `handleChange` function.

### Related issues

fixes #2367 
### Screenshots

ezgif com-gif-maker

### Checklist

### DIFF
--- a/frontend/src/components/NavBar.js
+++ b/frontend/src/components/NavBar.js
@@ -141,8 +141,6 @@ class NavBar extends React.Component<{
         });
     }
 
-    handleChange = (e, { value }) => this.setState({ value });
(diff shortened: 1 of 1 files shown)


epicfaace (approved):
Looks good!

PR #3194, opened 2021-01-19
Separate unit test to another row

yiboliu:
Split the unit test to another row to improve the efficiency and debug-ability. 
Before the change, the time for unit tests and all the other tests in that line to finish is about 9m30s to 10m. 
After the change, the time for unit tests to finish is 4m and for other tests in that line to finish is 8m44s. 

### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,8 +80,9 @@ jobs:
     strategy:
       matrix:
         test:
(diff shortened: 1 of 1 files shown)


teetone:
There is a cap to the number of processes that can run in parallel in GHA, so this change might actually make the build process longer. It looks like the build is failing for your PR, but can you later include how long a passing build takes before and after this change in the PR description?

teetone:
@yiboliu With your changes, the total time the build takes is actually 19 minutes now which makes it about 2 minutes slower:

Screen Shot 2021-01-20 at 11 29 51 PM

teetone:
@yiboliu I don't think we should make this change. Can we just close this for now?

PR #3197, opened 2021-01-19
Fix bug of cursoring with keyboard shortcut after cutting a bundle

### LINKED ISSUE #3176: cut interacts badly with cursoring
To reproduce, create a worksheet with three bundles b1, b2, b3.  Cursor over to b2 and detach it using 'a d'.  The cursor is now on b3.  Hit 'k' and it will skip over b1.

yuqijin:

The bug is due to we use keyboard shortcut `a d` to cut a bundle, but `a` is not a standard modifier in Mousetrap, and there also exists a keyboard shortcut `a k`, so `a` is treated as `k` in Mousetrap's array. We need to reset the array after using the keyboard shortcut to avoid the cursoring was executed twice when we only press one `k`

### Related issues

fixes #3176 

### Screenshots

ezgif com-gif-maker (1)

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -1056,6 +1056,7 @@ class Worksheet extends React.Component {
                     return;
                 }
                 this.toggleCmdDialogNoEvent('copy');
+                Mousetrap.reset();
             });
             if (this.state.ws.info.edit_permission) {
                 Mousetrap.bind(['backspace', 'del'], () => {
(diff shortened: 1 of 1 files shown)


epicfaace (approved):
Looks good! I'm still a bit confused as to what exactly was the problem with Mousetrap, though, what do you mean when you say "a is treated as k in Mousetrap's array"?

yuqijin:
> Looks good! I'm still a bit confused as to what exactly was the problem with Mousetrap, though, what do you mean when you say "a is treated as k in Mousetrap's array"?

Not quite sure though, but I suppose it is due to some internal mechanism of mousetrap. I have such thought because the bug only exists for `k`, but not for any other button without such binding, like `m` or `up`.

PR #3198, opened 2021-01-20
Allow configuring and uploading to Azure Blob Storage (split up into different PRs)

epicfaace:
Allow configuring and uploading to Azure Blob Storage. Add the `CODALAB_AZURE_BLOB_CONNECTION_STRING` and `CODALAB_ALWAYS_USE_AZURE_BLOB_BETA` environment variables (which are documented in the PR).

Running `cl upload -a` will use Blob Storage to store the file.

Don't merge this PR; it's broken down into several PRs, all listed on this megaissue:

**Azure Blob Storage megaissue - PR breakdown - #3298**


codalab/lib/beam/seekable_zipfile.py
```diff
@@ -0,0 +1,2143 @@
+# type: ignore
```
epicfaace: Had to include this ZipFile shim from Python 3.7 in order to get .zip file handling to properly work. If / when we switch back to a different archive format (.tar.gz), we can remove this.

codalab/worker/file_util.py
```diff
@@ -200,11 +205,65 @@ def do_unzip(filename):
...
+    ):
+        # If file path is a zip file on Azure, open the specified path within the
+        # zip file. zipfile.open only supports 'r', not 'rb', so we always pass in 'r'.
+        with SeekableZipFile(
+            FileSystems.open(linked_bundle_path.bundle_path, compression_type)
+        ) as f:
+            zinfo = f.getinfo(linked_bundle_path.zip_subpath + ("/" if sub_zip_directory else ""))
+            if zinfo.is_dir():
+                # If streaming a folder within an Azure bundle, we need to download its contents,
+                # re-archive the folder, and return the .tar.gz version of that folder.
```
epicfaace: Note that we have to extract and re-zip parts of the archive if the user is streaming a particular folder that is located within a bundle.

codalab/lib/beam/seekable_zipfile.py
yiboliu: I understand the comments below. But why the first line? what is it ignoring?

yiboliu (commented):
This PR is gigantic. Do you want to split it into smaller ones, so that it's easier to debug in case it breaks anything?

codalab/lib/beam/seekable_zipfile.py
epicfaace: It ensures that flake8 isn't run on this file.

epicfaace:
@yiboliu sure, I can split it up.

PR #3201, opened 2021-01-21
Add test for stats retrieval

yiboliu:
Add a test for stats retrieval

### Related issues
#3003 

### Checklist

### DIFF
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -211,22 +211,6 @@ def wait(uuid, expected_exit_code=0):
         raise e
 
 
-def wait_until_running(uuid: str) -> None:
-    # Wait until the bundle is running. The maximal waiting time is 30s.
-    running: bool = False
-    for i in range(10):
-        state: str = get_info(uuid, 'state')
-        if state == 'running':
-            running = True
-            break
-        elif state == 'failed':
-            assert False, f'the bundle {uuid} has failed'
-        else:
-            time.sleep(3)
-    if not running:
-        assert False, f'the bundle {uuid} is not ready after running 30s'
-
-
 def check_equals(true_value, pred_value):
     assert true_value == pred_value, "expected '%s', but got '%s'" % (true_value, pred_value)
     return pred_value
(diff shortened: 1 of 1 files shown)


tests/cli/test_cli.py
```diff
@@ -1470,6 +1470,27 @@ def test_run2(ctx):
         [cl, 'run', 'foo:%s' % dir3, 'foo/bar:%s' % dir1, 'cat foo/bar/f1'], expected_exit_code=1
     )
 
+    # Test that cpu_usage and memory_limit are properly populated
+    uuid = _run_command([cl, 'run', "bash -c 'for i in {1..90}; do sleep 1; done'"])
+    state = get_info(uuid, 'state')
+    # Wait until the bundle is running
```
epicfaace: Can we just replace this loop with `wait(uuid)`?

yiboliu: Oh I thought `wait(uuid)` is waiting for the bundle to finish. If it's waiting for it to start run, yes.

yiboliu: Looks like `wait(uuid)` is waiting for the bundle to finish as I experimented. When a bundle is finished, the `cpu_usage` is 0. So it's probably better to stick with the loop. I can make it as a method like `wait_for_bundle_running` if you like.

tests/cli/test_cli.py
```diff
@@ -211,6 +211,16 @@ def wait(uuid, expected_exit_code=0):
         raise e
 
 
+def wait_until_running(uuid):
+    # Wait until the bundle is running
+    state = get_info(uuid, 'state')
+    while state != 'running':
```
teetone: How long does it usually take for a bundle to start running? Also, this can loop infinitely if the `state` is never `running`. Can we fail after certain number of iterations?

tests/cli/test_cli.py
```diff
@@ -211,6 +211,16 @@ def wait(uuid, expected_exit_code=0):
...
+        if state == 'failed':
+            assert False
+        time.sleep(6)
```
teetone: Why 6 seconds? Can we shorten the time so the test runs faster?

tests/cli/test_cli.py
```diff
@@ -1470,6 +1480,21 @@ def test_run2(ctx):
         [cl, 'run', 'foo:%s' % dir3, 'foo/bar:%s' % dir1, 'cat foo/bar/f1'], expected_exit_code=1
     )
 
+    # Test that cpu_usage and memory_limit are properly populated
+    uuid = _run_command([cl, 'run', "bash -c 'for i in {1..90}; do sleep 1; done'"])
+    wait_until_running(uuid)
+
+    for i in range(10):
```
teetone: Can you add a comment before the for-loop that explains what is being tested here?

tests/cli/test_cli.py
```diff
@@ -1470,6 +1480,21 @@ def test_run2(ctx):
         [cl, 'run', 'foo:%s' % dir3, 'foo/bar:%s' % dir1, 'cat foo/bar/f1'], expected_exit_code=1
     )
 
+    # Test that cpu_usage and memory_limit are properly populated
+    uuid = _run_command([cl, 'run', "bash -c 'for i in {1..90}; do sleep 1; done'"])
```
teetone: I think 90 seconds is too long. I think 30 seconds should be enough.

tests/cli/test_cli.py
```diff
@@ -1470,6 +1480,21 @@ def test_run2(ctx):
...
+    for i in range(10):
+        _run_command([cl, 'info', uuid])
+        if float(get_info(uuid, 'cpu_usage')) <= 0 or int(get_info(uuid, 'memory_limit')) <= 0:
+            # In this case, we would like to retry, in case that docker stats are not fully ready
+            time.sleep(6)
+        else:
+            # When we have both stats ready, we are done testing.
+            break
+        if i == 9:
+            assert False
```
teetone: Can you add a message to the assert?

teetone (changes requested):

tests/cli/test_cli.py
```diff
@@ -1470,6 +1486,22 @@ def test_run2(ctx):
         [cl, 'run', 'foo:%s' % dir3, 'foo/bar:%s' % dir1, 'cat foo/bar/f1'], expected_exit_code=1
     )
 
+    # Test that cpu_usage and memory_limit are properly populated
+    uuid = _run_command([cl, 'run', "bash -c 'for i in {1..30}; do sleep 1; done'"])
+    wait_until_running(uuid)
```
epicfaace: I think we can replace this function with just using the existing function, `wait_until_state`.

teetone (commented):
Waiting for the docker stats feature to get reimplemented.

epicfaace:
Closing this for now.

Issue #3187, opened 2021-01-15
Set up sentry integration with worksheets / dev / prod environments


epicfaace:
Set up sentry integration with:
- worksheets.codalab.org - should be the "prod" environment
- worksheets-dev.codalab.org - should be the "dev" environment

This will make it easier to track common backend errors. Right now, we only use sentry to track errors coming from workers on codalab.stanford.edu.

We should also set up integration with [Bottle](https://docs.sentry.io/platforms/python/guides/bottle/) so that we can start adding logs from the REST server.

epicfaace:
Ask @nelson-liu

PR #3207, opened 2021-01-22
Integrate sentry with bottle

yiboliu:
Integrate bottle with sentry, so that errors occurring in rest server, which uses bottle, can be reported to sentry. 
Ref - https://docs.sentry.io/platforms/python/guides/bottle/

### Related issues
#3187 

### Checklist

### DIFF
--- a/codalab/server/rest_server.py
+++ b/codalab/server/rest_server.py
@@ -50,10 +50,7 @@ import codalab.rest.worksheets
 
 logger = logging.getLogger(__name__)
 
-sentry_sdk.init(
-    dsn="https://a9654419762745c3b255bf299f28a656@o431750.ingest.sentry.io/5383515",
-    integrations=[BottleIntegration()]
-)
+sentry_sdk.init(dsn=os.getenv('CODALAB_SENTRY_INGEST_URL'), integrations=[BottleIntegration()])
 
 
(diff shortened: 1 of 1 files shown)


codalab/server/rest_server.py
```diff
@@ -48,6 +50,11 @@
 
 logger = logging.getLogger(__name__)
 
+sentry_sdk.init(
+    dsn="https://a9654419762745c3b255bf299f28a656@o431750.ingest.sentry.io/5383515",
```
epicfaace: We shouldn't hard code this URL. Instead, we should use the CODALAB_SENTRY_INGEST_URL environment variable.

epicfaace (approved):

epicfaace:
@yiboliu I don't see any REST API logs in sentry.io -- do you know what is going on? Is Sentry being properly initialized here?

yiboliu:
> @yiboliu I don't see any REST API logs in sentry.io -- do you know what is going on? Is Sentry being properly initialized here?

Sorry, what logs are you looking for? like this one? - `2021-02-01 07:39:18,498 http://localhost:None "POST /v1.39/networks/create HTTP/1.1" 409 None`

epicfaace:
I'm only seeing things from workers, not any logs that seem to correspond to REST API calls:

image

PR #3212, opened 2021-01-23
Speed up CLI by lazy loading some imports

### LINKED ISSUE #3211: "cl" has a slow startup time
When typing "cl" plus anything, there's at least a few seconds of startup time before the command is actually run. This makes the CodaLab CLI just feel pretty slow for users in general.

We should look at resources such as https://dev.to/methane/how-to-speed-up-python-application-startup-time-nkf to see how we can best improve the CLI startup time.

Some options:
- Speed up imports by lazily importing / removing unnecessary imports
- Run a codalab client process in the background, and the CLI just communicates with it

epicfaace:
Fixes #3211.

Speeds up the CLI by lazy loading imports for the following modules:
- apache_beam
- docker

These don't need to be loaded on startup. This changes the import delay from ~2 seconds to around ~1 second (and the CodaLab CLI feels faster too!)

image


epicfaace:
@percyliang I think it would eventually be best to split out all the functions in file_util, zip_util, etc. into one file per function (so that we can minimize unnecessary imports to the greatest extent possible), but in this PR I'm just doing the minimum necessary to stop the most heavy imports (Apache Beam, Docker, SQLAlchemy) from being loaded on CLI startup.

tests/cli/test_cli.py
```diff
@@ -1639,6 +1639,15 @@ def test_status(ctx):
     check_contains("Commands for bundles", cl_output)
     check_equals(cl_output, help_output)
 
+    # Ensure that the "cl help" command doesn't take too long to start up.
```
epicfaace: I'm adding a test here to try to ensure this doesn't regress. At least, I hope that if the regular "cl" command suddenly takes a long time to start up, this will cause people to look into it in the future and ensure no large imports are taking up too much time.

epicfaace: Actually, removing this test because it seems too flaky (it sometimes takes up to 3.5 seconds in CI)

codalab/lib/upload_manager.py
```diff
@@ -13,11 +13,14 @@ class UploadManager(object):
     """
 
     def __init__(self, bundle_model, bundle_store):
+        from codalab.lib import zip_util
```
yiboliu, line 16: question - why does making imported module a variable can accelerate running? Since the importing also happens when the constructor is called, seems like the importing isn't delayed too much.

yiboliu (commented):
Do you think using `__getattr__ ` could be a better way? - https://snarky.ca/lazy-importing-in-python-3-7/

codalab/lib/upload_manager.py
epicfaace: When it's moved to the constructor, it's only called when someone calls something like `cl upload`, which initializes `UploadManager`. It won't be called if someone makes another command such as `cl --help` or `cl work`, though.

epicfaace:
> Do you think using `__getattr__ ` could be a better way? - https://snarky.ca/lazy-importing-in-python-3-7/

I'm not exactly clear as to how that works / what the blog post is saying; do you know what it is suggesting?

yiboliu:
> > Do you think using `__getattr__ ` could be a better way? - https://snarky.ca/lazy-importing-in-python-3-7/
> 
> I'm not exactly clear as to how that works / what the blog post is saying; do you know what it is suggesting?

Basically it's suggesting using this module - https://pypi.org/project/lazy-import/. As mentioned in the description, with this module, the cost of importing only occurs when the imported module is actually used.

epicfaace:
Hmm, I'm not sure if I'm comfortable using that library -- it doesn't seem to be widely used (~52 stars) and doesn't seem to be actively maintained (last commit was in January 2018). https://github.com/mnmelo/lazy_import

If we can find another, more widely used / maintained library for lazy imports, we can use that; otherwise, I'm most comfortable just using Python imports instead.

yiboliu:
> Hmm, I'm not sure if I'm comfortable using that library -- it doesn't seem to be widely used (~52 stars) and doesn't seem to be actively maintained (last commit was in January 2018). https://github.com/mnmelo/lazy_import
> 
> If we can find another, more widely used / maintained library for lazy imports, we can use that; otherwise, I'm most comfortable just using Python imports instead.

Ok, sounds good to me.

epicfaace:
Are you good to approve this PR then @yiboliu ?

yiboliu (approved):

Issue #3208, opened 2021-01-22
Large worksheets do not properly refresh


epicfaace:
Steps to reproduce:
- Create a large worksheet with ~200 bundles.
- Do something to make the page refresh, for example, kill some bundles.
- The following shows up:
![image](https://user-images.githubusercontent.com/1689183/105502652-a6018f00-5c93-11eb-92d3-4fa7a0ee3443.png)

It appears that all the bundle IDs are being sent over to the backend here, which causes the refresh request to fail.

This has to do with #1999 as well.

PR #3216, opened 2021-01-25
Bundle status refresh endpoint refreshes only the specified bundle 

### LINKED ISSUE #1999: Bundle status refresh endpoint refreshes the entire worksheet
It should refresh just the particular bundle IDs specified in the query string.

yuqijin:

We had a `partial update` mechanism in our system, however, it seems like this mechanism lost consistency between backend and frontend during the development, so even if a bundle status refresh endpoint always refreshes the entire worksheet. Now the bugs are fixed and only the bundles that need updating aren't `None` during network transmission.

### Related issues

fixes #1999 

### Screenshots
Before: 
Screen Shot 2021-01-24 at 6 54 41 PM

After:
Screen Shot 2021-01-24 at 6 53 16 PM
Screen Shot 2021-01-24 at 6 47 10 PM

### Checklist

### DIFF
--- a/codalab/rest/interpret.py
+++ b/codalab/rest/interpret.py
@@ -300,17 +300,17 @@ def fetch_interpreted_worksheet(uuid):
     # The bundle_infos for bundles that don't need updating are also None.
     if bundle_uuids:
         for i, block in enumerate(interpreted_blocks['blocks']):
-            if 'bundle_info' not in block:
+            if not ('bundles_spec' in block and 'bundle_infos' in block['bundles_spec']):
                 interpreted_blocks['blocks'][i] = None
             else:
-                if isinstance(block['bundle_info'], dict):
-                    block['bundle_info'] = [block['bundle_info']]
+                if isinstance(block['bundles_spec']['bundle_infos'], dict):
+                    block['bundles_spec']['bundle_infos'] = [block['bundles_spec']['bundle_infos']]
                 is_relevant_block = False
-                for j, bundle in enumerate(block['bundle_info']):
+                for j, bundle in enumerate(block['bundles_spec']['bundle_infos']):
                     if bundle['uuid'] in bundle_uuids:
                         is_relevant_block = True
                     else:
-                        block['bundle_info'][j] = None
+                        block['bundles_spec']['bundle_infos'][j] = None
                 if not is_relevant_block:
                     interpreted_blocks['blocks'][i] = None
     # Grouped individual items into blocks
@@ -347,7 +347,7 @@ def fetch_interpreted_worksheet(uuid):
     # Frontend doesn't use individual 'items' for now
     del worksheet_info['items']
     if bundle_uuids:
-        return {'blocks': worksheet_info['blocks']}
+        return {'blocks': worksheet_info['blocks'], 'uuid': uuid}
     return worksheet_info
 
 
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -1531,6 +1531,8 @@ class Worksheet extends React.Component {
                         'bundle_infos' in partialUpdateItems[i]['bundles_spec']
                     )
                 )
+                    // Partial Update mechanism only designs for the blocks consisting of bundles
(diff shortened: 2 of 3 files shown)


codalab/rest/interpret.py
```diff
@@ -300,17 +300,17 @@ def fetch_interpreted_worksheet(uuid):
     # The bundle_infos for bundles that don't need updating are also None.
     if bundle_uuids:
         for i, block in enumerate(interpreted_blocks['blocks']):
```
epicfaace, line 302: This looks good. Could we add a test within the `test/unit/rest` directory that ensures that the partial refresh endpoint never regresses again?

yuqijin: Sure. The test has been added

codalab/rest/interpret.py
```diff
@@ -347,7 +347,7 @@ def fetch_interpreted_worksheet(uuid):
     # Frontend doesn't use individual 'items' for now
     del worksheet_info['items']
     if bundle_uuids:
-        return {'blocks': worksheet_info['blocks']}
+        return {'blocks': worksheet_info['blocks'], 'uuid': uuid}
```
epicfaace, line 350: What is `uuid` equal to? I thought `bundle_uuids` can specify multiple bundle uuids?

codalab/rest/interpret.py
```diff
@@ -300,17 +300,17 @@ def fetch_interpreted_worksheet(uuid):
     # The bundle_infos for bundles that don't need updating are also None.
     if bundle_uuids:
         for i, block in enumerate(interpreted_blocks['blocks']):
-            if 'bundle_info' not in block:
+            if not ('bundles_spec' in block and 'bundle_infos' in block['bundles_spec']):
```
epicfaace, line 303: Can you explain a bit what your code changes do here? Maybe just in the PR description / a comment here

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1524,9 +1524,26 @@ class Worksheet extends React.Component {
         } else {
             var ws = _.clone(this.state.ws);
             for (var i = 0; i < partialUpdateItems.length; i++) {
-                if (!partialUpdateItems[i]) continue;
+                if (
+                    !partialUpdateItems[i] ||
+                    !(
```
epicfaace, line 1532: Can you add a comment explaining what this is looking for?

codalab/rest/interpret.py
yuqijin: `uuid` refers to the worksheet's uuid. `bundle_uuids` refers to the bundles' uuids.

codalab/rest/interpret.py
yuqijin: It is used to check whether the block contains the field of 'bundle_infos' to determine whether it is a non-None block containing a list of bundles.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Added!

yuqijin:
Also fixes #3208 that not all the bundle id will be included in the GET URL:

Screen Shot 2021-02-02 at 10 17 54 PM

codalab/rest/interpret.py
```diff
@@ -300,17 +300,17 @@ def fetch_interpreted_worksheet(uuid):
...
                 interpreted_blocks['blocks'][i] = None
             else:
-                if isinstance(block['bundle_info'], dict):
-                    block['bundle_info'] = [block['bundle_info']]
+                if isinstance(block['bundles_spec']['bundle_infos'], dict):
+                    block['bundles_spec']['bundle_infos'] = [block['bundles_spec']['bundle_infos']]
```
epicfaace, line 307: Why did you change `block['bundle_info']` to `block['bundles_spec']['bundle_infos']`? Can each block now contain multiple bundle infos (which wasn't possible before)?

yuqijin: Because the object `block` only contains the field of `bundles_spec`, I believe it is because the code has not been updated during development. 
And yes, each block can contain multiple bundle Infos, since a block can be a table, then it will make sense.

epicfaace (approved):

PR #3218, opened 2021-01-25
Fix none concatenation error when interpreting worksheet

### LINKED ISSUE #3217: Server error during interpret worksheet: can't concatenate None + str
Server logs:

```
  File "/opt/codalab-worksheets/codalab/lib/worksheet_util.py", line 926, in interpret_items
    flush_bundles(bundle_block_start_index)
  File "/opt/codalab-worksheets/codalab/lib/worksheet_util.py", line 807, in flush_bundles
    for (name, genpath, post, _) in schema
  File "/opt/codalab-worksheets/codalab/lib/worksheet_util.py", line 807, in <dictcomp>
    for (name, genpath, post, _) in schema
  File "/opt/codalab-worksheets/codalab/lib/worksheet_util.py", line 438, in interpret_genpath
    args.append(friendly_render_dep(dep)[1])
  File "/opt/codalab-worksheets/codalab/lib/worksheet_util.py", line 422, in friendly_render_dep
    + '}'
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

epicfaace:

Fixes #3217. I'm not sure exactly when / why this issue is happening, but `key` is the only possible variable here that can be equal to `None`, so we might as well just give it a fallback to a string "None" if it is equal to `None` so that it doesn't cause this error.

### DIFF
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -412,7 +412,7 @@ def interpret_genpath(bundle_info, genpath, db_model=None, owner_cache=None):
     elif genpath == 'summary':
 
         def friendly_render_dep(dep):
-            key = dep['child_path'] or dep['parent_name']
+            key = dep['child_path'] or dep['parent_name'] or ''
             friendly_parent_name = formatting.verbose_contents_str(dep['parent_name'])
             value = (
                 key
--- a/codalab/lib/worksheet_util.py
+++ b/codalab/lib/worksheet_util.py
@@ -412,7 +412,7 @@ def interpret_genpath(bundle_info, genpath, db_model=None, owner_cache=None):
     elif genpath == 'summary':
 
         def friendly_render_dep(dep):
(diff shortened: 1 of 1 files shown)


yiboliu (changes requested):
Can we use some more descriptive text as the default? Something like `missing child_path and parent_name` could tell us more insight about what's going on.

epicfaace:
I'm not exactly sure why this error is happening, i.e. in what instances `parent_name` will not be defined. I think if / once we know why this is happening, we'll better know what to put as a default value.

epicfaace:
Also, I just changed this PR to fallback to a string "None" instead of an empty string.

yiboliu:
> Also, I just changed this PR to fallback to a string "None" instead of an empty string.

Cool I think 'None' is good enough.

yiboliu (approved):

teetone (approved):

PR #3221, opened 2021-01-25
[DRAFT] Give default values to cpu_usage and memory_limit

yiboliu:
Give default values to cpu_usage and memory_limit when they are missing, in case that workers are not updated so the values are not added. 

### Related issues
#3217 

### Checklist


teetone:
Will reproduce the issue in dev and see if this changes fixes it afterwards.

teetone (approved):

Issue #3204, opened 2021-01-21
Better worker logging


epicfaace:
We currently have not much visibility into what exactly a worker is doing, just by looking at its logs. Most of its logs includes just GET / POST requests to Docker network APIs, and we occasionally have some messages logged (such as " Connected! Successful check in!").

![image](https://user-images.githubusercontent.com/1689183/105380227-a4749000-5bdb-11eb-8b6e-eef768ad42fc.png)

This makes it much harder to have much insight into what workers do, as well as which steps might be slowing down the worker's behavior. It's hard to debug issues about workers when we don't have the right logging available.

We should:
- Add logging for common tasks; for example, logs for "starting bundle X", "uploading results for bundle X", etc.
- Add verbose logging that logs each message sent to / from the worker and the bundle manager. We might want to add a command-line option to enable / disable verbose logging.
- Remove extraneous logs from the worker logs, such as the Docker network logs, so that we get the most relevant information when looking at the logs.

epicfaace:
Whenever there's an error, log stack trace using `traceback.format_exc()`

yiboliu:
we already have a `--verbose` for [worker](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/main.py#L110) and [worker-manager](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/main.py#L28). Are these what are you looking for? @epicfaace

epicfaace:
Could use similar code to this to silence output for Docker: https://github.com/codalab/codalab-worksheets/pull/3198/files#diff-0cd49d98d537f4929cae7db78a7209506ab09f5ce007148dfa05d209fcb19746R24-R25

yiboliu:
call stack of those logs:
```  
  File "/usr/local/bin/cl-worker", line 33, in <module>
    sys.exit(load_entry_point('codalab', 'console_scripts', 'cl-worker')())
  File "/opt/codalab/worker/main.py", line 289, in main
    worker.start()
  File "/opt/codalab/worker/worker.py", line 270, in start
    self.process_runs()
  File "/opt/codalab/worker/worker.py", line 469, in process_runs
    self.init_docker_networks(self.docker_network_prefix, verbose=False)
  File "/opt/codalab/worker/worker.py", line 181, in init_docker_networks
    docker_network_prefix + "_int", internal=True, verbose=verbose
  File "/opt/codalab/worker/worker.py", line 145, in create_or_get_network
[...]

PR #3223, opened 2021-01-26
Add logging for common tasks

yiboliu:
Add logging for common tasks like uploading reports and finalizing bundles. 

### Related issues
#3204 

### Screenshots
```
2021-02-04 08:16:43,481 Connecting to http://rest-server:2900 /opt/codalab/worker/main.py 173
2021-02-04 08:16:43,539 Loaded 0 dependencies, 0 paths from cache. /opt/codalab/worker/dependency_manager.py 111
2021-02-04 08:16:43,567 Looking for auth config /usr/local/lib/python3.6/dist-packages/docker/auth.py 41
2021-02-04 08:16:43,568 Using credentials store "ecr-login" /usr/local/lib/python3.6/dist-packages/docker/auth.py 233
2021-02-04 08:16:43,568 Looking for auth entry for 'https://index.docker.io/v1/' /usr/local/lib/python3.6/dist-packages/docker/auth.py 261
2021-02-04 08:16:43,577 No entry found /usr/local/lib/python3.6/dist-packages/docker/auth.py 277
2021-02-04 08:16:43,577 No entry in credstore - fetching from auth dict /usr/local/lib/python3.6/dist-packages/docker/auth.py 238
2021-02-04 08:16:43,577 Looking for auth entry for 'docker.io' /usr/local/lib/python3.6/dist-packages/docker/auth.py 242
2021-02-04 08:16:43,577 No entry found /usr/local/lib/python3.6/dist-packages/docker/auth.py 253
2021-02-04 08:16:43,577 No auth config found /usr/local/lib/python3.6/dist-packages/docker/auth.py 58
2021-02-04 08:16:44,833 Cannot initialize NVIDIA runtime, no GPU support: Problem getting NVIDIA devices: 400 Client Error: Bad Request ("Unknown runtime specified nvidia") /opt/codalab/worker/docker_utils.py 102
2021-02-04 08:16:44,834 Trying paths: ['/root/.docker/config.json', '/root/.dockercfg'] /usr/local/lib/python3.6/dist-packages/docker/utils/config.py 21
2021-02-04 08:16:44,834 Found file at path: /root/.docker/config.json /usr/local/lib/python3.6/dist-packages/docker/utils/config.py 25
2021-02-04 08:16:44,835 Found 'credsStore' section /usr/local/lib/python3.6/dist-packages/docker/auth.py 189
2021-02-04 08:16:44,835 Trying paths: ['/root/.docker/config.json', '/root/.dockercfg'] /usr/local/lib/python3.6/dist-packages/docker/utils/config.py 21
2021-02-04 08:16:44,835 Found file at path: /root/.docker/config.json /usr/local/lib/python3.6/dist-packages/docker/utils/config.py 25
2021-02-04 08:16:44,835 Found 'credsStore' section /usr/local/lib/python3.6/dist-packages/docker/auth.py 189
2021-02-04 08:16:44,841 Creating docker network codalab-worker-network_general /opt/codalab/worker/worker.py 146
2021-02-04 08:16:44,847 Network codalab-worker-network_general already exists, reusing /opt/codalab/worker/worker.py 153
2021-02-04 08:16:44,854 Creating docker network codalab-worker-network_ext /opt/codalab/worker/worker.py 146
2021-02-04 08:16:44,859 Network codalab-worker-network_ext already exists, reusing /opt/codalab/worker/worker.py 153
2021-02-04 08:16:44,866 Creating docker network codalab-worker-network_int /opt/codalab/worker/worker.py 146
2021-02-04 08:16:44,873 Network codalab-worker-network_int already exists, reusing /opt/codalab/worker/worker.py 153
2021-02-04 08:16:44,882 Worker started! /opt/codalab/worker/main.py 287
2021-02-04 08:16:44,886 Starting docker image manager /opt/codalab/worker/docker_image_manager.py 55
2021-02-04 08:16:44,887 Starting local dependency manager /opt/codalab/worker/dependency_manager.py 164
2021-02-04 08:16:47,966 Connected! Successful check in! /opt/codalab/worker/worker.py 429
```

### Checklist

### DIFF
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -209,7 +209,7 @@ def main():
 
     # Configure logging
     logging.basicConfig(
-        format='%(asctime)s %(message)s', level=(logging.DEBUG if args.verbose else logging.INFO)
+        format='%(asctime)s %(message)s %(pathname)s %(lineno)d', level=(logging.DEBUG if args.verbose else logging.INFO)
     )
 
     logging.getLogger('urllib3').setLevel(logging.INFO)
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -214,7 +214,6 @@ def main():
     )
 
     logging.getLogger('urllib3').setLevel(logging.INFO)
-
     # Initialize sentry logging
     if using_sentry():
         initialize_sentry()
--- a/codalab/worker/main.py
+++ b/codalab/worker/main.py
@@ -209,7 +209,8 @@ def main():
 
     # Configure logging
     logging.basicConfig(
-        format='%(asctime)s %(message)s %(pathname)s %(lineno)d', level=(logging.DEBUG if args.verbose else logging.INFO)
+        format='%(asctime)s %(message)s %(pathname)s %(lineno)d',
+        level=(logging.DEBUG if args.verbose else logging.INFO),
     )
 
     logging.getLogger('urllib3').setLevel(logging.INFO)
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -205,7 +205,10 @@ class RunStateMachine(StateTransitioner):
             #   dependency_path:docker_dependency_path:ro
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
+        logger.info(f'{run_state.bundle.uuid} is preparing')
+
         if run_state.is_killed or run_state.is_restaged:
+            logger.info(f'{run_state.bundle.uuid} is cleaning up')
             return run_state._replace(stage=RunStage.CLEANING_UP)
 
         # Check CPU and GPU availability
@@ -219,6 +222,7 @@ class RunStateMachine(StateTransitioner):
             )
             logger.error(message)
             logger.error(traceback.format_exc())
+            logger.info(f'{run_state.bundle.uuid} is changing state to {message}')
             return run_state._replace(run_status=message)
 
         dependencies_ready = True
@@ -238,6 +242,7 @@ class RunStateMachine(StateTransitioner):
                     dependencies_ready = False
                 elif dependency_state.stage == DependencyStage.FAILED:
                     # Failed to download dependency; -> CLEANING_UP
+                    logger.info(f'{run_state.bundle.uuid} is cleaning up')
                     return run_state._replace(
                         stage=RunStage.CLEANING_UP,
                         failure_message='Failed to download dependency %s: %s'
@@ -256,6 +261,7 @@ class RunStateMachine(StateTransitioner):
             # Failed to pull image; -> CLEANING_UP
             message = 'Failed to download Docker image: %s' % image_state.message
             logger.error(message)
+            logger.info(f'{run_state.bundle.uuid} is cleaning up')
             return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
 
         # stop proceeding if dependency and image downloads aren't all done
@@ -265,6 +271,7 @@ class RunStateMachine(StateTransitioner):
                 status_message += "(and downloading %d other dependencies and docker images)" % len(
                     status_messages
                 )
+            logger.info(f'{run_state.bundle.uuid} is changing state to {status_message}')
             return run_state._replace(run_status=status_message)
 
         # All dependencies ready! Set up directories, symlinks and container. Start container.
@@ -278,7 +285,13 @@ class RunStateMachine(StateTransitioner):
                         "your worker is mounted properly or contact your administrators."
                     )
                     logger.error(message)
+                    logger.info(
+                        f'{run_state.bundle.uuid} is cleaning up and changing state to {message}'
+                    )
                     return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
+                logger.info(
+                    f'{run_state.bundle.uuid} is waiting for bundle directory to be created by the server'
+                )
                 return run_state._replace(
                     run_status="Waiting for bundle directory to be created by the server",
                     bundle_dir_wait_num_tries=run_state.bundle_dir_wait_num_tries - 1,
@@ -333,6 +346,9 @@ class RunStateMachine(StateTransitioner):
                 try:
                     mount_dependency(dependency, self.shared_file_system)
                 except OSError as e:
+                    logger.info(
+                        f'{run_state.bundle.uuid} is cleaning up and changing state to {str(e)}'
+                    )
                     return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=str(e))
 
         if run_state.resources.network:
@@ -358,6 +374,7 @@ class RunStateMachine(StateTransitioner):
         except docker_utils.DockerUserErrorException as e:
             message = 'Cannot start Docker container: {}'.format(e)
             logger.warning(message)
+            logger.info(f'{run_state.bundle.uuid} is cleaning up and changing state to {message}')
             return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
         except Exception as e:
             message = 'Cannot start Docker container: {}'.format(e)
@@ -395,9 +412,11 @@ class RunStateMachine(StateTransitioner):
         2- If run is killed, kill the container
         3- If run is finished, move to CLEANING_UP state
         """
+        logger.info(f'{run_state.bundle.uuid} is running')
 
         def check_and_report_finished(run_state):
             try:
+                logger.info(f'{run_state.bundle.uuid} is checking finished')
                 finished, exitcode, failure_msg = docker_utils.check_finished(run_state.container)
             except docker_utils.DockerException:
                 logger.error(traceback.format_exc())
@@ -406,7 +425,15 @@ class RunStateMachine(StateTransitioner):
                 finished=finished, exitcode=exitcode, failure_message=failure_msg
             )
 
-        def check_resource_utilization(run_state):
+
+        def check_resource_utilization(run_state: RunState):
+            logger.info(f'{run_state.bundle.uuid} is checking resource utilization')
+            cpu_usage, memory_limit = docker_utils.get_container_stats_with_docker_stats(
+                run_state.container
+            )
+            run_state = run_state._replace(cpu_usage=cpu_usage)
+            run_state = run_state._replace(memory_limit=memory_limit)
+
             kill_messages = []
 
             run_stats = docker_utils.get_container_stats(run_state.container)
@@ -450,6 +477,7 @@ class RunStateMachine(StateTransitioner):
             return run_state
 
         def check_disk_utilization():
+            logger.info(f'{run_state.bundle.uuid} is checking disk utilization')
             running = True
             while running:
                 start_time = time.time()
@@ -506,6 +534,7 @@ class RunStateMachine(StateTransitioner):
             move to UPLOADING_RESULTS state
            Otherwise move to FINALIZING state
         """
+        logger.info(f'{run_state.bundle.uuid} is cleaning up')
 
         def remove_path_no_fail(path):
             try:
@@ -571,6 +600,7 @@ class RunStateMachine(StateTransitioner):
         If uploading and finished:
             Move to FINALIZING state
         """
+        logger.info(f'{run_state.bundle.uuid} is uploading results')
         if run_state.is_restaged:
             return run_state._replace(stage=RunStage.RESTAGED)
 
@@ -635,6 +665,7 @@ class RunStateMachine(StateTransitioner):
                 if run_state.failure_message
                 else run_state.kill_message
             )
+            logger.info(f'{run_state.bundle.uuid} is killed')
             run_state = run_state._replace(failure_message=failure_message)
         return run_state._replace(stage=RunStage.FINALIZING, run_status="Finalizing bundle")
 
@@ -644,11 +675,14 @@ class RunStateMachine(StateTransitioner):
         server, if bundle is going be sent back to the server, move on to the RESTAGED state. Otherwise,
         move on to the FINISHED state. Can also remove bundle_path now.
         """
+        logger.info(f'{run_state.bundle.uuid} is finalizing')
         if run_state.is_restaged:
+            logger.info(f'{run_state.bundle.uuid} is restaged')
             return run_state._replace(stage=RunStage.RESTAGED)
         elif run_state.finalized:
             if not self.shared_file_system:
                 remove_path(run_state.bundle_path)  # don't remove bundle if shared FS
+            logger.info(f'{run_state.bundle.uuid} is finished')
             return run_state._replace(stage=RunStage.FINISHED, run_status='Finished')
         else:
             return run_state
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -126,12 +126,17 @@ DependencyToMount = namedtuple('DependencyToMount', 'docker_path, child_path, pa
 
 
 def log_bundle_transition(
-    bundle_uuid: str, previous_state: str, next_state: str, reason: str, warn: bool = False
+    bundle_uuid: str,
+    previous_state: str,
+    next_state: str,
+    reason: str = '',
+    level: int = logging.INFO,
+    warn: bool = False,
 ):
     info = f'Bundle {bundle_uuid} is transitioning from {previous_state} to {next_state}'
     if reason != '':
         info = info + f' due to: {reason}'
-    logger.info(info)
+    logger.log(level=level, msg=info)
     if warn:
         logger.warning(traceback.format_exc())
 
@@ -216,7 +221,7 @@ class RunStateMachine(StateTransitioner):
             #   dependency_path:docker_dependency_path:ro
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
-        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.PREPARING, '')
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.PREPARING)
 
         if run_state.is_killed or run_state.is_restaged:
             log_bundle_transition(
@@ -282,8 +287,13 @@ class RunStateMachine(StateTransitioner):
         elif image_state.stage == DependencyStage.FAILED:
             # Failed to pull image; -> CLEANING_UP
             message = 'Failed to download Docker image: %s' % image_state.message
-            logger.error(
-                f'Bundle {run_state.bundle.uuid} is transitioning from {run_state.stage} to {RunStage.CLEANING_UP} due to {message}'
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.CLEANING_UP,
+                reason=message,
+                level=logging.ERROR,
+                warn=True,
             )
             return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
 
@@ -372,7 +382,8 @@ class RunStateMachine(StateTransitioner):
                         run_state.bundle.uuid,
                         run_state.stage,
                         RunStage.CLEANING_UP,
-                        f'since {str(e)}',
+                        reason=str(e),
+                        level=logging.ERROR,
                     )
                     return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=str(e))
 
@@ -398,9 +409,13 @@ class RunStateMachine(StateTransitioner):
             self.worker_docker_network.connect(container)
         except docker_utils.DockerUserErrorException as e:
             message = 'Cannot start Docker container: {}'.format(e)
-            logger.warning(message)
             log_bundle_transition(
-                run_state.bundle.uuid, run_state.stage, RunStage.CLEANING_UP, f'since the {message}'
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.CLEANING_UP,
+                reason=message,
+                level=logging.WARNING,
+                warn=True,
             )
             return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
         except Exception as e:
@@ -520,7 +535,12 @@ class RunStateMachine(StateTransitioner):
         run_state = check_resource_utilization(run_state)
 
         if run_state.is_killed or run_state.is_restaged:
-            log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.CLEANING_UP, '')
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.CLEANING_UP,
+                'the bundle is' 'killed' if run_state.is_killed else 'restaged',
+            )
             if docker_utils.container_exists(run_state.container):
                 try:
                     run_state.container.kill()
@@ -589,12 +609,12 @@ class RunStateMachine(StateTransitioner):
         self.paths_to_remove = []
 
         if run_state.is_restaged:
-            log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED, '')
+            log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED)
             return run_state._replace(stage=RunStage.RESTAGED)
 
         if not self.shared_file_system and run_state.has_contents:
             log_bundle_transition(
-                run_state.bundle.uuid, run_state.stage, RunStage.UPLOADING_RESULTS, ''
+                run_state.bundle.uuid, run_state.stage, RunStage.UPLOADING_RESULTS
             )
             return run_state._replace(
                 stage=RunStage.UPLOADING_RESULTS, run_status='Uploading results', container=None
@@ -622,9 +642,13 @@ class RunStateMachine(StateTransitioner):
         If uploading and finished:
             Move to FINALIZING state
         """
-        # logger.info(f'{run_state.bundle.uuid} is uploading results')
-        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED, '')
         if run_state.is_restaged:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.RESTAGED,
+                'the bundle has not started uploading yet',
+            )
             return run_state._replace(stage=RunStage.RESTAGED)
 
         def upload_results():
@@ -657,10 +681,24 @@ class RunStateMachine(StateTransitioner):
         )
 
         if self.uploading[run_state.bundle.uuid].is_alive():
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                run_state.stage,
+                'uploading has not finished yet',
+            )
             return run_state._replace(
                 run_status=self.uploading[run_state.bundle.uuid]['run_status']
             )
         elif not self.uploading[run_state.bundle.uuid]['success']:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.RESTAGED,
+                'uploading failed',
+                level=logging.ERROR,
+                warn=True,
+            )
             # upload failed
             failure_message = run_state.failure_message
             if failure_message:
@@ -681,7 +719,7 @@ class RunStateMachine(StateTransitioner):
         """
         Prepare the finalize message to be sent with the next checkin
         """
-        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.FINALIZING, '')
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.FINALIZING)
         if run_state.is_killed:
             # Append kill_message, which contains more useful info on why a run was killed, to the failure message.
             failure_message = (
@@ -689,7 +727,7 @@ class RunStateMachine(StateTransitioner):
                 if run_state.failure_message
                 else run_state.kill_message
             )
-            logger.error(f'{run_state.bundle.uuid} is killed, since {failure_message}')
+            logger.error(f'{run_state.bundle.uuid} is killed: {failure_message}')
             run_state = run_state._replace(failure_message=failure_message)
         return run_state._replace(stage=RunStage.FINALIZING, run_status="Finalizing bundle")
 
@@ -700,17 +738,22 @@ class RunStateMachine(StateTransitioner):
         move on to the FINISHED state. Can also remove bundle_path now.
         """
         if run_state.is_restaged:
-            logger.info(f'{run_state.bundle.uuid} is restaged')
             log_bundle_transition(
-                run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED, 'it has finalized'
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.RESTAGED,
+                'the bundle is going to be sent back to the server',
             )
             return run_state._replace(stage=RunStage.RESTAGED)
         elif run_state.finalized:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.FINISHED,
+                'a full worker cycle has passed',
+            )
             if not self.shared_file_system:
                 remove_path(run_state.bundle_path)  # don't remove bundle if shared FS
-            log_bundle_transition(
-                run_state.bundle.uuid, run_state.stage, RunStage.FINISHED, 'it has finalized'
-            )
             return run_state._replace(stage=RunStage.FINISHED, run_status='Finished')
         else:
             return run_state
--- a/codalab/worker/worker_run_state.py
+++ b/codalab/worker/worker_run_state.py
@@ -205,7 +205,7 @@ class RunStateMachine(StateTransitioner):
             #   dependency_path:docker_dependency_path:ro
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
-        logger.info(f'{run_state.bundle.uuid} is preparing')
+        logger.info(f'Bundle {run_state.bundle.uuid} is preparing')
 
         if run_state.is_killed or run_state.is_restaged:
             logger.info(f'{run_state.bundle.uuid} is cleaning up')
@@ -222,7 +222,6 @@ class RunStateMachine(StateTransitioner):
             )
             logger.error(message)
             logger.error(traceback.format_exc())
-            logger.info(f'{run_state.bundle.uuid} is changing state to {message}')
             return run_state._replace(run_status=message)
 
         dependencies_ready = True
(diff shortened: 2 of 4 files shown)


codalab/worker/worker_run_state.py
```diff
@@ -221,6 +224,7 @@ def mount_dependency(dependency, shared_file_system):
             )
             logger.error(message)
             logger.error(traceback.format_exc())
+            logger.info(f'{run_state.bundle.uuid} is changing state to {message}')
```
teetone: The message is already being logged in the previous lines.  I don't think it's necessary.

codalab/worker/worker_run_state.py
```diff
@@ -207,7 +207,10 @@ def mount_dependency(dependency, shared_file_system):
             #   dependency_path:docker_dependency_path:ro
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
+        logger.info(f'{run_state.bundle.uuid} is preparing')
```
teetone: For most of the log messages in this file, can you prepend "Bundle"?

codalab/worker/worker_run_state.py
```diff
@@ -258,6 +263,7 @@ def mount_dependency(dependency, shared_file_system):
             # Failed to pull image; -> CLEANING_UP
             message = 'Failed to download Docker image: %s' % image_state.message
             logger.error(message)
+            logger.info(f'{run_state.bundle.uuid} is cleaning up')
```
teetone: Maybe combine with the previous log line

codalab/worker/worker_run_state.py
```diff
@@ -280,7 +287,13 @@ def mount_dependency(dependency, shared_file_system):
                         "your worker is mounted properly or contact your administrators."
                     )
                     logger.error(message)
+                    logger.info(
+                        f'{run_state.bundle.uuid} is cleaning up and changing state to {message}'
```
teetone: `message` is not the same as bundle state.

codalab/worker/worker_run_state.py
```diff
@@ -335,6 +348,9 @@ def mount_dependency(dependency, shared_file_system):
                 try:
                     mount_dependency(dependency, self.shared_file_system)
                 except OSError as e:
+                    logger.info(
```
teetone: `logger.error`

codalab/worker/worker_run_state.py
```diff
@@ -360,6 +376,7 @@ def mount_dependency(dependency, shared_file_system):
         except docker_utils.DockerUserErrorException as e:
             message = 'Cannot start Docker container: {}'.format(e)
             logger.warning(message)
+            logger.info(f'{run_state.bundle.uuid} is cleaning up and changing state to {message}')
```
teetone: `logger.error`

teetone (changes requested):
There are repeat log messages, which makes it hard to distinguish why the bundle transitions to certain states (e.g. "... is cleaning up"). I think the log message should also give a brief reason with the event.

epicfaace (approved):

codalab/worker/worker_run_state.py
```diff
@@ -450,6 +469,7 @@ def check_resource_utilization(run_state):
             return run_state
 
         def check_disk_utilization():
+            logger.info(f'{run_state.bundle.uuid} is checking disk utilization')
```
teetone: This would result in a log statement like `0x111 is checking disk utilization`, which doesn't really make sense. Can you change it to "Checking disk utilization for bundle {run_state.bundle.uuid}"?

codalab/worker/worker_run_state.py
```diff
@@ -644,11 +667,14 @@ def _transition_from_FINALIZING(self, run_state):
         server, if bundle is going be sent back to the server, move on to the RESTAGED state. Otherwise,
         move on to the FINISHED state. Can also remove bundle_path now.
         """
+        logger.info(f'{run_state.bundle.uuid} is finalizing')
```
teetone: The bundle is already finalizing at this point.  This line should belong in `finalize_run`.

teetone (commented):
Thinking more about this, it would be cleaner if you created a util function `log_bundle_transition` that logs the UUID of the bundle, previous state, next state and reason for transition. The function signature might look something like this:

```
def log_bundle_transition(bundle_uuid, previous_state, next_state, reason, log_level): ...
```

This would make debugging easier and logging in this file more uniform. Other log lines can just be logged with `logger` directly.

codalab/worker/worker_run_state.py
teetone: Can you also update the other log lines that have the form `{uuid} is checking {something}`?

codalab/worker/worker_run_state.py
```diff
@@ -205,7 +215,16 @@ def mount_dependency(dependency, shared_file_system):
             #   dependency_path:docker_dependency_path:ro
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.PREPARING, '')
```
teetone: This will result in  "...from <state> to PREPARING due to". We should modify the util function to check if reason exists or not.

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,16 @@ class RunStage(object):
 DependencyToMount = namedtuple('DependencyToMount', 'docker_path, child_path, parent_path')
 
 
+def log_bundle_transition(
+    bundle_uuid: str, previous_state: str, next_state: str, reason: str, warn: bool = False
```
teetone: There should be a type for states that's not `str`.

codalab/worker/worker_run_state.py
```diff
@@ -205,7 +215,16 @@ def mount_dependency(dependency, shared_file_system):
...
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.PREPARING, '')
+
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.CLEANING_UP,
+                'the bundle is either killed or restaged.',
```
teetone: I think we can capitalize the sentence and put a colon before stating the reason.

codalab/worker/worker_run_state.py
```diff
@@ -255,7 +281,9 @@ def mount_dependency(dependency, shared_file_system):
         elif image_state.stage == DependencyStage.FAILED:
             # Failed to pull image; -> CLEANING_UP
             message = 'Failed to download Docker image: %s' % image_state.message
-            logger.error(message)
+            logger.error(
```
teetone: The new util function can replace this.

codalab/worker/worker_run_state.py
```diff
@@ -238,6 +257,13 @@ def mount_dependency(dependency, shared_file_system):
                     dependencies_ready = False
                 elif dependency_state.stage == DependencyStage.FAILED:
                     # Failed to download dependency; -> CLEANING_UP
+                    log_bundle_transition(
+                        run_state.bundle.uuid,
+                        run_state.stage,
+                        RunStage.CLEANING_UP,
+                        f'dependency key {dep_key} has failed for this bundle',
```
teetone: Should say `The dependency {dep} has failed for this bundle`.

codalab/worker/worker_run_state.py
```diff
@@ -333,6 +367,12 @@ def mount_dependency(dependency, shared_file_system):
                 try:
                     mount_dependency(dependency, self.shared_file_system)
                 except OSError as e:
+                    log_bundle_transition(
+                        run_state.bundle.uuid,
+                        run_state.stage,
+                        RunStage.CLEANING_UP,
+                        f'since {str(e)}',
```
teetone: This should be logged as an error not info or warning.

codalab/worker/worker_run_state.py
```diff
@@ -358,6 +398,9 @@ def mount_dependency(dependency, shared_file_system):
         except docker_utils.DockerUserErrorException as e:
             message = 'Cannot start Docker container: {}'.format(e)
             logger.warning(message)
```
teetone: We can get rid of this line and log it together using the util function.

codalab/worker/worker_run_state.py
```diff
@@ -542,9 +588,13 @@ def remove_path_no_fail(path):
         self.paths_to_remove = []
 
         if run_state.is_restaged:
+            log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED, '')
             return run_state._replace(stage=RunStage.RESTAGED)
 
         if not self.shared_file_system and run_state.has_contents:
+            log_bundle_transition(
+                run_state.bundle.uuid, run_state.stage, RunStage.UPLOADING_RESULTS, ''
```
teetone: Can we add a default argument for an empty or None reason?

codalab/worker/worker_run_state.py
```diff
@@ -571,6 +621,8 @@ def _transition_from_UPLOADING_RESULTS(self, run_state):
         If uploading and finished:
             Move to FINALIZING state
         """
+        # logger.info(f'{run_state.bundle.uuid} is uploading results')
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED, '')
```
teetone: Shouldn't this line be inside the if statement?

codalab/worker/worker_run_state.py
```diff
@@ -571,6 +621,8 @@ def _transition_from_UPLOADING_RESULTS(self, run_state):
         If uploading and finished:
             Move to FINALIZING state
         """
+        # logger.info(f'{run_state.bundle.uuid} is uploading results')
```
teetone: Clean up commented out code.

codalab/worker/worker_run_state.py
```diff
@@ -628,13 +680,15 @@ def finalize_run(self, run_state):
...
         """
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.FINALIZING, '')
         if run_state.is_killed:
             # Append kill_message, which contains more useful info on why a run was killed, to the failure message.
             failure_message = (
                 "{}. {}".format(run_state.failure_message, run_state.kill_message)
                 if run_state.failure_message
                 else run_state.kill_message
             )
+            logger.error(f'{run_state.bundle.uuid} is killed, since {failure_message}')
```
teetone: Get rid of "since". It's not needed if we have a colon before the reason.

codalab/worker/worker_run_state.py
```diff
@@ -645,10 +699,17 @@ def _transition_from_FINALIZING(self, run_state):
         move on to the FINISHED state. Can also remove bundle_path now.
         """
         if run_state.is_restaged:
+            logger.info(f'{run_state.bundle.uuid} is restaged')
```
teetone: This line seems redundant if we are going to log the same thing the next line.

codalab/worker/worker_run_state.py
```diff
@@ -645,10 +699,17 @@ def _transition_from_FINALIZING(self, run_state):
...
+            log_bundle_transition(
+                run_state.bundle.uuid, run_state.stage, RunStage.RESTAGED, 'it has finalized'
```
teetone: This doesn't make sense to me because the log will state "...to RESTAGED due to it has finalized".

codalab/worker/worker_run_state.py
```diff
@@ -645,10 +699,17 @@ def _transition_from_FINALIZING(self, run_state):
...
+            )
             return run_state._replace(stage=RunStage.RESTAGED)
         elif run_state.finalized:
             if not self.shared_file_system:
                 remove_path(run_state.bundle_path)  # don't remove bundle if shared FS
+            log_bundle_transition(
+                run_state.bundle.uuid, run_state.stage, RunStage.FINISHED, 'it has finalized'
```
teetone: Stating that the bundle transitioned to FINISHED is enough. I don't think we need the reason here.

codalab/worker/worker_run_state.py
```diff
@@ -474,6 +519,7 @@ def check_disk_utilization():
         run_state = check_resource_utilization(run_state)
 
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.CLEANING_UP, '')
```
teetone: We should include if the bundle was killed or restaged in the reason.

teetone (changes requested):

codalab/worker/worker_run_state.py
yiboliu: Like check if the reason exists, then we append the reason; otherwise just truncate the last two words?

codalab/worker/worker_run_state.py
yiboliu: I think we are actually logging transitions of stages in this function. It's only updating `stage` of the run_bundle. Stages like `RunStage.CLEANING_UP` are defined as strs right?

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,22 @@ class RunStage(object):
 DependencyToMount = namedtuple('DependencyToMount', 'docker_path, child_path, parent_path')
 
 
+def log_bundle_transition(
+    bundle_uuid: str,
+    previous_state: str,
+    next_state: str,
+    reason: str = '',
+    level: int = logging.INFO,
+    warn: bool = False,
```
teetone: Why do we still need `warn` if we now have `level`? We should show the stacktrace if it's an error.

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,22 @@ class RunStage(object):
...
+):
+    info = f'Bundle {bundle_uuid} is transitioning from {previous_state} to {next_state}'
+    if reason != '':
+        info = info + f' due to: {reason}'
```
teetone: `info = f'{info} due to: {reason}'`

codalab/worker/worker_run_state.py
```diff
@@ -205,7 +221,16 @@ def mount_dependency(dependency, shared_file_system):
...
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
+        log_bundle_transition(run_state.bundle.uuid, run_state.stage, RunStage.PREPARING)
+
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.CLEANING_UP,
+                'The bundle is either killed or restaged.',
```
teetone: Should be more specific here.

codalab/worker/worker_run_state.py
```diff
@@ -333,6 +378,13 @@ def mount_dependency(dependency, shared_file_system):
                 try:
                     mount_dependency(dependency, self.shared_file_system)
                 except OSError as e:
+                    log_bundle_transition(
+                        run_state.bundle.uuid,
```
teetone: I think it would be easier to read if we specify the parameters here like you did with `reason`.

codalab/worker/worker_run_state.py
```diff
@@ -572,6 +643,12 @@ def _transition_from_UPLOADING_RESULTS(self, run_state):
             Move to FINALIZING state
         """
         if run_state.is_restaged:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.RESTAGED,
+                'the bundle has not started uploading yet',
```
teetone: I'm not sure what this means.

codalab/worker/worker_run_state.py
```diff
@@ -645,8 +738,20 @@ def _transition_from_FINALIZING(self, run_state):
...
+                RunStage.RESTAGED,
+                'the bundle is going to be sent back to the server',
+            )
             return run_state._replace(stage=RunStage.RESTAGED)
         elif run_state.finalized:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.FINISHED,
+                'a full worker cycle has passed',
```
teetone: Do we need a reason here?

codalab/worker/worker_run_state.py
```diff
@@ -645,8 +738,20 @@ def _transition_from_FINALIZING(self, run_state):
         move on to the FINISHED state. Can also remove bundle_path now.
         """
         if run_state.is_restaged:
+            log_bundle_transition(
+                run_state.bundle.uuid,
+                run_state.stage,
+                RunStage.RESTAGED,
+                'the bundle is going to be sent back to the server',
```
teetone: For what case does restaged occur? I think that should be explained.

yiboliu: comes from here - https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L644

codalab/worker/worker_run_state.py
yiboliu: a brief explanation from comments - https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L643

codalab/worker/worker_run_state.py
yiboliu: it should be this case - https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker_run_state.py#L565

codalab/worker/worker_run_state.py
yiboliu: Is it helpful / necessary to show the stacktrace if it's a warning?

codalab/worker/worker_run_state.py
teetone: Can you include it in the logs?

codalab/worker/worker_run_state.py
teetone: I'm not sure why the boolean is needed when `level` has the same information.

codalab/worker/worker_run_state.py
teetone: Can you incorporate that into the log message? It's still unclear to me.

codalab/worker/worker_run_state.py
teetone: I think the reason why the bundle is being sent back should be included in the log message.

codalab/worker/worker_run_state.py
teetone: Usually stacktraces are shown with errors and not warnings.

yiboliu: oh I thought if the level is warning, probably stacktrace is also useful so I changed the code accordingly. You can check the most up to date code.

codalab/worker/worker_run_state.py
yiboliu: I don't know why exactly the bundle is being sent back as it's commented by Jing.

codalab/worker/worker_run_state.py
yiboliu: I think including the link to the log message is a bit awkward. Why don't we just ask the developer to check the comments in the code?

codalab/worker/worker_run_state.py
yiboliu: I know why exactly is it sent back to the server as it's commented by Jing. It comes from this PR - https://github.com/codalab/codalab-worksheets/pull/2451 where I saw your approval.

codalab/worker/main.py
```diff
@@ -209,9 +209,13 @@ def main():
 
     # Configure logging
     logging.basicConfig(
-        format='%(asctime)s %(message)s', level=(logging.DEBUG if args.verbose else logging.INFO)
+        format='%(asctime)s %(message)s %(pathname)s %(lineno)d',
+        level=(logging.DEBUG if args.verbose else logging.INFO),
     )
 
+    if not args.verbose:
+        logging.getLogger('urllib3').setLevel(logging.INFO)
```
epicfaace: Can you make a separate PR that just does this change, and silences the urllib3 logs even when the worker is set to verbose?

yiboliu: https://github.com/codalab/codalab-worksheets/pull/3280

codalab/worker/worker_run_state.py
teetone: Can you include that in the reason? The `reason` parameter should explain why the bundle is being restaged.

codalab/worker/worker_run_state.py
teetone: I just don't understand `the bundle has not started uploading yet`. Can we include a better, clearer reason?

codalab/worker/worker_run_state.py
```diff
@@ -205,7 +220,19 @@ def mount_dependency(dependency, shared_file_system):
...
+            previous_state=run_state.stage,
+            next_state=RunStage.PREPARING,
+        )
+
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_state=run_state.stage,
+                next_state=RunStage.CLEANING_UP,
+                reason='the bundle is ' + 'killed' if run_state.is_killed else 'restaged',
```
teetone: Use f string for consistency.

codalab/worker/worker_run_state.py
```diff
@@ -474,6 +534,12 @@ def check_disk_utilization():
         run_state = check_resource_utilization(run_state)
 
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_state=run_state.stage,
+                next_state=RunStage.CLEANING_UP,
+                reason='the bundle is ' + 'killed' if run_state.is_killed else 'restaged',
```
teetone: Same here.

codalab/worker/worker_run_state.py
yiboliu: This explanation is incorrect. It's just checking if `run_state.is_restaged` is specified; if yes, then restage it. It comes from [here](https://github.com/codalab/codalab-worksheets/pull/2186#discussion_r423395813).

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,24 @@ class RunStage(object):
...
+    previous_state: str,
+    next_state: str,
+    reason: str = '',
+    level: int = logging.INFO,
+):
+    info = f'Bundle {bundle_uuid} is transitioning from {previous_state} to {next_state}'
+    if reason != '':
+        info = f'{info} due to: {reason}'
+    logger.log(level=level, msg=info)
+    if level >= logging.WARNING:
```
teetone: Can we only do this for `SEVERE` errors?

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,24 @@ class RunStage(object):
...
+        logger.warning(traceback.format_exc())
```
teetone: Should be `logger.error`

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,24 @@ class RunStage(object):
...
+
+
+restage_common_reason = 'The bundle is not in terminal states when the worker checks termination'
```
teetone: If this is used only in `RunStateMachine`, can you move it inside the class? Also, constants should be capitalized.

teetone: Should specify the terminal states in the message.

codalab/worker/worker_run_state.py
```diff
@@ -125,6 +125,24 @@ class RunStage(object):
 DependencyToMount = namedtuple('DependencyToMount', 'docker_path, child_path, parent_path')
 
 
+def log_bundle_transition(
+    bundle_uuid: str,
+    previous_state: str,
```
teetone: I think I pointed out this before in an earlier review, but the type should not be `str` for the states.

codalab/worker/worker_run_state.py
```diff
@@ -205,7 +223,19 @@ def mount_dependency(dependency, shared_file_system):
...
+            previous_state=run_state.stage,
+            next_state=RunStage.PREPARING,
+        )
+
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_state=run_state.stage,
+                next_state=RunStage.CLEANING_UP,
+                reason=f'the bundle is {"killed" if run_state.is_killed else "restaged"}',
```
teetone: was instead of is

codalab/worker/worker_run_state.py
```diff
@@ -255,7 +291,13 @@ def mount_dependency(dependency, shared_file_system):
         elif image_state.stage == DependencyStage.FAILED:
             # Failed to pull image; -> CLEANING_UP
             message = 'Failed to download Docker image: %s' % image_state.message
-            logger.error(message)
+            log_bundle_transition(
```
teetone: Will this show a stacktrace? Maybe we should have a `show_stacktrace` parameter.

codalab/worker/worker_run_state.py
```diff
@@ -205,7 +223,19 @@ def mount_dependency(dependency, shared_file_system):
             #   dependency_path:docker_dependency_path:ro
             docker_dependencies.append((dependency.parent_path, dependency.docker_path))
 
+        log_bundle_transition(
```
teetone: The function is called `_transition_from_PREPARING`. Why is `next_state` preparing?

codalab/worker/worker_run_state.py
```diff
@@ -277,8 +320,13 @@ def mount_dependency(dependency, shared_file_system):
                         "Please ensure the shared fileystem between the server and "
                         "your worker is mounted properly or contact your administrators."
                     )
-                    logger.error(message)
+                    logger.error(
```
teetone: Use the util function here

codalab/worker/worker_run_state.py
```diff
@@ -265,6 +307,7 @@ def mount_dependency(dependency, shared_file_system):
                 status_message += "(and downloading %d other dependencies and docker images)" % len(
                     status_messages
                 )
+            logger.warning(f'{run_state.bundle.uuid} is not ready yet since {status_message}')
```
teetone: Why is this a warning? Also, is this necessary to log? We don't want to overcrowd the logs.

codalab/worker/worker_run_state.py
```diff
@@ -277,8 +320,13 @@ def mount_dependency(dependency, shared_file_system):
                         "Please ensure the shared fileystem between the server and "
                         "your worker is mounted properly or contact your administrators."
                     )
-                    logger.error(message)
+                    logger.error(
+                        f'Bundle {run_state.bundle.uuid} has trouble with bundle path: {message}'
+                    )
                     return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
+                logger.info(
+                    f'{run_state.bundle.uuid} is waiting for bundle directory to be created by the server'
```
teetone: Can we include the current `bundle_dir_wait_num_tries` in this message too?

codalab/worker/worker_run_state.py
```diff
@@ -333,6 +381,13 @@ def mount_dependency(dependency, shared_file_system):
                 try:
                     mount_dependency(dependency, self.shared_file_system)
                 except OSError as e:
+                    log_bundle_transition(
+                        bundle_uuid=run_state.bundle.uuid,
+                        previous_state=run_state.stage,
+                        next_state=RunStage.CLEANING_UP,
+                        reason=str(e),
```
teetone: Isn't the stacktrace already printed?

codalab/worker/worker_run_state.py
```diff
@@ -628,13 +727,19 @@ def finalize_run(self, run_state):
...
+            next_state=RunStage.FINALIZING,
+        )
         if run_state.is_killed:
             # Append kill_message, which contains more useful info on why a run was killed, to the failure message.
             failure_message = (
                 "{}. {}".format(run_state.failure_message, run_state.kill_message)
                 if run_state.failure_message
                 else run_state.kill_message
             )
+            logger.error(f'{run_state.bundle.uuid} is killed: {failure_message}')
```
teetone: Killing a bundle is not necessarily an error.

codalab/worker/worker_run_state.py
```diff
@@ -645,8 +750,20 @@ def _transition_from_FINALIZING(self, run_state):
...
+                next_state=RunStage.RESTAGED,
+                reason='the bundle is restaged, as `pass-down-termination` is specified for worker',
+            )
             return run_state._replace(stage=RunStage.RESTAGED)
         elif run_state.finalized:
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_state=run_state.stage,
+                next_state=RunStage.FINISHED,
+                reason='a full worker cycle has passed and `pass-down-termination` is not specified for worker',
```
teetone: I don't think a reason needs to be specified here.

codalab/worker/worker_run_state.py
```diff
@@ -604,10 +690,23 @@ def progress_callback(bytes_uploaded):
...
+                reason='uploading has not finished yet',
+            )
             return run_state._replace(
                 run_status=self.uploading[run_state.bundle.uuid]['run_status']
             )
         elif not self.uploading[run_state.bundle.uuid]['success']:
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_state=run_state.stage,
+                next_state=RunStage.RESTAGED,
```
teetone: Why is this restaged?

codalab/worker/worker_run_state.py
```diff
@@ -604,10 +690,23 @@ def progress_callback(bytes_uploaded):
         )
 
         if self.uploading[run_state.bundle.uuid].is_alive():
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_state=run_state.stage,
```
teetone: I don't think showing the same stage twice is helpful.

codalab/worker/worker_run_state.py
```diff
@@ -357,7 +412,13 @@ def mount_dependency(dependency, shared_file_system):
             self.worker_docker_network.connect(container)
         except docker_utils.DockerUserErrorException as e:
             message = 'Cannot start Docker container: {}'.format(e)
-            logger.warning(message)
+            log_bundle_transition(
```
teetone, line 407: This should actually be an error and not just a warning.

codalab/worker/worker_run_state.py
```diff
@@ -238,6 +268,12 @@ def mount_dependency(dependency, shared_file_system):
                     dependencies_ready = False
                 elif dependency_state.stage == DependencyStage.FAILED:
                     # Failed to download dependency; -> CLEANING_UP
+                    log_bundle_transition(
+                        bundle_uuid=run_state.bundle.uuid,
+                        previous_state=run_state.stage,
+                        next_state=RunStage.CLEANING_UP,
+                        reason=f'Dependency {dep} has failed for this bundle',
```
teetone: Will this output the entire dependency object?

teetone (changes requested):

codalab/worker/worker_run_state.py
yiboliu: by `SEVERE` do you mean critical error from [this chart](https://docs.python.org/3/library/logging.html#logging-levels) or just error?

yiboliu: `SEVERE` means error.

codalab/worker/worker_run_state.py
yiboliu: Will include an example of what `traceback.format_exc()` look like. Show the stacktrace in a string, not an array.

codalab/worker/worker_run_state.py
yiboliu: list 3 terminal states.

codalab/worker/worker_run_state.py
yiboliu: change `is` to `was`

codalab/worker/worker_run_state.py
yiboliu: keep the original code

codalab/worker/worker_run_state.py
yiboliu: delete this code

codalab/worker/worker_run_state.py
yiboliu: parameter reason would be the first line.

codalab/worker/worker_run_state.py
yiboliu: change to info

codalab/worker/worker_run_state.py
yiboliu: example of `traceback.format_exc()`: 
```
  File "error.py", line 4, in <module>
    a = 3 / 0
ZeroDivisionError: division by zero
```

codalab/worker/worker_run_state.py
```diff
@@ -474,6 +529,12 @@ def check_disk_utilization():
         run_state = check_resource_utilization(run_state)
 
         if run_state.is_killed or run_state.is_restaged:
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_stage=run_state.stage,
+                next_stage=RunStage.CLEANING_UP,
+                reason=f'the bundle is {"killed" if run_state.is_killed else "restaged"}',
```
teetone: Should be `was` instead of `is`.

codalab/worker/worker_run_state.py
```diff
@@ -137,6 +152,7 @@ class RunStateMachine(StateTransitioner):
 
     _ROOT = '/'
     _CURRENT_DIRECTORY = '.'
+    RESTAGE_COMMON_REASON = 'The bundle is not in terminal states {READY, FAILED, KILLED} when the worker checks termination'
```
teetone: Can we rename this to just `RESTAGED_REASON`?

codalab/worker/worker_run_state.py
```diff
@@ -357,7 +404,13 @@ def mount_dependency(dependency, shared_file_system):
             self.worker_docker_network.connect(container)
         except docker_utils.DockerUserErrorException as e:
             message = 'Cannot start Docker container: {}'.format(e)
-            logger.warning(message)
+            log_bundle_transition(
+                bundle_uuid=run_state.bundle.uuid,
+                previous_stage=run_state.stage,
+                next_stage=RunStage.CLEANING_UP,
+                reason=message,
```
teetone: Since the level is `ERROR` it will show the traceback twice: once in message and another time when calling `traceback.format_exc()`. Can we just put `Cannot start Docker Container.` as the reason and keep `message` the same?

teetone (approved):
Looks good! Just a few more comments.

codalab/worker/worker_run_state.py
```diff
@@ -137,6 +152,7 @@ class RunStateMachine(StateTransitioner):
 
     _ROOT = '/'
     _CURRENT_DIRECTORY = '.'
+    RESTAGEG_REASON = 'The bundle is not in terminal states {READY, FAILED, KILLED} when the worker checks termination'
```
```suggestion
    RESTAGED_REASON = 'The bundle is not in terminal states {READY, FAILED, KILLED} when the worker checks termination'
```

PR #3241, opened 2021-02-03
FIx the bug of `discard` button not closing the source editor

yuqijin:

 "Discard" button cannot close the source editor.

### Screenshots

ezgif com-gif-maker

### Checklist


epicfaace (commented):
Can you briefly summarize what exactly the issue with the previous code was (was it a race condition)?

yuqijin:
> Can you briefly summarize what exactly the issue with the previous code was (was it a race condition)?

Sure. So previously, the function for closing the source editor is a callback function for the SAVE button, so that's why clicking on the DISCARD button will not close the editor. Not a race condition but some logic mistake though.

epicfaace (approved):

PR #3244, opened 2021-02-05
Azure Blob part 1: Add MockBlobStorageFileSystem

epicfaace:

Part 1 of #3198.

Adds the testing utility MockBlobStorageFileSystem, which is a mock Azure Blob Storage filesystem that is backed by a local filesystem instead.

This means that while URLs will still start with azfs://, adding / removing files
will actually change files in the local /tmp/codalab/azfs-mock/ directory instead.

This is used only for unit tests for speed, so that unit tests do not need to depend on
a Blob Storage container / Azurite running in the background.

### DIFF
--- a/codalab/lib/beam/filesystems.py
+++ b/codalab/lib/beam/filesystems.py
@@ -2,9 +2,9 @@ import os
 from azure.storage.blob import BlobServiceClient
 
 # Test connection string for Azurite (local development)
-TEST_CONN_STR = "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;"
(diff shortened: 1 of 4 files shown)


yiboliu (commented):
Is the test still being worked on or finished? Looks like there isn't any test ready yet.

epicfaace:
@yiboliu the actual tests are in #3198, which I split. I couldn't add the tests without adding most of the entire PR. This PR contains only testing _utilities_ that are used there.

yiboliu:
> @yiboliu the actual tests are in #3198, which I split. I couldn't add the tests without adding most of the entire PR. This PR contains only testing _utilities_ that are used there.

Ok, that makes sense.

yiboliu (approved):

PR #3250, opened 2021-02-09
Table cell render as 'Forbidden' instead of 'Loading...' when the user has no access

### LINKED ISSUE #3239: schema tables are rendering as 'Loading...'
The columns "Exact" and "F1" are just showing "Loading..." forever:
https://worksheets.codalab.org/worksheets/0x816632b5306d41d3ba80ccecc84f1b6e
Javascript console shows a bunch of Forbidden (403) links.  Maybe I'm not supposed to be able to view these bundles?  But then I do have access...  User `pliang`, which is in the `squad` group, which these bundles have given `all` access to.

yuqijin:

Based on Percy's suggestion, use 'Forbidden' instead to avoid confusion when permission error occurs since 'Loading...' is misleading

### Related issues

fixes #3239 

### Screenshots

Screen Shot 2021-02-09 at 12 09 55 AM

### Checklist

### DIFF
--- a/codalab/rest/interpret.py
+++ b/codalab/rest/interpret.py
@@ -16,7 +16,7 @@ import json
 import yaml
 from bottle import get, post, local, request, abort, httplib
 
-from codalab.common import UsageError, NotFoundError
+from codalab.common import UsageError, NotFoundError, PermissionError
 from codalab.lib import formatting, spec_util
 from codalab.lib.worksheet_util import (
     TYPE_DIRECTIVE,
@@ -680,6 +680,10 @@ def interpret_file_genpath(target_cache, bundle_uuid, genpath, post):
                             info = ''.join(contents)
         except NotFoundError:
             pass
+        except PermissionError:
+            # Use the array of length 3 to pass the PermissionError to the frontend
+            info = ["Forbidden", "Forbidden", "Forbidden"]
+            return info
 
         # Try to interpret the structure of the file by looking inside it.
         target_cache[target] = info
--- a/codalab/rest/interpret.py
+++ b/codalab/rest/interpret.py
@@ -682,7 +682,7 @@ def interpret_file_genpath(target_cache, bundle_uuid, genpath, post):
             pass
         except PermissionError:
             # Use the array of length 3 to pass the PermissionError to the frontend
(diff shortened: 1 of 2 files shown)


codalab/rest/interpret.py
```diff
@@ -680,6 +680,10 @@ def interpret_file_genpath(target_cache, bundle_uuid, genpath, post):
                             info = ''.join(contents)
         except NotFoundError:
             pass
+        except PermissionError:
+            # Use the array of length 3 to pass the PermissionError to the frontend
+            info = ["Forbidden", "Forbidden", "Forbidden"]
```
epicfaace: Would it be easy to update the worksheet interpret unit tests to cover this PermissionError? If it's not too difficult, let's try to add a test case.

frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -165,8 +165,13 @@ class BundleRow extends Component {
             }
             if (Array.isArray(rowContent) && rowContent.length === 3) {
                 // Cell is a bundle genpath triple -- see is_bundle_genpath_triple() in backend.
-                // This means that the cell is only briefly loaded.
-                rowContent = <span style={{ color: 'grey' }}>Loading...</span>;
+                if (rowContent[0] === 'Forbidden') {
```
epicfaace: Rather than returning `["Forbidden", "Forbidden", "Forbidden"]`, would it make more sense to return a single `["Forbidden"]` and then add a `if (rowContent.length === 1)` condition for this? (because "Forbidden" is not really a genpath triple)

yuqijin: updated.

epicfaace (approved):

codalab/rest/interpret.py
```diff
@@ -682,7 +682,7 @@ def interpret_file_genpath(target_cache, bundle_uuid, genpath, post):
             pass
         except PermissionError:
             # Use the array of length 3 to pass the PermissionError to the frontend
```
epicfaace:
```suggestion
            # Use an array of length 1 to pass the PermissionError to the frontend
```

epicfaace (approved):

Issue #3245, opened 2021-02-06
Make request_queue tags more flexible


nelson-liu:
It'd be nice if there was a way for request_queue to target more complex specifications of tags. For instance, say I have 4 workers with tags `a`, `b`, `c`, and `d`.

It'd be useful to run bundles with `request_queue` of:
- `NOT a` (i.e., run on any worker that isn't tagged with `a`)
- `a OR b` (i.e., run on a worker that is either tagged as `a` or `b`)

Not really sure how to best encode these logical operators into the `request_queue` field, but seems like a solvable problem.

Some more motivation:

Sometimes, bundles require specific worker properties that aren't reflected by `request_cpus` or `request_gpus`. For example (an actual example that I'm running into), suppose you're running a bundle that requires a GPU. However, not any GPU works---you need a GPU that isn't old (e.g., not a NVIDIA Tesla K40). So, if you assign a tag to each worker dependent on its gpu type (e.g., some have tag `k40`, some have tag `titanx`, some have tag `titanrtx`), you can set your `request_queue` to `!k40`.

Another example: I've been running bundles that use FP16 / automatic mixed precision. GPU support for this isn't good across the board---only the newer GPUs (e.g., titan RTX, V100, or GTX 3090) have tensor cores that make them run efficiently. In this case, any bundle that uses FP16 would have a `request_queue` of `(titanrtx | v100 | gtx3090)` to select the desired subset of bundles.

percyliang:
Do we want to proceed with this or close it?

PR #3252, opened 2021-02-10
Force worker tags to be alphanumeric

nelson-liu:

First part of #3245 , to make parsing out operations easier / possibly safer as well.


teetone (approved):

epicfaace:
@nelson-liu can you fix tests?

PR #3255, opened 2021-02-11
Undo referencing to unassigned container

yiboliu:
here - https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/docker_utils.py#L206: the container is defined outside the except block. In case that an APIError is thrown, it can't find the definition of container. 

The reason of this approach comes from the `remove` api of the `Container`:
```    
def remove(self, **kwargs):
        """
        Remove this container. Similar to the ``docker rm`` command.

        Args:
            v (bool): Remove the volumes associated with the container
            link (bool): Remove the specified link and not the underlying
                container
            force (bool): Force the removal of a running container (uses
                ``SIGKILL``)

        Raises:
            :py:class:`docker.errors.APIError`
                If the server returns an error.
        """
        return self.client.api.remove_container(self.id, **kwargs)
```

### Checklist

### DIFF
--- a/codalab/worker/docker_utils.py
+++ b/codalab/worker/docker_utils.py
@@ -203,7 +203,7 @@ def start_bundle_container(
         # because a container with the same name already exists. So, we try to remove
         # the container here.
         try:
-            container.remove(force=True)
+            client.api.remove_container(container_name, force=True)
         except Exception:
             logger.warning("Failed to clean up Docker container after failed launch.")
             traceback.print_exc()


epicfaace (commented):
What's the point of this change? `container.remove(force=True)` is already wrapped in a try-except block, so it won't throw an error even if `container` is null.

nelson-liu:
i think the issue is that, even if container is null, we'd like to remove containers that've failed to start (since the next time we try to launch it, we'll get an uninformative name overlap error).

yiboliu:
> What's the point of this change? `container.remove(force=True)` is already wrapped in a try-except block, so it won't throw an error even if `container` is null.

Also, in except block, it's printing extra stacktrace of the can't find reference of container. This was meant to reduce the amount of stracktraces.

nelson-liu (approved):

PR #3256, opened 2021-02-12
DO NOT MERGE

adiprerepa:
DO NOT MERGE

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -2013,6 +2013,38 @@ class BundleCLI(object):
                 print(uuid, file=self.stdout)
 
     @Commands.command(
+        'ancestors',
+        help='Find ancestors of a bundle.',
+        arguments=(
+                Commands.Argument(
+                    'bundle_spec', help=BUNDLE_SPEC_FORMAT, nargs='+', completer=BundlesCompleter
+                ),
+                Commands.Argument(
+                    '-w',
+                    '--worksheet-spec',
+                    help='Operate on this worksheet (%s).' % WORKSHEET_SPEC_FORMAT,
+                    completer=WorksheetsCompleter,
+                ),
+        ),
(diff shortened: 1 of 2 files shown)


epicfaace:
Thanks! We'll take a look.

adiprerepa:
could you reopen? wanted to add one more commit

epicfaace:
Feel free to add more commits! You can add commits to a closed PR

adiprerepa:
All right, added a commit.

codalab/lib/bundle_cli.py
```diff
@@ -2012,6 +2013,44 @@ def do_rm_command(self, args):
...
+        arguments=(
+            Commands.Argument(
+                'bundle_spec', help=BUNDLE_SPEC_FORMAT, nargs='+', completer=BundlesCompleter
+            ),
+            Commands.Argument(
+                '-w',
+                '--worksheet-spec',
+                help='Operate on this worksheet (%s).' % WORKSHEET_SPEC_FORMAT,
+                completer=WorksheetsCompleter,
+            ),
```
epicfaace, line 2028: What is `indent_count`? Maybe you should add it to `arguments` so that it's documented.

adiprerepa: It's supposed to be internal - it lets the function know how many indents to make based on how many parents there are, just like in the example. If bundle z depends on x and y, and x depends on foo, it would look like:
```
- z
  - x
    - foo
  - y
```
Foo is 2 layers in, so there would be two indents.

Do we want the user to be able to set how many indents out the output starts?

adiprerepa: Actually, should probably prefix with xxx to convey it's internal

epicfaace: Ah, I see, thanks!

PR #3257, opened 2021-02-12
Store tracebacks in bundle metadata and send them to sentry

### LINKED ISSUE #3227: Send tracebacks to sentry when bundles fail
Normally, a bundle fails with the error that's pretty cryptic, such as, "Dependency download failed: not a gzip file."

I modified the code to print a traceback using `traceback.format_exc()` for ease of development, and this helps a lot since the error looks like this:

image

It might be useful to, in general, always store the actual traceback of each exception. It doesn't have to show up in `failure_message`, but perhaps it can be stored in another metadata field such as `error_traceback`, which regular users wouldn't see / interact with. This would help the codalab development team to more quickly figure out why things are failing.

yiboliu:
Send the stacktrace of errors happening at bundle failures to sentry. 

### Related issues
fixes #3227 

### Screenshots

Sample traceback output if a bundle fails during upload:
image
image

### Checklist


codalab/worker/dependency_manager.py
```diff
@@ -486,4 +486,7 @@ def interruptable_read(*args, **kwargs):
         else:
             with self._paths_lock:
                 self._paths.remove(dependency_state.path)
+            from codalab.lib import telemetry_util
```
epicfaace: I think it should be OK to move this dependency out, to the top of the file. I don't think there's much overhead in importing it.

codalab/worker/docker_image_manager.py
```diff
@@ -282,6 +283,7 @@ def download():
                             image_spec, ex
                         )
                         logger.error(failure_msg)
+                        capture_message(failure_msg)
```
epicfaace: Do we need this line? We're already logging it via `logger.error` -- is that automatically sent to Sentry?

yiboliu: Oh that's right. I read [this](https://docs.sentry.io/platforms/python/guides/logging/).

codalab/worker/dependency_manager.py
epicfaace: @yiboliu can you take a look at this?

codalab/worker/docker_image_manager.py
epicfaace: Can we replace all instances of `capture_message` with `logger.error` then? I don't see the point of using `capture_message` when we can just log them so that they are sent both to the console and Sentry.

codalab/worker/docker_image_manager.py
```diff
@@ -293,6 +293,7 @@ def download():
                     message=self._downloading[image_spec]['status'],
                 )
         except Exception as ex:
+            capture_exception(ex)
```
epicfaace: You should replace this with logger.error

epicfaace (changes requested):

epicfaace (commented):
Looks good. @jzwang43 can you show a screenshot of how `error_traceback` looks like for a sample failed bundle once you run `cl info` on that bundle?

epicfaace:
Ensure the traceback doesn't contain any sensitive information / things we don't want to make public.

jzwang43:
> 
> 
> Looks good. @jzwang43 can you show a screenshot of how `error_traceback` looks like for a sample failed bundle once you run `cl info` on that bundle?

I have updated the screen shots.

jzwang43:
> 
> 
> Looks good. @jzwang43 can you show a screenshot of how `error_traceback` looks like for a sample failed bundle once you run `cl info` on that bundle?



> 
> 
> Ensure the traceback doesn't contain any sensitive information / things we don't want to make public.

I think the trackbacks will only include the functions in the call stack that cause the error. Since our code is open-source, it shouldn't include any sensitive information.

epicfaace (commented):
Sounds good. Perhaps we could not show error_traceback in the web UI or when running `cl info` by default, and only show it if someone explicitly runs `cl info -f error_traceback`? This would just be for ease of use for the user, because we probably don't want regular users to be seeing tracebacks when just looking at bundle info in general (which is a bad user experience).

I'm thinking that maybe passing a kwarg such as `hidden=True` to the `MetadataSpec` constructor could trigger this behavior.

percyliang:
Is this necessary?  What does the output look like?  If you do a `cl info`, you're getting pretty raw information anyway, and it seems simpler if we didn't have this extra constraint.

epicfaace:
@percyliang the output looks like this (note how the traceback can be multiple lines long, unlike the rest of the metadata fields):
image

Although it won't hurt to show this to the user, I just don't think this data is relevant for people other than codalab developers.
 This is because `error_traceback` will never show user-code-related errors, but rather only codalab-related errors.

jzwang43:
> 
> 
> @percyliang the output looks like this (note how the traceback can be multiple lines long, unlike the rest of the metadata fields):
> image
> 
> Although it won't hurt to show this to the user, I just don't think this data is relevant for people other than codalab developers.
> This is because `error_traceback` will never show user-code-related errors, but rather only codalab-related errors.

I made the change so that error traceback only shows up with `-f` just in case we need it, but either way should be fine. Screen shot is updated in the PR description.

epicfaace:
> I made the change so that error traceback only shows up with `-f` just in case we need it, but either way should be fine

Cool, can you add a test in test_cli.py that ensures that error_traceback is accessible through `-f`, and that it doesn't show up without `-f`? Then we should be good to merge.

codalab/objects/metadata_spec.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         completer=None,
         hide_when_anonymous=False,
         optional=False,
+        hidden=False,
```
epicfaace: Can we add a comment here explaining what the `hidden` attribute does?

epicfaace (approved):
We decided there's no need to add this test because it's hard to run a command that results in the error_traceback field being set.

PR #3264, opened 2021-02-12
DO NOT MERGE

eicherseiji:

Hey all, thanks for the interesting problem! Please find my solution attached here.

### Checklist


epicfaace:
Thanks for your submission!

PR #3270, opened 2021-02-15
Fix the Editing Avatar Permission Error

### LINKED ISSUE #3267: Avatar Editing Permission Error
**Describe the bug**

Screen Shot 2021-02-12 at 3 49 36 PM

yuqijin:

The reason why the permission error occurred when a user tries to edit his/her avatar is normal user has no ALL permission to his/her dashboard worksheet, instead, the user only has READ permission. Only the root user has ALL permission to the dashboard worksheet. So the user cannot add the avatar bundle to the dashboard worksheet.

In order to avoid this problem, the current solution is to treat the avatar bundle as a special bundle that it does not need a host worksheet when being created. But the user can still access his/her avatar bundle by using the CLI, e.g. `search .mine`

### Related issues

fixes #3267 

### Screenshots

Screen Shot 2021-02-14 at 11 36 11 PM

### Checklist


codalab/rest/bundles.py
```diff
@@ -224,12 +224,16 @@ def _create_bundles():
       "uploading" regardless of the bundle type, or 0 otherwise. Used when
       copying bundles from another CodaLab instance, this prevents these new
       bundles from being executed by the BundleManager. Default is 0.
+    - `avatar`: 1 if the bundle is the user's uploaded avatar,
```
epicfaace: Should we replace this parameter with something more general, such as `parent_worksheet_required`, instead?

yuqijin: I think that's a good idea. Update the parameter to be `parent_worksheet_required`

codalab/rest/bundles.py
```diff
@@ -224,12 +224,16 @@ def _create_bundles():
...
+      (e.g. It is the user's uploaded avatar) or 1 otherwise.
+      Default is 1 since the parent worksheet need to be specified by default.
     """
     worksheet_uuid = request.query.get('worksheet')
     shadow_parent_uuid = request.query.get('shadow')
     after_sort_key = request.query.get('after_sort_key')
     detached = query_get_bool('detached', default=False)
-    if worksheet_uuid is None:
+    parent_worksheet_required = query_get_bool('parent_worksheet_required', default=True)
+    if parent_worksheet_required and worksheet_uuid is None:
```
percyliang: Why not just allow `worksheet_uuid` to be None, in which case the bundle doesn't get added to any worksheet?

yuqijin: I did not do so because if we set `worksheet_uuid` to None for avatar bundles, then we have to remove the check of whether the `parent_worksheet` is specified in URL, which may result in some error when creating normal bundles, since the normal bundles still need hosted worksheet to inherit the permission from.

codalab/rest/bundles.py
```diff
@@ -243,9 +247,10 @@ def _create_bundles():
     )
```
epicfaace, line 246: For simplicity, let's set `parent_worksheet_required = not detached` and remove the `parent_worksheet_required` query parameter.

yuqijin: Yeah, it makes sense based on the discussion! Just committed the update :)

codalab/rest/bundles.py
```diff
@@ -277,20 +282,21 @@ def _create_bundles():
...
-                {
-                    'object_uuid': bundle_uuid,
-                    'group_uuid': p['group_uuid'],
-                    'permission': p['permission'],
-                }
-                for p in group_permissions
-            ]
-        )
+        if not detached:
+            # Inherit worksheet permissions
```
yipenghe: Interestingly no bug was thrown for not having this check before. Maybe add a comment to say that if it's detached, then only the user will have all permissions on the bundle.

yuqijin:
```suggestion
            # Inherit worksheet permissions; else, only the user will have all permissions on the bundle
```

yuqijin: thanks for the advice. appiled!

codalab/rest/bundles.py
```diff
@@ -219,6 +219,7 @@ def _create_bundles():
     - `shadow`: UUID of the bundle to "shadow" (the new bundle will be added
       as an item immediately after this bundle in its parent worksheet).
     - `detached`: 1 if should not add new bundle to any worksheet,
+      so the bundle does not have a hosted worksheet(e.g. It is the user's uploaded avatar)
```
```suggestion
      so the bundle does not have a hosted worksheet.
      This is set to 1, for example, if the user is uploading their avatar as a bundle.
```

frontend/src/components/Dashboard/EditableAvatar.js
```diff
@@ -239,15 +239,13 @@ class EditableAvatar extends React.Component {
...
             }
 
-            // Fetch the uuid of user's dashboard worksheet for storing the profile-image bundle
-            const response = await fetch(`/rest/worksheets?specs=dashboard`).then((e) => e.json());
-            const worksheetUUID = response.data[0].id;
             const file = this.state.file;
-
             // Rename the file to let it be related to the user
             const fileName: string = this.props.userInfo.user_name + '.JPG';
+            // Specify the bundle is an avatar
```
epicfaace:
```suggestion
```

frontend/src/components/Dashboard/EditableAvatar.js
```diff
@@ -239,15 +239,13 @@ class EditableAvatar extends React.Component {
...
+            // So no parent worksheet ID is needed that the bundle is a detached bundle
```
epicfaace:
```suggestion
            // Create a detached bundle, so that no parent worksheet ID is needed for the avatar
```

epicfaace (approved):

PR #3271, opened 2021-02-15
Manually bind context to `this` for calllback function to solve "undefined" error

### LINKED ISSUE #3265: Cannot read property 'setState' of undefined
This error just happened randomly when I was running codalab locally.

image

yuqijin:

Manually bind context to `this` to have access inside the callback function. 

### Related issues

fixes #3265 

### Checklist

### DIFF
--- a/frontend/src/components/Bundle/Bundle.js
+++ b/frontend/src/components/Bundle/Bundle.js
@@ -153,21 +153,23 @@ class Bundle extends React.Component<
                     this.setState(stateUpdate);
                 }
             })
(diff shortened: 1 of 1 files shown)


epicfaace (approved):
Looks good. Do you think it might be clearer to just do `(xhr, status, err) => ...` instead of running `.bind(this)`?

yuqijin:
> Looks good. Do you think it might be clearer to just do `(xhr, status, err) => ...` instead of running `.bind(this)`?

Personally, I would prefer to keep a callback function here.

PR #3272, opened 2021-02-16
Redo the stats of docker images

### LINKED ISSUE #3003: View CPU / memory / network load stats for bundles

### LINKED ISSUE #3234: Re-add docker stats

yiboliu:
This reverts commit dd81f5f0f5668ba4009a79418dd6c45e10a7d24e.
Re-add the stats of docker images, such as cpu usage and memory limit. 

### Related issues

fixes https://github.com/codalab/codalab-worksheets/issues/3003, fixes https://github.com/codalab/codalab-worksheets/issues/3234

### Checklist

### DIFF
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -37,9 +37,9 @@ class RunBundle(DerivedBundle):
                                                                 'priority bundles are queued behind bundles with no specified priority.', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
     METADATA_SPECS.append(
-        MetadataSpec('cpu_usage', float, 'Portion of CPU used by this container (e.g., 0.24)', default=0.0, generated=True))
+        MetadataSpec('cpu_usage', float, 'Proportion of current CPU usage for a running bundle. This field is only relevant for running bundles. (e.g., 0.24)', default=0.0, generated=True))
     METADATA_SPECS.append(
-        MetadataSpec('memory_limit', int, 'Limit of Memory available to this container (e.g., 2085326848)', default=0, generated=True))
+        MetadataSpec('memory_limit', int, 'How much memory this container can use at any given time.', default=0, generated=True))
 
     METADATA_SPECS.append(MetadataSpec('exclude_patterns', list, 'Exclude these file patterns from being saved into the bundle contents.', default=[]))
 
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -39,7 +39,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(
         MetadataSpec('cpu_usage', float, 'Proportion of current CPU usage for a running bundle. This field is only relevant for running bundles. (e.g., 0.24)', default=0.0, generated=True))
     METADATA_SPECS.append(
-        MetadataSpec('memory_limit', int, 'How much memory this container can use at any given time.', default=0, generated=True))
+        MetadataSpec('memory_limit', int, 'The maximum amount of memory the container can use.', default=0, generated=True))
 
     METADATA_SPECS.append(MetadataSpec('exclude_patterns', list, 'Exclude these file patterns from being saved into the bundle contents.', default=[]))
 
--- a/codalab/bundles/run_bundle.py
+++ b/codalab/bundles/run_bundle.py
@@ -39,7 +39,7 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(
         MetadataSpec('cpu_usage', float, 'Proportion of current CPU usage for a running bundle. This field is only relevant for running bundles. (e.g., 0.24)', default=0.0, generated=True))
     METADATA_SPECS.append(
-        MetadataSpec('memory_limit', int, 'The maximum amount of memory the container can use.', default=0, generated=True))
+        MetadataSpec('memory_usage', float, 'Proportion of current memory usage based on the memory limit.', default=0.0, generated=True))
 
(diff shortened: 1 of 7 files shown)


epicfaace:
0. Make sure GHA builds the docker images
1. Start a bundle that says `sleep 1800`
2. Deploy this branch to dev (venv/bin/...) (set the codalab_version to the branch name - `redostats`)
3. See if the bundle is still running / finishes successfully. (the bug is that the bundle fails / becomes worker_offline)

adiprerepa:
blocked until dev issue gets fixed

adiprerepa:
Works on dev

codalab/bundles/run_bundle.py
```diff
@@ -36,6 +36,11 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important). Negative '
                                                                 'priority bundles are queued behind bundles with no specified priority.', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(
+        MetadataSpec('cpu_usage', float, 'Portion of CPU used by this container (e.g., 0.24)', default=0.0, generated=True))
+    METADATA_SPECS.append(
+        MetadataSpec('memory_limit', int, 'Limit of Memory available to this container (e.g., 2085326848)', default=0, generated=True))
```
epicfaace: I'm a bit confused by what memory_limit actually means, according to this description. Can you look into this and update the description to be clearer about what is actually means?

codalab/bundles/run_bundle.py
```diff
@@ -36,6 +36,11 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important). Negative '
                                                                 'priority bundles are queued behind bundles with no specified priority.', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(
+        MetadataSpec('cpu_usage', float, 'Portion of CPU used by this container (e.g., 0.24)', default=0.0, generated=True))
```
epicfaace: I think we should update the description to clarify that it refers to **current** CPU usage of a running bundle and that it is only set / relevant for bundles that are currently running.

codalab/bundles/run_bundle.py
adiprerepa: sgtm

codalab/bundles/run_bundle.py
adiprerepa: sgtm

codalab/bundles/run_bundle.py
```diff
@@ -36,6 +36,11 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('request_priority', int, 'Job priority (higher is more important). Negative '
                                                                 'priority bundles are queued behind bundles with no specified priority.', default=None))
     METADATA_SPECS.append(MetadataSpec('request_network', bool, 'Whether to allow network access.', default=False))
+    METADATA_SPECS.append(
+        MetadataSpec('cpu_usage', float, 'Proportion of current CPU usage for a running bundle. This field is only relevant for running bundles. (e.g., 0.24)', default=0.0, generated=True))
+    METADATA_SPECS.append(
+        MetadataSpec('memory_limit', int, 'How much memory this container can use at any given time.', default=0, generated=True))
```
epicfaace: sorry, still not clear what this means. Does memory_limit change during the duration of a run? It sounds like it might be a constant value if it's "how much memory this container **can use**"

adiprerepa: https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory

according to docker docs "The maximum amount of memory the container can use.". I believe that it's a constant.

epicfaace: Huh, well if it's just a constant, then maybe we should just remove it? See #3003 for the original issue of adding docker stats; not sure if a constant max memory value is really that useful.

adiprerepa: you are right.

I am not sure why yibo wanted to have `memory_limit` as a statistic. A more appropriate statistic would be the memory usage, so that would be `memory_usage/memory_stats`. I'll make this change.

codalab/worker/worker_run_state.py
```diff
@@ -461,6 +463,12 @@ def check_and_report_finished(run_state):
 
         def check_resource_utilization(run_state: RunState):
             logger.info(f'Checking resource utilization for bundle. uuid: {run_state.bundle.uuid}')
+            cpu_usage, memory_usage = docker_utils.get_container_stats_with_docker_stats(
+                run_state.container
+            )
+            run_state = run_state._replace(cpu_usage=cpu_usage)
```
epicfaace: you should be able to just combine this into a single line

```suggestion
            run_state = run_state._replace(cpu_usage=cpu_usage, memory_usage=memory_usage)
```

epicfaace (approved):
lgtm, did you verify that memory_usage actually shows up as a non-zero value when getting stats of an actual bundle?

Also, it would be great to add a test -- maybe we can add a unit test for get_cpu_usage  (since the logic is kind of complicated), and feel free to add a test in test_cli.py if you think it's needed.

PR #3282, opened 2021-02-19
Implement partial refresh of worksheet for updating markup_block

### LINKED ISSUE #2174: When refreshing the page after a worksheet item is saved, don't refresh the search results
This will improve the user experience for users, especially on worksheets with many search results on it (such as the dashboard worksheet).

yuqijin:

Improve performance by not refreshing the search results when only updating markup_block.

The improvement only makes sense to markup_block.

**Need refresh search results:**
- table_block
- schema_block

**Can't update block through UI:**
- record_block
- contents_block
- image_block
- graph_block
- subworksheets_block
- placeholder_block

### Related issues

fixes #2174 

### Screenshots

ezgif com-gif-maker

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -98,6 +98,7 @@ class Worksheet extends React.Component {
                 showMessage: false,
                 messageContent: null,
             },
+            searchResultCache: {}, // Store the fetched search results
         };
         this.copyCallbacks = [];
         this.bundleTableID = new Set();
@@ -133,6 +134,18 @@ class Worksheet extends React.Component {
             success: function(info) {
                 info['date_created'] = addUTCTimeZone(info['date_created']);
                 info['date_last_modified'] = addUTCTimeZone(info['date_last_modified']);
+                // Use the cache to update directive block if we are in the partial update mode
+                if (!props.searchRefresh) {
+                    for (let i: number = 0; i < info.blocks.length; i++) {
+                        const directive: string = info.blocks[i]['directive'];
+                        if (
+                            'directive' in info.blocks[i] &&
+                            directive in this.state.searchResultCache
+                        ) {
+                            info.blocks[i] = this.state.searchResultCache[directive];
+                        }
+                    }
+                }
                 this.setState({
                     ws: {
                         ...this.state.ws,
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/items/TextEditorItem/TextEditorItem.js
```diff
@@ -93,7 +93,9 @@ class TextEditorItem extends React.Component<{
             type: 'POST',
             success: (data, status, jqXHR) => {
                 const moveIndex = true ? mode === 'create' : false;
-                const param = { moveIndex };
+                // No need to refresh the search results when only updating markdown items
+                const searchRefresh = false;
```
epicfaace: Instead of making the prop `searchRefresh`, can we rename it to `skipSearchRefresh`? This way, it will be clearer what the prop does, and we only set it when we want to do something special (which is skipping the search refresh).

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1404,13 +1431,15 @@ class Worksheet extends React.Component {
             textDeleted = false,
             fromDeleteCommand = false,
             uploadFiles = false,
+            searchRefresh = true,
```
epicfaace: Can you add a comment here briefly explaining what this parameter means?

yuqijin: added.

frontend/src/components/worksheets/items/TextEditorItem/TextEditorItem.js
yuqijin: Sure. Updated

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -133,6 +134,18 @@ class Worksheet extends React.Component {
             success: function(info) {
                 info['date_created'] = addUTCTimeZone(info['date_created']);
                 info['date_last_modified'] = addUTCTimeZone(info['date_last_modified']);
+                // Use the cache to update directive block if we are in the partial update mode
+                if (props.skipSearchRefresh) {
```
epicfaace, line 138: Actually, it's considered not the best practice to store cache in state. Can we use something like `useSWR` (https://swr.vercel.app/getting-started) for the cache instead?

epicfaace:
Closing this in favor of https://github.com/codalab/codalab-worksheets/pull/3353.

PR #3215, opened 2021-01-24
Stress test a worker uploading large bundle results (2)

epicfaace:
Recreating #3072, as we had to revert this PR for the latest release. We can merge this PR back in when #2722 is fixed.

PR #3285, opened 2021-02-19
Fix "uploading results" error by sending data to keep the socket open

### LINKED ISSUE #2722: [Prod] Run bundles are stuck in uploading results
In production, some bundles are stuck with the status `Uploading results: 14.5g done (archived size)` and continues to occupy the worker. Killing the bundles doesn't do anything either.

epicfaace:
Fixes #2722 by sending data to keep the socket open.

When the worker calls `_upload_with_chunked_encoding` to upload results to the rest server, sometimes the rest server takes a long time to send its response, because it takes a very long time (> 20 minutes) to extract the uploaded `.tar.gz` file, and then the connection is dropped, causing #2722.

This PR fixes that issue by adding a 60-second timeout to `http.client`. The connection will wait 60 seconds for the server response; if no response is received, it will send a null byte to keep the connection open and repeat the wait.

I've tested this on dev with the 16 GB file used in the scenario from #3215, and the bundle is no longer stuck in "uploading results"!


epicfaace:
@teetone can you take a look when you get a chance?

codalab/worker/rest_client.py
```diff
@@ -160,7 +163,21 @@ def _upload_with_chunked_encoding(
...
             # Read the response.
-            response = conn.getresponse()
+            logging.debug("About to read the response... url: %s", url)
+
+            # Sometimes, it may take a while for the server to process
+            # the data and send the response. In this case, we want to
+            # periodically keep sending empty bytes so that the
+            # connection doesn't drop before the response is available.
+            got_response = False
+            while not got_response:
```
yipenghe, line 173: Any chance this results in a infinite loop? Feeling at some point we should fail the request after a set number of tries or a dynamic try attempt number depending on the upload size (if possible to get that info). The initial attempt number can be larger, before we get a better understanding of the retries it generally requires to finish the progress

codalab/worker/rest_client.py
```diff
@@ -160,7 +163,21 @@ def _upload_with_chunked_encoding(
...
+                try:
+                    response = conn.getresponse()
+                    got_response = True
+                except socket.timeout:
+                    logging.debug("Socket timeout, retrying")
```
yipenghe: We should probably log the number of retries to get, so we have a sense of how many retries in general we need to upload large bundles. (If we fail after a set number of tries / dynamic number of tries, we should log that failing event too)

codalab/worker/rest_client.py
epicfaace: We're actually already in an infinite loop with the code from `master` -- `conn.getresponse()` just hangs forever, which is causing the error. Since this PR is just a strict improvement to what was there earlier, I'd rather just do the bare minimum to fix the problem for now. I think your suggestions make sense for future improvements to this function though!

codalab/worker/rest_client.py
epicfaace:
```suggestion
                    logging.debug("Socket timeout, retrying url: %s", url)
```

epicfaace: Good idea, I just logged the URL for now, so that we can distinguish the socket retries and count the number of logs per URL if need be.

codalab/worker/rest_client.py
yipenghe: Sounds good

yipenghe (approved):
Looks good, since it's a strict improvement. Let's also test on dev on this change before deploying  to prod

epicfaace:
Yeah, I tested this on dev earlier. We can also add in #3215 after this PR, so that the stress tests will always test this scenario in the future.

Issue #3287, opened 2021-02-20
AttributeError: module 'codalab.worker.file_util' has no attribute 'un_gzip_stream'


epicfaace:
This error appears to happen whenever you try to stream a file but don't include the `Accept-Encoding: gzip` header in your request.

Prod error:

```
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body style="word-break: break-word;"><p>Error on request by public(None):

GET /bundles/0x02022dd69ab94632882f67603d344f88/contents/blob/id_val_eval.csv

Query params:


Form params:


JSON body:
&lt;none&gt;

Local variables:
fileobj=&lt;_io.BufferedReader name=25&gt;
read_args={'type': 'stream_file'}
response_socket_id=6091145
worker={'user_id': '0x6f747813c550409e9fd1fdab32a7c7da', 'worker_id': '$(hostname -s)-323e408f8fad4981b0a7bebfcc1cdb47', 'shared_file_system': False, 'socket_id': 6090453}
gzipped=False
target=&lt;codalab.worker.download_util.BundleTarget object at 0x7f31d0582c88&gt;
self=&lt;codalab.lib.download_manager.DownloadManager object at 0x7f31f232ee10&gt;

Traceback (most recent call last):
  File "/opt/codalab-worksheets/codalab/server/rest_server.py", line 110, in wrapper
    return callback(*args, **kwargs)
  File "/usr/local/bin/bottle.py", line 1748, in wrapper
    rv = callback(*a, **ka)
  File "/opt/codalab-worksheets/codalab/server/json_api_plugin.py", line 18, in wrapper
    result = callback(*args, **kwargs)
  File "/opt/codalab-worksheets/codalab/server/authenticated_plugin.py", line 136, in wrapper
    return callback(*args, **kwargs)
  File "/opt/codalab-worksheets/codalab/rest/bundles.py", line 672, in _fetch_bundle_contents_blob
    fileobj = local.download_manager.stream_file(target, gzipped_stream)
  File "/opt/codalab-worksheets/codalab/lib/download_manager.py", line 21, in wrapper
    return f(*args, **kwargs)
  File "/opt/codalab-worksheets/codalab/lib/download_manager.py", line 212, in stream_file
    fileobj = self.file_util.un_gzip_stream(fileobj)
[...]

PR #3288, opened 2021-02-20
Revert "Speed up CLI by lazy loading some imports"

epicfaace:
Reverts codalab/codalab-worksheets#3212 because it caused #3287.


epicfaace:
Closing this in favor of #3289 instead.

Issue #2771, opened 2020-08-20
Separate out cleanup logic from the worker


teetone:
**Describe the bug**

Currently, the cleanup logic resides inside the worker loop and only executes on a happy path. We should see if this can be separated out and cleaned up no matter what.

https://github.com/codalab/codalab-worksheets/blob/bf744068aaddccd2509e31a4c3666768bc3b0d2f/codalab/worker/worker.py#L252-L297

yiboliu:
@teetone sorry I'm a bit confused. To my understanding the worker loop is going exit anyway, and `cleanup()` will be called in the end as it's outside of the loop. Why do we still want to cleanup no matter what?

PR #3291, opened 2021-02-21
Ensure cleanup is executed

yiboliu:
Make sure cleanup is executed in all cases. 

### Related issues
#2771 

### Checklist

### DIFF
--- a/codalab/worker/worker.py
+++ b/codalab/worker/worker.py
@@ -260,17 +260,22 @@ class Worker:
 
     def start(self):
         """Return whether we ran anything."""
-        self.load_state()
-        self.sync_state()
-        self.image_manager.start()
-        if not self.shared_file_system:
-            self.dependency_manager.start()
+        try:
+            self.load_state()
+            self.sync_state()
+            self.image_manager.start()
+            if not self.shared_file_system:
+                self.dependency_manager.start()
+        except Exception as ex:
+            # if we encounter any problem of starting up before it goes to the loop, we still want to cleanup.
+            logger.error(ex)
+            self.cleanup()
         while not self.terminate:
             try:
                 self.process_runs()
                 self.save_state()
                 self.checkin()
-                self.check_termination()
+                self.check_termination()  # if terminate_and_restage and no restaged bundles, then it will terminate.
                 self.save_state()
                 if self.check_idle_stop() or self.check_num_runs_stop():
                     self.terminate = True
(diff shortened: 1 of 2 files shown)


codalab/worker/worker.py
```diff
@@ -309,6 +314,7 @@ def start(self):
                             except Exception:
                                 traceback.print_exc()
                             self.terminate = True
+                        # else: if it's not ready to terminate and restage, I think we can't clean up.
```
epicfaace: We don't need this comment

codalab/worker/worker.py
```diff
@@ -260,17 +260,22 @@ def check_num_runs_stop(self):
...
+            # if we encounter any problem of starting up before it goes to the loop, we still want to cleanup.
+            logger.error(ex)
+            self.cleanup()
         while not self.terminate:
             try:
                 self.process_runs()
                 self.save_state()
                 self.checkin()
-                self.check_termination()
+                self.check_termination()  # if terminate_and_restage and no restaged bundles, then it will terminate.
```
epicfaace: You should move this comment to be a docstring within the `check_termination` method definition itself.

yiboliu: Oh this comment was actually a note for me to understand the logic. It's not meant to be there. Sorry

codalab/worker/worker.py
```diff
@@ -260,11 +260,16 @@ def check_num_runs_stop(self):
...
-        self.image_manager.start()
-        if not self.shared_file_system:
-            self.dependency_manager.start()
+        try:
+            self.load_state()
+            self.sync_state()
+            self.image_manager.start()
+            if not self.shared_file_system:
+                self.dependency_manager.start()
+        except Exception as ex:
```
epicfaace, line 269: Let's rename `ex` to `e`, for consistency with the rest of the codebase.

codalab/worker/worker.py
```diff
@@ -260,11 +260,16 @@ def check_num_runs_stop(self):
 
     def start(self):
         """Return whether we ran anything."""
-        self.load_state()
-        self.sync_state()
-        self.image_manager.start()
-        if not self.shared_file_system:
-            self.dependency_manager.start()
+        try:
```
epicfaace, line 263: Would it be harmful to run `cleanup()` at, say, the line where we call `self.load_state()` (because the image manager hasn't been started yet)? @teetone , what do you think?

epicfaace:
Closing this for now -- we can revisit https://github.com/codalab/codalab-worksheets/issues/2771 later.

PR #3294, opened 2021-02-25
Cache the directives being fetched to avoid duplicate fetch of the same directives

### LINKED ISSUE #3277: Duplicate requests are made to resolve directives
Only a single request should be made, but 2-4 requests are made.

image

yuqijin:

Previously, in order to resolve a directive, multiple requests will be made.
The problem is due to the Worksheet component rerender several times, and every time it re-render, it creates the placeholder items if the results of directives have not been fetched since the fetch of directives is an async process, which results in multiple requests.

To solve it and improve performance, create a cache to store the directives being fetched to avoid duplicate fetch of the same directives. 
Here 'same' is determined by both `directive` and `version`, e.g. `wsearch .mine` in version `1`

### Related issues

fixes #3277 

### Screenshots

Prev:
Screen Shot 2021-02-24 at 1 40 49 AM

After:
Screen Shot 2021-02-24 at 6 05 23 PM

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/WorksheetItemList.js
+++ b/frontend/src/components/worksheets/WorksheetItemList.js
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import Immutable from 'seamless-immutable';
 import $ from 'jquery';
 import * as Mousetrap from '../../util/ws_mousetrap_fork';
 import { getAfterSortKey, getIds } from '../../util/worksheet_utils';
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -187,6 +106,117 @@ class WorksheetItemList extends React.Component {
...
+        // we must show the new upload / new run buttons after each row in the table (in the BundleRow component)
+        // as opposed to at the end of the table (in ItemWrapper).
+        props.showNewButtonsAfterEachBundleRow =
+            props.item.mode === 'table_block' && !props.item.loadedFromPlaceholder;
+        const constructor = BLOCK_TO_COMPONENT[item.mode];
+
+        // for a placeholder block, if the directive in current version is already being fetched
+        // no need to fetch the same directive again
+        let skipDirectiveFetch = false;
+        if (item.mode === 'placeholder_block') {
```
epicfaace, line 147: You mention that the problem is due to the "Worksheet component rerender several times." Is it possible to perhaps only make the worksheet component render once, and then perhaps we won't need a cache? Or is that not possible?

yuqijin: I considered the solution but it was difficult though, since may async process are related to the worksheet

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -187,6 +106,117 @@ class WorksheetItemList extends React.Component {
...
+            const directive = item.directive;
+            if (
+                props.version <= this.state.fetchingDirectives.length - 1 &&
+                this.state.fetchingDirectives[props.version].has(directive)
+            ) {
+                skipDirectiveFetch = true;
```
epicfaace, line 153: It just feels like this caching logic is pretty complicated. Could we instead always call `fetchData`, but then in the implementation of `fetchData` itself, we add the caching logic? That way, we can keep the logic within react components much more simple.

epicfaace: We could use something like `useSWR` (https://swr.vercel.app/getting-started) instead.

epicfaace:
Closing this in favor of https://github.com/codalab/codalab-worksheets/pull/3353.

Issue #3150, opened 2021-01-06
[FEATURE REQUEST] "Locking" worksheets and bundles to prevent them from being modified or deleted


nelson-liu:
**Is your feature request related to a problem? Please describe.**

CodaLab serves dual purposes as an experimental platform and as a place to share results. For the latter, it'd be nice if there was some way to "lock" bundles or worksheets such that they can no longer be modified (unless they're unlocked).

As someone who has many such bundles and worksheets, I'm always afraid that an errant `cl rm $(cl search)` is going to wipe away existing results that I want on CodaLab for archival.

**Describe the solution you'd like**

Commands `cl lock` and `cl unlock` for worksheets and bundles that change whether or not their metadata can be modified / whether or not they can be deleted.

**Describe alternatives you've considered**

Not sold on the `lock` and `unlock` names, `freeze` and `unfreeze` might be better, for instance.

percyliang:
`cl wedit --freeze` exists at least...  Freezing bundles seems reasonable too.

Issue #2293, opened 2020-05-23
cl rm with xargs doesn't respect current cl instance


nelson-liu:
Unfortunately, I learned this the hard way (luckily, not much was lost). Maybe this is intended behavior, let me know if so. Here's a reproduction:

1. Switch to main
```
$ cl work main::
Switched to worksheet: https://worksheets.codalab.org/worksheets/0x6c26647dad70404eb6ea6a01ed8b6a99 (home-nfliu)
```

2. list worksheet:

```
### Worksheet: https://worksheets.codalab.org::home-nfliu(0x6c26647dad70404eb6ea6a01ed8b6a99)
### Title: <none>
### Tags:
### Owner: nfliu(0x23c379b450cb4552b314268c413e2468)
### Permissions: public(0xc573c2):read
ref  uuid      name          summary  owner  created              data_size  state
----------------------------------------------------------------------------------
 ^1  0xd9fb67  run-allennlp           nfliu  2020-02-19 16:54:43       1.4g  ready
```

3. Run a test bundle

```
$ cl run "echo hi"
0x72fe7c01677146a2a98e8c3e6565054b
```

4. Verify it exists

```
$ cl i 0x72fe7c01677146a2a98e8c3e6565054b
bundle_type               : run
uuid                      : 0x72fe7c01677146a2a98e8c3e6565054b
data_hash                 : 0x35b085d2d5fbb7decdccdefecddb520280ba3a26
state                     : ready
command                   : echo hi
is_anonymous              : False
owner                     : nfliu(0x23c379b450cb4552b314268c413e2468)
name                      : run-echo
created                   : 2020-05-22 17:58:22
data_size                 : 4.0k
allow_failed_dependencies : False
request_memory            : 2g
request_cpus              : 1
request_gpus              : 0
request_priority          : 0
request_network           : False
time                      : 0.1s
time_user                 : 0.0s
time_system               : 0.0s
started                   : 2020-05-22 17:58:26
last_updated              : 2020-05-22 17:58:53
run_status                : Finished
docker_image              : codalab/default-cpu@sha256:ff7de716ac5f94d9b4ff8d6341d60166ed5b0163cc05d8d51da9418511caf5e1
exitcode                  : 0
remote                    : jagupard4.stanford.edu(12976)
```

5. Show it in search:

```
$ cl search run-echo .mine
uuid      name      summary  owner  created              data_size  state
-------------------------------------------------------------------------
[...]

PR #3295, opened 2021-02-25
Disallow editing metadata of frozen worksheets; enable unfreezing worksheets

nelson-liu:

Part of #3150 . Enables unfreezing bundles. This didn't require any backend changes, just some changes to the CLI, since the `frozen` field in the DB is nullable anyway.


percyliang:
The original point of freezing a worksheet is that you commit to a particular set of contents which is truly immutable.  The idea is that when you make an executable paper, you freeze it so people visiting a worksheet have some sort of guarantee (which you get for bundles).  But in practice no one uses this feature...

What's your use case?

nelson-liu:
I actually want to use this feature :) . I also want a parallel `freeze` / `unfreeze` for bundles (that would prevent deletion). 

However, I'm nervous about accidentally freezing the worksheet (and needing to change something later) / the "permanent" aspect. I personally care less about showing people that it was permanently frozen on a date, and more that I don't accidentally delete it in the future by doing something dumb. Besides, the permanence isn't all that permanent, since you can get around it client-side----might as well expose it.

> you freeze it so people visiting a worksheet have some sort of guarantee (which you get for bundles)

There's currently no way to tell if / when a worksheet is frozen, right? (it doesn't show up on the web UI, and there's no `winfo`)

percyliang:
Exposing the frozen bit in the UI is pretty easy to do (just never got around to it).

Freezing the worksheet I think doesn't protect the bundles on the worksheet if you do `cl rm -f`.  Can you say a bit more about the protection you have in mind?

In general, since I don't think anyone else is using this feature, you might as well repurpose it for your use case, but just want to understand it a bit better.

nelson-liu:
> Freezing the worksheet I think doesn't protect the bundles on the worksheet if you do cl rm -f. Can you say a bit more about the protection you have in mind?

Right, I think that's fine (I recognize that `-f` is very dangerous). A lot of how i interact with codalab is via `cl search`, and I often want to bulk-delete bundles (`cl rm $(cl search <something>)`). The "leaf" experiment bundles that I might put into an executable paper (i.e., things that have no downstream dependents) have no protection from an errant `cl rm $(cl search)`. I'm also perhaps overcautious after getting bitten by this once, and I just want to ensure that I can work with reckless abandon and not worry about modifying any executable papers i have. https://github.com/codalab/codalab-worksheets/issues/2293

nelson-liu:
I made the freeze feature even stronger---i don't see a good reason for being able to update the metadata of a frozen worksheet, especially if you can unfreeze it. Now, frozen worksheets cannot be modified at all [1], contents or metadata.

[1] There is only one attribute of a frozen worksheet that you can modify: whether or not it's frozen.

tests/cli/test_cli.py
```diff
@@ -1077,26 +1077,47 @@ def test(ctx):
...
+    _run_command([cl, 'wedit', '--unfreeze'])
+
+    # After unfreezing: can modify everything
+    _run_command([cl, 'detach', uuid1])  # would remove an item
+    _run_command([cl, 'rm', uuid1])  # would remove an item
+    _run_command([cl, 'add', 'text', 'message'])  # would add an item
     _run_command([cl, 'wedit', '-t', 'new_title'])  # can edit
     _run_command([cl, 'wperm', wuuid, 'public', 'a'])  # can edit
 
+    # Verify that we can make multiple edits to a frozen worksheet,
```
epicfaace: I think what this code is really verifying is that "we can make a single edit to a frozen edit, as long as we unfreeze at the same time" -- can we update the comment accordingly?

epicfaace (approved):

tests/cli/test_cli.py
epicfaace: @nelson-liu I just slightly updated the tests for clarity regarding this comment; let me know if you have any concerns!

nelson-liu: Oops sorry, I must have missed this---you're correct, the new comment is more accurate. Thanks for following up!

PR #3300, opened 2021-02-26
Blob Storage: Add storage_type and is_dir columns

epicfaace:

- Add `storage_type` and `is_dir` columns. These two columns are required in order to distinguish Azure Blob bundles from regular bundles. Because both files and folders will be stored in a `.tar.gz` format, we need `is_dir` to distinguish between files and folders on Blob Storage.
  - @percyliang I know we initially thought we could just store individual-file bundles directly on blob storage without tarring / gzipping them, but I realized that we need to use `.tar.gz` to properly preserve permissions and other properties. Additionally, gzipped files are smaller.
- Change the `StorageType` class to have string values, as those are the enum values that will be in the database for the `storage_type` column.
- Move `LinkedBundlePath` and `parse_linked_bundle_url` to `common.py` (so they can be used by both the server and CLI), and update them to use the new enum values from `StorageType`.

- The `storage_type` and `is_dir` columns are initially set to null on creation of a bundle. When calling `upload_to_bundle_store`, the bundle will get these two columns set based on how the user uploads the bundle contents.
  - As a consequence of this change (uploading bundle contents now mutates the database), I had to refactor the bundle manager unit tests as well.

- Add the `get_bundle_storage_info` and `get_bundle_storage_infos` methods to bundle model, which retrieve the `storage_type` and `is_dir` of the input bundle(s).


codalab/common.py
```diff
@@ -133,3 +135,73 @@ def urlopen_with_retry(request: urllib.request.Request, timeout: int = URLOPEN_T
...
+
+
+def parse_linked_bundle_url(url):
+    """Parses a linked bundle URL. This bundle URL usually refers to:
+        - an archive file on Blob Storage: "azfs://storageclwsdev0/bundles/uuid/contents.tar.gz"
+        - a single file that is stored within a subpath of an archive file on Blob Storage: "azfs://storageclwsdev0/bundles/uuid/contents.tar.gz/file1"
+
+        Returns a LinkedBundlePath instance to encode this information.
+    """
+    if url.startswith("azfs://"):
```
teetone: Can we turn this into a constant?

teetone (approved):
Looks good! Your comments made it easy to understand and review.

codalab/common.py
epicfaace: Done.

PR #3304, opened 2021-02-26
Blob Storage: update requirements, add new args to codalab-service

epicfaace:

- Update requirements to upgrade Apache Beam and use a forked version of ratarmount (used to read / index .tar.gz files). I needed to add `git` to the worker so that it could download pip packages from git.
- Add new arguments to codalab-service: `azure_blob_connection_string` and `always_use_azure_blob_beta`


requirements.txt
```diff
@@ -1,4 +1,4 @@
-apache-beam==2.25.0
+apache-beam[azure]==2.27.0
```
teetone, line 1: Just curious, what does `[azure]` do?

codalab_service.py
```diff
@@ -353,6 +353,11 @@ def has_callable_default(self):
         help='Comma-separated list of directories that are mounted on the REST server, allowing their contents to be used in the --link argument.',
         default='/tmp/codalab/link-mounts',
     ),
+    CodalabArg(name='azure_blob_connection_string', help='Azure Blob storage connection string'),
+    CodalabArg(
```
teetone, line 357: Why do we need this new argument?

teetone (approved):

requirements.txt
epicfaace: It adds some Azure Blob Storage dependencies. Functionally, I don't think it makes a difference, since we already have `azure-storage-blob` and `azure-core` in our dependencies. If Beam ends up adding additional dependencies to use Azure (or requires different versions of the Azure client) in the future, though, that would be useful. Additionally, once we support something like Google Cloud Storage, we can just say `apache-beam[azure,gcp]` to pull down all the required dependencies for GCP as well.

codalab_service.py
epicfaace: I use it in GitHub Actions tests so that everything (including all uploads from the workers) use Azure Blob Storage (these tests will come up in a future PR once all the functionality for Azure Blob Storage is in).

PR #3305, opened 2021-02-26
Blob Storage: Add --use-azure-blob-beta argument to bundle CLI

epicfaace:

- Add the `--use-azure-blob-beta` (or `-a`) argument to the `cl upload` command in the bundle CLI. This argument currently doesn't do anything, but it's simply passed on to `upload_to_bundle_store` in the REST server.
- Move the logic for skipping bundle uploads for linked bundles out of `upload_to_bundle_store` and into `bundles.py` instead.


adiprerepa:
curious: why should the user care about if a bundle is being uploaded to blob storage or not?

epicfaace:
The user doesn't care -- this is just meant as a feature flag to be deployed to production so we can gradually roll out the blob storage uploads. Eventually, we'll remove the flag.

codalab/rest/bundles.py
```diff
@@ -747,7 +753,12 @@ def _update_bundle_contents_blob(uuid):
         if request.query.filename:
             filename = request.query.get('filename', default='contents')
             sources = [(filename, request['wsgi.input'])]
-        if sources:
+        bundle_link_url = getattr(bundle.metadata, "link_url", None)
```
yipenghe, line 756: why do we move this check into this function?

yipenghe:
> The user doesn't care -- this is just meant as a feature flag to be deployed to production so we can gradually roll out the blob storage uploads. Eventually, we'll remove the flag.

Might be good to have a master issue tracking all the beta code that needs clean up latter

codalab/rest/bundles.py
epicfaace: It seems to semantically make more sense -- why would we call `upload_to_bundle_store` if we just end up not doing anything in that function linked bundles? Better to just not call `upload_to_bundle_store` in the first place.

Later, I also add more complexity to `upload_to_bundle_store` (#3308), so that's why I go ahead and move this check out of that function in this PR.

yipenghe: Makes sense, got a bit lost on the hierarchy on functions lol

yipenghe (approved):
Looks good

PR #3306, opened 2021-02-26
Blob Storage: Update MultiDiskBundleStore to use blob storage

epicfaace:

Update the MultiDiskBundleStore bundle store to use blob storage. The docstring should explain it well enough:

    A multi-disk bundle store that also supports storing bundles in a CodaLab-managed
    Blob Storage container.

    If bundles are indicated to be stored in Blob Storage, they are retrieved from Blob
    Storage. Otherwise, the bundle is retrieved from the underlying disk bundle store.

    In Blob Storage, each bundle is stored in the format:
    azfs://{container name}/bundles/{bundle uuid}/contents.tar.gz.

    If the bundle is a directory, the entire contents of the bundle is stored in the .tar.gz file;
    otherwise, if the bundle is a single file, the file is stored in the .tar.gz file as an archive
    member with name equal to the bundle uuid and is_dir is set to False in the database.

I've added a base class `BundleStore` abstraction. I've renamed the old MultiDiskBundleStore class to `_MultiDiskBundleStoreBase`, so it's just an internal class used to make the implementation of `MultiDiskBundleStore` a bit cleaner.


codalab/model/bundle_model.py
```diff
@@ -1210,7 +1210,7 @@ def get_bundle_states(self, uuids):
     def get_bundle_storage_info(self, uuid):
         result_dict = self.get_bundle_storage_infos([uuid])
         if uuid not in result_dict:
-            raise NotFoundError('Could not find bundle with uuid %s' % uuid)
```
teetone, line 1213: Should at least log something?

teetone (approved):

codalab/lib/bundle_store.py
```diff
@@ -368,3 +385,32 @@ def _check_other_paths(other_paths, db_bundle_by_uuid):
                 '\tBundles that need data_hash recompute: %d' % data_hash_recomputed,
                 file=sys.stderr,
             )
+
+
+class MultiDiskBundleStoreWithBlobStorage(MultiDiskBundleStore):
```
percyliang: I suggest we get `Azure` into the name and remove `MultiDisk`.

codalab/lib/codalab_manager.py
```diff
@@ -247,9 +248,13 @@ def bundle_store(self):
         depends on what the user has configured, but if no bundle store is configured manually then it defaults to a
         MultiDiskBundleStore.
         """
-        store_type = self.config.get('bundle_store', 'MultiDiskBundleStore')
+        store_type = self.config.get('bundle_store', 'MultiDiskBundleStoreWithBlobStorage')
```
percyliang: Not sure we should change this default...by default, people spinning up an instance will likely use the MultiDiskBundleStore.

epicfaace: Ok, yeah, this is one abstraction I wasn't sure if we needed. Do you think we really need a separate `MultiDiskBundleStoreWithBlobStorage` class, or do you think it would be best to just modify the existing `MultiDiskBundleStore` to also check Blob Storage first?

Also, to clarify, the behavior of `MultiDiskBundleStoreWithBlobStorage` currently is that it will check if the bundle uses Azure first and then fall back to the multi-disk bundle store.

codalab/model/bundle_model.py
epicfaace: No, because this would happen pretty often. Any bundle which was uploaded before the introduction of Blob Storage will trigger this code path, because `is_dir` and `storage_type` wouldn't be set.

codalab/lib/codalab_manager.py
percyliang: FWIW, we had abstractions in the past (SingleDisk and MultiDisk) and we just removed it because it was kind of annoying.

epicfaace:
Okay, I've changed this PR to just build upon `MultiDiskBundleStore` instead, so we don't have to bother with making the user have to configure different kinds of bundle stores. Can one of you take a look and let me know if this seems okay?

Issue #3298, opened 2021-02-26
Azure Blob Storage megaissue - PR breakdown


epicfaace:
I'm breaking down the various PRs involved with Azure Blob Storage in this megaissue (this is all for solving issue #3112).

## Entire PR
https://github.com/codalab/codalab-worksheets/pull/3198 - don't review this, but it may be good to use as a reference, as it includes the state of the codebase when all the individual PRs are combined.

## PRs that are ready to be reviewed
These PRs split the work from the above PR and are ready to be reviewed:
- [x] Add bundle `storage_type` and `is_dir` columns to the database - #3300
- [x] Update requirements, add new args to codalab-service - #3304
- [x] Add `--use-azure-blob-beta` argument to bundle CLI - #3305
- [x] Add the `MultiDiskBundleStoreWithBlobStorage` bundle store - #3306
- [ ] Add download_util.py changes to work with downloading / reading from Azure Blob Storage - #3309

## PRs that should be reviewed in the future
These PRs are in draft status and depend on other PRs from above, so they should not be reviewed until the ones above are merged:

- [ ] Update upload and download logic to work with Azure Blob Storage - #3308

epicfaace:
Closing this in favor of the other blob storage issues; most of the PRs in this issue (except for #3308) have already been merged.

PR #3309, opened 2021-02-26
Blob Storage: add download_util changes

epicfaace:

Add download_util changes (and related file_util changes) needed for opening, previewing, and gzipping files that are located in Blob Storage.

See the megaissue for full context of where this PR fits in the Azure Blob PRs: https://github.com/codalab/codalab-worksheets/issues/3298


setup.cfg
```diff
@@ -46,7 +46,7 @@ exclude = venv/*,var/*,codalab/lib/beam/*
 # F401 - module imported but not used
 # E501 - line too long, exceeds max-line-length
 
-ignore = E203,E731,W503,W605
+ignore = E203,E731,W503,W605,E231
```
epicfaace, line 49: E231 is a conflict between black and flake8 (regarding whitespace after commas), so I'm ignoring it in flake8 so that black prevails.

epicfaace:
Ready for review @jzwang43 @teetone @adiprerepa !

codalab/worker/download_util.py
```diff
@@ -110,7 +139,8 @@ def _get_target_path(bundle_path, path):
         return bundle_path
 
 
-def _compute_target_info(path, depth):
+def _compute_target_info_local(path, depth):
```
adiprerepa: curious: why the underscore? is it internal or something?

adiprerepa (commented):
I don't have any outstanding issues with this but I suspect that's due to my relative naivete with the codebase, will yield for an experienced member to approve.

codalab/worker/download_util.py
epicfaace: Yeah, it's internal.

codalab/server/bundle_manager.py
```diff
@@ -213,9 +213,15 @@ def _make_bundle(self, bundle):
...
                 parent_bundle_link_url = parent_bundle_link_urls.get(dep.parent_uuid)
-                parent_bundle_path = parent_bundle_link_url or os.path.normpath(
-                    self._bundle_store.get_bundle_location(dep.parent_uuid)
-                )
+                try:
+                    parent_bundle_path = parent_bundle_link_url or os.path.normpath(
+                        self._bundle_store.get_bundle_location(dep.parent_uuid)
+                    )
+                except NotFoundError:
+                    raise Exception(
```
percyliang, line 221: Is there a more specific type of Exception to raise?  Basically, is this treated as a user error or a system error?

codalab/worker/download_util.py
```diff
@@ -59,15 +67,28 @@ def get_target_info(bundle_path, target, depth):
     If reading the given path is not secure, raises a PathException.
     """
     final_path = _get_normalized_target_path(bundle_path, target)
+    if parse_linked_bundle_url(final_path).uses_beam:
+        # If the target is on Azure, use a special method using Apache Beam
```
percyliang: Comment shouldn't be specific to Azure

codalab/worker/download_util.py
```diff
@@ -91,8 +112,16 @@ def get_target_path(bundle_path, target):
 
 
 def _get_normalized_target_path(bundle_path, target):
```
percyliang: Add type annotations along the way

codalab/worker/download_util.py
```diff
@@ -125,9 +155,100 @@ def _compute_target_info(path, depth):
...
-                _compute_target_info(os.path.join(path, file_name), depth - 1)
+                _compute_target_info_local(os.path.join(path, file_name), depth - 1)
                 for file_name in os.listdir(path)
             ]
     if result is None:
         raise PathException()
     return result
+
+
+def _compute_target_info_beam(path, depth):
```
percyliang: Type annotations

codalab/worker/download_util.py
```diff
@@ -125,9 +155,100 @@ def _compute_target_info(path, depth):
...
+    if not FileSystems.exists(linked_bundle_path.bundle_path):
+        raise PathException(linked_bundle_path.bundle_path)
+    if not linked_bundle_path.is_archive:
+        # Single file
+        file = FileSystems.match([path])[0].metadata_list[0]
+        return {
+            'name': linked_bundle_path.bundle_uuid,
+            'type': 'file',
+            'size': file.size_in_bytes,
+            'perm': 0o755,
```
percyliang: We're just making up these permissions?  Shouldn't this come from the actual bundle?

codalab/worker/download_util.py
```diff
@@ -125,9 +155,100 @@ def _compute_target_info(path, depth):
...
+
+    tf, _ = open_indexed_tar_gz_file(linked_bundle_path.bundle_path)
+    islink = lambda finfo: stat.S_ISLNK(finfo.mode)
+    readlink = lambda finfo: finfo.linkname
+
+    isfile = lambda finfo: finfo.type in tarfile.REGULAR_TYPES
+    isdir = lambda finfo: finfo.type == tarfile.DIRTYPE
+    listdir = lambda path: tf.getFileInfo(path, listDir=True)
+
+    def _get_info(path, depth):
```
percyliang: Provide some higher-level documentation on what the general strategy is

codalab/worker/download_util.py
```diff
@@ -125,9 +155,100 @@ def _compute_target_info(path, depth):
...
+        if not path.startswith("/"):
+            path = "/" + path
+        finfo = tf.getFileInfo(path)
+        if finfo is None:
+            # Not found
+            raise PathException
+        result = {}
+        result['name'] = path.split("/")[-1]  # get last part of path
```
percyliang: `os.path.basename`?

codalab/worker/download_util.py
```diff
@@ -125,9 +155,100 @@ def _compute_target_info(path, depth):
...
+        # Return the contents of a subpath within a directory.
+        return _get_info(linked_bundle_path.archive_subpath, depth)
+    else:
+        # No subpath, return the entire directory.
+        file = FileSystems.match([path])[0].metadata_list[0]
+        result = {
+            'name': linked_bundle_path.bundle_uuid,
+            'type': 'directory',
+            'size': file.size_in_bytes,
+            'perm': 0o755,
```
percyliang: Same as above...permissions should come from bundle?

codalab/worker/download_util.py
epicfaace: This code path is actually never reached, because everything on Azure Blob Storage is stored as an archive. I'll raise an exception instead.

codalab/worker/download_util.py
epicfaace: This code path is only called when someone gets the target info of the _entire bundle_. The _entire bundle_ technically doesn't have a file permission -- only the constituent files within the bundle do.

codalab/worker/download_util.py
```diff
@@ -125,9 +155,100 @@ def _compute_target_info(path, depth):
...
+        }
+        if depth > 0:
+            result['contents'] = [
+                _get_info(file_name, depth - 1) for file_name in listdir("/") if file_name != "."
+            ]
+        return result
+
+
+def compute_target_info_beam_descendants_flat(path):
```
percyliang: type annotations

codalab/worker/file_util.py
```diff
@@ -135,8 +146,7 @@ def zip_directory(
         try:
             proc = subprocess.Popen(args, stdout=subprocess.PIPE, cwd=directory_path)
             proc.wait()
-            with open(tmp_zip_name, "rb") as out:
-                return BytesIO(out.read())
+            return open(tmp_zip_name, "rb")
```
percyliang: Was there a reason that we needed to read first and not just do the simpler thing you're doing?

percyliang: Seems like these subtle changes can have a large performance effect on the system for large files, just want to understand a bit better.

codalab/worker/file_util.py
```diff
@@ -171,11 +184,192 @@ def do_unzip(filename):
         do_unzip(fileobj_or_name)
 
 
-def open_file(file_path, mode='r', compression_type=CompressionTypes.UNCOMPRESSED):
+def open_indexed_tar_gz_file(path):
```
percyliang: This seems like a very specific structure...should we build an abstraction around this instead of returning a tuple (either way, should type annotate).  We might consider moving this logic into a separate file so that `file_util.py` doesn't become too cluttered.

codalab/worker/file_util.py
```diff
@@ -171,11 +184,192 @@ def do_unzip(filename):
...
+        return self._buffer.write(*args, **kwargs)
+
+    def __del__(self, *args, **kwargs):
+        self._cleanup_fn(*self._cleanup_args)
+        return self._buffer.__del__(*args, **kwargs)
+
+
+def open_file(file_path, mode='r'):
+    """
+    Opens the file indicated by the given file path. Can be in a directory.
```
percyliang: Elaborate on this a lot more - it's not just files and directories....

codalab/worker/file_util.py
```diff
@@ -171,11 +184,192 @@ def do_unzip(filename):
...
+    """
+    linked_bundle_path = parse_linked_bundle_url(file_path)
+    if (
+        linked_bundle_path.uses_beam
+        and linked_bundle_path.is_archive
+        and linked_bundle_path.archive_subpath
+    ):
+        # If file path is a .tar.gz file on Azure, open the specified path within the archive.
```
percyliang: Not specific to Azure

codalab/worker/file_util.py
```diff
@@ -171,11 +184,192 @@ def do_unzip(filename):
...
+        isdir = lambda finfo: finfo.type == tarfile.DIRTYPE
+
+        fpath = "/" + linked_bundle_path.archive_subpath
+        finfo = tf.getFileInfo(fpath)
+        if finfo is None:
+            raise FileNotFoundError(fpath)
+        if isdir(finfo):
+            from codalab.worker.download_util import compute_target_info_beam_descendants_flat
+
+            # If streaming a folder within an Azure bundle, we need to download its contents,
```
percyliang: This is potentially a big bottleneck if this is happening on the server...should be able to stream it somehow, right?

codalab/worker/file_util.py
```diff
@@ -171,11 +184,192 @@ def do_unzip(filename):
...
+        # If file path is a .tar.gz file on Azure, open the specified path within the archive.
+        tf, _ = open_indexed_tar_gz_file(linked_bundle_path.bundle_path)
+
+        isdir = lambda finfo: finfo.type == tarfile.DIRTYPE
+
+        fpath = "/" + linked_bundle_path.archive_subpath
+        finfo = tf.getFileInfo(fpath)
+        if finfo is None:
+            raise FileNotFoundError(fpath)
+        if isdir(finfo):
```
percyliang: Would be good to somehow separate this logic into a separate function, since this function is getting pretty long...

codalab/worker/file_util.py
```diff
@@ -171,11 +184,192 @@ def do_unzip(filename):
...
+            return ClosingStreamWrapper(
+                tar_gzip_directory(extracted_path), lambda t: t.cleanup(), (tmp_dir,),
+            )
+        else:
+            # TODO: Implement a tf.open() function so that we don't have to read the entire file.
+            return BytesIO(tf.read(path="", fileInfo=finfo, size=finfo.size, offset=0))
+    return FileSystems.open(file_path, mode, compression_type=CompressionTypes.UNCOMPRESSED)
+
+
+class GzipStream:
```
percyliang: I thought the Python version of this is not performant and that's why we call out to `gzip`?

codalab/worker/file_util.py
epicfaace: No, I don't think there was a good reason for doing that earlier. However, it looks like `zip_directory` is not actually used in the codalab codebase (I think I had added it initially when we were using .zip files for Blob Storage -- but it's just used in this test now: https://github.com/codalab/codalab-worksheets/blob/43eaddc8013d484fdfdde91eb07dc6eaf7902601/tests/unit/worker/file_util_test.py#L138).

codalab/worker/file_util.py
epicfaace: Yes -- but we can't easily use the command line `gzip` to read from a Python file-like object (earlier we would just be passing a file path to `gzip`). See comment on line 383:

```
        # We run gzip in the same process if the file is on Azure, so that
        # we can use Apache Beam methods to read the file. This may cause
        # performance issues once we switch all files to use Azure, but
        # they should be mitigated when we allow users to connect to Azure
        # servers directly to download files.
```

Potentially it might be possible to read from the Python file-like object and pipe that into stdin to the `gzip` process, instead -- do you think that might be worth trying out?

codalab/server/bundle_manager.py
epicfaace: This is a system error, but I'm not sure what kind of exception to raise here. I'm just following the convention used elsewhere in this file of raising `Exception` classes on error -- no other types of exceptions are raised here. We can probably refactor them into system-error-specific Exception subclasses in a future PR.

codalab/worker/file_util.py
epicfaace: I found a better way to do this -- the tuple was initially created to work around a bug where the file kept getting closed. Now, I use a context manager that just returns the SQLiteIndexedTar file object itself.

codalab/worker/file_util.py
epicfaace: I think I'm going to keep the Azure references in the comments for now. We don't support anything other than Azure, and I can't think of a good term that's general enough for now, so I think "Azure" is the clearest way to denote the concept we're trying to convey here.

Once we support another storage backend (such as GCP), hopefully then we'll be able to find a more general name (maybe something related to "blob"?), and it would make sense to also substitute the instances of "Azure" with the general name.

codalab/worker/download_util.py
epicfaace: See my message on Azure below

codalab/worker/download_util.py
epicfaace: Done

codalab/worker/download_util.py
epicfaace: Done

codalab/worker/download_util.py
epicfaace: Done

codalab/worker/download_util.py
epicfaace: Done

codalab/worker/download_util.py
epicfaace: Done

codalab/worker/download_util.py
epicfaace: So 0o755 is just a placeholder in some sense, because we can't really store the permission of the entire directory (the directory with name equal to `[bundle id]`) on blob storage. However, in CodaLab, we never care about the actual permission of this directory, as we only care about what is in the actual bundle itself.

Adding a comment to clarify.

epicfaace: Looking at the analogous situation for disk-based storage -- the permission of the bundle directory isn't configurable by the user, either. It's just a constant value determined by the umask setting of the system. That's why hardcoding the permission to 0o755 here would accomplish equivalent behavior.

codalab/worker/download_util.py
epicfaace: done

codalab/worker/file_util.py
epicfaace: Moved to tar_subdir_stream.py

codalab/worker/file_util.py
epicfaace: For simplicity, I'm just going to keep using GzipStream in this PR. We can think about doing what I mentioned (pipe into stdin for `gzip`) in another PR as that would add additional complexity to everything.

codalab/worker/file_util.py
epicfaace: Okay, I was able to stream this (see https://github.com/codalab/codalab-worksheets/blob/beam-du/codalab/worker/tar_subdir_stream.py and https://github.com/codalab/codalab-worksheets/blob/beam-du/codalab/worker/tar_file_stream.py) -- it's a bit complicated but is working well.

epicfaace:
@percyliang @teetone  this should be ready for another review now.

jzwang43 (approved):
Looks good. On a side note, I think we might need more QA and manual testing once everything is finished and merged in.

codalab/worker/download_util.py
```diff
@@ -125,9 +175,118 @@ def _compute_target_info(path, depth):
...
+    [{name, type, size, perm}], where `name` is equal to the full path of each item.
+
+    Also includes an entry for the specified directory with `name` equal to an empty string.
+
+    This function is used by TarSubdirStream in order to determine the list of descendants
+    that exist inside a given subdirectory in a .tar.gz file on Blob Storage.
+    """
+    target_info = _compute_target_info_beam(
+        path, math.inf
+    )  # We want to calculate the target info at infinite depth to expand *all* descendants.
```
percyliang: Couldn't this be very slow?  Might be safer to always have a depth bound (was this the case before?)  For example, what if you have cyclic symlinks?

epicfaace: This function (`compute_target_info_beam_descendants_flat`) is used only when streaming a subdirectory within a tar file on Blob Storage to get the initial list of files to combine to create the new .tar file to be streamed (see `TarSubdirStream`). So, normally, when computing target info from the regular REST endpoints, there's still a depth bound as usual.

Also, it doesn't resolve symlinks.

The only real slowdown that would happen is if there are too many files within a subdirectory, it would block a bit before starting to stream the file. We could eventually stop this blocking problem by changing this function to return a generator that yields the list of all descendants as well, but perhaps we could do that later once we first see the performance implications to see if it's worth that additional complexity.

codalab/worker/download_util.py
```diff
@@ -125,9 +175,118 @@ def _compute_target_info(path, depth):
...
+    """Computes target info for a file that is externalized on a location
+    such as Azure, by using the Apache Beam FileSystem APIs."""
+
+    linked_bundle_path = parse_linked_bundle_url(path)
+    if not FileSystems.exists(linked_bundle_path.bundle_path):
+        raise PathException(linked_bundle_path.bundle_path)
+    if not linked_bundle_path.is_archive:
+        # Single file
+        raise PathException(
+            "Single files on Blob Storage are not supported; only a path within a .tar.gz fils is supported."
```
adiprerepa: nit: `fils` -> `file`?

codalab/worker/download_util.py
```diff
@@ -125,9 +175,118 @@ def _compute_target_info(path, depth):
...
+        isdir = lambda finfo: finfo.type == tarfile.DIRTYPE
+        listdir = lambda path: tf.getFileInfo(path, listDir=True)
+
+        def _get_info(path: str, depth: Union[int, float]) -> TargetInfo:
+            """This function is called to get the target info of the specified path.
+            If the specified path is a directory and additional depth is requested, this
+            function is recursively called to retrieve the target info of files within
+            the directory, much like _compute_target_info_local.
+            """
+            if not path.startswith("/"):
```
adiprerepa, line 232: can these paths be relative?

if path is `foo.txt`, it would look for the file in the root directory?

codalab/worker/file_util.py
```diff
@@ -336,6 +469,15 @@ def get_path_size(path, exclude_names=[], ignore_nonexistent_path=False):
...
     0 is returned. Else, an exception is raised (FileNotFoundError).
     """
+    if parse_linked_bundle_url(path).uses_beam:
+        # On Azure, use Apache Beam methods, not native os methods,
+        # to get the path size.
+
+        # Get the size of the specified path (file / directory).
+        # This will only get the right size of files, not of directories (but we don't
+        # store any bundles as directories on Azure).
+        return get_file_size(path)
```
adiprerepa, line 477: Would it make sense to reuse the `get_path_size` function in `codalab.worker.file_util`?

codalab/worker/tar_file_stream.py
```diff
@@ -0,0 +1,75 @@
...
+
+    The general idea is that whenever .read() is called on this class,
+    it will read the specified number of bytes through ratarmount's tf.open()
+    API on the associated file and return those bytes.
+
+    TODO (Ashwin): If we can add tf.open() support upstream to the ratarmount API
+    (right now it only supports tf.read()), we may not have a need for this class anymore.
+    """
+
+    BUFFER_SIZE = 100 * 1024 * 1024  # Read in chunks of 100MB
```
adiprerepa: how did you arrive at this value?

adiprerepa (approved):
lgtm with nits

codalab/worker/file_util.py
epicfaace: No, this code is within the `get_path_size` function in `codalab.worker.file_util`, so we can't do that.

codalab/worker/tar_file_stream.py
epicfaace: I just used the same buffer size in the other stream class we had, `GzipStream`, which @nelson-liu initially added in https://github.com/codalab/codalab-worksheets/commit/df9a7fbb783f3ae4c5b63f4fff94e33e06a308c3. The tradeoffs here are:
- if BUFFER_SIZE is too small, the Python code may add too much overhead and slow down streaming a file if we're trying to stream a large amount at a time.
- if BUFFER_SIZE is too large, then if we just try to stream a small amount (like a couple of bytes), we'll add more overhead when trying to stream just a few bytes at a time.

Now that you mention it, though, maybe a smaller value would be better. I'll do some testing to see what the optimal value would be.

codalab/worker/download_util.py
epicfaace:
```suggestion
            "Single files on Blob Storage are not supported; only a path within a .tar.gz filseis supported."
```

epicfaace:
```suggestion
            "Single files on Blob Storage are not supported; only a path within a .tar.gz file is supported."
```

codalab/worker/download_util.py
epicfaace: Due to the way ratarmount parses paths, `foo.txt` would return no results, but `/foo.txt` refers to a relative path that would look for the file in the root directory (within the tar file, of course).

codalab/worker/tar_file_stream.py
epicfaace: I talked with @nelson-liu about the design of the buffer size and we came to the conclusion that it's not really necessary; it adds too much overhead by reading too much from the file if we only want to read a small number of bytes. The concern I mentioned about BUFFER_SIZE being too small doesn't seem to be a problem because when a file is streamed from a REST endpoint, it's read 8192 bytes at a time.

I removed it from this PR.

codalab/worker/download_util.py
epicfaace: @percyliang I ended up changing `compute_target_info_blob_descendants_flat` to return a generator instead of a list of all descendants, so listing the contents of a directory for use in `TarSubDirStream` should never block -- I believe this should alleviate your concern.

PR #3315, opened 2021-02-27
Fix schema change detection on text field

### LINKED ISSUE #3316: [BUG] Changing text field in schema item doesn't enable the save and revert button
Currently, when you only change the text field in the schema item, you cannot save or revert. You can only do that if you also changed the schema name. This is because of a bug introduced when trying to fix a eslint6 warning in #2912 . The fix in the PR made the textFieldChanged variable always false as long as you don't change all rows at the same time.

To reproduce:
Open any schema item/ create a schema item that has at least two rows. Change one field on one of the rows, the save and revert button isn't enabled until you change at least one text field on each row

yipenghe:

Currently, when you only change the text field in the schema item, you cannot save or revert. You can only do that if you also changed the schema name. This is because of a bug introduced when trying to fix a eslint6 warning in https://github.com/codalab/codalab-worksheets/pull/2912 . The fix in the PR made the textFieldChanged variable always false as long as you don't change all rows at the same time.

The correct way to fix the warning is to use the traditional loop https://stackoverflow.com/questions/58816244/debugging-eslint-warning-function-declared-in-a-loop-contains-unsafe-reference

### Test
Open an existing schema item, change any text field but not schema name, the save and revert button are now enabled again

### Screenshots
Before changing any text field, the save and revert are disabled
Screenshot from 2021-02-27 15-09-31

After changing the text field, the save and revert are enabled
Screenshot from 2021-02-27 15-09-45

### Checklist


yipenghe:
The bug was in production for since the PR was merged since Oct 2020, probably would need to add a UI test. Also the component looks a bit weird as the icon is overflowing the boundary

yipenghe:
Weirdly, I can't reproduce this on prod....

yipenghe:
The bug has been in production since Oct, probably need to add a test in the future to detect the regression. Also flagging that the UI component looks weird on my computer, as the icon overflows the bottom boundary

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -150,17 +150,15 @@ class SchemaItem extends React.Component<{
...
         const headerKeys = this.props.item.header;
-        let textFieldChanged = false;
         for (let ind = 0; ind < originalRows.length; ind++) {
-            let candTextFieldChanged = false;
-            headerKeys.forEach((key) => {
+            for (let keyInd = 0; keyInd < headerKeys.length; keyInd++) {
+                const key = headerKeys[keyInd];
                 if (originalRows[ind][key] !== this.state.rows[ind][key]) {
-                    candTextFieldChanged = true;
+                    return true;
```
yipenghe, line 157: Changing to the traditional for loop also allows us to early return

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -150,17 +150,15 @@ class SchemaItem extends React.Component<{
...
                 }
-            });
-            textFieldChanged = candTextFieldChanged;
```
yipenghe, line 161: originally, if one row remains unchanged, the function will return false, but it should return true as long as one field has changed

yuqijin (approved):
Thanks for fixing the bug!

PR #3318, opened 2021-02-28
Enable freezing and unfreezing bundles

nelson-liu:
This PR is a branch of #3295 , so probably makes most sense to go look at that first.

This PR adds `--freeze` and `--unfreeze` to bundles. Bundles can only be frozen if they're in a final state. When frozen, bundles are truly immutable---their metadata cannot be changed, and they can't be removed unless `cl rm -f` is used.

The intention is that after you publish an executable paper, you will (hopefully) never have to change it again, so freezing the bundles (and the worksheet, see #3295 ) prevents any accidental modification in the future. If you do indeed need to modify, it's easy enough to unfreeze, make the changes, and then freeze again.


epicfaace:
Converting this to a draft until https://github.com/codalab/codalab-worksheets/pull/3295 is merged.

codalab/lib/bundle_cli.py
```diff
@@ -1812,6 +1812,16 @@ def do_run_command(self, args):
                 '-T', '--tags', help='Change tags (must appear after worksheet_spec).', nargs='*'
             ),
             Commands.Argument('-d', '--description', help='New bundle description.'),
+            Commands.Argument(
+                '--freeze',
+                help='Freeze bundle to prevent future metadata modification.',
```
epicfaace, line 1817: I feel like "freeze" is associated with "preventing content modification" (i.e., for worksheets), not really "metadata modification", so I'm a bit wary of overloading the term with this PR. But maybe that's okay -- what do you think @nelson-liu ?

epicfaace (approved):
Looks good -- can you fix the tests? Then we can merge.

epicfaace:
@nelson-liu just bumping this 🙂

nelson-liu:
Oops. sorry for missing this! I'll fix up the tests.

codalab/lib/bundle_cli.py
nelson-liu: that's a good point, i'm happy to change it to something else but I'm not really sure what the right word is.

alembic/versions/2021022802_add_frozen_field_to_bundle_6c013a88862f.py
```diff
@@ -0,0 +1,22 @@
+"""Add frozen field to bundle.
+
+Revision ID: 6c013a88862f
+Revises: db12798a7cf6
```
nelson-liu:
```suggestion
Revises: e24a53ae9241
```

alembic/versions/2021022802_add_frozen_field_to_bundle_6c013a88862f.py
```diff
@@ -0,0 +1,22 @@
...
+Create Date: 2021-02-28 02:31:07.648416
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '6c013a88862f'
+down_revision = 'db12798a7cf6'
```
nelson-liu:
```suggestion
down_revision = 'e24a53ae9241'
```

codalab/lib/bundle_cli.py
epicfaace: Yeah, I think we're okay with just using "freeze", so let's just keep it.

PR #3323, opened 2021-03-03
Implement Image UI button & Fix after_sort_key bug

### LINKED ISSUE #3278: Add UI buttons for record block, contents, images

### LINKED ISSUE #3281: New Run Inserted to wrong place
**Describe the bug**
- Cursor to the end of the worksheet (shift-G) 
- Add a run (a r)

The bundle gets inserted at the top rather than at the end:
https://worksheets-dev.codalab.org/worksheets/0x2a0a3904a3e04840823aa9c7dd44f7e1

yuqijin:

**I. Fix after_sort_key bug**
- default after_sort_key should always be -1
- Fix bug when setting after_sort_key for the items

**II. Implement Image UI button**
- User can directly upload one image by using the button

### Related issues

fixes #3281 
fixes #3278 

### Screenshots

ezgif com-gif-maker (2)

### Checklist


frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -287,6 +289,7 @@ class NewUpload extends React.Component<{
         return (
             <React.Fragment>
                 <input
+                    key={this.props.after_sort_key} // force the NewUpload to update when after_sort_key changes
```
epicfaace: What is this change for?

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -526,7 +526,7 @@ class Worksheet extends React.Component {
         } else {
             // If no location for the insertion is specified,
             // insert the new item at the top of the worksheet in default.
-            actualData['after_sort_key'] = -1;
+            actualData['after_sort_key'] = 0;
```
epicfaace: Why change this from -1 to 0?

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
yuqijin: without setting the key, previously the props.after_sort_key passed to the NewUpload won't change even if the focused item is changed, which results in the new upload bundles might be inserted into the wrong place.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: use -1 instead

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -251,6 +254,7 @@ class WorksheetItemList extends React.Component {
                         onAsyncItemLoad: (item) => this.props.onAsyncItemLoad(index, item),
                         updateSchemaItem: this.props.updateSchemaItem,
                         setDeleteSchemaItemCallback: this.props.setDeleteSchemaItemCallback,
+                        addImageDispay: this.props.addImageDispay,
```
epicfaace:
```suggestion
                        addImageDisplay: this.props.addImageDisplay,
```

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -128,7 +128,9 @@ class NewUpload extends React.Component<{
         let promises = [...files].map(async file => {
             const createBundleData = getDefaultBundleMetadata(name || file.name, description);
             let url = `/rest/bundles?worksheet=${ worksheetUUID }`;
-            if (after_sort_key) {
+            // if after_sort_key == 0, we still need to pass it to the backend
+            // otherwise the backend will treat the after_sort_key as None and then set the sort_key to be NULL
```
epicfaace: Can we replace this entire comment with something shorter,

```
// after_sort_key can be equal to 0
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1626,6 +1651,38 @@ class Worksheet extends React.Component {
         this.setState({ uploadAnchor: e.currentTarget });
     };
 
+    showNewImage = (e) => {
```
epicfaace: Can we rename this? It seems like this maybe should be called `uploadNewImage` instead.

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -325,6 +326,17 @@ class WorksheetItemList extends React.Component {
                         // the file hasn't changed)
                         onUploadFinish={(e) => this.setState({ newUploadKey: Math.random() + '' })}
                     />
+                    <ImageEditor
+                        key={this.state.newUploadKey + 1}
+                        after_sort_key={getAfterSortKey(focusedItem, this.props.subFocusIndex)}
+                        worksheetUUID={info.uuid}
+                        reloadWorksheet={this.props.reloadWorksheet}
+                        onUploadFinish={(e) => this.setState({ newUploadKey: Math.random() + '' })}
+                        addImageDisplay={this.props.addImageDisplay}
```
epicfaace: addimagedisplay isn't used in the ImageEditor prop. Can you check for unused props and remove them?

epicfaace (changes requested):

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
yuqijin: solved.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: solved

frontend/src/components/worksheets/WorksheetItemList.js
yuqijin: thanks for pointing it out. solved.

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -287,6 +288,7 @@ class NewUpload extends React.Component<{
         return (
             <React.Fragment>
                 <input
+                    key={this.props.after_sort_key} // force the NewUpload to update when after_sort_key changes
```
epicfaace:
```suggestion
                    key={this.props.after_sort_key} // force the NewUpload to update when the focused item changes

```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1454,6 +1455,10 @@ class Worksheet extends React.Component {
                             items[focus].mode !== 'table_block'
                         ) {
                             this.setFocus(focus >= 0 ? focus + 1 : 'end', 0);
+                        } else if (this.state.focusIndex === -1) {
+                            // if currently the virtual item is focused and a new bundle has been uploaded to the worksheet
```
epicfaace: What does "the virtual item" mean? Maybe we can clarify this comment a bit.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1507,11 +1512,31 @@ class Worksheet extends React.Component {
...
-                    });
+                    this.setState(
+                        {
+                            updating: false,
+                            version: this.state.version + 1,
+                            numOfBundles: numOfBundles,
+                        },
+                        () => {
+                            if (addImage) {
+                                const subFocusIndex = this.state.subFocusIndex
```
epicfaace: Let's simplify this to
```suggestion
                                const subFocusIndex = this.state.subFocusIndex || 0;
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1507,11 +1512,31 @@ class Worksheet extends React.Component {
...
+                                    ? this.state.subFocusIndex
+                                    : 0;
+                                let focusIndexPair = this.state.focusIndex + ',' + subFocusIndex;
+                                let index = this.state.ws.info.block_to_raw[focusIndexPair];
+                                if (index === undefined) {
+                                    // the newly uploaded image currently does not create a new block
```
epicfaace: Can you update this comment to be clearer? When does "the newly uploaded image currently does not create a new block"? Does this always happen or only if `index === undefined`?

frontend/src/components/worksheets/items/ActionButtons.js
```diff
@@ -182,6 +185,20 @@ class ActionButtons extends React.Component<{
                         Schema
                     </Button>
                 </Tooltip>
+                <Tooltip title='Add an image'>
+                    <Button
+                        size='small'
+                        color='inherit'
+                        aria-label='image'
+                        onClick={uploadNewImage}
```
epicfaace: I don't see the point in passing `uploadNewImage` as a prop from a parent many levels above to this component, given that its implementation just calls a js function that isn't dependent on state (`document.querySelector('label[for=codalab-image-upload-input]').click();
`). Could you just define `uploadNewImage` within this component instead? That would simplify it.

frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -347,7 +347,7 @@ class BundleRow extends Component {
                         <TableCell colSpan='100%' classes={{ root: classes.insertPanel }}>
                             <div className={classes.insertBox}>
                                 <NewRun
-                                    after_sort_key={this.props.after_sort_key}
+                                    after_sort_key={this.props.after_sort_key + 1} // insert the new run below the bundle row, so add 1 to after_sort_key
```
epicfaace: Can you move this comment to the "Rerun" comment above (line 325), so it's clear that this comment applies to this entire section?

frontend/src/util/worksheet_utils.js
```diff
@@ -301,7 +301,13 @@ export function createHandleRedirectFn(worksheetUuid) {
 // Return the sort key at index subFocusIndex, if subFocusIndex is defined.
 // Otherwise, return the largest sort_key.
 export function getAfterSortKey(item, subFocusIndex) {
-    if (!item) return 0;
+    // the default after_sort_key is -1 when inserting an item on top of the worksheet
```
epicfaace:
```suggestion
    // The default after_sort_key is -1 when inserting an item on top of the worksheet,
```

frontend/src/util/worksheet_utils.js
```diff
@@ -301,7 +301,13 @@ export function createHandleRedirectFn(worksheetUuid) {
...
+    // so the items' sort_key should always >= 0 (sort_key > after_sort_key)
```
epicfaace:
```suggestion
    // so the item's sort_key should always be >= 0 (sort_key > after_sort_key)
```

frontend/src/util/worksheet_utils.js
```diff
@@ -301,7 +301,13 @@ export function createHandleRedirectFn(worksheetUuid) {
...
+    if (!item) return -1;
+    if (item.mode === 'image_block') {
+        // image_block has different format with other blocks
```
epicfaace:
```suggestion
        // image_block stores the sort_key in a different property than do other blocks
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: change to the `top of the worksheet`

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: comment updated. e.g. be added as a bundle row to a table; so `does not create a new block` <==> `index === undefined`

frontend/src/components/worksheets/items/TableItem/BundleRow.js
yuqijin: updated.

frontend/src/components/worksheets/items/ActionButtons.js
yuqijin: thanks for the suggestion! Updated.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1456,7 +1456,7 @@ class Worksheet extends React.Component {
                         ) {
                             this.setFocus(focus >= 0 ? focus + 1 : 'end', 0);
                         } else if (this.state.focusIndex === -1) {
-                            // if currently the virtual item is focused and a new bundle has been uploaded to the worksheet
+                            // if currently the top of the worksheet is focused and a new bundle has been uploaded to the worksheet
```
```suggestion
                            // If currently the top of the worksheet is focused and a new bundle has been uploaded to the worksheet,
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1456,7 +1456,7 @@ class Worksheet extends React.Component {
...
                             // then the first item on the worksheet should be focused
```
```suggestion
                            // the first item on the worksheet should be focused.
```

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -304,7 +303,7 @@ class WorksheetItemList extends React.Component {
                                 ],
                                 header: ['field', 'generalized-path', 'post-processor'],
                                 schema_name: '',
-                                sort_keys: [-1],
+                                sort_keys: [0],
```
epicfaace, line 306:
```suggestion
                                sort_keys: [-1],
```

epicfaace (approved):

frontend/src/components/worksheets/WorksheetItemList.js
yuqijin: I think the sort_keys should still be [0]. Here it is the sort_keys, not the after_sort_key. All the sort_keys (except for the dummy item) should be >=0

epicfaace: Oh, I see, thanks for pointing that out.

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -200,7 +198,7 @@ class WorksheetItemList extends React.Component {
                     isDummyItem: true,
                     text: '',
                     mode: 'markup_block',
-                    sort_keys: [-1],
+                    sort_keys: [-1], // default sort_key is -1
```
epicfaace: Wait so why is this -1 and not 0 @yuqijin ?

yuqijin: Because it is the dummy item, which represents the top of the worksheet, and so it is an exception as I mentioned

epicfaace:
```suggestion
                    sort_keys: [-1], // the dummy item represents the top of the worksheet, so its sort key is -1
```

PR #3313, opened 2021-02-27
Add checkbox to the delete worksheet dialog to prevent accidental deletion

yipenghe:
### Reasons for making this change

Adding more friction to the delete worksheet flow on the UI side by adding a checkbox and capitalizing the worksheet term

### Related issues

fix #3312

### Test
Create a new worksheet and use the delete button on top right to delete the worksheet

### Screenshots

Not checked (default when open)
Screenshot from 2021-02-27 14-12-28

Checked:
Screenshot from 2021-02-27 14-12-41

### Checklist

* [x] I've added a screenshot of the changes, if this is a frontend change
* [x] I've added and/or updated tests, if this is a backend change
* [x] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [x] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

PR #3296, opened 2021-02-25
Add a filter of workers  by disk space

yuqijin:
### Related issues

fixes #2323 

### Screenshots

when the disk space a bundle required > the worker's disk space: not assigned to the worker
ezgif com-gif-maker (1)

### Checklist

* [x] I've added a screenshot of the changes, if this is a frontend change
* [x] I've added and/or updated tests, if this is a backend change
* [x] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [x] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

PR #3273, opened 2021-02-16
Add tutorial for uploading / downloading bundles

epicfaace:
### Reasons for making this change

Add tutorial for uploading / downloading bundles. Also upgrade mkdocs-material and add some improvements to the menu on the left (such as making it look a bit better / more usable).

### Related issues

Fixes #3269.

### Screenshots

image

PR #3324, opened 2021-03-04
Bump to version 0.5.41

yuqijin:
Bump new version

Full diff: https://github.com/codalab/codalab-worksheets/compare/v0.5.40...rc0.5.41

# Draft release notes

## Frontend
- Add checkbox to the delete worksheet dialog to prevent accidental deletion (#3313)
- Fix schema change detection on the text field (#3315)
- Fix the Editing Avatar Permission Error (#3270)

## Backend
- Add logging for common tasks (#3223)
- Add a filter of workers by disk space (#3296)
- Fix the "uploading results" error by sending data to keep the socket open (#3285)

## Dev / docs / CI
- Fix test formatting, run black on all files (#3301)
- Unit test - don't expect a fixed time_quota / disk_quota (#3302)
- Add tutorial for uploading / downloading bundles (#3273)
- Change the comment of AUTO_HIDDEN_DURATION (#3314)
- Stress test a worker uploading large bundle results (2) (#3215)

### DIFF
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -13,7 +13,7 @@ from retry import retry
 
 # Increment this on master when ready to cut a release.
 # http://semver.org/
-CODALAB_VERSION = '0.5.40'
+CODALAB_VERSION = '0.5.41'
 BINARY_PLACEHOLDER = '<binary>'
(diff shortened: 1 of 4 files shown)


epicfaace:
@yuqijin can you not include "Blob Storage: Add --use-azure-blob-beta argument to bundle CLI " in the release notes? This is only part of the actual blob storage feature, and isn't fully functional at the moment, so let's not announce it or it may confuse users.

PR #3325, opened 2021-03-04
Simplify file uploads with inbuilt chunk encoding

### LINKED ISSUE #3286: Simplify _upload_with_chunked_encoding by using python's built-in chunk encoding
We can greatly simplify _upload_with_chunked_encoding by using python's built-in chunk encoding.

See docs: https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection.request

> The encode_chunked argument is only relevant if Transfer-Encoding is specified in headers. If encode_chunked is False, the HTTPConnection object assumes that all encoding is handled by the calling code. If it is True, the body will be chunk-encoded.

> Changed in version 3.6: If neither Content-Length nor Transfer-Encoding are set in headers, file and iterable body objects are now chunk-encoded. The encode_chunked argument was added. No attempt is made to determine the Content-Length for file objects.

Thus, we can either explicitly specify the `encode_chunked` argument, or we can rely on Python 3.6's default behavior of already using chunk-encoding for normal file objects.

adiprerepa:
fixes #3286

### DIFF
--- a/codalab/worker/rest_client.py
+++ b/codalab/worker/rest_client.py
@@ -140,3 +140,29 @@ class RestClient(object):
             }
             # use in-built encoding
             conn.request(method, parsed_base_url.path + path, body=fileobj, headers=headers, encode_chunked=True)
+
+            # Read the response.
+            logging.debug("About to read the response... url: %s", url)
+
+            # Sometimes, it may take a while for the server to process
+            # the data and send the response. In this case, we want to
+            # periodically keep sending empty bytes so that the
+            # connection doesn't drop before the response is available.
+            got_response = False
+            while not got_response:
+                try:
+                    response = conn.getresponse()
+                    got_response = True
+                except socket.timeout:
+                    logging.debug("Socket timeout, retrying url: %s", url)
+                    conn.send(b'\0')
+            logging.debug("Finished reading the response, url: %s", url)
+            if response.status != 200:
+                # Low-level httplib module doesn't throw HTTPError
+                raise urllib.error.HTTPError(
+                    self._base_url + path,
+                    response.status,
+                    response.reason,
+                    dict(response.getheaders()),
+                    StringIO(response.read().decode()),
+                )
--- a/codalab/worker/rest_client.py
+++ b/codalab/worker/rest_client.py
@@ -139,7 +139,13 @@ class RestClient(object):
                 'X-Requested-With': 'XMLHttpRequest',
             }
             # use in-built encoding
-            conn.request(method, parsed_base_url.path + path, body=fileobj, headers=headers, encode_chunked=True)
+            conn.request(
+                method,
+                parsed_base_url.path + path,
+                body=fileobj,
+                headers=headers,
+                encode_chunked=True
+            )
 
             # Read the response.
             logging.debug("About to read the response... url: %s", url)
--- a/codalab/worker/rest_client.py
+++ b/codalab/worker/rest_client.py
@@ -138,7 +138,7 @@ class RestClient(object):
                 'Authorization': 'Bearer ' + self._get_access_token(),
                 'X-Requested-With': 'XMLHttpRequest',
             }
-            # use in-built encoding
+            # Enable Python's in-built chunked encoding by passing encode_chunked=True
             conn.request(
                 method,
                 parsed_base_url.path + path,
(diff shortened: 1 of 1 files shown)


codalab/worker/rest_client.py
```diff
@@ -135,55 +134,9 @@ def _upload_with_chunked_encoding(
...
-            logging.debug("About to read the response... url: %s", url)
-
-            # Sometimes, it may take a while for the server to process
-            # the data and send the response. In this case, we want to
-            # periodically keep sending empty bytes so that the
-            # connection doesn't drop before the response is available.
-            got_response = False
-            while not got_response:
-                try:
-                    response = conn.getresponse()
```
epicfaace: I think we still need to keep this response handling code

codalab/worker/rest_client.py
```diff
@@ -135,32 +134,12 @@ def _upload_with_chunked_encoding(
...
-                    break
-                conn.send(b'%X\r\n%s\r\n' % (len(to_send), to_send))
-                bytes_uploaded += len(to_send)
-                if progress_callback is not None:
-                    should_resume = progress_callback(bytes_uploaded)
-                    if not should_resume:
-                        raise Exception('Upload aborted by client')
-            conn.send(b'0\r\n\r\n')
+            # use in-built encoding
+            conn.request(method, parsed_base_url.path + path, body=fileobj, headers=headers, encode_chunked=True)
```
epicfaace: Can we add a unit test for `_upload_with_chunked_encoding` that ensures that the correct headers, for example, are sent? It could just check, for example, for the presence of the `Transfer-Encoding` header.

adiprerepa: where would this unit test be?

and how would it be set up? Would this function be called and then you would do a server-side check for the header?

epicfaace: See https://codalab-worksheets.readthedocs.io/en/latest/Code-Overview/#unit-tests for information about unit tests.

I was initially thinking you could maybe mock the http client, but I guess now that I'm thinking about it, unit tests are probably not the best way to do this, because we are trying to just ensure that http.client works properly when passing in encode_chunked.

Can you just manually test it by running `cl upload` and ensuring that chunked uploading is being used by logging http.client calls? see httpclient_logging_patch  in https://stackoverflow.com/questions/16337511/log-all-requests-from-the-python-requests-module

Or let me know if you think of an easier way to test it manually.

codalab/worker/rest_client.py
```diff
@@ -135,32 +134,18 @@ def _upload_with_chunked_encoding(
...
-                if not to_send:
-                    break
-                conn.send(b'%X\r\n%s\r\n' % (len(to_send), to_send))
-                bytes_uploaded += len(to_send)
-                if progress_callback is not None:
-                    should_resume = progress_callback(bytes_uploaded)
-                    if not should_resume:
-                        raise Exception('Upload aborted by client')
-            conn.send(b'0\r\n\r\n')
+            # use in-built encoding
```
epicfaace:
```suggestion
            # Enable Python's in-built chunked encoding by passing encode_chunked=True
```

codalab/worker/rest_client.py
```diff
@@ -135,32 +134,18 @@ def _upload_with_chunked_encoding(
...
-            # Use chunked transfer encoding to send the data through.
-            bytes_uploaded = 0
-            while True:
-                to_send = fileobj.read(CHUNK_SIZE)
-                if not to_send:
-                    break
-                conn.send(b'%X\r\n%s\r\n' % (len(to_send), to_send))
-                bytes_uploaded += len(to_send)
-                if progress_callback is not None:
-                    should_resume = progress_callback(bytes_uploaded)
```
percyliang, line 160: But now we're not invoking `progress_callback`, which is important...

adiprerepa: true - I assumed this was a tradeoff we were making for simplicity. If we need `progress_callback`, I do not think we can use python's chunk encoding. @epicfaace thoughts?

epicfaace: Actually, yeah, `progress_callback` is important and we need to keep it. I guess that means we can't use Python's chunk encoding then.

adiprerepa: Closing pr

PR #3333, opened 2021-03-09
 Add UI Button for Content Block that User Can Specify Subpath Inside a Bundle

### LINKED ISSUE #3278: Add UI buttons for record block, contents, images

yuqijin:
Duplicate the bundle for creating a separate content block.

- User can specify a subpath inside a bundle for content block
  - default path is `/`
- use different Icons for cut and copy bundle operations; 
- fix infinite after_sort_key bug.

### Related issues

fixes #3278 

### Screenshots

ezgif com-gif-maker (5)

### Checklist


percyliang:
One often wants to display contents for a subpath inside a bundle, not just a bundle...so it'd be nice to be able to support this.

epicfaace:
Looks good -- I just wonder if we want to use an alert to specify the subpath, or maybe a less intrusive method (maybe a popout, like the panel that shows up when you click the "run" button) would look better. Let's discuss this in the meeting today.

epicfaace:
Duplicate the bundle

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1677,6 +1683,75 @@ class Worksheet extends React.Component {
...
+            const items = ['% display contents ' + path, newSource[index], ''];
+            newSource = [
+                ...newSource.slice(0, index + 3 * i + 1),
+                ...items,
+                ...newSource.slice(index + 3 * i + 1),
+            ];
+        });
+        const toastString =
+            showContentCounts > 0
+                ? 'Show contents for ' +
```
epicfaace: Let's use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for clarity / conciseness. So that would look like:

```
`Show contents for ${showContentCounts} bundle...`
```

frontend/src/components/worksheets/WorksheetDialogs.js
```diff
@@ -253,6 +259,39 @@ class WorksheetDialogs extends React.Component {
...
+                    aria-labelledby='create-content-block-title'
+                    aria-describedby='create-content-block-description'
+                >
+                    <DialogTitle id='create-content-block-title' style={{ color: 'red' }}>
+                        Specify the subpath inside the bundle you want to display content for
+                    </DialogTitle>
+                    <DialogContent>
+                        <TextField
+                            autoFocus
+                            onChange={this.setPathValue}
```
epicfaace: I think we can inline this function (just add it directly here instead of defining it above). It's a bit confusing to have to scroll all the way up to line 29 for a one-line function that's used once.

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -115,6 +116,19 @@ class TableItem extends React.Component<{
         });
     };
 
+    showCheckedBundleRowsContents = () => {
+        let item = this.props.item;
+        let rows = [...Array(item.rows.length).keys()].filter((item, index) => {
```
epicfaace, line 121: Why can't we just use `item.bundles_spec.bundle_infos` here like we did in the `copyCheckedBundleRows` function?

frontend/src/util/worksheet_utils.js
```diff
@@ -304,15 +304,16 @@ export function getAfterSortKey(item, subFocusIndex) {
...
+        // image_block & content_block stores the sort_key in a different property than do other blocks
         return item['bundles_spec']['bundle_infos'][0]['sort_key'];
     }
     const sort_keys = item.sort_keys || [];
     if (sort_keys[subFocusIndex] || sort_keys[subFocusIndex] === 0) {
         return sort_keys[subFocusIndex];
     }
-    return Math.max(...sort_keys);
+    const afterSortKey: number = Math.max(...sort_keys);
+    return isFinite(afterSortKey) ? afterSortKey : -1;
```
epicfaace, line 316: What edge case are we checking for here? Is it just that afterSortKey is not a number -- and in that case, when does that happen? (in that case, we should probably use `isNumber` instead)

frontend/src/util/worksheet_utils.js
```diff
@@ -304,15 +304,16 @@ export function getAfterSortKey(item, subFocusIndex) {
     // The default after_sort_key is -1 when inserting an item on top of the worksheet,
     // so the item's sort_key should always be >= 0 (sort_key > after_sort_key)
     if (!item) return -1;
-    if (item.mode === 'image_block') {
-        // image_block stores the sort_key in a different property than do other blocks
+    if (item.mode === 'image_block' || item.mode === 'contents_block') {
+        // image_block & content_block stores the sort_key in a different property than do other blocks
```
epicfaace:
```suggestion
        // image_block and content_block store the sort_key in a different property than do other blocks
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1677,6 +1683,75 @@ class Worksheet extends React.Component {
...
+            });
+        });
+
+        let newSource = [...this.state.ws.info.source];
+
+        // Add %display line to the worksheet source right before each checked bundle
+        validBundles.forEach((index, i) => {
+            const items = ['% display contents ' + path, newSource[index], ''];
+            newSource = [
+                ...newSource.slice(0, index + 3 * i + 1),
```
epicfaace: I'm a bit confused where these numbers such as 3 and 1 come from. Could you add a comment?

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: good suggestion!

frontend/src/components/worksheets/items/TableItem/TableItem.js
yuqijin: basically, it is because the filtering of the checked bundles are different in `copyCheckedBundleRows` and `showCheckedBundleRowsContents`. In `copyCheckedBundleRows`, we only care about the bundle itself, while in `showCheckedBundleRowsContents`, we also care about the raw index of each checked bundle.

frontend/src/util/worksheet_utils.js
yuqijin: we are checking for whether the afterSortKey is inf or not, so I think we better use `isFinite` to check. WIth inf afterSortKey, the frontend will report an error.

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: added

frontend/src/util/worksheet_utils.js
epicfaace: When would afterSortKey be equal to inf?

yuqijin: when some error occurs, you can find inf check in NewUpload as well.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1677,6 +1683,73 @@ class Worksheet extends React.Component {
...
+            });
+        });
+
+        let newSource = [...this.state.ws.info.source];
+
+        validBundles.forEach((index, i) => {
+            const items = ['% display contents ' + path, newSource[index], ''];
+            // For each checked bundle, we are adding 3 new lines to the source right after each checked bundle
+            newSource = [
+                ...newSource.slice(0, index + 3 * i + 1),
```
epicfaace: Can we instead use `newSource = [...newSource].splice(...)` to insert the 3 new lines? it might be a bit clearer: https://stackoverflow.com/a/586189

epicfaace (approved):

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: updated

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -1708,13 +1708,13 @@ class Worksheet extends React.Component {
...
 
         validBundles.forEach((index, i) => {
-            const items = ['% display contents ' + path, newSource[index], ''];
-            // For each checked bundle, we are adding 3 new lines to the source right after each checked bundle
-            newSource = [
-                ...newSource.slice(0, index + 3 * i + 1),
-                ...items,
-                ...newSource.slice(index + 3 * i + 1),
-            ];
+            newSource.splice(
```
epicfaace, line 1711: @yuqijin Oh, I meant `newSource = [...newSource].splice(...`. This way, we're still immutably changing `newSource`.

yuqijin: I don't think we need to do 'newSource=...' since the splice function will change the original array (while slice function does not)

epicfaace: Oh, you're right -- sorry, didn't realize that `newSource` was already immutably created from `this.state.ws.info.source`.

PR #3335, opened 2021-03-11
Allow workers to have multiple tags

nelson-liu:

Another step toward #3245 . This PR enables workers to have multiple tags. Now, instead of checking for exact match, a bundle can run on a worker if any of the worker's tags matches the bundle's `request_queue`. For instance, if a bundle has `request_queue=B`, and a worker has tags `[A, B]`, then said worker can run the bundle.

In the `Worker` object, `self.tag` is changed from from `str` to `List[str]`.

The worker model stays the same, since we can serialize a list of tags as a single string that's comma-separated (i.e., tags `[A, B]` are stored as `A,B` in MySQL). I'm a little annoyed that the semantics of "tag" have slightly changed (i.e., it's not a single tag, but multiple tags), but I'm not sure if it's worth changing `tag` to `tags` throughout the codebase / doing a DB migration just for this...thoughts?


tests/unit/server/bundle_manager/mocked_test.py
```diff
@@ -215,18 +229,23 @@ def test_filter_and_sort_workers_tag_exclusive_priority(self):
         sorted_workers_list = self.bundle_manager._filter_and_sort_workers(
             self.workers_list, self.bundle, self.bundle_resources
         )
-        self.assertEqual(len(sorted_workers_list), 2)
+        self.assertEqual(len(sorted_workers_list), 3)
         self.assertEqual(sorted_workers_list[0]['worker_id'], 6)
-        self.assertEqual(sorted_workers_list[1]['worker_id'], 5)
+        self.assertEqual(
```
epicfaace, line 234: For consistency / clarity, let's just split this up into two different `self.assertEqual` statements.

epicfaace (commented):
Looks good -- I think it might actually be worth it, for readability in the future, to create a migration to rename `tag` to `tags`. It seems like such a migration would probably be pretty simple, so I think it's worth it to go ahead and add it in.

tests/unit/server/bundle_manager/mocked_test.py
nelson-liu: Sorry if I'm missing something obvious, but I'm not sure what you mean, could you clarify?

I added this `assertEqual(set(), set())`because the ordering of these two items is non-deterministic (since ties are broken randomly), so I just wanted to ensure that these two positions hold these two items, but not assert that any item must occur at any particular position.

epicfaace: Oh, I didn't realize that the ordering is nondeterministic. Let's just keep it then.

tests/unit/worker_manager/slurm_batch_worker_manager_test.py
```diff
@@ -15,7 +15,7 @@ def test_base_command(self):
             partition='some_partition',
             worker_executable='cl-worker',
             worker_idle_seconds='888',
-            worker_tag='some_tag',
+            worker_tag='sometag',
```
epicfaace, line 18: Why remove the underscores here?

nelson-liu: worker tags have to be alphanumeric, so these tags aren't actually valid. This is just a mocked test / I haven't added the functionality to take multiple tags to the workermanager yet, but I'm planning on doing that in a subsequent PR.

nelson-liu: I opened a PR at https://github.com/codalab/codalab-worksheets/pull/3350 that builds upon this one

alembic/versions/2021031301_rename_worker_tag_column_to__6cee23f1e02f.py
```diff
@@ -0,0 +1,24 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '6cee23f1e02f'
+down_revision = 'db12798a7cf6'
+
+
+def upgrade():
+    op.add_column('worker', sa.Column('tags', sa.Text(), nullable=True))
```
epicfaace: I don't think this would work. This would essentially drop all existing tags.

Let's use `op.alter_column` instead (https://stackoverflow.com/a/46452540/1950269).

codalab/model/tables.py
```diff
@@ -431,7 +433,7 @@
     ),  # Whether the worker and the server have a shared filesystem.
     Column(
         'tag_exclusive', Boolean, nullable=False
-    ),  # Whether worker runs bundles if and only if they match tags.
+    ),  # Whether worker runs bundles that are untagged.
```
epicfaace: Why did you change this? I thought `tag_exclusive` set to `True` means "the worker doesn't run untagged bundles" (whereas this description makes it sound like `True` = "the worker runs untagged bundles"). I think the earlier description is probably clearer.

codalab/rest/workers.py
```diff
@@ -28,7 +28,8 @@ def checkin(worker_id):
     socket_id = local.worker_model.worker_checkin(
         request.user.user_id,
         worker_id,
-        request.json.get("tag"),
+        # TODO(nfliu): deprecate passing "tag" in the request. The right key is "tags".
```
epicfaace, line 31: Can we make a GitHub issue for this, then link to the issue in the comment, too?

codalab/server/bundle_manager.py
```diff
@@ -841,11 +840,27 @@ def _get_matched_workers(request_queue, workers):
         :param workers: a list of workers
         :return: a list of matched workers
         """
-        tag_match = re.match('(?:tag=)?(.+)', request_queue)
-        if tag_match is not None:
-            return [worker for worker in workers if worker['tag'] == tag_match.group(1)]
+        prefix = "tag="
```
epicfaace, line 890: It probably would've been most consistent to let people use `tags=` instead... but do you think that's going to be too disruptive to users' workflows?

nelson-liu:
(i must have missed the comments on this too, thanks for the review @epicfaace . I'll take a look at fixing them shortly)

codalab/model/tables.py
nelson-liu: Good point, I'll leave it as is.

alembic/versions/2021031301_rename_worker_tag_column_to__6cee23f1e02f.py
nelson-liu: good catch, thanks!

codalab/server/bundle_manager.py
nelson-liu: I don't think that anyone actually explicitly passes in `tag=` anymore (the default isnot to use `tag=`, i think). Maybe we should just remove all of the logic with `tag=` entirely. https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/worker_manager.py#L188-L194

nelson-liu:
@epicfaace this is ready to be looked at again, if you have a chance.

codalab/worker/main.py
```diff
@@ -33,7 +33,9 @@
 def parse_args():
     parser = argparse.ArgumentParser(description='CodaLab worker.')
     parser.add_argument(
-        '--tag', help='Alphanumeric tag that allows for scheduling runs on specific workers.'
+        '--tags',
```
epicfaace, line 36: The one thing I'm not comfortable about is that this is a breaking change (passing --tags instead of --tag), so this would probably necessitate a major version bump. Maybe would it be less disruptive just to allow people to pass in `--tag` instead?

codalab/server/bundle_manager.py
```diff
@@ -888,11 +887,27 @@ def _get_matched_workers(request_queue, workers):
...
+        if request_queue.startswith(prefix):
```
epicfaace, line 891: Can you explain why we removed the regex? just to make the logic simpler?

codalab/worker/main.py
nelson-liu: thoughts on supporting both? or should we just keep using `--tag`?

codalab/server/bundle_manager.py
nelson-liu: Yeah, iirc that was the reason.

codalab/worker/main.py
epicfaace: Yeah, I would prefer to support `--tag` and just let the user pass in multiple tags in that argument if needed. We could support both if you think it's better, though, not super opinionated on that point.

epicfaace: Hey @nelson-liu would you like to / still have time to make this change? renaming it to `--tag`?

PR #3341, opened 2021-03-12
GCP worker manager

### LINKED ISSUE #3155: Support GCP workers using GKE and a Kubernetes worker manager
- Support GCP workers
- Create a Kubernetes worker manager which can be used with GKE (as well as AWS / Azure's managed Kubernetes services)

teetone:

GCP worker manager with Kubernetes.

### Related issues

Resolves #3155


codalab/worker_manager/gcp_batch_worker_manager.py
```diff
@@ -0,0 +1,114 @@
...
+        self.codalab_password = os.environ.get('CODALAB_PASSWORD')
+        if not self.codalab_username or not self.codalab_password:
+            raise EnvironmentError(
+                'Valid credentials need to be set as environment variables: CODALAB_USERNAME and CODALAB_PASSWORD'
+            )
+
+    def get_worker_jobs(self) -> List[WorkerJob]:
+        # Use kubectl to get the current running pods in JSON format
+        output_json: str = self.run_command(
+            ['kubectl', 'get', 'pod', '--field-selector=status.phase==Running', '--output=json']
```
epicfaace: Should we use the Python Kubernetes client (https://github.com/kubernetes-client/python) instead? This way, the pip package will handle the dependency instead of requiring people to install `kubectl` on the worker manager machine beforehand.

adiprerepa: +1 for this - this is probably the best approach

epicfaace: @teetone I think all the documentation is available on this file (https://raw.githubusercontent.com/kubernetes-client/python/master/kubernetes/docs/CoreV1Api.md).

For example, this particular command corresponds to the `list_pod_for_all_namespaces` function:

image

teetone: @epicfaace Thanks for finding this. I will take a look.

codalab/worker_manager/gcp_batch_worker_manager.py
```diff
@@ -0,0 +1,159 @@
...
+        subparser.add_argument(
+            '--gpus', type=int, default=0, help='Default number of GPUs to request for each worker'
+        )
+        subparser.add_argument(
+            '--memory-mb', type=int, default=2048, help='Default memory (in MB) for each worker'
+        )
+
+        # Output path
+        subparser.add_argument(
+            '--output-path', type=str, default='.', help='Path to the output directory'
```
epicfaace: What does "output" mean? Should we rename it "cert_path" instead? Also, what's the point of specifying this path -- would it be better to use a temporary file instead?

teetone: I'm fine with either specifying a file or a directory. The user might prefer to have the file be at a certain path.

epicfaace (approved):

codalab/worker_manager/gcp_batch_worker_manager.py
teetone: I renamed it to `cert_path` for now.

codalab/worker_manager/gcp_batch_worker_manager.py
```diff
@@ -0,0 +1,157 @@
...
+        work_dir_prefix: str = (
+            self.args.worker_work_dir_prefix if self.args.worker_work_dir_prefix else '/tmp/'
+        )
+        worker_id: str = uuid.uuid4().hex
+        worker_name: str = f'cl-worker-{worker_id}'
+        work_dir: str = os.path.join(work_dir_prefix, f'{worker_name}_work_dir')
+        command: List[str] = self.build_command(worker_id, work_dir)
+        worker_image: str = 'codalab/worker:' + os.environ.get('CODALAB_VERSION', 'latest')
+
+        config: Dict[str, Any] = {
```
adiprerepa, line 116: a bit late but it might be a good idea to use a template from a yaml file, execute it, and maybe pass that into kubernetes. 

I'm not sure what the python best practice is (maybe jinja) but I would think something like this: https://golang.org/pkg/text/template/

PR #3345, opened 2021-03-12
Enable finer-grained selection of workers via richer request_queue

### LINKED ISSUE #3245: Make request_queue tags more flexible

nelson-liu:

Fixes #3245 . This enables bundles to more-flexibly select workers to run on. This is built upon #3335 , so go look at that first.

### Changes:

- Instead of simply checking whether a bundle's `request_queue` exactly matches a worker's `tag` (or any of the tags, after #3335), this uses a small DSL to parse out finer-grained specifications. The defaults remain unchanged---if you provide a `request_queue` that is a single string, and have a single worker tag, everything should work as it did before.
- `request_queue` can now contain logical NOT (`!`), OR (`|`), and AND (`&`). A worker will be matched with a bundle only if the workers tags satisfy the `request_queue`.

For example, the tag:
- `sometag` matches any worker that has the tag `sometag`
- `!sometag` matches any worker that has a tag, but does not have the tag `sometag`
- `sometag | othertag` matches any worker that has either the `sometag` or `othertag` tags
- `sometag & othertag` matches any worker that has _both_ the `sometag` or `othertag` tags.

Operator precedence follows Python---NOT has the highest precedence, followed by AND and then OR. When in doubt, parens are also supported.

Implementation-wise, I wrote a simple grammar to parse the `request_queue` values. The parsing is done by the [Lark parser](https://github.com/lark-parser/lark). The lark parser produces an expression tree, which is then transformed into a string conditional. For example, for the `request_queue` value `sometag`, the conditional is `sometag in worker_tags` (where `worker_tags` is a `List[str]` with the tags of any worker). This string is _evaluated_ into a lambda function, and then run on a given worker's tags to determine if the worker satisfies the `request_queue`.

Some points of discussion:
- It's a reasonable concern that parsing the `request_queue` might be a bit slow. I've used `functools.lru_cache` to cache the calls to the parser, so recently-used `request_queue` values don't need to be parsed every time.
- There are some obvious security implications here, since user-input code is being run. I put in a few safeguards, but I'm curious what people think:

(1) The grammar only successfully parses `request_queue` if all characters are alphanumeric (excluding the operators). So, no `()` or `__` for python trickery.
(2) TODO: I want to put a max length on the `request_queue` field, since a malicious user could try to crash the server by inserting very long values that take awhile to parse. I think even very long values should be fine (since the grammar won't infinitely generate), but a length limit seems like a reasonable precaution anyway.

Anyway, looking forward to hearing your feedback on this!

### DIFF
--- a/codalab/lib/parsing_util.py
+++ b/codalab/lib/parsing_util.py
@@ -28,10 +28,10 @@ class FunctionTransformer(Transformer):
         return f"({' and '.join(exprs)})"
 
     def not_expr(self, atom):
-        return f" not {atom}"
+        return f"not ({atom})"
 
     def name(self, n):
-        return f"'{n}' in worker_tag"
+        return f"'{n}' in worker_tags"
 
 
 class RequestQueueParser:
@@ -46,4 +46,4 @@ class RequestQueueParser:
         # NOTE: this is potentially dangerous. The parser is designed to only accept alphanumeric input
         # (excepting its operators). Similarly, self.collection_name should be alphanumeric.
         # These two constraints make it hard to call functions or do bad things with double-underscores.
-        return eval(f"lambda worker_tag : {function_string}")
+        return eval(f"lambda worker_tags : {function_string}")
--- a/codalab/lib/parsing_util.py
+++ b/codalab/lib/parsing_util.py
@@ -1,0 +1,49 @@
+import functools
+from typing import Callable
+
+from lark import Lark, Transformer, v_args
+
+REQUEST_QUEUE_GRAMMAR = r"""
+    ?start: or_expr
+    ?or_expr: and_expr ("|" and_expr)*
+    ?and_expr: atom ("&" atom)*
+    ?atom: ALPHANUMERIC -> name
+        | "!" atom -> not_expr
+        | "(" or_expr ")"
+
+    ALPHANUMERIC: (LETTER|DIGIT)+
+    %import common.LETTER
+    %import common.DIGIT
+    %import common.WS
+    %ignore WS
+"""
+
+
+@v_args(inline=True)
+class FunctionTransformer(Transformer):
+    def or_expr(self, *exprs):
+        return f"({' or '.join(exprs)})"
+
+    def and_expr(self, *exprs):
+        return f"({' and '.join(exprs)})"
+
+    def not_expr(self, atom):
+        return f" not {atom}"
+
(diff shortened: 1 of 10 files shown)


epicfaace:
Interesting! I'm curious, why create our own DSL instead of maybe trying to use an existing DSL? Or did existing languages not fit the right needs we need here?

nelson-liu:
I couldn't find anything that was appropriate, though if you find something it'd be great to use an out-of-the-box solution. The selector language is actually pretty inspired by what SLURM uses for routing jobs to nodes (go to https://slurm.schedmd.com/sbatch.html and then ctrl-f `-C, --constraint=<list>`, sorry there's no direct link to it).

percyliang:
Should we try to get this merged soon?

epicfaace (commented):
Sorry for the delay with reviewing! Can you document the DSL somewhere in the codalab documentation and/or CLI reference so that users are aware with how to specify complex values for request_queue?

nelson-liu:
I think  #3335 needs to go in first, so i fixed merge conflicts in that.

PR #3350, opened 2021-03-13
Allow WorkerManager to take multiple alphanumeric tags

nelson-liu:
Continuation of https://github.com/codalab/codalab-worksheets/pull/3252 and https://github.com/codalab/codalab-worksheets/pull/3335

This is a branch of https://github.com/codalab/codalab-worksheets/pull/3335, so i'd recommend looking at that first and then I can clean up the diff.

ref: https://github.com/codalab/codalab-worksheets/pull/3335#discussion_r593548780


epicfaace:
Marking this as a draft until https://github.com/codalab/codalab-worksheets/pull/3335 is merged.

Issue #3174, opened 2021-01-12
show status of docker image download


percyliang:
Currently, the `run_status` is stuck in `Pulling docker image: Download starting.` for a very long time and it's not clear what's happening.  We used to show the progress of the docker download, but it disappeared during a refactor.  It'd be nice to get it back.

epicfaace:
Show percentage (such as: `Pulling docker image: Download starting (50%).`)

yiboliu:
am thinking about how to redirect this console info to our frontend:
![image](https://user-images.githubusercontent.com/18689351/105802961-33bdd280-5f51-11eb-85a2-15e16119100b.png)

yiboliu:
@epicfaace looks like [it's pretty hard to get how many layers are there to download](https://stackoverflow.com/questions/25830005/how-can-i-get-percentage-of-pull-docker-image). So I propose that we just show how many layers have been downloaded and what's the progress of downloading the current layer.

epicfaace:
Sounds good.

epicfaace:
Fixed by https://github.com/codalab/codalab-worksheets/pull/3351

adiprerepa:
this can't close until #3382 gets merged

PR #3351, opened 2021-03-14
Make docker image status more granular with the percentage done

adiprerepa:

Set status of docker image more granular.

### Related issues

#3174 

### Screenshots

image

### Checklist


codalab/worker/worker_run_state.py
```diff
@@ -290,8 +291,9 @@ def mount_dependency(dependency, shared_file_system):
         if not dependencies_ready:
             status_message = status_messages.pop()
             if status_messages:
-                status_message += "(and downloading %d other dependencies and docker images)" % len(
-                    status_messages
+                status_message += (
+                    "\n(and downloading %d other dependencies and docker images)"
```
epicfaace: Do we need this newline?

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,14 @@ def image_availability_state(image_spec, success_message, failure_message):
                 def download():
                     logger.debug('Downloading Docker image %s', image_spec)
                     try:
-                        self._docker.images.pull(image_spec)
+                        # client = docker.APIClient(base_url='unix://var/run/docker.sock')
```
epicfaace: Can you remove the comments?

epicfaace (commented):
Were you able to test this? Can you show a screenshot of how the status looks like now?

codalab/worker/worker_run_state.py
adiprerepa: I think it looks weird without it.

adiprerepa:
image

I used percentages, this is how it looks.

codalab/worker/worker_run_state.py
epicfaace: Ok -- the screenshot on this PR seems to not have the newline -- can you change the screenshot to the latest one so I can see how it looks like with the newline added?

image

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,13 @@ def image_availability_state(image_spec, success_message, failure_message):
                 def download():
                     logger.debug('Downloading Docker image %s', image_spec)
                     try:
-                        self._docker.images.pull(image_spec)
+                        client = docker.APIClient(base_url='unix://var/run/docker.sock')
+                        self.generator = client.pull(image_spec, stream=True, decode=True)
```
epicfaace: Why do we need to create a new client as opposed to just running `self._docker.images.pull`?

codalab/worker/worker_run_state.py
```diff
@@ -276,22 +277,32 @@ def mount_dependency(dependency, shared_file_system):
         docker_image = run_state.resources.docker_image
         image_state = self.docker_image_manager.get(docker_image)
         if image_state.stage == DependencyStage.DOWNLOADING:
-            status_messages.append(
-                'Pulling docker image: ' + (image_state.message or docker_image or "")
-            )
+            msg = 'Pulling docker image %s' % docker_image
+            if isinstance(image_state.message, dict) and image_state.message['progressDetail']:
+                msg += '(%d%%)' % (
```
epicfaace: I think it would be cleaner to instead just make sure that `image_state.message` is always a string (it's good practice in general not to have the same variable have multiple possible types in Python). This means that this code here would probably be moved over to `docker_image_manager.py` and you'd set `self._downloading[image_spec]['status']` equal to `"Pulling docker image (XX%)"`

codalab/worker/worker_run_state.py
```diff
@@ -276,22 +277,32 @@ def mount_dependency(dependency, shared_file_system):
...
+
+            status_messages.append(msg)
             dependencies_ready = False
         elif image_state.stage == DependencyStage.FAILED:
             # Failed to pull image; -> CLEANING_UP
             message = 'Failed to download Docker image: %s' % image_state.message
             logger.error(message)
             return run_state._replace(stage=RunStage.CLEANING_UP, failure_message=message)
 
+        logger.debug(dependencies_ready)
```
epicfaace: rm this line

epicfaace:
> image
> 
> I used percentages, this is how it looks.

Can you change this so that there's a space before the parenthesis `(` as well?

codalab/worker/docker_image_manager.py
adiprerepa: When I do this, the `generator` returned is not an iterator. I am not sure why this is the case - but it only works when you have a new `APIClient`

epicfaace: Ok. What exactly is returned from the generator coming from `self._docker.images.pull`? Is `self._docker.images` also an `APIClient`?

adiprerepa: its an `ImageCollection` I believe - I will have to double check.

adiprerepa: It looks like when we use `self._docker.images.pull`, it calls the same `pull` as we do with the new `APIClient`, but omits the output.

Reference: https://github.com/docker/docker-py/blob/master/docker/models/images.py#L449

adiprerepa:
image

This is what it looks like when there are dependencies. Imo it looks good

codalab/worker/docker_image_manager.py
epicfaace: Ah, I see. In that case, can we just replace this with `client.client.api.pull` then? This way, we don't have to instantiate a new `APIClient`.

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,21 @@ def image_availability_state(image_spec, success_message, failure_message):
...
+                        for line in self.generator:
+                            self._downloading[image_spec]['status'] = (
```
epicfaace: I think we can pull this line out of the for loop

codalab/worker/worker_run_state.py
```diff
@@ -290,8 +289,9 @@ def mount_dependency(dependency, shared_file_system):
         if not dependencies_ready:
             status_message = status_messages.pop()
             if status_messages:
-                status_message += "(and downloading %d other dependencies and docker images)" % len(
-                    status_messages
+                status_message += (
+                    "\n(and downloading %d other dependencies and docker images)"
```
epicfaace: I don't see a newline in the latest screenshot you sent. I'm guessing perhaps the newline just doesn't show up in the HTML; in that case, let's just remove the newline.

codalab/worker/worker.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         # A flag indicating if the worker will exit if it encounters an exception
         exit_on_exception=False,  # type: bool
     ):
+
```
epicfaace: undo this change

codalab/worker/worker_run_state.py
```diff
@@ -17,6 +17,7 @@
 from codalab.worker.fsm import DependencyStage, StateTransitioner
 from codalab.worker.worker_thread import ThreadDict
 
+
```
epicfaace: undo this change

codalab/worker/worker_run_state.py
```diff
@@ -276,9 +277,7 @@ def mount_dependency(dependency, shared_file_system):
         docker_image = run_state.resources.docker_image
         image_state = self.docker_image_manager.get(docker_image)
         if image_state.stage == DependencyStage.DOWNLOADING:
-            status_messages.append(
-                'Pulling docker image: ' + (image_state.message or docker_image or "")
-            )
+            status_messages.append(image_state.message)
```
epicfaace: I think we should keep the "Pulling docker image [image_spec]" here (and remove it from docker_image_manage.py), just for consistency with the rest of the code.

codalab/worker/docker_image_manager.py
adiprerepa: how do you mean?

epicfaace: Oh, I see what you were doing. Never mind my comment

codalab/worker/worker_run_state.py
adiprerepa: I see - but keep the percent stuff in the message in `docker_image_manager.py`

adiprerepa:
> Ah, I see. In that case, can we just replace this with client.client.api.pull then? This way, we don't have to instantiate a new APIClient.

@epicfaace what do you mean? `client` is the field being instantiated

adiprerepa:
I just used `self._docker.api.pull`

epicfaace:
> I just used `self._docker.api.pull`

Oh, ok, that makes sense then, thanks for updating

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,18 @@ def image_availability_state(image_spec, success_message, failure_message):
                 def download():
                     logger.debug('Downloading Docker image %s', image_spec)
                     try:
-                        self._docker.images.pull(image_spec)
+                        self.generator = self._docker.api.pull(image_spec, stream=True, decode=True)
+                        for line in self.generator:
+                            self._downloading[image_spec]['status'] = ''
+                            if line['status'] == 'Downloading':
+                                self._downloading[image_spec]['status'] = ' (%d%%)' % (
```
epicfaace: Let's move the space to worker_run_state.py instead
```suggestion
                                self._downloading[image_spec]['status'] = '(%d%%)' % (
```

codalab/worker/worker_run_state.py
```diff
@@ -277,7 +277,7 @@ def mount_dependency(dependency, shared_file_system):
         image_state = self.docker_image_manager.get(docker_image)
         if image_state.stage == DependencyStage.DOWNLOADING:
             status_messages.append(
-                'Pulling docker image: ' + (image_state.message or docker_image or "")
+                'Pulling docker image %s%s' % (docker_image, image_state.message)
```
```suggestion
                'Pulling docker image %s %s' % (docker_image, image_state.message)
```

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,18 @@ def image_availability_state(image_spec, success_message, failure_message):
                 def download():
                     logger.debug('Downloading Docker image %s', image_spec)
                     try:
-                        self._docker.images.pull(image_spec)
+                        self.generator = self._docker.api.pull(image_spec, stream=True, decode=True)
```
epicfaace: No need to assign to an instance variable
```suggestion
```

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,18 @@ def image_availability_state(image_spec, success_message, failure_message):
...
+                        for line in self.generator:
```
epicfaace:
```suggestion
                        for line in self._docker.api.pull(image_spec, stream=True, decode=True):
```

codalab/worker/docker_image_manager.py
```diff
@@ -243,10 +243,18 @@ def image_availability_state(image_spec, success_message, failure_message):
...
+                            self._downloading[image_spec]['status'] = ''
+                            if line['status'] == 'Downloading':
```
epicfaace, line 249:
```suggestion
                            # Set the status to the percent completed
                            if line['status'] == 'Downloading':
```

epicfaace (approved):
Looks good! Just a few more changes and then we'll be all set to merge.

adiprerepa:
@epicfaace should be good to go!

PR #3352, opened 2021-03-17
Fix the bug of dependencies with same shorten uuid

### LINKED ISSUE #3322: Can't start a run on the UI if two or more bundles have the same shortened UUID
**Describe the bug**
I can't start a run on the UI if two or more bundles have the same shortened UUID and one of them is specified as a dependency.

**To Reproduce**
1. Start a run with https://worksheets.codalab.org/bundles/0x9a2add5219db4ebc89965d7f42719750 as a dependency.
2. Expand the bundle and click "Edit and Rerun".
3. Notice that the bundle does not get restarted.
4. Open the browser's network tab to see the error.

**Expected behavior**
Bundle should restart.

**Screenshots**
codalab_bug

yuqijin:

When two dependencies have the same shorten uuid, a new run cannot be started.
It is because previously the dependencies are represented by shorten uuid & name when being passed to the backend. When the backend validated the dependency, it will fail. 

### Related issues

fixes #3322 

### Screenshots

Screen Shot 2021-03-17 at 4 25 16 PM

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/NewRun/NewRun.jsx
+++ b/frontend/src/components/worksheets/NewRun/NewRun.jsx
@@ -287,7 +287,7 @@ class NewRun extends React.Component<{
 
         for (let dep of dependencies) {
             const key = dep.alias;
-            let value = shorten_uuid(dep.target.uuid);
+            let value = dep.target.uuid;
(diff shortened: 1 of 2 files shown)


teetone (approved):
Thanks for fixing this!

PR #3353, opened 2021-03-19
Use useSWR  as a cache for the search directive

### LINKED ISSUE #2174: When refreshing the page after a worksheet item is saved, don't refresh the search results

### LINKED ISSUE #3277: Duplicate requests are made to resolve directives

### LINKED ISSUE #3307: Use useSWR for data fetching
We should start using `useSWR` (https://swr.vercel.app/getting-started) to wrap most of our ajax / fetch calls made on the frontend. This allows allows for duplicate calls to be cached and deduped easily.

In the short-term, this will be a good way to solve issues like #2174 (caching for partial refresh of worksheets), #3277 (duplicate requests are made to resolve directives), and #3203 (two requests being made for data fetching).

In the long-term, we can treat useSWR as a common data store for network data, which allows us to not have to pass data down too many props and makes our frontend code more simplified.

yuqijin:

Use swr as a global cache to solve the problem of duplicate requests for a search directive, 
and achieve partial refresh of the worksheet by not refreshing the search results when the search results do not change.

### Related issues

fixes #3277 
fixes #2174 

### Screenshots

Screen Shot 2021-03-18 at 6 03 32 PM

ezgif com-gif-maker (3)

### Checklist


frontend/src/components/worksheets/items/PlaceholderItem.js
```diff
@@ -1,56 +1,58 @@
...
+
+    useEffect(() => {
+        if (cache.has(url)) {
+            setBlocks(cache.get(url));
+        }
+    });
+
+    // fetch data only once
+    useSWR(url, fetcher, {
+        revalidateOnMount: !cache.has(url),
```
epicfaace, line 50: Why can't we just use `revalidateOnMount: false`? Trying to simplify the logic if possible.

yuqijin: Because we need to fetch the data for the first time when the data does not exist.

frontend/src/components/worksheets/items/PlaceholderItem.js
```diff
@@ -1,56 +1,58 @@
...
+    }
+    const url =
+        '/rest/interpret/worksheet/' + worksheetUUID + '?' + queryString.stringify({ directive });
+    // use data stored in cache
+
+    useEffect(() => {
+        if (cache.has(url)) {
+            setBlocks(cache.get(url));
+        }
+    });
```
epicfaace, line 46: Let's add the dependency array.

```suggestion
    }, []);
```

frontend/src/components/worksheets/items/PlaceholderItem.js
```diff
@@ -1,56 +1,58 @@
...
+
+    // fetch data only once
+    useSWR(url, fetcher, {
+        revalidateOnMount: !cache.has(url),
+        onSuccess: (data, key, config) => {
```
epicfaace, line 51: Do we need this callback? Given that we already have `useEffect` earlier

yuqijin: Yes, cuz we need to update the page after the data is successfully fetched.

epicfaace:
Looks like you need to change the dependencies in useEffect, according to CI:

image

epicfaace (commented):
Sounds good. Let's merge this into master for now and if we find a more efficient way, we can update it later.

epicfaace (approved):

PR #3359, opened 2021-03-19
Upgrade urllib3 to 1.26.4

epicfaace:
Fixes a security vulnerability (https://github.com/advisories/GHSA-5phf-pp7p-vc2r). Technically, this vulnerability doesn't affect urllib3 < 1.26 (so it doesn't impact codalab; seems to be an issue with github showing a false positive), but we might as well upgrade the package so that github doesn't keep showing this alert as a false positive.

I had to bump the requests and boto3 versions so that urllib3 > 1.26 would be supported, as well.

### DIFF
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -7,7 +7,7 @@
 # Only for development
 -r requirements.dev.txt
 
-boto3==1.15.7
+boto3==1.17.33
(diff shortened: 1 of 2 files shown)


adiprerepa (approved):
lgtm

PR #3365, opened 2021-03-23
Add meta tag to home

### LINKED ISSUE #3140: Improve SEO for home page
The home page description looks like this:

image

Ideally, we'd probably want it to say something like "A collaborative platform for reproducible research ... Run your machine learning experiments in the cloud. Manage them in a digital lab notebook. Publish them so other researchers can reproduce your results.", like what we have in our actual home page:

image

jzwang43:

Add meta tag to improve google search result for Codalab Worksheets.

### Related issues

fixes #3140 

### Screenshots

image

The meta tag is installed into `<head>` correctly.
Will need to double check once deployed to verify Google search is displaying the description correctly.

### Checklist


frontend/package.json
```diff
@@ -26,19 +26,20 @@
...
     "query-string": "^6.13.1",
+    "rc-slider": "^9.7.1",
     "react": "^16.13.1",
+    "react-avatar-editor": "^11.1.0",
     "react-bootstrap": "^0.33.1",
     "react-circular-progressbar": "^2.0.3",
     "react-cookie": "^4.0.3",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.13.1",
+    "react-meta-tags": "^1.0.1",
```
epicfaace: Can we use `react-helmet` instead? It's a more widely used library / more of a standard for adding meta tags or tags in the head.

epicfaace (changes requested):

frontend/package.json
jzwang43: I have switched to use `react-helmet`. Thanks for the feedback!

frontend/package.json
```diff
@@ -26,19 +26,20 @@
     "marked": "^2.0.0",
     "material-ui-chip-input": "^1.1.0",
     "query-string": "^6.13.1",
+    "rc-slider": "^9.7.1",
     "react": "^16.13.1",
+    "react-avatar-editor": "^11.1.0",
```
epicfaace, line 31: Can you ensure you're running npm 6 (not npm 7), then undo the changes to package-lock.json, then redo them? It appears that because you're using npm 7, the lockfile version has changed to 2, but we want to keep it at version 1 for now (at least until we can upgrade default versions of node / npm on CI at a later time).

epicfaace (approved):

Issue #3369, opened 2021-03-26
investigate time_user and time_system accuracy


adiprerepa:
When I was working on https://github.com/codalab/codalab-worksheets/issues/3328, I noticed that `time_user` and `time_system` seemed to be always set to 0. 

I'm not 100% sure that this is an issue with in the prod database, need to check, but I am not sure how much it matters because it doesn't seem to be used anywhere. The `time` field that we see on worksheets.codalab.org on a bundle's completion seems to be accurate.

See #3328 's associated pr for more info.

epicfaace:
We should remove `time_user`, and `time_system`.

Issue #3328, opened 2021-03-05
Profile the slowness of each stage of a bundle


epicfaace:
Store the speed of each stage of a bundle in the bundle metadata.

- How long in STAGED
- How long in RUNNING
- etc.

adiprerepa:
Note to self:

Should be able to mimic behavior of `time_used` & how its sent over the API in metadata and stuff.

benchmark transition here: https://github.com/codalab/codalab-worksheets/blob/dd81f5f0f5668ba4009a79418dd6c45e10a7d24e/codalab/worker/worker.py#L478

Open Questions:
- store time elapsed in database or start&end times?

epicfaace:
Document the different kinds of times

adiprerepa:
I think I will store start and end times in whatever `time.time()` gives me, and then get the diff when I send it over the API

PR #3370, opened 2021-03-26
benchmark bundle stages

adiprerepa:
for https://github.com/codalab/codalab-worksheets/issues/3328

The start and end times for each stage are measured in the worker and then sent over the API in a new field `bundle_profile_stats`, which contains the start, end and elapsed time in seconds for each stage.

The fields added to the metadata for the `bundle_metadata` table are `time_preparing`, `time_running`, `time_cleaning_up`, and `time_uploading_results`. 

As for the other time fields we have in the bundle metadata table (cc @percyliang ), we seem to store the fields `time`, `time_user`, and `time_system`. However, `time_user` and `time_system` were always 0 in my investigation, they may be broken. They might not be sent over the API in the first place from what I can tell. Issue filed: #3369 

Screenshot of table:
image


codalab/bundles/run_bundle.py
```diff
@@ -54,6 +54,10 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('last_updated', int, 'Time when information about this bundle was last updated.', generated=True, formatting='date'))
     METADATA_SPECS.append(MetadataSpec('run_status', str, 'Execution status of the bundle.', generated=True))
     METADATA_SPECS.append(MetadataSpec('staged_status', str, 'Information about the status of the staged bundle.', generated=True))
+    METADATA_SPECS.append(MetadataSpec('time_preparing', float, 'Amount of system time in the PREPARING stage.', generated=True, formatting='duration'))
```
epicfaace, line 57: Can we also add `time_staged`? Bundles sometimes spend a considerable amount of time being stuck in staged, so it would be useful to have that metric too.

codalab/worker/worker.py
```diff
@@ -476,6 +476,8 @@ def process_runs(self):
         for uuid in self.runs:
             run_state = self.runs[uuid]
             self.runs[uuid] = self.run_state_manager.transition(run_state)
+            # all the _replace calls for the namedtuple are propogated here
```
epicfaace: What do you mean?

codalab/worker/worker.py
```diff
@@ -759,6 +772,23 @@ def bundle_service_reply(self, socket_id, err, message, data):
         else:
             self.bundle_service.reply(self.id, socket_id, message)
 
+    def save_time_stats(self, uuid, from_stage):
```
epicfaace: Can you add docstrings under each of these new methods?

codalab/worker/worker.py
```diff
@@ -622,6 +625,15 @@ def initialize_run(self, bundle, resources):
                 bundle=bundle,
                 bundle_path=os.path.realpath(bundle_path),
                 bundle_dir_wait_num_tries=Worker.BUNDLE_DIR_WAIT_NUM_TRIES,
+                bundle_profile_stats={
+                    RunStage.PREPARING: self.initstats(),
```
epicfaace: If we store all this information, why are we only storing times in metadata for preparing, running, cleaning up, and uploading results?

codalab/worker/worker.py
```diff
@@ -759,6 +772,23 @@ def bundle_service_reply(self, socket_id, err, message, data):
...
+
+    def end_stage(self, uuid, stage):
+        self.runs[uuid].bundle_profile_stats[stage]['end'] = time.time()
+        self.runs[uuid].bundle_profile_stats[stage]['elapsed'] = (
+            self.runs[uuid].bundle_profile_stats[stage]['end']
+            - self.runs[uuid].bundle_profile_stats[stage]['start']
+        )
+
+    def initstats(self):
+        return {'start': 0, 'end': 0, 'elapsed': -1}
```
epicfaace: Why -1 and not 0?

epicfaace (commented):
Can we add some kind of automated test to cover this functionality as well?

codalab/worker/worker.py
adiprerepa: was just a note to self - will delete

codalab/worker/worker.py
adiprerepa: i wanted to have it marked as an "impossible" value, elapsed can technically be 0

codalab/worker/worker.py
adiprerepa: I should probably remove the extraneous stuff.

it goes from uploading results->finalizing, and that seems to be the last stage (there is a start time set but no end time). I could not find where it gets out of finalizing, maybe if I set the end time there we could have the finalizing stage.

codalab/bundles/run_bundle.py
adiprerepa: the starting state seems to be `PREPARING` in the worker, would I record this in the bundle manager or something?

epicfaace: Oh, if it's in a different spot, we don't need to add it in this PR then.

codalab/worker/worker.py
```diff
@@ -622,6 +625,15 @@ def initialize_run(self, bundle, resources):
                 bundle=bundle,
                 bundle_path=os.path.realpath(bundle_path),
                 bundle_dir_wait_num_tries=Worker.BUNDLE_DIR_WAIT_NUM_TRIES,
+                bundle_profile_stats={
+                    RunStage.PREPARING: self.initstats(),
+                    RunStage.RUNNING: self.initstats(),
+                    RunStage.CLEANING_UP: self.initstats(),
+                    RunStage.UPLOADING_RESULTS: self.initstats(),
+                    RunStage.FINALIZING: self.initstats(),
+                    RunStage.FINISHED: self.initstats(),
```
percyliang: Should FINISHED (and RESTAGED) be actual stages where we have stats for?  It seems like it's more of an endpoint.

codalab/worker/worker.py
percyliang: Maybe `None`?

codalab/worker/worker.py
```diff
@@ -759,6 +772,23 @@ def bundle_service_reply(self, socket_id, err, message, data):
         else:
             self.bundle_service.reply(self.id, socket_id, message)
 
+    def save_time_stats(self, uuid, from_stage):
+        self.end_stage(uuid, from_stage.stage)
+        self.start_stage(uuid, self.runs[uuid].stage)
+
+    def start_stage(self, uuid, stage):
```
percyliang: If these methods are only going to measure stats, then perhaps the name should reflect that, since `start_stage` in the context of the worker could mean a lot of things.

percyliang:
Note that the `state` field on the server is different from the run stage on the worker, and there's a mapping between the two.  We should confirm that the worker fields are what we want.

Also, given all the variants of `time`, it would be good to have a docstring somewhere that details what the convention is - there's `time`, `time_user`, `time_system` and `time_<WORKER_STAGE>`.  Is `time` the same as `time_running`?

codalab/worker/worker.py
adiprerepa: Yeah, good catch. I just initially included all the stages for quick iteration.

adiprerepa:
@percyliang I can make a docstring, but it seems that:
- `time` is the amount of time spent in the docker container - we get this *from docker*
- `time_<stage>` is measured outside the docker container, so it is probably a little longer.
- As far as I can tell, `time_user` and `time_system` are broken and not set anywhere - this would be confirmed by my investigation in issue #3369. We may want to remove these altogether.

adiprerepa:
> Can we add some kind of automated test to cover this functionality as well?

Would the test just check if the fields exist? I'm not sure if we can enforce a strict rule for benchmarks (checking if the time spent is a certain value).

epicfaace:
> > Can we add some kind of automated test to cover this functionality as well?
> 
> Would the test just check if the fields exist? I'm not sure if we can enforce a strict rule for benchmarks (checking if the time spent is a certain value).

Yeah, it could test to see if the fields are set to a non-zero value.

tests/unit/server/bundle_manager/schedule_run_bundles_test.py
```diff
@@ -92,3 +92,27 @@ def test_finalizing_bundle_gets_finished(self):
 
         bundle = self.bundle_manager._model.get_bundle(bundle.uuid)
         self.assertEqual(bundle.state, State.READY)
+        self.assertNotEqual(
```
epicfaace: In fact, I think here, we can actually use `assertEqual` and ensure it's equal to the proper values, since the start / end times are mocked anyway (so in this case, it would be 5)

adiprerepa: makes sense to me

adiprerepa:
@epicfaace anything else I should address?

epicfaace:
@adiprerepa it looks good, can you just add a screenshot of the output when someone runs `cl info` (or the table on the Web UI) on a bundle that's running / just ran? The screenshot of the database table is good but it would be best to have a screenshot of what the user actually sees.

codalab/worker/worker.py
```diff
@@ -643,6 +655,7 @@ def initialize_run(self, bundle, resources):
                 finalized=False,
                 is_restaged=False,
             )
+            self.start_stage_stats(bundle.uuid, RunStage.PREPARING)
```
epicfaace, line 660:
```suggestion
            # Start measuring bundle stats for the initial bundle state. Subsequent states
            # are measured by the call to save_time_stats in process_runs.
            self.start_stage_stats(bundle.uuid, RunStage.PREPARING)
```

codalab/worker/worker.py
```diff
@@ -759,6 +772,27 @@ def bundle_service_reply(self, socket_id, err, message, data):
         else:
             self.bundle_service.reply(self.id, socket_id, message)
 
+    def save_time_stats(self, uuid, from_stage):
```
epicfaace: It's a state, not a stage
```suggestion
    def save_time_stats(self, uuid, from_state):
```

codalab/worker/worker.py
```diff
@@ -476,6 +476,7 @@ def process_runs(self):
         for uuid in self.runs:
             run_state = self.runs[uuid]
             self.runs[uuid] = self.run_state_manager.transition(run_state)
+            self.save_time_stats(uuid, run_state)
```
epicfaace: I think we should inline the function `save_time_stats` directly in this function, so it's clearer what is actually going on. It's not clear that  `save_time_stats` not only ends the current stage, but starts measuring the next stage, and it's only two lines of code, so I think it may be clearer to inline.

```
        self.end_stage(uuid, from_stage.stage)
        self.start_stage_stats(uuid, self.runs[uuid].stage)
```

Moreover, there's an edge case that this code doesn't apply to. It's possible for `self.run_state_manager.transition` to return the same stage without changing it. In that case, we don't want to reset the stats for the stage, so maybe we should add some code like this:

```
            prev_state = self.runs[uuid]
            self.runs[uuid] = self.run_state_manager.transition(prev_state)
            if prev_state.stage != self.runs[uuid].stage:
                 self.end_stage(uuid, prev_state.stage)
                 self.start_stage_stats(uuid, self.runs[uuid].stage)
```

epicfaace (changes requested):

codalab/worker/worker.py
adiprerepa: makes sense to me. thanks for the detailed feedback!

epicfaace:
> @adiprerepa it looks good, can you just add a screenshot of the output when someone runs `cl info` (or the table on the Web UI) on a bundle that's running / just ran? The screenshot of the database table is good but it would be best to have a screenshot of what the user actually sees.

@adiprerepa , could you address this comment?

codalab/worker/worker.py
```diff
@@ -759,6 +775,19 @@ def bundle_service_reply(self, socket_id, err, message, data):
         else:
             self.bundle_service.reply(self.id, socket_id, message)
 
+    def start_stage_stats(self, uuid, stage):
+        self.runs[uuid].bundle_profile_stats[stage]['start'] = time.time()
+
+    def end_stage(self, uuid, stage):
```
epicfaace: For consistency, let's rename end_stage to end_stage_stats and initstats to init_stage_stats().

Also, please add docstrings to start_stage_stats, end_stage_stats, and initstats.

adiprerepa:
@epicfaace heres the screenshot
image

codalab/worker/worker.py
epicfaace: @adiprerepa can you add type annotations and docstrings to all the functions you added? See this for an example: https://github.com/codalab/codalab-worksheets/blob/a715ecbadf93bbe682dab45a3015d5195d331e1a/codalab/worker/tar_subdir_stream.py#L82-L90

adiprerepa: didnt know python had those, will do

adiprerepa: @epicfaace are the docstrings I have right now fine or do they need to be changed?

codalab/worker/worker.py
```diff
@@ -759,6 +775,29 @@ def bundle_service_reply(self, socket_id, err, message, data):
...
+        self.runs[uuid].bundle_profile_stats[stage]['elapsed'] = (
+            self.runs[uuid].bundle_profile_stats[stage]['end']
+            - self.runs[uuid].bundle_profile_stats[stage]['start']
+        )
+
+    def init_stage_stats(self) -> Dict:
+        """
+        Returns a stage stats dict with default empty values for start, end, and elapsed.
+        """
+        return {'start': 0, 'end': 0, 'elapsed': None}
```
epicfaace: Let's use None as a deafult
```suggestion
        return {'start': None, 'end': None, 'elapsed': None}
```

tests/unit/server/bundle_manager/schedule_run_bundles_test.py
epicfaace: @adiprerepa can you address this comment when you get a chance?

codalab/worker/bundle_state.py
```diff
@@ -218,6 +220,7 @@ def from_dict(cls, dct):
             remote=dct['remote'],
             exitcode=dct['exitcode'],
             failure_message=dct['failure_message'],
+            bundle_profile_stats=dct['bundle_profile_stats'],
```
epicfaace: Let's do this:
```suggestion
            bundle_profile_stats=dct.get('bundle_profile_stats'),
```

because that way existing bundles that are running during deployment (which won't have the bundle_profile_stats key set) will be backwards compatible

codalab/worker/worker.py
```diff
@@ -622,6 +626,16 @@ def initialize_run(self, bundle, resources):
                 bundle=bundle,
                 bundle_path=os.path.realpath(bundle_path),
                 bundle_dir_wait_num_tries=Worker.BUNDLE_DIR_WAIT_NUM_TRIES,
+                bundle_profile_stats={
+                    RunStage.PREPARING: self.init_stage_stats(),
```
epicfaace, line 632: Let's do this for all the keys, so that the key will actually be equal to the value of `RunStage.PREPARING` rather than the actual literal string `"RunStage.PREPARING"`
```suggestion
                    [RunStage.PREPARING]: self.init_stage_stats(),
```

codalab/worker/bundle_state.py
adiprerepa: Should we make that the case for every field? for the sake of consistency

tests/unit/server/bundle_manager/schedule_run_bundles_test.py
adiprerepa: done

codalab/worker/worker.py
epicfaace: @adiprerepa I believe you didn't address this yet -- could you take a look?

codalab/worker/bundle_state.py
epicfaace: Sure, good point -- feel free to

codalab/worker/worker.py
adiprerepa: is it not addressed? just putting brackets around each `RunStage.*`?

epicfaace: Looks like you addressed it in these particular lines of code, but there aren't brackets here: https://github.com/codalab/codalab-worksheets/pull/3370/files#diff-2faf6aac04322764cf1ab1ca223e08f26783c1a7aa03b125a5468c246c1916a9R272-R278

Moreover, by adding brackets, that means that this code should also change: https://github.com/codalab/codalab-worksheets/pull/3370/files#diff-66dcf239a8feb8b8d4031040dca05ead4e0d27ec4f580e7987e858028c46a82fR948-R961

epicfaace: @adiprerepa , I think you still forgot to add brackets to `tests/unit/server/bundle_manager/__init__.py`.

That also might be why the tests are failing. Could you address this issue and ensure CI is passing before requesting another review?

adiprerepa: Doesn't adding the brackets make the element a list? `bundle_profile_stats` is a dictionary, so the way it is right now, the value of `RunStage.PREPARING` (`'RUN_STAGE.PREPARING'`) is the key of the element.

epicfaace: We discussed this over slack and decided to remove the brackets.

codalab/worker/worker.py
```diff
@@ -474,8 +474,11 @@ def process_runs(self):
 
         # 1. transition all runs
         for uuid in self.runs:
-            run_state = self.runs[uuid]
-            self.runs[uuid] = self.run_state_manager.transition(run_state)
+            prev_state = self.runs[uuid]
+            self.runs[uuid] = self.run_state_manager.transition(prev_state)
+            if prev_state.stage != self.runs[uuid].stage:
```
epicfaace, line 480: Add a comment here explaining what this if statement does, and why we want to run this code only in this if statement

codalab/worker/worker.py
```diff
@@ -622,6 +626,16 @@ def initialize_run(self, bundle, resources):
...
                 bundle_dir_wait_num_tries=Worker.BUNDLE_DIR_WAIT_NUM_TRIES,
+                bundle_profile_stats={
+                    RunStage.PREPARING: self.init_stage_stats(),
+                    RunStage.RUNNING: self.init_stage_stats(),
+                    RunStage.CLEANING_UP: self.init_stage_stats(),
+                    RunStage.UPLOADING_RESULTS: self.init_stage_stats(),
+                    RunStage.FINALIZING: self.init_stage_stats(),
+                    # note: this is not sent over the api, but
+                    # the start time needs to be set
+                    RunStage.FINISHED: self.init_stage_stats(),
```
epicfaace: I think it would be cleaner to remove RunStage.FINISHED altogether from this dictionary, and then in the line where we call `self.start_stage_stats(uuid, self.runs[uuid].stage)`, add an if statement to ensure it's called only if `self.runs[uuid].stage != RunState.FINISHED`. This way, we're only storing what we need in this dictionary.

adiprerepa: We would add this check only when starting stage statistics, correct?

codalab/worker/worker.py
adiprerepa: done

codalab/worker/worker.py
```diff
@@ -476,9 +476,13 @@ def process_runs(self):
         for uuid in self.runs:
             prev_state = self.runs[uuid]
             self.runs[uuid] = self.run_state_manager.transition(prev_state)
+            # it's possible that a bundle can return to the same stage, so we don't want to reset that initial stage.
```
epicfaace: I think this would be a clearer comment that could replace the two comments you have here (it's currently not clear what "reset" means until you actually read & understand the code):
"Only start keeping stats for a new stage when the run has actually transitioned to that stage."

epicfaace (approved):

PR #3372, opened 2021-03-30
Fix target sizes for directories -- return data_size if user requests the whole bundle

### LINKED ISSUE #3120: Target sizes for directories are 4096, instead of the size of the entire directory

jzwang43:

The target size of directories should be reported as the size of the entire directory.

### Related issues

fixes #3120  

### Screenshots

Correct target size for a bundle:

image

Target size for a subdir is set to 0:

image

Target size for a file:

image

### Checklist

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -691,8 +691,15 @@ def _fetch_bundle_contents_blob(uuid, path=''):
     else:
         response.set_header('Content-Disposition', 'attachment; filename="%s"' % filename)
     response.set_header('Target-Type', target_info['type'])
-    response.set_header('X-Codalab-Target-Size', target_info['size'])
-
+    if target_info['type'] == 'file':
+        size = target_info['size']
+    elif not path and bundle_name:
+        # return data_size if the user requests the actual bundle
+        size = local.model.get_bundle(target.bundle_uuid).metadata.data_size
+    else:
+        # if request is for a subdir in a bundle then return 0
+        size = 0
+    response.set_header('X-Codalab-Target-Size', size)
     return fileobj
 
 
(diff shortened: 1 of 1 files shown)


codalab/rest/bundles.py
```diff
@@ -691,8 +691,15 @@ def _fetch_bundle_contents_blob(uuid, path=''):
     else:
         response.set_header('Content-Disposition', 'attachment; filename="%s"' % filename)
     response.set_header('Target-Type', target_info['type'])
-    response.set_header('X-Codalab-Target-Size', target_info['size'])
-
+    if target_info['type'] == 'file':
+        size = target_info['size']
+    elif not path and bundle_name:
+        # return data_size if the user requests the actual bundle
+        size = local.model.get_bundle(target.bundle_uuid).metadata.data_size
```
epicfaace: Let's use the `local.model.get_bundle_metadata` function instead.

epicfaace (changes requested):

codalab/rest/bundles.py
jzwang43: I have updated the code to use `get_bundle_metadata `.

epicfaace:
@jzwang43 the tests are failing -- can you look into it?

image

I found this error message by looking at the logs that I downloaded that correspond to the failing run:

image

jzwang43:
> 
> 
> @jzwang43 the tests are failing -- can you look into it?
> 
> image
> 
> I found this error message by looking at the logs that I downloaded that correspond to the failing run:
> 
> image

The tests are passing now, looks like the failure was caused by some bundles' metadata not having `data_size` attribute for some reason.

epicfaace (approved):

Issue #3100, opened 2020-12-07
Enforce disk quota before upload starts


epicfaace:
It would be useful if a user is alerted that they don’t have enough disk quota before the file is uploaded (with `cl upload`), rather than after waiting a long time for the file to upload.

This is especially important if the file is really big, say, tens of GBs, as then the user wastes a lot of time waiting for an upload that then eventually fails.

adiprerepa:
Ideas:

concurrently start upload and checking of disk quota, and if quota exceeded, maybe stop responding on the server (or respond with an error) on that specific upload stream.

Alternatively the server could just respond on the unary quota call. In go we would pass in a context to both threads and then cancel the context if the quota was exceeded.

percyliang:
As a first pass, it might be wise to do a full check before uploading - just to keep things simple instead of fancy threading stuff.  This gives us the opportunity to report the size, number of files, and check the permissions, etc. before uploading begins, and therefore actually have a progress bar with an ETA, which would be cool.  This might be useful anyway since sometimes, it's easy to do a `cl upload` of a directory containing way too many files and a user will now be able to say "whoa, I didn't realize that was 5GB!".

adiprerepa:
do we want to do a full check for uploads of URLs too?

We would probably make a `HEAD` request and check the `Content-Length` header. The nice thing about `HEAD` requests here is that it does not depend on the actual size of the file - so the request would take the same amount of time for 1MB or 100GB.

percyliang:
Don't worry about URLs.  The URL downloading actually happens on the server.

PR #3379, opened 2021-04-03
enforce disk quota on bundle uploads

adiprerepa:
For  #3100


codalab/lib/bundle_util.py
```diff
@@ -338,3 +339,23 @@ def recurse(old_bundle_uuid):
                 )
 
     return plan
+
+
+def get_path_size(path):
```
epicfaace: I think we already have a get_path_size somewhere else -- can we reuse that?

epicfaace (commented):
Can you add a test?

adiprerepa:
aditya: write test in `test_cli.py`, use `cl uedit` to change disk quota, and hand it file that is less than that.

epicfaace:
Testing steps for test_cli.py:
1. `cl uedit` command to change disk quota to ~ 10 MB
2. Upload a > 10 MB file
3. Ensure it fails with the expected error

tests/cli/test_cli.py
```diff
@@ -739,6 +739,14 @@ def test_upload1(ctx):
         2 + 1, _run_command([cl, 'cat', uuid])
     )  # Directory listing with 2 headers lines and one file
 
+    # Upload a file that exceeds the disk quota
+    username = _run_command([cl, 'uinfo', '-f', 'user_name'])
+    prev_dq = _run_command([cl, 'uinfo', username, '-f', 'disk']).split(' ')[2]
+    _run_command([cl, 'uedit', username, '--disk-quota', '2'])
+    # expect to fail when we upload something more than 2 bytes
+    _run_command([cl, 'upload', 'codalab.png'], expected_exit_code=1)
+    _run_command([cl, 'uedit', username, '--disk-quota', prev_dq])
```
epicfaace: This command should be run regardless of whether this test fails or not. Could we maybe move this command to the "cleaning up" section of this file so that that is the behavior of it?

adiprerepa: how would it know the previous disk quota? Should I make it a global variable? Or just set it to the default of 100g

epicfaace: You can run `cl uinfo` to get the previous disk quota

codalab/lib/bundle_cli.py
```diff
@@ -1362,6 +1364,15 @@ def do_upload_command(self, args):
 
             # Canonicalize paths (e.g., removing trailing /)
             sources = [path_util.normalize(path) for path in args.path]
+            # if sources is an array check the cumulative size with the users disk quota
```
epicfaace: Not sure if "if sources is an array" is relevant? let's clarify this comment by saying something like "calculate size of sources"

tests/cli/test_cli.py
adiprerepa: if the test fails and it goes to cleanup, wouldn't the previous disk quota from `cl uinfo` be the wrong value? In this case, 2? Seeing as the default is 100g, I think we should just reset to that no matter what. Thoughts?

codalab/lib/bundle_cli.py
adiprerepa: done

tests/cli/test_cli.py
epicfaace: Sorry if I wasn't clear. I meant that you can run `cl uinfo` in the `__enter__` function to get the original disk quota (which would run before the tests start in the first place), store this disk quota in a class property, then restore the disk quota using `cl uedit` in `__exit__`. I think it's probably better to keep this general than hardcoding 100g in the code base here.

tests/cli/test_cli.py
```diff
@@ -376,6 +377,7 @@ def __enter__(self):
         temp_worksheet = _run_command([cl, 'new', random_name()])
         self.worksheets.append(temp_worksheet)
         _run_command([cl, 'work', temp_worksheet])
+        self.disk_quota = _run_command([cl, 'uinfo', 'codalab', '-f', 'disk']).split(' ')[2]
```
epicfaace: We can omit "codalab" to get the current user's info
```suggestion
        self.disk_quota = _run_command([cl, 'uinfo', '-f', 'disk']).split(' ')[2]
```

tests/cli/test_cli.py
```diff
@@ -425,7 +427,7 @@ def __exit__(self, exc_type, exc_value, tb):
                 _run_command([cl, 'rm', '--force', bundle])
 
         # Reset disk quota
-        _run_command([cl, 'uedit', 'codalab', '--disk-quota', '100g'])
+        _run_command([cl, 'uedit', 'codalab', '--disk-quota', self.disk_quota])
```
epicfaace: We should initialize self.disk_quota to `None` in `__enter__`, then run this command only if `self.disk_quota is not None`

tests/cli/test_cli.py
```diff
@@ -743,12 +745,11 @@ def test_upload1(ctx):
     )  # Directory listing with 2 headers lines and one file
 
     # Upload a file that exceeds the disk quota
-    username = _run_command([cl, 'uinfo', '-f', 'user_name'])
-    prev_dq = _run_command([cl, 'uinfo', username, '-f', 'disk']).split(' ')[2]
-    _run_command([cl, 'uedit', username, '--disk-quota', '2'])
+    prev_dq = _run_command([cl, 'uinfo', 'codalab', '-f', 'disk']).split(' ')[2]
```
epicfaace: we don't need this line anymore

tests/cli/test_cli.py
```diff
@@ -743,12 +745,11 @@ def test_upload1(ctx):
...
+    _run_command([cl, 'uedit', 'codalab', '--disk-quota', '2'])
     # expect to fail when we upload something more than 2 bytes
     _run_command([cl, 'upload', 'codalab.png'], expected_exit_code=1)
-    _run_command([cl, 'uedit', username, '--disk-quota', prev_dq])
+    _run_command([cl, 'uedit', 'codalab', '--disk-quota', prev_dq])
```
epicfaace: we don't need this line anymore

adiprerepa: so `__exit__` runs at the end of every module, not the entire suite of tests?

tests/cli/test_cli.py
adiprerepa: do you mean initialize `self.disk_quota` to `None` in `__init__`?

tests/cli/test_cli.py
adiprerepa: would I get the value through `ctx.disk_quota`?

tests/cli/test_cli.py
epicfaace: I believe so. You can test it out by printing something there and/or following the code for test modules.

tests/cli/test_cli.py
epicfaace: We won't be using this variable if we remove the line below.

tests/cli/test_cli.py
epicfaace: Yes

tests/cli/test_cli.py
adiprerepa: yep it does

tests/cli/test_cli.py
adiprerepa: ok, removed the variable

codalab/lib/bundle_cli.py
```diff
@@ -1362,6 +1364,15 @@ def do_upload_command(self, args):
 
             # Canonicalize paths (e.g., removing trailing /)
             sources = [path_util.normalize(path) for path in args.path]
+            # Calculate size of sources
+            total_bundle_size = sum([get_path_size(source) for source in sources])
+            user = client.fetch('users', client.fetch('user')['user_name'])
+            disk_left = user['disk_quota'] - user['disk_used']
+            if disk_left - total_bundle_size <= 0:
+                raise DiskQuotaExceededError(
+                    'Attempted to upload bundle of size %d with only %d remaining.'
```
epicfaace, line 1373:
```suggestion
                    'Attempted to upload bundle of size %d with only %d remaining in user\'s disk quota.'
```

epicfaace (approved):

codalab/lib/bundle_cli.py
epicfaace: @adiprerepa can you make this suggested change in another PR?

adiprerepa:
Sure

codalab/lib/bundle_cli.py
adiprerepa: #3409

PR #3355, opened 2021-03-19
Blob Storage: Use new forked version of ratarmount

epicfaace:
Use new forked version of ratarmount that upgrades indexed_gzip to use the new released v1.5.3 version, rather than my forked version. We can do this, now that https://github.com/codalab/codalab-worksheets/issues/3320 has been fully fixed. 

Full changes between these two versions: https://github.com/epicfaace/ratarmount/compare/71f6dd3b545b255e21a290f3d50c74019989c4ad...83d310d8efdf34351650ee98e7f44cf1c00dbf6e

image

PR #3380, opened 2021-04-03
bump version to v0.5.44

adiprerepa:
Bump new version

Full diff: https://github.com/codalab/codalab-worksheets/compare/v0.5.43...rc0.5.44

# Draft release notes

## Frontend
- Add meta description to home page (#3365)
- Make username clickable (#3366)
- Add UI Button for Content Block that User Can Specify Subpath Inside a Bundle (#3333)
- Fix target sizes for directories -- return data_size if user requests the whole bundle (#3372)
- Use useSWR as a cache for the search directive (#3353)

## Backend
- Make docker image status more granular with the percentage done (#3351)
- Disallow editing metadata of frozen worksheets; enable unfreezing worksheets (#3295)
- Implement GCP worker manager (#3341) 
- Blob Storage: Use new forked version of ratarmount (#3355)
- Blob Storage: Update MultiDiskBundleStore to use blob storage (#3306)

### DIFF
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -15,7 +15,7 @@ from enum import Enum
 
 # Increment this on master when ready to cut a release.
 # http://semver.org/
-CODALAB_VERSION = '0.5.43'
+CODALAB_VERSION = '0.5.44'
 BINARY_PLACEHOLDER = '<binary>'
 URLOPEN_TIMEOUT_SECONDS = int(os.environ.get('CODALAB_URLOPEN_TIMEOUT_SECONDS', 5 * 60))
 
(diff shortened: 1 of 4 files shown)


epicfaace:
Failing bundle on stress tests: https://worksheets-dev.codalab.org/bundles/0x25eb6936b59e463c95cf884caa0e4f7f

adiprerepa:
Looks like the issue is that sometimes the total image size is not available.

```
Exception in thread Thread-2:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/codalab/worker/docker_image_manager.py", line 253, in download
    / line['progressDetail']['total']
KeyError: 'total'
```

I think just adding a `hasattr` should fix it.

PR #3390, opened 2021-04-09
Fix LATEX rendering bug

### LINKED ISSUE #3283: Bundle detail view renders latex when it shouldn't
The bundle detail view renders latex when it shouldn't. [This bundle](https://worksheets.codalab.org/bundles/0xae2d1316bc0049d3bcbf59744cfb400f) renders as the following:

image

It should actually just show `python wilds/examples/run_expt.py --root_dir $HOME --log_dir $HOME --dataset amazon --algorithm ERM --seed 0`.

yuqijin:
The LATEX bug is caused by the 2 `$` signs in the text, so the MathJax treats the command between the $ signs as a Latex formula. 
The solution is when we want to use `$` sign in a markdown box, we should use `\$` instead.  The Markdown item will then render it as the `$`.

The bug should only exist for Markdown items. For regular bundles, no latex formula will be rendered. 

### Related issues

fixes #3283 

### Screenshots

**Raw text**
Screen Shot 2021-04-09 at 7 05 40 AM

**Before**
Screen Shot 2021-04-09 at 7 05 48 AM

**After**
Screen Shot 2021-04-09 at 7 06 05 AM

### Checklist


percyliang:
Just to confirm, latex should still be rendered the same in the worksheet markdown ($ should still work).  However, we don't want to apply MathJax latex rendering to anything else (table contents, display contents, etc.).

Also, $\text{TrainLoss}$ is not rendering properly above.

yuqijin:
> Just to confirm, latex should still be rendered the same in the worksheet markdown ($ should still work). However, we don't want to apply MathJax latex rendering to anything else (table contents, display contents, etc.).
> 
> Also, $\text{TrainLoss}$ is not rendering properly above.

Sorry, the incorrect rendering for $\text{TrainLoss}$ is due to the wrong latex grammar. I just updated the screenshots!

And yes, I inspected the codebase and found the MathJax is only used in the Markdown item. And this fix could render `$` properly now, while previously a single `$` could not be rendered correctly (see the Before example).

frontend/src/components/worksheets/items/MarkdownItem.js
```diff
@@ -191,6 +196,13 @@ class MarkdownItem extends React.Component {
             //   start = 0, inStart = 2, inEnd = 5, end = 7
             var start = text.indexOf('$', curr);
             if (start === -1) break; // No more math blocks
+            if (start > 0 && text[start - 1] === '\\') {
+                // \$ --> \\$
+                // Avoid the $ sign to be escaped
```
epicfaace:
```suggestion
                // Escape the $ sign to avoid inadvertent latex rendering
```

frontend/src/components/worksheets/items/MarkdownItem.js
```diff
@@ -191,6 +196,13 @@ class MarkdownItem extends React.Component {
             //   start = 0, inStart = 2, inEnd = 5, end = 7
             var start = text.indexOf('$', curr);
             if (start === -1) break; // No more math blocks
+            if (start > 0 && text[start - 1] === '\\') {
+                // \$ --> \\$
```
epicfaace, line 200: I'm a bit confused ... when would we find `\$` in the markdown cell text? I would have expected to only see `$`.

yuqijin: When a user types a text like `$ THIS IS PLAIN TEXT $`, the text will be treated as a latex formula, and no `$` will be displayed in the markdown cell. So if a user wants to display `$`, he should use type `\$` instead.

jzwang43 (approved):
Tested locally and looks good.

epicfaace (approved):

PR #3393, opened 2021-04-10
Run cl run --interactive as a non-root user 

### LINKED ISSUE #3051: Run cl run --interactive as a non-root user
**Describe the bug**
Run `cl run --interactive` as a non-root user to match the same behavior as CodaLab worker running jobs.

teetone:

Run interactive session as a non-root user.

### Related issues

Resolves #3051

### Checklist


codalab/lib/interactive_session.py
```diff
@@ -74,15 +75,15 @@ def __init__(
 
         self._docker_client = docker.from_env(timeout=InteractiveSession._MAX_SESSION_TIMEOUT)
         self._session_uuid = generate_uuid()
+        self._host_bundle_path: str = os.path.join(
```
epicfaace, line 78: Why move this command out of self.start?

codalab/lib/interactive_session.py
```diff
@@ -142,30 +143,34 @@ def get_docker_path(sub_path):
...
             ]
         )
-        command.append(
-            '-v {}:{}:rw'.format(
-                self._host_bash_history_path, InteractiveSession._BASH_HISTORY_CONTAINER_PATH
-            )
+        command.extend(
+            [
+                f'-v {self._host_bash_history_path}:{InteractiveSession._BASH_HISTORY_CONTAINER_PATH}:rw',
+                f'-v {self._host_bundle_path}:{container_working_directory}:rw',
```
epicfaace, line 170: Why did we not mount _host_bundle_path before?

codalab/lib/interactive_session.py
```diff
@@ -142,30 +143,34 @@ def get_docker_path(sub_path):
...
             'docker run',
             '-it',
             f'--name {name}',
-            f'-w {os.path.sep}{self._session_uuid}',
-            # TODO: figure out a way to run as a non-root user and be able to save the bash_history on any machine
-            # '-u 1',
+            f'-w {container_working_directory}',
+            f'-e HOME={container_working_directory}',
+            f'-e HISTFILE={InteractiveSession._BASH_HISTORY_CONTAINER_PATH}',
+            '-e PROMPT_COMMAND="history -a"',
```
epicfaace, line 157: Can you explain your approach here? Why is it needed to set PROMPT_COMMAND?

codalab/lib/interactive_session.py
```diff
@@ -23,15 +24,15 @@ class InteractiveSession:
...
+           exit the bash session. All the commands the user tried out will be stored at path `/usr/sbin/.bash_history`
            in the container.
         6. Copy `.bash_history` out to the host machine.
         7. Open an editor and allow the user to select and edit commands for the official run.
         8. Return the final command.
         9. Stop and remove the interactive session container.
     """
 
-    _BASH_HISTORY_CONTAINER_PATH = "/root/.bash_history"
+    _BASH_HISTORY_CONTAINER_PATH = "/usr/sbin/.bash_history"
```
percyliang, line 35: Is there anything special about this particular path?  More specifically, what property do we need to satisfy here...

teetone: It just has to be on a non-root path. I think `/usr/bin` should work too.

codalab/lib/interactive_session.py
teetone: This was for testing purposes. This static fields needs to be set in order to test `get_docker_run_command`.

codalab/lib/interactive_session.py
teetone: Setting the environment variable ensures that the history will always gets appended to `$HISTFILE`.

codalab/lib/interactive_session.py
teetone: Before the root user had access to the working directory without mounting. Since we are running as a non-root user, we have to mount this directory from the host so the non-root user has access in the container.

codalab/lib/interactive_session.py
epicfaace: Ok, so does this happen by default for root users without needing to set PROMPT_COMMAND?

teetone: I don't think was an issue before. The bash_history would get updated properly.

epicfaace (approved):
@teetone This all looks good -- do you think it would be possible to add an automated test for `cl run --interactive` in the future (probably could make as a separate issue / PR)? This code is pretty complex here, so I'd like to add a test so we can ensure nothing regresses. Maybe we could use something like https://ptyprocess.readthedocs.io/en/latest/ to make a simple test.

teetone:
> @teetone This all looks good -- do you think it would be possible to add an automated test for `cl run --interactive` in the future (probably could make as a separate issue / PR)? This code is pretty complex here, so I'd like to add a test so we can ensure nothing regresses. Maybe we could use something like https://ptyprocess.readthedocs.io/en/latest/ to make a simple test.

I think it's a bit difficult to test end-to-end because we expect user input in the container and most of the complexity is in how we start the interactive session container i.e. `get_docker_run_command`. I'm open to adding e2e tests in a separate PR though.

PR #3400, opened 2021-04-15
A More Central Call for Ajax Calls

### LINKED ISSUE #1037: Eliminate Jquery from codalab-worksheets
Move jquery show / hide functions -> use react state instead

### LINKED ISSUE #2223: [Code cleanup] A more central call for Ajax calls
It might better to wrap some ajax call into a function, not sure what a cleaner version would look like, maybe @epicfaace has more thoughts on it?

jzwang43:

We are looking for a cleaner way to call rest APIs and remove jquery from our code.

Since this is a pretty big pull request, I have moved some of the changes to pull request #3426 to make review easier, please review #3426 first.

All of the `$.ajax `  calls have been replaced with `axios`. Most of the api calls used are now in the `apiWrapper` class. This is a step towards removing jquery from our codebase. Initially I considered using fetch API but it turns out that fetch does not support `onUploadProgress` yet, which is used when uploading files/folders.

Since is mainly a refactor that involves many functionalities, so please play with the front end and help check if everything is working fine.

### Related issues

fixes #2223 
fixes #1037 

### Checklist


frontend/src/components/NavBar.js
```diff
@@ -69,21 +70,15 @@ class NavBar extends React.Component<{
...
 
     fetchName() {
-        $.ajax({
-            url: '/rest/user',
-            dataType: 'json',
-            cache: false,
-            type: 'GET',
-            success: (data) => {
+        fetchWrapper
+            .get('/rest/user')
```
epicfaace: Interesting approach. I was thinking it might be good to create different functions for each route? So, for example, a function called `getUser` would then call `fetchWrapper.get('/rest/user')`.

adiprerepa: as far as I can tell from these changes, we only call the `/rest/user` endpoint three times. does this constitute a function? might be good to keep it as general as possible

jzwang43: I agree with @adiprerepa , many api calls are only used once or twice. I was just focusing on removing $.ajax and replacing those with fetch.

epicfaace: Hmm, but I don't know if there's much value in replacing $.ajax with fetchWrapper, then -- we're not changing the complexity of the code at all; we're just renaming a single function.

I was thinking it would be good to create functions for each endpoint, because some endpoints may also require additional processing of input / output that can then be factored out as well -- for example, see the code that calls `/rest/cli/command` (appears [3x in the codebase](https://github.com/codalab/codalab-worksheets/blob/43286e6206b20dc8ff8ee75d94286ed2bc6c1306/frontend/src/components/worksheets/WorksheetTerminal.js)) and the `/contents/blob` endpoint that [appears 6x in the codebase](https://github.com/codalab/codalab-worksheets/search?l=JavaScript&q=%22contents%2Fblob%22). Both the input processing here and the additional headers (such as content-type, etc.) can be cleanly factored out and thus make the React code much simpler.

What do you think?

jzwang43: I think it makes more sense to move all related ajax calls to one place since we are looking into using SWR and removing jquery, I am working on this and have pushed some new changes.

frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -80,26 +81,19 @@ export class FileBrowser extends React.Component<
...
                 $('.file-browser').hide();
-            },
-        });
+            }
+        };
+        const errorHandler = () => {
+            this.setState({ fileBrowserData: {} });
+            $('.file-browser').hide();
+        };
+        apiWrapper.updateFileBrowser(this.props.uuid, folder_path, callback, errorHandler);
```
epicfaace: Return promises, so they can be called like this:

```suggestion
        apiWrapper.updateFileBrowser(this.props.uuid, folder_path).then(callback).catch(errorHandler);
```

jzwang43: This has been changed in the recent commits as suggested.

frontend/src/components/Bundle/Bundle.js
```diff
@@ -62,114 +62,80 @@ class Bundle extends React.Component<
...
-            })
-            .fail(function(xhr, status, err) {
-                this.setState({
-                    bundleInfo: null,
-                    fileContents: null,
-                    stdout: null,
-                    stderr: null,
-                    errorMessages: this.state.errorMessages.concat([xhr.responseText]),
-                });
+        let callback = (response) => {
```
adiprerepa, line 63: Is there any value in it being a variable? Why not inline it inside the `.then()`

frontend/src/components/Bundle/Bundle.js
```diff
@@ -62,114 +62,80 @@ class Bundle extends React.Component<
...
-                            stderr: null,
-                            errorMessages: this.state.errorMessages.concat([xhr.responseText]),
-                        });
-                    } else {
-                        // If contents aren't available yet, then also clear stdout and stderr.
-                        this.setState({ fileContents: null, stdout: null, stderr: null });
-                    }
-                }.bind(this),
-            );
+        callback = async (response) => {
```
adiprerepa, line 94: same comment as above

adiprerepa: actually, not this one, it's used twice

frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -80,26 +81,22 @@ export class FileBrowser extends React.Component<
...
+            }
+        };
+        const errorHandler = (error) => {
+            console.error(error);
+            this.setState({ fileBrowserData: {} });
+            $('.file-browser').hide();
+        };
+        updateFileBrowser(this.props.uuid, folder_path)
+            .then(callback)
+            .catch(errorHandler);
```
adiprerepa, line 102: inline?

frontend/src/components/NavBar.js
```diff
@@ -69,21 +69,13 @@ class NavBar extends React.Component<{
...
-            },
-            error: (xhr, status, err) => {
-                console.error(xhr.responseText);
-            },
-        });
+        const callback = (data) => {
+            const userInfo = data.data.attributes;
+            userInfo.user_id = data.data.id;
+            this.fetchImg(userInfo.avatar_id);
+            this.setState({ userInfo: userInfo, newWorksheetName: `${userInfo.user_name}-` });
```
adiprerepa, line 76: why was there a dash (`{userInfo.user_name}-`) in the first place?

frontend/src/util/apiWrapper.js
```diff
@@ -0,0 +1,178 @@
+import { Semaphore } from 'await-semaphore';
+import axios from 'axios';
+import { createDefaultBundleName, pathIsArchive, getArchiveExt } from './worksheet_utils';
+
+export const get = (url, params) => {
+    const requestOptions = {
+        params,
+    };
+    return axios.get(url, requestOptions).then((res) => res.data);
```
adiprerepa, line 9: is there value in returning the `Promise` without resolving it?

frontend/src/util/apiWrapper.js
```diff
@@ -0,0 +1,178 @@
...
+    const url = '/rest/cli/command';
+    return post(url, {
+        worksheet_uuid: worksheet_uuid || null,
+        command: command,
+        autocomplete: true,
+    });
+};
+
+// Limit concurrent requests for async resolving items
+const MAX_CONCURRENT_REQUESTS = 3;
```
adiprerepa, line 81: how did you arrive at this number?

frontend/src/util/apiWrapper.js
```diff
@@ -0,0 +1,178 @@
...
+const semaphore = new Semaphore(MAX_CONCURRENT_REQUESTS);
+
+export const fetchAsyncBundleContents = async ({ contents }) => {
+    // used in table and record items
+    return semaphore.use(async () => {
```
adiprerepa, line 86: probably my naivete but why do we want to use this as opposed to https://www.npmjs.com/package/await-semaphore#promise-style-javascript

Might be more consistent with the rest of the codebase, maybe not.

adiprerepa (commented):
not really a javascript guy but thought I would take a shot at this

frontend/src/components/Bundle/Bundle.js
jzwang43: I think it is cleaner this way otherwise the .then().catch() is going to be super long.

frontend/src/components/NavBar.js
jzwang43: I was just refactoring the code, did not change anything functionality wise.

frontend/src/util/apiWrapper.js
jzwang43: I was just refactoring the code, did not change anything functionality wise. It was written this way so for now can just keep it the way it was.

frontend/src/components/Bundle/Bundle.js
jzwang43: The two functions are to handle two different API calls here I think.

frontend/src/util/apiWrapper.js
jzwang43: This automatically handles acquire/release, the behavior should be the same as other styles.

frontend/src/util/apiWrapper.js
jzwang43: axios.get is asynchronous so everything needs to be chained with .then() and .catch(), when the axios call is finished it will automatically resolve with response. We were thinking about doing async await for everything but that will probably be later and for a separate ticket.

frontend/src/components/NavBar.js
jzwang43: I think this is the default name since by default we want the name of the new worksheet to be username-xxxx, so the user just need to enter what's after the dash.

frontend/src/components/Bundle/Bundle.js
epicfaace: Actually, the cleanest way to do this would be to use async-await. Then you could do something like this:

```
async refreshBundle = () => {
   try {
    const response = await fetchBundleMetadata(this.props.uuid);
    const bundleInfo = new JsonApiDataStore().sync(response);
    // ... rest of callback here
   } catch (error) {
     if (error.response.status === 404) {
     // ... rest of error handler here
     }
   }
}
```

Note how this code is much cleaner because you don't have to define separate callback functions at all and just have to use a try-catch pattern.

@jzwang43 , I was thinking it wasn't worth doing async-await for now so we could keep the changes as small as possible. We could probably do this in a later refactor, though. But let me know what you think about this.

frontend/src/components/worksheets/WorksheetTerminal.js
```diff
@@ -186,34 +177,25 @@ class WorksheetTerminal extends React.Component {
...
-                deferred.resolve(data.completions);
-            },
-            error: function(jqHXR, status, error) {
-                console.error(jqHXR.responseText);
-                deferred.reject();
-            },
+        return new Promise((resolve, reject) => {
+            apiWrapper
+                .completeCommand(command, this.props.ws.info.uuid)
+                .then((data) => resolve(data.completions))
```
epicfaace: We don't need to construct a new Promise here. We can just remove that and change this line to

```suggestion
                .then((data) => data.completions)
```

and it should automatically return a promise.

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -119,19 +124,9 @@ class Worksheet extends React.Component {
...
         };
 
-        $.ajax({
-            type: 'GET',
-            url:
-                '/rest/interpret/worksheet/' +
-                this.state.ws.uuid +
-                '?' +
-                queryString.stringify(queryParams),
-            // TODO: migrate to using main API
```
epicfaace, line 129: Is this TODO no longer relevant? (I believe so, but just wanted to make sure)

frontend/src/components/Bundle/Bundle.js
jzwang43: I think we should do it at a later time since this pull request is already very big.

frontend/src/components/worksheets/Worksheet/Worksheet.js
jzwang43: I think it's no longer relevant, the API call works fine as is.

frontend/src/components/worksheets/WorksheetTerminal.js
jzwang43: suggestion applied.

jzwang43: suggestion applied.

frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -527,11 +525,12 @@ export const FileBrowserLite = ({ uuid, startCollapsed, isRunningBundle, bundle_
         currentWorkingDirectory +
         '?' +
         new URLSearchParams({ depth: 1 });
-
+    console.log('fileBrower');
```
yuqijin: nit: remove debug log statement

frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -527,11 +525,12 @@ export const FileBrowserLite = ({ uuid, startCollapsed, isRunningBundle, bundle_
...
     useSWR(url, fetcher, {
         revalidateOnMount: true,
         refreshInterval: isRunningBundle ? 4000 : 0,
         onSuccess: (data, key, config) => {
+            console.log(data);
```
yuqijin: nit: remove log statement

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -174,111 +124,91 @@ class NewUpload extends React.Component<{
...
+        } catch (error) {
             this.clearProgress();
-            alert(
-                createAlertText(
-                    url,
-                    error.responseText,
-                    'refresh and try again.',
-                ),
-            );
+            console.log(error);
```
yuqijin: nit: remove it

frontend/src/components/worksheets/WorksheetTerminal.js
```diff
@@ -50,6 +51,7 @@ class WorksheetTerminal extends React.Component {
                         self.renderHyperlinks(data.structured_result.refs);
                     }
                 } catch (error) {
+                    console.error(error);
```
yuqijin, line 54: nit: remove

yuqijin (approved):
The PR lgtm, thanks for your work, Jizhen! But don't forget to remove the redundant log statements before merging it.

frontend/src/components/FileBrowser/FileBrowser.js
jzwang43: Thanks for pointing out, these have been removed.

PR #3412, opened 2021-04-18
Simplify uploads by removing local path support

epicfaace:

- Simplify uploads by removing local path support -- I did this to simplify the amount of changes that need to go into blob storage (#3298 ). From what I can tell, local paths aren't used at all at the moment when uploading (when uploading from the CLI, you either upload strings, URLs, or fileobj's; and workers upload a fileobj directly with BundleServiceClient). Local paths only seem to be used in upload_manager_test.py, so we should be able to safely remove them.
- follow_symlinks, exclude_patterns, and remove_sources have also been removed from UploadManager, because these were only used during local path support. Note that `--follow-symlinks` and `--exclude-patterns` still exist and work properly on the Bundle CLI (they are used when zipping up directories from the CLI). It's just that their existence in UploadManager is no longer functional and can be removed.
- Removed logic around using `_ignore_file_in_archive` to determine whether to simplify archives in UploadManager (see comment below on upload_manager.py).
- I've added additional tests (unit and CLI tests) that exercise more of the functionality of UploadManager.

Although this PR is technically a breaking change by removing functionality, all the removed functionality seems to be either unused / unintentional, and removing them would greatly simplify the implementation of Blob Storage.

### DIFF
--- a/codalab/lib/path_util.py
+++ b/codalab/lib/path_util.py
@@ -235,12 +235,11 @@ def hash_file_contents(path):
 ################################################################################
 
 
-def copy(source_path, dest_path, follow_symlinks=False, exclude_patterns=None):
+def copy(source_path, dest_path, follow_symlinks=False):
     """
     Copy |source_path| to |dest_path|.
     Assume dest_path doesn't exist.
     |follow_symlinks|: whether to follow symlinks
-    |exclude_patterns|: patterns to not copy
     Note: this only works in Linux.
     """
     if os.path.exists(dest_path):
@@ -266,9 +265,6 @@ def copy(source_path, dest_path, follow_symlinks=False, exclude_patterns=None):
             + ('/' if not os.path.islink(source_path) and os.path.isdir(source_path) else ''),
             dest_path,
         ]
-        if exclude_patterns is not None:
-            for pattern in exclude_patterns:
-                command.extend(['--exclude', pattern])
         if subprocess.call(command) != 0:
             raise path_error('Unable to copy %s to' % source_path, dest_path)
 
--- a/codalab/lib/path_util.py
+++ b/codalab/lib/path_util.py
@@ -21,6 +21,7 @@ import os
 import shutil
 import subprocess
 import sys
+from typing import Optional
 
 from codalab.common import precondition, UsageError
 from codalab.lib import file_util
@@ -235,7 +236,7 @@ def hash_file_contents(path):
 ################################################################################
 
 
-def copy(source_path, dest_path, follow_symlinks=False):
+def copy(source_path: str, dest_path: str, follow_symlinks: Optional[bool] = False):
     """
     Copy |source_path| to |dest_path|.
     Assume dest_path doesn't exist.
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -24,7 +24,7 @@ class UploadManager(object):
 
     def upload_to_bundle_store(
         self,
-        bundle: Any,
+        bundle: Bundle,
         sources: List[Any],
         git: bool,
         unpack: bool,
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -120,8 +120,7 @@ class UploadManager(object):
     def _simplify_archive(self, path):
         """
         Modifies |path| in place: If |path| is a directory containing exactly
-        one file / directory that is not ignored, then replace |path| with that
-        file / directory.
+        one file / directory, then replace |path| with that file / directory.
         """
         if not os.path.isdir(path):
             return
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -125,7 +125,7 @@ class UploadManager(object):
         if not os.path.isdir(path):
             return
 
-        files = [f for f in os.listdir(path)]
+        files = os.listdir(path)
         if len(files) == 1:
             self._simplify_directory(path, files[0])
 
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -17,19 +17,14 @@ class UploadManager(object):
         from codalab.lib import zip_util
 
         # exclude these patterns by default
-        DEFAULT_EXCLUDE_PATTERNS = ['.DS_Store', '__MACOSX', '^\._.*']
         self._bundle_model = bundle_model
         self._bundle_store = bundle_store
-        self._default_exclude_patterns = DEFAULT_EXCLUDE_PATTERNS
         self.zip_util = zip_util
 
     def upload_to_bundle_store(
         self,
         bundle,
         sources,
-        follow_symlinks,
-        exclude_patterns,
-        remove_sources,
         git,
         unpack,
         simplify_archives,
@@ -40,11 +35,6 @@ class UploadManager(object):
 
         |sources|: specifies the locations of the contents to upload. Each element is
                    either a URL, a local path or a tuple (filename, binary file-like object).
-        |follow_symlinks|: for local path(s), whether to follow (resolve) symlinks,
-                           but only if remove_sources is False.
-        |exclude_patterns|: for local path(s), don't upload these patterns (e.g., *.o),
-                            but only if remove_sources is False.
-        |remove_sources|: for local path(s), whether |sources| should be removed
         |git|: for URLs, whether |source| is a git repo to clone.
         |unpack|: for each source in |sources|, whether to unpack it if it's an archive.
         |simplify_archives|: whether to simplify unpacked archives so that if they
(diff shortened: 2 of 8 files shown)


codalab/lib/upload_manager.py
```diff
@@ -168,14 +126,10 @@ def _simplify_archive(self, path):
         if not os.path.isdir(path):
             return
 
-        files = [f for f in os.listdir(path) if not self._ignore_file_in_archive(f)]
+        files = [f for f in os.listdir(path)]
```
epicfaace, line 128: I went ahead and removed this logic, as well. Although it's technically a breaking change in how CodaLab treats certain files, I don't think we want to keep it because the current behavior is problematic.

Previous behavior (current behavior on master):
- If you upload an archive to CodaLab with exactly two files: `README` and `.DS_Store`, CodaLab will just treat the bundle as a single file called `README`.

New behavior (with this PR):
- If you upload an archive to CodaLab with exactly two files: `README` and `.DS_Store`, CodaLab will treat the bundle as a folder with two files: `README` and `.DS_Store`.

The previous behavior is problematic because the user **has already included both files** in the fileobj they are uploading. Presumably, if they really wanted to ignore `.DS_Store`, they would have done so with `cl upload --exclude-patterns`, and the fileobj wouldn't have included `.DS_Store` in the first place.

Because fileobj _does_ include both these files, the new behavior in this PR makes the most sense.

I removed this logic now so I wouldn't have to replicate it for Blob Storage (thus simplifying blob storage implementation).

This scenario is also covered in the `test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive` test case.

codalab/lib/path_util.py
```diff
@@ -235,12 +235,11 @@ def hash_file_contents(path):
 ################################################################################
 
 
-def copy(source_path, dest_path, follow_symlinks=False, exclude_patterns=None):
+def copy(source_path, dest_path, follow_symlinks=False):
```
epicfaace: With this PR, `exclude_patterns` is no longer used in `copy`, so I removed it to simplify `copy`'s implementation.

codalab/lib/upload_manager.py
```diff
@@ -159,27 +127,23 @@ def _unpack_fileobj(self, source_filename, source_fileobj, dest_path, simplify_a
...
         Modifies |path| in place: If |path| is a directory containing exactly
-        one file / directory that is not ignored, then replace |path| with that
-        file / directory.
+        one file / directory, then replace |path| with that file / directory.
         """
         if not os.path.isdir(path):
             return
 
-        files = [f for f in os.listdir(path) if not self._ignore_file_in_archive(f)]
+        files = os.listdir(path)
```
epicfaace, line 139: I went ahead and removed this logic, as well. Although it's technically a breaking change in how CodaLab treats certain bundles, I don't think we want to keep it because the current behavior is problematic.

Current behavior:
- If you upload an archive to CodaLab with exactly two files: `README` and `.DS_Store`, CodaLab will just treat the bundle as a single file called `README`.
- If you upload an archive to CodaLab with exactly three files: `README`, `README2`, and `.DS_Store`, CodaLab will treat the bundle as a folder with three files: `README`, `README2`, and `.DS_Store`.

New behavior (with this PR):
- If you upload an archive to CodaLab with exactly two files: `README` and `.DS_Store`, CodaLab will treat the bundle as a folder with two files: `README` and `.DS_Store`.
- If you upload an archive to CodaLab with exactly three files: `README`, `README2`, and `.DS_Store`, CodaLab will treat the bundle as a folder with three files: `README`, `README2`, and `.DS_Store`.

For the two-file scenario, the current behavior is problematic because the user **has already included the .DS_Store file** in the fileobj they are uploading. Presumably, if they really wanted to ignore `.DS_Store`, they would have done so with `cl upload --exclude-patterns`, and the fileobj wouldn't have included `.DS_Store` in the first place. Because fileobj _does_ include both these files, the new behavior in this PR makes the most sense.

Moreover, note the inconsistency in the current behavior -- `.DS_Store` is removed from the final bundle _only in_ the 2-files case, but it is included in the 3-files case. This also shows why the current behavior isn't ideal.

I removed this logic now so I wouldn't have to replicate this kind of problematic / unnecessary logic for Blob Storage (thus simplifying blob storage implementation).

These scenarios are also covered in the `test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive` and `test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive_2` test cases.

nelson-liu:
> follow_symlinks, exclude_patterns, and remove_sources have also been removed from UploadManager, because these were only used during local path support. Note that --follow-symlinks and --exclude-patterns still exist and work properly on the Bundle CLI (they are used when zipping up directories from the CLI). It's just that their existence in UploadManager is no longer functional and can be removed.

I'm a bit confused---I use the `exclude_patterns` metadata key of bundles and it seems to work on local paths, is this not the case? Sorry if I'm missing something obvious here.

codalab/lib/upload_manager.py
```diff
@@ -129,18 +97,18 @@ def upload_to_bundle_store(
             raise
 
     def _interpret_source(self, source):
```
percyliang: Add type annotations

codalab/lib/upload_manager.py
percyliang: Yeah, this function is annoyingly complicated, so it's great to simplify.  I think what you're doing makes sense...I can't remember why we had local file paths.  However, this is a potentially disruptive change, so we should double check and test everything before deploying.

epicfaace:
> I'm a bit confused---I use the `exclude_patterns` metadata key of bundles and it seems to work on local paths, is this not the case? Sorry if I'm missing something obvious here.

@nelson-liu The `exclude_patterns` metadata key should continue to work. That key is only used by the worker in https://github.com/codalab/codalab-worksheets/blob/c2d8960bbdcda127e0ea8800b6ecd4b7849c48f7/codalab/worker/worker_run_state.py#L626 and https://github.com/codalab/codalab-worksheets/blob/c2d8960bbdcda127e0ea8800b6ecd4b7849c48f7/codalab/worker/worker_run_state.py#L670.

You'll see that those instances of `exclude_patterns` are only used for local operations, and they aren't actually passed on to the upload endpoint and thus `UploadManager.upload_to_bundle_store`.

Confusingly, `UploadManager.upload_to_bundle_store` has its own `exclude_patterns` parameter, which is only used for files with a local path (here, "local path" refers to "a path local to the REST server"). However, because files with a local path are only used in tests and not used in normal CodaLab usage, it's good (and removes duplication / confusion between the different kinds of `exclude_patterns` variables) to remove it here.

codalab/lib/upload_manager.py
epicfaace: Done

epicfaace:
@teetone @nelson-liu @percyliang could I get another review from one of you on this?

codalab/lib/upload_manager.py
epicfaace: Ah, I think I see why this might have been implemented in the first place. If you zip up a single file on a Mac from the Finder, it automatically adds a `__MACOSX` directory to the zip archive. I guess CodaLab wanted to ignore that directory so it could just treat such archives as a single file, rather than storing them as a bundle directory on CodaLab.

Nevertheless, the inconsistency I mentioned above still persists with `__MACOSX` directories, so I think we're still good to remove it.

tests/unit/lib/upload_manager_test.py
```diff
@@ -23,110 +29,113 @@ def update_bundle(self, *args, **kwargs):
...
+        the archive should not be simplified because we have more than one file in the archive.
+        """
+        source = os.path.join(self.temp_dir, 'source_dir')
+        os.mkdir(source)
+        self.write_string_to_file('testing', os.path.join(source, 'README'))
+        self.write_string_to_file('testing', os.path.join(source, '.DS_Store'))
+        self.do_upload([('source.tar.gz', tar_gzip_directory(source))])
+        self.assertEqual(['.DS_Store', 'README'], sorted(os.listdir(self.bundle_location)))
+
+    def test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive_2(self):
```
adiprerepa, line 83: how is this any different in purpose than `test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive`?

adiprerepa:
ff

tests/unit/lib/upload_manager_test.py
```diff
@@ -23,110 +29,113 @@ def update_bundle(self, *args, **kwargs):
...
+        )
+        self.assertEqual(
+            ['bc-1.06.95', 'source1', 'source2'], sorted(os.listdir(self.bundle_location))
+        )
         self.check_file_contains_string(os.path.join(self.bundle_location, 'source1'), 'testing1')
         self.check_file_contains_string(os.path.join(self.bundle_location, 'source2'), 'testing2')
+        self.assertIn('README', os.listdir(os.path.join(self.bundle_location, 'bc-1.06.95')))
+
+    def test_url_git(self):
+        self.do_upload(['https://github.com/codalab/test'], git=True)
```
adiprerepa, line 138: this url doesn't exist, right? what's the purpose?

adiprerepa (approved):
lgtm

funny - this pr enables #3432 which enables #3434 which enables #3435

tests/unit/lib/upload_manager_test.py
epicfaace: Actually, it exists -- this test hits the actual network, as I couldn't find an easy way to mock a git repository url. See https://github.com/codalab/test

tests/unit/lib/upload_manager_test.py
epicfaace: It has 3 files instead of 2 files. Before this PR, the archive would be simplified in test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive and not be simplified in test_fileobj_single_tar_gz_with_dsstore_should_not_simplify_archive_2. With this PR, the archive would not be simplified in both cases.

PR #3423, opened 2021-04-23
Fix jQuery invalid selector bug for the search bar

### LINKED ISSUE #3403: Invalid selector error occurred when using search bar on the profile page
**Describe the bug**
Only occurred at localhost.
Screen Shot 2021-04-15 at 9 22 25 PM

**To Reproduce**
1. Go to MY PROFILE page
2. Click on the search bar
3. See error

yuqijin:

Add check of whether the worksheet terminal exists on the current page

### Related issues

fixes #3403 

### Checklist

### DIFF
--- a/frontend/src/components/NavBar.js
+++ b/frontend/src/components/NavBar.js
@@ -204,15 +204,10 @@ class NavBar extends React.Component<{
 
     handleSearchFocus = () => {
         // Disable the terminal to avoid the search bar text being mirrored in the terminal
-        if ($('#command_line').length > 0) {
-            if (
-                $('#command_line')
-                    .terminal()
-                    .enabled()
-            ) {
-                $('#command_line')
-                    .terminal()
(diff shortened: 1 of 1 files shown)


frontend/src/components/NavBar.js
```diff
@@ -204,14 +204,16 @@ class NavBar extends React.Component<{
...
         // Disable the terminal to avoid the search bar text being mirrored in the terminal
-        if (
-            $('#command_line')
-                .terminal()
-                .enabled()
-        ) {
-            $('#command_line')
-                .terminal()
-                .focus(false);
+        if ($('#command_line').length > 0) {
```
epicfaace: Let's add `const $cmd = $('#command_line');` at the beginning and rewrite the code in terms of `$cmd` to make this code cleaner.

epicfaace (approved):

PR #3424, opened 2021-04-23
Fix installation of "cl" command with git dependencies

epicfaace:

Stress tests were failing for the rc0.5.45 release because the git dependency (fork of ratarmount) was not properly added to setup.py.

I've also added a check in CI to ensure this doesn't regress in the future.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,14 @@ jobs:
           restore-keys: |
             pip-
       - run: ./pre-commit.sh && git diff --exit-code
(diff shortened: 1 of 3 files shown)


epicfaace:
@teetone can you take a look so we can add this to the latest rc?

teetone (approved):

epicfaace:
@teetone I've also changed the requirement to point to a forked version of ratarmount under codalab rather than under my personal repo, epicfaace. This is probably best, in case something happens to my personal account, so it's best to keep everything under the codalab account.

PR #3426, opened 2021-04-23
Move API Calls to a Central Place

### LINKED ISSUE #2223: [Code cleanup] A more central call for Ajax calls

jzwang43:

This pull request is related to #3400, the pull request has been broken down to make review easier.

### Related issues

fixes #2223 

### Checklist

### DIFF
--- a/frontend/src/components/Bundle/Bundle.js
+++ b/frontend/src/components/Bundle/Bundle.js
@@ -69,7 +69,7 @@ class Bundle extends React.Component<
         };
 
         let errorHandler = (error) => {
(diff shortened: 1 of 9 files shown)


jzwang43:
I will go ahead and merge this into fixes/2223-ajax for now since review is going on on that pull request #3400. And the changes here are closely related to the main pull request. It makes more sense to just review #3400 instead.
Next time I will break down the larger pull request better so they can be reviewed separately.

PR #3432, opened 2021-04-25
Remove support for uploads of multiple URLs

epicfaace:

This PR effectively removes support for uploads of multiple URLs, greatly simplifying the upload logic, which is needed for the current implementation plan of Blob Storage.

Essentially, it changes `sources` in `upload_to_bundle_store` to just be a single value `source`. Note that you can still upload multiple files from the CLI, as normal (because in those cases, those files are tarred up from the client and show up as a single "source"). **This change only affects uploading multiple URLs.**

### DIFF
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -52,8 +52,7 @@ class UploadManager(object):
         bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
         try:
             path_util.make_directory(bundle_path)
-            # Note that for uploads with a single source, the directory
-            # structure is simplified at the end.
+            # Note that the directory structure is simplified at the end.
             is_url, is_fileobj, filename = self._interpret_source(source)
             source_output_path = os.path.join(bundle_path, filename)
             if is_url:
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -24,7 +24,7 @@ class UploadManager(object):
 
     def upload_to_bundle_store(
         self,
-        bundle: Any,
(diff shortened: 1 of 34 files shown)


codalab/lib/upload_manager.py
```diff
@@ -36,57 +36,52 @@ def upload_to_bundle_store(
...
+                            remove_source=True,
                             simplify_archive=simplify_archives,
                         )
-                    else:
-                        with open(source_output_path, 'wb') as out:
-                            shutil.copyfileobj(cast(IO, source[1]), out)
-
-            if len(sources) == 1:
-                self._simplify_directory(bundle_path)
+            elif is_fileobj:
```
adiprerepa, line 72: if we remove support for multiple URLs, wouldn't we be effectively only allowing one fileobj to be uploaded? since the `for source in sources` doesn't exist - where is the multiple filenames logic being handled?

adiprerepa (approved):

adiprerepa (approved):

adiprerepa:
OK, I should stop using the jetbrains github integration. super janky. 
pretend the "approved" is "reviewed"

codalab/lib/upload_manager.py
adiprerepa: oh, is it tarred into a single file?

epicfaace: Yeah, the bundle cli tars it into a single file.

epicfaace:
@adiprerepa you can leave another review with "request changes" to cancel out your approval.

adiprerepa (approved):
LGTM - awesome sauce

PR #3434, opened 2021-04-25
Simplify uploads by extracting directly to bundle_path, streaming downloaded URLs

epicfaace:

Merge this only after https://github.com/codalab/codalab-worksheets/pull/3432 is merged. This PR greatly simplifies file structure during upload, in preparation for upload changes needed during Blob Storage.

This PR **does not make any breaking changes.** The file `tests/unit/lib/upload_manager_test.py` pretty thoroughly covers UploadManager's functionality through unit tests, and all tests still pass.

#### Changes in this PR

- Essentially, previously we would extract each source in `sources` to the path `source_output_path = os.path.join(bundle_path, filename)`. In the vast majority of cases when we had a single source, we would then have to call `self. _simplify_directory` and move the extracted archive back to `bundle_path`. Now, because we only have one source due to https://github.com/codalab/codalab-worksheets/pull/3432, we can simplify this logic in this PR a lot by just directly extracting the single source to `bundle_path` directly, skipping the renaming / directly simplifying process entirely.
- Previously, when uploading a URL that is an archive, the server would first download it to disk, then extract it, then finally delete the archive file itself. With this PR, we simplify that process by replacing `file_util.download_url` with `urlopen_with_retry` directly. This saves disk space on the server and ensures that the URL uploads go through the same code path as direct fileobj uploads. Both are prerequisites for the streaming uploads needed for Blob Storage.


codalab/lib/zip_util.py
```diff
@@ -47,11 +48,14 @@ def strip_archive_ext(path):
     raise UsageError('Not an archive: %s' % path)
 
 
-def unpack(ext, source, dest_path):
-    """
-    Unpack the archive |source| to |dest_path|.
-    Note: |source| can be a file handle or a path.
-    |ext| contains the extension of the archive.
+def unpack(ext: str, source: Union[str, IO[bytes]], dest_path):
```
adiprerepa: should we have a type annotation for `dest_path`?

```suggestion
def unpack(ext: str, source: Union[str, IO[bytes]], dest_path: str):
```

codalab/lib/zip_util.py
```diff
@@ -77,7 +81,7 @@ def unpack(ext, source, dest_path):
         raise UsageError('Invalid archive upload.')
     finally:
         if close_source:
-            source.close()
+            cast(IO[bytes], source).close()
```
adiprerepa, line 84: what if `source` is a `str`? I'm not familiar with how python union type casting works.

adiprerepa (commented):
general lgtm

codalab/lib/zip_util.py
epicfaace: source is actually always `IO[bytes]` if close_source is true. The python type checker isn't smart enough to infer that, so I had to cast it.

adiprerepa: ok, makes sense. thanks!

adiprerepa (approved):
lgtm

PR #3435, opened 2021-04-25
Remove simplify_archive

### LINKED ISSUE #3439: Implement Blob Storage uploading breaking changes
Implement Blob Storage uploading changes. This will involve 1) making some breaking changes that apply to both non-Blob and Blob uploads, and 2) ensuring these uploading changes work for Blob Storage as well.

The table summarizing these changes, taken from [this doc](https://docs.google.com/document/d/1l4kOqi9irBjOApmn4E6vlzsjAXDJbetIyVw8gMRHrpU/edit#):

Scenario | Sample command | Description
-- | -- | --
Single file | cl upload item.txt | The file is compressed to a .gz file on Blob Storage.
Single archive / zip | cl upload item.gzcl upload item.tar.gzcl upload item.zip | The archive is unpacked and then saved to Blob Storage as either a .tar.gz (folder) or .gz (file).
Multiple files / directories | cl upload a.txt b.txt dir | The files / directories are added to a .tar.gz file on Blob Storage.
Multiple archives | cl upload item.gz dir.tar.gz | Breaking change: Archives are not unpacked when multiple archives are specified. They are treated as files in the “multiple files / directories” scenario.
Single URL - file | cl upload https://a.txt | The file is streamed and compressed, like the “single file” scenario above.
Single URL - archive | cl upload https://a.gzcl upload https://a.tar.gzcl upload https://a.zip | The archive is unpacked, similar to the “single archive / zip” scenario above.
Multiple URLs | cl upload https://a.txt https://a.zip | Breaking change: This scenario is no longer supported. Entering multiple URLs into cl upload will throw an error.

epicfaace:

Merge this only after https://github.com/codalab/codalab-worksheets/pull/3434 is merged. This is a **breaking change** that removes the simplify_archive functionality -- this simplifies the upload process even further in preparation for usage with Blob Storage.

The simplify archives functionality essentially meant that if an archive that was unpacked to create a bundle just had a single file, the bundle would just be that single file rather than that extracted folder with one item.

Practically, this change only affects the following workflows:

- If someone creates a file `a.tar.gz` with one file / folder `X` in it, running `cl upload a.tar.gz` previously would create a bundle with just `X` at the top-level of it. Now, the bundle will be a directory bundle with `X` in it.
- The same as above applies to URLs, such as `https://a.tar.gz`.

The "worker uploading bundle results" workflow is not affected, because these archives will always have greater than one file (they'll at least have stdout and stderr), so the archives are never simplified anyway.

For example, consider the archive. This archive contains all its files within a single "bc-1.06.95" directory:

image

Previously when running `cl upload http://alpha.gnu.org/gnu/bc/bc-1.06.95.tar.bz2`, it would simplify the archive by default and the bundle contents would look like this:

image

With this PR, running `cl upload http://alpha.gnu.org/gnu/bc/bc-1.06.95.tar.bz2` will give a different output, with the bundle contents all included within the "bc-1.06.95" directory:

image

### DIFF
--- a/codalab/lib/upload_manager.py
+++ b/codalab/lib/upload_manager.py
@@ -1,9 +1,9 @@
 import os
 import shutil
-from typing import Optional, Union, Tuple, IO, cast
+from typing import Union, Tuple, IO, cast
 
 from codalab.common import UsageError, StorageType, urlopen_with_retry
-from codalab.lib import crypt_util, file_util, path_util
+from codalab.lib import file_util, path_util
 from codalab.objects.bundle import Bundle
 
 Source = Union[str, Tuple[str, IO[bytes]]]
@@ -24,24 +24,16 @@ class UploadManager(object):
         self.zip_util = zip_util
 
     def upload_to_bundle_store(
-        self,
-        bundle: Bundle,
-        source: Source,
-        git: bool,
-        unpack: bool,
-        simplify_archives: bool,
-        use_azure_blob_beta: bool,
+        self, bundle: Bundle, source: Source, git: bool, unpack: bool, use_azure_blob_beta: bool,
     ):
         """
         Uploads contents for the given bundle to the bundle store.
 
+        |bundle|: specifies the bundle associated with the contents to upload.
         |source|: specifies the location of the contents to upload. Each element is
                    either a URL or a tuple (filename, binary file-like object).
         |git|: for URLs, whether |source| is a git repo to clone.
         |unpack|: whether to unpack |source| if it's an archive.
-        |simplify_archives|: whether to simplify unpacked archives so that if they
-                             contain a single file, the final path is just that file,
-                             not a directory containing that file.
         |use_azure_blob_beta|: whether to use Azure Blob Storage.
 
         Exceptions:
@@ -61,9 +53,7 @@ class UploadManager(object):
                     source = (filename, urlopen_with_retry(source))
             if is_fileobj:
                 if unpack and self.zip_util.path_is_archive(filename):
-                    self._unpack_fileobj(
-                        source[0], source[1], bundle_path, simplify_archive=simplify_archives,
-                    )
+                    self._unpack_fileobj(source[0], source[1], bundle_path)
                 else:
                     with open(bundle_path, 'wb') as out:
                         shutil.copyfileobj(cast(IO, source[1]), out)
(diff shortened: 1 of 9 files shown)


codalab/lib/upload_manager.py
```diff
@@ -24,13 +24,7 @@ def __init__(self, bundle_model, bundle_store):
...
 
     def upload_to_bundle_store(
-        self,
-        bundle: Bundle,
-        source: Source,
-        git: bool,
-        unpack: bool,
-        simplify_archives: bool,
-        use_azure_blob_beta: bool,
+        self, bundle: Bundle, source: Source, git: bool, unpack: bool, use_azure_blob_beta: bool,
```
percyliang, line 27: Add `bundle` to the docstring

tests/cli/test_cli.py
```diff
@@ -816,11 +816,11 @@ def test_upload3(ctx):
 
     # Upload URL that's an archive
     uuid = _run_command([cl, 'upload', 'http://alpha.gnu.org/gnu/bc/bc-1.06.95.tar.bz2'])
-    check_contains(['README', 'INSTALL', 'FAQ'], _run_command([cl, 'cat', uuid]))
+    check_contains(['bc-1.06.95'], _run_command([cl, 'cat', uuid]))
```
percyliang, line 819: Why this change?

percyliang:
This is a pretty big backward incompatible change and I think breaks a lot of things...  I definitely like how things are simpler, but currently, the contract is that the contents of a bundle are contained in a single directory in the tar.gz.  This is because when you download a bundle as a tar.gz and unpack it, you want it to go into one directory rather than polluting your current directory.

But this makes me realize that we could have a the tar.gz contain either a single contents file or directory as we discussed last Friday...

tests/cli/test_cli.py
epicfaace: See the description of the case with http://alpha.gnu.org/gnu/bc/bc-1.06.95.tar.bz2 in this PR description. It contains a single directory, bc-1.06.95.

epicfaace:
@percyliang 
> but currently, the contract is that the contents of a bundle are contained in a single directory in the tar.gz

I don't believe that's correct. See this bundle (https://worksheets.codalab.org/bundles/0x49d745d0d3864666b38892402b5fe526), for example -- the contents of a bundle are not contained in a single directory in the tar.gz, but rather are just direct descendants of it:

image

> This is a pretty big backward incompatible change and I think breaks a lot of things

I don't think so. This PR, I believe, only affects the workflows where you run `cl upload **.tar.gz` (or any archive file) where the archive to be uploaded contains either exactly one file or exactly one folder at the root level. You can see that by the fact that I only had to make a couple of changes to test_cli.py to get the tests to pass.

epicfaace:
@percyliang and I discussed this and we agreed on proceeding with these changes.

@teetone @adiprerepa could you take a look and review?

codalab/lib/zip_util.py
```diff
@@ -112,8 +112,7 @@ def pack_files_for_upload(
         'fileobj': <file object of archive>,
         'filename': <name of archive file>,
         'filesize': <size of archive in bytes, or None if unknown>,
-        'should_unpack': <True iff archive should be unpacked at server>,
-        'should_simplify': <True iff directory should be 'simplified' at server>
+        'should_unpack': <True iff archive should be unpacked at server>
```
adiprerepa, line 115:
```suggestion
        'should_unpack': <True if the archive should be unpacked at server>
```

adiprerepa (approved):
lgtm with a nit

codalab/lib/zip_util.py
epicfaace: I think this was intentional ("if and only if") and is used in other parts of the codebase, but we could probably think about style maybe in a separate issue.

adiprerepa: oh I wasn' t aware that `iff` was `if and only if`. It's fine then.

Issue #3416, opened 2021-04-18
Sentry is not working with dev / prod environments


epicfaace:
- Sentry is not working with dev / prod environments. It only works with the stanford environment.

![image](https://user-images.githubusercontent.com/1689183/115161061-821ed100-a069-11eb-8bc1-d69c70c00a4f.png)

adiprerepa:
Looks like the sentry environment (stanford, dev, prod...) is set https://github.com/codalab/deployment/blob/master/ansible/roles/server/vars/main.yml#L12

But when I run `env | grep SENTRY` on the docker container for our current dev instance, there is no environment variable set. I assume this means the system thinks `mode` is blank, but I think we set it here: https://github.com/codalab/deployment/blob/master/ansible/roles/dev/vars/main.yml#L2

When we deploy, we also include the `-e mode=dev` flag, maybe that's overwriting it? It shouldn't really matter either way.

Thoughts @epicfaace?

epicfaace:
I'm not sure. Maybe you can try creating a new variable called sentry_mode and see if that fixes it?

adiprerepa:
yeah maybe that will work. 

It looks like some of the other values in `dev/vars/main.yaml` are also not being populated into the template. This may be a deeper issue...

epicfaace:
See this: https://github.com/codalab/codalab-worksheets/blob/master/docker_config/compose_files/docker-compose.yml#L11

PR #3444, opened 2021-04-26
Fix sentry setup by including sentry environment variables in docker-compose.yml 

adiprerepa:
This is the reason sentry isn't working for `dev` and `prod`.

Related: https://github.com/codalab/codalab-worksheets/issues/3416

After this is merged, https://github.com/codalab/deployment/blob/master/ansible/roles/server/vars/main.yml#L11-L12 env vars should work.


adiprerepa:
once the docker images build, I'll deploy this branch to dev

epicfaace (approved):
Looks good. Have you been able to test that this PR actually fixes it?

adiprerepa:
was waiting for GHA to finish - will deploy to dev and test

epicfaace (commented):
Ok, let me know once you successfully test, and then feel free to merge.

adiprerepa:
So this is odd..

`CODALAB_SENTRY_INGEST_URL` is set to the right URL but `CODALAB_SENTRY_ENVIRONMENT` is just blank, even when hard-coded, no different from `public_hostname` in `dev/main.yml`. no idea why.

codalab/worker_manager/azure_batch_worker_manager.py
```diff
@@ -133,6 +137,9 @@ def start_worker_job(self) -> None:
             task_container_run_options.append(
                 '--env CODALAB_SENTRY_INGEST_URL=%s' % CODALAB_SENTRY_INGEST
             )
+            task_container_run_options.append(
+                '--env CODALAB_SENTRY_ENVIRONMENT=%s' % CODALAB_SENTRY_ENVIRONMENT
```
adiprerepa, line 141: @epicfaace do you think we need this here? Is the sentry sdk initialized in the worker managers?

epicfaace: I'm not sure, but we should probably make an issue if that's not the case (I believe that nevertheless, these env variables will probably be passed on to the workers, which **do** have the sentry sdk initialized).

epicfaace (approved):

Issue #1255, opened 2019-06-24
suppress PreconditionViolation if worker unreachable


percyliang:
`

```JSON body:
<none>

Local variables:
message='Unable to reach worker'
condition=False

Traceback (most recent call last):
  File "/opt/codalab-worksheets/codalab/server/rest_server.py", line 148, in wrapper
    return callback(*args, **kwargs)
  File "/opt/codalab-worksheets/venv/local/lib/python2.7/site-packages/bottle.py", line 1732, in wrapper
    rv = callback(*a, **ka)
  File "/opt/codalab-worksheets/codalab/server/json_api_plugin.py", line 18, in wrapper
    result = callback(*args, **kwargs)
  File "/opt/codalab-worksheets/codalab/rest/bundles.py", line 570, in _fetch_bundle_contents_blob
    uuid, path, head_lines, tail_lines, max_line_length, truncation_text, gzipped_stream
  File "/opt/codalab-worksheets/codalab/lib/download_manager.py", line 21, in wrapper
    return f(*args, **kwargs)
  File "/opt/codalab-worksheets/codalab/lib/download_manager.py", line 209, in summarize_file
    string = self._get_read_response_string(response_socket_id)
  File "/opt/codalab-worksheets/codalab/lib/download_manager.py", line 290, in _get_read_response_string
    with closing(self._get_read_response_stream(response_socket_id)) as fileobj:
  File "/opt/codalab-worksheets/codalab/lib/download_manager.py", line 279, in _get_read_response_stream
    precondition(header_message is not None, 'Unable to reach worker')
  File "/opt/codalab-worksheets/codalab/common.py", line 95, in precondition
    raise PreconditionViolation(message)
PreconditionViolation: Unable to reach worker```

This happens, especially with own workers...should tolerate it.

percyliang:
Fixed by #1627

epicfaace:
Understand why the previous fix didn't work, and decide whether it's a deeper issue or should just be suppressed.

epicfaace:
Server code that raises this error: https://github.com/codalab/codalab-worksheets/blob/c2d8960bbdcda127e0ea8800b6ecd4b7849c48f7/codalab/lib/download_manager.py#L36 -- try printing out header_message

error_code instances in the worker (where the actual JSON messages are produced): https://github.com/codalab/codalab-worksheets/search?q=error_code

Example of preconditionviolation on prod (April 5, 8:14 PM ET):
[...]

PR #3445, opened 2021-04-26
suppress preconditionviolation when worker unreachable

adiprerepa:
related: https://github.com/codalab/codalab-worksheets/issues/1255

read https://github.com/codalab/codalab-worksheets/issues/1255#issuecomment-826405871

### DIFF
--- a/codalab/lib/download_manager.py
+++ b/codalab/lib/download_manager.py
@@ -361,7 +361,8 @@ class DownloadManager(object):
     def _get_read_response_stream(self, response_socket_id):
         with closing(self._worker_model.start_listening(response_socket_id)) as sock:
             header_message = self._worker_model.get_json_message(sock, 60)
-            precondition(header_message is not None, 'Unable to reach worker')
+            if header_message is None:
+                logging.info('Unable to reach worker')
             if 'error_code' in header_message:
                 raise http_error_to_exception(
                     header_message['error_code'], header_message['error_message']
@@ -378,6 +379,7 @@ class DownloadManager(object):
 
 class Deallocating(object):
     """
+    Deallocates the socket when closed.
     Deallocates the socket when closed.
     """
 


codalab/lib/download_manager.py
```diff
@@ -361,7 +361,8 @@ def _send_netcat_message(self, worker, response_socket_id, uuid, port, message):
     def _get_read_response_stream(self, response_socket_id):
         with closing(self._worker_model.start_listening(response_socket_id)) as sock:
             header_message = self._worker_model.get_json_message(sock, 60)
-            precondition(header_message is not None, 'Unable to reach worker')
+            if header_message is None:
+                logging.info('Unable to reach worker')
```
percyliang, line 365: I'm mildly worried that there's an important case that we're suppressing.  Could we have these go to sentry?

adiprerepa: If we want the error to go to sentry, I think we should leave it as is. Raising `PreconditionViolation` will create a Sentry event

epicfaace:
@adiprerepa sorry, I'm not clear why we need to suppress PreconditionViolation and how that follows from your investigation in #1255. Could you explain?

adiprerepa:
afaik, when you kill the worker, the `socket_id` in the worker table becomes invalid, and when the bundle manager tries to connect, the preconditionviolation is thrown. I'm not really sure if this is expected. wdyt @epicfaace ?

epicfaace:
What does "invalid" mean? Maybe this is expected because the socket is no longer open (since the worker has been killed).

adiprerepa:
right, but should we be throwing an exception for that? or just logging

adiprerepa:
> What does "invalid" mean? Maybe this is expected because the socket is no longer open (since the worker has been killed).

'invalid' means the socket no longer exists because the worker is gone

epicfaace:
> right, but should we be throwing an exception for that? or just logging

@adiprerepa I'm fine with whatever you choose, can you just pick a choice and explain your reasoning for it? The key to this issue (and https://github.com/codalab/codalab-worksheets/issues/1255) is just to pick one of these options and have a solid justification for it that's based on the investigation you already did in https://github.com/codalab/codalab-worksheets/issues/1255.

adiprerepa:
> > right, but should we be throwing an exception for that? or just logging
> 
> @adiprerepa I'm fine with whatever you choose, can you just pick a choice and explain your reasoning for it? The key to this issue (and #1255) is just to pick one of these options and have a solid justification for it that's based on the investigation you already did in #1255.

I think since dead workers are a "fact of life", we shouldn't treat this occurrence any differently than we would if a worker were offline and we tried to send it a message (https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/download_manager.py#L346). We should just log and move on.

epicfaace:
Makes sense, thanks! Also, why did https://github.com/codalab/codalab-worksheets/pull/1627 not fix this? Did that maybe only silence a subset of the preconditionviolation errors?

adiprerepa:
yeah, it didn't suppress everything

adiprerepa:
this should throw an error in the API I think

epicfaace:
Sounds good. What does the error look like on the API end? (Status code, response body?) Just want to make sure it's what we want.

adiprerepa:
reach out to user to see if `PreconditionViolation` is impacting their experience

adiprerepa:
So this isn't good
image
Functionality seems to be affected when a `PreconditionViolation` happens, unless this is a one-off

adiprerepa:
yeah, I can't reach this bundle, so can't reproduce.

PR #3448, opened 2021-04-26
Fix the bug of cannot go back to profile page after visiting dashboard

### LINKED ISSUE #3427: new profile page -> dashboard -> can't go back
To reproduce, go to the new profile page.  Click 'View dashboard' to go to the worksheet with the old dashboard.  When hit the back button in the browser, doesn't go back to the profile page, but is just stuck.

yuqijin:

The bug is due to previously I used `/worksheets?name=dashboard` to visit the dashboard, which actually will first make a request to interpret the URL into the /worksheets/<uuid> format. 
Also, update the default page for VerifySuccess & DeleteWorksheet.

### Related issues

fixes #3427 

### Screenshots

ezgif com-gif-maker (4)

### Checklist

### DIFF
--- a/frontend/src/components/Dashboard/SideBar.js
+++ b/frontend/src/components/Dashboard/SideBar.js
@@ -171,7 +171,7 @@ class SideBar extends React.Component {
             contentType: 'application/json; charset=utf-8',
             success: (data) => {
                 const uuid = data.response[0].uuid;
-                window.location = '/worksheets/' + uuid;
+                window.location.href = '/worksheets/' + uuid;
             },
             error: (xhr, status, err) => {
                 console.error(xhr.responseText);
--- a/frontend/src/components/Dashboard/SideBar.js
+++ b/frontend/src/components/Dashboard/SideBar.js
@@ -160,6 +160,24 @@ class SideBar extends React.Component {
         fetchBundles(0, {});
     }
 
+    viewDashboard() {
+        const url = '/rest/interpret/wsearch';
+        $.ajax({
+            url: url,
+            dataType: 'json',
+            type: 'POST',
+            cache: false,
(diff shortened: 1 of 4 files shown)


epicfaace (commented):
Your approach in this PR requires a bit more complicated code whenever you want to create a button that goes to the dashboard. The root problem is that whenever you go to the /worksheets?name=dashboard page, you can't go back anymore.

Instead of this approach, maybe we could fix the root cause of this problem. The root cause is that in the component rendered at /worksheets?name=dashboard, it uses `history.push` instead of `history.replace`.

Can you try this approach: using history.replace here?: https://github.com/codalab/codalab-worksheets/blob/ae726fde1ca12aa227ccbc179601334e43c8ea3d/frontend/src/components/worksheets/WorksheetNameSearch.js#L24

If that fixes the issue, I think we should go with that approach instead of the approach in this PR.

yuqijin:
> Your approach in this PR requires a bit more complicated code whenever you want to create a button that goes to the dashboard. The root problem is that whenever you go to the /worksheets?name=dashboard page, you can't go back anymore.
> 
> Instead of this approach, maybe we could fix the root cause of this problem. The root cause is that in the component rendered at /worksheets?name=dashboard, it uses `history.push` instead of `history.replace`.
> 
> Can you try this approach: using history.replace here?:
> 
> https://github.com/codalab/codalab-worksheets/blob/ae726fde1ca12aa227ccbc179601334e43c8ea3d/frontend/src/components/worksheets/WorksheetNameSearch.js#L24
> 
> If that fixes the issue, I think we should go with that approach instead of the approach in this PR.

thanks for the suggestion! I have updated the PR.

epicfaace (approved):

Issue #3437, opened 2021-04-26
Ensure files are streamed from the bundle CLI when zipping


epicfaace:
Right now, files are not actually streamed from the bundle CLI while they are being zipped. Rather, the files must be zipped first (and stored in a SpooledTemporaryFile) because they can actually be sent over to the server.

See: https://github.com/codalab/codalab-worksheets/blob/75dd2d6d34dabc958d166338a5c7d7fc822132ee/codalab/lib/zip_util.py#L172-L177

PR #3449, opened 2021-04-26
Don't unpack from the bundle CLI when multiple archives are uploaded

epicfaace:

For https://github.com/codalab/codalab-worksheets/issues/3439. Don't unpack from the bundle CLI when multiple archives are uploaded.

**This is a breaking change.** This means that running, for example, `cl upload item.gz dir.tar.gz` will not unpack the each individual archive file; rather, these files will just be added as usual to the entire bundle directory.

I've also added unit tests that exercise all of the functionality of `zip_util.pack_file_for_upload`, as we'll be changing this function later for streaming (https://github.com/codalab/codalab-worksheets/issues/3437).

### DIFF
--- a/tests/unit/lib/zip_util_test.py
+++ b/tests/unit/lib/zip_util_test.py
@@ -1,12 +1,36 @@
+import tempfile
+import zipfile
+import tarfile
+import os
+import gzip
+import bz2
 import unittest
+
 from codalab.common import UsageError
-from codalab.lib.zip_util import get_archive_ext, strip_archive_ext, path_is_archive
+from codalab.lib.zip_util import (
+    get_archive_ext,
+    strip_archive_ext,
+    path_is_archive,
+    pack_files_for_upload,
+)
+from codalab.worker.file_util import tar_gzip_directory, zip_directory
+from io import BytesIO
+
+SAMPLE_CONTENTS = b"hello world"
+
+
+def tar_bz2_directory(*args, **kwargs):
+    """Method used for creating a .tar.bz2 archive from a directory.
+    This is just used for tests; it's not performance optimized
+    and should not be used in production code."""
+    output = tar_gzip_directory(*args, **kwargs)
+    return BytesIO(bz2.compress(gzip.decompress(output.read())))
 
 
 class ZipUtilTest(unittest.TestCase):
     """Test for zip util methods."""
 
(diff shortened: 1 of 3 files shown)


adiprerepa (commented):
why are we doing this?

epicfaace:
See https://docs.google.com/document/d/1l4kOqi9irBjOApmn4E6vlzsjAXDJbetIyVw8gMRHrpU/edit# for the design doc. Basically, this is not a feature that is used very often; it's not really a common use case for users of the CodaLab CLI.

It makes it harder to implement Blob Storage and/or streaming uploads, because in order to create streaming uploads with multiple files (i.e. `cl upload a.txt b.txt c.txt`), we have to construct an archive on-the-fly with `a.txt`, `b.txt`, and `c.txt` in it.

On the other hand, if we kept the original behavior of also unpacking each archive source, a command such as `cl upload a.tar.gz b.tar.gz c.tar.gz` would require us to, while we write to the output .tar.gz file in a stream, 1) unpack `a.tar.gz`, then for every item in `a.tar.gz`, prepend `a/` to its path, and add it to the final output archive in the stream; 2) do the same with `b.tar.gz`, but prepend `b/` to these paths; and 3) do the same with `c.tar.gz`. A lot of complexity, and it's possible, but it's probably not worth the effort.

tests/unit/lib/zip_util_test.py
```diff
@@ -1,37 +1,211 @@
...
         self.assertEqual(get_archive_ext(path), ".tar.gz")
         self.assertEqual(strip_archive_ext(path), "https://codalab.org/file")
+
+    def test_pack_single_file(self):
+        """Pack a single file."""
+        with tempfile.NamedTemporaryFile() as f:
+            f.write(SAMPLE_CONTENTS)
+            f.flush()
+            packed = pack_files_for_upload(
+                sources=[f.name], should_unpack=False, follow_symlinks=False
```
jzwang43, line 69: Do we still need `should_unpack` since we are not doing unpack anymore?

epicfaace: We still unpack single archives (such as `cl upload a.zip`). We're just not unpacking archives if multiple ones are specified.

jzwang43 (approved):
Looks good.

Issue #3453, opened 2021-04-30
cl upload fails when 'disk_quota' is not present in user


adiprerepa:
Figure out why `disk_quota` isn't present. Worst case, just add a `hasattr` check and enforce disk quote based on that (downside isn't that high)

![image](https://user-images.githubusercontent.com/43913660/116633094-75259b80-a90d-11eb-8bae-6e68c694939c.png)

adiprerepa:
We seem to check if the field exists when API calls are made because it is not guaranteed in the response:
- https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L4067
- https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/bundle_cli.py#L3965

I think the appropriate response would be to do a key check.

PR #3454, opened 2021-04-30
Fix disk quota enforcement error for non-root users by using correct endpoint

adiprerepa:
For https://github.com/codalab/codalab-worksheets/issues/3453

This seems to be the correct remedy because `disk_quota` and analogous fields are not guaranteed to be in the api response.

cc @teetone @epicfaace


epicfaace:
Wait, but what does it mean if disk_quota does not exist? Isn't every user supposed to have a disk quota? How would we check to ensure the bundle does not exceed the user's disk quota in these cases?

Maybe you can see how disk quota enforcement is done on the REST server currently for users, after the bundle is finished uploading.

adiprerepa:
I believe quota enforcement is done here: https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L886

Would it not be better to follow the rest of the behavior in `bundle_cli.py` and just check for the field's existence?

And any idea how I would debug that user's specific instance?

epicfaace:
> Would it not be better to follow the rest of the behavior in `bundle_cli.py` and just check for the field's existence?

Well, but my question is: what does it mean if the field doesn't exist? Every user should have a disk quota set (unlike maybe other fields in bundle_cli.py), and it's always accessible as you linked to at https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L886. Perhaps this means that we should use another method to find the user's disk quota, or there might be a bug in the implementation of `client.fetch('users', client.fetch('user')['user_name'])` that doesn't properly get the user's current disk quota.

> And any idea how I would debug that user's specific instance?

Not sure, maybe check the rest server logs and search for that user's user id

adiprerepa:
@teetone what was that user's id?

adiprerepa:
yeah, nothing in `codalab_rest-server_1`'s logs

user id: `pangwei`

adiprerepa:
having trouble reproducing...

running `uinfo` after switching to workshet `0xacd40b3e4991410b98643b9cc0b10347` gives me a valid disk quota.

adiprerepa:
look into the endpoint (just returned `AuthenticatedUser`)

adiprerepa:
the endpoint seems to never not return `disk_quota` as far as I can tell...

@epicfaace I think I'll fix the user fetching logic in disk quota enforcement, but should we leave the attribute check in? Seems like a dead end to investigate

epicfaace (changes requested):
I looked at the code a bit further and figured out why the issue was happening to Pang Wei.

Here is the code for the "users" endpoint:

https://github.com/codalab/codalab-worksheets/blob/32cd704ef92290fe5cf6af77f86de144c3f08c77/codalab/rest/users.py#L79-L94

Note how if the user is not a root user, then they essentially get only the fields in UserSchema, which does not include disk_quota.

Can you please remove the disk_quota and disk_used checks, then, as we don't need them anymore? Additionally, please update test_cli.py for the disk quota enforcement to also test with a non-root user signed in (so this issue doesn't regress).

adiprerepa:
> Note how if the user is not a root user, then they essentially get only the fields in UserSchema, which does not include disk_quota.

@epicfaace isn't the endpoint `/users` being called? not `/users/<user_spec>`? So now  there is no root user check.

I'll add the test

epicfaace:
> isn't the endpoint `/users` being called? not `/users/<user_spec>`? So now there is no root user check.

I don't think so, I'm pretty sure it's calling `/users/<user_spec>`. The code that calls it is:

```
client.fetch('users', client.fetch('user')['user_name'])
```

Where else would the value of `client.fetch('user')['user_name']` be passed on to the REST API?

Feel free to trace the code / test it out to double check, though.

adiprerepa:
Oh yeah, that used to be the case. I think we decided that was redundant, so now the code is just:

```python3
user = client.fetch('user')
```

codalab/lib/bundle_cli.py
```diff
@@ -1368,13 +1368,15 @@ def do_upload_command(self, args):
...
             # Calculate size of sources
             total_bundle_size = sum([get_path_size(source) for source in sources])
-            user = client.fetch('users', client.fetch('user')['user_name'])
-            disk_left = user['disk_quota'] - user['disk_used']
-            if disk_left - total_bundle_size <= 0:
-                raise DiskQuotaExceededError(
-                    'Attempted to upload bundle of size %d with only %d remaining in user\'s disk quota.'
-                    % (total_bundle_size, disk_left)
-                )
+            user = client.fetch('user')
```
adiprerepa, line 1371: @epicfaace here was where we changed the logic

tests/cli/test_cli.py
```diff
@@ -101,14 +101,22 @@ def current_user():
...
 
 
-def create_user(context, username, password='codalab'):
+def create_user(context, username, password='codalab', disk_quota='100g'):
     # Currently there isn't a method for creating a user with the CLI. Use CodaLabManager instead.
     manager = CodaLabManager()
     model = manager.model()
 
     # Creates a user without going through the full sign-up process
     model.add_user(
```
epicfaace, line 113: You could simplify this PR (and avoid the need to add new arguments to `add_user` that are only used in tests) by just adding:

```
model.default_user_info['disk_quota'] = disk_quota
```

adiprerepa: yeah that's a lot simpler

codalab/lib/bundle_cli.py
```diff
@@ -1368,7 +1368,8 @@ def do_upload_command(self, args):
             sources = [path_util.normalize(path) for path in args.path]
             # Calculate size of sources
             total_bundle_size = sum([get_path_size(source) for source in sources])
-            user = client.fetch('users', client.fetch('user')['user_name'])
+            user = client.fetch('user')
+            # Only attempt to enforce disk quota when disk information is returned.
```
epicfaace: remove this comment, it's no longer relevant

epicfaace (approved):

PR #3463, opened 2021-05-03
Add CAPTCHA for Registration

### LINKED ISSUE #3447: Add a CAPTCHA for user registration
We should look into adding a CAPTCHA for user registration. This will prevent too many automated registrations.

jzwang43:

We should look into adding a CAPTCHA for user registration. This will prevent too many automated registrations.

### Related issues

fixes #3447 

### Screenshots

ezgif-6-806d14c029ba

### Checklist


codalab_service.py
```diff
@@ -366,6 +366,13 @@ def has_callable_default(self):
     ),
     # Public workers
     CodalabArg(name='public_workers', help='Comma-separated list of worker ids to monitor'),
+    CodalabArg(
+        name='recaptcha_site_key',
+        env_var='REACT_APP_RECAPTCHA_SITE_KEY',
```
epicfaace: Let's rename this REACT_APP_CODALAB_RECAPTCHA_SITE_KEY instead. Also, let's not add a custom env_var here -- instead, in the mapping in docker-compose.yml, we can do REACT_APP_CODALAB_RECAPTCHA_SITE_KEY=${CODALAB_RECAPTCHA_SITE_KEY}

frontend/src/components/SignUp.js
```diff
@@ -179,7 +182,20 @@ export class SignUp extends React.Component {
                             name='error_uri'
                             value={this.props.location.pathname}
                         />
-                        <button type='submit'>Sign Up &raquo;</button>
+                        <div style={{ display: this.state.captchaPassed && 'none' }}>
```
epicfaace: I think we also have to configure a reCAPTCHA secret key and server-side validation to do the reCAPTCHA validation properly -- see https://developers.google.com/recaptcha/docs/verify

codalab_service.py
percyliang: Can we make all environment variables start with `$CODALAB_` and not have to have two?

jzwang43: https://create-react-app.dev/docs/adding-custom-environment-variables/
Looks like for create-react-app, it must start with `REACT_APP_`

frontend/src/components/SignUp.js
jzwang43: Yes, I have added the logic in the latest commit. Thanks for pointing it out.

codalab_service.py
jzwang43: > 
> 
> Let's rename this REACT_APP_CODALAB_RECAPTCHA_SITE_KEY instead. Also, let's not add a custom env_var here -- instead, in the mapping in docker-compose.yml, we can do REACT_APP_CODALAB_RECAPTCHA_SITE_KEY=${CODALAB_RECAPTCHA_SITE_KEY}

I wanted to have default keys so the pages can run out of the box, are there any other ways to do it? Otherwise developers cannot use the signup page by default.

epicfaace: Default keys are fine. My suggestion was that we could still make all environment variables to codalab_service.py start with `$CODALAB_` (as @percyliang suggested) by adding the line `REACT_APP_CODALAB_RECAPTCHA_SITE_KEY=${CODALAB_RECAPTCHA_SITE_KEY}` in docker-compose.yml. 

This way, we're essentially hiding the implementation logic (the fact that environment variables in the frontend must start with `REACT_APP_`) and presenting a consistent interface to the user who is actually running codalab_service.py (the interface being that "all environment variables to codalab_service.py must start with `$CODALAB_`").

It shouldn't be too confusing to have two, because all environment variables that are exclusively passed on to the frontend container would have to have this format (`REACT_APP_CODALAB_ENV_VAR_NAME=${CODALAB_ENV_VAR_NAME}`).

codalab_service.py
```diff
@@ -366,6 +366,20 @@ def has_callable_default(self):
...
+    CodalabArg(
+        name='recaptcha_site_key',
+        env_var='REACT_APP_CODALAB_RECAPTCHA_SITE_KEY',
+        type=str,
+        default='6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI',  # test key provided by Google, will always pass
+        help='API site key needed for recaptcha',
+    ),
+    CodalabArg(
+        name='recaptcha_secret_key',
+        env_var='CODALAB_RECAPTCHA_SECRET_KEY',
```
epicfaace: We don't need to specify env_var here because it will be set to `CODALAB_RECAPTCHA_SECRET_KEY` by default

codalab_service.py
```diff
@@ -366,6 +366,20 @@ def has_callable_default(self):
     ),
     # Public workers
     CodalabArg(name='public_workers', help='Comma-separated list of worker ids to monitor'),
+    CodalabArg(
+        name='recaptcha_site_key',
+        env_var='REACT_APP_CODALAB_RECAPTCHA_SITE_KEY',
+        type=str,
+        default='6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI',  # test key provided by Google, will always pass
+        help='API site key needed for recaptcha',
```
```suggestion
        help='API site key needed for reCAPTCHA',
```

codalab_service.py
```diff
@@ -366,6 +366,20 @@ def has_callable_default(self):
...
+    ),
+    CodalabArg(
+        name='recaptcha_secret_key',
+        env_var='CODALAB_RECAPTCHA_SECRET_KEY',
+        type=str,
+        default='6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe',  # test key provided by Google, will always pass
+        help='API secret key needed for recaptcha',
```
```suggestion
        help='API secret key needed for reCAPTCHA',
```

codalab/rest/account.py
```diff
@@ -67,23 +68,38 @@ def do_signup():
...
+        errors.append(str(e))
+
     if request.user.is_authenticated:
         errors.append(
             "You are already logged in as %s, please log out before "
             "creating a new account." % request.user.user_name
         )
 
-    if request.forms.get('confirm_password') != password:
+    if request.json.get('confirm_password') != password:
```
adiprerepa: why the change here?

codalab_service.py
jzwang43: @epicfaace  I see what you mean. I have updated the code.

codalab_service.py
jzwang43: Suggestions applied in the new commits.

codalab/rest/account.py
jzwang43: Yeah I was sending json to the backend, I switched back to sending form to the backend so the change here is not needed anymore.

codalab/rest/account.py
```diff
@@ -75,8 +76,23 @@ def do_signup():
     last_name = request.forms.get('last_name')
     password = request.forms.get('password')
     affiliation = request.forms.get('affiliation')
+    token = request.forms.get('token')
 
     errors = []
+
+    url = 'https://www.google.com/recaptcha/api/siteverify?secret={}&response={}'.format(
+        os.environ['CODALAB_RECAPTCHA_SECRET_KEY'], token
```
epicfaace: Will this throw an error if `CODALAB_RECAPTCHA_SECRET_KEY` is not defined? Is that variable always defined (so we don't have to worry about this edge case)?

codalab/rest/account.py
```diff
@@ -75,8 +76,23 @@ def do_signup():
     last_name = request.forms.get('last_name')
     password = request.forms.get('password')
     affiliation = request.forms.get('affiliation')
+    token = request.forms.get('token')
```
epicfaace, line 79: Let's add a test called account_test.py that tests the signup endpoint (let's structure it similarly to https://github.com/codalab/codalab-worksheets/blob/master/tests/unit/rest/user_test.py). We can mock the response from the reCAPTCHA API -- this would just ensure we have this endpoint covered in our tests, which is good since we're changing it now.

frontend/src/components/SignUp.js
```diff
@@ -173,13 +190,20 @@ export class SignUp extends React.Component {
                                 onChange={this.handleInputChange}
                             />
                         </div>
-                        <input type='hidden' name='success_uri' value='/account/signup/success' />
-                        <input
-                            type='hidden'
-                            name='error_uri'
-                            value={this.props.location.pathname}
+                        <div style={{ display: this.state.captchaPassed && 'none' }}>
+                            Please complete the captcha below.
```
epicfaace: I think we probably don't need this text

epicfaace (changes requested):

codalab/rest/account.py
jzwang43: Since the variable will either be the default key or overridden by the developer, I don't since it will ever be undefined unless the env variable was manually unset, which is very unlikely.

frontend/src/components/SignUp.js
jzwang43: Sure, this has been removed since the reCAPTCHA should be self-explanatory.

codalab/rest/account.py
jzwang43: I have added the tests for the signup endpoint. We don't need mock since the default secret key will just accept any token (as long as something is given). I am just directly calling the backend API for the tests.

codalab/rest/account.py
```diff
@@ -80,18 +80,21 @@ def do_signup():
...
-        os.environ['CODALAB_RECAPTCHA_SECRET_KEY'], token
-    )
-    res = requests.post(url)
+    if not token:
+        errors.append('Google reCAPTCHA token is missing.')
+    else:
+        url = 'https://www.google.com/recaptcha/api/siteverify?secret={}&response={}'.format(
+            os.environ['CODALAB_RECAPTCHA_SECRET_KEY'], token
+        )
+        res = requests.post(url)
```
epicfaace: Instead of using format strings to add in the url parameters, let's pass them in as a dictionary to the params keyword in requests.post (so that they'll be automatically URL-encoded)

codalab/rest/account.py
epicfaace: Ideally, we would mock so that we don't have to hit the network for these tests. See if you can do so, but maybe if it takes too much work we can just keep it as it is.

epicfaace (approved):

codalab/rest/account.py
jzwang43: Updated.

codalab_service.py
```diff
@@ -716,6 +716,12 @@ def build_image(self, image):
         else:
             build_args = ''
 
+        if image == 'frontend' and not self.args.dev:
+            CODALAB_RECAPTCHA_SITE_KEY = os.environ['CODALAB_RECAPTCHA_SITE_KEY']
+            build_args += '--build-arg REACT_APP_CODALAB_RECAPTCHA_SITE_KEY={}'.format(
```
epicfaace: Let's do this to simplify instead:
```suggestion
            build_args += '--build-arg REACT_APP_CODALAB_RECAPTCHA_SITE_KEY=$CODALAB_RECAPTCHA_SITE_KEY'
```

docker_config/dockerfiles/Dockerfile.frontend
```diff
@@ -3,6 +3,7 @@ FROM node:14-slim
 ENV NPM_CONFIG_LOGLEVEL warn
 
 ARG dev=false
+ARG REACT_APP_CODALAB_RECAPTCHA_SITE_KEY
```
epicfaace: Will this arg be accessible as an environment variable?

codalab_service.py
```diff
@@ -716,6 +716,12 @@ def build_image(self, image):
         else:
             build_args = ''
 
+        if image == 'frontend' and not self.args.dev:
+            CODALAB_RECAPTCHA_SITE_KEY = os.environ['CODALAB_RECAPTCHA_SITE_KEY']
```
epicfaace: Add a comment mentioning that we need to do this to pass along the frontend variables during build time.

codalab_service.py
epicfaace: Actually, even better, I wonder if we could add a parameter to `CodalabArg` called `frontend` -- when this parameter is set, this step will automatically add in all frontend-related variables to build_args. This way, we don't have to duplicate all these variable names.

epicfaace (changes requested):
Interesting complication you ran into regarding passing env variables to the frontend container at build-time. Is there an easy way to ensure this is tested in our CI pipeline?

docker_config/dockerfiles/Dockerfile.frontend
jzwang43: Yes, I have tested both in dev and production this is accessible.

codalab_service.py
jzwang43: Can we just always pass all the frontend-related variables to build_args when building? These variables should always be needed anyways. We can have a list of these variables and loop through those too add them to build_args.

epicfaace: Yeah, I was suggesting a similar approach, but rather than maintaining a separate list of frontend-related variables, we can just add a kwarg `frontend=True` that is passed to `CodalabArg` that would let us mark specific arguments as frontend-related. What do you think?

codalab_service.py
jzwang43: Comment added.

codalab_service.py
jzwang43: I added kwarg and also I switched to use .env.production instead of ARG in dockerfile. This allow us to programmatically dump all the env variables instead of writing them in docker file. Values in `.env.production` will autormatically be picked up when doing `npm build`.

frontend/.env.production
```diff
@@ -0,0 +1 @@
+REACT_APP_CODALAB_RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
```
epicfaace: We should .gitignore this file

codalab_service.py
```diff
@@ -716,11 +718,13 @@ def build_image(self, image):
         else:
             build_args = ''
 
+        # We need the frontend env variables during build time
         if image == 'frontend' and not self.args.dev:
-            CODALAB_RECAPTCHA_SITE_KEY = os.environ.get('CODALAB_RECAPTCHA_SITE_KEY')
-            build_args += '--build-arg REACT_APP_CODALAB_RECAPTCHA_SITE_KEY={}'.format(
-                CODALAB_RECAPTCHA_SITE_KEY
-            )
+            with open('frontend/.env.production', "w") as f:
```
epicfaace: Need w+ so it will overwrite an existing env file. Also, let's just call it `.env` for simplicity.

```suggestion
            with open('frontend/.env', "w+") as f:
```

codalab_service.py
```diff
@@ -133,13 +133,14 @@ def var_path(name):
 
 # An configuration argument.
 class CodalabArg(object):
-    def __init__(self, name, help, type=str, env_var=None, flag=None, default=None):
+    def __init__(self, name, help, type=str, env_var=None, flag=None, default=None, frontend=False):
```
epicfaace, line 142: Let's add a docstring here. Mainly, we just need somewhere to document what `frontend` means.

frontend/.env.production
jzwang43: Yes I though it was in there already but it only has .`env`.

codalab_service.py
jzwang43: Added. Let me know what you think.

codalab_service.py
jzwang43: Fixed. Thanks:)

codalab_service.py
epicfaace: Let's also talk about the fact that the frontend argument indicates that the env variable will be passed down to the frotnend container at build time, prefixed by REACT_APP_.

docker_config/compose_files/docker-compose.yml
```diff
@@ -67,6 +67,8 @@ x-codalab-env: &codalab-env
   - CODALAB_ALWAYS_USE_AZURE_BLOB_BETA=${CODALAB_ALWAYS_USE_AZURE_BLOB_BETA}
   - CODALAB_SENTRY_INGEST_URL=${CODALAB_SENTRY_INGEST_URL}
   - CODALAB_SENTRY_ENVIRONMENT=${CODALAB_SENTRY_ENVIRONMENT}
+  - CODALAB_RECAPTCHA_SECRET_KEY=${CODALAB_RECAPTCHA_SECRET_KEY}
+  - REACT_APP_CODALAB_RECAPTCHA_SITE_KEY=${CODALAB_RECAPTCHA_SITE_KEY}
```
epicfaace: I think because we already create the .env file, we won't need this line anymore.

epicfaace (approved):

docker_config/compose_files/docker-compose.yml
jzwang43: Without this line users cannot supply their own site_key in development mode i think? Since in dev mode the react app actually reads the docker's env vars.

codalab_service.py
jzwang43: Updated, thanks.

docker_config/compose_files/docker-compose.yml
epicfaace: I'm pretty sure that the env var will still be passed to the react app (due to the presence of .env file) in dev mode. Maybe you could try and see if dev mode still gets the env variable when taking this line out?

jzwang43: I think the edge case is when user only uses `-d` instead of `-bd`, then the .env is not outputted thus the key is the last one when they built the image. I think most people use `-bd` all the time so it is probably not an issue?

jzwang43: Yeah I think this is probably the expected behavior so I went ahead and applied the suggestion.

PR #3464, opened 2021-05-05
Fix Schema Deletion Bug

### LINKED ISSUE #3461: deleting schema on worksheet has some rough edges
Go to a worksheet and create a schema.  If it is selected and you hit backspace, it doesn't delete it and tries to delete bundles instead.

You have to open up the schema, and then click the trash can.  This works, but the message asks if you want to delete the markdown rather than deleting the schema.

yuqijin:

update schema deletion confirmation message;
enable to use the short key(`backspace`/`del`) to delete a schema

### Related issues

fixes #3461 
### Screenshots

ezgif com-gif-maker (4)

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -397,9 +397,13 @@ class Worksheet extends React.Component {
             this.setState({ openedDialog: null, deleteItemCallback: null });
             return;
         }
-        if (cmd_type === 'deleteItem') {
+        if (cmd_type === 'deleteMarkdown') {
             // This is used to delete markdown blocks
             this.setState({ openedDialog: DIALOG_TYPES.OPEN_DELETE_MARKDOWN });
+        }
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -397,10 +397,14 @@ class Worksheet extends React.Component {
             this.setState({ openedDialog: null, deleteItemCallback: null });
             return;
         }
-        if (cmd_type === 'deleteItem') {
```
epicfaace, line 400: What was deleteItem used for earlier? Do we not need a handler for that case anymore?

yuqijin: previously `deleteitem` refers to `deleteMarkdown` only. I think the naming is ambiguous since we now have `delete Schema` as well, so I rename it to be `deleteMarkdown` & `deleteSchema`

epicfaace (approved):

Issue #3417, opened 2021-04-18
Sentry frontend integration


epicfaace:
- Integrate sentry with the frontend so we can start cataloging frontend errors, too (see https://www.npmjs.com/package/@sentry/react)

PR #3467, opened 2021-05-06
setup sentry for frontend

adiprerepa:
for https://github.com/codalab/codalab-worksheets/issues/3417

using https://docs.sentry.io/platforms/javascript/guides/react/


adiprerepa:
okay, that commit had a lot of files changed. Do files get generated when I add a library or something?

epicfaace:
@adiprerepa I think you installed sentry in the wrong place. You have to run `npm install` after running a `cd frontend`. This way, sentry will be installed in `frontend/node_modules`, which is ignored by git.

adiprerepa:
@epicfaace should be ready now

jzwang43:
We still need this https://docs.sentry.io/platforms/javascript/guides/react/usage/ to capture errors? I think we need to do something like `Sentry.captureException(err);`  when errors are cought. This pull request is just adding the module?

epicfaace:
[The docs say,
](https://docs.sentry.io/platforms/javascript/guides/react/)
> Once this is done, all unhandled exceptions are automatically captured by Sentry.

Nevertheless, we probably do want to test this out to ensure the errors are actually being caught.

adiprerepa:
I tested it out, sentry does automatically capture errors. Here is what it looks like when I just added `throw new Error('foo');`:
image
@epicfaace @jzwang43 should we merge this?

epicfaace (commented):
We need to make sure that REACT_APP_CODALAB_SENTRY_INGEST_URL and REACT_APP_CODALAB_SENTRY_ENVIRONMENT are actually passed down to the frontend container through docker-compose.yml, though. We may also have to add the variables at build-time as @jzwang43 did in #3463 (https://github.com/codalab/codalab-worksheets/pull/3463/files#diff-d5ad1f3e91f51a1ae9de4052cad0da8231afafb60dd0ef4ce3298c97b101d28eR719). Let's consider this PR blocked until #3463 is merged and we have a better idea of how to pass variables down to the frontend.

adiprerepa:
@epicfaace #3643 merged, can we merge this?

epicfaace:
@adiprerepa Can we add the sentry frontend variable in a way similar to REACT_APP_CODALAB_RECAPTCHA_SITE_KEY  in https://github.com/codalab/codalab-worksheets/pull/3501? Then, once https://github.com/codalab/codalab-worksheets/pull/3501 is merged, we should be good to merge this in as well.

adiprerepa:
@epicfaace should be good to go

epicfaace (approved):

PR #3468, opened 2021-05-06
Fix shared worksheet bug due to code refactoring

yuqijin:

We should include this bug fix PR before we deploy the new version; otherwise, the shared worksheet tab on the new profile page will always be empty.


jzwang43 (approved):
Looks good, thanks for fixing this.

PR #3470, opened 2021-05-06
Blob Storage: allow downloading from single contents.gz files, improve gzip handling when downloading gzipped files

### LINKED ISSUE #3440: Update downloading logic to work with plain .gz files on Blob Storage
Right now, the downloading logic for Blob Storage only works with .tar.gz files. However, according to the [design doc](https://docs.google.com/document/d/1l4kOqi9irBjOApmn4E6vlzsjAXDJbetIyVw8gMRHrpU/edit#), we should use plain .gz files to store single-file bundles.

We should update the downloading logic to work with "contents.gz" bundles. They can still be read through random access by using the "index.sqlite" index files.

Here's the downloading flow table (taken from the doc) whose specifications should be followed:

Scenario | Command | Description
-- | -- | --
Read from a bundle that is a single file | cl download 0xacc8252ae6f646368a749b28bcbefd2d | The contents.gz file is streamed back to the client. If random access is needed (such as reading portions of a file), ratarmount will be used for random access.
Download a bundle that is a single directory | cl download 0x722fadbeb50d4af68174a325f2e86b26 | The entire .tar.gz file on Blob Storage is streamed to the client.
Download a file path from within a bundle | cl download 0x722fadbeb50d4af68174a325f2e86b26/file.txt | The member of the archive corresponding to the given path on Blob Storage is streamed back to the client (uses TarFileStream).
Download a subdirectory path from within a bundle | cl download 0x722fadbeb50d4af68174a325f2e86b26/dir | The members of the archive that begin to the given path on Blob Storage are streamed into a .tar.gz file for the subdirectory, which is then streamed back to the client (uses TarSubdirStream).

epicfaace:
- Updates the downloading logic to work with "contents.gz" bundles, which represent single files. They can still be read through random access by using the "index.sqlite" index files.
- Improve gzip handling when downloading gzipped files. I do this by adding a `gzipped` kwarg to `OpenFile`. This allows us to directly stream the `contents.gz` / `contents.tar.gz` files if one is requesting a gzipped version of the entire bundle, so that we don't have to un-gzip and then re-gzip these files.

fixes https://github.com/codalab/codalab-worksheets/issues/3440.

### DIFF
--- a/codalab/worker/file_util.py
+++ b/codalab/worker/file_util.py
@@ -8,7 +8,7 @@ import bz2
 
 from codalab.common import BINARY_PLACEHOLDER, UsageError
 from codalab.common import parse_linked_bundle_url
-from codalab.worker.un_gzip_stream import BytesBuffer
+from codalab.worker.un_gzip_stream import BytesBuffer, un_gzip_stream
 from codalab.worker.tar_subdir_stream import TarSubdirStream
 from codalab.worker.tar_file_stream import TarFileStream
 from apache_beam.io.filesystem import CompressionTypes
@@ -178,13 +178,13 @@ def unzip_directory(fileobj_or_name, directory_path, force=False):
         do_unzip(fileobj_or_name)
 
 
-class OpenIndexedTarGzFile(object):
-    """Open a .tar.gz file specified by the provided path on Azure Blob Storage.
+class OpenIndexedArchiveFile(object):
+    """Open an archive file (.tar.gz / .gz) specified by the provided path on Azure Blob Storage.
     Also reads this file's associated index.sqlite file, then opens the file as an
     SQLiteIndexedTar object.
 
-    This way, the .tar.gz file can be read and specific files can be extracted without
-    needing to download the entire .tar.gz file.
+    This way, the archive file can be read and specific files can be extracted without
+    needing to download the entire archive file.
 
     Returns the SQLiteIndexedTar object.
     """
@@ -196,7 +196,9 @@ class OpenIndexedTarGzFile(object):
             self.index_file_name = index_fileobj.name
             shutil.copyfileobj(
                 FileSystems.open(
-                    path.replace("/contents.tar.gz", "/index.sqlite"),
+                    path.replace("/contents.tar.gz", "/index.sqlite").replace(
+                        "/contents.gz", "/index.sqlite"
+                    ),
                     compression_type=CompressionTypes.UNCOMPRESSED,
                 ),
                 index_fileobj,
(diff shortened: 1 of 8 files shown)


codalab/worker/file_util.py
```diff
@@ -219,40 +221,69 @@ class OpenFile(object):
...
-                    return TarFileStream(tf, finfo)
-        # Stream a single file that is directly on Blob Storage (such as the entire .tar.gz file)
-        return FileSystems.open(
-            self.path, self.mode, compression_type=CompressionTypes.UNCOMPRESSED
-        )
+                    fs = TarFileStream(tf, finfo)
+                    return GzipStream(fs) if self.gzipped else fs
+        else:
+            # Stream a file from disk storage.
+            return open(self.path, self.mode)
```
epicfaace, line 282: I'm switching back to our original disk storage implementations in this `else` statement, because at this point we know that we're only accessing disk storage (so there's no need to use the extra complexity of adding Beam).

codalab/worker/file_util.py
```diff
@@ -351,24 +377,23 @@ def get_file_size(file_path):
...
             if finfo is None:
                 raise FileNotFoundError(fpath)
             return finfo.size
     if not get_path_exists(file_path):
         raise FileNotFoundError(file_path)
-    # TODO: add a FileSystems.size() method to Apache Beam to make this less verbose.
-    filesystem = FileSystems.get_filesystem(file_path)
-    return filesystem.size(file_path)
+    # Local path
+    return os.stat(file_path).st_size
```
epicfaace, line 392: I'm switching back to our original disk storage implementations here, because at this point we know that we're only accessing disk storage (so there's no need to use the extra complexity of adding Beam).

codalab/worker/file_util.py
```diff
@@ -196,7 +196,10 @@ def __init__(self, path: str):
             self.index_file_name = index_fileobj.name
             shutil.copyfileobj(
                 FileSystems.open(
-                    path.replace("/contents.tar.gz", "/index.sqlite"),
+                    # path can end in either "contents.tar.gz" (if a directory) or "contents.gz" (if a file).
+                    path.replace("/contents.tar.gz", "/index.sqlite").replace(
```
adiprerepa, line 200: does regex work? `/contents.*` -> `/index.sqlite`

codalab/worker/file_util.py
```diff
@@ -219,40 +222,64 @@ class OpenFile(object):
...
-            and linked_bundle_path.is_archive
-            and linked_bundle_path.archive_subpath
-        ):
-            # If a file path is specified within a .tar.gz file on Azure, open the specified path within the archive.
-            with OpenIndexedTarGzFile(linked_bundle_path.bundle_path) as tf:
+        if linked_bundle_path.uses_beam and linked_bundle_path.is_archive:
+            # Stream an entire, single .gz file from Blob Storage. This is gzipped by default,
+            # so if the user requested a gzipped version of the entire file, just read and return it.
+            if not linked_bundle_path.is_archive_dir and self.gzipped:
+                return FileSystems.open(self.path, compression_type=CompressionTypes.UNCOMPRESSED)
```
adiprerepa, line 251: we seem to do the same thing whether or not the file is a `.gz` or a `.tar.gz` file (`FileSystems.open(...`)). Would it be possible to condense this logic to make it simpler?

adiprerepa (commented):
basically lgtm

codalab/worker/file_util.py
epicfaace: No, because we throw an error on the second condition. It seems simpler to just repeat the code because that way the code reads clearer and we can comment out a description of each scenario.

adiprerepa: fair enough. lgtm then

adiprerepa (approved):

codalab/worker/file_util.py
epicfaace: Probably, but I'd rather actually just restrict it to these two extensions (contents.tar.gz and contents.gz).

PR #3473, opened 2021-05-07
Implement multipart upload for Azure Blob Storage uploads

### LINKED ISSUE #3438: Improve BlobStorageIO to upload files in chunks
We need to improve BlobStorageIO to upload files in chunks. Right now, the way it's implemented in Apache Beam is just to create a temporary file locally, which causes the current available disk space to be a major limitation on how large files can be sent to Blob Storagae:

https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/azure/blobstorageio.py#L649-L657

We should change it to use blob blocks so that the file is uploaded in parts, similar to how S3IO's multipart upload is implemented: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/aws/s3io.py#L588-L615

Moreover, see this table (https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#remarks) for storage limitations.

With the current implementation, we can only store files up to 5000 MiB in a single put blob operation. On the other hand, once we switch to uploading files in chunks (via a block list), each file can be up to 190 TiB.

epicfaace:
Implement multipart upload for Azure Blob Storage uploads in BlobStorageIO. Previously, BlobStorageIO would store everything in a local temporary file before uploading, which made it impossible to upload large files because they couldn't be streamed. To fix this, we implement multipart upload for BlobStorageIO.

This PR just monkey-patches Apache Beam to have this functionality, but we'll eventually want to add a PR to Apache Beam to add this feature upstream. Fixes https://github.com/codalab/codalab-worksheets/issues/3438.


epicfaace:
Original Beam implementation: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/azure/blobstorageio.py#L636

epicfaace:
S3Uploader: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/aws/s3io.py#L536

codalab/lib/beam/blobstorageuploader.py
```diff
@@ -0,0 +1,98 @@
...
+    self.block_list = []
+
+  def put(self, data):
+    # Note that Blob Storage currently can hold a maximum of 100,000 uncommitted blocks.
+    # This means that with this current implementation, we can upload a file with a maximum
+    # size of 10 TiB to Blob Storage. To exceed that limit, we must either increase MIN_WRITE_SIZE
+    # or modify the implementation of this class to call commit_block_list more often (and not
+    # just at the end of the upload). 
+    MIN_WRITE_SIZE = 100 * 1024 * 1024
+    MAX_WRITE_SIZE = 2 * 1024 * 1024 * 1024
```
adiprerepa:
```suggestion
    MAX_WRITE_SIZE = 5 * 1024 * 1024 * 1024
```

codalab/lib/beam/blobstorageuploader.py
```diff
@@ -0,0 +1,98 @@
...
+      chunk = self.buffer.read(MAX_WRITE_SIZE)
+      self._write_to_blob(chunk)
+
+  def _write_to_blob(self, data):
+    # block_id's have to be base-64 strings normalized to have the same length.
+    block_id = base64.b64encode('{0:-32d}'.format(self.block_number).encode()).decode()
+    
+    self._blob_to_upload.stage_block(block_id, data)
+    self.block_list.append(BlobBlock(block_id))
+    self.block_number = self.block_number + 1
```
adiprerepa, line 51: Does azure auto-delete blocks if an upload stops midway?

codalab/lib/beam/blobstorageuploader.py
```diff
@@ -0,0 +1,98 @@
...
+"""
+
+if __name__ == '__main__':
+    from azure.core.exceptions import ResourceExistsError
+    from apache_beam.io.azure.blobstorageio import BlobStorageIO
+    import filecmp
+    import os
+    import shutil
+    
+    os.environ["CODALAB_AZURE_BLOB_CONNECTION_STRING"] = (
```
adiprerepa: remove test

adiprerepa (approved):
mostly lgtm

codalab/lib/beam/blobstorageuploader.py
```diff
@@ -0,0 +1,99 @@
...
+    self.block_list = []
+
+  def put(self, data):
+    # Note that Blob Storage currently can hold a maximum of 100,000 uncommitted blocks.
+    # This means that with this current implementation, we can upload a file with a maximum
+    # size of 10 TiB to Blob Storage. To exceed that limit, we must either increase MIN_WRITE_SIZE
+    # or modify the implementation of this class to call commit_block_list more often (and not
+    # just at the end of the upload). 
+    MIN_WRITE_SIZE = 100 * 1024 * 1024
+    # Maximum block size is 4000 MiB (https://docs.microsoft.com/en-us/rest/api/storageservices/put-block#remarks).
```
epicfaace, line 35: @adiprerepa FYI I changed the max write size to 4000 MiB because that's the max size blob storage supports.

codalab/lib/beam/blobstorageuploader.py
epicfaace: Yes -- looks like it auto-deletes blocks after a week: https://stackoverflow.com/questions/62021724/azure-block-blob-staging-area-block-expiration

PR #3474, opened 2021-05-07
Ensure image block is a separate block

### LINKED ISSUE #3405: Focus on an ImageItem and upload a file
**Describe the bug**
The file bundle will automatically be rendered as an image item as well.
image

**Expected behavior**
The uploaded file should be treated as a regular table row.

yuqijin:

when focusing on an image block, before uploading a bundle, add a new line to separate the image block and the table block.

### Related issues

fixes #3405 

### Screenshots

ezgif com-gif-maker (4)

### Checklist

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -305,6 +305,7 @@ def _create_bundles():
         if not detached:
             if shadow_parent_uuid is None:
                 items = [worksheet_util.bundle_item(bundle_uuid)]
+                # Add a blank line after the image block in the worksheet source to ensure it is a separate block
                 if after_image:
                     items.insert(0, worksheet_util.markup_item(''))
                 local.model.add_worksheet_items(worksheet_uuid, items, after_sort_key)
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -232,6 +232,7 @@ def _create_bundles():
     shadow_parent_uuid = request.query.get('shadow')
     after_sort_key = request.query.get('after_sort_key')
     detached = query_get_bool('detached', default=False)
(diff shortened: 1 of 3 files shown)


codalab/rest/bundles.py
```diff
@@ -303,9 +304,10 @@ def _create_bundles():
         # Add as item to worksheet
         if not detached:
             if shadow_parent_uuid is None:
-                local.model.add_worksheet_items(
-                    worksheet_uuid, [worksheet_util.bundle_item(bundle_uuid)], after_sort_key
-                )
+                items = [worksheet_util.bundle_item(bundle_uuid)]
+                if after_image:
```
epicfaace, line 309: Let's add a comment here to explain what this code is doing

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
```diff
@@ -92,7 +92,10 @@ class NewUpload extends React.Component<{
             if (after_sort_key || after_sort_key === 0) {
                 url += `&after_sort_key=${after_sort_key}`;
             }
-
+            // current focused item is an image block
```
epicfaace, line 95: In the comment, explain why we need to pass after_image to the backend.

epicfaace (approved):

frontend/src/components/worksheets/NewUpload/NewUpload.jsx
yuqijin: solved

codalab/rest/bundles.py
yuqijin: solved

PR #3482, opened 2021-05-07
Fix delete bundle issue

jzwang43:

Delete bundle was not working since we used `done` instead of `then` for an API call. The changes should fix the issue. This pull request also fixes the error display for terminal commands since `error.responseText` was for jquery.

### Checklist


yuqijin:
lgtm. Thanks for fixing it.

yuqijin (approved):

Issue #3477, opened 2021-05-07
Make zip_util.unpack streaming and able to output to .tar.gz files


epicfaace:
Change the implementation of zip_util.unpack (https://github.com/codalab/codalab-worksheets/blob/570edaacdeb608a1eabacacaf838d02845410484/codalab/lib/zip_util.py#L51) so that it is:
- streaming. The implementation of `unzip_directory`, for example, writes to a temporary file, so it is limited by the size of the input zip file. We should instead be able to just unzip it by reading each file entry, one by one, but most unzippers require random access as they first read the central directory of the .zip file first. Check if ZipFile supports this, otherwise implement something like https://github.com/madler/sunzip in Python.
- be able to construct an output .tar.gz fileobj instead of writing to disk. This will be needed in order to finally stream the .tar.gz file directly to Blob Storage.

PR #3485, opened 2021-05-08
Make unpacking zip files streaming

epicfaace:
Make unpacking zip files streaming; currently, zip files have to be stored in a temporary file before they can be unpacked. I created a wrapper around python's ZipFile called StreamingZipFile that allows reading a .zip file by just streaming it in.

Normally, using ZipFile to read files in a .zip archive requires random access
to the file, because ZipFile reads the central directory at the end of the archive first
to retrieve the metadata of each file in the archive to create the list of ZipInfo objects.

Instead, StreamingZipFile can be iterated over in order to read each file one by one,
using each file header to construct ZipInfo objects for each file. Sample usage:

```python
with StreamingZipFile(fileobj) as zf:
    for zinfo in zf:
        print(zinfo)
        print(zf.open(zinfo).read())
```

This is analogous to the `r|` mode that can be used to open tar files in Python.

Addresses part of https://github.com/codalab/codalab-worksheets/issues/3477.


codalab/lib/zip_util.py
```diff
@@ -48,20 +48,16 @@ def strip_archive_ext(path):
     raise UsageError('Not an archive: %s' % path)
 
 
-def unpack(ext: str, source: Union[str, IO[bytes]], dest_path: str):
+def unpack(ext: str, source: IO[bytes], dest_path: str):
```
epicfaace, line 51: The only usage of `unpack` now passes in a fileobj to `source`.

codalab/worker/file_util.py
```diff
@@ -148,7 +149,7 @@ def zip_directory(
             raise IOError(e.output)
 
 
-def unzip_directory(fileobj_or_name, directory_path, force=False):
+def unzip_directory(fileobj: IO[bytes], directory_path: str, force: bool = False):
```
epicfaace, line 152: The only usage of `unzip_directory` now passes in a fileobj as the first argument, not a string.

setup.cfg
```diff
@@ -33,7 +33,7 @@ ignore_missing_imports = True
 
 [flake8]
 max-line-length = 200
-exclude = venv/*,var/*,alembic/*,frontend/*,codalab/lib/beam/ratarmount.py
+exclude = venv/*,var/*,alembic/*,frontend/*,codalab/lib/beam/ratarmount.py,codalab/lib/beam/streamingzipfile.py
```
epicfaace, line 36: I'm excluding streamingzipfile.py from black / flake8 because much of the code in it comes from and is written in the style of https://github.com/python/cpython/blob/3.9/Lib/zipfile.py. Eventually, we may want to contribute a PR back to cpython that adds this feature in to zipfile (see https://github.com/codalab/codalab-worksheets/issues/3488).

tests/unit/worker/file_util_test.py
```diff
@@ -171,7 +173,8 @@ def test_has_files(self):
         self.assertNotIn('b.txt', output_dir_entries)
         self.assertTrue(os.path.exists(os.path.join(output_dir, 'dir1', 'f1')))
         self.assertFalse(os.path.exists(os.path.join(output_dir, 'dir1', 'f2')))
-        self.assertTrue(os.path.islink(os.path.join(output_dir, 'a-symlink.txt')))
+        if not self.skip_symlinks:
+            self.assertTrue(os.path.islink(os.path.join(output_dir, 'a-symlink.txt')))
```
epicfaace, line 178: Technically, this is a breaking change -- when uploading a .zip file, symlinks are no longer preserved -- but I don't think this is a feature that our users rely on. We used to never preserve symlinks in .zip file uploads; we only started doing so after https://github.com/codalab/codalab-worksheets/pull/2952, and we never documented that fact.

So I think it should be OK to remove support for symlinks now. (We can't preserve symlinks anymore since we're using Python's ZipFile now, which doesn't have support for symlinks -- unless we make further modifications to it).

codalab/lib/beam/streamingzipfile.py
```diff
@@ -0,0 +1,169 @@
...
+    # Stores the position of the next file header to be read.
+    _next_header_pos = 0
+
+    # Whether all entries have been loaded.
+    _loaded = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        try:
+            self.fp.tell()
```
adiprerepa, line 63: is this basically a null check to initialize it if not null? Is this the recommended way?

codalab/lib/beam/streamingzipfile.py
```diff
@@ -0,0 +1,169 @@
...
+
+        (x.extract_version, x.reserved, x.flag_bits, x.compress_type, t, d,
+        x.CRC, x.compress_size, x.file_size) = fheader[1:10]
+        if x.extract_version > MAX_EXTRACT_VERSION:
+            raise NotImplementedError("zip file version %.1f" %
+                                    (x.extract_version / 10))
+
+        # Convert date/time code to (year, month, day, hour, min, sec)
+        x._raw_time = t
+        x.date_time = ( (d>>9)+1980, (d>>5)&0xF, d&0x1F,
```
adiprerepa, line 145: does it make sense to add a comment explaining this logic? simpletons like me may have a hard time understanding

tests/unit/worker/file_util_test.py
adiprerepa: what's wrong with making further modifications to `ZipFile`? I think we should be careful here

adiprerepa (commented):
nice

tests/unit/worker/file_util_test.py
epicfaace: symlinks themselves aren't part of the .zip standard. They're just sometimes added as part of extra metadata fields.

In fact, normally if you create a .zip file, symlinks aren't preserved by default unless you actually run it with the `zip --symlinks` option (see https://serverfault.com/questions/265675/how-can-i-zip-compress-a-symlink).

The majority of cases in which we are handling .zip files anyway will be through uploads that happen through the web interface -- and in this case, the zip file is created without the `--symlinks` option.

So what this really breaks is the use case in which someone 1) created a zip file with symlinks in it with `zip --symlinks` and then 2) uploads it to codalab and it gets unpacked. It seems like this is an edge-enough use case (which would only have worked since October 2020) that we can just remove it.

codalab/lib/beam/streamingzipfile.py
epicfaace: It's a check to see if the `.tell()` method is implemented on the file object. Otherwise, the file object is wrapped in a wrapper that implements `.tell()` for it.

codalab/lib/beam/streamingzipfile.py
epicfaace: I don't understand it either 😆 It comes (as does most of the implementation of this function) from https://github.com/python/cpython/blob/048f54dc75d51e8a1c5822ab7b2828295192aaa5/Lib/zipfile.py#L1316 though, so maybe I'll just comment a link to it.

tests/unit/worker/file_util_test.py
adiprerepa: makes sense to me

codalab/lib/beam/streamingzipfile.py
adiprerepa: ok, great

epicfaace:
@adiprerepa can you re-review?

jzwang43 (approved):

Issue #3479, opened 2021-05-07
Remove unnecessary Apache Beam code


epicfaace:
We have some unnecessary Apache Beam code -- for example, see `un_bz2_file`, where we don't really need to write to Blob Storage. https://github.com/codalab/codalab-worksheets/blob/3e886d8bd20fc2c99013ed74ff00fd19e027b662/codalab/worker/file_util.py#L322

We should take a pass around the codebase and remove unneeded FileSystems code, and replace them with their disk-based equivalents (such as `open`).

PR #3490, opened 2021-05-10
Remove unneeded Beam code from un_bz2_file

epicfaace:
Remove unneeded Beam code from un_bz2_file, as un_bz2_file is only called with reference to local file paths. Addresses part of https://github.com/codalab/codalab-worksheets/issues/3479


epicfaace:
@jzwang43 can you take a look?

jzwang43 (approved):
Looks good.

PR #3491, opened 2021-05-10
Only allow relative urls in redirect_uri

### LINKED ISSUE #3481: Only allow relative urls in redirect_uri
We currently allow absolute urls when calling the logout endpoint, for example, see https://worksheets.codalab.org/rest/account/logout?redirect_uri=https://google.com

This should not be allowed as it's a security risk -- see https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html. We've also been seeing abuse of this endpoint in our logs.

Moreover, when you actually click the "Logout" button, we only use relative urls in redirect_uri -- it sends you to https://worksheets-dev.codalab.org/rest/account/logout?redirect_uri=%2Fusers%2Fcodalab.

jzwang43:

We should disallow redirect to absolute URIs to prevent potential exploits.

### Related issues
fixes #3481 

### Screenshots

image

### Checklist


codalab/lib/server_util.py
```diff
@@ -166,11 +166,19 @@ def bottle_patch(path=None, **options):
     return app().route(path, 'PATCH', **options)
 
 
+def safe_uri(redirect_uri):
+    """Check if an URI is relative, otherwise raise an error.
+    """
+    if '://' in redirect_uri or '//' in redirect_uri:
```
epicfaace: Let's use this method instead: https://stackoverflow.com/questions/8357098/how-can-i-check-if-a-url-is-absolute-using-python/8357518

epicfaace (changes requested):

codalab/lib/server_util.py
jzwang43: Updated, thanks!

epicfaace (approved):

PR #3494, opened 2021-05-11
Update developers on "About" page

epicfaace:

Update developers on "About" page, as they were outdated.


yuqijin (approved):

adiprerepa (approved):
sauce

PR #3496, opened 2021-05-11
Blob Storage Upload

### LINKED ISSUE #3476: Support azure blob uploading in UploadManager.upload_to_bundle_store()
Support azure blob uploading in UploadManager.upload_to_bundle_store(). When the `use_azure_blob_beta` argument is set to true, the bundle should be uploaded to Blob Storage and an index should be created for it.

epicfaace:
Blob Storage upload. Fixes https://github.com/codalab/codalab-worksheets/issues/3476 by supporting azure blob uploading in UploadManager.upload_to_bundle_store().

This PR is based on https://github.com/codalab/codalab-worksheets/pull/3489, so please merge that one first before this one.


requirements.txt
```diff
@@ -6,7 +6,7 @@ marshmallow-jsonapi==0.15.1
 marshmallow==2.15.1
 setuptools>=40.0.0
 argcomplete==1.12.1
-indexed_gzip==1.5.3
+indexed_gzip_fileobj_fork_epicfaace==1.5.4
```
epicfaace, line 9: This fork includes a fix for https://github.com/codalab/codalab-worksheets/issues/3514. This fix was required, otherwise the tests in this PR wouldn't pass.

codalab/lib/upload_manager.py
```diff
@@ -1,74 +1,113 @@
...
+            unpack_archive (bool): Whether fileobj is an archive that should be unpacked.
         """
-        bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
+        raise NotImplementedError
+
+    def upload_to_bundle_store(self, bundle: Bundle, source: Source, git: bool, unpack: bool):
+        """Uploads the given source to the bundle store.
+        Given arguments are the same as UploadManager.upload_to_bundle_store()"""
         try:
+            # bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
```
adiprerepa: remove
```suggestion
```

codalab/lib/upload_manager.py
```diff
@@ -1,74 +1,113 @@
...
             is_url, is_fileobj, filename = self._interpret_source(source)
             if is_url:
                 assert isinstance(source, str)
                 if git:
-                    file_util.git_clone(source, bundle_path)
+                    bundle_path = self._update_and_get_bundle_location(bundle, is_directory=True)
```
adiprerepa, line 60: `is_directory` is true because all git repos are directories?

codalab/lib/upload_manager.py
```diff
@@ -1,74 +1,113 @@
...
-            is_directory = os.path.isdir(bundle_path)
-            self._bundle_model.update_bundle(
-                bundle, {'storage_type': StorageType.DISK_STORAGE.value, 'is_dir': is_directory},
-            )
         except UsageError:
             if os.path.exists(bundle_path):
                 path_util.remove(bundle_path)
             raise
 
+    def _update_and_get_bundle_location(self, bundle: Bundle, is_directory: bool) -> str:
```
adiprerepa, line 85: why did we combine this functionality (update & get)? would it be better for the code to be more granular?

codalab/lib/upload_manager.py
```diff
@@ -82,9 +121,100 @@ def _interpret_source(self, source: Source):
...
+
+    def write_fileobj(
+        self, source_ext: str, source_fileobj: IO[bytes], bundle_path: str, unpack_archive: bool
+    ):
+        if unpack_archive:
+            output_fileobj = zip_util.unpack_to_archive(source_ext, source_fileobj)
+        else:
+            output_fileobj = GzipStream(source_fileobj)
+        # Write archive file.
+        with FileSystems.create(bundle_path, compression_type=CompressionTypes.UNCOMPRESSED) as out:
```
adiprerepa, line 167: aren't archives compressed?

tests/unit/lib/path_util_test.py
```diff
@@ -129,6 +129,9 @@ def test_single_file(self):
         )
         self.assertEqual(linked_bundle_path.is_archive, True)
         self.assertEqual(linked_bundle_path.is_archive_dir, False)
+        self.assertEqual(
```
adiprerepa, line 132: do we also need to check for the existence of the archive?

tests/unit/lib/upload_manager_test.py
```diff
@@ -1,66 +1,90 @@
...
+    def listdir(self):
+        """List the files in the current bundle location."""
+        raise NotImplementedError
+
+    @property
+    def bundle_location(self):
+        """Get bundle location of the currently created bundle."""
+        return self.codalab_manager.bundle_store().get_bundle_location(self.bundle.uuid)
+
+    def setUp(self):
```
adiprerepa, line 51: is this camel case for a reason?
```suggestion
    def set_up(self):
```

tests/unit/lib/upload_manager_test.py
```diff
@@ -119,20 +141,69 @@ def test_url_tar_gz_should_not_simplify_archives(self):
...
+                        expected_contents,
+                    )
+
+    def listdir(self):
+        with FileSystems.open(
+            self.bundle_location, compression_type=CompressionTypes.UNCOMPRESSED
+        ) as f, tarfile.open(fileobj=f, mode='r:gz') as tf:
+            # Remove "." entry and "./" prefixes to make the file listing similar to that
+            # in an ordinary directory on disk.
+            return [i.replace("./", "") for i in tf.getnames() if i != "."]
```
adiprerepa, line 209: why do we have the `if` here? that's kind of implicit, right?

adiprerepa (commented):
great stuff

codalab/lib/upload_manager.py
epicfaace: Yeah

codalab/lib/upload_manager.py
epicfaace: Yes, but the fileobj we're writing to is already compressed. If we don't specify `compression_type=CompressionTypes.UNCOMPRESSED` to `FileSystems.create()`, Apache Beam will automatically _re-compress_ the given fileobj.

tests/unit/lib/path_util_test.py
epicfaace: The archive won't exist here as we haven't uploaded anything yet. This is just testing the parsing logic of `parse_linked_bundle_url`

tests/unit/lib/upload_manager_test.py
epicfaace: Yes, see https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

tests/unit/lib/upload_manager_test.py
epicfaace: We want to not return the `.` entry, so that the file listing is similar to that in an ordinary directory that one gets by running `os.listdir`. See https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions for more about the syntax.

codalab/lib/upload_manager.py
adiprerepa: makes sense

tests/unit/lib/upload_manager_test.py
adiprerepa: oh, I see. List comprehensions trip me up.

adiprerepa (approved):
lgtm

codalab/lib/upload_manager.py
epicfaace: Because they're only ever called together. Additionally, the "update" part of the code affects the result of "getting" the bundle location (get_bundle_location constructs a bundle location based on the storage_type and is_directory fields).

PR #3501, opened 2021-05-11
Read Runtime Environment Variables for Frontend

epicfaace:
Sample code that shows how you could pass runtime env variables to the frontend container.

--Jizhen we need to allow environment variables to be read by the frontend image at runtime instead of build time.


frontend/public/index.prod.html
```diff
@@ -69,5 +68,6 @@
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
     <script type="text/javascript" src="%PUBLIC_URL%/js/collapsible.js"></script>
+    <script type="text/javascript" src="%PUBLIC_URL%/js/env.js"></script>
```
epicfaace: we can probably just have one index.html with this setup (instead of needing an index.prod.html and an index.html)

jzwang43: Yeah I just removed `index.prod.html` and cleaned up the code.

docker_config/compose_files/docker-compose.yml
```diff
@@ -236,7 +237,7 @@ services:
 
   frontend:
     image: codalab/frontend:${CODALAB_VERSION}
-    command: serve -s build -l ${CODALAB_FRONTEND_PORT}
+    command: bash -c "node setup-frontend.js && serve -s build -l ${CODALAB_FRONTEND_PORT}"
```
jzwang43, line 240: By running `setup-frontend.js` we read react app environment variables from the runtime and allows the production website to read them.

codalab_service.py
```diff
@@ -136,7 +136,7 @@ class CodalabArg(object):
     """CodalabArg defines the arguments needed for starting the service.
 
     Attributes:
-        frontend (bool): frontend arg will be passed down to the frotnend container at build time, prefixed by REACT_APP_.
+        frontend (bool): frontend arg will be passed down to the frotnend container, prefixed by REACT_APP_.
```
```suggestion
        frontend (bool): frontend arg will be passed down to the frontend container, prefixed by REACT_APP_.
```

setup-frontend.js
```diff
@@ -0,0 +1,8 @@
+const FRONTEND_ARGS = ["REACT_APP_CODALAB_RECAPTCHA_SITE_KEY"];
```
epicfaace: Instead of hardcoding this, let's just iterate through process.env and pick out the environment variables that start with REACT_APP_.

frontend/src/components/SignUp.js
```diff
@@ -192,7 +192,11 @@ export class SignUp extends React.Component {
                         </div>
                         <ReCAPTCHA
                             ref={this.recaptchaRef}
-                            sitekey={process.env.REACT_APP_CODALAB_RECAPTCHA_SITE_KEY}
+                            sitekey={
+                                process.env.NODE_ENV === 'production'
```
epicfaace: 1. Is it possible to just always use `window.env.REACT_APP_CODALAB_RECAPTCHA_SITE_KEY`, even for development? This would simplify the code a bunch. We could do this by modifying the command in docker-compose.dev.yml to include `node setup-frontend.js` in it.
2. If we can do 1., then we can actually remove the REACT_APP_ prefix altogether (so we can do something like `window.env.CODALAB_RECAPTCHA_SITE_KEY`, further simplifying env variable logic).

epicfaace: Actually, let's not do suggestion 2 (removing the REACT_APP_ prefix). This is because then, in setup-frontend.js, we'd have no way to distinguish between frontend (non-secret) and non-frontend (secret) env variables

setup-frontend.js
jzwang43: Updated.

frontend/src/components/SignUp.js
jzwang43: Just applied suggestion 1.

setup-frontend.js
```diff
@@ -1,8 +1,14 @@
-const FRONTEND_ARGS = ["REACT_APP_CODALAB_RECAPTCHA_SITE_KEY"];
+const fs = require("fs");
 
-fs = require("fs");
-vars = FRONTEND_ARGS.map(
-  (arg) => `window.env.${arg}="${process.env[arg]}"`
+const path =
```
epicfaace, line 6: Add a comment here explaining what this script does

epicfaace (commented):
Looks good! I can't approve this PR because it was made originally by me, but consider it approved

adiprerepa (approved):
saucy

PR #3502, opened 2021-05-13
Add Delete Account Option for non-root user

### LINKED ISSUE #3459: Users should be allowed to delete their account themselves
Allow users to run `ufarewell` on themselves, check to make sure it’s restricted to root otherwise.

yuqijin:

The user can use `ufarewell` cmd or the `Delete your account` button on the account page to delete his own account if the user does not own any bundles, worksheets, or groups.

Provide a delete account dialog to avoid unintentional deletion.

### Related issues

fixes #3459

### Screenshots

ezgif com-gif-maker (5)

### Checklist

### DIFF
--- a/codalab/rest/users.py
+++ b/codalab/rest/users.py
@@ -101,10 +101,7 @@ def delete_user():
 
     request_user_id = request.user.user_id
 
-    if not (
-        request_user_id == local.model.root_user_id
-        or (user_ids == [request_user_id])
-    ):
+    if not (request_user_id == local.model.root_user_id or (user_ids == [request_user_id])):
         abort(http.client.UNAUTHORIZED, 'As a non-root user, you can only delete your own account.')
 
     for user_id in user_ids:
--- a/codalab/rest/users.py
+++ b/codalab/rest/users.py
@@ -103,7 +103,7 @@ def delete_user():
 
     if not (
         request_user_id == local.model.root_user_id
-        or (len(user_ids) == 1 and request_user_id == user_ids[0])
+        or (user_ids == [request_user_id])
     ):
         abort(http.client.UNAUTHORIZED, 'As a non-root user, you can only delete your own account.')
 
--- a/codalab/rest/users.py
+++ b/codalab/rest/users.py
@@ -101,8 +101,11 @@ def delete_user():
 
     request_user_id = request.user.user_id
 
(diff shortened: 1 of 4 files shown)


adiprerepa:
why did the gif show an error?

yuqijin:
> why did the gif show an error?

@adiprerepa Because the user owns bundles/worksheets/groups, so the user cannot delete the account.

codalab/rest/users.py
```diff
@@ -101,8 +101,11 @@ def delete_user():
 
     request_user_id = request.user.user_id
 
-    if request_user_id != local.model.root_user_id:
-        abort(http.client.UNAUTHORIZED, 'Only root user can delete other users.')
+    if not (
+        request_user_id == local.model.root_user_id
+        or (len(user_ids) == 1 and request_user_id == user_ids[0])
```
```suggestion
        or (user_ids == [request_user_id])
```

codalab/rest/users.py
```diff
@@ -135,7 +138,7 @@ def delete_user():
 
         groups = local.model.batch_get_groups(owner_id=user_id)
         if groups is not None and len(groups) > 0:
-            group_uuids = [group.uuid for group in groups]
+            group_uuids = [group['uuid'] for group in groups]
```
epicfaace, line 138: Why this change? Was this a bug earlier?

frontend/src/components/UserInfo.js
```diff
@@ -342,4 +353,83 @@ class AccountProfileField extends React.Component {
...
+            .catch((error) => alert(error));
+    };
+
+    return (
+        <div>
+            <Divider style={{ marginBottom: '30px' }} />
+            <h2 style={{ color: '#bc3638' }}>Delete account</h2>
+            <p>
+                To be safe, you can only delete your account if you do not own any bundles,
+                worksheets, or groups.
```
epicfaace, line 390: what does owning a group mean? Does this just mean being part of a group?

codalab/rest/users.py
yuqijin: yes, some bug existing due to the object/dict difference

frontend/src/components/UserInfo.js
yuqijin: A group only has one owner, and owning a group means that the user creates the group

epicfaace (approved):

PR #3504, opened 2021-05-14
Catch Parsing Error of Datetime

### LINKED ISSUE #3483: keyword search formatted wrong throws error
https://sentry.io/organizations/codalab/issues/2387369760/

better error cataloging/don't throw errors when they should be user-facing

jzwang43:

isoparse is throwing an unhandled error if an invalid date is provided, triggering sentry.io

### Related issues

fixes #3483 

### Screenshots

image

### Checklist

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -531,7 +531,7 @@ class BundleModel(object):
             elif key in ('.before', '.after'):
                 try:
                     target_datetime = parser.isoparse(value)
-                except Exception:
+                except ValueError:
                     raise UsageError("Unable to parse datatime, please check the format.")
 
                 subclause = None
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -532,7 +532,9 @@ class BundleModel(object):
                 try:
                     target_datetime = parser.isoparse(value)
                 except ValueError:
-                    raise UsageError("Unable to parse datatime, please check the format.")
+                    raise UsageError(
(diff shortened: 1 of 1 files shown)


codalab/model/bundle_model.py
```diff
@@ -529,7 +529,10 @@ def make_condition(key, field, value):
                         alias(select([cl_worksheet_item.c.bundle_uuid]).where(condition))
                     )
             elif key in ('.before', '.after'):
-                target_datetime = parser.isoparse(value)
+                try:
+                    target_datetime = parser.isoparse(value)
+                except Exception:
+                    raise UsageError("Unable to parse datatime, please check the format.")
```
```suggestion
                    raise UsageError("Unable to parse datetime. Datetime must be specified as an ISO-8601 datetime string such as YYYY-MM-DD.")
```

codalab/model/bundle_model.py
```diff
@@ -529,7 +529,10 @@ def make_condition(key, field, value):
                         alias(select([cl_worksheet_item.c.bundle_uuid]).where(condition))
                     )
             elif key in ('.before', '.after'):
-                target_datetime = parser.isoparse(value)
+                try:
+                    target_datetime = parser.isoparse(value)
+                except Exception:
```
epicfaace: more specific exception

epicfaace (changes requested):

codalab/model/bundle_model.py
jzwang43: Updated.

codalab/model/bundle_model.py
jzwang43: Changed to `ValueError` which is thrown by `isoparse`.

epicfaace (approved):

PR #3516, opened 2021-05-17
Blob Storage download

### LINKED ISSUE #3478: Update download_manager.stream_tarred_gzipped_directory to work with Blob Storage
Update the logic in download_manager.stream_tarred_gzipped_directory to work with Blob Storage. Right now, this function only works with local paths (see https://github.com/codalab/codalab-worksheets/blob/5f72f4d2d86857ccef2abc016a2b071c1f3dc54c/codalab/lib/download_manager.py#L170). We need to modify it to use the `OpenFile` context manager instead.

epicfaace:
Implements download_manager changes needed for blob storage. Fixes https://github.com/codalab/codalab-worksheets/issues/3478.

This PR is based on https://github.com/codalab/codalab-worksheets/pull/3496/


codalab/lib/beam/ratarmount.py
```diff
@@ -753,6 +753,7 @@ def _createIndex(
                     # fmt:on
                 )
             except tarfile.ReadError:
+                print("RE")
```
adiprerepa: is this the correct way to handle this? Might be ambiguous to the reader
```suggestion
                print("Tar read error")
```

codalab/lib/upload_manager.py
```diff
@@ -53,6 +53,7 @@ def upload_to_bundle_store(self, bundle: Bundle, source: Source, git: bool, unpa
         """Uploads the given source to the bundle store.
         Given arguments are the same as UploadManager.upload_to_bundle_store()"""
         try:
+            # bundle_path = self._bundle_store.get_bundle_location(bundle.uuid)
```
adiprerepa: remove lol
```suggestion
```

codalab/worker/file_util.py
```diff
@@ -268,7 +268,15 @@ def __enter__(self) -> IO[bytes]:
...
                     return GzipStream(fs) if self.gzipped else fs
         else:
-            # Stream a file from disk storage.
+            # Stream a directory or file from disk storage.
+            if os.path.isdir(self.path):
+                if not self.gzipped:
+                    raise IOError("Directories must be gzipped.")
+                return tar_gzip_directory(self.path)
+            if self.gzipped:
+                raise IOError(
```
adiprerepa, line 278: does it make sense to add a comment to the function so future callers know this?

adiprerepa (commented):
awesome sauce

codalab/lib/beam/ratarmount.py
epicfaace:
```suggestion
```

epicfaace: I meant to remove this. This was the original code from the ratarmount library and I just added it in for debugging.

codalab/worker/file_util.py
epicfaace: done

adiprerepa (approved):

PR #3517, opened 2021-05-17
Add subset of Blob Storage E2E tests

epicfaace:
I've added a subset of Blob Storage E2E tests (from test_cli.py), which now work since we have both upload and download functionality working 🎉. I had to also change some logic related to hashing directories / removing directories to get the E2E tests to work.

This PR is based on https://github.com/codalab/codalab-worksheets/pull/3516, so merge that one first.


.github/workflows/test.yml
```diff
@@ -295,6 +295,67 @@ jobs:
...
+          # - resources
+          # - memoize
+          # - copy netcat netcurl
+          # - edit
+          # - open wopen
+    steps:
+      - name: Clear free space
+        run: |
+            sudo rm -rf /opt/ghc
+            df -h
```
adiprerepa, line 320: why do we run `df -h` here? do we need the output?

adiprerepa: ah well it exists in the rest of the test setup as well

adiprerepa (approved):
:confetti_ball:

.github/workflows/test.yml
epicfaace: Yeah, don't think it's needed, but maybe we can remove in a later refactor

PR #3519, opened 2021-05-17
DO NOT MERGE

### LINKED ISSUE #1073: DO NOT MERGE

penghaolu:
### Screenshots
Screen Shot 2021-05-17 at 4 58 54 PM

### DIFF
--- a/frontend/src/components/worksheets/items/MarkdownItem.js
+++ b/frontend/src/components/worksheets/items/MarkdownItem.js
@@ -23,8 +23,10 @@ marked.setOptions({
     breaks: true,
     smartLists: true,
     smartypants: true,
-    highlight: function(code) {
-        return hljs.highlightAuto(code).value;
+    highlight: function(code, lang) {
+        if (lang) {
+            return hljs.highlight(lang, code).value;
+        }
     },
 });
 
--- a/frontend/src/components/worksheets/items/MarkdownItem.js
+++ b/frontend/src/components/worksheets/items/MarkdownItem.js
@@ -2,6 +2,8 @@ import * as React from 'react';
 import Immutable from 'seamless-immutable';
 import { worksheetItemPropsChanged, getAfterSortKey } from '../../../util/worksheet_utils';
 import marked from 'marked';
+import hljs from 'highlight.js';
+import 'highlight.js/styles/atom-one-light.css';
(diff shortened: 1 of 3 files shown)


frontend/src/components/worksheets/items/MarkdownItem.js
```diff
@@ -12,6 +14,20 @@ import TextEditorItem from './TextEditorItem';
...
+marked.setOptions({
+    renderer: new marked.Renderer(),
+    gfm: true,
+    pedantic: false,
+    tables: true,
+    breaks: true,
+    smartLists: true,
+    smartypants: true,
+    highlight: function(code) {
+        return hljs.highlightAuto(code).value;
```
epicfaace: Does this function end up reading the "```python" at the beginning of the code block? It appears that this function, rather, just auto-detects language based on the content. Do you think it'd be possible to change this to do the former, and if no language is specified, just not do any highlighting?

epicfaace (commented):
Can you also specify which initial worksheet source code made the screenshots show up?

frontend/src/components/worksheets/items/MarkdownItem.js
penghaolu: For the screenshot, the source code is the following
<pre>
```python
x = [1 for i in range(0, 100)]
# This program prints Hello, world!
print('Hello, world!')
```

```js
componentDidMount() {
    this.processMathJax();
}
componentDidUpdate() {
    this.processMathJax();
    if (this.props.focused) this.capture_keys();
}
```
</pre>

As for the language, when I was reading the hightlight.js documentation, it stated that the language is provided through the class attribute, which the original already does. I believe that the package will always try to automatically detect the language of the code block if no language is provided through the class attribute unless the block is given a "plaintext" class.

penghaolu: I think it should be possible to default to "plaintext" as the provided language prop if none is provided by the user. Another possible solution would be to edit the documentation so that ```plaintext would denote the start of a block of plaintext code without highlighting.

penghaolu: Pushed a new update that disables automatic detection and has the desired functionality (highlights with the language given in code block, does not highlight if none provided).

PR #3520, opened 2021-05-18
Blob Storage: fix dependency download for make bundles

### LINKED ISSUE #3443: Implement support for downloading dependencies through blob storage (bundle manager handling "cl make")
See [this TODO](https://github.com/codalab/codalab-worksheets/pull/3433/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R306). This logic includes:
- creating make bundles from bundles on Blob Storage
- ensuring that workers can download / mount bundles properly from blob storage

This code for accomplishing this task is available in this PR (https://github.com/codalab/codalab-worksheets/pull/3308).

epicfaace:
Fixes https://github.com/codalab/codalab-worksheets/issues/3443: Implement support for downloading dependencies through blob storage (bundle manager handling "cl make").

I'm also enabling the rest of the Blob Storage E2E tests, as they should now work (except for `link`, which I will need to fix in a future PR).

This PR is based off of https://github.com/codalab/codalab-worksheets/pull/3517, so merge that one first.


codalab/lib/beam/mockblobstoragefilesystem.py
```diff
@@ -34,7 +34,10 @@ def _azfs_to_local(self, path):
         if not path.startswith("azfs://"):
             return path
         path = MockBlobStorageFileSystem.AZFS_MOCK_LOCATION + path[len("azfs://") :]
-        os.makedirs(os.path.dirname(path), exist_ok=True)
+        try:
```
epicfaace, line 37: This is only used during tests. This change is needed because `os.makedirs(os.path.dirname(path), exist_ok=True)` fails when the path has a trailing slash.

epicfaace:
GHA error:

```

>> cl info -f / 0x87575bb644804894acf73d6deacd5815
 (exit code 0, expected 0)



[!] ERROR: expected 'This is a simple text file for CodaLab. It has two lines in it.', but got ''

```

codalab/worker/file_util.py
```diff
@@ -463,7 +469,7 @@ def ensure_ends_with_newline(lines, remove_line_without_newline=False):
                 lines = tail_lines
         else:
             try:
-                lines = fileobj.readlines()
+                lines = fileobj.read().splitlines(True)
```
epicfaace, line 491: I had to make this change because the underlying fileobj when reading from a tar archive on Blob Storage is BytesBuffer, which doesn't support the `readlines()` operation at the moment. This code is equivalent.

adiprerepa (approved):
lgtm, @teetone should take a look as well for good measure

PR #3521, opened 2021-05-18
Show hidden fields for root

### LINKED ISSUE #3518: Show error_traceback by default only for the root user
Right now, error_traceback is a hidden field and only accessible through `cl info -f error_traceback`. We should show it on the UI and on the cli output for `cl info` if the current user is the root user, for ease of debugging.

Discussion on Slack from @epicfaace :

> I know this was against my original suggestion on https://github.com/codalab/codalab-worksheets/pull/3257, but I wonder if maybe we should show error_traceback by default, though. This might be useful for someone on call, who can, for example, just click through failed bundles and easily view tracebacks for them without having to run an extra CLI command
> 
> 3:49
> It's also a bit hard for, say, a new developer to the codalab team to remember / know to do cl info -f error_traceback  (so this increases the learning curve for them). It might be easier if this message just comes up "by default" so debugging is much simpler (edited) 
> 3:50
> maybe we could either 1) just show error_traceback by default for everyone, or 2) show error_traceback by default only for the root user?

jzwang43:

Right now, error_traceback is a hidden field and only accessible through cl info -f error_traceback. We should show it on the UI and on the cli output for cl info if the current user is the root user, for ease of debugging.

### Related issues

fixes #3518 

### Checklist


codalab/lib/worksheet_util.py
```diff
@@ -186,7 +186,7 @@ def get_worksheet_lines(worksheet_info):
     return lines
 
 
-def get_formatted_metadata(cls, metadata, raw=False):
+def get_formatted_metadata(cls, metadata, raw=False, show_hidden=False):
```
epicfaace, line 189: Is show_hidden used at all?

jzwang43: Yeah, if user is root then show_hidden will be true else false.

codalab/lib/bundle_cli.py
```diff
@@ -2368,7 +2368,11 @@ def print_basic_info(self, client, info, raw):
 
         # Metadata fields (standard)
         cls = get_bundle_subclass(info['bundle_type'])
-        for key, value in worksheet_util.get_formatted_metadata(cls, metadata, raw):
+
+        # Show all hidden fields for root user
+        is_root = client.fetch('user')['is_root_user']
```
epicfaace: Let's rename is_root to show_hidden for clarity.

epicfaace (approved):

PR #3529, opened 2021-05-19
Remove jquery show and hide

### LINKED ISSUE #3528: Remove jQuery .show() and .hide()
Remove jQuery .show() and .hide() functions, this is one of the steps to remove jQuery completely.
Related to fixes #1037

jzwang43:

We want to remove jquery from the code base.

### Related issues

fixes #3528 

### Checklist


frontend/src/components/FileBrowser/FileBrowser.js
```diff
@@ -236,16 +234,18 @@ export class FileBrowser extends React.Component<
...
         }
         return (
-            <div className='file-browser'>
-                {checkbox}
-                {header}
-                <div className={content_class_name}>
-                    <div className='panel panel-default'>
-                        {bread_crumbs}
-                        <div className='panel-body'>{file_browser}</div>
+            this.state.hidden || (
```
epicfaace: If hidden is false, to be safe, we should return null with a ternary operator. Right now, we return false if hidden is false, and I'm not sure if it's valid to return false from a React component.

epicfaace (commented):
Thanks, can you update the PR name / description as well?

frontend/src/components/FileBrowser/FileBrowser.js
jzwang43: This is safe I believe, I have been using this for a long time. Check this: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

yuqijin: I think it would be safer to return `<div>{false}</div>`, as shown here https://react-cn.github.io/react/tips/false-in-jsx.html

jzwang43: @yuqijin I believe this actually explains why rendering false is safe and false will not be rendered as string. React will ignore false and null render output.

epicfaace: Hmm, but that example uses && instead of ||. Do both have the same behavior in this case?

jzwang43: I think this applies to any boolean expressions for conditional rendering. This is just De Morgan's laws in this case since we are checking for `hidden` .

yuqijin (approved):

frontend/src/components/FileBrowser/FileBrowser.js
epicfaace: @jzwang43 I see your point, but I think to be consistent, we should use && as we do that in the rest of the codebase. It also reads easier -- "if !this.state.hidden, then this component is rendered" -- which is probably why the React docs just default to talking about && and don't even mention ||. https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator Given that the docs use &&, it seems like && is a more standard convention than using || with React conditional rendering.

```suggestion
            !this.state.hidden && (
```

PR #3533, opened 2021-05-21
Stricter Linting

### LINKED ISSUE #3527: Stricter linting for code in the frontend
- syntax and whitespace
- code complexity

jzwang43:

We need stricter linting rules to improve frontend code quality.

I suppressed many warning for now since these issues require non-trivial fixes. The next step will be to fix the errors that are suppressed for now.

### Related issues

fixes #3527 

### Checklist


frontend/.eslintrc.json
```diff
@@ -1,13 +1,30 @@
 {
-    "extends": "react-app",
-    "rules": {},
-    "overrides": [
-        {
-            "files": ["./src/util/ws_mousetrap_fork.js"],
```
epicfaace, line 6: Is it OK that we are no longer overriding ws_mousetrap_fork.js? It's external code and we don't want it to be subject to linting.

At the same time, it looks like this PR is passing, so perhaps ws_mousetrap_fork.js is ignored anyway...?

epicfaace (approved):

PR #3539, opened 2021-05-23
Continuously build default-cpu and default-gpu dockerfiles

### LINKED ISSUE #2753: Continuously build default-cpu and default-gpu dockerfiles
It looks like these dockerfiles haven't been updated in a year—it'd be nice if they could be built whenever a change is made.

jzwang43:

Add a GHA workflow that runs on master whenever the default-cpu/gpu dockerfiles are changed. This should build and push the images to docker hub.

### Related issues

fixes #2753 

### Checklist

### DIFF
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -1,11 +1,11 @@
 name: Build default worker images
 on:
   push:
-    branches: [ master ]
-      paths: ['docker_config/dockerfiles/default-*']
+    branches: [master]
+    paths: ["docker_config/dockerfiles/default-*"]
   release:
     types: [published]
-    paths: ['docker_config/dockerfiles/default-*']
+    paths: ["docker_config/dockerfiles/default-*"]
 
 jobs:
   build_default:
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -5,6 +5,7 @@ on:
       paths: ['docker_config/dockerfiles/default-*']
   release:
     types: [published]
+    paths: ['docker_config/dockerfiles/default-*']
 
 jobs:
   build_default:
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -5,12 +5,12 @@ name: Build default worker images
 on:
   push:
     branches: [master]
-    paths: ["docker_config/dockerfiles/default-*"]
+    paths: ["docker_config/dockerfiles/Dockerfile.default-*"]
   release:
     types: [published]
-    paths: ["docker_config/dockerfiles/default-*"]
+    paths: ["docker_config/dockerfiles/Dockerfile.default-*"]
   pull_request:
-    paths: ["docker_config/dockerfiles/default-*"]
+    paths: ["docker_config/dockerfiles/Dockerfile.default-*"]
 
 jobs:
   build_default:
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -1,4 +1,7 @@
 name: Build default worker images
+# Default images are built and pushed when pushed to master, release, and pull request, 
+# and only when the corresponding dockerfiles are changed.
+# This workflow ensures the worker images are up-to-date on docker hub.
 on:
   push:
     branches: [master]
@@ -32,7 +35,8 @@ jobs:
         env:
           CODALAB_DOCKER_USERNAME: ${{ secrets.CODALAB_DOCKER_USERNAME }}
           CODALAB_DOCKER_PASSWORD: ${{ secrets.CODALAB_DOCKER_PASSWORD }}
-          # Gives us the branch name of the PR if on a pull_request-triggered build,
-          # otherwise, "master" if on a push-triggered build
+          # Set tag to latest if triggered by release,
+          # use the branch name of the PR if triggered by pull request,
+          # "master" if on a push-triggered build
           VERSION: ${{ (github.event_name == 'release' && 'latest') || github.head_ref || 'master' }}
           SERVICE: ${{ matrix.service }}
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -34,5 +34,5 @@ jobs:
           CODALAB_DOCKER_PASSWORD: ${{ secrets.CODALAB_DOCKER_PASSWORD }}
           # Gives us the branch name of the PR if on a pull_request-triggered build,
           # otherwise, "master" if on a push-triggered build
-          VERSION: ${{ github.head_ref || 'master' }}
+          VERSION: ${{ (github.event_name == 'release' && 'latest') || github.head_ref || 'master' }}
           SERVICE: ${{ matrix.service }}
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
+name: Push Default
+on:
+  # push:
+  #   branches: [ master ]
+  push:
```
jzwang43: Don't want to accidentally trigger a push so the it's commented out, will change back to the correct command before merging.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
...
+    branches: [ auto-build ]
+    # paths: ['docker_config/dockerfiles/default-*']
+
+jobs:
+  build_default:
+    name: Build Default Images
```
epicfaace:
```suggestion
    name: Build and push default worker images
```

.github/workflows/docker.yml
epicfaace: Not sure what you meant -- this workflow is already being triggered in this PR because of the auto-build condition you added. It's ok to trigger a push because it's being built with the "auto_build" tag rather than the "latest" tag (which is what is used by default).

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
+name: Push Default
```
epicfaace: Rename this to build_worker_images.yml

epicfaace:
```suggestion
name: Build default worker images
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  # push:
+  #   branches: [ master ]
+  push:
+    branches: [ auto-build ]
+    # paths: ['docker_config/dockerfiles/default-*']
```
epicfaace: We should also trigger this on a release (see release.yml), in which case it should build and push these images with tag "latest".

epicfaace:
It looks like it only takes 40 minutes to build both images in parallel. Because of that, here's a proposal about when we can run the workflow:
- Run the workflow on PRs, but only whenever the dockerfiles are changed. The image should be pushed with a tag equal to the branch name so that future builds can be cached from that branch.
- Run the workflow on master, but only whenever the dockerfiles are changed. The image should be pushed with a tag equal to "master".
- Run the workflow on release, but only whenever the dockerfiles are changed. The image should be pushed with a tag equal to "latest".

What do you think @jzwang43 ? (also cc'ing @percyliang on this)

jzwang43:
@epicfaace I think we can run on all of the occasions you mentioned since the build time is not too long. Right now I have it run when pushed to master and release. Not sure if running on PRs are needed since if they can build locally then it should have no issues building with git action.

.github/workflows/docker.yml
jzwang43: I have updated to trigger this on master and release.

.github/workflows/docker.yml
jzwang43: Updated

.github/workflows/docker.yml
jzwang43: I updated the code so the workflow is run on PRs, push to masters, and release.

.github/workflows/build_worker_images.yml
```diff
@@ -0,0 +1,38 @@
...
+          restore-keys: |
+            pip-
+      - run: pip install -r requirements.txt
+      - run: python3 codalab_service.py build ${SERVICE} $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
+        env:
+          CODALAB_DOCKER_USERNAME: ${{ secrets.CODALAB_DOCKER_USERNAME }}
+          CODALAB_DOCKER_PASSWORD: ${{ secrets.CODALAB_DOCKER_PASSWORD }}
+          # Gives us the branch name of the PR if on a pull_request-triggered build,
+          # otherwise, "master" if on a push-triggered build
+          VERSION: ${{ github.head_ref || 'master' }}
```
epicfaace: We need to ensure we run an additional build step in which we set VERSION to "latest" when this build is triggered by a release.

.github/workflows/build_worker_images.yml
```diff
@@ -0,0 +1,38 @@
...
+          path: ~/.cache/pip
+          key: pip-${{ hashFiles('requirements.txt') }}
+          restore-keys: |
+            pip-
+      - run: pip install -r requirements.txt
+      - run: python3 codalab_service.py build ${SERVICE} $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
+        env:
+          CODALAB_DOCKER_USERNAME: ${{ secrets.CODALAB_DOCKER_USERNAME }}
+          CODALAB_DOCKER_PASSWORD: ${{ secrets.CODALAB_DOCKER_PASSWORD }}
+          # Gives us the branch name of the PR if on a pull_request-triggered build,
```
epicfaace: Can you update this comment to reflect what VERSION means now (including "latest")?

.github/workflows/build_worker_images.yml
```diff
@@ -0,0 +1,38 @@
+name: Build default worker images
+on:
```
epicfaace, line 5: Can you add a comment here explaining when the default worker images are built, and why this decision was made?

docker_config/dockerfiles/Dockerfile.default-cpu
```diff
@@ -64,9 +67,9 @@ ENV SCALA_HOME /usr/share/java
 
 ## Python packages
 
-RUN pip2 install -U pip
+RUN python -m pip install --upgrade "pip < 21.0"
```
epicfaace, line 68: Why this change?

jzwang43: Newer pip dropped the support for python 2, the installation will fail.

.github/workflows/build_worker_images.yml
jzwang43: Comments added, thanks for pointing out.

docker_config/dockerfiles/Dockerfile.default-cpu
epicfaace: Let's add a comment so it's clear about that

docker_config/dockerfiles/Dockerfile.default-cpu
```diff
@@ -1,7 +1,10 @@
 FROM ubuntu:xenial
 MAINTAINER CodaLab Team "codalab.worksheets@gmail.com"
 
+RUN apt-get update && apt-get install -y \
+    software-properties-common
```
epicfaace, line 5: Why is this change needed? Also, why did we remove python-software-properties in default-gpu but not in this file?

.github/workflows/build_worker_images.yml
```diff
@@ -0,0 +1,42 @@
+name: Build default worker images
+# Default images are built and pushed when pushed to master, release, and pull request, 
+# and only when the corresponding dockerfiles are changed.
+# This workflow ensures the worker images are up-to-date on docker hub.
```
epicfaace:
```suggestion
# This workflow ensures the worker images are up-to-date on Docker Hub.
```

docker_config/dockerfiles/Dockerfile.default-cpu
jzwang43: The step after this `add-apt-repository` will fail if we don't install software-properties-common first.

jzwang43: Nevermind I think apt install cannot find python-software-properties. software-properties-common package is an alternate one for python-software-properties and should work fine.

epicfaace (approved):

PR #3553, opened 2021-05-26
Ensure CPU-only bundles don't have access to GPUs in slurm jobs

nelson-liu:
Right now, if you launch a job that requests only CPUs on a slurm node with GPUs available, it'll parse the GPUSET as "ALL", giving the job access to all of the GPUs. This explicitly sets the GPUSET to `NONE`, a new sentinel value for an empty gpuset (there are some issues with shell quotes and passing down `''` to the slurm worker).


nelson-liu:
hmm, bug is still there...will keep looking around

nelson-liu:
ok, now it's ready for a look.

epicfaace:
Interesting, how did you notice this issue in the first place?

nelson-liu:
Usually i run cpu-only jobs on nodes that don't have GPUs, but some GPU nodes were recently moved to the queue that was formerly cpu-only. when the bundles were running on those nodes, I was noticing that they were inadvertently using the GPU.

teetone (approved):

PR #3557, opened 2021-05-27
Optimize size of default-cpu / gpu dockerfiles, update to latest ubuntu and cuda, remove python 2

### LINKED ISSUE #3526: Update and make default-cpu / gpu dockerfiles smaller
Update (to latest ubuntu) and make default-cpu / gpu dockerfiles smaller

jzwang43:
Update (to latest ubuntu) and make default-cpu / gpu dockerfiles smaller

### Related issues

fixes #3526 

### Checklist

### DIFF
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -38,7 +38,7 @@ jobs:
           sudo apt clean
           docker rmi $(docker image ls -aq)
           df -h
-      - run: python3 codalab_service.py build ${SERVICE} $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
+      - run: python3 codalab_service.py build --version ${VERSION} ${SERVICE} $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
         env:
           CODALAB_DOCKER_USERNAME: ${{ secrets.CODALAB_DOCKER_USERNAME }}
           CODALAB_DOCKER_PASSWORD: ${{ secrets.CODALAB_DOCKER_PASSWORD }}
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -31,6 +31,12 @@ jobs:
           restore-keys: |
             pip-
       - run: pip install -r requirements.txt
+      - run: | 
+          sudo swapoff -a
+          sudo rm -f /swapfile
+          sudo apt clean
+          docker rmi $(docker image ls -aq)
+          df -h
       - run: python3 codalab_service.py build ${SERVICE} $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
         env:
           CODALAB_DOCKER_USERNAME: ${{ secrets.CODALAB_DOCKER_USERNAME }}
--- a/.github/workflows/build_worker_images.yml
+++ b/.github/workflows/build_worker_images.yml
@@ -34,6 +34,7 @@ jobs:
       - run: | 
           sudo swapoff -a
           sudo rm -f /swapfile
+          sudo rm -rf /opt/ghc
           sudo apt clean
           docker rmi $(docker image ls -aq)
           df -h
--- a/docker_config/dockerfiles/Dockerfile.default-cpu
+++ b/docker_config/dockerfiles/Dockerfile.default-cpu
@@ -6,6 +6,7 @@ RUN apt-get update && apt-get install -y \
     software-properties-common
 RUN add-apt-repository ppa:deadsnakes/ppa 
 
+
 ############################################################
 # Common steps (must be the same in the CPU and GPU images)
 
@@ -70,5 +71,5 @@ RUN python3 -m pip install -U --no-cache-dir \
       torch==1.1.0 \
       torchvision
 
-RUN python3 -m spacy download en
+RUN python3 -m spacy download en_core_web_sm
 
--- a/docker_config/dockerfiles/Dockerfile.default-cpu
+++ b/docker_config/dockerfiles/Dockerfile.default-cpu
@@ -1,10 +1,10 @@
 FROM ubuntu:20.04
 MAINTAINER CodaLab Team "codalab.worksheets@gmail.com"
 
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y \
     software-properties-common
 RUN add-apt-repository ppa:deadsnakes/ppa 
-ENV DEBIAN_FRONTEND=noninteractive
 
 ############################################################
 # Common steps (must be the same in the CPU and GPU images)
--- a/docker_config/dockerfiles/Dockerfile.default-cpu
+++ b/docker_config/dockerfiles/Dockerfile.default-cpu
@@ -1,7 +1,10 @@
-FROM ubuntu:xenial
+FROM ubuntu:20.04
 MAINTAINER CodaLab Team "codalab.worksheets@gmail.com"
 
+RUN apt-get update && apt-get install -y \
+    software-properties-common
 RUN add-apt-repository ppa:deadsnakes/ppa 
+ENV DEBIAN_FRONTEND=noninteractive
 
 ############################################################
 # Common steps (must be the same in the CPU and GPU images)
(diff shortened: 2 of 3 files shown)


epicfaace:
How small are they now?

jzwang43:
> 
> 
> How small are they now?

I think docker slim reduces the size based on what is used and not used during execution. But we are providing a general image which does not have a specific behavior since user can do anything with it. So I am not sure docker-slim can help much. I checked the images and looks like they are already removing the apt files.

epicfaace:
Okay, sounds good. From my earlier comment "How small are they now?" I meant, how large are the images now (in bytes?) I know they used to be 5 GB, so I'd like to know how much space we saved in this PR.

jzwang43:
> 
> 
> Okay, sounds good. From my earlier comment "How small are they now?" I meant, how large are the images now (in bytes?) I know they used to be 5 GB, so I'd like to know how much space we saved in this PR.

I am not sure why but locally if you do `docker images`, the reported size for the original image is 10+ GB. After my changes it seems to be 1GB smaller. I think on docker hub reports around 5GB for the size and it might be related to this issue: https://github.com/docker/hub-feedback/issues/331

Just as an experiment, I tried removing python2, the docker image size is halved.

docker_config/dockerfiles/Dockerfile.default-gpu
```diff
@@ -1,25 +1,30 @@
...
     build-essential \
     cmake \
     libhdf5-dev \
     swig \
-    wget \
-    curl
+    wget
+
+RUN curl https://bootstrap.pypa.io/pip/2.7/get-pip.py --output get-pip.py
+RUN python2 get-pip.py
```
epicfaace: We should be able to remove python 2 now.

docker_config/dockerfiles/Dockerfile.default-cpu
```diff
@@ -13,21 +13,15 @@ RUN apt-get update && apt-get install -y \
...
     swig \
-    wget \
-    curl
+    wget
 
 ## Python 3.6
-RUN add-apt-repository ppa:deadsnakes/ppa && \
-    apt-get update -y  && \
-    apt-get install python3.6 -y \
+RUN apt-get install python3.6 -y \
```
epicfaace: Let's combine this with the previous apt-get command

docker_config/dockerfiles/Dockerfile.default-cpu
```diff
@@ -13,21 +13,15 @@ RUN apt-get update && apt-get install -y \
...
         python3.6-venv \
         python3.6-dev \
         python3-pip \
```
epicfaace: At the end, we should also add `&& rm -rf /var/lib/apt/lists/*`, which will save more space -- see https://stackoverflow.com/questions/61990329/dockerfile-benefits-of-repeated-apt-cache-cleans. let's also add this to the end of all apt-get install commands.

docker_config/dockerfiles/Dockerfile.default-cpu
```diff
@@ -60,26 +54,11 @@ RUN wget http://scala-lang.org/files/archive/scala-$SCALA_VERSION.deb && \
     apt-get clean && \
     apt-get autoremove && \
```
epicfaace, line 48: We should combine all apt-get install commands into one single command if possible. I believe this should save a bunch of space (as we'd store far less layers).

docker_config/dockerfiles/Dockerfile.default-gpu
```diff
@@ -94,22 +78,13 @@ RUN python3 -m pip install -U \
...
 
-RUN pip2 install spacy
-RUN python -m spacy download en
 RUN python3 -m spacy download en
 
-RUN pip2 install -U \
-      tensorflow-gpu==1.12.0 \
-      tensorboard \
-      keras
 RUN python3 -m pip install -U \
```
epicfaace: For both dockerfiles: we should combine all pip install commands into a single command, and also use the `--no-cache-dir` option to save space. See https://stackoverflow.com/questions/45594707/what-is-pips-no-cache-dir-good-for (we can probably keep the pip install command at the end with apex as-is, though)

jzwang43:
@epicfaace looks like the newer nvidia/cuda image is much larger than the older version we used before. The old base image was under 1GB ([link](https://hub.docker.com/layers/nvidia/cuda/9.0-devel-ubuntu16.04/images/sha256-87e7cdd5ff3e90ae8d9f28f8ea52cc6a4517aed312b40f00aa6a51578c42ff35?context=explore)) and if we upgrade the newer image is 2.61GB ([link](https://hub.docker.com/layers/nvidia/cuda/11.1.1-devel-ubuntu20.04/images/sha256-88212522722ca13de87ea88a441be9f99c8dcec9ab64f01e12b76d4c1604d226?context=explore)), this will increase our default-gpu image size by a lot. What should we do in this case?

epicfaace:
As per our discussion, let's try the following:
- clearing out disk space (https://github.com/codalab/codalab-worksheets/blob/8857244823146c0fe0169fabdc29c4d698d84667/.github/workflows/test.yml#L103-L106 and https://github.com/flashlight/wav2letter/actions/runs/74797824/workflow#L14-L18)
- multistage build for nvidia apex (see https://github.com/cisagov/crossfeed/blob/master/backend/Dockerfile.worker for an example)

.github/workflows/build_worker_images.yml
```diff
@@ -31,6 +31,12 @@ jobs:
           restore-keys: |
             pip-
       - run: pip install -r requirements.txt
+      - run: | 
+          sudo swapoff -a
+          sudo rm -f /swapfile
+          sudo apt clean
+          docker rmi $(docker image ls -aq)
```
epicfaace, line 39: I think we should also add

`sudo rm -rf /opt/ghc`

(from https://github.com/codalab/codalab-worksheets/blob/8857244823146c0fe0169fabdc29c4d698d84667/.github/workflows/test.yml#L103-L106)

.github/workflows/build_worker_images.yml
```diff
@@ -31,6 +31,13 @@ jobs:
...
             pip-
       - run: pip install -r requirements.txt
+      - run: | 
+          sudo swapoff -a
+          sudo rm -f /swapfile
+          sudo rm -rf /opt/ghc
+          sudo apt clean
+          docker rmi $(docker image ls -aq)
+          df -h
       - run: python3 codalab_service.py build ${SERVICE} $([ -z "${CODALAB_DOCKER_USERNAME}" ] || echo "--push")
```
epicfaace: Right now, it's pushing to a tag called HEAD, rather than the actual PR branch name. We should change this

image

epicfaace:
Image size changes with this PR:
default-cpu (https://hub.docker.com/r/codalab/default-cpu/tags?page=1&ordering=last_updated): 5.1 GB -> 1.63 GB :tada: 
default-gpu (https://hub.docker.com/r/codalab/default-gpu/tags?page=1&ordering=last_updated): 6.18 GB -> 7 GB

It might still be worth it to consider using multistage builds for default-gpu to save more space -- what do you think?

jzwang43:
> 
> 
> Image size changes with this PR:
> default-cpu (https://hub.docker.com/r/codalab/default-cpu/tags?page=1&ordering=last_updated): 5.1 GB -> 1.63 GB 🎉
> default-gpu (https://hub.docker.com/r/codalab/default-gpu/tags?page=1&ordering=last_updated): 6.18 GB -> 7 GB
> 
> It might still be worth it to consider using multistage builds for default-gpu to save more space -- what do you think?

I think it might be fine for now, we can have a separate PR to do multistage.

.github/workflows/build_worker_images.yml
jzwang43: Update the github action file, should be working now: https://hub.docker.com/layers/codalab/default-cpu/update_images/images/sha256-ee7b6df243b8e51db27fccc8f1f125dee21bddff29fa21da1377b11b75c4acd4?context=explore

epicfaace:
> Image size changes with this PR:
> default-cpu (https://hub.docker.com/r/codalab/default-cpu/tags?page=1&ordering=last_updated): 5.1 GB -> 1.63 GB 🎉
> default-gpu (https://hub.docker.com/r/codalab/default-gpu/tags?page=1&ordering=last_updated): 6.18 GB -> 7 GB
> 
> It might still be worth it to consider using multistage builds for default-gpu to save more space -- what do you think?

FYI @percyliang -- default-cpu size decreased because of optimizations / removing Python 2. default-gpu size increased because we upgraded ubuntu (which required us to upgrade cuda, which is much larger than the earlier version).

We could decrease the size of both images, especially default-gpu, by using docker multi-stage builds. However, we decided that this is beyond the scope of this PR and could do so in another PR.

docker_config/dockerfiles/Dockerfile.default-gpu
```diff
@@ -64,23 +51,10 @@ RUN wget http://scala-lang.org/files/archive/scala-$SCALA_VERSION.deb && \
 ENV SCALA_HOME /usr/share/java
 
 ## Python packages
-# Newer pip dropped the support for python 2 so we need to specify the version.
-RUN python -m pip install --upgrade "pip < 21.0"
+
+## Python packages
```
epicfaace: Remove this line (duplicated from above)

epicfaace (approved):

docker_config/dockerfiles/Dockerfile.default-gpu
jzwang43: Thanks, it has been removed.

PR #3558, opened 2021-05-27
print status of bundle disk quota check

adiprerepa:
as requested by @percyliang 

image

A disk quota check is not performed on the server or client for `--link` bundles. You only get a `UsageError: Out of disk quota` whenever the disk quota is *currently* invalid (`disk_use`>`disk_quota`) for all bundles including linked -- but I do not think this is really an issue.

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1375,7 +1375,7 @@ class BundleCLI(object):
                     'Attempted to upload bundle of size %d with only %d remaining in user\'s disk quota.'
                     % (total_bundle_size, disk_left)
                 )
-            print("Bundle of size %d meets the disk quota requirement of %d, continuing..." % (total_bundle_size, disk_left), file=self.stderr)
+            print("Bundle of size %s meets the disk quota requirement of %s, continuing..." % (FileTransferProgress.format_size(total_bundle_size), FileTransferProgress.format_size(disk_left)), file=self.stderr)
             print("Preparing upload archive...", file=self.stderr)
             if args.ignore:
                 print(
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1375,7 +1375,7 @@ class BundleCLI(object):
                     'Attempted to upload bundle of size %d with only %d remaining in user\'s disk quota.'
                     % (total_bundle_size, disk_left)
                 )
-
+            print("Bundle of size %d meets the disk quota requirement of %d, continuing..." % (total_bundle_size, disk_left), file=self.stderr)
             print("Preparing upload archive...", file=self.stderr)
             if args.ignore:
                 print(


codalab/lib/bundle_cli.py
```diff
@@ -1375,7 +1375,7 @@ def do_upload_command(self, args):
                     'Attempted to upload bundle of size %d with only %d remaining in user\'s disk quota.'
                     % (total_bundle_size, disk_left)
                 )
-
+            print("Bundle of size %d meets the disk quota requirement of %d, continuing..." % (total_bundle_size, disk_left), file=self.stderr)
```
percyliang: Could we print out a message earlier 'Checking files...{size} so far...\r' to have a bit more of a status message, and then make sure at the end of the day, only the uuid shows up in the stdout.

epicfaace:
> A disk quota check is not performed on the server or client for `--link` bundles. You only get a `UsageError: Out of disk quota` whenever the disk quota is _currently_ invalid (`disk_use`>`disk_quota`) for all bundles including linked -- but I do not think this is really an issue.

I'm a bit confused -- what does this PR have to do with linked bundles?

codalab/lib/bundle_cli.py
epicfaace: We have a format function that formats bytes in terms of megabytes / gigabytes. We should use that here.

adiprerepa:
> I'm a bit confused -- what does this PR have to do with linked bundles?

@epicfaace We thought a disk quota check would be performed with --link, I'm just saying that it does not. See the meeting notes for the actual issue report.

codalab/lib/bundle_cli.py
adiprerepa: sorry, why would we print `Checking files...{size}`? The user gets the information regardless, either they get the `DiskQuotaExceededError` or the print statement after. And everything goes to `stderr` here.

adiprerepa: done

epicfaace:
> See the meeting notes for the actual issue report.

Can you either describe the issue report in this PR or make a github issue and link it to this? The meeting notes are just a temporary place to take notes, ideally the source of truth for issues / etc. will be in GitHub issues / PR, and we should always transfer issues from meeting notes over to GitHub.

epicfaace:
@adiprerepa what's the status of this PR -- is it still relevant?

adiprerepa:
@epicfaace i think it's still relevant...maybe review?

epicfaace:
@adiprerepa I don't get what the point of this PR is -- why are we just showing this message?

percyliang:
What's the status of this PR?

epicfaace:
> A disk quota check is not performed on the server or client for `--link` bundles. You only get a `UsageError: Out of disk quota` whenever the disk quota is _currently_ invalid (`disk_use`>`disk_quota`) for all bundles including linked -- but I do not think this is really an issue.

@adiprerepa I think the ball is in your court for the PR here. I don't really understand your PR description above. Can you update it to say what issue exactly this is fixing?

PR #3560, opened 2021-05-27
Fix Scrolling Issue

### LINKED ISSUE #3556: Scrolling Issue on the Frontend
~~When you are viewing a worksheet in a window that's narrow, if you press shift-G to go to the end of the worksheet, it doesn't seem to scroll all the way down.~~
~~And sometimes it scrolls too far down.~~
extra scrollbar on the right in the tables in a worksheet
extra scrollbar on the page when the window is narrow

yuqijin:

remove extra scrolling bar

### Related issues

fixes #3556 

### Screenshots

image

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
+++ b/frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
@@ -46,7 +46,7 @@ export default ({
     toggleWorksheetSize,
     showBundleContent,
 }) => (
-    <Sticky top={6} innerZ={1059}>
+    <Sticky top='#codalab-app-bar' innerZ={1059}>
         <div className='worksheet_content' style={styles}>
             <div className='header-row'>
                 <Grid container direction='column'>
--- a/frontend/src/components/worksheets/items/TableItem/TableItem.js
+++ b/frontend/src/components/worksheets/items/TableItem/TableItem.js
@@ -262,7 +262,6 @@ class TableItem extends React.Component<{
         return (
             <div className='ws-item'>
                 <TableContainer style={{ overflowX: 'auto', overflowY: 'hidden' }}>
-                    {/*The 0.5px margin space is for the divider between table header and table body*/}
                     <Table className={tableClassName} style={{ marginBottom: '0.5px' }}>
                         <TableHead>
                             <TableRow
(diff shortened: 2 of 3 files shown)


epicfaace (approved):
Looks good. I think we should remove the "fixes https://github.com/codalab/codalab-worksheets/issues/3556" wording because this only fixes part of that issue, but not the Ctrl+G issue.

frontend/src/components/worksheets/items/TableItem/TableItem.js
```diff
@@ -261,8 +261,9 @@ class TableItem extends React.Component<{
         });
         return (
             <div className='ws-item'>
-                <TableContainer style={{ overflowX: 'auto' }}>
-                    <Table className={tableClassName} style={{ marginBottom: '0' }}>
+                <TableContainer style={{ overflowX: 'auto', overflowY: 'hidden' }}>
+                    {/*The 0.5px margin space is for the divider between table header and table body*/}
```
epicfaace: I don't think we need this comment here, it seems too trivial to need a comment

yuqijin:
> Looks good. I think we should remove the "fixes #3556" wording because this only fixes part of that issue, but not the Ctrl+G issue.

The remaining issue is the redundant scrolling bar when the window is narrow. The latest commit fix it.

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
```diff
@@ -46,7 +46,7 @@ export default ({
     toggleWorksheetSize,
     showBundleContent,
 }) => (
-    <Sticky top={6} innerZ={1059}>
+    <Sticky top='#codalab-app-bar' innerZ={1059}>
```
epicfaace, line 49: Can we actually factor out "codalab-app-bar" into a constant in constants.js? so that we don't actually refactor the header component and remove the id and make this break again

epicfaace (approved):

frontend/src/components/worksheets/Worksheet/WorksheetHeader.js
epicfaace: Let's do this later

PR #3569, opened 2021-05-27
singularity image manager

### LINKED ISSUE #3655: [singularity] abstract image pulling logic to be platform-agnostic
We should abstract `DockerImageManager` into an `ImageManager` and have a `SingularityImageManager` to be platform agnostic as to how we pull our images. 

We should also run a separate docker container (i.e. a codalab service) for singularity workers.

PR: https://github.com/codalab/codalab-worksheets/pull/3569

adiprerepa:
fixes https://github.com/codalab/codalab-worksheets/issues/3655

new pr because I accidentally uploaded a `.sif` file and then screwed up git.

design doc: https://docs.google.com/document/d/1QfH3xk84xByjrK19BwcbrZ87_F2LJe4MAbts5hx4nBE/edit

issue: https://github.com/codalab/codalab-worksheets/issues/3661

change description:
- `ImageManager` base class is created. This has a few key methods (`get`, `cleanup`) that are used by the worker to pull singularity images -- these do not have default implementations. It also has methods like `start` and `stop` which have default implementations.
- `DockerImageManager` starts to inherit from `ImageManager`, and implements it's own `get` and `cleanup` based on the docker api.
- `SingularityImageManager` is created and inherits from `ImageManager`. It implements its own `get` and `cleanup`, like the docker image manager.
  - An argument in the constructor is `image_folder`, which is the path to the folder where singularity images are saved after `get` is called. When `cleanup` is called, it deletes all the files in this folder.
 
In the future, when someone runs a bundle and requests an image, my hope is that before the Image Manager is called, the image spec is converged into some unique, platform-specific format. For example, for a docker image, the image spec would be `docker://ubuntu:latest`, while a sylabs container image would be `library://go.sif`. `--request-docker-image` would remain, of course, but there would also be a `--request-image` flag that would accept any image in the right schema. These two flags would converge to the same place.

singularity jobs can run (any) docker image. The dedicated "docker" worker would have access to the docker socket (therefore root), and is just for running jobs natively. I don't think we want to group the launching of native docker jobs and singularity jobs together as the perms required for a worker using just singularity jobs are lesser than the root perms required for the native docker worker. If we do this, I think it defeats the purpose of singularity.


codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,120 @@
...
+                    return ImageAvailabilityState(
+                        digest=None, stage=DependencyStage.FAILED, message=failure_msg
+                    )
+
+            self._downloading.add_if_new(image_spec, threading.Thread(target=download, args=[]))
+            return ImageAvailabilityState(
+                digest=None,
+                stage=DependencyStage.DOWNLOADING,
+                message=self._downloading[image_spec]['status'],
+            )
```
epicfaace, line 83: Can we add some tests for this perhaps?

epicfaace:
@adiprerepa is this PR still in progress, or is it done? What stage exactly is it within the general task of adding singularity support?

adiprerepa:
> @adiprerepa is this PR still in progress, or is it done? What stage exactly is it within the general task of adding singularity support?

@epicfaace this PR is to introduce some of the key components of singularity. I figured it would be easier to introduce it in more granular stages as opposed to all in one mega PR, kind of like how you handled blob storage.

epicfaace:
> this PR is to introduce some of the key components of singularity.

Okay, if you could outline in the PR description what components this PR introduces, that will help with the review. Also, let me know once it's ready for review.

adiprerepa:
@epicfaace should be ready for review. I tested it and it works, let me know if I can make any clarifications.

codalab/worker/singularity_image_manager.py
adiprerepa: will do

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,120 @@
+import logging
+import os
+import threading
+
+from codalab.lib.formatting import size_str
+from codalab.worker.image_manager import ImageAvailabilityState
+from codalab.worker.fsm import DependencyStage
+from codalab.worker.image_manager import ImageManager
+from codalab.worker.singularity_utils import get_singularity_container_size
+from spython.main import Client  # type: ignore
```
epicfaace: Add spython to https://github.com/codalab/codalab-worksheets/blob/master/setup.cfg instead of using "type: ignore"

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
+import logging
+import threading
+import time
+import traceback
+from collections import namedtuple
+
+from codalab.worker.worker_thread import ThreadDict
+
+logger = logging.getLogger(__name__)
```
epicfaace, line 11: Add space below this line

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
...
+# Stores the download state of a Docker image (also includes the digest being pulled, digest string,
+# DependencyStage and relevant status message from the download)
+ImageAvailabilityState = namedtuple('ImageAvailabilityState', ['digest', 'stage', 'message'])
+
+
+class ImageManager:
+    """
+    ImageManager is the base interface for all image managers (docker, singularity, etc.)
+    No actual logic is encapsulated in this class.
```
epicfaace: Instead of this, just say it's an abstract class. Explain what an "image manager" is. Maybe also say which methods subclasses need to implement.

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
...
+        logger.debug("Stopping image manager: stop the downloads threads")
+        self._downloading.stop()
+        if self._cleanup_thread:
+            logger.debug("Stopping image manager: stop the cleanup thread")
+            self._cleanup_thread.join()
+        logger.info("Stopped image manager")
+        pass
+
+    def get(self, image_spec: str) -> ImageAvailabilityState:
+        raise NotImplementedError("get should not be called in the ImageManager base class")
```
epicfaace:
```suggestion
        raise NotImplementedError
```

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # should be implemented in subclasses
+    def _cleanup(self):
+        raise NotImplementedError("cleanup should not be called in the ImageManager base class")
```
epicfaace:
```suggestion
        raise NotImplementedError
```

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
...
+        if self._cleanup_thread:
+            logger.debug("Stopping image manager: stop the cleanup thread")
+            self._cleanup_thread.join()
+        logger.info("Stopped image manager")
+        pass
+
+    def get(self, image_spec: str) -> ImageAvailabilityState:
+        raise NotImplementedError("get should not be called in the ImageManager base class")
+
+    # should be implemented in subclasses
```
epicfaace: Remove this and replace this with a docstring explaining what _cleanup should do

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
...
+        self._stop = True
+        logger.debug("Stopping image manager: stop the downloads threads")
+        self._downloading.stop()
+        if self._cleanup_thread:
+            logger.debug("Stopping image manager: stop the cleanup thread")
+            self._cleanup_thread.join()
+        logger.info("Stopped image manager")
+        pass
+
+    def get(self, image_spec: str) -> ImageAvailabilityState:
```
epicfaace, line 80: Add a docstring here

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,63 @@
...
+ImageAvailabilityState = namedtuple('ImageAvailabilityState', ['digest', 'stage', 'message'])
+
+
+class ImageManager:
+    """
+    ImageManager is the base interface for all image managers (docker, singularity, etc.)
+    No actual logic is encapsulated in this class.
+    """
+
+    def __init__(self, max_image_size: int, max_image_cache_size: int):
```
epicfaace, line 27: Add docstrings for all methods -- see https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html for format

codalab/worker/docker_image_manager.py
```diff
@@ -14,20 +14,17 @@
...
-# DependencyStage and relevant status message from the download)
-ImageAvailabilityState = namedtuple('ImageAvailabilityState', ['digest', 'stage', 'message'])
 # Stores information relevant about caching about docker images
 ImageCacheEntry = namedtuple(
     'ImageCacheEntry', ['id', 'digest', 'last_used', 'virtual_size', 'marginal_size']
 )
 
 
-class DockerImageManager:
+class DockerImageManager(ImageManager):
```
epicfaace, line 26: in __init__, you should call `super().__init__(commit_file, max_image_cache_size, max_image_size)`

codalab/worker/main.py
```diff
@@ -116,6 +117,11 @@ def parse_args():
         action='store_true',
         help='If specified the worker quits if it finds itself with no jobs after a checkin',
     )
+    parser.add_argument(
+        '--singularity',
```
epicfaace: Maybe make an argument called `--container_runtime` which defaults to "docker" and can be set to "singularity".

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class SingularityImageManager(ImageManager):
+    def __init__(self, max_image_size: int, max_image_cache_size: int, image_folder: str):
+        """
+        image_folder needs to be an absolute path
+        """
+        super().__init__(max_image_size, max_image_cache_size)
+        if not os.path.isdir(image_folder):
+            raise ValueError("image_folder %s is not an absolute path" % image_folder)
```
epicfaace: This error is not accurate. Maybe say it's not a directory.

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,120 @@
...
+from codalab.worker.singularity_utils import get_singularity_container_size
+from spython.main import Client  # type: ignore
+
+logger = logging.getLogger(__name__)
+
+
+class SingularityImageManager(ImageManager):
+    def __init__(self, max_image_size: int, max_image_cache_size: int, image_folder: str):
+        """
+        image_folder needs to be an absolute path
```
epicfaace: Make the docstrings match the convention (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

codalab/worker/main.py
```diff
@@ -245,20 +251,36 @@ def main():
...
+    else:
+        singularity_folder = None
     # Set up local directories
     if not os.path.exists(args.work_dir):
         logging.debug('Work dir %s doesn\'t exist, creating.', args.work_dir)
         os.makedirs(args.work_dir, 0o770)
     if local_bundles_dir and not os.path.exists(local_bundles_dir):
         logger.info('%s doesn\'t exist, creating.', local_bundles_dir)
         os.makedirs(local_bundles_dir, 0o770)
+    if singularity_folder and not os.path.exists(singularity_folder):
```
epicfaace: Combine this chunk with the above chunk about singularity_folder

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def get(self, image_spec):
+        """
+        This will for now be built without caching - that's a little more complicated.
+        image_spec is in singularity format:
+        - Docker: 'docker://<image>'
+        - Sylabs Cloud hub: 'library://<image>'
+        - Singularity Hub (deprecated): 'shub://<image>'
+        """
+        if image_spec in self._downloading:
```
epicfaace: Let's move more of this logic to the base class ImageManager.

codalab/worker/singularity_utils.py
```diff
@@ -0,0 +1,16 @@
+def is_shub(img):
+    return img.startswith("shub://")
+
+
+def is_sylabs(img):
+    return img.startswith("library://")
+
+
+def is_singularity(image_spec):
```
epicfaace: This function isn't used

codalab/worker/singularity_utils.py
```diff
@@ -0,0 +1,16 @@
...
+    # sylabs container registry or (deprecated) singlarity hub
+    return is_shub(image_spec) or is_sylabs(image_spec)
+
+
+def get_singularity_container_size(image_spec):
```
epicfaace: This is used, but not implemented

codalab/worker/worker_run_state.py
```diff
@@ -179,7 +179,7 @@ def __init__(
         self.add_terminal(RunStage.RESTAGED)
 
         self.dependency_manager = dependency_manager
-        self.docker_image_manager = docker_image_manager
+        self.docker_image_manager = image_manager
```
```suggestion
        self.image_manager = image_manager
```

codalab_service.py
```diff
@@ -75,7 +77,9 @@ def should_run_service(args, service):
     # `default no-worker` is generally used for real deployment since we don't want a worker running on the same machine.
     services = [] if args.services is None else args.services
     if 'default' in args.services:
+        print("IM RUNNING DEFAULT")
```
epicfaace: rm these print statements

docker_config/compose_files/docker-compose.yml
```diff
@@ -367,6 +367,30 @@ services:
...
+      --work-dir ${CODALAB_WORKER_DIR}
+      --network-prefix ${CODALAB_WORKER_NETWORK_PREFIX}
+      --id ${HOSTNAME}
+      --singularity
+      --verbose
+    <<: *codalab-base
+    depends_on:
+      - rest-server
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
epicfaace: rm this

requirements.txt
```diff
@@ -26,3 +26,4 @@ markdown2==2.3.10
 wheel==0.35.1
 urllib3==1.26.4
 retry==0.9.2
+spython==0.1.13
```
epicfaace:
```suggestion
spython==0.1.14
```

codalab/worker/singularity_utils.py
adiprerepa: I am planning on using it in the next PR for singularity support. That function will determine whether or not to run in a singularity runtime. I would like to keep it for now, and possibly remove it later.

codalab/worker/singularity_utils.py
adiprerepa: If ever shub or sylabs will start supporting querying image sizes (which they said is one of their goals), this function will be useful. I think we should keep it here, as it will be also useful to unify the logic in the base class.

epicfaace: Not sure if there's value in just keeping it here and leaving it unimplemented. If you really want to keep it, maybe raise a NotImplementedError and add a docstring explaining why?

codalab/worker/singularity_utils.py
epicfaace: I think it's better to remove it and then re-add it in the future PR. I'm not able to give meaningful feedback for this function since I don't know how it's going to be used. If we want to keep it, though, we should add a docstring and explain how it can be used.

epicfaace:
Update design doc

codalab/worker/docker_image_manager.py
```diff
@@ -13,21 +12,20 @@
...
 logger = logging.getLogger(__name__)
 
-# Stores the download state of a Docker image (also includes the digest being pulled, digest string,
-# DependencyStage and relevant status message from the download)
-ImageAvailabilityState = namedtuple('ImageAvailabilityState', ['digest', 'stage', 'message'])
 # Stores information relevant about caching about docker images
 ImageCacheEntry = namedtuple(
     'ImageCacheEntry', ['id', 'digest', 'last_used', 'virtual_size', 'marginal_size']
 )
+URI_PREFIX = 'https://hub.docker.com/v2/repositories/'
```
epicfaace: import this from docker_utils instead

codalab/worker/docker_image_manager.py
```diff
@@ -38,44 +36,16 @@ def __init__(self, commit_file, max_image_cache_size, max_image_size):
...
+        super().__init__(max_image_size, max_image_cache_size)
         self._state_committer = JsonStateCommitter(commit_file)  # type: JsonStateCommitter
         self._docker = docker.from_env(timeout=DEFAULT_DOCKER_TIMEOUT)  # type: DockerClient
         self._downloading = ThreadDict(
             fields={'success': False, 'status': 'Download starting'}, lock=True
         )
-        self._max_image_cache_size = max_image_cache_size
-        self._max_image_size = max_image_size
-
         self._stop = False
```
epicfaace: Remove this line because it's already in imagemanager

codalab/worker/docker_image_manager.py
```diff
@@ -38,44 +36,16 @@ def __init__(self, commit_file, max_image_cache_size, max_image_size):
...
         self._sleep_secs = 10
```
epicfaace: Remove this line because it's already in imagemanager

codalab/worker/docker_image_manager.py
```diff
@@ -38,44 +36,16 @@ def __init__(self, commit_file, max_image_cache_size, max_image_size):
...
         self._cleanup_thread = None
```
epicfaace: Move this initialization of _cleanup_thread from here to imagemanager

codalab/worker/docker_image_manager.py
```diff
@@ -156,156 +126,157 @@ def get(self, image_spec):
...
+             When this is the case, different images API methods act differently:
+             - pull pulls all tags of the image
+             - get tries to get `latest` by default
+             That means if someone requests a docker image without a tag, and the image does not have a latest
+             tag pushed to Dockerhub, pull will succeed since it will pull all other tags, but later get calls
+             will fail since the `latest` tag won't be found on the system.
+             We don't want to assume what tag the user wanted so we want the pull step to fail if no tag is specified
+             and there's no latest tag on dockerhub.
+             Hence, we append the latest tag to the image spec if there's no tag specified otherwise at the very beginning
+            image_spec += ':latest'
```
epicfaace, line 124: Uncomment this

codalab/worker/docker_image_manager.py
```diff
@@ -156,156 +126,157 @@ def get(self, image_spec):
...
-        :returns: A DockerAvailabilityState object with the state of the docker image
+        :returns: An ImageAvailabilityState object with the state of the docker image
         """
 
-        def image_availability_state(image_spec, success_message, failure_message):
+        if ':' not in image_spec:
             """
-            Try to get the image specified by image_spec from host machine.
-            Return ImageAvailabilityState.
+             Both digests and repo:tag kind of specs include the : character. The only case without it is when
```
epicfaace: Change this comment to a `#`-style comment.

codalab/worker/docker_image_manager.py
```diff
@@ -156,156 +126,157 @@ def get(self, image_spec):
             * READY if the image was downloaded successfully.
```
epicfaace: Move the parts of this docstring about threads etc. to the parent class docstring.

codalab/worker/docker_image_manager.py
epicfaace: Maybe just move it to the docstring

codalab/worker/docker_image_manager.py
```diff
@@ -156,156 +126,157 @@ def get(self, image_spec):
...
-                        except docker.errors.NotFound as err:
-                            # It's possible that we get a 404 not found error here when removing the image,
-                            # since another worker on the same system has already done so. We just
-                            # ignore this 404, since any extraneous tags will be removed during the next iteration.
-                            logger.warning(
-                                "Attempted to remove image %s from cache, but image was not found: %s",
-                                tag,
-                                err,
-                            )
+        ImageManager.get(self, image_spec)
```
epicfaace:
```suggestion
        return super().get(image_spec)
```

codalab/worker/docker_image_manager.py
```diff
@@ -156,156 +126,157 @@ def get(self, image_spec):
...
-                )
-            except Exception as ex:
-                logger.error(ex)
-                return ImageAvailabilityState(
-                    digest=None, stage=DependencyStage.FAILED, message=failure_message % ex
-                )
+    def _download(self, image_spec) -> None:
+        """
+        Download the container image from DockerHub to the host machine.
+        This function will update the _downloading ThreadDict with the status and progress of the
```
epicfaace: Move this comment to the parent class

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,217 @@
...
+                            )
+                        self._downloading.remove(image_spec)
+                        return status
+            else:
+                if self._max_image_size:
+                    try:
+                        image_size_bytes = self._image_size_without_pulling(image_spec)
+                        if image_size_bytes is None:
+                            failure_msg = (
+                                "Unable to find Docker image: {} from Docker HTTP Rest API V2. "
```
epicfaace: make this more generic

codalab/worker/docker_image_manager.py
```diff
@@ -38,44 +36,16 @@ def __init__(self, commit_file, max_image_cache_size, max_image_size):
         :param max_image_cache_size: Total size in bytes that the image cache can use
         :param max_image_size: Total size in bytes that the image can have
         """
+        super().__init__(max_image_size, max_image_cache_size)
```
epicfaace, line 37: Update https://github.com/codalab/codalab-worksheets/blob/ea6f19297309eeebac508f00855f08637ea75ad9/codalab/worker/main.py#L93 to 1) make descriptions of args more generic and 2) specify that max image size does not work with singularity.

codalab/worker/main.py
```diff
@@ -116,6 +117,11 @@ def parse_args():
         action='store_true',
         help='If specified the worker quits if it finds itself with no jobs after a checkin',
     )
+    parser.add_argument(
+        '--container_runtime',
```
epicfaace:
```suggestion
        '--container-runtime',
```

codalab/worker/main.py
```diff
@@ -116,6 +117,11 @@ def parse_args():
...
+        default='docker',
```
epicfaace, line 127: Use `choices` to limit values: https://stackoverflow.com/a/15836714/1950269

codalab/worker/main.py
epicfaace: Test running this with docker / singularity to make sure it works.

codalab/worker/main.py
```diff
@@ -245,6 +251,25 @@ def main():
             args.work_dir,
             args.max_work_dir_size,
         )
+
+    if args.container_runtime == "singularity":
+        singularity_folder = os.path.join(args.work_dir, 'codalab_singularity_images')
+        if not os.path.exists(singularity_folder):
+            logger.info(
+                'codalab local singularity image location %s doesn\'t exist, creating.',
```
epicfaace:
```suggestion
                'Local singularity image location %s doesn\'t exist, creating.',
```

codalab/worker/main.py
```diff
@@ -245,6 +251,25 @@ def main():
...
+                singularity_folder,
+            )
+            os.makedirs(singularity_folder, 0o770)
+        image_manager = SingularityImageManager(
+            args.max_image_size, args.max_image_cache_size, singularity_folder,
+        )
+    else:
+        # assume anything else is docker -- no need to error
```
epicfaace:
```suggestion
```

codalab/worker/main.py
```diff
@@ -245,6 +251,25 @@ def main():
             args.work_dir,
             args.max_work_dir_size,
         )
+
+    if args.container_runtime == "singularity":
+        singularity_folder = os.path.join(args.work_dir, 'codalab_singularity_images')
```
epicfaace, line 268: Double check -- maybe check in deployment to see what work_dir is set to? to ensure it can store all these images?

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,98 @@
...
+
+        Args:
+            max_image_size: Maximum image size of any pulled singularity image.
+            max_image_cache_size: The maximum size of the image cache (folder images are pulled into)
+            image_folder: folder images are pulled into.
+        """
+        super().__init__(max_image_size, max_image_cache_size)
+        if not os.path.isdir(image_folder):
+            raise ValueError("image_folder %s is not a directory" % image_folder)
+        if not os.path.isabs(image_folder):
```
epicfaace: rm this

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,98 @@
...
+            # prefix docker, no scheme exists
+            image_spec = "docker://" + image_spec
+        ImageManager.get(image_spec)
+
+    def _download(self, image_spec):
+        logger.debug('Downloading image %s', image_spec)
+        try:
+            # stream=True for singularity doesnt return progress or anything really - for now no progress
+            self._downloading[image_spec]['message'] = "Starting Download"
+            img, puller = Client.pull(image_spec, pull_folder=self.image_folder, stream=True)
```
epicfaace: Remove stream=True

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,98 @@
...
+            failure_message: message to store in the ImageAvailabilityState if the image does not exist
+
+        Returns:
+
+        """
+        if len(image_spec.split(":")) <= 1:
+            # error, we should have a version
+            raise ValueError
+        img = image_spec.split(":")[0]
+        version = image_spec.split(":")[-1]
```
epicfaace: This is more Pythonic:
`img, version = image_spec.split(":")`

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        hypofile = os.path.join(self.image_folder, img + "_" + version)
+        if os.path.isfile(hypofile):
+            return ImageAvailabilityState(
+                digest=None, stage=DependencyStage.READY, message=success_message
```
epicfaace: See if we can get a digest for this -- see https://docs.docker.com/engine/reference/commandline/images/#list-image-digests and https://singularityhub.github.io/singularityhub-docs/docs/interact.

This is important for `docker_image` when we specify the latest tag.

image

codalab_service.py
```diff
@@ -32,6 +32,7 @@
     'rest-server',
     'bundle-manager',
     'worker',
+    # 'worker-singularity',
```
```suggestion
```

adiprerepa:
Updated the design doc: https://docs.google.com/document/d/1QfH3xk84xByjrK19BwcbrZ87_F2LJe4MAbts5hx4nBE/edit#

codalab/worker/singularity_image_manager.py
adiprerepa: Looking into it. We definitely can (`singularity cache list -v`) will list the image ids, but I need to double check if theres a better way with the singularity community

adiprerepa:
@epicfaace could you re-review for everything except the comments I responded to above? Want to make sure its fine now

codalab/worker/main.py
adiprerepa: will do

adiprerepa:
megaissue: https://github.com/codalab/codalab-worksheets/issues/3661

adiprerepa:
todo aditya: there are issues with references to docker and accessing the runtime freaking EVERYWHERE -- make sure to have it so whenever it fails (due to a docker socket not found error) fall back on singularity

codalab/worker/docker_image_manager.py
```diff
@@ -156,45 +119,9 @@ def get(self, image_spec):
             * READY if the image was downloaded successfully.
```
epicfaace: - rm this docstring
- rm all docstrings in DockerImageManager that are repeated verbatim in ImageManager

(this helps us to maintain docstrings more easily so they don't get out of sync)

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        image_spec is in singularity format:
+        - Docker: 'docker://<image>'
+        - Sylabs Cloud hub: 'library://<image>'
+        - Singularity Hub (deprecated): 'shub://<image>'
+        Images will be assumed to be docker if no prefix is included.
+        """
+        if len(image_spec.split("://")) <= 1:
+            # prefix docker, no scheme exists
+            image_spec = "docker://" + image_spec
```
epicfaace, line 42: append :latest to it if no version is specified

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        ImageManager.get(image_spec)
```
epicfaace:
```suggestion
        return ImageManager.get(image_spec)
```

codalab/worker/singularity_image_manager.py
epicfaace: do this in ImageManager.get since it's common across both platforms

codalab/worker/docker_image_manager.py
```diff
@@ -206,106 +133,138 @@ def image_availability_state(image_spec, success_message, failure_message):
...
             image_spec += ':latest'
+        return super().get(image_spec)
+
+    def _download(self, image_spec) -> None:
+        """
+        Download the container image from DockerHub to the host machine.
+        Args:
+            image_spec: docker image (just image, no prefix docker://)
+
+        Returns: None; all status is set in the _downloading ThreadDict
```
epicfaace:
```suggestion
        Returns: None
```

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def _download(self, image_spec):
+        logger.debug('Downloading image %s', image_spec)
+        try:
+            # stream=True for singularity doesnt return progress or anything really - for now no progress
+            self._downloading[image_spec]['message'] = "Starting Download"
```
epicfaace:
```suggestion
            self._downloading[image_spec]['message'] = "Starting download"
```

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+            img = Client.pull(image_spec, pull_folder=self.image_folder)
+            # maybe add img to a list and in cleanup rm the files in the list
```
epicfaace: rm comment

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,225 @@
...
+            image_spec: container-runtime specific spec
+            success_message: message to return in the ImageAvailabilityState if the image was successfully found
+            failure_message: message to return if there were any issues in getting the image.
+
+        Returns: ImageAvailabilityState
+
+        """
+        raise NotImplementedError
+
+    def _download(self, image_spec) -> None:
```
epicfaace: Change this so that
- subclasses override the `pull` function which just pulls the image
- when ImageManager creates a threaddict with `self._download`, it can instead wrap the `pull` function in the status- / threaddict- updating logic

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,225 @@
...
+                                success_message='Image {} can not be downloaded from the cloud '
+                                'but it is found locally'.format(image_spec),
+                                failure_message=self._downloading[image_spec]['message'] + ": %s",
+                            )
+                        self._downloading.remove(image_spec)
+                        return status
+            else:
+                if self._max_image_size:
+                    try:
+                        image_size_bytes = self._image_size_without_pulling(image_spec)
```
epicfaace: NotImplementedError nicer than returning None

```suggestion
                        try: 
                            image_size_bytes = self._image_size_without_pulling(image_spec)
                        except NotImplementedError:
                            ...
```

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        except Exception as ex:
+            logger.debug('Download for Singularity image %s failed: %s', image_spec, ex)
+            self._downloading[image_spec]['success'] = False
+            self._downloading[image_spec]['message'] = "Can't download image: {}".format(ex)
+
+    def _image_availability_state(
+        self, image_spec, success_message, failure_message
+    ) -> ImageAvailabilityState:
+        """
+        Returns the state of a said image on the codalab singularity image folder.
```
epicfaace:
```suggestion
        Returns the state of a specified image on the codalab singularity image folder.
```

codalab/worker/singularity_image_manager.py
epicfaace: Let's use the SHA256 hashes of the file (add a comment to this effect, maybe find a link)

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+            image_spec: image specification that has a version associated (format image:version)
+            success_message: message to store in the ImageAvailabilityState if the image exists
+            failure_message: message to store in the ImageAvailabilityState if the image does not exist
+
+        Returns:
+
+        """
+        if len(image_spec.split(":")) <= 1:
+            # error, we should have a version
+            raise ValueError
```
epicfaace:
```suggestion
            raise ValueError("Version must be specified")
```

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        Args:
+            image_spec: image specification that has a version associated (format image:version)
+            success_message: message to store in the ImageAvailabilityState if the image exists
+            failure_message: message to store in the ImageAvailabilityState if the image does not exist
+
+        Returns:
+
+        """
+        if len(image_spec.split(":")) <= 1:
+            # error, we should have a version
```
epicfaace:
```suggestion
```

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+            raise ValueError
+        img, version = image_spec.split(":")
+        hypofile = os.path.join(self.image_folder, img + "_" + version)
```
epicfaace: Add a comment mentioning that the image is located at a file specified by its name and version

epicfaace:
```suggestion
        image_path = os.path.join(self.image_folder, img + "_" + version)
```

epicfaace:
```suggestion
        hypofile = os.path.join(self.image_folder, image_spec.replace(":", "_"))
```

codalab/worker/singularity_image_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        Assumes image_spec has a version associated (in format image:version).
+        Args:
+            image_spec: image specification that has a version associated (format image:version)
+            success_message: message to store in the ImageAvailabilityState if the image exists
+            failure_message: message to store in the ImageAvailabilityState if the image does not exist
+
+        Returns:
+
+        """
+        if len(image_spec.split(":")) <= 1:
```
epicfaace: rm this check because we already ensure this upstream

codalab/worker/main.py
```diff
@@ -353,6 +380,8 @@ def parse_gpuset_args(arg):
         all_gpus = docker_utils.get_nvidia_devices()  # Dict[GPU index: GPU UUID]
     except docker_utils.DockerException:
         all_gpus = {}
+    except Exception:
```
epicfaace: Try to get a more specific exception handler. You can find the type of the exception by doing:

```python
except Exception as e:
    print("EXCEPTION", e, type(e))
```

codalab/worker/docker_utils.py
```diff
@@ -104,23 +104,40 @@ def get_available_runtime():
...
+        client.images.pull(cuda_image)
+        output = client.containers.run(
+            cuda_image, nvidia_command, runtime=NVIDIA_RUNTIME, detach=False, stdout=True, remove=True
+        )
+        output = output.decode()
+        # Get newline delimited gpu-index, gpu-uuid list
+        logger.info("GPUs: " + str(output.split('\n')[:-1]))
+        return {gpu.split(',')[0].strip(): gpu.split(',')[1].strip() for gpu in output.split('\n')[:-1]}
+    else:
+        # we don't really care where the image is because we are just using
```
epicfaace: Add a comment here saying that we're using the singularity runtime

codalab/worker/docker_utils.py
```diff
@@ -104,23 +104,40 @@ def get_available_runtime():
...
+        # it for nvidia-smi
+        img, puller = Client.pull('docker://' + cuda_image, pull_folder='/tmp', stream=True)
```
epicfaace: Don't stream, remove logs

codalab/worker/docker_utils.py
```diff
@@ -104,23 +104,40 @@ def get_available_runtime():
...
+        for l in puller:
+            logger.error(l)
+        logger.error(img)
+        output = Client.execute(img, nvidia_command, options=['--nv'])
+        if output['return_code'] != 0:
+            return {}
```
epicfaace: Throw an exception instead, for consistency

codalab/worker/docker_utils.py
```diff
@@ -104,23 +104,40 @@ def get_available_runtime():
...
-    logger.info("GPUs: " + str(output.split('\n')[:-1]))
-    return {gpu.split(',')[0].strip(): gpu.split(',')[1].strip() for gpu in output.split('\n')[:-1]}
+    if use_docker:
+        client.images.pull(cuda_image)
+        output = client.containers.run(
+            cuda_image, nvidia_command, runtime=NVIDIA_RUNTIME, detach=False, stdout=True, remove=True
+        )
+        output = output.decode()
+        # Get newline delimited gpu-index, gpu-uuid list
+        logger.info("GPUs: " + str(output.split('\n')[:-1]))
```
epicfaace: Pull these two lines (log and return dictionary) out of the if-else or a function (then we could unit test)

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,225 @@
...
+            )
+        except Exception as ex:
+            logger.error(ex)
+            return ImageAvailabilityState(
+                digest=None, stage=DependencyStage.FAILED, message=str(ex)
+            )
+
+    def _cleanup(self):
+        """
+        _cleanup should prune and clean up images in accordance with the image cache and
```
epicfaace:
```suggestion
        Prune and clean up images in accordance with the image cache and
```

codalab/worker/image_manager.py
```diff
@@ -0,0 +1,225 @@
...
+            how much space the images take up. The logic will vary per container runtime.
+            Therefore, this function should be implemented in the subclasses.
+        """
+        raise NotImplementedError
+
+    def _image_availability_state(
+        self, image_spec, success_message, failure_message
```
epicfaace: Add type annotations for all arguments

codalab/worker/main.py
```diff
@@ -92,7 +93,9 @@ def parse_args():
         help='Limit the size of Docker images to download from the Docker Hub'
         '(e.g. 3, 3k, 3m, 3g, 3t). If the limit is exceeded, '
         'the requested image will not be downloaded. '
-        'The bundle depends on this image will fail accordingly.',
+        'The bundle depends on this image will fail accordingly.'
+        'If running an image on the singularity runtime, there is no size'
```
epicfaace:
```suggestion
        'If running an image on the singularity runtime, there is no size '
```

codalab/worker/main.py
```diff
@@ -92,7 +93,9 @@ def parse_args():
         help='Limit the size of Docker images to download from the Docker Hub'
         '(e.g. 3, 3k, 3m, 3g, 3t). If the limit is exceeded, '
         'the requested image will not be downloaded. '
-        'The bundle depends on this image will fail accordingly.',
+        'The bundle depends on this image will fail accordingly.'
```
epicfaace:
```suggestion
        'The bundle depends on this image will fail accordingly. '
```

codalab/worker/worker.py
```diff
@@ -78,6 +78,8 @@ def __init__(
         delete_work_dir_on_exit=False,  # type: bool
         # A flag indicating if the worker will exit if it encounters an exception
         exit_on_exception=False,  # type: bool
+        # mainly to avoid trying to use the docker socket which we may not have permissions to.
+        runtime_docker=True,  # type: bool
```
epicfaace: Take this parameter out in this branch (you can add it in another branch if you need though)

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -8,11 +8,12 @@ WORKDIR /root/go/src/github.com/awslabs/amazon-ecr-credential-helper
 RUN make
 
 FROM ubuntu:16.04
+ARG default_user=codalab
 MAINTAINER CodaLab Worksheets <codalab.worksheets@gmail.com>
+RUN useradd -ms /bin/bash $default_user
```
epicfaace: Add a comment explaining

codalab/worker/singularity_image_manager.py
adiprerepa: i think its better to keep the `:latest` appendation where it is.

We shouldn't modify the `image_spec` in `image_manager.py` because theres no way to tell whether there is a version for singularity (the scheme (`docker://`) contains a colon)

epicfaace: > We shouldn't modify the `image_spec` in `image_manager.py` because theres no way to tell whether there is a version for singularity (the scheme (`docker://`) contains a colon)

There is, right? You could strip the scheme first and then check if there's a colon. Versions will not look like `://`.

codalab/worker/image_manager.py
adiprerepa: I still feel like maybe we should have `_download` be subclass-specific. Right now, the docker progress updating/streaming logic is pretty specific as is, and I feel like having it as is might be better than subclass-specific pulling.

codalab/worker/singularity_image_manager.py
adiprerepa: yeah. fixed

codalab/worker/file_util.py
```diff
@@ -576,3 +577,12 @@ def path_is_parent(parent_path, child_path):
     # the parent path will regularize the path name in the same way as the
     # comparison that deals with both paths, removing any trailing path separator.
     return os.path.commonpath([parent_path]) == os.path.commonpath([parent_path, child_path])
+
+
+def sha256(file: str) -> str:
```
epicfaace, line 582: Add a docstring (calculates sha256 of contents of the file)

codalab/worker/main.py
```diff
@@ -380,9 +380,14 @@ def parse_gpuset_args(arg):
         all_gpus = docker_utils.get_nvidia_devices()  # Dict[GPU index: GPU UUID]
     except docker_utils.DockerException:
         all_gpus = {}
-    except Exception:
-        all_gpus = docker_utils.get_nvidia_devices(use_docker=False)
+    # docker socket cant be used
```
epicfaace:
```suggestion
    # Docker socket can't be used
```

codalab/worker/main.py
```diff
@@ -380,9 +380,14 @@ def parse_gpuset_args(arg):
...
+    except requests.exceptions.ConnectionError:
+        try:
+            all_gpus = docker_utils.get_nvidia_devices(use_docker=False)
+        except SingularityError:
+            all_gpus = {}
 
+    logger.error(all_gpus, arg)
```
epicfaace: rm

codalab/worker/singularity_image_manager.py
```diff
@@ -39,15 +40,16 @@ def get(self, image_spec):
         if len(image_spec.split("://")) <= 1:
```
epicfaace:
```suggestion
        if "://" not in image_spec:
```

codalab/worker/singularity_image_manager.py
```diff
@@ -39,15 +40,16 @@ def get(self, image_spec):
...
             # prefix docker, no scheme exists
             image_spec = "docker://" + image_spec
-        ImageManager.get(image_spec)
+        if len(image_spec.split(":")) <= 2:
```
epicfaace:
```suggestion
        scheme, rest = image_spec.split("://")
        if ":" not in rest:
```

codalab/worker/singularity_image_manager.py
epicfaace: todo

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -10,6 +10,7 @@ RUN make
 FROM ubuntu:16.04
 ARG default_user=codalab
 MAINTAINER CodaLab Worksheets <codalab.worksheets@gmail.com>
+# the singularity container runtime needs to be able to lookup the uid of the default user
```
epicfaace:
```suggestion

# The Singularity container runtime needs to be able to lookup the uid of the default user
```

codalab/worker/docker_image_manager.py
```diff
@@ -206,106 +119,138 @@ def image_availability_state(image_spec, success_message, failure_message):
...
-                # Download images if size cannot be obtained
-                if self._max_image_size:
+        return image_size_bytes
+
+    def _image_availability_state(
+        self, image_spec: str, success_message: str, failure_message: str
+    ) -> ImageAvailabilityState:
+        """
+        Try to get the image specified by image_spec from host machine.
+        Return ImageAvailabilityState.
```
epicfaace, line 229: - Document arguments
- specify that failure_message must have a %s in it

codalab/worker/main.py
```diff
@@ -252,6 +263,25 @@ def main():
...
+        singularity_folder = os.path.join(args.work_dir, 'codalab_singularity_images')
+        if not os.path.exists(singularity_folder):
+            logger.info(
+                'Local singularity image location %s doesn\'t exist, creating.', singularity_folder,
+            )
+            os.makedirs(singularity_folder, 0o770)
+        image_manager = SingularityImageManager(
+            args.max_image_size, args.max_image_cache_size, singularity_folder,
+        )
+        docker_runtime = None
```
epicfaace, line 278: put a TODO comment that singularity workers don't currently work and should be implemented in the future

codalab/worker/main.py
```diff
@@ -341,6 +364,10 @@ def parse_cpuset_args(arg):
 def parse_gpuset_args(arg):
     """
     Parse given arg into a set of strings representing gpu UUIDs
+    By default, we will try to start a docker container with nvidia-smi to get the GPUs
```
epicfaace:
```suggestion
    By default, we will try to start a Docker container with nvidia-smi to get the GPUs.
```

codalab/worker/main.py
```diff
@@ -341,6 +364,10 @@ def parse_cpuset_args(arg):
...
+    If we get an exception that the docker socket does not exist, which will be the case
```
epicfaace:
```suggestion
    If we get an exception that the Docker socket does not exist, which will be the case
```

codalab/worker/main.py
```diff
@@ -341,6 +364,10 @@ def parse_cpuset_args(arg):
...
+    on singularity workers, because they do not have root access, and therefore, access to
+    the docker socket, we should try to get the GPUs with singularity.
```
epicfaace:
```suggestion
    the Docker socket, we should try to get the GPUs with Singularity.
```

codalab/worker/main.py
```diff
@@ -341,6 +364,10 @@ def parse_cpuset_args(arg):
 def parse_gpuset_args(arg):
     """
     Parse given arg into a set of strings representing gpu UUIDs
+    By default, we will try to start a docker container with nvidia-smi to get the GPUs
+    If we get an exception that the docker socket does not exist, which will be the case
+    on singularity workers, because they do not have root access, and therefore, access to
```
epicfaace:
```suggestion
    on Singularity workers, because they do not have root access, and therefore, access to
```

docker_config/compose_files/docker-compose.yml
```diff
@@ -86,6 +86,9 @@ x-codalab-base: &codalab-base
   user: ${CODALAB_UID}
   networks:
     - service
+  volumes:
```
epicfaace: rm this

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -23,11 +25,36 @@ RUN apt-get update && \
...
     unzip && \
     rm -rf /var/lib/apt/lists/*;
 
+RUN wget https://golang.org/dl/go1.16.5.linux-amd64.tar.gz && \
+  rm -rf /usr/local/go && \
+  tar -C /usr/local -xzf go1.16.5.linux-amd64.tar.gz;
+
+ENV PATH="/usr/local/go/bin:${PATH}"
+ENV SINGULARITY_VERSION=3.7.4
+ENV USER=$default_user
```
epicfaace: rm?

docker_config/dockerfiles/Dockerfile.worker
epicfaace: todo @adiprerepa : figure out if this `useradd` command is needed

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -23,11 +25,36 @@ RUN apt-get update && \
...
+    squashfs-tools \
+    libseccomp-dev \
+    pkg-config \
     zip \ 
     git \
+    wget \
     unzip && \
     rm -rf /var/lib/apt/lists/*;
 
+RUN wget https://golang.org/dl/go1.16.5.linux-amd64.tar.gz && \
```
epicfaace: Move singularity install to the earlier stage (ecr-login-installation)

docker_config/dockerfiles/Dockerfile.worker
adiprerepa: alright, it is required -- https://www.gitmemory.com/issue/hpcng/singularity/5962/828040364 -- a user needs to be present with a UID

epicfaace: I don't understand, could you explain further?

adiprerepa: @epicfaace updated comment as discussed

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -56,3 +77,7 @@ RUN python3 -m pip install --no-cache-dir -e .
 
 # Allow non-root to read everything
 RUN chmod -R og=u-w .
+# when the worker docker container is run, the uid is the same as the uid that is used to run
```
epicfaace: Let's use this comment instead:

```suggestion
# When running the worker container as a non-root user on the host,
# we need to create a user `codalab` in the Docker container that corresponds to the specified uid / gid.
# See https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf
```

epicfaace (approved):

adiprerepa:
will add tests in a later iteration, when we actually implement running

PR #3578, opened 2021-05-28
Add license to ratarmount code

epicfaace:
Realized that because ratarmount is licensed under the MIT license, we need to include a copy of it.


adiprerepa (approved):
saucy

PR #3580, opened 2021-05-29
Make wide screen view the default for all users

### LINKED ISSUE #3565: Make wide screen view the default for all users, so it would save the state when people go back
Create a new key in localstorage to store this preference (default to Y)

yuqijin:

create a new key `newWorksheetWidth` in localStorage

### Related issues

fixes #3565 

### Screenshots

Screen Shot 2021-05-29 at 12 07 11 AM

### Checklist

### DIFF
--- a/frontend/src/constants.js
+++ b/frontend/src/constants.js
@@ -9,7 +9,7 @@ export const HEADER_HEIGHT = 170;
 
 // Worksheet width
 export const EXPANDED_WORKSHEET_WIDTH = '99%';
-export const DEFAULT_WORKSHEET_WIDTH = '65%';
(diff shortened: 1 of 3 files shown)


frontend/src/constants.js
```diff
@@ -12,7 +12,7 @@ export const EXPANDED_WORKSHEET_WIDTH = '99%';
 export const DEFAULT_WORKSHEET_WIDTH = '65%';
```
epicfaace: We should rename this to NARROW_WORKSHEET_WIDTH, as it's no longer the default.

yuqijin: good catch. Renamed it.

epicfaace (approved):

PR #3591, opened 2021-06-03
Show complete info in Bundle detail side bar when clicking "More Details"

### LINKED ISSUE #3568: Show all bundle info in bundle details
`more details` in the bundle view just shows all the details right there
with a `less details` button showing the info we have right now
with a `view bundle` button to link to the bundle page

yuqijin:

with a `less details` button showing the info we have right now
with a `view bundle` button to link to the bundle page

### Related issues

fixes #3568 

### Screenshots

ezgif com-gif-maker

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
@@ -7,7 +7,7 @@ import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
 import Grid from '@material-ui/core/Grid';
 import { withStyles } from '@material-ui/core/styles';
-import {renderDuration} from '../../../util/worksheet_utils';
+import { renderDuration } from '../../../util/worksheet_utils';
 import { Icon } from '@material-ui/core';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
```diff
@@ -83,6 +98,42 @@ class BundleDetailSideBar extends React.Component<
...
+                keys.push(property);
+        }
+        keys.sort();
+        for (let i = 0; i < keys.length; i++) {
+            let key = keys[i];
+            items.push(
+                <div>
+                    <ConfigLabel label={this.renderInfoLabel(key)} inline={true}/>
+                    <div className={classes.dataText}>
+                        {renderFormat(metadata[key], metadataType[key])}
```
epicfaace, line 131: Are we sure we want to do this in the first place? I'm not sure we really need to show all metadata fields inline; it might look not good if the metadata field is very large, for example, error_traceback.

Was this a suggestion from an earlier demo?

yuqijin: yes, it is a suggestion in last Thursday's dry run demo.

epicfaace:
From meeting: Let's think about how exactly we want to show these fields / which fields we want to show (cc @percyliang )

PR #3592, opened 2021-06-03
Provide better space usage for bundle detail info in a worksheet

### LINKED ISSUE #3566: Better space usage for file browser under bundle detail
File browser of a bundle view in a worksheet should match the size of the screen (including stdout)
When the window narrows or expands, the stdout should always fit the screen

yuqijin:

File browser of a bundle view in a worksheet matches the size of the screen

### Related issues

fixes #3566 

### Screenshots
ezgif com-gif-maker (1)

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
@@ -159,7 +159,7 @@ class BundleDetailSideBar extends React.Component<
                                     </span>
                                 </div>        
                         </CopyToClipboard>
-                        <div style={{ maxWidth: 300, flexWrap: 'wrap', flexShrink: 1}}>      
+                        <div style={{  flexWrap: 'wrap', flexShrink: 1}}>
                             {bundleInfo.command}
                         </div>
                     </Grid>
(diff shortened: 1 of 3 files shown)


percyliang:
The side panel starts out quite wide.  What happens if you have a stdout with very long lines?

yuqijin:
> The side panel starts out quite wide. What happens if you have a stdout with very long lines?
it looks like this
Screen Shot 2021-06-02 at 10 01 08 PM

percyliang:
What if the stdout were even longer, but the command was short (so the side panel doesn't need that much horizontal space)?

yuqijin:
> What if the stdout were even longer, but the command was short (so the side panel doesn't need that much horizontal space)?
Looks like this: it still occupies lots width since we did not set the max width for it.
ezgif com-gif-maker (2)

percyliang:
Perhaps the side panel could be made narrower, since there's a lot of unused space?

yuqijin:
> Perhaps the side panel could be made narrower, since there's a lot of unused space?

I see. How about this? 
image

percyliang:
Looks great!

epicfaace (approved):

PR #3601, opened 2021-06-04
Have the bundle manager check for unresponsive bundles less frequently

### LINKED ISSUE #3563: Don't check _fail_unresponsive_bundles() every iteration of the bundle manager
Currently, we are checking `_fail_unresponsive_bundles()` every iteration of the bundle manager. This can be expensive as we are gather all running, staged and uploading bundles.

teetone:

The bundle manager checking for unresponsive bundles is quite expensive. This PR simply reduces the number of times we check for unresponsive bundles.

### Related issues

Resolves https://github.com/codalab/codalab-worksheets/issues/3563

### Checklist

### DIFF
--- a/codalab/bin/bundle_manager.py
+++ b/codalab/bin/bundle_manager.py
@@ -19,9 +19,19 @@ def main():
         type=int,
         default=60,
     )
+    parser.add_argument(
+        '--unresponsive-bundles-check-frequency-hours',
+        help='Number of hours between checks for failing unresponsive bundles.',
+        type=int,
+        default=8,
+    )
     args = parser.parse_args()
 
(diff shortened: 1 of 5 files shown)


epicfaace (commented):
Question: when it checks for unresponsive bundles, does it block up the entire bundle manager so that it can't process other bundles?

If so, I think that rather than performing the check less often, we may want to reduce the scope of _fail_unresponsive_bundles instead (so it only checks a subset of bundles, perhaps the 100 oldest ones).

epicfaace (approved):

epicfaace:
Closing in favor of doing more research as mentioned in https://github.com/codalab/codalab-worksheets/issues/3646

PR #3616, opened 2021-06-08
Eliminate delay for deleting bundles

### LINKED ISSUE #3571: Delay for deleting bundles
When deleting a lot of bundles from the web UI, there's a progress bar for 'Deleting 20 bundles...' which takes a while but then after that, there's still a delay.
Ideally, all the waiting would be in the progress bar.

yuqijin:
### Related issues

fixes #3571 

### Screenshots

ezgif com-gif-maker (1)

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -308,6 +308,7 @@ class Worksheet extends React.Component {
             closeOnClick: true,
             pauseOnHover: false,
             draggable: true,
+            autoClose: false,
         });
         executeCommand(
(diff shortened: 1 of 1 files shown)


percyliang:
What happens when bundle deletion actually takes a while?  Is there a
progress bar?

epicfaace:
> What happens when bundle deletion actually takes a while? Is there a progress bar?

The progress bar still shows up with this PR change. I think the only difference with this PR change is that the toast doesn't close early (it was closing too early before this PR) -- thus, the toast is visible throughout the entire bundle deletion process.

epicfaace (approved):

PR #3617, opened 2021-06-08
Fix bug of cut & paste bundles

### LINKED ISSUE #3572: Cut and Paste Bundles
Cut a bundle (a d) and then go to the last bundle and try to paste (a v), it still pastes it at the top.

yuqijin:
### Related issues
fixes #3572 

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/Worksheet/Worksheet.js
+++ b/frontend/src/components/worksheets/Worksheet/Worksheet.js
@@ -103,6 +103,9 @@ class Worksheet extends React.Component {
                 showMessage: false,
                 messageContent: null,
             },
+            showUpdateProgress: false,
+            showWorksheetContent: false,
+            showWorksheetContainer: true,
         };
         this.copyCallbacks = [];
         this.showContentCallbacks = [];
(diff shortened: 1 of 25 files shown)


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -519,7 +519,7 @@ class Worksheet extends React.Component {
         } else {
             // If no location for the insertion is specified,
             // insert the new item at the top of the worksheet in default.
```
```suggestion
            // insert the new item at the top of the worksheet by default.
```

frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -519,7 +519,7 @@ class Worksheet extends React.Component {
...
-            actualData['after_sort_key'] = -1;
+            actualData['after_sort_key'] = 0;
```
epicfaace, line 525: To solve issues like this in the future, maybe we should define a constant called `WORKSHEET_TOP_POS` and set it to -1, and use it in all places that depend on the worksheet top position. What do you think? (we could probably do this in another PR as this is just a general refactoring thing though)

epicfaace (approved):

frontend/src/components/worksheets/Worksheet/Worksheet.js
yuqijin: Yes, I think it is a good suggestion!

epicfaace: Tracking in https://github.com/codalab/codalab-worksheets/issues/3623

epicfaace (approved):

Issue #3579, opened 2021-05-28
Manual upload zip file stuck 


lynn-he:
Manual upload of three zip files stuck in "uploading" for about 45 minutes

**Codalab Information**
- *Your username*: DeepLearning.AI
- *UUID of worksheet where problem occurred*: 0xaece8d397dd54037bf12603b2aea0d1f
- *UUID of problem bundle (please don't delete)*: 0xc620f44d71b3451f9f135a28c7302d1b, 0x7df31975843c4ecb8b3459324c26d26a, 0xa7c4c320e32f4b0ebea63626877d8578
- *Error message received from Codalab (if applicable)*: N/A
 - *OS (if on CLI)*:
 - *Codalab CLI Version (if on CLI)*: 
 - *Browser (if on Web)*; Chrome

**To Reproduce**
Steps to reproduce the behavior :
Try manual uploading this file: 
[val.zip](https://github.com/codalab/codalab-worksheets/files/6563345/val.zip)

**Expected behavior**
Upload process to end within 5 minutes

**Screenshots**
<img width="1189" alt="Screen Shot 2021-05-28 at 3 04 45 PM" src="https://user-images.githubusercontent.com/51978114/120042218-a4e0c580-bfd7-11eb-9e54-112b336dfae7.png">

lynn-he:
[val.zip](https://github.com/codalab/codalab-worksheets/files/6563663/val.zip)
have also tried with only png files and no __MACOSX folders

epicfaace:
@lynn-he as a temporary workaround, it looks like this does work if you use `cl upload val.zip` from the CLI. We're working on fixing this issue on the frontend as well, though!

lynn-he:
@epicfaace thank you!!

epicfaace:
Stack trace for zip file uploads:


[...]

PR #3618, opened 2021-06-08
Inline function to attempt to fix .zip file problem

epicfaace:
Addresses https://github.com/codalab/codalab-worksheets/issues/3579. It still appears that the temporary file is not being fully written / flushed to disk (or maybe it's being removed prematurely?). Inlining the do_unzip function to see if that fixes it.


epicfaace:
@jzwang43 can you also merge this one before deployment?

jzwang43 (approved):

PR #3622, opened 2021-06-11
Revert "Monitor: Show results of df in human-readable format"

### LINKED ISSUE #3621: Monitor script broken --  invalid literal for int() with base 10: '8.1G'
image

epicfaace:
Reverts codalab/codalab-worksheets#3610. Fixes https://github.com/codalab/codalab-worksheets/issues/3621 because the code assumes that everything should be ints, not human-formatted.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -251,7 +251,7 @@ def backup_db():
 
 
 def check_disk_space(paths):
-    lines = run_command(['df', '-h'] + paths).split('\n')[1:]
+    lines = run_command(['df'] + paths).split('\n')[1:]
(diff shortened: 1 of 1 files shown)


adiprerepa:
Should be instead change the code's behavior to parse the human readable format?

epicfaace:
> Should be instead change the code's behavior to parse the human readable format?

Yeah, that's probably the long-term solution, but given that the monitor is completely broken now, I just wanted to fix it with this PR.

adiprerepa (approved):
Lgtm

PR #3626, opened 2021-06-18
Fix "python" executable in default-cpu and default-gpu

epicfaace:

Fix "python" executable in default-cpu and default-gpu dockerfiles so that it points to Python 3.6.


jzwang43 (approved):
Looks good

PR #3652, opened 2021-06-29
Shared work directory

teetone:

Have the workers share the same working directory to download dependencies to. This was the way it was done with static workers.


epicfaace (approved):

adiprerepa (approved):

percyliang (commented):
Isn't there a problem if two workers share the same scratch directory, there can be a clash?  With static workers, we didn't have this problem because we just ensured that one machine just ran one worker.

epicfaace:
It's not a problem on Azure Batch because we only have one worker running
at a time on Azure Batch workers. Essentially, workers will consecutively
run on an Azure Batch VM (but no more than one worker will run at the same
time).

--
Ashwin Ramaswami

percyliang:
This won't affect the slurm workers though?

epicfaace:
@teetone ?

epicfaace:
Actually -- this change only affects azure_batch_worker_manager.py, so it won't affect the slurm workers.

teetone:
This change will only affect Azure Batch workers. We can run stress tests in dev to make sure everything is okay.

PR #3663, opened 2021-07-02
Worker to retry downloading dependencies to deal with transient errors

### LINKED ISSUE #3457: Failed to download dependency celebA: Dependency download failed: IncompleteRead(0 bytes read)
See https://worksheets.codalab.org/bundles/0xc26e560d35d94a2aa475ac8a2acb2df9

teetone:

Added `--worker-download-dependencies-max-retries` parameter to the worker.

### Related issues

Resolves #3457

### Checklist

### DIFF
--- a/tests/unit/worker_manager/slurm_batch_worker_manager_test.py
+++ b/tests/unit/worker_manager/slurm_batch_worker_manager_test.py
@@ -15,6 +15,7 @@ class SlurmBatchWorkerManagerTest(unittest.TestCase):
             partition='some_partition',
             worker_executable='cl-worker',
             worker_idle_seconds='888',
+            worker_download_dependencies_max_retries=5,
             worker_tag='some_tag',
             worker_group='some_group',
             worker_exit_after_num_runs=8,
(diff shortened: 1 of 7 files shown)


tests/unit/worker_manager/slurm_batch_worker_manager_test.py
```diff
@@ -15,6 +15,7 @@ def test_base_command(self):
             partition='some_partition',
             worker_executable='cl-worker',
             worker_idle_seconds='888',
+            worker_download_dependencies_max_retries=5,
```
epicfaace, line 18: Why 5 here instead of 3?

epicfaace (commented):
Looks good, but I'm not convinced this needs to be configurable. Do you think it's worth the additional complexity and arguments added?

adiprerepa:
do we know for a fact that retries fixes the issue?

epicfaace:
We're pretty sure it's a transient error during deployment; see https://github.com/codalab/codalab-worksheets/issues/3457#issuecomment-855001732 re. our conversation last meeting.

tests/unit/worker_manager/slurm_batch_worker_manager_test.py
teetone: Just for to test if setting a value other than the default value of 3 gets respected.

epicfaace (approved):

PR #3671, opened 2021-07-06
Fix blob upload issue for large files

### LINKED ISSUE #3665: Blob storage error when uploading large (11 GB) file
I get this error when uploading a 11 GB file using `cl upload -a`:

Traceback (most recent call last): File "/opt/codalab-worksheets/codalab/rest/bundles.py", line 798, in _update_bundle_contents_blob use_azure_blob_beta=use_azure_blob_beta, File "/opt/codalab-worksheets/codalab/lib/upload_manager.py", line 217, in upload_to_bundle_store bundle, source, git, unpack File "/opt/codalab-worksheets/codalab/lib/upload_manager.py", line 78, in upload_to_bundle_store source_ext, source_fileobj, bundle_path, unpack_archive=False File "/opt/codalab-worksheets/codalab/lib/upload_manager.py", line 169, in write_fileobj shutil.copyfileobj(output_fileobj, out) File "/usr/local/lib/python3.6/dist-packages/apache_beam/io/filesystemio.py", line 220, in close self._uploader.finish() File "/opt/codalab-worksheets/codalab/lib/beam/blobstorageuploader.py", line 54, in finish self._write_to_blob(self.buffer) File "/opt/codalab-worksheets/codalab/lib/beam/blobstorageuploader.py", line 49, in _write_to_blob self._blob_to_upload.stage_block(block_id, data) File "/usr/local/lib/python3.6/dist-packages/azure/core/tracing/decorator.py", line 83, in wrapper_use_tracer return func(*args, **kwargs) File "/usr/local/lib/python3.6/dist-packages/azure/storage/blob/_blob_client.py", line 1977, in stage_block return self._client.block_blob.stage_block(**options) File "/usr/local/lib/python3.6/dist-packages/azure/storage/blob/_generated/operations/_block_blob_operations.py", line 349, in stage_block raise models.StorageErrorException(response, self._deserialize) File "/usr/local/lib/python3.6/dist-packages/azure/storage/blob/_generated/models/_models_py3.py", line 1817, in __init__ self.error = deserialize(model_name, response) File "/usr/local/lib/python3.6/dist-packages/msrest/serialization.py", line 1367, in __call__ data = self._unpack_content(response_data, content_type) File "/usr/local/lib/python3.6/dist-packages/msrest/serialization.py", line 1541, in _unpack_content raw_data.headers File "/usr/local/lib/python3.6/dist-packages/msrest/pipeline/universal.py", line 226, in deserialize_from_http_generics return cls.deserialize_from_text(body_bytes, content_type) File "/usr/local/lib/python3.6/dist-packages/msrest/pipeline/universal.py", line 203, in deserialize_from_text raise DeserializationError("Cannot deserialize content-type: {}".format(content_type)) msrest.exceptions.DeserializationError: Cannot deserialize content-type: text/html

Bundle: https://worksheets-dev.codalab.org/bundles/0x72a7d091bb3d4388946982bdc88d367a

epicfaace:
Fixes #3665. The problem was that upon `finish()` being called on the BlobStorageUploader, we were calling _write_to_blob (and thus stage_block) directly with the BytesBuffer `self.buffer` as the argument. This has two problems:
- I don't think Blob Storage plays well with us directly passing in a custom class to stage_block. It really just expects a bytestring instead.
- If self.buffer is longer than the maximum block size, then this will cause an error.

Thus, the solution is to split it into chunks, similar to the logic in `put` that periodically calls `_write_to_blob`.

### DIFF
--- a/codalab/lib/beam/blobstorageuploader.py
+++ b/codalab/lib/beam/blobstorageuploader.py
@@ -51,8 +51,6 @@ class BlobStorageUploader(Uploader):
     self.block_number = self.block_number + 1
 
   def finish(self):
-    while len(self.buffer) > 0:
-      # Take the first chunk off the buffer and write it to Blob Storage
-      chunk = self.buffer.read(BlobStorageUploader.MAX_WRITE_SIZE)
-      self._write_to_blob(chunk)
+    # The buffer will have a size smaller than MIN_WRITE_SIZE, so its contents can fit into memory.
+    self._write_to_blob(self.buffer.read())
     self._blob_to_upload.commit_block_list(self.block_list, content_settings=self._content_settings)
--- a/codalab/lib/beam/blobstorageuploader.py
+++ b/codalab/lib/beam/blobstorageuploader.py
@@ -12,6 +12,15 @@ class BlobStorageUploader(Uploader):
   that handles multipart streaming (block-by-block) uploads.
   TODO (Ashwin): contribute this back upstream to Apache Beam (https://github.com/codalab/codalab-worksheets/issues/3475).
   """
+  # Note that Blob Storage currently can hold a maximum of 100,000 uncommitted blocks.
+  # This means that with this current implementation, we can upload a file with a maximum
+  # size of 10 TiB to Blob Storage. To exceed that limit, we must either increase MIN_WRITE_SIZE
+  # or modify the implementation of this class to call commit_block_list more often (and not
+  # just at the end of the upload). 
+  MIN_WRITE_SIZE = 100 * 1024 * 1024
+  # Maximum block size is 4000 MiB (https://docs.microsoft.com/en-us/rest/api/storageservices/put-block#remarks).
+  MAX_WRITE_SIZE = 4000 * 1024 * 1024
+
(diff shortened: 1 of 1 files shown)


codalab/lib/beam/blobstorageuploader.py
```diff
@@ -51,5 +51,8 @@ def _write_to_blob(self, data):
     self.block_number = self.block_number + 1
 
   def finish(self):
-    self._write_to_blob(self.buffer)
+    while len(self.buffer) > 0:
```
adiprerepa: why is `finish()` being called if there are still more bytes in the buffer than `MAX_WRITE_SIZE`?

epicfaace: Ok, that's true -- there will always be less bytes in the buffer than `MIN_WRITE_SIZE` at this point. Will change it to just do:

`self._write_to_blob(self.buffer.read())`

instead.

adiprerepa (approved):
lgtm -- the change that matters then is the `self.buffer` -> `self.buffer.read()` to pass in a raw bytestring, right?

epicfaace:
Yes

PR #3699, opened 2021-07-08
Fix signup error when user is already logged in, fixes #3698

### LINKED ISSUE #3698: [USER ISSUE] Can't sign up for an account from China
**Describe the bug**
when I sign up on https://worksheets.codalab.org/account/signup，the “sign up>>” button in the bottom has no response so I can't regist.

**To Reproduce**
1. Go to 'https://worksheets.codalab.org/account/signup'
2. Fill in all the fields and click on 'sign up' and it doesn't work！

**Screenshots**
image

epicfaace:
Fix signup error when user is already logged in, fixes part of #3698. Removed redundant error-handling code as this error case is already dealt with later on in the file:

https://github.com/codalab/codalab-worksheets/blob/e0a317573c35cf4fd7899d39bf6e9e98bba9ebd4/codalab/rest/account.py#L96-L100

Now, it will show this error:

image

### DIFF
--- a/codalab/rest/account.py
+++ b/codalab/rest/account.py
@@ -2,7 +2,7 @@
 Login and signup views.
 Handles create new user accounts and authenticating users.
 """
-from bottle import request, response, template, local, redirect, default_app, get, post
+from bottle import request, response, template, local, redirect, get, post
 import requests
 import os
 from codalab.lib import crypt_util, spec_util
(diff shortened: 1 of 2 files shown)


epicfaace:
@adiprerepa can you review?

adiprerepa (approved):

Issue #2971, opened 2020-10-16
Blob Storage: bypass codalab server and contact blob storage directly


epicfaace:
- Let the worker upload / download directly to S3 / Azure Blob

epicfaace:
Design doc: https://docs.google.com/document/d/1wr5xPDimlai-2IEiRExV1p062fWj_einp2gpj6h5rKs/edit

PR #3733, opened 2021-07-30
Upgrade azure-storage-blob==12.8.1, azure-core==1.10.0

epicfaace:
These are required to properly generate SAS tokens for #2971 .

Doing this in this PR so that we can get the requirements into master, so that future Docker builds will be faster / cached.


adiprerepa (approved):

epicfaace:
Actually, this isn't needed. The feature I needed to use (`is_directory`) doesn't work well for my use case, actually, so I don't need it.

PR #3734, opened 2021-07-30
Fix link to REST API reference on https://worksheets.codalab.org/

epicfaace:
The previous link was broken.

### DIFF
--- a/frontend/src/routes/HomePage.js
+++ b/frontend/src/routes/HomePage.js
@@ -270,7 +270,7 @@ class HomePage extends React.Component<{
                                         {this.renderTableItem(
                                             'REST API Reference',
                                             'Develop your own application against our REST API.',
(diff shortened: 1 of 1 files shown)


adiprerepa (approved):

frontend/src/routes/HomePage.js
```diff
@@ -270,7 +270,7 @@ class HomePage extends React.Component<{
                                         {this.renderTableItem(
                                             'REST API Reference',
                                             'Develop your own application against our REST API.',
```
pranavjain, line 272: As a suggestion, can we move these strings to separate JSON files rather than directly hardcoding them in JS? It would be useful and easy to localize them if we decide to support multiple languages later on.

pranavjain (approved):

frontend/src/routes/HomePage.js
epicfaace: Yeah, do you want to make an issue for that so we could address it later?

PR #3737, opened 2021-07-30
Blob Storage: Use SAS tokens to bypass server when downloading entire bundles through one's browser or the CLI

epicfaace:

Blob Storage: Use SAS tokens to bypass server when downloading entire bundles either through one's browser (through the `/contents/blob/` endpoint) or through the CLI (through a command such as `cl cat` or `cl download`). Implements part of https://github.com/codalab/codalab-worksheets/issues/2971.

Although the files are stored as .gz or .tar.gz files, in the case of single files, we use transparent decompression to automatically decompress single bundles (so that the user can directly see a text file / image in their browser). In the case of .tar.gz files, we just download it as usual.

These provisions allow us to have the same interface that users used to have when downloading non-Blob Storage files. The benefit here is that **the server is bypassed entirely** and all decompression, if it happens, happens on the client / browser side.

### support_redirect query parameter

This query parameter is for the GET /contents/blob endpoint.

- `support_redirect`: Set to 1 if the client supports bypassing the server
  and redirecting to another URL (such as Blob Storage). If so, the Target-Type and
  X-CodaLab-Target-Size headers will not be present in the response.

  If this endpoint is called from a web browser (`Referer` header is set), this parameter
  defaults to 1. Otherwise, it defaults to 0, meant for compatibility
  with older clients / CLI versions that depend on the Target-Type and
  X-CodaLab-Target-Size headers. In a future release, this parameter will always
  be set to 1.

### Examples

A bundle that is a text file:

text

A bundle that is an image file:

image

A bundle that is a directory (.tar.gz file):

folder


epicfaace:
TODO: check to see if `--pack` works with this as well -- I think this is why the tests are failing.

Additionally, try to add tests for this endpoint through the REST tests.

epicfaace:
related: https://github.com/Azure/Azurite/issues/997

epicfaace:
@adiprerepa this should be ready for review now.

adiprerepa (commented):
querystring fix then approve

epicfaace:
> querystring fix then approve

Done @adiprerepa

codalab/common.py
```diff
@@ -209,6 +218,32 @@ class LinkedBundlePath:
...
+        blob_name = path.replace(
+            f"azfs://{AZURE_BLOB_ACCOUNT_NAME}/{AZURE_BLOB_CONTAINER_NAME}/", ""
+        )  # for example, "0x9955c356ed2f42e3970bdf647f3358c8/contents.gz"
+        sas_token = generate_blob_sas(
+            **kwargs,
+            account_name=AZURE_BLOB_ACCOUNT_NAME,
+            container_name=AZURE_BLOB_CONTAINER_NAME,
+            account_key=AZURE_BLOB_ACCOUNT_KEY,
+            permission=BlobSasPermissions(read=True),
+            expiry=datetime.datetime.now() + datetime.timedelta(hours=1),
```
adiprerepa, line 236: are there scenarios where bundle access takes more than an hour? I'm thinking downloads that are many many GB on a low throughput network

tests/cli/test_cli.py
```diff
@@ -878,6 +912,51 @@ def test_upload4(ctx):
...
+    CODALAB_ALWAYS_USE_AZURE_BLOB_BETA is set to 1."""
+
+    def fetch_contents_blob_no_redirect(uuid):
+        """Fetch blob contents without redirecting to Blob Storage.
+        We don't redirect to Blob Storage, and Target-Type and X-CodaLab-Target-Size
+        should always be present on the response, if we don't pass in the
+        `support_redirect` parameter. In a future version, we will change this so that
+        this guarantee no longer exists (see comment at bundles.py:_fetch_bundle_contents_blob).
+        """
+        return ctx.client._make_request(
```
adiprerepa, line 928: maybe the `supports_redirect=0` parameter should be included to be explicit?

codalab/client/json_api_client.py
```diff
@@ -621,7 +621,7 @@ def fetch_contents_blob(self, target, range_=None, head=None, tail=None, truncat
         headers = {'Accept-Encoding': 'gzip'}
         if range_ is not None:
             headers['Range'] = 'bytes=%d-%d' % range_
-        params = {}
+        params = {'support_redirect': 1}
```
adiprerepa, line 624: nit: should `support_redirect` be conditionally set here based on other things like `head` and `tail`? 

I'm fine with it being done in `bundles.py` as well.

adiprerepa (approved):
lgtm

codalab/common.py
epicfaace: I'm pretty sure this means that as long as the image has _started_ downloading within one hour, the SAS token should work for the remainder of that request.

tests/cli/test_cli.py
epicfaace: No, I want to test to ensure that the default behavior is to set it to 0.

codalab/client/json_api_client.py
epicfaace: I'd rather do it in `bundles.py`, as there are several other conditions (in addition to just head and tail) that it checks.

PR #3741, opened 2021-08-04
[DO NOT DELETE / MERGE] Fix multigpu shared memory issues

teetone:
Keeping this PR around so we can use the worker image until the next deployment.


epicfaace:
Should we close this PR in favor of https://github.com/codalab/codalab-worksheets/pull/3748?

teetone:
> Should we close this PR in favor of #3748?

I need it until we deploy.

PR #3748, opened 2021-08-10
Make the shared memory size of the run containers configurable

### LINKED ISSUE #3742: Multi-gpu + multi-worker data loaders not working in CodaLab
We need to make the shared memory configurable.

```
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 872, in _try_get_data
    data = self._data_queue.get(timeout=timeout)
  File "/opt/conda/lib/python3.8/queue.py", line 179, in get
    self.not_empty.wait(remaining)
  File "/opt/conda/lib/python3.8/threading.py", line 306, in wait
    gotit = waiter.acquire(True, timeout)
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/signal_handling.py", line 66, in handler
    _error_if_any_worker_fails()
RuntimeError: DataLoader worker (pid 52) is killed by signal: Bus error. It is possible that dataloader's workers are out of shared memory. Please try to raise your shared memory limit.
```

teetone:

Make the shared memory size of the run containers configurable.

### Related issues

Fixes https://github.com/codalab/codalab-worksheets/issues/3742

### Checklist


codalab/worker/main.py
```diff
@@ -184,6 +184,12 @@ def parse_args():
         default=3,
         help='The number of times to retry downloading dependencies after a failure (defaults to 3).',
     )
+    parser.add_argument(
+        '--shared-memory-size-gb',
+        type=int,
+        default=1,
```
epicfaace, line 190: Maybe it's better to make this argument similar to other arguments with byte inputs (such as `--max-memory`)? So people would input things like `1g`.

epicfaace (approved):
I'm also good with just merging this as-is.

PR #3749, opened 2021-08-10
Prevent script failures and better test competitiond.py

### LINKED ISSUE #2242: [USER ISSUE] - cl-competitiond fails if any one of the submission bundles are not accessible
**Describe the bug**
The cl-competitiond script fails during a run if at least one run bundle has a private dependency. Based on the fetch order, the script either fails at the first mimic or runs a few bundles and then fails.

**Codalab Information**
- *Your username*: imalhi
- *UUID of worksheet where problem occurred*: local cli
- *Error message received from Codalab (if applicable)*: 404 Bundle not found
 - *OS (if on CLI)*: Mac Os, Ubuntu 18
 - *Codalab CLI Version (if on CLI)*: 0.5.13
 - *Browser (if on Web)* : N/A

**To Reproduce**
1. Create a competition, or use an existing one
2. Make the src bundle of a submission bundle's upstream run bundle private.
3. Run cl-competitiond for the competition
4.  Script fails with a 404 error

**Expected behavior**
The script should skip the 404'd bundle and run the rest.

**Screenshots**
N/A

**Additional context**
I have a semi-working fix that uses a try-catch to ignore 404 errors when looping through submissions.

### LINKED ISSUE #3705: Competition daemon fails when bundle does not have public read permissions
Reported by @lynn-he

Just ran into this issue with the new competition daemon:

```
[INFO] 2021-07-09 17:21:16,459: Mimicking submission for naveenkb
[INFO] 2021-07-09 17:21:17,061: Submission 0xbeabde273acf442eb66a0d8f77cb4eb1 by naveenkb is missing expected dependencies.
[INFO] 2021-07-09 17:21:17,061: Aborting submission for naveenkb
  File "competitiond.py", line 768, in run
    self.run_once()
  File "competitiond.py", line 761, in run_once
    self.generate_leaderboard(num_total_submissions, num_period_submissions)
  File "competitiond.py", line 692, in generate_leaderboard
    or submit_bundle['metadata']['description'],  # Allow description override
KeyError: 'description'
```

would expect that it shouldn't quit the daemon

_Originally posted by @lynn-he in https://github.com/codalab/codalab-worksheets/issues/3700#issuecomment-877506123_

robinjia:

As discussed on the thread for #2242 . I'm including all my `competitiond.py` script changes, which handle a few different edge cases that have shown up in SQuAD 2.0. Apologies for not sending a PR sooner.

### Related issues

Should fix #2242 and #3705

### DIFF
--- a/scripts/competitiond.py
+++ b/scripts/competitiond.py
@@ -713,8 +713,13 @@ class Competition(object):
                         ],
                         'created': submit_bundle['metadata']['created'],
                     }
-                except KeyError as e:
-                    print('Error when processing bundle {}'.format(eval_bundle['id']))
+                except KeyError:
+                    logger.error(
+                        "Aborting submission {id} for "
+                        "{owner[user_name]}: Error when processing eval bundle".format(
+                            **eval_bundle
+                        )
+                    )
                     traceback.print_exc()
 
             leaderboard.append(
@@ -753,12 +758,15 @@ class Competition(object):
                 )
                 try:
                     predict_bundle = self.run_prediction(submit_bundle)
-                except Exception as e:
+                except Exception:
+                    logger.error(
+                        "Aborting submission {id} for " "{owner[user_name]}".format(**submit_bundle)
+                    )
                     traceback.print_exc()
                     continue
                 if predict_bundle is None:
-                    logger.info(
-                        "Aborting submission for " "{owner[user_name]}".format(**submit_bundle)
+                    logger.error(
+                        "Aborting submission {id} for " "{owner[user_name]}".format(**submit_bundle)
                     )
                     continue
                 self.run_evaluation(submit_bundle, predict_bundle)
--- a/scripts/competitiond.py
+++ b/scripts/competitiond.py
@@ -682,6 +682,18 @@ class Competition(object):
         leaderboard = []
         for eval_bundle in eval_bundles.values():
             meta = self._get_competition_metadata(eval_bundle)
+
+            # In case there isn't a corresponding submit bundle, use some sane
+            # defaults based on just the eval bundle.
+            submission_info = {
+                'description': eval_bundle['metadata']['description'],
+                'public': None,
+                'user_name': None,
+                'num_total_submissions': 0,
+                'num_period_submissions': 0,
+                'created': eval_bundle['metadata']['created'],
+            }
+
             if eval_bundle['id'] in eval2submit:
                 submit_bundle = eval2submit[eval_bundle['id']]
                 try:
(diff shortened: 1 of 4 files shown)


scripts/competitiond.py
```diff
@@ -738,10 +743,19 @@ def run_once(self):
...
+                        "Mimicking submission for " "{owner[user_name]}".format(**submit_bundle)
+                    )
+                else:
+                    logger.info(
+                        "Mimicking submission for " "{metadata[description]} (no username)".format(**submit_bundle)
+                    )
+                try:
+                    predict_bundle = self.run_prediction(submit_bundle)
+                except Exception as e:
+                    traceback.print_exc()
```
epicfaace, line 765: Should we also log something like this?

```suggestion
                    traceback.print_exc()
                    logger.info(
                        "Aborting submission for " "{owner[user_name]}".format(**submit_bundle)
                    )
```

epicfaace (commented):
This is great, thanks for the PR!

scripts/competitiond.py
robinjia: It's possible for the user_name field to not exist (that's why I added the if statement above), so we'd need a similar if statement here.

epicfaace:
@lynn-he would you like to try the script in this PR? It might fix the issue you've been facing with the daemon crashing when bundles with the right permissions aren't found.

You can try it by downloading the following file: https://github.com/codalab/codalab-worksheets/blob/3823061f05fa50af98498ac94dbc8bd1a7f6e817/scripts/competitiond.py

lynn-he:
Hey @epicfaace , thanks for your message. I'll try the script out! So far no issues.

scripts/competitiond.py
```diff
@@ -743,14 +748,9 @@ def run_once(self):
...
             for submit_bundle in submissions:
-                if 'user_name' in submit_bundle['owner']:
-                    logger.info(
-                        "Mimicking submission for " "{owner[user_name]}".format(**submit_bundle)
-                    )
-                else:
-                    logger.info(
-                        "Mimicking submission for " "{metadata[description]} (no username)".format(**submit_bundle)
-                    )
+                logger.info(
```
epicfaace, line 756: @robinjia I removed the `user_name` check because there seem to be many other places where we just assume that "user_name" exists, and I can't think of a way to replicate this error. Let me know if you have an idea about how to replicate this, and if so, we can add it back in.

epicfaace:
@robinjia I've added tests that specifically test some basic use cases with competitiond as well as #3705.

epicfaace (approved):

epicfaace:
@teetone FYI, as part of this PR, I removed the `competition` test from the protected mode test suite -- this is because we need to create a new user and log in as that user in order to properly test competitions.

Issue #2495, opened 2020-06-24
Reenable EditWorksheetTest in GHA


teetone:
The `EditWorksheetTest` selenium test was flaky in GHA. We need to investigate further and reenable this test.

Disable test PR: https://github.com/codalab/codalab-worksheets/pull/2493

PR #3759, opened 2021-08-22
Add UploadDownloadBundleTest (UI test)

epicfaace:
Tests uploading and downloading of bundles, as well as viewing them in the browser. I wanted some more robust tests for uploading / downloding bundles, especially with blob storage changes such as https://github.com/codalab/codalab-worksheets/pull/3737.

Related issue: https://github.com/codalab/codalab-worksheets/issues/2495

I think the edit worksheet tests were flaky earlier because we relied a bunch on calling sleep() to sleep for a static amount of time rather than explicitly tying all our waits to certain elements being visible / invisible (see https://www.lambdatest.com/blog/types-of-waits-in-selenium/ for a more detailed explanation of the drawbacks of using sleep()). -- cc @teetone


tests/ui/ui_tester.py
```diff
@@ -395,15 +489,16 @@ def test(self):
         self.refresh_worksheet()
 
         # Take screenshots and compare to the existing baseline images
-        num_of_screenshots = 1
-        self.output_images('worksheet_container', num_of_screenshots)
-        self.compare_to_baselines(num_of_screenshots)
+        # num_of_screenshots = 1
```
adiprerepa, line 492: remove?

adiprerepa (approved):
lgtm

PR #3767, opened 2021-08-27
Set up NFS for Kubernetes Worker Manager

### LINKED ISSUE #3717: Integrate PersistentVolumes with Kubernetes Worker Manager
Allow users to specify (or create?) Persistent Volumes when starting the Kubernetes Worker Manager. Test it out with GKE.

teetone:

Set up NFS for Kubernetes Worker Manager.

Instructions to set up the NFS server can be found here: https://github.com/codalab/deployment/pull/177

### Related issues

Resolves #3717

### DIFF
--- a/codalab/worker_manager/kubernetes_worker_manager.py
+++ b/codalab/worker_manager/kubernetes_worker_manager.py
@@ -40,6 +40,12 @@ class KubernetesWorkerManager(WorkerManager):
             help='Path to the SSL cert for the Kubernetes cluster',
             required=True,
         )
+        subparser.add_argument(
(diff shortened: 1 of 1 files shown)


adiprerepa (commented):
this lgtm, but why do we actually define the Pod as a dict? I would think the better practice would be to template a separate yaml file, no?

teetone:
> this lgtm, but why do we actually define the Pod as a dict? I would think the better practice would be to template a separate yaml file, no?

I'm using the `utils.create_from_dict` function. hmm I'm not sure if I understand why it would be better to create a new yaml file for this?

adiprerepa (approved):
lgtm

PR #3768, opened 2021-08-27
Simplify frontend test running on CI

epicfaace:

- Update UI test running instructions to show how to run from Docker, which will replicate the exact conditions through which UI tests are run 
- remove unneeded steps in CI for frontend tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -382,12 +382,8 @@ jobs:
       - run: pip install -r requirements.txt
       - name: Setup tests
         run: |
-          sudo ./scripts/test-setup.sh
-          export PATH=$PATH:$PWD/geckodriver
           sudo service mysql stop
           python3 codalab_service.py build services --version ${VERSION} --pull
-          pip install --upgrade diffimg==0.2.3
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -382,12 +382,8 @@ jobs:
       - run: pip install -r requirements.txt
       - name: Setup tests
         run: |
-          sudo ./scripts/test-setup.sh
```
epicfaace, line 385: @teetone this is interesting, do you know why these steps aren't actually needed? Maybe because the tests are actually running within the docker container?

.github/workflows/test.yml
```diff
@@ -382,12 +382,8 @@ jobs:
...
-          export PATH=$PATH:$PWD/geckodriver
           sudo service mysql stop
           python3 codalab_service.py build services --version ${VERSION} --pull
-          pip install --upgrade diffimg==0.2.3
```
adiprerepa, line 389: why arent these needed?

epicfaace: not sure, but the tests still pass

teetone (approved):

PR #3777, opened 2021-09-03
Pin apex version to fix default gpu dockerfile builds

epicfaace:
Pin apex version to fix default gpu dockerfile build.

Fixes the issue @teetone faced with v1.1.1 (see https://github.com/codalab/codalab-worksheets/runs/3500957811?check_suite_focus=true).


epicfaace:
@teetone this fix fixes the build, and I tested on CI. Can you review?

teetone (approved):

Issue #3775, opened 2021-09-03
[USER ISSUE] CLI installation fails when using Python 3.9


aniketsharma00411:
**Describe the bug**
I use Python 3.9 and I was not able to install it using pip as instructed in the [Quickstart guide](https://codalab-worksheets.readthedocs.io/en/latest/examples/quickstart/). codalab has apache-beam[azure]==2.27.0 as a requirement and apache-beam[azure]==2.27.0 has pyarrow<3.0.0,>=0.15.1 as a requirement and pyarrow 2.0.0 (the only version below 3.0.0) does not support Python 3.9.

I was able to install Codalab CLI after removing the version requirement from apache-beam[azure] i.e. changing `apache-beam[azure]==2.27.0` in requirements.txt to `apache-beam[azure]`.

**Codalab Information**
- *Your username*: aniketsharma00411
- *UUID of worksheet where problem occurred*: N/A
- *UUID of problem bundle (please don't delete)*: N/A
- *Error message received from Codalab (if applicable)*: N/A
 - *OS (if on CLI)*: Garuda Linux
 - *Codalab CLI Version (if on CLI)*: 1.1.1
 - *Browser (if on Web)*; N/A

**To Reproduce**
Steps to reproduce the behavior :
1. Update Python version to Python 3.9
2. Install Codalab CLI using `pip install codalab -U --user`

**Expected behavior**
CLI should have been installed.

**Screenshots**

**Additional context**

epicfaace:
I think we can only fix this once the latest version of Apache Beam supports Python 3.9.

aniketsharma00411:
It does. After installing the latest version of Apache Beam I am able to use the CLI properly. It gives a warning that the version is not tested properly though.

epicfaace:
I see. Would you be willing to make a PR that upgrades to the latest
version of Beam so that this is fixed?

aniketsharma00411:
Sure

Issue #3034, opened 2020-11-06
Can't use CodaLab with Python > 3.7


epicfaace:
When I try using Python 3.9 and run `pip install codalab`, I get an installation error.

```
ERROR: Command errored out with exit status 1: /Users/epicfaace/.pyenv/versions/3.9.0/bin/python3.9 /Users/epicfaace/.pyenv/versions/3.9.0/lib/python3.9/site-packages/pip install --ignore-installed --no-user --prefix /private/var/folders/_y/j1zg3yt16bx4t6523yj42c480000gn/T/pip-build-env-i1ccxr0h/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- 'cython >= 0.29' 'numpy==1.14.5; python_version<'"'"'3.7'"'"'' 'numpy==1.16.0; python_version>='"'"'3.7'"'"'' setuptools setuptools_scm wheel Check the logs for full command output.
```

This is probably because Apache Beam only supports Python up to 3.7.

epicfaace:
Apache Beam now supports Python 3.8. Hopefully when we upgrade to the latest version, this issue should be gone.

Issue #3029, opened 2020-11-06
How to use python 3.7 or 3.8 (or latest version) instead of 2. ? [USER ISSUE]


mellahysf:
As described in the title, I want to use python 3.7 or 3.8 (or the latest version) instead of python 2.

How to setup that in the Codalab platform? 

Thanks.

yiboliu:
Hi @mellahysf 

So we currently don't support python 3.7, for the reason [here](https://github.com/codalab/codalab-worksheets/pull/3039). I suggest you to use python 3.6. When you are running python command, use python3.6 rather than python. For example: `python codalab_service.py -> python3.6 codalab_service.py`.
Or a more convenient way would be to set python 3.6 as your default python version. To do that, run `echo "alias python=/usr/local/bin/python3.7" >> ~/.bashrc (or .zshrc, whichever you use)`.

yiboliu:
If you still have questions, feel free to reopen this issue and reach out to me.

PR #3778, opened 2021-09-03
Test installing codalab CLI on different python versions on CI

epicfaace:

Test running codalab on different python versions on CI. This will help us catch issues such as https://github.com/codalab/codalab-worksheets/issues/3775, https://github.com/codalab/codalab-worksheets/issues/3034, https://github.com/codalab/codalab-worksheets/issues/3029 and ensure support for Python versions doesn't regress.


adiprerepa (approved):
thank you -- this will help future codalab generations with the headaches that come with python versioning

PR #3782, opened 2021-09-10
Document development tips

epicfaace:

I end up explaining these to new people when onboarding, and they're not really documented, so I'm trying to document them better. This just includes a rough sketch now, but we can probably add more.

@adiprerepa  do you want to give this a pass? Feel free to add additional commands if you think they're useful when developing locally.


epicfaace:
@wwwjn can you please review?

docs/Development-Tips.md
```diff
@@ -0,0 +1,29 @@
+Below is a collection of development tips when running CodaLab locally.
+
```
wwwjn, line 2: For development, use Python3.6 is better. Other higher version of Python3 might brings problem when installing python packages.

wwwjn, line 2: First, start a virtual environment to develop.

virtualenv -p python3.6 venv
. venv/bin/activate

wwwjn:
It's really useful! And I have add something that I think it might be useful.

teetone (approved):

Issue #3631, opened 2021-06-20
"bundles" should be singular when deleting 1 bundle


adiprerepa:
![image](https://user-images.githubusercontent.com/43913660/122681361-b8331b00-d1a8-11eb-9b37-bffda5672140.png)

It should say "1 bundle deleting..."

adiprerepa:
to self: https://github.com/codalab/codalab-worksheets/blob/b61255dc7e5ed63b49c09d19f9c7197d4a1279ff/frontend/src/components/worksheets/Worksheet/Worksheet.js#L283

PR #3784, opened 2021-09-12
[frontend] fix grammar on bundle deletion

adiprerepa:
for https://github.com/codalab/codalab-worksheets/issues/3631


adiprerepa:
I keep messing up git

PR #3808, opened 2021-09-20
Provide default type for schema fields 

### LINKED ISSUE #3599: Provide default type for schema fields
For the schemas: time should default to duration, size should default to size

jzwang43:

Provide default types for some schema paths so it's easier to use.

### Related issues

#3599 

### Screenshots

test
### Checklist

### DIFF
--- a/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
+++ b/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
@@ -116,7 +116,9 @@ class SchemaItem extends React.Component<{
         const schemaHeaders = schemaItem.header;
         let newRow = { from_schema_name: schemaItem.schema_name };
         schemaHeaders.forEach((header) => {
-            newRow[header] = '';
+            // Set post-processor to null for new rows so we can tell if a row has been autofilled.
+            // This also makes new row default values consistent with the first row.
+            newRow[header] = header === 'post-processor' ? null : '';
         });
         let curRow = [...this.state.rows];
         curRow.splice(idx + 1, 0, newRow);
--- a/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
+++ b/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
@@ -118,7 +118,6 @@ class SchemaItem extends React.Component<{
         let newRow = { from_schema_name: schemaItem.schema_name };
         schemaHeaders.forEach((header) => {
             // Set post-processor to null for new, unedited rows to indicate that the row can be autofilled.
-            // This also makes new row default values consistent with the first row.
             newRow[header] = header === 'post-processor' ? null : '';
         });
         let curRow = [...this.state.rows];
--- a/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
+++ b/frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
@@ -143,7 +143,7 @@ class SchemaItem extends React.Component<{
     };
 
     autofillTypes = (row) => {
-        // Only autofill if how to render is null.
+        // Only autofill if the post-processor is null.
         // Null means the post-processor has never been manually changed or autofilled.
         if (row['post-processor'] !== null) {
             return row;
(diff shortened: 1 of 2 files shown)


frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -128,13 +130,38 @@ class SchemaItem extends React.Component<{
...
         this.setState({ rows: [...copiedRows] });
     };
 
+    autofillTypes = (row) => {
+        // Do not fill if how to render is defined.
+        if (row['post-processor'] !== null) {
+            return row;
+        }
+
+        switch (row['generalized-path'].toLowerCase()) {
```
epicfaace: Can we instead use a dictionary with a name like `DEFAULT_POST_PROCESSOR` so it's easier to configure in the future?

jzwang43: Thanks, I added it to constants.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -128,13 +131,29 @@ class SchemaItem extends React.Component<{
...
+            this.autofillTypes({
+                ...rows[idx],
+                [key]: e.target.value.replace(/\n/g, ' ').replace("'", ''),
+            }),
+        );
         this.setState({ rows: [...copiedRows] });
     };
 
+    autofillTypes = (row) => {
+        // Do not fill if how to render is defined.
```
epicfaace: I don't understand this comment.

Can you also explain in the comment what `null` means? does it mean that it needs to be autofilled?

jzwang43: Yeah `null` means the type has never been auto-filled so we can go ahead and auto-fill it. Let me update the comment so it's more clear.

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -128,13 +131,30 @@ class SchemaItem extends React.Component<{
...
+            this.autofillTypes({
+                ...rows[idx],
+                [key]: e.target.value.replace(/\n/g, ' ').replace("'", ''),
+            }),
+        );
         this.setState({ rows: [...copiedRows] });
     };
 
+    autofillTypes = (row) => {
+        // Only autofill if how to render is null.
```
epicfaace:
```suggestion
        // Only autofill if the post-processor is null.
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -116,7 +117,9 @@ class SchemaItem extends React.Component<{
         const schemaHeaders = schemaItem.header;
         let newRow = { from_schema_name: schemaItem.schema_name };
         schemaHeaders.forEach((header) => {
-            newRow[header] = '';
+            // Set post-processor to null for new rows so we can tell if a row has been autofilled.
```
epicfaace:
```suggestion
            // Set post-processor to null for new, unedited rows to indicate that the row can be autofilled.
```

frontend/src/components/worksheets/items/SchemaItem/SchemaItem.js
```diff
@@ -116,7 +117,9 @@ class SchemaItem extends React.Component<{
...
+            // This also makes new row default values consistent with the first row.
```
epicfaace: What does this mean?

jzwang43: We need to set the default post-processor to null so we can tell the difference between initialized value and when user manually deleted the input for post-processor.

epicfaace: > We need to set the default post-processor to null so we can tell the difference between initialized value and when user manually deleted the input for post-processor.

I think that's what the first comment ("Set post-processor to null for new, unedited rows to indicate that the row can be autofilled.") means. I don't understand what the second comment ("This also makes new row default values consistent with the first row.") means. Can you modify it to make it clearer?

jzwang43: > 
> 
> > We need to set the default post-processor to null so we can tell the difference between initialized value and when user manually deleted the input for post-processor.
> 
> I think that's what the first comment ("Set post-processor to null for new, unedited rows to indicate that the row can be autofilled.") means. I don't understand what the second comment ("This also makes new row default values consistent with the first row.") means. Can you modify it to make it clearer?

The old code sometime sets post-processor to `null` and sometimes `''` and it's confusing.
The second line was referring to this issue. But I this the first line of the comment should be enough to explain why `null` is used so I removed the second line.

epicfaace (approved):

PR #3817, opened 2021-09-24
Don't wipe out schema / text / runs when clicking on something else

### LINKED ISSUE #3595: Don't wipe out schema / text / runs when clicking on something else
problem: For schema / text / runs editing, when clicking on something else (refreshed), the schema gets wiped out

jzwang43:

 For schema / text / runs editing, when clicking on something else (refreshed), the schema gets wiped out

### Related issues
#3595 

### Screenshots

a

### Checklist


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -73,6 +73,9 @@ class Worksheet extends React.Component {
             editorEnabled: false, // Whether the editor is actually showing (sometimes lags behind inSourceEditMode)
             showTerminal: false, // Whether the terminal is shown
             focusIndex: -1, // Which worksheet items to be on (-1 is none)
+            newTextIndex: -1, // Which index new text to be on (-1 is at the beginning).
```
epicfaace, line 79: I'm a bit confused as to how these variables fix the issue. Can you explain how they work?

jzwang43: The index stores where the new text/run/schema should be added. Every time an item is focused we update the index. If no item is selected then we add new ones to the top of the worksheet.
Before we didn't have this and the default behavior was just remove the item whenever focus changes.

jzwang43: @epicfaace can I get a review on this, I think it should be ready to go.

frontend/src/components/worksheets/WorksheetItemList.js
```diff
@@ -275,7 +289,7 @@ class WorksheetItemList extends React.Component {
                             }}
                         />
                     )}
-                    {this.props.showNewRun && !focusedItem && (
+                    {this.props.showNewRun && this.props.newRunIndex === -1 && (
```
epicfaace, line 293: Why remove `!focusedItem` here? I see that you kept it below, on line 304

frontend/src/components/worksheets/Worksheet/Worksheet.js
epicfaace: Okay, i see what you're doing, but can you update the comments here to make it clearer what's going on and why these state properties are useful? At the moment, it might be hard for someone new looking at the code for the first time to see what they're used for.

epicfaace (changes requested):

frontend/src/components/worksheets/Worksheet/Worksheet.js
jzwang43: Thanks, I have added more comments to explain the indices.

frontend/src/components/worksheets/WorksheetItemList.js
epicfaace: @jzwang43 I think you accidentally resolved this -- what do you think about this? I'm also not clear why we replace `!focusedItem` with `this.props.newTextIndex === -1`

jzwang43: @epicfaace Sorry I accidentally resolved this.  `focusedItem` should be removed for all of the three types.

jzwang43: @epicfaace can I get a stamp on this, the focusedItem have been removed.

epicfaace: I don't understand the logic here, can we have a quick zoom call to review it live? https://calendly.com/aramaswamis/meeting?month=2021-11

jzwang43: @epicfaace Actually let me test it more tonight and I'll schedule a meeting with you.

epicfaace: Okay, then why is `focusedItem` still set? should we remove it completely then?

yeah, let's discuss this further.

PR #3818, opened 2021-09-25
Add DB changes for multiple bundle stores

adiprerepa:
for multi bundle store

design doc:https://docs.google.com/document/d/1TOsO4gnxUm1XMmlGUb7BJ4Z6y_olFbv9APbx0-IEy6M/edit#heading=h.kh5hv3iuzlpg

### DIFF
--- a/codalab/model/tables.py
+++ b/codalab/model/tables.py
@@ -105,36 +105,27 @@ bundle_store = Table(
         autoincrement=True,
     ),
     Column('uuid', String(63), nullable=False),
-
     # Usually root. For BYO bundle stores (not yet supported),
     # it is the user who created the bundle store.
     Column('owner_id', String(255), nullable=True),
-
     # Name used to refer to this bundle store from the CLI.
     Column('name', String(255), nullable=True),
-
     # Storage type. This is usually redundant with information already in the url column, but set for efficiency reasons. When updating this column, sync it with codalab.common.StorageType.
     Column(
         'storage_type',
         Enum(StorageType.DISK_STORAGE.value, StorageType.AZURE_BLOB_STORAGE.value),
         nullable=True,
     ),
-
     # The format through which the bundle is stored.
     Column(
-        'storage_format',
-        Enum(StorageFormat.UNCOMPRESSED.value, StorageFormat.COMPRESSED_V1.value)
+        'storage_format', Enum(StorageFormat.UNCOMPRESSED.value, StorageFormat.COMPRESSED_V1.value)
     ),
-
     # URL that uniquely identifies the bundle store.
     Column('url', String(255), nullable=True),
-    
     # Authentication / password to access the bundle store.
     Column('authentication', String(255), nullable=True),
-
     # Name of environment variable through which the authentication variable from above can be passed to codalab-service to access the bundle store.
     Column('authentication_env', String(255), nullable=True),
-
     UniqueConstraint('uuid', name='uix_1'),
     Index('bundle_store_name_index', 'name'),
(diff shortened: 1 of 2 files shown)


codalab/model/tables.py
```diff
@@ -93,6 +93,73 @@
...
+    # URL that uniquely identifies the bundle store.
+    Column('url', String(255), nullable=True),
+    
+    # Authentication / password to access the bundle store.
+    Column('authentication', String(255), nullable=True),
+
+    # Name of environment variable through which the authentication variable from above can be passed to codalab-service to access the bundle store.
+    Column('authentication_env', String(255), nullable=True),
+
+    UniqueConstraint('uuid', 'owner_id', name='uix_1'),
```
epicfaace: I think we discussed that we would make it unique on uuid only (instead of `(uuid, owner_id)`)

epicfaace (changes requested):

codalab/model/tables.py
adiprerepa: done

codalab/model/tables.py
```diff
@@ -93,6 +93,73 @@
...
+        autoincrement=True,
+    ),
+    Column('uuid', String(63), nullable=False),
+
+    # Usually root. For BYO bundle stores (not yet supported),
+    # it is the user who created the bundle store.
+    Column('owner_id', String(255), nullable=True),
+
+    # Name used to refer to this bundle store from the CLI.
+    Column('name', String(255), nullable=True),
```
epicfaace: Let's create an index on name and on owner, just like we do for worksheets -- see https://github.com/codalab/codalab-worksheets/blob/7544ef4ace20722881d1d25f7c864341a97617bd/codalab/model/tables.py#L190

epicfaace (changes requested):

codalab/model/tables.py
adiprerepa: done

epicfaace (approved):

codalab/model/tables.py
```diff
@@ -93,6 +93,64 @@
...
+        'id',
+        BigInteger().with_variant(Integer, "sqlite"),
+        primary_key=True,
+        nullable=False,
+        autoincrement=True,
+    ),
+    Column('uuid', String(63), nullable=False),
+    # Usually root. For BYO bundle stores (not yet supported),
+    # it is the user who created the bundle store.
+    Column('owner_id', String(255), nullable=True),
```
epicfaace, line 110: Make this nullable False

codalab/model/tables.py
```diff
@@ -93,6 +93,64 @@
...
+    # Name used to refer to this bundle store from the CLI.
```
epicfaace, line 111: Make `name` nullable False

epicfaace: We should do this to correspond with the worksheet table. Can you also check all other columns with worksheet to ensure they correspond in terms of things being nullable, etc.?

epicfaace (approved):
Feel free to merge after addressing these last comments

codalab/model/tables.py
adiprerepa: owner_id on bundle worksheet is nullable=True

adiprerepa: matched everything with worksheet

epicfaace: Oh, I guess that's kind of strange (having it nullable = True) but let's just keep it that way to be consistent

epicfaace (approved):

Issue #3651, opened 2021-06-29
Frontend blank screen


epicfaace:
This happened after I edited source and then clicked Save.

![image](https://user-images.githubusercontent.com/1689183/123869732-fd68f280-d8f6-11eb-8c9d-adf23809feae.png)

![image](https://user-images.githubusercontent.com/1689183/123869711-f510b780-d8f6-11eb-827e-d21d9f714501.png)

PR #3819, opened 2021-09-27
Fix blank screen for frontend errors

jzwang43:

Blank screen show after edit source and then click save.

This is because error is an object and cannot be directly rendered. Because the error message is just error 404, I added clearer message to explain the error to user.

### Related issues

#3651 

### Screenshots

a

### Checklist


epicfaace (commented):
This seems fine, but does the error object already contain a message that we could show to the user? or is it better to hard-code the error message as we do in this PR?

jzwang43:
> This seems fine, but does the error object already contain a message that we could show to the user? or is it better to hard-code the error message as we do in this PR?

The `error.message` just says something like "error 404" so I think it's better to give some more user-friendly error message.

epicfaace (approved):

PR #3820, opened 2021-09-28
Better formatting of disk quota error message

### LINKED ISSUE #3130: Better formatting of disk quota error message
When you use `cl upload` to upload a file whose size exceeds your disk quota, the error message just gives the number of bytes in the disk quota / bundle size. Instead, we should format it (such as `100g`), just like we do all other file sizes on CodaLab when presented to the user.

```
(venv) epicfaace@ashwins-mbp codalab-worksheets % cl upload ~/codalab/largefile
Preparing upload archive...
Uploading largefile (0xc1e0c4f5b21344118080e68cd934e381) to http://localhost
Sent 102400.00MiB / 102400.00MiB (100.0%) [18.51MiB/sec]        		
UsageError: Upload failed: Can't save bundle, bundle size 107374182400 greater than user's disk quota left: 101919140000.0
```

andyjin2000:

When you use cl upload to upload a file whose size exceeds your disk quota, the error message just gives the number of bytes in the disk quota / bundle size. Instead, we should format it (such as 100g), just like we do all other file sizes on CodaLab when presented to the user.

### Related issues

fixes https://github.com/codalab/codalab-worksheets/issues/3130


epicfaace (commented):
Let's use formatting.size_str (https://github.com/codalab/codalab-worksheets/blob/953b3705d02c27800b0aebfc6da00d52799ad79e/codalab/lib/formatting.py#L34) instead because I think that's used everywhere else in the codebase to format disk quota (see https://github.com/codalab/codalab-worksheets/blob/953b3705d02c27800b0aebfc6da00d52799ad79e/codalab/lib/bundle_cli.py#L3931).

Also, can you update this test (https://github.com/codalab/codalab-worksheets/blob/953b3705d02c27800b0aebfc6da00d52799ad79e/tests/cli/test_cli.py#L749) to ensure that the formatting of the error message is correct?

epicfaace:
> Let's use formatting.size_str (

@andyjin2000 Can you update this PR to use formatting.size_str instead? And can you push the test you made earlier to this PR so we can see how to move forward with this PR / debug further?

epicfaace (approved):

PR #3822, opened 2021-10-05
Move away from using an indexed_gzip fork

### LINKED ISSUE #3535: Move away from using an indexed_gzip fork
Move away from using an indexed_gzip fork, which we had to use in https://github.com/codalab/codalab-worksheets/pull/3496 in order to fix https://github.com/codalab/codalab-worksheets/issues/3514. Once the upstream repository is updated to address the trailing null bytes bug, we can switch back to using indexed_gzip instead of indexed_gzip_fileobj_fork_epicfaace again.

Just waiting for https://github.com/pauldmccarthy/indexed_gzip/pull/72 to be merged / released and we can then switch back to the upstream repo.

epicfaace:
fixes #3535


epicfaace:
@adiprerepa can you take a look when you get a chance?

teetone (approved):

Issue #3800, opened 2021-09-15
Create new bundle store REST endpoints


epicfaace:
Part of this megaissue: https://github.com/codalab/codalab-worksheets/issues/3799
Design doc: https://docs.google.com/document/d/1TOsO4gnxUm1XMmlGUb7BJ4Z6y_olFbv9APbx0-IEy6M/edit#

---

GET /bundle_stores/
List bundle stores available to (owned by) the user.

POST /bundle_stores/
Add a new bundle store.

PUT /bundle_store/{id}
Update a bundle store.

GET /bundle_store/{id}
Get a bundle store.

DELETE /bundle_store/{id}
Delete a bundle store (you can’t delete a bundle store unless there are no BundleLocations associated with it)

PR #3850, opened 2021-10-20
[Multi Bundle Store] Create new bundle store endpoints

adiprerepa:
Creates new endpoints for bundle stores (create, update, etc)

for https://github.com/codalab/codalab-worksheets/issues/3800


codalab/model/bundle_model.py
```diff
@@ -30,6 +30,8 @@
     bundle as cl_bundle,
     bundle_dependency as cl_bundle_dependency,
     bundle_metadata as cl_bundle_metadata,
+    bundle_store as cl_bundle_store,
```
epicfaace: Make sure you run `./pre-commit.sh` to fix format

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
     return dict(data=uuids_to_locations)
 
 
+@get('/bundle_stores/', apply=AuthenticatedProtectedPlugin())
+def _fetch_bundle_stores():
+    """
+    Fetch the bundle stores available to the user.
+
+    Returns a dictionary in which the keys are the bundle store uuids, and the values
```
epicfaace: Is this the standard format we use in other REST API endpoints to return multiple items?

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
     return dict(data=uuids_to_locations)
 
 
+@get('/bundle_stores/', apply=AuthenticatedProtectedPlugin())
```
epicfaace: Maybe remove the trailing slash to be consistent 
```suggestion
@get('/bundle_stores', apply=AuthenticatedProtectedPlugin())
```

Do this for all the new endpoints

Also test to see if `/bundle_stores` and `/bundle_stores/` both work after you remove the trailing slash.

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+    Returns a dictionary in which the keys are the bundle store uuids, and the values
+    are tuples with the owner_id, name, and url.
+    """
+    return local.model.get_bundle_stores(request.user)
+
+
+@post('/bundle_stores/', apply=AuthenticatedProtectedPlugin())
+def _add_bundle_store():
+    """
+    Add a bundle store that the user can access.
```
epicfaace, line 493: Document the parameters.

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+    """
+    return local.model.create_bundle_store(
+        request.user,
+        request.forms.get('name'),
```
epicfaace: Don't use `request.forms` because that would use form bodies. We use JSON bodies everywhere else. See https://stackoverflow.com/questions/23118249/whats-the-difference-between-request-payload-vs-form-data-as-seen-in-chrome for more info on the difference.

Let's use `request.json` and create a marshmallow schema to validate input, similar to `BundlePermissionSchema` (see https://github.com/codalab/codalab-worksheets/blob/6cf9a3159c3b332d173c521ca516ed69247fe2f6/codalab/rest/bundles.py#L406 for an example)

epicfaace: Also do this for the update bundle store endpoint

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+        request.user,
+        request.forms.get('name'), 
+        request.forms.get('storage_type'), 
+        request.forms.get('storage_format'), 
+        request.forms.get('url'), 
+        request.forms.get('authentication')
+    )
+
+
+@put('/bundle_stores/<id:re:%s>', apply=AuthenticatedProtectedPlugin())
```
```suggestion
@put('/bundle_stores/<uuid:re:%s>', apply=AuthenticatedProtectedPlugin())
```

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+def _update_bundle_store(uuid):
+    """
+    Update a bundle store that the user can access.
```
epicfaace, line 517: Document the parameters.

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+    """
+    Fetch the bundle store corresponding to the specified id.
+
+    Returns a dictionary in which the key id the bundle store uuid, and the value
+    is a tuple with the owner_id, name, and url.
+    """
+    return local.model.get_bundle_store(request.user, uuid)
+
+
+@delete('/bundle_stores/<id:re:%s>', apply=AuthenticatedProtectedPlugin())
```
epicfaace: id -> uuid

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+        uuid,
+        request.forms.get('name'), 
+        request.forms.get('storage_type'), 
+        request.forms.get('storage_format'), 
+        request.forms.get('url'), 
+        request.forms.get('authentication')
+    )
+
+
+@get('/bundle_stores/<id:re:%s>', apply=AuthenticatedProtectedPlugin())
```
epicfaace: id -> uuid

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+def _fetch_bundle_store(uuid):
+    """
+    Fetch the bundle store corresponding to the specified id.
```
```suggestion
    Fetch the bundle store corresponding to the specified uuid.
```

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+    Returns a dictionary in which the key id the bundle store uuid, and the value
+    is a tuple with the owner_id, name, and url.
+    """
+    return local.model.get_bundle_store(request.user, uuid)
+
+
+@delete('/bundle_stores/<id:re:%s>', apply=AuthenticatedProtectedPlugin())
+def _delete_bundle_store(uuid):
+    """
+    Delete the bundle store that the user can access, only if there is an associated BundleLocation.
```
epicfaace: This is the opposite of what should be there -- the bundle store can't be deleted _unless_ there are no associated bundlelocations -- update the comment

codalab/rest/bundles.py
```diff
@@ -426,6 +425,67 @@ def _fetch_locations():
...
+    )
+
+
+@get('/bundle_stores/<id:re:%s>', apply=AuthenticatedProtectedPlugin())
+def _fetch_bundle_store(uuid):
+    """
+    Fetch the bundle store corresponding to the specified id.
+
+    Returns a dictionary in which the key id the bundle store uuid, and the value
+    is a tuple with the owner_id, name, and url.
```
epicfaace: Check `_fetch_bundle` to see how its output is returned, and return the output in a similar way.

Goal: we want all rest endpoints to return output in a predictable manner.

epicfaace (commented):
Try to mimic the structure of these endpoints as closely as possible for the return values:

https://worksheets.codalab.org/rest/bundles?keywords=test - /bundle_stores/
https://worksheets.codalab.org/rest/bundles/0xcd4b74497cff47db84e98d4daf8488c6 - /bundle_store/<uuid>/

codalab/model/bundle_model.py
```diff
@@ -30,6 +30,8 @@
...
+    bundle_location as cl_bundle_location,
     group as cl_group,
```
epicfaace: Add tests for the new endpoints in a file similar to https://github.com/codalab/codalab-worksheets/blob/bundle-store-rest-endpoints/tests/unit/rest/bundles_test.py#L6

codalab/model/bundle_model.py
```diff
@@ -2901,3 +2903,112 @@ def delete_oauth2_auth_code(self, auth_code_id):
...
+                        cl_bundle_store.c.url,
+                    ]
+                ).where(
+                    or_(
+                        cl_bundle_store.c.owner_id == 'codalab(0)',
+                        cl_bundle_store.c.owner_id == user,
+                    )
+                )
+            ).fetchall()
+            logger.error(rows)
```
epicfaace: rm this line

codalab/model/bundle_model.py
```diff
@@ -2901,3 +2903,112 @@ def delete_oauth2_auth_code(self, auth_code_id):
...
+                        cl_bundle_store.c.uuid,
+                        cl_bundle_store.c.owner_id,
+                        cl_bundle_store.c.name,
+                        cl_bundle_store.c.storage_type,
+                        cl_bundle_store.c.storage_format,
+                        cl_bundle_store.c.url,
+                    ]
+                ).where(
+                    or_(
+                        cl_bundle_store.c.owner_id == 'codalab(0)',
```
epicfaace: Document this in a comment that it returns all bundles owned by the root user or current user

codalab/model/bundle_model.py
```diff
@@ -2901,3 +2903,112 @@ def delete_oauth2_auth_code(self, auth_code_id):
...
+                'owner_id': user,
+                'name': name,
+                'storage_type': storage_type,
+                'storage_format': storage_format,
+                'url': url,
+                'authentication': authentication,
+            }
+            connection.execute(
+                cl_bundle_store.update()
+                .where(cl_bundle_store.c.uuid == id)
```
epicfaace: This is a problem because anyone can update any bundle store. We need to add a check to ensure that a user can only update a bundle store they own.

codalab/model/bundle_model.py
```diff
@@ -2901,3 +2903,112 @@ def delete_oauth2_auth_code(self, auth_code_id):
...
+                'url': row.url,
+            }
+
+    def delete_bundle_store(self, user, id):
+        """
+        return the bundle store corresponding to the specified id
+        """
+        with self.engine.begin() as connection:
+            bundle_store_row = connection.execute(
+                select([cl_bundle_store.c.uuid]).where(cl_bundle_store.c.uuid == id)
```
epicfaace: Same security problem where user can delete any bundle store

codalab/model/bundle_model.py
```diff
@@ -2901,3 +2903,112 @@ def delete_oauth2_auth_code(self, auth_code_id):
...
+            logger.error(row)
+            return {
+                'owner_id': row.owner_id,
+                'name': row.name,
+                'url': row.url,
+            }
+
+    def delete_bundle_store(self, user, id):
+        """
+        return the bundle store corresponding to the specified id
```
epicfaace: update this comment

codalab/rest/bundles.py
adiprerepa: I mean, does it have to be the same format? I think a mapping of uuid->object is pretty intuitive/good

codalab/rest/bundles.py
adiprerepa: done

codalab/rest/bundles.py
epicfaace: Yes, we want consistency in our REST API endpoints, so I think that's what we should aim for.

codalab/rest/schemas.py
```diff
@@ -220,6 +220,18 @@ class BundleSchema(Schema):
     class Meta:
         type_ = 'bundles'
 
+class BundleStoreSchema(Schema):
+    id = fields.String(validate=validate_uuid, attribute='uuid')
+    uuid = fields.String(attribute='uuid')
+    owner = fields.String(attribute='owner_id')
+    name = fields.String()
+    storage_type = fields.String()
+    storage_format = fields.String(validate=validate.OneOf({'UNCOMPRESSED', 'COMPRESSED_V1'}))
```
epicfaace: Let's just do this for now

```suggestion
    storage_format = fields.String()
```

so we only need to maintain the enum list in one place, which is the db schema

tests/cli/test_cli.py
```diff
@@ -599,7 +599,7 @@ def test_localhost(ctx):
 @TestModule.register('unittest')
 def test_unittest(ctx):
     """Run backend unit tests."""
-    _run_command(['coverage', 'run', '--rcfile=tests/unit/.coveragerc', '-m', 'nose', 'tests.unit'])
+    _run_command(['coverage', 'run', '--rcfile=tests/unit/.coveragerc', '-m', 'nose', 'tests.unit.rest'])
```
epicfaace: revert this

epicfaace (commented):
One other test you could do is something similar to https://github.com/codalab/codalab-worksheets/pull/3899/files#diff-9b8ded0d95380fbcd0b71e371298b0b5c77df7e40b12de58fcbb29f4d76fc045

andyjin2000:
Perhaps also add documentation and type annotations for the function args in bundle_model? Otherwise lgtm; will let @epicfaace have a final look at the updates before accepting!

codalab/model/bundle_model.py
epicfaace: We can't assume the root user is called "codalab" or has id "0". Also, isn't owner_id supposed to be an int and not a string "codalab(0)"?

You should use the root user id configured from here: https://github.com/codalab/codalab-worksheets/blob/be5d92d038471da59eee1a493c60f346a6b94188/codalab/lib/codalab_manager.py#L403

Also I'd test this out to double check it actually gets the root user's bundle.

epicfaace: Just use `self.root_user_id` which is set to the root user's id.

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
-    def save_oauth2_auth_code(self, grant):
-        with self.engine.begin() as connection:
-            result = connection.execute(oauth2_auth_code.insert().values(grant.columns))
-            grant.id = result.lastrowid
-        return grant
-
-    def delete_oauth2_auth_code(self, auth_code_id):
+    def update_bundle_store(self, user, uuid, update_fields):
+        """
+        update a bundle store
```
```suggestion
        Update a bundle store
```

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
+        """
         with self.engine.begin() as connection:
             connection.execute(
-                oauth2_auth_code.delete().where(oauth2_auth_code.c.id == auth_code_id)
+                cl_bundle_store.update()
+                .where(and_(cl_bundle_store.c.uuid == uuid, cl_bundle_store.c.owner_id == user,))
```
epicfaace: What is `user`? Should we rename it to `user_id`? Can you also rename other parameters called `user` appropriately in other methods as well?

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
         return uuid
 
-    def save_oauth2_auth_code(self, grant):
-        with self.engine.begin() as connection:
-            result = connection.execute(oauth2_auth_code.insert().values(grant.columns))
-            grant.id = result.lastrowid
-        return grant
-
-    def delete_oauth2_auth_code(self, auth_code_id):
+    def update_bundle_store(self, user, uuid, update_fields):
```
epicfaace: Add type annotations to all new parameters. Can you also add it to create_bundle_store above?

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
-                oauth2_auth_code.delete().where(oauth2_auth_code.c.id == auth_code_id)
+                cl_bundle_store.update()
+                .where(and_(cl_bundle_store.c.uuid == uuid, cl_bundle_store.c.owner_id == user,))
+                .values(update_fields)
             )
 
+    def get_bundle_store(self, user, uuid):
+        """
+        Return the bundle store corresponding to the specified uuid.
+        Just returns the owner_id, name, and url of the bundle store.
```
epicfaace: This line is not correct, we return more. Let's just remove this line?

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
+                        cl_bundle_store.c.name,
+                        cl_bundle_store.c.storage_type,
+                        cl_bundle_store.c.storage_format,
+                        cl_bundle_store.c.url,
+                    ]
+                ).where(
+                    and_(
+                        cl_bundle_store.c.uuid == uuid,
+                        or_(
+                            cl_bundle_store.c.owner_id == 'codalab(0)',
```
epicfaace: same issue with root user as mentioned above

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
+                'owner_id': row.owner_id,
+                'name': row.name,
+                'url': row.url,
+                'storage_type': row.storage_type,
+                'storage_format': row.storage_format,
+            }
+
+    def delete_bundle_store(self, user, uuid):
+        """
+        Delete a bundle store given its uuid. We can only delete the Bundle Store if there are
```
```suggestion
        Delete a bundle store given its uuid. We can only delete the bundle store if there are
```

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
             connection.execute(
-                oauth2_auth_code.delete().where(oauth2_auth_code.c.id == auth_code_id)
+                cl_bundle_store.update()
+                .where(and_(cl_bundle_store.c.uuid == uuid, cl_bundle_store.c.owner_id == user,))
+                .values(update_fields)
             )
 
+    def get_bundle_store(self, user, uuid):
+        """
+        Return the bundle store corresponding to the specified uuid.
```
epicfaace, line 3000: Can you update all docstrings to document argument names under an  "Arguments:" heading, such as in https://github.com/codalab/codalab-worksheets/blob/d83b8b81e735a6789feda92255654e35311c68ca/codalab/worker/worker.py#L513?

codalab/model/bundle_model.py
```diff
@@ -2910,18 +2961,71 @@ def create_bundle_store(self, user, name, storage_type, storage_format, url, aut
...
+                select([cl_bundle_store.c.uuid]).where(
+                    and_(cl_bundle_store.c.uuid == uuid, cl_bundle_store.c.owner_id == user,)
+                )
+            ).fetchone()
+            bundle_location_row = connection.execute(
+                select([cl_bundle_location.c.id]).where(
+                    cl_bundle_location.c.bundle_store_uuid == bundle_store_row.uuid
+                )
+            ).fetchone()
+            # delete only if there is BundleLocation associated
```
epicfaace: I'm confused by this comment, I thought we delete bundle stores only if there are no BundleLocations associated with it?

codalab/rest/bundles.py
```diff
@@ -469,6 +470,65 @@ def _fetch_bundle_location(bundle_uuid: str, bundle_store_uuid: str):
...
+    are tuples with the owner_id, name, and url.
+    """
+    bundle_stores = local.model.get_bundle_stores(request.user)
+    return BundleStoreSchema(many=True).dump(bundle_stores).data
+
+
+@post('/bundle_stores', apply=AuthenticatedProtectedPlugin())
+def _add_bundle_store():
+    """
+    Add a bundle store that the user can access. Request body must contain the fields in BundleStoreSchema.
```
epicfaace: We can't just reference "BundleStoreSchema" in these docstrings, because these docstrings are directly copied over to the REST API reference and the user won't know what to look at. We should document each parameter, and it's okay if it ends up being duplicative across docstrings.

codalab/rest/bundles.py
epicfaace: @adiprerepa

codalab/rest/bundles.py
```diff
@@ -469,6 +470,65 @@ def _fetch_bundle_location(bundle_uuid: str, bundle_store_uuid: str):
...
+    is a tuple with the owner_id, name, and url.
+    """
+    bundle_store = local.model.get_bundle_store(request.user, uuid)
+    return BundleStoreSchema().dump(bundle_store).data
+
+
+@delete('/bundle_stores/<uuid:re:%s>' % spec_util.UUID_STR, apply=AuthenticatedProtectedPlugin())
+def _delete_bundle_store(uuid):
+    """
+    Delete the bundle store that the user can access. Note that you can’t delete a bundle store
```
```suggestion
    Delete the specified bundle store. Note that you can't delete a bundle store
```

tests/unit/server/bundle_store_test.py
```diff
@@ -3,6 +3,41 @@
 
 
 class BundleStoreTest(BaseBundleManagerTest):
+    def test_add_bundle_store(self):
```
epicfaace: This test tests a bunch of endpoints, but you called it "test_add_bundle_store" and said "Tests the workflow for creating bundles" in the docstring. Can you update both so it better reflects what this test does?

epicfaace (commented):
Looks good, just a few more changes and then let's merge.

codalab/model/bundle_model.py
adiprerepa: sorry, whats the issue?

epicfaace: I said in the other comment:

```
We can't assume the root user is called "codalab" or has id "0". Also, isn't owner_id supposed to be an int and not a string "codalab(0)"?

You should use the root user id configured from here:

codalab-worksheets/codalab/lib/codalab_manager.py

Line 403 in be5d92d

 def root_user_id(self): 
Also I'd test this out to double check it actually gets the root user's bundle.

 
Member
@epicfaace epicfaace 22 hours ago
Just use self.root_user_id which is set to the root user's id.
```

epicfaace (approved):

PR #3877, opened 2021-11-09
[frontend] Handling 403 error

### LINKED ISSUE #3756: [USER ISSUE] 403 screen is blank
**Describe the bug**
When you try to view a bundle that is private, the screen goes blank.

chrome log traceback:

```
Failed to load resource: the server responded with a status of 403 (Forbidden)
instrument.ts:129 Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=Error%3A%20Request%20failed%20with%20status%20code%20403&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Eo (react-dom.production.min.js:139)
    at react-dom.production.min.js:149
    at Pa (react-dom.production.min.js:173)
    at vu (react-dom.production.min.js:265)
    at ls (react-dom.production.min.js:246)
    at us (react-dom.production.min.js:246)
    at Ju (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Ui (react-dom.production.min.js:122)
    at Gi (react-dom.production.min.js:123)
    at $i (react-dom.production.min.js:122)
    at Yu (react-dom.production.min.js:230)
    at Object.enqueueSetState (react-dom.production.min.js:132)
    at a._.setState (react.production.min.js:12)
    at e (Bundle.js:68)
(anonymous) @ instrument.ts:129
Bundle.js:72 Uncaught (in promise) TypeError: Cannot read property 'status' of undefined
    at t (Bundle.js:72)
o431750.ingest.sentry.io/api/5383515/store/?sentry_key=a9654419762745c3b255bf299f28a656&sentry_version=7:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
```

**To Reproduce**
1. Make a private bundle (e.g., `0x6ffce9a9d4244f0a931dae1822ecc759`)
2. go to https://worksheets.codalab.org/bundles/0x6ffce9a9d4244f0a931dae1822ecc759
3. see blank screen

pranavjain:

Currently, users are not shown proper errors if appropriate bundle permissions are not provided (403).

### Related issues
fixes #3756 

### Screenshots
Codalab-403

### Checklist


frontend/src/components/Bundle/Bundle.js
```diff
@@ -136,6 +145,20 @@ class Bundle extends React.Component<
     /** Renderer. */
     render() {
         const bundleInfo = this.state.bundleInfo;
+
+        // Below block can be used to map any errorCode to custom message for user
+        if (this.state.errorCode == 403) {
+            return (
+                <ErrorMessage
+                    message={
+                        "You don't have permission to access: '/bundles/" + this.props.uuid + "'"
```
epicfaace:
```suggestion
                        "Permission denied: '/bundles/" + this.props.uuid + "'"
```

frontend/src/components/Bundle/Bundle.js
```diff
@@ -136,6 +145,20 @@ class Bundle extends React.Component<
     /** Renderer. */
     render() {
         const bundleInfo = this.state.bundleInfo;
+
+        // Below block can be used to map any errorCode to custom message for user
```
epicfaace:
```suggestion
        // Show custom messages based on error code
```

epicfaace: We should add this entire block into the `if (!bundleInfo) {` block, for consistency with the other error handling logic in this component

epicfaace (changes requested):

frontend/src/components/Bundle/Bundle.js
pranavjain: I did that earlier but decided to keep it outside the `if` block because we do get some info in `bundleInfo` variable at times and error is not triggered in those cases.

jzwang43: I think this PR looks good, just update the comment as suggested and it should be good to go.

epicfaace: > I did that earlier but decided to keep it outside the `if` block because we do get some info in `bundleInfo` variable at times and error is not triggered in those cases.

When would `bundleInfo` be non-null and `this.state.errorCode` be 403/404? I would have thought `bundleInfo` is supposed to be null in all those cases.

pranavjain: Even in the case of 403 error we do get some content in bundleInfo. I am making the changes in the comment as per your suggestion. We can merge this one and test on dev to see if it works fine.

pranavjain: > I think this PR looks good, just update the comment as suggested and it should be good to go.

Done! Please +1 if it's good to go from your side @jzwang43

epicfaace: @pranavjain 
> Even in the case of 403 error we do get some content in bundleInfo

Got it, I'm just asking if you understand why that is the case -- what is in bundleInfo when we get a 403 error? I'm trying to see if this is expected behavior or another bug that we should file.

pranavjain: I think there is another bug in the code. We can follow up on that one separately. I will investigate and file a separate issue. Let's close this one. Thanks!

epicfaace: Ok

epicfaace (approved):

PR #3882, opened 2021-11-10
Support bundle checkpoint / preemptible workers

### LINKED ISSUE #3874: Bundle manager changes required for model checkpoints
As per https://docs.google.com/document/d/1ifcuXLPldSWJehZgl9RTmPogllbyUJ_vRIOiW2BF1i0/edit#

epicfaace:
For documentation on how to use this feature, see: https://github.com/codalab/codalab-worksheets/blob/6423cc3ba1e6d9779224c6c66e97b59a9c208197/docs/Checkpoints.md

Original design doc: https://docs.google.com/document/d/1ifcuXLPldSWJehZgl9RTmPogllbyUJ_vRIOiW2BF1i0/edit#

Checklist:

- [x] Add --preemptible flag to cl-worker and --worker-preemptible flag to cl-worker-manager
- [x] Update bundle manager logic so that: - Ashwin Ramaswami
  - Worker checkins include the “preemptible” field
  - A bundle in worker_offline that is assigned to a preemptible worker will be moved to staged
- [x] Add tests in test_cli.py to ensure this logic works


codalab/model/bundle_model.py
```diff
@@ -915,7 +915,15 @@ def transition_bundle_preparing(self, bundle, user_id, worker_id, start_time, re
 
             bundle_update = {
                 'state': State.PREPARING,
-                'metadata': {'started': start_time, 'last_updated': start_time, 'remote': remote},
+                'metadata': {
+                    'started': start_time,
+                    'last_updated': start_time,
+                    'remote': remote,
+                    'remotes': (bundle.metadata.remotes or [])
```
percyliang: `remotes` is defined as previous, which suggests it shouldn't include `remote`.  Perhaps this should be named `previous_remotes`?

codalab/bundles/run_bundle.py
```diff
@@ -69,7 +69,9 @@ class RunBundle(DerivedBundle):
     METADATA_SPECS.append(MetadataSpec('exitcode', int, 'Exitcode of the process.', generated=True))
     METADATA_SPECS.append(MetadataSpec('job_handle', str, 'Identifies the job handle (internal).', generated=True, hide_when_anonymous=True))
     METADATA_SPECS.append(MetadataSpec('remote', str, 'Where this job is/was run (internal).', generated=True, hide_when_anonymous=True))
-    METADATA_SPECS.append(MetadataSpec('remotes', list, 'Previous workers where this job was run (internal); multiple values indicate that the bundle was preempted and moved to a different worker.', generated=True, hide_when_anonymous=True))
+    METADATA_SPECS.append(MetadataSpec('remotes', list, 'Previous workers where this job was run (internal); multiple values indicate that the bundle was preempted and moved to a different worker.',
```
percyliang: Remote names aren't verified / stable names, so I'm worried about depending on them for anything serious.

codalab/bundles/run_bundle.py
```diff
@@ -69,7 +69,9 @@ class RunBundle(DerivedBundle):
...
+    generated=True, hide_when_anonymous=True))
+    METADATA_SPECS.append(MetadataSpec('preemptible', bool, 'Whether the bundle is currently running / finished on a preemptible worker.', generated=True, hide_when_anonymous=True, default=False))
```
percyliang: I thought `preemptible` would just specify whether we want to restart the job again if the worker died, but the description here seems to be something else...

codalab/bundles/run_bundle.py
epicfaace: We don't actually depend on them -- this field is not used otherwise at all. I just thought it would be good to keep track of which remotes it ran on for debugging purposes (as opposed to not collecting this info at all). Ideally we would store a list of {preemption time, remote name} -- but how useful do you think this is? Maybe we don't need to keep it at all?

codalab/model/bundle_model.py
```diff
@@ -987,6 +995,8 @@ def transition_bundle_worker_offline(self, bundle):
         Transitions bundle to WORKER_OFFLINE state:
             Updates the last_updated metadata.
             Removes the corresponding row from worker_run if it exists.
+
+        If the worker is preemptible, move the bundle to the STAGED state instead.
```
```suggestion
        If the bundle is preemptible, move the bundle to the STAGED state instead.
```

codalab/model/bundle_model.py
```diff
@@ -915,7 +915,15 @@ def transition_bundle_preparing(self, bundle, user_id, worker_id, start_time, re
...
                 'state': State.PREPARING,
-                'metadata': {'started': start_time, 'last_updated': start_time, 'remote': remote},
+                'metadata': {
+                    'started': start_time,
+                    'last_updated': start_time,
+                    'remote': remote,
+                    'remotes': (bundle.metadata.remotes or [])
+                    + [
+                        remote
+                    ],  # Store the history of which workers ran this bundle before in the bundle metadata.
```
epicfaace, line 931: Store worker ID, time

codalab/model/bundle_model.py
epicfaace: Renamed it to `remote_history` instead. It should include the current remote as well.

codalab/bundles/run_bundle.py
epicfaace: Just going to store list of remote names for now. At the very least, even if the names aren't useful, we'll know how many times the bundle was preempted.

codalab/bundles/run_bundle.py
epicfaace: No -- `preemptible` means that the bundle is currently running / finished on a preemptible worker. The user cannot set it and it's only used internally by the bundle manager to handle bundle state.

The way we target bundles to run on preemptible workers is to target the bundles using `--tag` to a tag that specifies a preemptible worker pool. The preemptible workers should be started using `--tag` and `--tag-exclusive`.

epicfaace: @percyliang 
I've added documentation explaining this better:

image

percyliang: I think `on_preemptible_worker` might be a more accurate description, since bundles are not preemptible, workers are.  What do you think?  This is a flag that's automatically set by CodaLab, not by the user.

codalab/model/bundle_model.py
```diff
@@ -1000,15 +1010,20 @@ def transition_bundle_worker_offline(self, bundle):
                 # The user deleted the bundle or the bundle finished
                 return False
 
+            if getattr(bundle.metadata, "preemptible", False):
```
percyliang: Why can't we access the preemptible field directly and have to use `getattr`?

codalab/model/bundle_model.py
```diff
@@ -1000,15 +1010,20 @@ def transition_bundle_worker_offline(self, bundle):
...
+                bundle_update = {
```
percyliang, line 1023: Put a comment on what this logic, since this is the core thing.

docs/Checkpoints.md
```diff
@@ -0,0 +1,14 @@
...
+!!! warning "Unreleased feature"
+        This feature is still in development. It does not fully work yet and will only work once [#3710](https://github.com/codalab/codalab-worksheets/issues/3710) is resolved.
+
+CodaLab supports the concept of bundle checkpoints, which allow bundles to be stopped and then resumed on different workers. This is especially useful if you want to use preemptible compute nodes, such as [EC2 Spot](https://aws.amazon.com/ec2/spot/) instances. When a worker is preempted, the bundle should be resumed on another worker.
+
+## Workflow steps
+
+This section describes a sample workflow for running bundles on a pool of preemptible workers. The workflow uses the tag `preemptiblepool1` to organize compute, but you can change this tag name as needed.
+
+1. Start a worker manager that runs workers on preemptible compute notes, using the following flags: `--preemptible --worker-tag-exclusive --worker-tag="preemptiblepool1"`. Ensure that all the workers started by the worker manager are configured to share the same network disk for their bundle run working directory / dependency cache.
```
percyliang: Should be `--worker-preemptible`

percyliang (commented):
I might have missed it, but I didn't see any logic on the worker side on how bundles are resumed.  How will the resumed bundle be run in the same working directory as its previous incarnation?

I do think that the bundle needs a `preemptible` flag, because in some cases, you don't want your job to be resumed from a partially completed working directory because you'll get invalid results.  In those cases, the bundle should not be restaged.  Otherwise, I'm afraid that we're going to get into some bad infinite loops.

epicfaace:
TODO @epicfaace -- add a `--preemptible` flag: https://docs.google.com/document/d/1ifcuXLPldSWJehZgl9RTmPogllbyUJ_vRIOiW2BF1i0/edit?disco=AAAAUURsO9U&usp=comment_email_document&ts=61e76abb&usp_dm=false

codalab/model/bundle_model.py
epicfaace: I'm worried that if we do `bundle.metadata["preemptible"]`, this will break bundles that started running in the previous deploy (but continue running during the deploy) that don't yet have this metadata key yet.

codalab/model/bundle_model.py
epicfaace, line 1023:
```suggestion
                # If the bundle is preemptible, move the bundle to the STAGED state instead.
                bundle_update = {
```

docs/Checkpoints.md
```diff
@@ -0,0 +1,14 @@
...
+        This feature is still in development. It does not fully work yet and will only work once [#3710](https://github.com/codalab/codalab-worksheets/issues/3710) is resolved.
+
+CodaLab supports the concept of bundle checkpoints, which allow bundles to be stopped and then resumed on different workers. This is especially useful if you want to use preemptible compute nodes, such as [EC2 Spot](https://aws.amazon.com/ec2/spot/) instances. When a worker is preempted, the bundle should be resumed on another worker.
+
+## Workflow steps
+
+This section describes a sample workflow for running bundles on a pool of preemptible workers. The workflow uses the tag `preemptiblepool1` to organize compute, but you can change this tag name as needed.
+
+1. Start a worker manager that runs workers on preemptible compute notes, using the following flags: `--worker-preemptible --worker-tag-exclusive --worker-tag="preemptiblepool1"`. Ensure that all the workers started by the worker manager are configured to share the same network disk for their bundle run working directory / dependency cache.
+1. User can run a bundle by running `cl run --tag="preemptiblepool1"`.
```
epicfaace:
```suggestion
1. User can run a bundle by running `cl run --preemptible --tag="preemptiblepool1"`.
```

epicfaace: todo

codalab/bundles/run_bundle.py
epicfaace: This flag is actually set by the user. We wanted users to explicitly specify `cl run --preemptible` so they are aware that their command must work on preemptible workers. For that reason, I'm going to change it back to `preemptible` instead of `on_preemptible_worker` (because `--on_preemptible_worker` is awkward to type).

epicfaace: Never mind, I just removed `--preemptible` entirely from `cl run` and so we can keep calling it `on_preemptible_worker`.

tests/cli/test_cli.py
```diff
@@ -1053,6 +1054,26 @@ def fetch_contents_blob_from_web_browser(uuid):
...
+    wait_until_state(uuid, State.RUNNING)
+    remote_preemptible_worker = get_info(uuid, 'remote')
+    check_equals("True", get_info(uuid, 'on_preemptible_worker'))
+    # Kill worker
+    client = docker.from_env()
+    client.containers.get("codalab_worker-preemptible_1").kill()
+    # Wait for bundle to be re-assigned
+    wait_until_state(uuid, State.STAGED)
+    wait_until_state(uuid, State.RUNNING)
+    # bundle should be resumed on the other worker
```
epicfaace: consider testing to make sure it reuses the same work directory

.github/workflows/test.yml
```diff
@@ -308,6 +308,55 @@ jobs:
...
+      - name: Setup tests
+        run: |
+          sudo service mysql stop
+          python3 codalab_service.py build services --version ${VERSION} --pull
+        env:
+          VERSION: ${{ github.head_ref || 'master' }}
+      - name: Run tests
+        run: |
+          python3 codalab_service.py start --services default worker-preemptible --version ${VERSION} -d
+          python3 test_runner.py --version ${VERSION} ${TEST}
```
epicfaace:
```suggestion
          python3 codalab_service.py start --services default worker-preemptible --version ${VERSION} -d
          # keep running "cl status" until we have one bundle running with the "--preemptible" tag, and then kill the worker
          python3 test_runner.py --version ${VERSION} ${TEST}
```

epicfaace:
@percyliang @teetone One more thing -- with this PR, the bundle stats (such as time taken) only include the time for the last worker that ran a bundle (rather than the total time across all workers that ran a bundle). We might want to update that in the future, though, as that's just nice-to-have functionality that could take longer to implement.

epicfaace:
@teetone @wwwjn tests pass now, this is ready for review.

codalab/worker/worker_run_state.py
```diff
@@ -344,8 +344,11 @@ def mount_dependency(dependency, shared_file_system):
                     bundle_dir_wait_num_tries=next_bundle_dir_wait_num_tries,
                 )
         else:
-            remove_path(run_state.bundle_path)
-            os.makedirs(run_state.bundle_path)
+            if len(run_state.bundle.metadata.get("remotes", [])) > 1:
```
```suggestion
            if len(run_state.bundle.metadata.get("remote_history", [])) <= 1:
```

teetone (approved):
Went over the PR offline with Ashwin

PR #3885, opened 2021-11-16
Fix worksheets display limit

### LINKED ISSUE #3838: Show more than 10 worksheets on user dashboard
image

On a user’s page, there’s a note that says “Note: this page only displays up to the first 100 worksheets.”  Only 10 worksheets are returned now.  We should  fetch more than 10 worksheets.

jzwang43:

The profile page is only displaying up to 10 worksheets instead of 100 for "my worksheets".

By default the limit is 10, we just need to specify limit to be 100.

### Related issues

fixes #3838 

### Screenshots

image

### Checklist

### DIFF
--- a/frontend/src/components/Dashboard/MainPanel.js
+++ b/frontend/src/components/Dashboard/MainPanel.js
@@ -189,7 +189,7 @@ class MainPanel extends React.Component<{
 
     componentDidMount() {
         // Fetch worksheets owned by the current user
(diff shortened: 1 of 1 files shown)


epicfaace (commented):
lgtm, @pranavjain can you review and approve this?

pranavjain (approved):
LGTM.

We might want to change the limit to 100 directly in the backend to avoid setting this from the frontend.

PR #3895, opened 2021-11-30
Fix login process doesn't remember the worksheets page

### LINKED ISSUE #3892: login process doesn't remember the worksheets page
To reproduce:

1. Log out.
2. Go to a worksheet page.
3. Log in.

CodaLab should take us back to the same worksheets page, but it forgets and takes us back to the user's home page instead.

jzwang43:

Currently, logging out and back in doesn't bring the user back to the previous worksheet page.

### Related issues

fixes #3892 

### Screenshots

To fix this, we just need to path in the correct props to the login page.

a

### Checklist

### DIFF
--- a/frontend/src/components/NavBar.js
+++ b/frontend/src/components/NavBar.js
@@ -359,7 +359,12 @@ class NavBar extends React.Component<{
                                 <Link to='/account/signup'>
                                     <Button color='inherit'>Sign Up</Button>
                                 </Link>
-                                <Link to='/account/login'>
+                                <Link
+                                    to={{
+                                        pathname: '/account/login',
+                                        state: { from: this.props.location },
+                                    }}
+                                >
                                     <Button color='inherit'>Login</Button>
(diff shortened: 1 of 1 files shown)


epicfaace (approved):

frontend/src/components/NavBar.js
```diff
@@ -359,7 +359,12 @@ class NavBar extends React.Component<{
                                 <Link to='/account/signup'>
                                     <Button color='inherit'>Sign Up</Button>
                                 </Link>
-                                <Link to='/account/login'>
+                                <Link
```
epicfaace, line 362: Should we do the same for the sign up button?

epicfaace (approved):

frontend/src/components/NavBar.js
jzwang43: I think it's outside the scope of this issue. But we can talk about it in the meeting? Thanks!

PR #3901, opened 2021-12-05
Re-add upstream ratarmount

epicfaace:
With ratarmount 0.1.3, should fix the earlier CI issues.


epicfaace:
huh, I'm still seeing the same error that @mxmlnkn fixed in 0.1.12... will investigate

image

mxmlnkn:
No need to investigate. It was a dumb error, which I have a strong déja-vu of fixing already. I tested for major version 2 instead of 3. But maybe I only fixed it in my manual tests, not in the actual code. You know that Github Actions does not even offer Ubuntu 16.04 runners anymore (or I did use a wrong version when trying it), so I'm kinda fixing this in the dark here. I released ratarmountcore 0.1.3.

epicfaace:
Oh, one way you could test for ubuntu 16 is to specify the `container` key in GHA -- see https://stackoverflow.com/questions/58930529/github-action-how-do-i-run-commands-inside-a-docker-container

mxmlnkn:
> Oh, one way you could test for ubuntu 16 is to specify the `container` key in GHA -- see https://stackoverflow.com/questions/58930529/github-action-how-do-i-run-commands-inside-a-docker-container

Thanks for the pointer. I set up a very rudimentary CI with an Ubuntu 16.04 Docker container, which can reproduce the bug for ratarmountcore 0.1.2 and which works for 0.1.3.

epicfaace (commented):
It works, thanks for your help @mxmlnkn ! 🎉

epicfaace:
@teetone can you review?

jzwang43 (approved):

epicfaace:
Blocked on merging until we can get https://github.com/codalab/codalab-worksheets/issues/3906 resolved.

PR #3828, opened 2021-10-10
Bundle store CLI commands

jzwang43:
Flow is documented here: https://github.com/codalab/codalab-worksheets/blob/f864bb2f1303bcf722e1c204d35f0c4741bc716d/docs/Multiple-Bundle-Stores.md

### Reasons for making this change

**Included in this PR:**

cl store add --name [store name]
Creates a new bundle store, allows one to configure credentials
cl store ls
Lists all bundle stores
cl store rm [store id]
Deletes bundle store with given id
cl upload --bundle-store [store name]
Uploads a bundle and specifies the bundle store. If no bundle store is specified, the CLI will pick the optimal available bundle store.

**Not included in this PR:**
cl replicate [bundle uuid] --bundle-store [store name]
Replicates the contents of the specified bundle to the specified bundle store.

cl store work [store name]
All bundles will by default be uploaded to this bundle store.

cl run --bundle-store [store name]
Runs a bundle and specifies the bundle store where results are uploaded. If no bundle store is specified, the worker will pick the optimal available bundle store.

cl rm -d --bundle-store [store name]
Updates existing command “cl rm -d”; keeps the bundle, but removes the bundle contents from the specified bundle store.

### Related issues
#3803 
#3827 

### Checklist

* [x] I've added a screenshot of the changes, if this is a frontend change
* [x] I've added and/or updated tests, if this is a backend change
* [x] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [x] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

PR #3905, opened 2021-12-08
Support GCS

### LINKED ISSUE #3904: Support GCP for bundle storage

epicfaace:
Fixes #3904. Depends on https://github.com/codalab/codalab-worksheets/pull/3828 to be merged first.


codalab/lib/bundle_store.py
```diff
@@ -422,5 +426,8 @@ def get_bundle_location(self, uuid):
         if storage_type == StorageType.AZURE_BLOB_STORAGE.value:
             file_name = "contents.tar.gz" if is_dir else "contents.gz"
             return f"azfs://{self._azure_blob_account_name}/bundles/{uuid}/{file_name}"
+        # if storage_type == StorageType.GCS_STORAGE.value:
+        #     file_name = "contents.tar.gz" if is_dir else "contents.gz"
```
epicfaace: This is blocked on getting bundle locations / bundle store methods working, as that's the way we'll distinguish between Azure Blob and GCS in this method.

alembic/versions/2021121222_rm_storage_type_enum_a325a0756797.py
```diff
@@ -0,0 +1,28 @@
...
+revision = 'a325a0756797'
+down_revision = '685cb4bc0b79'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('bundle_store', 'storage_type', type_=sa.String(255), existing_type=sa.Enum("disk", "azure_blob"), nullable=True)
```
epicfaace, line 19: This is a one-time change. Realized that it's difficult to modify enums so I changed the columns to just be Text types instead, so we don't need migrations for future expansions of supporting new storage types.

epicfaace:
@percyliang FYI this PR is about the unit of work required to support a new storage type that Apache Beam supports. Relatively straightforward and minimal -- a similar PR could add support for S3.

docs/Multiple-Bundle-Stores.md
```diff
@@ -36,9 +36,18 @@ cl store add --name blob --storage-type azure_blob --url azfs://devstoreaccount1
 cl upload --store blob mkdocs.yml
 ```
 
-## Steps to set up GCP
+## Steps to set up and test with GCP
+
+First, make sure the `CODALAB_GOOGLE_APPLICATION_CREDENTIALS` env var is set on the server that runs the REST server. It should be set to a path to a credentials JSON file.
+
+```
+export CODALAB_GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json
```
epicfaace, line 44: @teetone Flagging this for you -- you'll need to set up this environment variable when setting up GCP on codalab.stanford.edu.

jzwang43 (approved):

PR #3913, opened 2021-12-13
Fix store info display

### LINKED ISSUE #3910: Bundle store frontend error
image

jzwang43:

Legacy bundles will not have bundle stores so we won't show the store info if API returns empty.

`BundleLocationListSchema(many=True).dump` doesn't seem to be working so I'm using `json.dumps` for now.

### Related issues

fixes #3910 

### Screenshots

For an old file upload:
image

New bundle:
image

### Checklist


epicfaace:
Looks good except for the `BundleLocationListSchema(many=True).dump(bundle_locations).data` part. Why didn't that work? did you get a specific error? Ideally, we would use that instead of using json.dump.

jzwang43:
> 

The schema is defined incorrectly I believe. The error is `Unexpected Internal Error (Must have an `id` field). The administrators have been notified.`

epicfaace:
Are you sure that's the error you get? BundleLocationListSchema doesn't have an id field defined (https://github.com/codalab/codalab-worksheets/blob/fix-store-info/codalab/rest/schemas.py#L247-L255) (only BundleStoreSchema does)

jzwang43:
> Are you sure that's the error you get? BundleLocationListSchema doesn't have an id field defined (https://github.com/codalab/codalab-worksheets/blob/fix-store-info/codalab/rest/schemas.py#L247-L255) (only BundleStoreSchema does)

Yes I can confirm that's the error. If you look at the definition of the Schema class the BundleLocationListSchema extends, it requires an id field to be defined.

epicfaace:
> > Are you sure that's the error you get? BundleLocationListSchema doesn't have an id field defined (https://github.com/codalab/codalab-worksheets/blob/fix-store-info/codalab/rest/schemas.py#L247-L255) (only BundleStoreSchema does)
> 
> Yes I can confirm that's the error. If you look at the definition of the Schema class the BundleLocationListSchema extends, it requires an id field to be defined.

Can you try reverting the BundleLocationListSchema(many=True).dump(bundle_locations).data -> json.dump change? I think the json change you did is causing the CI tests to fail. We should maybe get on a call to talk through this further to see what exactly the issue.

jzwang43:
> > > Are you sure that's the error you get? BundleLocationListSchema doesn't have an id field defined (https://github.com/codalab/codalab-worksheets/blob/fix-store-info/codalab/rest/schemas.py#L247-L255) (only BundleStoreSchema does)
> > 
> > 
> > Yes I can confirm that's the error. If you look at the definition of the Schema class the BundleLocationListSchema extends, it requires an id field to be defined.
> 
> Can you try reverting the BundleLocationListSchema(many=True).dump(bundle_locations).data -> json.dump change? I think the json change you did is causing the CI tests to fail. We should maybe get on a call to talk through this further to see what exactly the issue.

If I revert the json.dumps change then the endpoint wouldn't work and this page will be broken.
The previous API I used was an old one and doesn't support multiple bundle locations.

codalab/model/bundle_model.py
```diff
@@ -3090,6 +3090,7 @@ def get_bundle_locations(self, bundle_uuid: str) -> List[dict]:
             rows = connection.execute(
                 select(
                     [
+                        cl_bundle_store.c.id,
```
epicfaace: let's revert this file (which I changed in an attempt to fix the json issue)

epicfaace (approved):

PR #3917, opened 2021-12-14
Allow hyphens in worker tags

teetone:


epicfaace (approved):
Looks good, but what's the reason for doing so?

And have you tested with a hyphen tag to make sure it works / doesn't cause any unexpected errors?

teetone:
> Looks good, but what's the reason for doing so?
> 
> And have you tested with a hyphen tag to make sure it works / doesn't cause any unexpected errors?

This is so we can name queues better (e.g., `gcp-small` vs `gcpsmall`).

epicfaace:
Sounds good

--
Ashwin Ramaswami

Issue #3911, opened 2021-12-13
"cl store add" should infer --storage-type from --url


epicfaace:
"cl store add" should infer --storage-type from --url so that --storage-type is not required.

So, current command:

```
cl store add --name blob --storage-type azure_blob --url azfs://devstoreaccount1/bundles
```

Ideal command that we want:

```
cl store add --name blob --url azfs://devstoreaccount1/bundles
```

epicfaace:
@wwwjn link to store documentation: https://codalab-worksheets.readthedocs.io/en/latest/Multiple-Bundle-Stores/

https://github.com/codalab/codalab-worksheets/pull/3928

pranavjain:
@wwwjn To update documentation.

PR #3928, opened 2021-12-17
infer storage type of bundle store

adiprerepa:
for https://github.com/codalab/codalab-worksheets/issues/3911

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1054,6 +1054,11 @@ class BundleCLI(object):
                 "url": args.url,
                 "authentication": args.authentication,
             }
+            if args.storage_type is None:
+                if args.url.startswith("azfs://"):
+                    bundle_store_info["storage_type"] = "azure_blob"
+                if args.url.startswith("gs://"):
+                    bundle_store_info["storage_type"] = "gcs"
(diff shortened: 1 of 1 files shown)


codalab/lib/bundle_cli.py
```diff
@@ -1054,6 +1054,11 @@ def do_store_command(self, args):
                 "url": args.url,
                 "authentication": args.authentication,
             }
+            if args.storage_type is None:
```
epicfaace, line 1057: Great! Can you update the cli tests so they include one example of not passing in the --storage-type argument?

codalab/lib/bundle_cli.py
```diff
@@ -1054,6 +1054,11 @@ def do_store_command(self, args):
...
+                if args.url.startswith("azfs://"):
+                    bundle_store_info["storage_type"] = "azure_blob"
```
epicfaace, line 1059: Also, let's actually move this logic to here (https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/bundles.py#L514-L522) to be consistent with how storage_format is inferred.

epicfaace (changes requested):

epicfaace:
@adiprerepa can you please make these changes?

PR #3930, opened 2022-01-06
Update bundle locations endpoint to use schema dump

### LINKED ISSUE #3925: `/bundles/<bundle_uuid:re:%s>/locations/` should use `BundleLocationListSchema(many=True).dump`
Investigate the issue with `BundleLocationListSchema(many=True).dump`.
We should use this function instead of `json.dumps`

jzwang43:

We need to use `BundleLocationListSchema(many=True).dump` for the endpoint instead of `json.dump`.

### Related issues

fixes #3925 

### Screenshots

Nothing should change.

image

### Checklist


epicfaace (commented):
Looks good. Just want to check -- is the output of this endpoint the exact same as it was before?

jzwang43:
> Looks good. Just want to check -- is the output of this endpoint the exact same as it was before?

Yes, the data is now in "data" field of the json object, which we used for other endpoints too.

epicfaace (approved):
Great!

codalab/rest/bundles.py
```diff
@@ -442,7 +441,7 @@ def _fetch_bundle_locations(bundle_uuid: str):
     - `bundle_uuid`: Bundle UUID to get the locations for
     """
     bundle_locations = local.model.get_bundle_locations(bundle_uuid)
-    return json.dumps(bundle_locations)
+    return {"data": BundleLocationListSchema(many=True).dump(bundle_locations).data}
```
epicfaace, line 444: Wait, why is it that in the other endpoints we don't need to explicitly specify the `{"data": ` data structure?

frontend/src/components/Bundle/Bundle.js
```diff
@@ -131,8 +132,8 @@ class Bundle extends React.Component<
             .then(callback)
             .catch(errorHandler);
 
-        callback = (storeInfo) => {
-            if (!storeInfo || storeInfo.length == 0) return;
+        callback = ({ data: storeInfo }) => {
```
epicfaace, line 135: Do we need to modify the `cl bundle ls` command to properly handle the new output structure as well?

epicfaace (changes requested):
Great!

codalab/rest/bundles.py
jzwang43: The dump returns a list, which was causing the error because it's not a JSON object.

frontend/src/components/Bundle/Bundle.js
jzwang43: Do you mean `cl store ls`? I tested it's working fine.

codalab/rest/bundles.py
epicfaace: Hmm, but I'm jsut confused because for example this endpoint (https://github.com/codalab/codalab-worksheets/blob/7120ca863f0d9eb67400f54b956d181dba966404/codalab/rest/groups.py#L26-L42) also has the dump return a list, but we don't need the `{"data":` section.

jzwang43: It calls `include_group_relationships` which does a bunch of different operations, so it's not directly returning the list.

epicfaace (approved):

PR #3932, opened 2022-01-07
Show bundle stores when running "cl info"

### LINKED ISSUE #3914: Show bundle stores when running "cl info"
When running "cl info [uuid]", it should show the associated bundle locations / bundle stores for that bundle.

### LINKED ISSUE #3973: Implement "cl store info" command
The "cl store info [bundle_store_uuid]" command should list information about a specific bundle store, such as its url, etc.

jzwang43:

When running "cl info [uuid]", it should show the associated bundle locations / bundle stores for that bundle.

### Related issues

fixes #3914 

### Screenshots

correct store info for new bundles:
    
image

legacy bundle:

image

### Checklist


epicfaace (commented):
Can we show the bundle store name instead of ID?

Also, how does it look like if there are multiple bundle stores for a single bundle?

jzwang43:
> Can we show the bundle store name instead of ID?
> 
> Also, how does it look like if there are multiple bundle stores for a single bundle?

@epicfaace Thanks, I have updated the code and also the screenshot.

jzwang43:
@epicfaace can I get a stamp on this? thanks

codalab/lib/bundle_cli.py
```diff
@@ -2446,7 +2446,16 @@ def print_basic_info(self, client, info, raw):
         show_hidden = client.fetch('user')['is_root_user']
 
         for key, value in worksheet_util.get_formatted_metadata(cls, metadata, raw, show_hidden):
-            lines.append(self.key_value_str(key, value))
+            if key == 'store':
```
epicfaace: We actually shouldn't show this under the 'store' key for metadata because the 'store' key has a different meaning (it's the original bundle store that the bundle was uploaded to -- see https://github.com/codalab/codalab-worksheets/blob/c82a01e14272db382c460e0197ff813e7264282f/codalab/bundles/uploaded_bundle.py#L32).

Instead, we should:
- hide the store key from displaying -- that already happens because it's set to `hidden=True` in the link above
- below the metadata keys, add a separate section where we say "Bundle stores: store1, store2"

epicfaace (changes requested):
One more change!

codalab/lib/bundle_cli.py
jzwang43: Thanks, I have updated the code based on your suggestion:)

epicfaace: thanks -- can you add a screenshot? :)

jzwang43: Updated

codalab/client/json_api_client.py
```diff
@@ -706,3 +706,10 @@ def get_workers_info(self):
         request_path = '/workers/info'
         response = self._make_request('GET', request_path)
         return response['data']
+
+    @wrap_exception('Unable to get the locations of bundles')
+    def get_multiple_bundle_locations(self, bundle_uuid):
```
epicfaace: I think this name makes more sense?

```suggestion
    def get_bundle_locations(self, bundle_uuid):
```

epicfaace (commented):
One more thing -- can we add a line in test_cli.py (https://github.com/codalab/codalab-worksheets/blob/master/tests/cli/test_cli.py#L912) so that we run "cl info" on the uploaded bundle and ensure the bundle store's name shows up in the result?

codalab/client/json_api_client.py
jzwang43: Three's a function named get_bundle_locations in `bundle_model.py.` I didn't want to cause any confusion so I named my function that way.

epicfaace: I think it's fine to call this one `get_bundle_locations` -- as this method pretty much corresponds with `get_bundle_locations` in bundle_model.py, right?

pranavjain:
@epicfaace Do you want to merge your PR before this one?

jzwang43:
@epicfaace I have discovered a weird issue when` BundleLocationListSchema(many=True).dump` returns different format for storage_type=disk and storage_type=azure_blob.

You can reproduce this by creating one disk store and one azure store following the instruction here:
https://codalab-worksheets.readthedocs.io/en/latest/Multiple-Bundle-Stores/

Then if you inspect the rest API result for both bundles from `rest/bundles/{bundle_uuid}/locations/`, you can see the difference.

image
vs
image

epicfaace:
> @epicfaace I have discovered a weird issue when` BundleLocationListSchema(many=True).dump` returns different format for storage_type=disk and storage_type=azure_blob.
> 
> You can reproduce this by creating one disk store and one azure store following the instruction here: https://codalab-worksheets.readthedocs.io/en/latest/Multiple-Bundle-Stores/
> 
> Then if you inspect the rest API result for both bundles from `rest/bundles/{bundle_uuid}/locations/`, you can see the difference.
> 
> image vs image

Hmm, can you take a look and fix it in this PR? Worst case, we can revert to `json.dumps` if that makes the output format uniform.

jzwang43:
> > @epicfaace I have discovered a weird issue when` BundleLocationListSchema(many=True).dump` returns different format for storage_type=disk and storage_type=azure_blob.
> > You can reproduce this by creating one disk store and one azure store following the instruction here: https://codalab-worksheets.readthedocs.io/en/latest/Multiple-Bundle-Stores/
> > Then if you inspect the rest API result for both bundles from `rest/bundles/{bundle_uuid}/locations/`, you can see the difference.
> > image vs image
> 
> Hmm, can you take a look and fix it in this PR? Worst case, we can revert to `json.dumps` if that makes the output format uniform.

I was able to fix the schema dump by changing the url to string. all checks are passing now

epicfaace (approved):

epicfaace:
This PR isn't dependent on https://github.com/codalab/codalab-worksheets/pull/3981, so let's go ahead and merge this one

PR #3941, opened 2022-01-11
Auth worker manager

### LINKED ISSUE #3888: Worker manager prompts login even when specifying --password-file
Example command: 

```cl-worker-manager  --server https://codalab.stanford.edu --search "request_gpus=0" --min-seconds-between-workers 180 --sleep-time 60  --worker-shared-memory-size-gb 8 --min-workers 0 --max-workers 10 --worker-work-dir-prefix /scr/scr-with-most-space/ --worker-pass-down-termination --worker-delete-work-dir-on-exit --worker-exit-on-exception --worker-exit-after-num-runs 1 slurm-batch --partition john --cpus 10 --gpus 0 --memory-mb 240000 --password-file ~/.codalab/codalab-nlp.password --slurm-work-dir ~/scr/slurm_output/ --job-name codalab-slurm-worker-stanford-john ```

Expected behavior: The worker manager should start without having the user log in.

### PR DESCRIPTION
**Pull request by teetone**

- Have the temporary codalab manager initialize from the state file when starting up

### Related issues

Resolves #3888 

### Checklist

* [X] I've added and/or updated tests, if this is a backend change
* [X] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [X] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

### CODE REVIEW

File: codalab/server/oauth2_provider.py [hunk 1]
```diff
@@ -150,7 +151,7 @@ def validate_client_id(self, client_id):
 
             oauth._validator = MyValidator()
         """
-        expires_in = self.app.config.get('OAUTH2_PROVIDER_TOKEN_EXPIRES_IN')
+        expires_in = self.app.config.get('OAUTH2_PROVIDER_TOKEN_EXPIRES_IN', parse_duration("30d"))
```
epicfaace: cli vs worker manager

epicfaace: @teetone It seems like this change doesn't actually have to do with https://github.com/codalab/codalab-worksheets/issues/3888, because that issue doesn't have to do with an expiration. Can you revert this change and see if https://github.com/codalab/codalab-worksheets/issues/3888 is still fixed? In that case, we should then merge this PR in without this file changed.

teetone: You're right. This change doesn't directly fix #3888, but this change is also related because we use the same auth token to authenticate when the user starts a worker manager. We do want to extend the expiration too.

epicfaace: Can you revert this change, then, so we can merge this PR?

And add this change in a separate PR and link it to another issue? The reason is because I think this change is a bit more complicated / requires more discussion before we add it in.

codalab/lib/codalab_manager.py
```diff
@@ -127,8 +127,13 @@ def __init__(self, temporary=False, config={}, clients={}):
 
         if self.temporary:
             self.config = config
-            self.state = {'auth': {}, 'sessions': {}}
             self.clients = clients
+
+            # Initialize `state` with the content in `self.state.path`, if it exists.
```
epicfaace:
```suggestion
            # Initialize `state` with the content in `self.state_path`, if it exists.
            # Usually, a temp session is created when running cl-worker-manager. In this case, we want to have a temp session so that we don't interfere with state stored by "cl" processes. However, we want to initialize the temp session state with the state stored from the CodaLab CLI "cl", so that the user is not prompted again for their credentials.
```

codalab/lib/codalab_manager.py
```diff
@@ -127,8 +127,13 @@ def __init__(self, temporary=False, config={}, clients={}):
...
+            if os.path.exists(self.state_path):
```
epicfaace, line 137: Why is the codalab manager temporary when running the scenario in https://github.com/codalab/codalab-worksheets/issues/3888?

epicfaace: Because of https://github.com/codalab/codalab-worksheets/pull/3151

epicfaace (approved):

tests/unit/lib/codalab_manager_test.py
```diff
@@ -38,3 +41,19 @@ def test_temp_codalab_manager(self):
             os.path.exists(manager.state_path),
             msg='Assert that the current state is not written out to state_path for a temporary CodaLabManager',
         )
+
+    def test_temp_codalab_manager_initialize_state(self):
```
epicfaace, line 49:
```suggestion
    def test_temp_codalab_manager_initialize_state(self):
        """A codalab manager with temporary state should initialize its state from an existing state.json file if it is present."""
```

epicfaace (approved):

codalab/server/oauth2_provider.py
```diff
@@ -151,7 +150,7 @@ def validate_client_id(self, client_id):
 
             oauth._validator = MyValidator()
         """
-        expires_in = self.app.config.get('OAUTH2_PROVIDER_TOKEN_EXPIRES_IN', parse_duration("30d"))
+        expires_in = self.app.config.get('OAUTH2_PROVIDER_TOKEN_EXPIRES_IN')
```
epicfaace: @teetone make an issue for this --

try making the access token lifetime really short (15 secs), see if the refresh token works.

PR #3942, opened 2022-01-11
Update mergify to get rid of deprecated strict mode

epicfaace:
Mergify no longer works because strict mode has been deprecated -- following instructions here (https://blog.mergify.com/strict-mode-deprecation/) to get the same functionality.

Strict mode just makes sure that each PR is merged serially and that each PR must be up-to-date and have CI passing before being merged.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,3 +1,8 @@
+queue_rules:
+  - name: default
+    conditions:
+      - check-success=All CI tasks complete
+
 pull_request_rules:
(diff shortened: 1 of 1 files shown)


teetone:
I tried something similar, and the build still failed for me.

epicfaace:
> I tried something similar, and the build still failed for me.

It should work. We need to merge this manually (as mergify is reading config from the `master` branch) and then it should work.

teetone (approved):

PR #3954, opened 2022-01-18
Fix building docker files from scratch

### LINKED ISSUE #3955: Docker server images builds are broken
See https://github.com/codalab/codalab-worksheets/pull/3954

image

How to replicate:
- Check out https://github.com/codalab/codalab-worksheets/pull/3954
- Start codalab (`codalab-service start -bd`)
- Docker images don't build

epicfaace:
The previous ppa doesn't have python3.6 anymore, updated to use a new repository.

### DIFF
--- a/.github/workflows/postcommit.yml
+++ b/.github/workflows/postcommit.yml
@@ -1,0 +1,29 @@
+# Build Docker files from scratch after each commit to master.
+# We do this in case there are any issues with the Dockerfiles
+# that were missed in the Test CI pipelines by Docker caching.
+name: Postcommit
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        service: [rest-server, worker, frontend]
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


jzwang43:
@epicfaace I tested checking out the branch and building the images. They seem to be working? Anything else I need to do other than `start -bd` to reproduce?

epicfaace:
@jzwang43 it's working because it's currently using the cache as of my latest commit. Try undoing my most recent commit (https://github.com/codalab/codalab-worksheets/pull/3954/commits/52f26a440ef00ee4327ad253bccdd6c18f0c8a70) -- this will cause the cache to bust and then you'll see the error.

jzwang43:
@epicfaace I updated the apt repository and the build should be working now:)

epicfaace (commented):
Great! I think we might need to update the other dockerfiles too.

jzwang43:
> Great! I think we might need to update the other dockerfiles too.

Updated! Thanks!

jzwang43:
> Great! I think we might need to update the other dockerfiles too.

Actually we only need the update for 16.04. 20.04 doesn't require any changes.

epicfaace:
lgtm

PR #3971, opened 2022-01-25
Fix comment

epicfaace:

Fix comment to link to the right issue.


epicfaace:
this is included in https://github.com/codalab/codalab-worksheets/pull/3972

PR #3993, opened 2022-02-09
Improve handling if user does not have access to the server v2

### LINKED ISSUE #3961: cs324 - users who haven't been added to the protected mode server get ugly exceptions
- User creates an account on cs324.codalab.org
- User gets ugly experience (can't create worksheets, etc.)

Ideal behavior:
- users should just be presented with a message that says "please contact the admins to get access to this server"

jzwang43:

    User creates an account on cs324.codalab.org
    User gets ugly experience (can't create worksheets, etc.)

### Related issues

Fixes #3961 

### Screenshots

Shows alert if user doesn't have access to the server:
image

### Checklist

### DIFF
--- a/frontend/src/components/Dashboard/NewDashboard.js
+++ b/frontend/src/components/Dashboard/NewDashboard.js
@@ -68,7 +68,7 @@ class NewDashboard extends React.Component<{
 
     /** Renderer. */
     render() {
-        // If the ser is in protected mode and the user does not have access, show error message.
+        // If the server is in protected mode and the user does not have access, show error message.
         if (
             this.state.userInfo &&
             this.state.userInfo.protected_mode &&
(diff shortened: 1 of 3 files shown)


pranavjain:
@jzwang43 You might want to update the branch

frontend/src/components/Dashboard/NewDashboard.js
```diff
@@ -68,7 +68,12 @@ class NewDashboard extends React.Component<{
 
     /** Renderer. */
     render() {
-        if (this.state.userInfo && !this.state.userInfo.has_access) {
+        // If the ser is in protected mode and the user does not have access, show error message.
```
epicfaace:
```suggestion
        // If the server is in protected mode and the user does not have access, show error message.
```

epicfaace (approved):

PR #4002, opened 2022-02-15
Worker manager documentation for Azure Batch and GKE

### LINKED ISSUE #3603: Document batch worker managers
We should document batch worker managers and include steps for how to set them up / what needs to be configured to do so. We could start with the k8s batch worker manager, since that's what we most recently added in.

teetone:
Resolves https://github.com/codalab/codalab-worksheets/issues/3603


epicfaace (commented):
lgtm! is this copied from the deployment docs or are there other changes you added?

If it's the same as the deployment docs, maybe we should remove these sections from the deployment docs and just link to the regular codalab worksheets documentation site instead (to avoid duplication).

teetone:
> lgtm! is this copied from the deployment docs or are there other changes you added?
> 
> If it's the same as the deployment docs, maybe we should remove these sections from the deployment docs and just link to the regular codalab worksheets documentation site instead (to avoid duplication).

I changed it to make it more general. We should keep the docs in the deployment repo since they're specific to us.

epicfaace:
> I changed it to make it more general. We should keep the docs in the deployment repo since they're specific to us.

Hmm, which parts were changed though? There's a lot of duplication now between the deployment repo docs and public docs and it'll be harder to update in both places -- might it be possible to only include the commands specific to us in the deployment repo docs, and then just link to the main docs?

teetone:
> > I changed it to make it more general. We should keep the docs in the deployment repo since they're specific to us.
> 
> Hmm, which parts were changed though? There's a lot of duplication now between the deployment repo docs and public docs and it'll be harder to update in both places -- might it be possible to only include the commands specific to us in the deployment repo docs, and then just link to the main docs?

We're giving up convenience for less documentation - now people have to look at two sets of documentation to set something up. I like the convenience of just looking at the deployment docs to deploy without following another link and trying to put things together.

Anyways, we do want these changes in this PR to be merged, so if it looks good to you, can you approve the PR? We can follow up on the deployment docs separately.

epicfaace:
I just can't tell what has changed so I can't effectively review this PR yet. I will make a draft PR in deployment docs so it's easier to compare what has changed in the docs.

epicfaace:
OK @teetone , I've copied these files into https://github.com/codalab/deployment/pull/214 to make it easier to review and see what changed. let's continue the review there.

> We're giving up convenience for less documentation - now people have to look at two sets of documentation to set something up. I like the convenience of just looking at the deployment docs to deploy without following another link and trying to put things together.

We can still keep it convenient. For example, it might make sense to have just one set of deployment docs, but maybe we can just keep the kubernetes yaml files in one place (the CodaLab public repo) if nothing significant has changed. There's also the convenience of maintaining this documentation by the team -- especially once we both are gone -- that we should take into account.

> Anyways, we do want these changes in this PR to be merged, so if it looks good to you, can you approve the PR? We can follow up on the deployment docs separately.

I think it's better to get it right when we first merge it in. It's hard to manage too many docs and we might never get to it otherwise :)

teetone:
> OK @teetone , I've copied these files into [codalab/deployment#214](https://github.com/codalab/deployment/pull/214) to make it easier to review and see what changed. let's continue the review there.
> 
> > We're giving up convenience for less documentation - now people have to look at two sets of documentation to set something up. I like the convenience of just looking at the deployment docs to deploy without following another link and trying to put things together.
> 
> We can still keep it convenient. For example, it might make sense to have just one set of deployment docs, but maybe we can just keep the kubernetes yaml files in one place (the CodaLab public repo) if nothing significant has changed. There's also the convenience of maintaining this documentation by the team -- especially once we both are gone -- that we should take into account.
> 
> > Anyways, we do want these changes in this PR to be merged, so if it looks good to you, can you approve the PR? We can follow up on the deployment docs separately.
> 
> I think it's better to get it right when we first merge it in. It's hard to manage too many docs and we might never get to it otherwise :)

Sorry, I misunderstood then. You said it looks good to you, so I thought you reviewed it. 

I still think we should have two separate documentation specific to us and end-users because it's much more convenient to deploy, and these instructions rarely change. The AWS, Azure, and GCP setup instructions haven't really changed since the worker managers were implemented.

I will close this PR then and open a new one when we figure out how we want to maintain the docs. We can sync more offline.

epicfaace:
Ok, sounds good. Let's keep this PR open so we don't forget to address it?

Also this PR is really great -- I'm glad we're moving a lot of GKE setup (potentially) to the public repo! Will make it easier to set up GKE on a new instance like cs324 if we don't need access to the private deployment repo on that instance to do so.

epicfaace:
Todo: https://github.com/codalab/codalab-worksheets/issues/3603#issuecomment-1063395752

epicfaace (approved):

PR #4007, opened 2022-02-17
Be able to control seconds between workers for gpu vs cpu worker managers

teetone:
For GKE, the GPU pool takes much longer to spin up a VM than a CPU pool because of the Nvidia drivers installation. We need to change the variables in the deployment repo after.


codalab_service.py
```diff
@@ -474,6 +468,12 @@ def has_callable_default(self):
             type=int,
             default=1,
         ),
+        CodalabArg(
+            name=f'worker_manager_{worker_manager_type}_seconds_between_workers',
+            help='Number of seconds to wait between launching two workers',
```
```suggestion
            help=f'Number of seconds to wait between launching two {worker_manager_type} workers',
```

epicfaace (approved):

PR #4019, opened 2022-02-21
Fix bundle manager check in transition_bundle_worker_offline

epicfaace:
Fixes https://github.com/codalab/codalab-worksheets/pull/4019 by adding the proper `and_` operator in sqlalchemy and also allowing `FINALIZING` bundles to pass the check.

### DIFF
--- a/tests/unit/server/bundle_manager/__init__.py
+++ b/tests/unit/server/bundle_manager/__init__.py
@@ -158,8 +158,10 @@ class TestBase:
             uuid=generate_uuid(),
             state=state,
         )
-        bundle.is_frozen = None
+        bundle.frozen = None
         bundle.is_anonymous = False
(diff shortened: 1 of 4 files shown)


tests/unit/server/bundle_manager/__init__.py
```diff
@@ -158,8 +158,10 @@ def create_run_bundle(self, state=State.CREATED, metadata=None):
             uuid=generate_uuid(),
             state=state,
         )
-        bundle.is_frozen = None
+        bundle.frozen = None
```
epicfaace, line 161: We had some errors with test setup that only surfaced when writing tests for this PR.

jzwang43 (approved):

jzwang43 (approved):

PR #4020, opened 2022-02-21
Remove more chat legacy code

epicfaace:
Remove more chat legacy code -- noticed some of this code because it had similar issues to https://github.com/codalab/codalab-worksheets/pull/3882 (using `or` instead of `or_` or `|`).


pranavjain:
@epicfaace Can you give me more context about it? I don't have much idea to review this one.

epicfaace:
CodaLab used to store chats in a `chat` table very long ago. It's no longer used (we no longer support chats), so I'm just removing the code that interfaced with it. This way, our code files are simpler / cleaner / easier to read.

jzwang43 (approved):

PR #4022, opened 2022-02-21
Better logging for workers

epicfaace:
Better logging for workers. This includes:
- don't log unnecessary things from apache beam or docker
- don't log things such as just checking CPU / resource utilization for a job
- log every time a worker checks in


codalab/worker/worker_run_state.py
```diff
@@ -472,7 +472,6 @@ def check_and_report_finished(run_state):
             )
 
         def check_resource_utilization(run_state: RunState):
-            logger.info(f'Checking resource utilization for bundle. uuid: {run_state.bundle.uuid}')
```
teetone, line 475: Should we change these to `debug` instead of removing them?

epicfaace: I don't think so, it doesn't really seem to serve any purpose and I think it would be pretty noisy even in debug mode. We'll be doing this anyway on every iteration of the worker for every bundle, so I think it's better not to log.

epicfaace:
Please review when you get a chance @teetone

epicfaace:
@teetone please review

epicfaace:
@teetone Please review

teetone (approved):

PR #4029, opened 2022-02-22
Fix Stdout and File Section not working/refreshing

### LINKED ISSUE #3994: Stdout and File Section not working/refreshing
- User should be able to open the Stddout file and view its content in real time when the job is running.
Bug: The file keeps loading

- Files list take too long to show (refer to 0:45 in the video)
- Stddout output should be printed on the bundle infobox. (Randomly repros, Not visible in the video, visible in the attached screenshot)

https://user-images.githubusercontent.com/10134824/153292800-6e8addb6-6dfc-41f9-92c7-ec5a129176f2.mov

image

jzwang43:

Previously the front end was constantly sending request to the back end even if the previous one was pending, added state variables to keep track of pending request so we don't overwhelm the back end.
### Related issues
fixes #3994

### Screenshots

a

No extra request when the call is pending.

Also tested `while sleep 2; do echo thinking; done`, and it's updating file.

### Checklist


frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -126,18 +133,24 @@ const BundleDetail = ({
...
                     fetchRequests.push(
-                        fetchFileSummary(uuid, '/' + name).then(function(blob) {
-                            stateUpdate[name] = blob;
-                        }),
+                        fetchFileSummary(uuid, '/' + name)
+                            .then(function(blob) {
+                                stateUpdate[name] = blob;
+                            })
+                            .catch((e) => {
+                                setFetchingFileSummary(false);
```
epicfaace: This isn't correct -- what if one request (fetching stdout) fails but then the second request (fetching stderr) is still in progress -- then `fetchingFileSummary` will be set to False (when it should still be True).

I'd suggest instead setting `fetchingFileSummary` to `0` at the beginning so it stores a counter that stores the number of pending requests. Let's replace the following:
- `setFetchingFileSummary(true)` -> `setFetchingFileSummary(f => f + 1)`
- `setFetchingFileSummary(false)` -> `setFetchingFileSummary(f => f - 1)`
- `if (!info || fetchingFileSummary)` -> `if (!info || fetchingFileSummary > 0)`

Finally, maybe rename `fetchingFileSummary` to `pendingFileSummaryFetches`?

epicfaace (changes requested):

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
jzwang43: Thanks I have updated the state variable!

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -113,9 +118,11 @@ const BundleDetail = ({
...
         const info = response.data;
-        if (!info) return;
+        if (!info || pendingFileSummaryFetches > 0) return;
         if (info.type === 'file' || info.type === 'link') {
+            setPendingFileSummaryFetches((f) => f + 1);
             return fetchFileSummary(uuid, '/').then(function(blob) {
+                setPendingFileSummaryFetches((f) => f - 1);
                 setFileContents(blob);
                 setStderr(null);
                 setStdout(null);
```
epicfaace: we should replace this with `.finally(() => setPendingFileSummaryFetches((f) => f - 1))` (so this will be called whether the promise succeeds or fails)

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -126,18 +133,25 @@ const BundleDetail = ({
...
+                                setPendingFileSummaryFetches((f) => f - 1);
+                            }),
                     );
                 } else {
                     stateUpdate[name] = null;
                 }
             });
             Promise.all(fetchRequests).then((r) => {
+                setPendingFileSummaryFetches((f) => f - 1);
                 setFileContents(stateUpdate['fileContents']);
```
epicfaace: we should replace this with `.finally(() => setPendingFileSummaryFetches((f) => f - 1))` (so this will be called whether the promise succeeds or fails)

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -126,18 +133,25 @@ const BundleDetail = ({
...
+                    setPendingFileSummaryFetches((f) => f + 1);
                     fetchRequests.push(
-                        fetchFileSummary(uuid, '/' + name).then(function(blob) {
-                            stateUpdate[name] = blob;
-                        }),
+                        fetchFileSummary(uuid, '/' + name)
+                            .then(function(blob) {
+                                stateUpdate[name] = blob;
+                            })
+                            .catch((e) => {
```
epicfaace: we should replace this with `.finally(() => setPendingFileSummaryFetches((f) => f - 1))` (so this will be called whether the promise succeeds or fails)

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -61,22 +63,25 @@ const BundleDetail = ({
...
-            });
+                .then((r) => r.json())
+                .catch((error) => {
+                    setFetchingMetadata(false);
+                    setBundleInfo(null);
+                    setFileContents(null);
+                    setStderr(null);
+                    setStdout(null);
+                    setErrorMessages((errorMessages) => errorMessages.concat([error]));
+                });
```
epicfaace, line 85: we should replace this with `.finally(() => setFetchingMetadata(false)` (so this will be called whether the promise succeeds or fails)

epicfaace (commented):
Added a suggestion to use `.finally` instead. This is because we have to call setFetchingMetadata / setPendingFileSummaryFetches when each network promise either succeeds or fails, and `.finally` avoids duplication. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

Can you also test this locally to ensure it does refresh? I don't think the PR as it is now would have worked -- because `setFetchingMetadata(false)` is currently not called on success of `fetcherMetadata`.

jzwang43:
> Added a suggestion to use `.finally` instead. This is because we have to call setFetchingMetadata / setPendingFileSummaryFetches when each network promise either succeeds or fails, and `.finally` avoids duplication. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally
> 
> Can you also test this locally to ensure it does refresh? I don't think the PR as it is now would have worked -- because `setFetchingMetadata(false)` is currently not called on success of `fetcherMetadata`.

setFetchingMetadata(false) is called in `onSuccess` in `useSWR`, it should be moved to finally as well. But yeah I tested before and it's working.

Thanks for the suggestion of using finally, I'll update that.

epicfaace:
> setFetchingMetadata(false) is called in `onSuccess` in `useSWR`,

ooh interesting, where in the code? Can't seem to locate it

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -152,6 +166,7 @@ const BundleDetail = ({
         revalidateOnMount: true,
         refreshInterval: refreshInterval,
         onSuccess: (response) => {
+            setFetchingMetadata(false);
```
jzwang43: @epicfaace it's here:)

epicfaace: Wait, why do we call setFetchingMetadata after call fetcherContents? Shouldn't we call setFetchingMetadata only after calling fetcherMetadata?

jzwang43: Yeah just cleaned up everything with finally

epicfaace (approved):
One more thing -- I think we should also add the `finally...` code to `fetcherContents`? Because the same issue could happen with that network request as well?

Feel free to merge after adding that.

epicfaace:
Great!

--
Ashwin Ramaswami

Issue #3883, opened 2021-11-10
Investigate Material UI vs Bootstrap for CodaLab Frontend


pranavjain:
We are currently using both Material UI (MUI) and Bootstrap at different places.

The goal of this task is to compare both of these and eventually move CodaLab to use only one.

(This will be a mega issue and sub-tasks would be added.)

pranavjain:
Not a priority right now. Clearing assignment.

leilenah:
Moving towards MUI for all new features.

PR #4033, opened 2022-02-25
Wrap long command 

### LINKED ISSUE #4027: very long command should wrap
Right now, we have a huge scrollbar within a scrollbar

image

jzwang43:

very long command should wrap

### Related issues

fixes #4027

### Screenshots

image
### Checklist

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
@@ -169,7 +169,7 @@ class BundleDetailSideBar extends React.Component<{
                             </div>
                         </CopyToClipboard>
                         <div style={{ flexWrap: 'wrap', flexShrink: 1 }}>
-                            <pre style={{ whiteSpace: 'pre-wrap' }}>{bundleInfo.command} </pre>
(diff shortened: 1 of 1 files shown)


frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
```diff
@@ -169,7 +169,7 @@ class BundleDetailSideBar extends React.Component<{
                             </div>
                         </CopyToClipboard>
                         <div style={{ flexWrap: 'wrap', flexShrink: 1 }}>
-                            <pre>{bundleInfo.command} </pre>
+                            <pre style={{ whiteSpace: 'pre-wrap' }}>{bundleInfo.command} </pre>
```
jzwang43: Inline style is not the best but the file is using withStyles which is also deprecated, so for the scope of this fix we just put it here. This file should be refactored when we start #3883

epicfaace:
```suggestion
                            <pre style={{ whiteSpace: 'pre-wrap' }}>{bundleInfo.command}</pre>
```

epicfaace (approved):

PR #4040, opened 2022-03-04
Adjust header padding and fix body padding

leilenah:
## Issue 1

The header appears to be too small when the search bar is present. The search bar takes up the full height of the header, making the header appear visually unbalanced. 

image

### Solution

This change adds slight padding to the top and bottom of the header. **Note:** Using `spacing.larger` here might look even better.

image

## Issue 2

The hero is currently being pushed up behind the header. The embedded CodaLab YouTube video is calling attention to this stying issue.

image

### Solution

I've updated the global `body` styles to resolve this issue.

image

### Checklist

### DIFF
--- a/frontend/src/components/NavBar.js
+++ b/frontend/src/components/NavBar.js
@@ -320,7 +320,7 @@ class NavBar extends React.Component<{
 
         return (
             <MuiThemeProvider theme={overrideMedia}>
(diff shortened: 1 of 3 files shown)


epicfaace (approved):
lgtm 👍🏼

PR #4044, opened 2022-03-07
revising docker files to fix fatal issues

yuchenWYC:

The code to start the CodaLab service 
```
./codalab_service.py start -bd
```
would fail to someone who newly cloned the repo because deadsnakes ppa in the old docker file no longer supports Ubuntu 16.04 for any python version (https://github.com/deadsnakes/issues/issues/195). Therefore, I had to change the Ubuntu version to 20.04 to finish setting up the environment.

### DIFF
--- a/docker_config/dockerfiles/Dockerfile.default-gpu
+++ b/docker_config/dockerfiles/Dockerfile.default-gpu
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install -y \
 RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.6 1
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.6 1
 
-RUN curl https://bootstrap.pypa.io/pip/3.6/get-pip.py | python3.6
+RUN curl https://bootstrap.pypa.io/get-pip.py | python3.6
 
 ### Without this Python thinks we're ASCII and unicode chars fail
 ENV LANG C.UTF-8
--- a/docker_config/dockerfiles/Dockerfile.server
+++ b/docker_config/dockerfiles/Dockerfile.server
@@ -1,11 +1,11 @@
-FROM ubuntu:16.04
+FROM ubuntu:20.04
 MAINTAINER CodaLab Worksheets <codalab.worksheets@gmail.com>
 
 ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update && \
   apt-get install -y --no-install-recommends software-properties-common && \
-  add-apt-repository ppa:fkrull/deadsnakes && \
+  add-apt-repository ppa:deadsnakes/ppa && \
   apt-get update && \
   apt-get install -y --no-install-recommends \
     build-essential \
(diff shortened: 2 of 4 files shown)


docker_config/dockerfiles/Dockerfile.default-gpu
```diff
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install -y \
 RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.6 1
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.6 1
 
-RUN curl https://bootstrap.pypa.io/pip/3.6/get-pip.py | python3.6
```
epicfaace, line 35: @yuchenWYC are you sure these changes are needed (changes to pip url)? @jzwang43 just did the opposite of this to fix the docker builds a while ago -- see https://github.com/codalab/codalab-worksheets/pull/3992

docker_config/dockerfiles/Dockerfile.server
```diff
@@ -1,11 +1,11 @@
-FROM ubuntu:16.04
+FROM ubuntu:20.04
 MAINTAINER CodaLab Worksheets <codalab.worksheets@gmail.com>
 
 ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update && \
   apt-get install -y --no-install-recommends software-properties-common && \
-  add-apt-repository ppa:fkrull/deadsnakes && \
+  add-apt-repository ppa:deadsnakes/ppa && \
```
epicfaace, line 8: This is great, thanks! This does something similar to https://github.com/codalab/codalab-worksheets/pull/3954 but I forgot to update these dockerfiles in that PR.

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -62,7 +62,7 @@ RUN mkdir ${WORKDIR}/scripts
 
 # Install dependencies
 COPY requirements.txt requirements.txt
-RUN python3.6 -m pip install --user --upgrade pip==20.3.4; \
+RUN python3.6 -m pip install --user --upgrade pip; \
```
epicfaace, line 65: fyi @jzwang43 this reverts your change here https://github.com/codalab/codalab-worksheets/pull/3986

epicfaace:
@jzwang43 Can you please review -- this seems to relate to your earlier PRs, https://github.com/codalab/codalab-worksheets/pull/3992 and https://github.com/codalab/codalab-worksheets/pull/3986

epicfaace:
@yuchenWYC looks like the github actions docker build action is failing...

docker_config/dockerfiles/Dockerfile.default-gpu
yuchenWYC: For this line I'm not 100% sure, since I tried a bunch of mixed things and kept the code that finally worked

jzwang43: these changes will cause the GitHub actions to fail i think, I don't think the changes for default-gpu and default.cpu are needed since they have nothing to do with `start -bd`

jzwang43:
@epicfaace I think the root cause is that there's no 3.6 for 16.04 from deadsnake ppa anymore, that's why we switched to the legacy ppa in #3954.

However, that legacy ppa is only for x64 not for arm64 like apple m1, which is causing issue for m1 users.

jzwang43:
Closed since this only happens for apple m1.

PR #4045, opened 2022-03-07
Fix New bundle is added at the random place 

### LINKED ISSUE #3989: New bundle is added at a random place
### Implementation
- [x] Prevent a worksheet row from having a `null` sort key when rows are created using the cli
    * **Note:** cl commands should always add bundles to the end of the worksheet
- [x] Check for `null` sort keys when we initially fetch a worksheet and fix them on the spot if needed
- [x] Code documentation
- [x] Write tests if needed
- [x] QA
    - [x] Ensure that source editing is not impacted
    - [x] Test large worksheets that have lots of rows with null sort keys
    - [x] Test a worksheet with a mixture of bundles created in web and bundles created in CLI
    - [x] Remove test console logs 

### Background

The bundle should be added at the same place where the user has selected. Currently, it is getting placed at random places in the worksheet.

https://user-images.githubusercontent.com/10134824/152871649-3a46f325-000f-41e6-853b-421011812961.mov

jzwang43:

The bundle should be added at the same place where the user has selected. Currently, it is getting placed at random places in the worksheet.

The main reason this happens is because some sort_keys are null for some items for some reason. 
This PR should be able to resolve the issue from the front end for now, while we do more research to figure out the root cause.

The logic is to simply re-save the worksheet so the sort_keys are fixed by the back-end. This allows us to avoid any migration of existing user data.

### Related issues

Fixes #3989 

### Checklist

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1607,7 +1607,7 @@ class BundleModel(object):
             clause = cl_worksheet_item.c.worksheet_uuid == worksheet_uuid
             query = select([func.max(cl_worksheet_item.c.sort_key)]).where(clause)
             max_row = connection.execute(query).fetchone()
-            return -1 if max_row is None else max_row[0]
+            return -1 if (max_row is None or max_row[0] is None) else max_row[0]
 
     def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replace=[]):
         """
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1616,6 +1616,14 @@ class BundleModel(object):
             if len(items) == 0:
                 # Nothing to insert, return
                 return
+
+            # if after_sort_key not specified, insert the bundle to the end of the worksheet
+            if after_sort_key is None:
+                clause = cl_worksheet_item.c.worksheet_uuid == worksheet_uuid
+                query = select([func.max(cl_worksheet_item.c.sort_key)]).where(clause)
+                max_row = connection.execute(query).fetchone()
+                after_sort_key = -1 if max_row is None else max_row[0]
+
             if after_sort_key is not None:
                 after_sort_key = int(after_sort_key)
                 # Shift existing items' sort_keys for items that originally came after
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1617,7 +1617,7 @@ class BundleModel(object):
                 # Nothing to insert, return
                 return
 
-            # if after_sort_key not specified, insert the bundle to the end of the worksheet
+            # if after_sort_key not specified, insert the bundle at the end of the worksheet
             if after_sort_key is None:
                 clause = cl_worksheet_item.c.worksheet_uuid == worksheet_uuid
                 query = select([func.max(cl_worksheet_item.c.sort_key)]).where(clause)
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1602,6 +1602,13 @@ class BundleModel(object):
             result = connection.execute(cl_worksheet.insert().values(worksheet_value))
             worksheet.id = result.lastrowid
 
+    def get_max_sort_key(self, worksheet_uuid):
+        with self.engine.begin() as connection:
+            clause = cl_worksheet_item.c.worksheet_uuid == worksheet_uuid
+            query = select([func.max(cl_worksheet_item.c.sort_key)]).where(clause)
+            max_row = connection.execute(query).fetchone()
+            return -1 if max_row is None else max_row[0]
+
     def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replace=[]):
         """
         Add worksheet items *items* to the position *after_sort_key* to the worksheet,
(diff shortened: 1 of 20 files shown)


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -138,6 +138,17 @@ class Worksheet extends React.Component {
                         info,
                     },
                 });
+                if (info.blocks) {
+                    for (let block of info.blocks) {
+                        if (
+                            block.sort_keys &&
+                            Object.values(block.sort_keys).some((sort_key) => sort_key == null)
```
epicfaace: Add a comment mentioning this is a temporary solution, and put a link to a github issue that describes this issue (make the github issue if needed).

Great that you're able to fix it! Is there a chance this might cause an infinite loop though?

edit: never mind, I see there's no way to have an infinite loop, but it's just inefficient

percyliang: Isn't this really inefficient because you have to save the entire worksheet for each block?  At least you could just save it once?  But this is definitely not satisfying because we really want to only save the relevant block, not the whole worksheet.  This is really important for long worksheets, and this will cause a huge performance degradation.  Ideally, we'd get to the root cause and fix that.

jzwang43: @epicfaace @percyliang 
I think the root cause of this issue is:
If a user runs something like cl run "echo 2", this will be added to a worksheet without sort_key (null). This messes up the frontend logic to determine where to insert bundle.

I think the fix should be 
1. From the backend, add it to the beginning of a worksheet if it's added from CLI without -a.
2. for existing worksheet, we keep my current fix so they will get re-parsed and the sort key gets fixed.

Just pushed the fix, let me know what you think

percyliang: 1. Wait, `cl` commands should always add bundles to the end of the worksheet, not the beginning.
2. If we just make sure everything gets a sort key when bundles are added from the backend, then why do we need to do anything on the frontend?

jzwang43: 1. I updated the PR to insert it to the end instead of the beginning.
2. It's because existing worksheet will still have bundles will null sort key. The re-save will just happen once when they open those worksheets for the first time so those nulls get fixed. Or we can manually fix existing ones.

percyliang: I see - it's just a one time thing.  I think it's cleanest to just do a migration to fix the nulls.  Or if not that, then doing it on the backend seems like it'd make more sense (i.e., whenever we fetch a worksheet, if there are nulls, we fix it on the spot).

epicfaace: > then doing it on the backend seems like it'd make more sense (i.e., whenever we fetch a worksheet, if there are nulls, we fix it on the spot).

I like this approach.

codalab/model/bundle_model.py
```diff
@@ -1616,6 +1616,14 @@ def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replac
             if len(items) == 0:
                 # Nothing to insert, return
                 return
+
+            # if after_sort_key not specified, insert the bundle to the end of the worksheet
```
```suggestion
            # if after_sort_key not specified, insert the bundle at the end of the worksheet
```

epicfaace: What was the previous behavior if after_sort_key was not specified?

codalab/rest/interpret.py
```diff
@@ -245,6 +246,14 @@ def fetch_interpreted_worksheet(uuid):
     # Fetch items.
     worksheet_info['source'] = get_worksheet_lines(worksheet_info)
 
+    for item in worksheet_info['items']:
```
epicfaace: Add a comment explaining what this does.

codalab/rest/interpret.py
```diff
@@ -245,6 +246,14 @@ def fetch_interpreted_worksheet(uuid):
...
+        if item[-1] is None:
+            lines = worksheet_info['source']
+            new_items = parse_worksheet_form(lines, local.model, request.user, uuid)
+            worksheet_info = get_worksheet_info(uuid, fetch_items=True)
+            update_worksheet_items(worksheet_info, new_items)
```
epicfaace: Rather than re-saving the worksheet, can we just assign it an arbitrary sort key that is at the bottom of the worksheet? Or would that cause a problem?

codalab/model/bundle_model.py
jzwang43: it's just null

codalab/rest/interpret.py
jzwang43: yeah that should work too, let me update the code, rather than arbitrary we can use the `max_sort_key + 1`

jzwang43: updated, let me know what you think

jzwang43: @epicfaace can I get a review on the new approach? thanks!

codalab/model/bundle_model.py
```diff
@@ -1616,6 +1623,11 @@ def add_worksheet_items(self, worksheet_uuid, items, after_sort_key=None, replac
             if len(items) == 0:
                 # Nothing to insert, return
                 return
+
+            # if after_sort_key not specified, insert the bundle at the end of the worksheet
```
percyliang, line 1627: Provide more comments / context for this: going forward, this code should not be triggered, etc.

leilenah:
Fixed here https://github.com/codalab/codalab-worksheets/pull/4168

PR #4046, opened 2022-03-07
Fix JSON Display

### LINKED ISSUE #4037: Screen goes blank when looking at a JSON bundle
To reproduce, visit this page:

https://codalab.stanford.edu/bundles/0x6e3abaeb28e34640a12a9a66d27d8ddf

jzwang43:

Bundle detail screen goes blank sometimes if the file is JSON.

### Related issues

Fixes #4037 

### Screenshots

image

### Checklist

### DIFF
--- a/frontend/src/components/Bundle/Bundle.js
+++ b/frontend/src/components/Bundle/Bundle.js
@@ -494,7 +494,7 @@ function renderContents(bundleInfo, fileContents, stdout, stderr) {
                     </span>
                 </div>
                 <div className='collapsible-content bundle-meta'>
-                    <pre>{JSON.stringify(fileContents, null, 2)}</pre>
+                    <pre>{fileContents}</pre>
                 </div>
             </div>
         );
--- a/frontend/src/components/Bundle/Bundle.js
+++ b/frontend/src/components/Bundle/Bundle.js
@@ -102,7 +102,12 @@ class Bundle extends React.Component<
             if (!info) return;
             if (info.type === 'file' || info.type === 'link') {
                 return fetchFileSummary(this.props.uuid, '/').then((blob) => {
-                    this.setState({ fileContents: blob, stdout: null, stderr: null });
+                    this.setState({
+                        fileContents:
+                            typeof blob === 'object' ? JSON.stringify(blob, null, 2) : blob,
+                        stdout: null,
+                        stderr: null,
+                    });
                 });
             } else if (info.type === 'directory') {
                 // Get stdout/stderr (important to set things to null).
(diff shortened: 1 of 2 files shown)


frontend/src/components/Bundle/Bundle.js
```diff
@@ -102,7 +102,12 @@ class Bundle extends React.Component<
             if (!info) return;
             if (info.type === 'file' || info.type === 'link') {
                 return fetchFileSummary(this.props.uuid, '/').then((blob) => {
-                    this.setState({ fileContents: blob, stdout: null, stderr: null });
+                    this.setState({
+                        fileContents:
```
epicfaace: I think the cleanest way to do this would be to pass in `responseType: 'text'` when we call `get` from axios in the first place -- then it won't parse as json in the first place.

epicfaace (changes requested):

frontend/src/components/Bundle/Bundle.js
jzwang43: Hi, Ashwin thanks for the suggestion. Just applied.

frontend/src/util/apiWrapper.js
```diff
@@ -129,7 +130,14 @@ export const fetchFileSummary = (uuid, path) => {
     };
     const url =
         '/rest/bundles/' + uuid + '/contents/blob' + path + '?' + new URLSearchParams(params);
-    return get(url, { headers: { Accept: 'text/plain' } });
+    return get(
+        url,
+        {
+            headers: { Accept: 'text/plain' },
+        },
+        // need to define transformResponse to prevent json being parsed due to the issue discussed in: https://github.com/axios/axios/issues/811
```
epicfaace:
```suggestion
        // Get the raw text response and don't parse JSON. Need to define transformResponse to prevent JSON being parsed due to the issue discussed in: https://github.com/axios/axios/issues/811
```

frontend/src/util/apiWrapper.js
```diff
@@ -2,9 +2,10 @@ import { Semaphore } from 'await-semaphore';
 import axios from 'axios';
 import { createDefaultBundleName, pathIsArchive, getArchiveExt } from './worksheet_utils';
 
-export const get = (url, params) => {
+export const get = (url, params, options) => {
```
epicfaace:
```suggestion
export const get = (url, params, options={}) => {
```

epicfaace (approved):

frontend/src/util/apiWrapper.js
jzwang43: @epicfaace is this needed? the spread operator is capable of handling undefined i think

epicfaace: Not needed, but it would make it clearer that this argument is optional if someone reads the function definition.

epicfaace (approved):

PR #4061, opened 2022-04-05
Fix/3911-cl-store infer storage type from url

wwwjn:
Fix issue #3911 
#### Design choices
Infer storage type from url in bundle_cli.py rather than bundles.py line 514, because the field `storage_type` database model `BundleStoreSchema`  can not be NULL, so we need to infer the storage type before insert into database. And we need to know the storage type as soon as possible so I put it in cl command.

#### Changes
1. Add auto test cases in test_cli.py
2. Infer storage type from url in bundle_cli.py
 
### Related issues

#3911 

### Screenshots
N/A

### Checklist


tests/cli/test_cli.py
```diff
@@ -1077,7 +1077,64 @@ def test_upload_default_bundle_store(ctx):
...
+    # create a new azure_blob but specify wrong type
+    blob_id = _run_command(
+        [
+            cl,
+            "store",
+            "add",
+            "--name",
+            bundle_store_name,
+            '--storage-type',
+            'disk',  # the type does not aligns with url
```
```suggestion
            'disk',  # the type does not align with url
```

epicfaace:
@wwwjn thanks, can you fix format?

wwwjn:
> @wwwjn thanks, can you fix format?

Yes! It seems that after run pre-commit.sh, I mistakenly change a file. Fixed now!

codalab/lib/bundle_cli.py
```diff
@@ -1054,6 +1054,21 @@ def do_store_command(self, args):
                 "url": args.url,
                 "authentication": args.authentication,
             }
+            if args.url is not None:
```
epicfaace, line 1058: Rather than hardcoding azfs://, azure_blob, etc., can we instead just enumerate through the constants defined here (https://github.com/codalab/codalab-worksheets/blob/545106e1bf9599f608597d9dcbe82391788d6dab/codalab/common.py#L171-L190)?

This will reduce code duplication.

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,65 @@ def test_upload_default_bundle_store(ctx):
...
+            bundle_store_name,
+            '--storage-type',
+            'azure_blob',
+            '--url',
+            'azfs://devstoreaccount1/bundles',
+        ]
+    )
+    check_contains("azure_blob", _run_command([cl, "store", "ls"]))
+    _run_command([cl, "store", "rm", blob_id])
+    # TODO: set CODALAB_GOOGLE_APPLICATION_CREDENTIALS and test gcp blob storage
```
epicfaace: Since we're just running `cl store add`, we don't actually need to connect to GCS so we don't need to set CODALAB_GOOGLE_APPLICATION_CREDENTIALS here.

epicfaace (commented):
Nice work! Small update to make the code cleaner

codalab/lib/bundle_cli.py
wwwjn: Thanks Ashwin, already changed to use common classes.

tests/cli/test_cli.py
wwwjn: Ohh yes, so I will add some similar test cases of GCS

codalab/lib/bundle_cli.py
epicfaace: Can we simplify this further? We can just loop through the values of `StorageURLScheme` and `StorageType` -- or we might need to define a dict in common.py mapping `StorageURLScheme` to `StorageType`

wwwjn: Cool! I will define a dict in common.py to make it more clear, in case there are some new type of storage

codalab/common.py
```diff
@@ -190,6 +190,27 @@ class StorageURLScheme(Enum):
     GCS_STORAGE = "gs://"
 
 
+# A Dict mapping to
```
percyliang: Incomplete comment?

codalab/common.py
```diff
@@ -190,6 +190,27 @@ class StorageURLScheme(Enum):
...
+storage_url_type_dict = {
+    StorageURLScheme.DISK_STORAGE: StorageType.DISK_STORAGE,
+    StorageURLScheme.AZURE_BLOB_STORAGE: StorageType.AZURE_BLOB_STORAGE,
+    StorageURLScheme.GCS_STORAGE: StorageType.GCS_STORAGE,
+}
+
+
+def storage_url_to_type(url):
```
percyliang: Add type annotation

codalab/common.py
```diff
@@ -190,6 +190,27 @@ class StorageURLScheme(Enum):
...
+    """Find corresponding StorageType by StorageURLScheme.
+    If can not find, return DISK_STORAGE by default.
```
percyliang: Fix grammar: If we cannot find it, return `DISK_STORAGE`.

codalab/common.py
```diff
@@ -190,6 +190,27 @@ class StorageURLScheme(Enum):
...
+# A Dict mapping to
+storage_url_type_dict = {
+    StorageURLScheme.DISK_STORAGE: StorageType.DISK_STORAGE,
+    StorageURLScheme.AZURE_BLOB_STORAGE: StorageType.AZURE_BLOB_STORAGE,
+    StorageURLScheme.GCS_STORAGE: StorageType.GCS_STORAGE,
+}
+
+
+def storage_url_to_type(url):
+    """Find corresponding StorageType by StorageURLScheme.
```
percyliang: This comment is a bit cryptic - can you expand on it and relate to `url`, perhaps with an example?

percyliang: "Return the storage type (e.g., ...) corresponding to the storage URL scheme (e.g., ...) given by `url` (e.g., ...).

codalab/common.py
```diff
@@ -190,6 +190,27 @@ class StorageURLScheme(Enum):
...
+    If can not find, return DISK_STORAGE by default.
+    """
+    if url is None:
+        return StorageType.DISK_STORAGE.value
+    # Iterate over non-disk storage type
```
percyliang: Fix grammar: Iterate over non-disk storage types

codalab/common.py
wwwjn: Done! Thanks Percy, I will pay more attention to comments.

codalab/common.py
```diff
@@ -190,6 +190,28 @@ class StorageURLScheme(Enum):
     GCS_STORAGE = "gs://"
 
 
+# A Dict mapping from StorageURLSchema to StorageType
```
epicfaace:
```suggestion
# A Dict mapping from StorageURLScheme to StorageType
```

codalab/common.py
```diff
@@ -190,6 +190,28 @@ class StorageURLScheme(Enum):
...
+storage_url_type_dict = {
+    StorageURLScheme.DISK_STORAGE: StorageType.DISK_STORAGE,
+    StorageURLScheme.AZURE_BLOB_STORAGE: StorageType.AZURE_BLOB_STORAGE,
+    StorageURLScheme.GCS_STORAGE: StorageType.GCS_STORAGE,
+}
+
+
+def storage_url_to_type(url: str):
```
epicfaace: I think we should replace this method with just `parse_linked_bundle_url(url).storage_type`

epicfaace (changes requested):

codalab/common.py
wwwjn: @epicfaace Yeah If we can merge these functions together, it will be more clear. But for the function parse_linked_bundle_url, it seems that it only parses: an archive file on Blob Storage: "azfs://storageclwsdev0/bundles/uuid/contents.tar.gz". And when adding a storage using url (eg, "azfs://devstoreaccount1/bundles"), we do not know the uuid and file name (We only know after uploading some files). Do you think we should modify `parse_linked_bundle_url()` to parse  or still use `storage_url_to_type()` function?

epicfaace: I think we should modify `parse_linked_bundle_url` to parse other storage type urls as well -- because eventually we'll need to do that anyway to support these other bundle stores. If it's just a URL "azfs://devstoreaccount1/bundles", the scheme parsing should still work, so it should be OK. We might want to add that fact to the parse_linked_bundle_url documentation and update the parse_linked_bundle_url unit tests, though.

wwwjn: Thanks Ashwin! Now it will failed because of  this line `storage_account, container, bundle_uuid, contents_file, *remainder = url.split("/", 4)`. Because for  url like this "azfs://devstoreaccount1/bundles", we do not have enough parameter to parse. I will modify `parse_linked_bundle_url` to support parsing short url and modify unitests

codalab/common.py
```diff
@@ -279,18 +280,27 @@ def parse_linked_bundle_url(url):
...
             url = url[len(StorageURLScheme.GCS_STORAGE.value) :]
-            bucket_name, bundle_uuid, contents_file, *remainder = url.split("/", 3)
-            bundle_path = (
-                f"{StorageURLScheme.GCS_STORAGE.value}{bucket_name}/{bundle_uuid}/{contents_file}"
-            )
-
-        is_archive = contents_file.endswith(".gz") or contents_file.endswith(".tar.gz")
-        is_archive_dir = contents_file.endswith(".tar.gz")
+            bucket_name, *remainder = url.split("/", 1)
+            if len(remainder) and len(remainder[0].split("/")) >= 2:
```
epicfaace: this feels pretty complex / difficult to read through.

Maybe it would be simpler if we just did:

```
try:
    bucket_name, bundle_uuid, contents_file, *remainder = url.split("/", 3)
except ValueError:
    # url refers to bucket, e.g. gs://{bucket_name}
    bucket_name, *remainder = url.split("/", 1)
    bundle_uuid, contents_file, remainder = None, None, []
```

What do you think?

codalab/lib/bundle_cli.py
epicfaace: Feel feel to resolve a conversation when we're done with it

codalab/common.py
epicfaace: This would also fit more with the EAFP style recommended for python 🙂 https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

wwwjn: Oh it's much more simpler if we catch the error then deal with it.  I will change it.

codalab/lib/bundle_cli.py
```diff
@@ -1054,6 +1055,14 @@ def do_store_command(self, args):
...
+                inferred_type = parse_linked_bundle_url(args.url).storage_type
+                if args.storage_type is None:
+                    bundle_store_info["storage_type"] = inferred_type
+                elif args.storage_type != inferred_type:
+                    raise UsageError(
+                        f"cl store storage type '{args.storage_format}' conflicts with storage URL."
```
epicfaace: Let's say "Bundle store {url} only supports storage type: {inferred_type}"

codalab/lib/bundle_cli.py
```diff
@@ -1054,6 +1055,14 @@ def do_store_command(self, args):
                 "url": args.url,
                 "authentication": args.authentication,
             }
+            if args.url is not None:
+                inferred_type = parse_linked_bundle_url(args.url).storage_type
+                if args.storage_type is None:
+                    bundle_store_info["storage_type"] = inferred_type
```
epicfaace, line 1061: Is bundle_store_info["storage_type"] set when args.storage_type is specified?

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,102 @@ def test_upload_default_bundle_store(ctx):
     check_contains(bundle_store_name, _run_command([cl, "info", uuid]))
 
 
+@TestModule.register('store_add')
+def test_store_add(ctx):
+    """
+    Tests of command `cl store add` on different blob storage. (Not test --storage-format yet)
+    1. Specify storage type and url
```
epicfaace: I think we can remove 1./2./3., probably too much detail

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,102 @@ def test_upload_default_bundle_store(ctx):
     check_contains(bundle_store_name, _run_command([cl, "info", uuid]))
 
 
+@TestModule.register('store_add')
+def test_store_add(ctx):
+    """
+    Tests of command `cl store add` on different blob storage. (Not test --storage-format yet)
```
```suggestion
    Tests of command `cl store add` on different bundle stores. (Not testing --storage-format yet)
```

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,102 @@ def test_upload_default_bundle_store(ctx):
...
+            '--storage-type',
+            'azure_blob',
+            '--url',
+            'azfs://devstoreaccount1/bundles',
+        ]
+    )
+    check_contains("azure_blob", _run_command([cl, "store", "ls"]))
+    _run_command([cl, "store", "rm", blob_id])
+
+    # create a new azure_blob but not specify storage type
```
epicfaace, line 1104:
```suggestion
    # Create a new azure_blob bundle store without specifying storage type
```

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,102 @@ def test_upload_default_bundle_store(ctx):
...
+            "--name",
+            bundle_store_name,
+            '--url',
+            'azfs://devstoreaccount1/bundles',
+        ]
+    )
+    check_contains("azure_blob", _run_command([cl, "store", "ls"]))
+    _run_command([cl, "store", "rm", blob_id])
+
+    # create a new azure_blob but specify wrong type
```
```suggestion
    # Create a new azure_blob bundle store and specify the wrong storage type
```

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,102 @@ def test_upload_default_bundle_store(ctx):
...
+            "--name",
+            bundle_store_name,
+            '--storage-type',
+            'disk',  # the type does not align with url
+            '--url',
+            'azfs://devstoreaccount1/bundles',
+        ],
+        expected_exit_code=1,
+    )
+    # Test 3 kinds of conditions on GCP storage
```
```suggestion
    # Test these 3 conditions on GCS
```

tests/cli/test_cli.py
```diff
@@ -1078,6 +1078,102 @@ def test_upload_default_bundle_store(ctx):
...
+    check_contains("gcs", _run_command([cl, "store", "ls"]))
+    _run_command([cl, "store", "rm", blob_id])
+
+    blob_id = _run_command(
+        [cl, "store", "add", "--name", bundle_store_name, '--url', 'gs://codalab-test',]
+    )
+    check_contains("gcs", _run_command([cl, "store", "ls"]))
+    _run_command([cl, "store", "rm", blob_id])
+
+    # create a new gcs storage but specify wrong storage-type
```
epicfaace: rm this comment because the previous comment ("Test 3 kinds of conditions on GCP storage") already covers it

epicfaace (approved):

codalab/lib/bundle_cli.py
wwwjn: I think No? When storage_type is specified, it should have higher priority than inferred_type. And if a user use a wrong command like "--storage_type azure_blob --url gs://bucket", we will raise Error. If the args.storage_type is specified and equals inferred_type, it's ok not to set it.

PR #4070, opened 2022-04-19
Use signed URL to bypass server for GCS downloads

### LINKED ISSUE #3974: Extend SAS token logic to bypass server for GCS downloads
Right now we only support generating SAS tokens for Azure Blob downloads (and soon uploads). We should extend this logic to generate signed URLs that can be used in GCS as well.

wwwjn:
When user download bundles stored in GCS, need to bypass server to speed up the downloading process.
For now, we only support the following situation to bypass server.
image

We can not support bypassing server when download a single-file bundle. This is because GCS can not support transparent decompression without setting GCS object metadata. For a detailed test process and explanation, please refer to: https://docs.google.com/document/d/1q2zTXFPlkjnm-xm6_IcfenAm5twBsYOIEdopxEHCbx0/edit#

### Related issues
Finish part of the issue #3974. Still need to bypass server when uploading file to GCS.

### Screenshots
By inspecting the frontend, we can see the redirected url and request/response when bypass server downloading.
test3

### Checklist


codalab/common.py
```diff
@@ -258,12 +260,46 @@ def _get_sas_url(self, path, **kwargs):
         )
         return f"{AZURE_BLOB_HTTP_ENDPOINT}/{AZURE_BLOB_CONTAINER_NAME}/{blob_name}?{sas_token}"
 
+    def _get_signed_url(self, path, **kwargs):
+        """Generate GCS signed url that can be used to download the blob for 1 hour."""
+        print(kwargs, type(kwargs))
```
epicfaace: rm this

epicfaace (commented):
Should we add a test in `test_cli.py`? Maybe we should test it with a real GCS bucket.

wwwjn:
> 

Yes I'm also trying to do this. I noticed that there are a lot of test cases about Azure but only a few about GCS. I'm considering how to use credential to write some test cases and test a real GCS bucket.

epicfaace:
- See if we can monkey-patch GcsIO to allow using the local client, and if the e2e tests pass.
- If they work, then add the e2e tests. Merge this PR.
- Otherwise, let's not add the e2e tests. Make an issue to add GCS e2e tests. Maybe also make issues in the upstream open source repos (gcs-emulator / apache beam). Merge this PR.

epicfaace (changes requested):
Let's do a manual test where we set:
- CODALAB_DEFAULT_BUNDLE_STORE_NAME=gcs
- CODALAB_GOOGLE_APPLICATION_CREDENTIALS
and we should run these tests: https://github.com/codalab/codalab-worksheets/blob/master/.github/workflows/test.yml#L423-L434


Add a section to the [testing page](https://codalab-worksheets.readthedocs.io/en/latest/Server-Setup/#testing) called "manual tests" and then include the instructions to run this GCS test.

Don't add to github actions / CI.

wwwjn:
Add manual local tests to test directly using GCS, and the test result are in `.out` files: (have removed these file in final PR)
https://github.com/codalab/codalab-worksheets/pull/4070/commits/1fad99c58a40db772f2ceef0b031841628904756

Test Summary:
- Using GCS: Tests failed: upload2, blob, default_bundle_store, uls, time, link, worksheets, protected_mode
-  Not using GCS: Tests failed: blob, uls, time, link, worksheets, protected_mode

Analysis: 
- Fail upload2: Because test `upload2` checks the response header (need `Content-Encoding='identity'`), but the GCS response header does not contain `content-encoding`.
- Fail default_bundle_store:  the `default_bundle_store` test is expected to fail because it wants to create a new disk bundle storage that has the same name as our manually created one.

epicfaace:
Hmm, why are `blob, uls, time, link, worksheets, protected_mode` failing?

wwwjn:
> 

They failed because of the same reason `WARNING:retry.api:HTTP Error 403: Forbidden, retrying in 1 seconds...` Because they receive HTTP Error 403 after retry. I haven't look into detail about this test, but when I first set up my codalab, these tests fail. (It also might be because I use some wrong parameters..)

docs/Server-Setup.md
```diff
@@ -186,6 +186,31 @@ In sum, to start an instance and run tests on it:
...
+2. Set the environment variable:
+
+    export CODALAB_GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentail.json
+
+3. Start the codalab server, and manually create a bundle store using GCS.
+
+    ./codalab_service start -bds default
+    cl work http:://localhost::   # make sure it works using local instance
+    cl store add --name {your_bundle_store_name} --url gs://{test_bucket_name}
+    cl store ls  # make sure you have succcessfully add the
```
epicfaace: fix comment

epicfaace (approved):

docs/Server-Setup.md
epicfaace: @wwwjn can you update this comment, and then merge?

epicfaace:
@wwwjn feel free to merge (or add the `ready-to-merge` tag)!

PR #4073, opened 2022-04-19
Modify "Multiple-Bundle-Storage" docs and add doc for inferring store type

wwwjn:

To record the modification of the `cl store add` command usage.

### Checklist

### DIFF
--- a/docs/Multiple-Bundle-Stores.md
+++ b/docs/Multiple-Bundle-Stores.md
@@ -32,7 +32,7 @@ cl store rm [bundle store uuid]
 First, make sure you start Azurite locally by running `codalab-service start -bds default azurite`. Then run:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/Multiple-Bundle-Stores.md
```diff
@@ -62,6 +62,19 @@ Once a bundle store is created, you can restart the server and set the `CODALAB_
 CODALAB_DEFAULT_BUNDLE_STORE_NAME=blob codalab-service start ...
 ```
 
+## Infer store type from URL
```
epicfaace: I don't think we need a new section for this. Can we just modify the previous commands so we remove `--storage-type gcs`, etc. from the commands?

epicfaace (approved):

PR #4077, opened 2022-04-26
Fix problem that does not copy exclude-patterns when rerun a bundle

### LINKED ISSUE #3238: "exclude_patterns" is not copied over when rerunning a bundle
**Describe the bug**
`exclude_patterns` is not copied over when rerunning a bundle on the UI.

**To Reproduce**
1. Create a bundle and specify a value for `exclude_patterns`
2. Rerun the bundle from step 1
3. Notice that `exclude_patterns` is not specified for the new bundle

**Expected behavior**
I would expect the value of the field to get copied over.

wwwjn:
When rerun a bundle, the `exclude-patterns` parameter is not copied.

### Related issues

Fix #3238 

### Screenshots
test4

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/NewRun/NewRun.jsx
+++ b/frontend/src/components/worksheets/NewRun/NewRun.jsx
@@ -273,6 +273,7 @@ class NewRun extends React.Component<
             networkAccess,
             failedDependencies,
             queue,
+            exclude_patterns,
         } = this.state;
         const { after_sort_key } = this.props;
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -300,6 +301,14 @@ class NewRun extends React.Component<
 
         if (command) args.push(command);
 
+        // exclude_patterns a list of arguments, so put it at the end of commands
+        if (exclude_patterns) {
+            args.push(`--exclude-patterns`);
+            for (let i = 0; i < exclude_patterns.length; i++) {
```
epicfaace: We can just replace this with

```
args.push(...exclude_patterns);
```

See https://stackoverflow.com/questions/1374126/how-to-extend-an-existing-javascript-array-with-another-array-without-creating

frontend/src/components/worksheets/NewRun/NewRun.jsx
```diff
@@ -300,6 +301,14 @@ class NewRun extends React.Component<
 
         if (command) args.push(command);
 
+        // exclude_patterns a list of arguments, so put it at the end of commands
```
epicfaace:
```suggestion
        // exclude_patterns can take a list of arguments, so put it at the end
```

epicfaace (approved):

PR #3949, opened 2022-01-16
Worker manager fixes for cs324

epicfaace:
### Reasons for making this change

Fixes this issue when starting up codalab:

```
Traceback (most recent call last):
  File "/usr/local/bin/cl-worker-manager", line 33, in <module>
    sys.exit(load_entry_point('codalab', 'console_scripts', 'cl-worker-manager')())
  File "/opt/codalab-worksheets/codalab/worker_manager/main.py", line 141, in main
    worker_manager_type.add_arguments_to_subparser(worker_manager_subparser)
  File "/opt/codalab-worksheets/codalab/worker_manager/slurm_batch_worker_manager.py", line 77, in add_arguments_to
_subparser
    '--user', type=str, default=getpass.getuser(), help='User to run the Batch jobs as'
  File "/usr/lib/python3.6/getpass.py", line 169, in getuser
    return pwd.getpwuid(os.getuid())[0]
KeyError: 'getpwuid(): uid not found: 1003'
Traceback (most recent call last):
  File "/usr/local/bin/cl-worker-manager", line 33, in <module>
    sys.exit(load_entry_point('codalab', 'console_scripts', 'cl-worker-manager')())
  File "/opt/codalab-worksheets/codalab/worker_manager/main.py", line 141, in main
    worker_manager_type.add_arguments_to_subparser(worker_manager_subparser)
  File "/opt/codalab-worksheets/codalab/worker_manager/slurm_batch_worker_manager.py", line 77, in add_arguments_to
_subparser
    '--user', type=str, default=getpass.getuser(), help='User to run the Batch jobs as'
  File "/usr/lib/python3.6/getpass.py", line 169, in getuser
    return pwd.getpwuid(os.getuid())[0]
KeyError: 'getpwuid(): uid not found: 1003'
```

### Checklist

* [ ] I've added a screenshot of the changes, if this is a frontend change
* [ ] I've added and/or updated tests, if this is a backend change
* [ ] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [ ] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

PR #4088, opened 2022-04-30
Allow running a local kubernetes cluster using kind

epicfaace:
Allow running a local kubernetes cluster using kind (https://kind.sigs.k8s.io/). kind is a lightweight way to run kubernetes, which runs through a Docker container itself. This way, we can continue to use the Docker daemon for local development and not have to install additional things (for example, other local kubernetes deployments such as minikube require you to install and run a VM)


epicfaace:
Whoever reviews this, can you try the steps yourself? Note that while the cluster can communicate with codalab and run workers, it doesn't have access to Docker so it can't run actual bundles yet. https://github.com/codalab/codalab-worksheets/pull/4090 allows workers in the cluster to also launch jobs.

codalab/worker_manager/kubernetes_worker_manager.py
```diff
@@ -74,6 +74,12 @@ def __init__(self, args):
         configuration.api_key['authorization'] = args.auth_token
         configuration.host = args.cluster_host
         configuration.ssl_ca_cert = args.cert_path
+        if configuration.host == "https://codalab-control-plane:6443":
```
epicfaace, line 74: Allows anonymous access to a local cluster without SSL checks, for testing / development purposes only. I did this because it was getting really complicated to provision the right SSL certs -- we might in the future see if there's an automated / easy way to provision SSL certs for communication with the local cluster.

teetone: Makes sense, should we add a comment in the code?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -32,6 +32,11 @@ class SlurmBatchWorkerManager(WorkerManager):
 
     @staticmethod
     def add_arguments_to_subparser(subparser):
+        try:
+            user_id = getpass.getuser()
+        except Exception:
+            # Sometimes getpass.getuser() doesn't work.
```
teetone, line 38: Does it not work on the cluster? Should we error out instead?

docs/Server-Setup.md
```diff
@@ -327,3 +327,68 @@ If you need to send Slack notifications from monitor.py service, you can configu
    Slack email address which will show up in a designated Slack channel.
 * Note that this integration only works with workspaces on *the Slack Standard Plan and above*.
 
+
+## Start a local Kubernetes Batch Worker Manager (with kind, for testing / development only)
+
+If you want to test or develop with kubernetes locally, follow these steps to do so:
```
teetone, line 363: nit: capitalize Kubernetes

teetone (approved):
Thanks for making Kubernetes development more convenient!

codalab/worker_manager/slurm_batch_worker_manager.py
epicfaace: This error always comes up when I try to run a worker manager locally. I had to make the same change to get cs324 worker managers to work (https://github.com/codalab/codalab-worksheets/pull/3949) so I thought I might as well add it in this PR.

codalab/worker_manager/kubernetes_worker_manager.py
epicfaace: Yes, done

PR #4096, opened 2022-05-05
Use websockets to make on-demand worker file previews faster

### LINKED ISSUE #4084: Improve STDOUT for users
Currently, a user has to wait several minutes before seeing an output. Improve STDOUT logic to display output instantly, like in real-life code debugging. We tried to solve this earlier via #4023

epicfaace:
Use websockets to make on-demand worker file previews faster. File previews (such as loading stdout) of a worker that is currently running a bundle now take ~0.5 seconds.

Fixes https://github.com/codalab/codalab-worksheets/issues/4084. This PR is basically a POC of my comment https://github.com/codalab/codalab-worksheets/issues/4084#issuecomment-1117271699 -- I've added a thin websocket layer so that when a user requests to view one of the worker's files:
- the server will then ping the worker through the websocket
- the worker will then check in
- the worker will then send back the file.

This also gives us the latitude to change the worker's default checkin frequency from 5 seconds -> 20 seconds, further decreasing the load on the rest server.

TODOs:
- don't hardcode the `ws-server` URL
- test this on dev, with ~20 workers
- make sure worker websocket listening thread code is thread-safe


codalab/worker/main.py
```diff
@@ -138,7 +138,7 @@ def parse_args():
         '--checkin-frequency-seconds',
         help='Number of seconds to wait between worker check-ins',
         type=int,
-        default=5,
+        default=20,
```
percyliang: I still think it would be nice to update the worker status at least once every 5 seconds...we don't need this to be 20, right?

epicfaace: No -- we can keep it at 5

epicfaace:
image

codalab/worker/worker.py
```diff
@@ -265,18 +269,54 @@ def check_num_runs_stop(self):
...
+                    f"ws://ws-server:2901/worker/{self.id}", max_queue=1
+                ) as websocket:
+
+                    async def receive_msg():
+                        await websocket.send("a")
+                        data = await websocket.recv()
+                        logging.warn(
+                            f"Got websocket message, got data: {data}, going to check in now."
+                        )
+                        self.checkin()
```
epicfaace, line 305: For concurrency:
- hold a global lock when running checkin()
- also hold the lock in the worker loop 1) when running checkin() 2) when processing bundles

codalab/bin/ws_server.py
```diff
@@ -0,0 +1,77 @@
...
+worker_to_ws = {}
+
+
+async def rest_server_handler(websocket):
+    # Got a message from the rest server.
+    logger.warn(f"rest_server_handler")
+    print("RSH")
+    worker_id = await websocket.recv()
+    logger.warn(f"Got a message from the rest server, to ping worker: {worker_id}.")
+
```
epicfaace, line 24: document

docs/Server-Setup.md
```diff
@@ -43,6 +43,7 @@ Docker Container | Docker Image Used            | Purpose
  frontend        | `codalab/frontend:<version>` | Website (serves static pages)
  rest-server     | `codalab/server:<version>`   | REST API endpoint (used by website and CLI)
  bundle-manager  | `codalab/server:<version>`   | Schedules bundles to workers in the background
+ rest-server     | `codalab/server:<version>`   | Websocket API endpoint (used by workers)
```
epicfaace:
```suggestion
 ws-server     | `codalab/server:<version>`   | Websocket API endpoint (used by workers)
```

docs/Code-Overview.md
```diff
@@ -154,6 +154,40 @@ containers, and periodically report on the status of the runs.
 A worker also has to respond to various commands such as reading files in the
 bundle while it's running, killing bundles, etc.
 
+All data transfer between the worker and the server happens through a process known
```
epicfaace, line 157: @AndrewJGaut here's an explanation of my approach here, would love any feedback!

AndrewJGaut:
@epicfaace Can you please review the changes made since your last change? They are quite minor. Once you give the LGTM, I'll approve it and merge it

codalab/worker/worker.py
```diff
@@ -145,8 +145,9 @@ def __init__(
             bundle_runtime=bundle_runtime,
         )
 
-        # Lock ensures .checkin() is not run concurrently.
-        self._checkin_lock = Lock()
+        # Lock ensures listening thread and main thread don't simultaneously
+        # access the runs dictionary, thereby causing race conditions.
+        self._lock = RLock()
```
epicfaace, line 153: Why use RLock instead of Lock?

codalab/worker/worker.py
```diff
@@ -497,66 +499,76 @@ def checkin(self):
...
-                    self.kill(uuid)
-                elif action_type == 'mark_finalized':
-                    self.mark_finalized(uuid)
-                elif action_type == 'read':
-                    self.read(socket_id, uuid, response['path'], response['read_args'])
-                elif action_type == 'netcat':
-                    self.netcat(socket_id, uuid, response['port'], response['message'])
-                elif action_type == 'write':
-                    self.write(uuid, response['subpath'], response['string'])
+            if type(response) is not list:
```
epicfaace, line 505: What does this change do?

tests/cli/test_cli.py
```diff
@@ -1724,7 +1724,7 @@ def test_run(ctx):
     # Test that bundle fails when run without sufficient time quota
     _run_command([cl, 'uedit', 'codalab', '--time-quota', '2'])
     uuid = _run_command([cl, 'run', 'sleep 100000'])
-    wait_until_state(uuid, State.KILLED, timeout_seconds=60)
+    wait_until_state(uuid, State.KILLED, timeout_seconds=63)
```
epicfaace: why from 60 to 63?

epicfaace (commented):
lgtm

codalab/model/worker_model.py
```diff
@@ -360,7 +362,18 @@ def send_stream(self, socket_id, fileobj, timeout_secs):
 
         return False
 
-    def send_json_message(self, socket_id, message, timeout_secs, autoretry=True):
+    def _ping_worker_ws(self, worker_id):
+        async def ping_ws():
+            async with websockets.connect("ws://ws-server:2901/main") as websocket:
```
epicfaace: We need to ensure this isn't hardcoded.

codalab/worker/worker.py
```diff
@@ -268,18 +278,54 @@ def check_num_runs_stop(self):
...
         self.image_manager.start()
         if not self.shared_file_system:
             self.dependency_manager.start()
+
+        async def listen(self):
+            logging.warn("Started websocket listening thread")
+            while not self.terminate:
+                logging.warn(f"Connecting anew to: ws://ws-server:2901/worker/{self.id}")
+                async with websockets.connect(
+                    f"ws://ws-server:2901/worker/{self.id}", max_queue=1
```
epicfaace: We need to ensure this isn't hardcoded.

epicfaace:
@AndrewJGaut my main feedback is that ws-server:2901 shouldn't be hardcoded. we should ensure we make use of the `ws_port` configured in codalab_service.py. Once you add that, feel free to merge!

codalab/worker/worker.py
AndrewJGaut: It's just for elegance's sake.

RLock allows the lock to be acquired multiple times (provided it is also released precisely as many times at is acquired). This is desirable in this case because, to avoid race conditions, we must call self.process_runs _before_ the lock is released within the checkin() function. By using RLock, we allow the code to acquire the lock a second time so that we can run process_runs within checkin.

If we didn't do this, we could use a flag to tell process_runs we already acquired the lock and not acquire the lock in that case, but that seemed less elegant to me.

AndrewJGaut (approved):

PR #4108, opened 2022-05-20
Bypass upload for Azure blob storage

### LINKED ISSUE #3779: bypass codalab server for Azure blob storage uploads
Part of this megaissue: https://github.com/codalab/codalab-worksheets/issues/3799
Design doc: https://docs.google.com/document/d/1TOsO4gnxUm1XMmlGUb7BJ4Z6y_olFbv9APbx0-IEy6M/edit#

---

We should bypass the codalab server when doing blob storage uploads. First, the CLI should call `POST /bundles/{bundle_uuid}/locations/` in order to get a SAS token. Then it should use the SAS token to upload the .tar.gz and the index file directly to Blob Storage. Finally, it should tell the server that it's done (perhaps with another endpoint)?

wwwjn:
Bypass server upload for Azure bundle storage. Design doc: https://docs.google.com/document/d/1qFT6OWSUW5-2ZfnkK_q6P-Q-9sVe_fyBLX3UfEIWRiY/edit?pli=1#

### Related issues
#3779 

### Checklist


epicfaace:
Hi @wwwjn I saw https://github.com/codalab/codalab-worksheets/pull/4108/commits/2a654b1c0fcd9742ed730dfee813ec77c5db54e5, are you stuck on anything in this PR?

wwwjn:
> Hi @wwwjn I saw [2a654b1](https://github.com/codalab/codalab-worksheets/commit/2a654b1c0fcd9742ed730dfee813ec77c5db54e5), are you stuck on anything in this PR?

No, everything works fine and this PR is ready to review, except the upload process bar will stuck for a while. This is because we need to read the uploaded file and calculate `index.sqlite` file, which takes some time. The whole upload process looks like:
test4

codalab/client/json_api_client.py
```diff
@@ -663,9 +663,7 @@ def upload_contents_blob(self, bundle_id, fileobj=None, params=None, progress_ca
     @wrap_exception('Unable to get the locations of bundles')
     def get_bundles_locations(self, bundle_uuids):
         response = self._make_request(
-            method='GET',
-            path='/bundles/locations',
-            query_params=self._pack_params({'uuids': bundle_uuids}),
+            method='GET', path='/bundles', query_params=self._pack_params({'uuids': bundle_uuids}),
```
epicfaace: What is this change for? Do we need to change this?

codalab/client/json_api_client.py
```diff
@@ -713,3 +711,26 @@ def get_bundle_locations(self, bundle_uuid):
             method='GET', path='/bundles/{}/locations/'.format(bundle_uuid),
         )
         return response['data']
+
+    @wrap_exception('Unable to create the location of bundles')
+    def update_bundle_locations(self, bundle_uuid, bundle_store_uuid, params):
```
epicfaace: Rename to add_bundle_location or (something like "add location and prepare for upload")?

codalab/client/json_api_client.py
```diff
@@ -713,3 +711,26 @@ def get_bundle_locations(self, bundle_uuid):
...
+            path='/bundles/{}/locations/'.format(bundle_uuid),
+            data=self._pack_document(
+                [{'bundle_uuid': bundle_uuid, 'bundle_store_uuid': bundle_store_uuid}],
+                'bundle_locations',
+            ),
+            query_params=self._pack_params(params),
+        )
+        return response['data']
+
+    @wrap_exception("Unable to finilize the status of blob storge bundles")
```
epicfaace:
```suggestion
    @wrap_exception("Unable to finalize the status of blob storage bundles")
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +261,210 @@ def cleanup_existing_contents(self, bundle):
...
+                'bundle_stores',
+                params={
+                    'name': destination_bundle_store,
+                    'include': ['uuid', 'storage_type', 'url'],
+                },
+            )
+            bundle_store_uuid = storage_info['uuid']
+            if storage_info['storage_type'] in (
+                StorageType.AZURE_BLOB_STORAGE.value,
+                StorageType.GCS_STORAGE.value,
```
epicfaace:
```suggestion
 ```

File: codalab/lib/upload_manager.py [hunk 5]
```diff
@@ -258,3 +261,210 @@ def cleanup_existing_contents(self, bundle):
...
+        self,
+        bundle: Dict,
+        packed_source: Dict,
+        use_azure_blob_beta: bool,
+        destination_bundle_store=None,
+    ):
+
+        # By pass server upload:
+        # 1. If the user specify `-a`, upload to Azure blob storage
+        # 2. If the user specify `--store` and blob storage is on Azure or GCS
```
epicfaace:
```suggestion
        # 2. If the user specify `--store` and blob storage is on Azure
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +261,210 @@ def cleanup_existing_contents(self, bundle):
...
+            )
+            bundle_store_uuid = storage_info['uuid']
+            if storage_info['storage_type'] in (
+                StorageType.AZURE_BLOB_STORAGE.value,
+                StorageType.GCS_STORAGE.value,
+            ):
+                bypass_server = True
+
+        if bypass_server:
+            # Mimic the rest server behavior
```
epicfaace, line 426: Initialize

`Uploader(is_client=True, client=self._client)`


In `Uploader. _update_and_get_bundle_location`, if `is_client` is True, then it should call `add_bundle_location`, save the SAS URLs, then set them to the appropriate env variables right when uploading happens.

codalab/lib/upload_manager.py
```diff
@@ -258,3 +261,210 @@ def cleanup_existing_contents(self, bundle):
...
+        use_azure_blob_beta: bool,
+        destination_bundle_store=None,
+    ):
+
+        # By pass server upload:
+        # 1. If the user specify `-a`, upload to Azure blob storage
+        # 2. If the user specify `--store` and blob storage is on Azure or GCS
+        bypass_server = True if use_azure_blob_beta else False
+        bundle_store_uuid = None
+        if destination_bundle_store is not None and destination_bundle_store != '':
```
epicfaace, line 397: Remove support for -a, we only care about --store

codalab/lib/upload_manager.py
```diff
@@ -258,3 +261,210 @@ def cleanup_existing_contents(self, bundle):
...
+                packed_source['filename']
+            ):
+                unpack_before_upload = True
+                is_dir = source_ext in zip_util.ARCHIVE_EXTS_DIR
+            else:
+                unpack_before_upload = False
+                is_dir = False
+
+            params = {'need_sas': True, 'is_dir': is_dir}
+            data = self._client.update_bundle_locations(bundle['id'], bundle_store_uuid, params)[
```
epicfaace: Even if bypass_server is False, we should call update_bundle_locations nevertheless. If bundle_conn_str / index_conn_str are None, then we should just call the regular PUT endpoint to upload the bundle through the server.

wwwjn: > Even if bypass_server is False, we should call update_bundle_locations nevertheless. If bundle_conn_str / index_conn_str are None, then we should just call the regular PUT endpoint to upload the bundle through the server.

Fixed. Now all the uploads will first try to do bypass server upload. (except the user specify `--store {some_disk_storage}`

codalab/lib/upload_manager.py
wwwjn: > Initialize
> 
> `Uploader(is_client=True, client=self._client)`
> 
> In `Uploader. _update_and_get_bundle_location`, if `is_client` is True, then it should call `add_bundle_location`, save the SAS URLs, then set them to the appropriate env variables right when uploading happens.

I reuse the function write_fileobj() function and keep the `_update_and_get_bundle_location` the same.  It's not easy to reuse the Upload._update_and_get_bundle_location function because it uses a lot self._bundle_store/_bundle_model. If we reuse it, we basically need to have an if statement and rewrite all the parts. I feel it's more clear to handle this in ClientUploadManger.

codalab/client/json_api_client.py
```diff
@@ -713,3 +713,26 @@ def get_bundle_locations(self, bundle_uuid):
...
+            query_params=self._pack_params(params),
+        )
+        return response['data']
+
+    @wrap_exception("Unable to finalize the status of blob storage bundles")
+    def update_bundle_locations_blob(self, bundle_uuid, params):
+        response = self._make_request(
+            method='POST',
+            path='/bundles/{}/locations/blob'.format(bundle_uuid),
+            # data=self._pack_document([{'bundle_uuid': bundle_uuid, 'bundle_store_uuid': bundle_store_uuid}], 'bundle_locations'),
```
epicfaace: rm comment

codalab/common.py
```diff
@@ -249,12 +254,23 @@ def _get_sas_url(self, path, **kwargs):
             f"{StorageURLScheme.AZURE_BLOB_STORAGE.value}{AZURE_BLOB_ACCOUNT_NAME}/{AZURE_BLOB_CONTAINER_NAME}/",
             "",
         )  # for example, "0x9955c356ed2f42e3970bdf647f3358c8/contents.gz"
+
+        if permission == 'w':
+            # Use `create` instead of `write`. The `write` permission might modify other bundles.
```
epicfaace: What does this mean? Is this a todo?

codalab/common.py
```diff
@@ -272,25 +288,30 @@ def _get_signed_url(self, path, **kwargs):
             "/", 1
         )
         bucket = client.get_bucket(bucket_name)
-        blob = bucket.get_blob(blob_name)
+        blob = bucket.blob(blob_name)
```
epicfaace, line 291: What's the difference between `blob` and `get_blob`?

codalab/common.py
```diff
@@ -272,25 +288,30 @@ def _get_signed_url(self, path, **kwargs):
...
         signed_url = blob.generate_signed_url(
             version="v4",
             expiration=datetime.timedelta(hours=1),
-            method="GET",
-            response_disposition=kwargs["content_disposition"],
-            response_type=kwargs["content_type"],
+            method=kwargs.get("method", "GET"),  # HTTP method. eg, GET, PUT
+            # TODO: test what if kwargs does not have this keyword.
```
epicfaace: What does this mean?

codalab/common.py
```diff
@@ -272,25 +288,30 @@ def _get_signed_url(self, path, **kwargs):
...
+            content_type=kwargs.get("request_content_type", None),
+            response_disposition=kwargs.get("content_disposition", None),
+            response_type=kwargs.get("content_type", None),
         )
         return signed_url
 
-    def bundle_path_sas_url(self, **kwargs):
-        return self._get_sas_url(self.bundle_path, **kwargs)
+    def bundle_path_sas_url(self, permission='r', **kwargs):
```
epicfaace: We can remove this and just do 

```
def bundle_path_sas_url(self, **kwargs):
    return self._get_sas_url(self.bundle_path, **kwargs)
```

Because permission is a kwarg

codalab/lib/beam/filesystems.py
```diff
@@ -39,3 +39,17 @@
...
+def get_base_conn_str():
+    """
+    Get current connection string from environmental variable.
+    Remove the AccountKey. Used for bypass server upload to blob storage.
+    Return the Azure connection string without Account key.
+    """
+    conn_str = os.environ.get('AZURE_STORAGE_CONNECTION_STRING')
+    fields = conn_str.split(";")
+    for field in fields:
+        if "AccountKey" in field:
```
epicfaace: Always better to whitelist allowed fields than blacklist disallowed fields, for security purposes. Instead of removing "AccountKey", we should instead only include: "DefaultEndpointsProtocol","AccountName","BlobEndpoint".

codalab/lib/download_manager.py
```diff
@@ -349,6 +349,14 @@ def _get_target_path(self, target):
         except download_util.PathException as e:
             raise UsageError(str(e))
 
+    def get_target_sas_url(self, target, **kwargs):
```
epicfaace: This function isn't used.

codalab/lib/upload_manager.py
```diff
@@ -4,26 +4,41 @@
...
 
 class Uploader:
     """Uploader base class. Subclasses should extend this class and implement the
-    non-implemented methods that perform the uploads to a bundle store."""
+    non-implemented methods that perform the uploads to a bundle store.
+    Used when: 1. client -> blob storage
+               2. rest-server -> blob storage
+
+    params:
+    bundle_model: Used on rest-server.
```
epicfaace: move the params documentation to a docstring under __init__

codalab/lib/upload_manager.py
```diff
@@ -39,7 +54,14 @@ def write_git_repo(self, source: str, bundle_path: str):
...
     def write_fileobj(
-        self, source_ext: str, source_fileobj: IO[bytes], bundle_path: str, unpack_archive: bool
+        self,
+        source_ext: str,
+        source_fileobj: IO[bytes],
+        bundle_path: str,
+        unpack_archive: bool,
+        bundle_conn_str=None,
+        index_conn_str=None,
+        progress_callback=None,
```
epicfaace, line 65: Add progress_callback to the docstring for this method

codalab/lib/upload_manager.py
```diff
@@ -52,7 +74,8 @@ def write_fileobj(
 
     def upload_to_bundle_store(self, bundle: Bundle, source: Source, git: bool, unpack: bool):
         """Uploads the given source to the bundle store.
-        Given arguments are the same as UploadManager.upload_to_bundle_store()"""
+        Given arguments are the same as UploadManager.upload_to_bundle_store().
+        Used when upload from rest server."""
```
epicfaace:
```suggestion
        Used when uploading from rest server."""
```

codalab/lib/upload_manager.py
```diff
@@ -148,7 +171,14 @@ def write_git_repo(self, source: str, bundle_path: str):
...
     def write_fileobj(
-        self, source_ext: str, source_fileobj: IO[bytes], bundle_path: str, unpack_archive: bool
+        self,
+        source_ext: str,
+        source_fileobj: IO[bytes],
+        bundle_path: str,
+        unpack_archive: bool,
+        bundle_conn_str=None,
+        index_conn_str=None,
+        progress_callback=None,
```
epicfaace, line 185: Is it OK that we're not using progress_callback in this method?

codalab/lib/upload_manager.py
```diff
@@ -173,16 +203,38 @@ def write_git_repo(self, source: str, bundle_path: str):
...
             output_fileobj = zip_util.unpack_to_archive(source_ext, source_fileobj)
         else:
             output_fileobj = GzipStream(source_fileobj)
-        # Write archive file.
+
+        # write archive file.
+        if bundle_conn_str is not None:
+            conn_str = os.environ.get('AZURE_STORAGE_CONNECTION_STRING', '')
+            os.environ['AZURE_STORAGE_CONNECTION_STRING'] = bundle_conn_str
+        bytes_uploaded = 0
```
epicfaace: I would wrap this whole block in a `try:`. Then, at the very end, wrap the `os.environ['AZURE_STORAGE_CONNECTION_STRING'] = conn_str` code in a `finally:` block. This way, even if anything in this code throws an exception, we'll still eventually reset the connection string to the correct one.

codalab/lib/upload_manager.py
```diff
@@ -258,3 +326,168 @@ def cleanup_existing_contents(self, bundle):
...
         self._bundle_model.update_user_disk_used(bundle.owner_id)
+
+    def get_bundle_sas_token(self, path, **kwargs):
+        """
+        Get SAS token with write permission. Used for bypass server uploading.
+        """
+        return (
+            parse_linked_bundle_url(path)
+            .bundle_path_sas_url(permission='rw', **kwargs)
+            .split('?')[-1]
```
epicfaace: What is `.split('?')[-1]` for? Can you add a comment?

codalab/lib/upload_manager.py
```diff
@@ -258,3 +326,168 @@ def cleanup_existing_contents(self, bundle):
...
         self._bundle_model.update_bundle(bundle, bundle_update)
         self._bundle_model.update_user_disk_used(bundle.owner_id)
+
+    def get_bundle_sas_token(self, path, **kwargs):
+        """
+        Get SAS token with write permission. Used for bypass server uploading.
+        """
+        return (
+            parse_linked_bundle_url(path)
+            .bundle_path_sas_url(permission='rw', **kwargs)
```
epicfaace: Should this be

```suggestion
            .bundle_path_sas_url(permission='w', **kwargs)
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +326,168 @@ def cleanup_existing_contents(self, bundle):
...
+        """
+        Get SAS token of the index file with read and write permission. Used for uploading.
+        """
+        return (
+            parse_linked_bundle_url(path)
+            .index_path_sas_url(permission='rw', **kwargs)
+            .split('?')[-1]
+        )
+
+    def get_bundle_signed_url(self, path, **kwargs):
```
epicfaace, line 360: What's the difference between a signed url and sas url?

codalab/lib/upload_manager.py
epicfaace: Sounds good

codalab/lib/upload_manager.py
```diff
@@ -258,3 +326,168 @@ def cleanup_existing_contents(self, bundle):
...
+            # Mimic the rest server behavior
+            # decided the bundle type (file/directory) and decide whether need to unpack
+            bundle_conn_str = data.get('bundle_conn_str')
+            index_conn_str = data.get('index_conn_str')
+            bundle_url = data.get('bundle_url')
+            bundle_read_str = data.get('bundle_read_url', bundle_url)
+            try:
+                progress = FileTransferProgress('Sent ', f=self.stderr)
+                with closing(packed_source['fileobj']), progress:
+                    self.upload_Azure_blob_storage(
```
epicfaace, line 435: This function is small enough -- it's just one line -- that can we just inline the implementation of `upload_Azure_blob_storage` directly here instead?

codalab/rest/bundles.py
```diff
@@ -492,7 +590,14 @@ def _fetch_bundle_stores():
     - `storage_format`: the format in which storage is being stored (UNCOMPRESSED, COMPRESSED_V1, etc)
     - `url`: a self-referential URL that points to the bundle store.
     """
-    bundle_stores = local.model.get_bundle_stores(request.user.user_id)
+    bundle_store_name = query_get_type(str, 'name', None)
+    if bundle_store_name is not None:
+        # this function only fetch one record
```
epicfaace, line 611: What is this change for? Should we update the docstring as well?

codalab/worker/rest_client.py
```diff
@@ -69,7 +69,6 @@ def _make_request(
         access_token = self._get_access_token()
         if authorized and access_token:
             headers['Authorization'] = 'Bearer ' + self._get_access_token()
-
```
epicfaace: re-add this space

epicfaace (commented):
Nice work! more comments

codalab/common.py
wwwjn: Oh no, it's an out-dated comment, and I have remove it

codalab/common.py
wwwjn: get_blob can only be used when the blob store already exists in the GCS (when downloading). The blob is more general and can refer to a blob that does not exist on GCS (eg, we want to generate a signed URL for upload, and the blob_name does not exist on GCS yet.

codalab/common.py
wwwjn: It's an outdated comment. I will pay more attention to clean it.

codalab/rest/bundles.py
wwwjn: This is a improvement of this function: support query a bundle store by bundle store name. We don't need to update the docs, cause the description of the endpoint has already have the parameter `name`, but we did not support using the `name` field before.

codalab/lib/upload_manager.py
wwwjn: This fuction is small for Azure, but the upload_GCS_blob_storage function is long (since it can not use Apache Beam to upload). So I think it's better to have another function to handle the actual upload process.

codalab/lib/upload_manager.py
wwwjn: > Should this be

This token is used for upload first, then used for download this file to generate index file. So the permission is `read and write`

codalab/lib/upload_manager.py
wwwjn: This is to get the SAS token from the SAS URL. I have add a comment for this, Thanks Ashwin!

codalab/lib/upload_manager.py
wwwjn: The signed url is for GCS, and the SAS URL is for Azure. And they call different function to generate SAS URL and signed url.

codalab/lib/beam/filesystems.py
wwwjn: Thanks Ashwin, I have modified this!

codalab/lib/upload_manager.py
wwwjn: Yes, the default progress_callback function is set to None. But we need `progress_callback` in the function signature, since the ClientUploadManager need to know the upload progress by `_write_fileobj`

codalab/common.py
```diff
@@ -272,33 +288,33 @@ def _get_signed_url(self, path, **kwargs):
...
             return self._get_sas_url(self.bundle_path, **kwargs)
         elif self.storage_type == StorageType.GCS_STORAGE.value:
             return self._get_signed_url(self.bundle_path, **kwargs)
         else:
             raise UsageError(f"Does not support current storage type: {self.storage_type}")
 
+    def index_path_sas_url(self, **kwargs):
+        if self.storage_type == StorageType.AZURE_BLOB_STORAGE.value:
+            return self._get_sas_url(self.index_path, **kwargs)
+        elif self.storage_type == StorageType.GCS_STORAGE.value:
```
epicfaace, line 313: This PR doesn't include GCS -- it's only for azure blob... can you remove this from this PR?

codalab/rest/bundles.py
epicfaace: Wait, but the description of the endpoint doesn't already have the parameter `name`. So should we update the docs?

codalab/rest/bundles.py
```diff
@@ -479,6 +541,42 @@ def _fetch_bundle_location(bundle_uuid: str, bundle_store_uuid: str):
     return BundleLocationListSchema(many=True).dump(bundle_location).data
 
 
+@post(
+    '/bundles/<bundle_uuid:re:%s>/locations/blob' % spec_util.UUID_STR,
+    apply=AuthenticatedProtectedPlugin(),
+)
+def _update_bundle_location(bundle_uuid: str):
+    """
+    Finalize after uploading a file to the blob storage.
```
```suggestion
    Updates a bundle location. Used to finalize a bundle's upload status after it is uploaded by the client directly to the bundle store, such as uploading to blob storage and bypassing the server.
```

codalab/rest/bundles.py
```diff
@@ -479,6 +541,42 @@ def _fetch_bundle_location(bundle_uuid: str, bundle_store_uuid: str):
...
+
+    Query parameters:
```
epicfaace, line 565: Because this is a POST request, I think these should be part of the JSON body, rather than query string parameters.

codalab/rest/bundles.py
```diff
@@ -451,16 +458,71 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+            bundle_sas_token = local.upload_manager.get_bundle_sas_token(bundle_url)
+            index_sas_token = local.upload_manager.get_index_sas_token(bundle_url)
+            base_conn_str = get_base_conn_str()
+            if LOCAL_USING_AZURITE and get_request_source() == RequestSource.CLI:
+                # For test locally. Mannually typing `cl upload`
+                base_conn_str = base_conn_str.replace("azurite", "localhost", 1)
+            bundle_conn_str = f"{base_conn_str};SharedAccessSignature={bundle_sas_token};"
+            index_conn_str = f"{base_conn_str};SharedAccessSignature={index_sas_token};"
+
+        elif bundle_url.startswith(StorageURLScheme.GCS_STORAGE.value):
```
epicfaace: Remove GCS from this PR

codalab/rest/bundles.py
```diff
@@ -451,16 +458,71 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+    if need_sas:
+        if bundle_url is None:
+            bundle_conn_str, index_conn_str = None, None
+        elif bundle_url.startswith(StorageURLScheme.AZURE_BLOB_STORAGE.value):
+            # generate the SAS token and Azure connection string, and send it back to the client
+            bundle_sas_token = local.upload_manager.get_bundle_sas_token(bundle_url)
+            index_sas_token = local.upload_manager.get_index_sas_token(bundle_url)
+            base_conn_str = get_base_conn_str()
+            if LOCAL_USING_AZURITE and get_request_source() == RequestSource.CLI:
+                # For test locally. Mannually typing `cl upload`
```
```suggestion
                # For test locally. Manually typing `cl upload`
```

I don't understand this, we should discuss

codalab/rest/bundles.py
```diff
@@ -451,16 +458,71 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
     )
-    return BundleLocationSchema(many=True).dump([new_location]).data
+    if new_location.get('bundle_store_uuid', None) is None:
+        if os.environ.get('CODALAB_ALWAYS_USE_AZURE_BLOB_BETA') == '1':
+            # The rest-server use Azure as default storage. Use default azure path
+            local.model.update_bundle(
+                bundle, {'storage_type': StorageType.AZURE_BLOB_STORAGE.value, 'is_dir': is_dir},
+            )
+            bundle_url = local.bundle_store.get_bundle_location(bundle_uuid)
+        else:  # The rest-server does not support store on Azure.
```
```suggestion
        else:  # The rest-server does not use Azure as default storage.
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,71 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+    bundle = local.model.get_bundle(bundle_uuid)
     new_location = BundleLocationSchema(many=True).load(request.json).data[0]
-    new_location["uuid"] = local.model.add_bundle_location(
-        new_location['bundle_uuid'], new_location['bundle_store_uuid']
+    logging.info(
+        f"Try to bypass server upload, need_sas: {need_sas}, is_dir: {is_dir}, new_location: {new_location}"
     )
-    return BundleLocationSchema(many=True).dump([new_location]).data
+    if new_location.get('bundle_store_uuid', None) is None:
+        if os.environ.get('CODALAB_ALWAYS_USE_AZURE_BLOB_BETA') == '1':
```
epicfaace: It's not very clear what this function is doing -- I think we should add more comments to clarify

codalab/client/json_api_client.py
```diff
@@ -713,3 +713,25 @@ def get_bundle_locations(self, bundle_uuid):
...
+            data=self._pack_document(
+                [{'bundle_uuid': bundle_uuid, 'bundle_store_uuid': bundle_store_uuid}],
+                'bundle_locations',
+            ),
+            query_params=self._pack_params(params),
+        )
+        return response['data']
+
+    @wrap_exception("Unable to finalize the status of blob storage bundles")
+    def update_bundle_locations_blob(self, bundle_uuid, params):
```
epicfaace: Should this be `update_bundle_location_blob`?

codalab/common.py
wwwjn: This two lines are not specific for bypassing server upload, It is also used to bypass server download. I thinks we can live these two lines here.

epicfaace: Thanks

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,167 @@ def cleanup_existing_contents(self, bundle):
         bundle_update = {'data_hash': None, 'metadata': {'data_size': 0}}
         self._bundle_model.update_bundle(bundle, bundle_update)
         self._bundle_model.update_user_disk_used(bundle.owner_id)
+
+    def get_bundle_sas_token(self, path, **kwargs):
```
epicfaace, line 340: Let's replace _sas_ with _azure_sas_, and _signed_ with _gcs_signed_

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,167 @@ def cleanup_existing_contents(self, bundle):
...
+        """
+        Bypass server upload. Upload from client directly to different blob storage (Azure, GCS, Disk storage).
+        """
+
+        # By pass server upload:
+        # 1. The server support use Azure as default storage
+        # 2. If the user specify `--store` and blob storage is on Azure
+        upload_to_disk = False
+        bundle_store_uuid = None
+        if destination_bundle_store is not None and destination_bundle_store != '':
```
epicfaace, line 397:
```suggestion
        # 1) Read from --store if user has specified it
        if destination_bundle_store is not None and destination_bundle_store != '':
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,167 @@ def cleanup_existing_contents(self, bundle):
...
+                params={
+                    'name': destination_bundle_store,
+                    'include': ['uuid', 'storage_type', 'url'],
+                },
+            )
+            bundle_store_uuid = storage_info['uuid']
+            if storage_info['storage_type'] in (StorageType.DISK_STORAGE.value,):
+                upload_to_disk = True  # The user specify --store to upload to disk storage
+
+        source_ext = zip_util.get_archive_ext(packed_source['filename'])
```
epicfaace, line 410:
```suggestion
        # 2) Pack the files to be uploaded
        source_ext = zip_util.get_archive_ext(packed_source['filename'])
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,167 @@ def cleanup_existing_contents(self, bundle):
...
+        if packed_source['should_unpack'] and zip_util.path_is_archive(packed_source['filename']):
+            unpack_before_upload = True
+            is_dir = source_ext in zip_util.ARCHIVE_EXTS_DIR
+        else:
+            unpack_before_upload = False
+            is_dir = False
+
+        # try to bypass server upload
```
epicfaace:
```suggestion
        # 3) Create a bundle location for the bundle
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,167 @@ def cleanup_existing_contents(self, bundle):
...
+        params = {'need_sas': not upload_to_disk, 'is_dir': is_dir}
+        data = self._client.add_bundle_location(bundle['id'], bundle_store_uuid, params)[0].get(
+            'attributes'
+        )
+
+        if data.get('bundle_conn_str', None) is not None:
```
epicfaace, line 425:
```suggestion
        4) If bundle location has bundle_conn_str, we should bypass the server when uploading.
        if data.get('bundle_conn_str', None) is not None:
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,167 @@ def cleanup_existing_contents(self, bundle):
...
+            except Exception as err:
+                self._client.update_bundle_location_blob(
+                    bundle['id'],
+                    params={'success': False, 'error_msg': f'Bypass server upload error. {err}',},
+                )
+                raise err
+            else:
+                self._client.update_bundle_location_blob(bundle['id'], params={'success': True})
+        else:
+            # upload to rest-server
```
epicfaace:
```suggestion
            # 5) Otherwise, upload the bundle directly through the server.
```

codalab/client/json_api_client.py
```diff
@@ -713,3 +713,25 @@ def get_bundle_locations(self, bundle_uuid):
...
+            data=self._pack_document(
+                [{'bundle_uuid': bundle_uuid, 'bundle_store_uuid': bundle_store_uuid}],
+                'bundle_locations',
+            ),
+            query_params=self._pack_params(params),
+        )
+        return response['data']
+
+    @wrap_exception("Unable to finalize the status of blob storage bundles")
+    def update_bundle_location_blob(self, bundle_uuid, params):
```
epicfaace:
```suggestion
    def update_bundle_location(self, bundle_uuid, location_id, params):
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,60 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+    need_sas = query_get_bool('need_sas', default=False)
+    is_dir = query_get_bool('is_dir', default=False)
+
+    bundle = local.model.get_bundle(bundle_uuid)
     new_location = BundleLocationSchema(many=True).load(request.json).data[0]
-    new_location["uuid"] = local.model.add_bundle_location(
-        new_location['bundle_uuid'], new_location['bundle_store_uuid']
+    logging.info(
+        f"Try to bypass server upload, need_sas: {need_sas}, is_dir: {is_dir}, new_location: {new_location}"
     )
```
epicfaace, line 475: maybe flatten out if statements

```
if new_location.get('bundle_store_uuid', None) is None and os.environ.get('CODALAB_ALWAYS_USE_AZURE_BLOB_BETA') == '1':
   # Scenario 1: rest-server does not use azure
   ...
```

epicfaace: Also let's describe the scenarios in this doc: https://docs.google.com/document/d/1qFT6OWSUW5-2ZfnkK_q6P-Q-9sVe_fyBLX3UfEIWRiY/edit?pli=1#

codalab/client/json_api_client.py
wwwjn: The problem is: not every bundle has a corresponding location_id. For example, if a user does not specify destination storage, the rest server supports using Azure as default storage.  In this scenario, the rest server will pick the default azure storage to store it, and will not add a line in the `bundle_location` table.

Now, the behavior is same for bypass upload and go-through server upload. If we want to add a location in `bundle_location` table for every bundle (and stored in any bundle storage), should we make another PR to change bypass server upload and go-through server upload at the same time?

wwwjn: And for now, this function `update_bundle_location` does not really update the bundle location. It only update the bundle state  (success or failure), to finalize the bypass server upload. 
So I think a better way is to change the function name to `update_bundle_state`, and change endpoint URL to `'/bundles/<bundle_uuid:re:%s>/state'`.

epicfaace: > The problem is: not every bundle has a corresponding location_id. For example, if a user does not specify destination storage, the rest server supports using Azure as default storage. In this scenario, the rest server will pick the default azure storage to store it, and will not add a line in the `bundle_location` table.

@wwwjn If the rest server sets the CODALAB_ALWAYS_USE_AZURE_BLOB_BETA variable, then we should actually not support bypass server for uploads. This is because, just like cl upload -a, CODALAB_ALWAYS_USE_AZURE_BLOB_BETA is also a temporary / soon-to-be-deprecated solution because it uses storage_type instead of creating an entry in the bundle location table.

This means that we should only support bypass server for uploads where we do create a bundle store, so this means that each bundle here should have a corresponding location_id. This means that there are only two scenarios in which this would be the case:
1. User specifies --store [store name]
2. Server is configured with CODALAB_DEFAULT_BUNDLE_STORE_NAME=[store name]

Does that make sense? I've also added suggestions to the doc (https://docs.google.com/document/d/1qFT6OWSUW5-2ZfnkK_q6P-Q-9sVe_fyBLX3UfEIWRiY/edit?pli=1#) to reflect this change:

image

wwwjn: I think this is a good solution for this. But I'm a little bit confused about do we need to change this in this PR? Or we simply do not support CODALAB_ALWAYS_USE_BLOB_STORAGE?

If we want to make change for default storage, maybe we need another PR to change go through server at same time

epicfaace: I think we should not support bypassing server upload for CODALAB_ALWAYS_USE_AZURE_BLOB_BETA. I will indicate in another comment where I think you need to change this PR.

codalab/rest/bundles.py
```diff
@@ -451,16 +458,67 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
-    return BundleLocationSchema(many=True).dump([new_location]).data
+    # Scenario 1: User does not sepcify destination store, but rest-server support using Azure as default storage.
+    # Should bypass server and upload to default Azure store.
+    if (
+        new_location.get('bundle_store_uuid', None) is None
+        and os.environ.get('CODALAB_ALWAYS_USE_AZURE_BLOB_BETA') == '1'
+    ):
+        # The rest-server use Azure as default storage. Use default azure path
+        local.model.update_bundle(
```
epicfaace: @wwwjn I think this is where you need to change the PR. We should no longer check for CODALAB_ALWAYS_USE_AZURE_BLOB_BETA. Rather, we should check for CODALAB_DEFAULT_BUNDLE_STORE_NAME, and, in that case, we should create a bundle store.

Here's a simple (but just one) way to do it:
```
bundle_store_uuid = new_location.get('bundle_store_uuid', None)
if bundle_store_uuid is None and os.environ.get('CODALAB_DEFAULT_BUNDLE_STORE_NAME'):
    bundle_store_uuid = # get bundle store uuid from os.environ.get('CODALAB_DEFAULT_BUNDLE_STORE_NAME')
...
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,78 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
-        new_location['bundle_uuid'], new_location['bundle_store_uuid']
+    logging.info(
+        f"Try to bypass server upload, need_sas: {need_sas}, is_dir: {is_dir}, new_location: {new_location}"
     )
-    return BundleLocationSchema(many=True).dump([new_location]).data
+    # Scenario 1: User does not sepcify destination store, but rest-server set default storage name.
+    # Should bypass server and upload to default Azure store.
+    if (
+        new_location.get('bundle_store_uuid', None) is None
+        and os.environ.get('CODALAB_DEFAULT_BUNDLE_STORE_NAME') is not None
```
epicfaace: We should also add one more test in https://github.com/codalab/codalab-worksheets/blob/master/.github/workflows/test.yml where CODALAB_DEFAULT_BUNDLE_STORE_NAME is set to blob storage, so we can ensure that the server is bypassed during uploads.

epicfaace: We should also add one more test in https://github.com/codalab/codalab-worksheets/blob/master/.github/workflows/test.yml where CODALAB_DEFAULT_BUNDLE_STORE_NAME is set to blob storage, so we can ensure that the server is bypassed during uploads.

.github/workflows/test.yml
```diff
@@ -374,6 +374,61 @@ jobs:
...
+      - name: Setup tests
+        run: |
+          sudo service mysql stop
+          python3 codalab_service.py build services --version ${VERSION} --pull
+        env:
+          VERSION: ${{ github.head_ref || 'master' }}
+      - name: Run tests
+        run: |
+          python3 codalab_service.py start --services default azurite --version ${VERSION}
+          sleep 20
```
epicfaace: I don't think we need `sleep 20` since the previous command already waits for codalab server to be ready. Try it out (removing it)?
```suggestion
```

codalab/common.py
```diff
@@ -272,30 +288,30 @@ def _get_signed_url(self, path, **kwargs):
...
-            response_disposition=kwargs["content_disposition"],
-            response_type=kwargs["content_type"],
+            method=kwargs.get("method", "GET"),  # HTTP method. eg, GET, PUT
+            content_type=kwargs.get("request_content_type", None),
+            response_disposition=kwargs.get("content_disposition", None),
+            response_type=kwargs.get("content_type", None),
         )
         return signed_url
 
     def bundle_path_sas_url(self, **kwargs):
```
epicfaace: Maybe we should rename `bundle_path_sas_url` -> `bundle_path_bypass_url` and `index_path_sas_url` -> `bundle_path_bypass_url`. (this is because SAS is specific to Azure, so using a more general term). If you have a better idea than "bypass_url", feel free to rename it to that as well!

Also let's add docstring comments to each of these functions.

codalab/lib/beam/filesystems.py
```diff
@@ -39,3 +39,19 @@
 
 # HTTP endpoint used to directly access Blob Storage. Used to generate SAS URLs.
 AZURE_BLOB_HTTP_ENDPOINT = f"http://localhost:10000/{AZURE_BLOB_ACCOUNT_NAME}" if LOCAL_USING_AZURITE else f"https://{AZURE_BLOB_ACCOUNT_NAME}.blob.core.windows.net"
+
+
+def get_base_conn_str():
```
epicfaace: Rename this to `get_azure_bypass_conn_str`

codalab/lib/beam/filesystems.py
```diff
@@ -39,3 +39,19 @@
...
+    """
+    Get current connection string from environmental variable.
```
epicfaace:
```suggestion
    Get current Azure connection string from environment variables.
```

codalab/lib/beam/filesystems.py
```diff
@@ -39,3 +39,19 @@
...
+    Remove the AccountKey. Used for bypass server upload to blob storage.
```
epicfaace:
```suggestion
    Used for bypass server upload to blob storage.
```

codalab/lib/beam/filesystems.py
```diff
@@ -39,3 +39,19 @@
...
+    Return the Azure connection string without Account key.
```
epicfaace:
```suggestion
    Returns the Azure connection string without Account key.
```

codalab/lib/beam/filesystems.py
epicfaace: Could we add a small unit test for `get_base_conn_str`? just to make sure it doesn't regress in the future...

codalab/lib/download_manager.py
```diff
@@ -349,9 +349,12 @@ def _get_target_path(self, target):
         except download_util.PathException as e:
             raise UsageError(str(e))
 
-    def get_target_download_url(self, target, **kwargs):
-        return parse_linked_bundle_url(self._get_target_path(target)).bundle_path_download_url(
-            **kwargs
+    def get_target_sas_url(self, target, **kwargs):
```
epicfaace: Similarly, we should probably rename this to get_target_bypass_url? (since SAS is specific to Azure).

codalab/lib/upload_manager.py
```diff
@@ -4,26 +4,42 @@
...
+from codalab.lib.print_util import FileTransferProgress
 
 Source = Union[str, Tuple[str, IO[bytes]]]
 
 
 class Uploader:
     """Uploader base class. Subclasses should extend this class and implement the
-    non-implemented methods that perform the uploads to a bundle store."""
+    non-implemented methods that perform the uploads to a bundle store.
+    Used when: 1. client -> blob storage
```
epicfaace: mention that in option 1. is_client should be set to True, else False

codalab/lib/upload_manager.py
```diff
@@ -173,31 +207,72 @@ def write_git_repo(self, source: str, bundle_path: str):
...
-        ) as ttf, tempfile.NamedTemporaryFile(suffix=".sqlite") as tmp_index_file:
-            SQLiteIndexedTar(
-                fileObject=ttf,
-                tarFileName="contents",  # If saving a single file as a .gz archive, this file can be accessed by the "/contents" entry in the index.
-                writeIndex=True,
-                clearIndexCache=True,
-                indexFilePath=tmp_index_file.name,
-            )
+
+        # write archive file.
```
epicfaace:
```suggestion
        # Write archive file.
```

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,170 @@ def cleanup_existing_contents(self, bundle):
...
+        use_azure_blob_beta: bool,
+        destination_bundle_store=None,
+    ):
+        """
+        Bypass server upload. Upload from client directly to different blob storage (Azure, GCS, Disk storage).
+        """
+
+        # By pass server upload:
+        # 1. The server support use Azure as default storage
+        # 2. If the user specify `--store` and blob storage is on Azure
```
epicfaace: I would move this comment (these 3 lines) up to the docstring

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+
+    bundle = local.model.get_bundle(bundle_uuid)
     new_location = BundleLocationSchema(many=True).load(request.json).data[0]
-    new_location["uuid"] = local.model.add_bundle_location(
-        new_location['bundle_uuid'], new_location['bundle_store_uuid']
+    logging.info(
+        f"Try to bypass server upload, need_sas: {need_sas}, is_dir: {is_dir}, new_location: {new_location}"
     )
-    return BundleLocationSchema(many=True).dump([new_location]).data
+    # Scenario 1: User does not sepcify destination store, but rest-server set default storage name.
```
```suggestion
    # Scenario 1: User does not specify destination store, but rest-server has set a default storage name.
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+                bundle, {'storage_type': default_bundle_store['storage_type'], 'is_dir': is_dir},
+            )
+            bundle_url = local.bundle_store.get_bundle_location(
+                bundle_uuid, default_bundle_store['uuid']
+            )
+        else:  # default storage is disk, do not support bypass server
+            bundle_url = None
+
+    # Scenario 2: User does not specify destination store, and rest-server does not use Azure as default storage.
+    # Should go throught rest server and upload to disk storage.
```
epicfaace, line 501:
```suggestion
    # Should go through rest server and upload to disk storage.
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+    elif new_location.get('bundle_store_uuid', None) is None:
+        bundle_url = None
+
+    # Scenario 3: User specify destination store. Should upload to the specified storage.
```
epicfaace: @wwwjn one thing -- are all of these scenarios (both in bundles.py and the ones in upload_manager.py) covered in our tests? If not, we might want to add more tests that cover them.

codalab/lib/upload_manager.py
```diff
@@ -258,3 +336,170 @@ def cleanup_existing_contents(self, bundle):
...
+        upload_to_disk = False
```
epicfaace: Instead of making a variable called `upload_to_disk` and then setting `need_sas = not upload_to_disk` later, it would be simpler if we just made a variable called `need_sas` and then set it to True / False.

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
-        new_location['bundle_uuid'], new_location['bundle_store_uuid']
+    logging.info(
+        f"Try to bypass server upload, need_sas: {need_sas}, is_dir: {is_dir}, new_location: {new_location}"
     )
-    return BundleLocationSchema(many=True).dump([new_location]).data
+    # Scenario 1: User does not sepcify destination store, but rest-server set default storage name.
+    # Should bypass server and upload to default Azure store.
+    if (
+        new_location.get('bundle_store_uuid', None) is None
+        and os.environ.get('CODALAB_DEFAULT_BUNDLE_STORE_NAME') is not None
```
epicfaace: We can replace this line and the next line with just:

```suggestion
        and os.environ.get('CODALAB_DEFAULT_BUNDLE_STORE_NAME')
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+            local.model.update_bundle(
+                bundle, {'storage_type': default_bundle_store['storage_type'], 'is_dir': is_dir},
+            )
+            bundle_url = local.bundle_store.get_bundle_location(
+                bundle_uuid, default_bundle_store['uuid']
+            )
+        else:  # default storage is disk, do not support bypass server
+            bundle_url = None
+
+    # Scenario 2: User does not specify destination store, and rest-server does not use Azure as default storage.
```
```suggestion
    # Scenario 2: User does not specify destination store, and rest-server does not specify a default bundle store name.
```

codalab/rest/bundles.py
```suggestion
    # Scenario 3: User specifies destination store. Should upload to the specified storage.
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+    # Should go throught rest server and upload to disk storage.
+    elif new_location.get('bundle_store_uuid', None) is None:
+        bundle_url = None
+
+    # Scenario 3: User specify destination store. Should upload to the specified storage.
+    else:
+        local.model.add_bundle_location(
+            new_location['bundle_uuid'], new_location['bundle_store_uuid']
+        )
+        local.model.update_bundle(  # get_bundle_location() function uses this field
```
```suggestion
        local.model.update_bundle(
```

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+            bundle, {'is_dir': is_dir},
+        )
+        bundle_url = local.bundle_store.get_bundle_location(bundle_uuid)
+    data = BundleLocationSchema(many=True).dump([new_location]).data
+    logging.info(f"Bypass server upload, the URL is {bundle_url}")
+    if need_sas:
+        if bundle_url is None:
+            bundle_conn_str, index_conn_str = None, None
```
epicfaace, line 519: Add a comment explaining what this scenario means.

codalab/rest/bundles.py
```diff
@@ -451,16 +458,79 @@ def _fetch_bundle_locations(bundle_uuid: str):
...
+        elif bundle_url.startswith(StorageURLScheme.AZURE_BLOB_STORAGE.value):
+            # generate the SAS token and Azure connection string, and send it back to the client
+            bundle_sas_token = local.upload_manager.get_bundle_sas_token(bundle_url)
+            index_sas_token = local.upload_manager.get_index_sas_token(bundle_url)
+            base_conn_str = get_base_conn_str()
+            if LOCAL_USING_AZURITE and get_request_source() == RequestSource.CLI:
+                # For test locally. Manually typing `cl upload`
```
epicfaace: Can you explain this comment further?

codalab/rest/bundles.py
```diff
@@ -492,7 +599,14 @@ def _fetch_bundle_stores():
     - `storage_format`: the format in which storage is being stored (UNCOMPRESSED, COMPRESSED_V1, etc)
     - `url`: a self-referential URL that points to the bundle store.
     """
-    bundle_stores = local.model.get_bundle_stores(request.user.user_id)
+    bundle_store_name = query_get_type(str, 'name', None)
```
epicfaace, line 609: Add the `name` parameter to this function's docstring

epicfaace (approved):

codalab/common.py
wwwjn: I think `bypass_url` is a good name for our usage. Thanks Ashwin!

codalab/lib/beam/filesystems.py
wwwjn: Yes, sure! I put a test in `tests/unit/server/upload_download_test.py`, I think the base_conn_str() is a function on the server-side, and it is used for upload and download. I'm not very sure about whether this is the right place to put it..

codalab/rest/bundles.py
wwwjn: I modified `upload1` and `upload2` to make these test case cover all the scenarios.

epicfaace (commented):
Feel free to merge once you've addressed all the comments. Thanks for your work on this! 👍🏼

PR #4109, opened 2022-05-25
Update python from 3.6 to 3.7 in codalab docker containers

### LINKED ISSUE #4091: Upgrade from python 3.6
Python 3.6 is no longer supported by the Python core team and is now end-of-life; we should upgrade.

nelson-liu:
Python 3.6 is end-of-life.

I was banging my head trying to figure out why something I was working on in the rest server wasn't working, and it turns out the python version in the docker container was too old. I've upgraded it to 3.7 via conda and moved away from the defunct deadsnakes ppa

this also fixes a very noisy warning in the docker build process about `cryptography` not supporting 3.6.

### DIFF
--- a/docker_config/dockerfiles/Dockerfile.worker
+++ b/docker_config/dockerfiles/Dockerfile.worker
@@ -25,14 +25,9 @@ ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update && \
   apt-get install -y --no-install-recommends software-properties-common curl && \
-  add-apt-repository ppa:fkrull/deadsnakes && \
   apt-get update && \
   apt-get install -y --no-install-recommends \
     build-essential \
-    python3.6 \
-    python3.6-dev \
-    python3-pip \
-    python3-setuptools \
     libssl-dev \
     uuid-dev \
     libgpgme11-dev \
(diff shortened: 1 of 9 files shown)


epicfaace:
@nelson-liu great! Can you also update github actions to use python 3.7?

nelson-liu:
should i remove 3.6 from the test matrix as well? https://github.com/codalab/codalab-worksheets/pull/4109/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R39

nelson-liu:
oops, not sure if that link worked, here's a better one: https://github.com/nelson-liu/codalab-worksheets/blob/update_codalab_python/.github/workflows/test.yml#L39

nelson-liu:
I did the bare minimum here wrt resolving dependencies to make things build / pass, but maybe worth going through and updating other dependencies at some point? anyway, i think this is ready to be looked at @epicfaace

docker_config/dockerfiles/Dockerfile.worker
```diff
@@ -62,9 +62,9 @@ RUN mkdir ${WORKDIR}/scripts
 
 # Install dependencies
 COPY requirements.txt requirements.txt
-RUN python3.6 -m pip install --user --upgrade pip==20.3.4; \
-    python3.6 -m pip install setuptools --upgrade; \
-    python3.6 -m pip install --no-cache-dir -r requirements.txt;
+RUN python3 -m pip install --user --upgrade pip==20.3.4; \
+    python3 -m pip install setuptools --upgrade; \
+    python3 -m pip install --no-cache-dir -r requirements.txt;
```
epicfaace, line 67: Just making sure that python3 ends up pointing to 3.7, rather than the default system installation of python (3.5?), right?

epicfaace (approved):
lgtm

docker_config/dockerfiles/Dockerfile.worker
nelson-liu: yup, one of the lines above prepends the conda `bin` directory to $PATH, which should set the default python.

PR #4112, opened 2022-05-27
Change python3.6 to python3 in monitor compose entry

nelson-liu:
Monitor breaks because 3.6 isn't in the container anymore


epicfaace (approved):
Thanks

PR #4090, opened 2022-04-30
Support kubernetes runtime for launching jobs

epicfaace:
This PR builds off of https://github.com/codalab/codalab-worksheets/pull/4088. It includes a working proof-of-concept of CPU jobs with dependencies being launched through the kubernetes runtime.

image

Todos for future PRs:
- get GPUs to work
  - just that one line in the PR, install drivers needs to happen as a prereq (https://codalab-worksheets.readthedocs.io/en/latest/Worker-Managers/#installing-nvidia-driver-and-dependencies)
- make sure workers are now advertising their number of CPUs / disk space properly
  - max 1 pod per node, pod should advertise disk space / memory available for the node.
- make sure time calculations for bundles work

For #3975

Issue #3975, opened 2022-01-26
Launch jobs through a kubernetes runtime instead of a docker runtime (because GKE has deprecated the docker runtime)


epicfaace:
https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/

https://kubernetes.io/blog/2022/01/07/kubernetes-is-moving-on-from-dockershim/

pranavjain:
This is planned to be depreciated in Kubernetes 1.24 (ETA April 2022).

pranavjain:
@epicfaace will pick this task after the necessary tasks for the course are done.

epicfaace:
Two options:
1. Mount containerd socket (`/run/containerd/containerd.sock`) instead of docker socket (`/var/run/docker.sock`). (https://github.com/containerd/containerd/blob/main/docs/getting-started.md, https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/kubernetes_worker_manager.py#L133)
2. Rather than running a Docker container, a worker can just interface with the k8s API to launch a job.

2) is probably the right path forward.

pranavjain:
@epicfaace Are you working on this one actively? Do we have some rough ETA?

epicfaace:
The plan is for me to have something written up by Tuesday as to a possible
approach to solve the problem. Will work on this soon.

--
Ashwin Ramaswami

epicfaace:
Details: https://docs.google.com/document/d/1QsnUYAb0DzbH6uKA1L5oTbmxQYikGQ7F_FI4n3Tgp_g/edit

epicfaace:
https://cloud.google.com/kubernetes-engine/docs/deprecations/docker-containerd

epicfaace:
Good read / explanation: https://kubernetes.io/docs/tasks/administer-cluster/migrating-from-dockershim/check-if-dockershim-removal-affects-you/

pranavjain:
@teetone to help with the testing Nvidia containerd

pranavjain:
https://jamboard.google.com/d/1_wV8vM9A5nMo1i4r6A8ZmgBMUSGKe7hkXyHZscLV8w0/viewer?ts=62c74ff8

pranavjain:
ToDo: Code Cleanup by Ashwin

pranavjain:
#4214 is ready for review (ToDo: @wwwjn )
#4090 Still pending code clean up.

PR #4121, opened 2022-05-31
Allow passing empty --worker-tag to worker managers

epicfaace:
- Allow passing empty `--worker-tag`s to worker managers (previously this didn't work). This allows us to run k8s worker managers that are not tied to a particular tag.

Part of https://github.com/codalab/codalab-worksheets/pull/4090, for issue https://github.com/codalab/codalab-worksheets/issues/3975


codalab_service.py
```diff
@@ -432,7 +432,6 @@ def has_callable_default(self):
         CodalabArg(
             name='worker_manager_{}_tag'.format(worker_manager_type),
             help='Tag of worker for {} jobs'.format(worker_manager_type),
-            default='codalab-{}'.format(worker_manager_type),
```
epicfaace: undo this change -- we will just pass "" to the env variables

codalab/worker/main.py
```diff
@@ -323,7 +323,7 @@ def main():
         args.checkin_frequency_seconds,
         bundle_service,
         args.shared_file_system,
-        args.tag_exclusive,
+        args.tag_exclusive if args.tag else False,
```
epicfaace: undo this change and test

teetone (approved):

PR #4125, opened 2022-06-04
Read and authenticate with credentials from password file when starting Slurm worker manager

### LINKED ISSUE #4066: nlp-cl-worker-manager prompts for password
nlp-cl-worker-manager prompts for a password if it is restarted even when a password file exists.

teetone:
Resolves https://github.com/codalab/codalab-worksheets/issues/4066

### DIFF
--- a/codalab/worker_manager/slurm_batch_worker_manager.py
+++ b/codalab/worker_manager/slurm_batch_worker_manager.py
@@ -109,6 +109,10 @@ class SlurmBatchWorkerManager(WorkerManager):
         # A set of newly submitted job id to keep tracking worker status, as worker might not be created right away.
         self.submitted_jobs = self.load_worker_jobs()
 
+        # We're using a temporary CodaLabManager for the worker and worker managers by default,
+        # so we don't write out the state to state.json when users log in.
+        # If we always read from the password file, we can both leave state.json untouched,
+        # and users can always start the worker manager without logging in.
         if args.password_file:
             logger.info(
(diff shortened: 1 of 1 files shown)


codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -108,6 +109,14 @@ def __init__(self, args):
         # A set of newly submitted job id to keep tracking worker status, as worker might not be created right away.
         self.submitted_jobs = self.load_worker_jobs()
 
+        if args.password_file:
+            logger.info(
+                f"--password-file specified. Reading credentials from {args.password_file}..."
+            )
+            with open(args.password_file) as f:
+                os.environ["CODALAB_USERNAME"] = f.readline().strip()
+                os.environ["CODALAB_PASSWORD"] = f.readline().strip()
```
percyliang, line 122: Can you explain why we have to do this?

teetone: We're using a temporary `CodaLabManager` for the worker and worker managers by default, so we don't write out the state to `state.json` when users log in. That's why running `nlp-cl-worker-manager` multiple times always prompts the user to log in. If we always read from the password file, we can both leave state.json untouched, and users can always start the worker manager without logging in.

percyliang: Could you please add a comment that explains some of this?

percyliang (approved):

PR #4126, opened 2022-06-04
Set default for OAUTH2_PROVIDER_TOKEN_EXPIRES_IN to 7 days

### LINKED ISSUE #3948: Set oauth token expiration default to a month
extend oauth token expiration

teetone:
Resolves https://github.com/codalab/codalab-worksheets/issues/3948

Every token refresh by a user is a request to the server, so now it's a single refresh request every week instead of every day. Also, unless I missed something, the refresh token TTL is undocumented for Flask.

### DIFF
--- a/codalab/server/oauth2_provider.py
+++ b/codalab/server/oauth2_provider.py
@@ -151,7 +151,7 @@ class OAuth2Provider(object):
 
             oauth._validator = MyValidator()
         """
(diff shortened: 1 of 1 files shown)


percyliang:
I'm still a bit confused what impact this will have because right now I think I do `cl work main::` and it doesn't prompt me for my password even if it's been more than one day.

epicfaace (commented):
@teetone Can you explain what impact this has on the user's experience?

epicfaace:
@teetone checking on this?

PR #4128, opened 2022-06-24
Add --store validation during cl upload.

### LINKED ISSUE #4062: `cl upload` needs user-friendly message when specify wrong storage name
**Describe the bug**
Need to change the error message to user friendly message when user specify a wrong storage name in "cl upload --store xxx file.txt". For example, as the screenshot shows, there is no storage named  "gcs2"  but I use "cl upload --store gcs file.txt".  The error message  says internal error. Can we change to some user-friendly error message, eg "Might specify wrong store name"?

**Codalab Information**
- *Your username*: wwwjn
- *UUID of problem bundle (please don't delete)*: N/A
- *Error message received from Codalab (if applicable)*: Internal Server Error about Nonetype
 - *OS (if on CLI)*: MacOS
 - *Codalab CLI Version (if on CLI)*:  1.4.6
 - *Browser (if on Web)*;

**To Reproduce**
1. Specify a wrong store name in `cl upload --store store_name file.txt`, where `store_name` does not exist in current bundle stores. (List all the bundle store using `cl store ls`)

**Expected behavior**
A clear and user-friendly error message.

**Screenshots**
image
image

leilenah:

If the user types in a store that doesn't exist during `cl upload --store`, we currently show them a general `500: Internal Server Error`.

This change will let the user know that they entered in a store name that doesn't exist, as well as show them a list of valid stores.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4062

### Screenshots

https://user-images.githubusercontent.com/25855750/175749113-f4ed237b-dd9d-451f-b86e-709ad9647f57.mov

### Checklist

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -1372,6 +1372,26 @@ class BundleCLI(object):
             'metadata': metadata,
         }
 
+        # Validate store metadata.
+        store = metadata.get('store') or ''
+        if store != '':
+            print('Validating store...')
+            bundle_stores = client.fetch('bundle_stores')
+            if len(bundle_stores) == 0:
+                raise UsageError(f'{store} is not a valid store. No stores have been added.')
+
+            store_is_valid = False
+            for store_data in bundle_stores:
+                if store_data['name'] == store:
+                    store_is_valid = True
+                    break
+
+            if not store_is_valid:
+                print('Available Stores:\n')
+                self.print_table(['id', 'name', 'storage_type', 'storage_format'], bundle_stores)
+                print(' ')
+                raise UsageError(f' {store} is not a valid store.')
+
         # Option 1: --link
         if args.link:
             if len(args.path) != 1:
@@ -1399,7 +1419,7 @@ class BundleCLI(object):
                     'state_on_success': State.READY,
                     'finalize_on_success': True,
                     'use_azure_blob_beta': args.use_azure_blob_beta,
-                    'store': metadata.get('store') or '',
+                    'store': store,
                 },
             )
 
@@ -1420,7 +1440,7 @@ class BundleCLI(object):
                     'state_on_success': State.READY,
                     'finalize_on_success': True,
                     'use_azure_blob_beta': args.use_azure_blob_beta,
-                    'store': metadata.get('store') or '',
+                    'store': store,
                 },
             )
 
@@ -1486,7 +1506,7 @@ class BundleCLI(object):
                         'state_on_success': State.READY,
                         'finalize_on_success': True,
                         'use_azure_blob_beta': args.use_azure_blob_beta,
-                        'store': metadata.get('store') or '',
+                        'store': store,
                     },
                     progress_callback=progress.update,
                 )
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -923,6 +923,9 @@ def test_upload3(ctx):
     check_equals('hello', _run_command([cl, 'cat', uuid]))
     check_contains(bundle_store_name, _run_command([cl, 'info', uuid]))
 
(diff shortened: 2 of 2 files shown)


leilenah:
@wwwjn @epicfaace can either of you take a peek at this when you have a chance plz?

codalab/lib/bundle_cli.py
```diff
@@ -1372,6 +1372,26 @@ def do_upload_command(self, args):
             'metadata': metadata,
         }
 
+        # Validate store metadata.
+        store = metadata.get('store') or ''
```
epicfaace, line 1376: Thanks! I wonder -- in order to make this faster and simpler -- could we:
- not call `client.fetch('bundle_stores')` (which is an extra network request), but just catch the exception when we call the upload function?
- no need to print out available bundle stores on error, as a user can already do that with `cl store ls`. We could give a message "use cl store ls to view available bundle stores" if you'd like.

leilenah: @epicfaace from what I can see in `json_api_client.py`, the error that comes back is a generic 500 internal server error. So that that point, we don't know that it was specifically an incorrect store name that caused the error. Lmk if I'm looking in the wrong place.

If you're suggesting that we just catch the 500 and throw something generic (perhaps including `cl store ls` instructions), that seems fairly unhelpful for the user. However if you think that the extra `bundle_stores` call is going to be problematic performance-wise, I can make this change. Lmk!

leilenah: Also bear in mind that the extra `bundle_stores` call will only get called during upload for users who are specifying a --store, so users who are not doing this will not be slowed down by that extra call.

epicfaace: Perhaps we should change the backend code (https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/bundles.py#L1109) so it throws a 404 not found with a specific error about bundle store not found? Then we could catch it.

My main concern is that we're adding a new step where the user sees the "Validating bundle store..." message, and this isn't consistent with how other CLI commands do it. I guess I'd rather either 1) all CLI commands do this kind of extra call for validation or 2) all CLI commands don't, and 2) seems simplest here. What do you think?

leilenah: @epicfaace good point re: consistency. Option 1 is best for the user, but agreed that option 2 is simpler and maintains consistency. I will implement option 2 for now. If we want to go with option 1 later down the line, that can be a separate set of tasks. 👍

epicfaace: Sounds good. Note that -- I think we can do Option 2 with an equivalent user experience to Option 1 though! So long as we modify the backend code so it throws a 404 not found with a specific error about bundle store not found (in fact it would be a better user experience because the user does not need to wait for an additional request)

leilenah: @epicfaace won't we have to add a call in the backend to check if the store was incorrect in order to throw the 404? If not, how can we tell from the backend that it was an incorrect store name specifically that caused the error?

epicfaace: Yes, we can just wrap `local.model.get_bundle_store` (in [this line](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/bundles.py#L1109)) with a try-except in which if there is a not found error, we throw a 404.

leilenah: @epicfaace oh rad! on it

leilenah:
@epicfaace I just pulled the latest `master` and it looks like this issue was actually fixed recently by @wwwjn here: https://github.com/codalab/codalab-worksheets/blame/master/codalab/model/bundle_model.py#L2989

Going to close this.

PR #4102, opened 2022-05-16
Upgrade Ubuntu in Docker files

pranavjain:
This PR upgrades the Ubuntu to the 20.04 version. We can upgrade Python once this change is in to avoid any failures.

Please do check this one locally before merging.

PR #4147, opened 2022-07-13
Upgrade ubuntu to 18.04 in Dockerfile.server

epicfaace:
Splitting up https://github.com/codalab/codalab-worksheets/pull/4102 for simplicity.


epicfaace:
https://docs.sqlalchemy.org/en/13/changelog/migration_13.html#change-4393-threadlocal

PR #4148, opened 2022-07-18
Change upload expiration time to 3h

wwwjn:

Now, the signed URL for upload will expire in 1h, but when uploading large files, it needs more than 1h to upload.

### Screenshots

image

### Checklist


wwwjn:
Will set a proper expiration time after we shorten current index generating time.

PR #4150, opened 2022-07-19
Optimize & Consolidate Bundle Sidebar & Bundle View

### LINKED ISSUE #4131: Optimize & Consolidate Bundle Sidebar & Bundle View
## Documentation

[Bundle View UI Redesign](https://docs.google.com/document/d/1vqb8a2DuFqioAKwiIhws__8Y7SknZNAbqDzlSrpRJlI/edit#)

## High-Level Goals
1. Use the same code path for `BundleDetailSidebar` and `BundleView`
2. understand the purpose of each field shown in this view
3. create logical groupings, e.g. `requests` can be grouped together
4. determine the optimal UI for dense technical information
5. look into how sites with similar UI needs handle this
6. create UI mock-ups and present options to the team

### Research Tasks

1. ~~Start proposal document~~
2. ~~Review current bundle view~~
4. ~~List the different bundle types and their fields~~
5. ~~What is the purpose of each metadata field?~~
7. ~~What are the logical groupings of the fields~~
8. ~~What is most important to the user?~~
10. ~~UI wireframe~~
11. ~~UI prototype~~
12. ~~Present proposal document to Percy~~

## Implementation
1. ~~Add collapsible `command` above STDOUT~~
2. ~~Add collapsible `failure_message` below command~~
3. ~~Add command copy button w/ confirmation popup~~
4. ~~When expanded, replace more info button with bundle view button~~
5. ~~Chance <none> to <enter> for editable fields~~
6. ~~Run Bundle -- sidebar update~~
7. ~~Run Bundle -- states graphic~~
    * ~~Processing time is shown as the bundle is running~~
    * ~~States graphic should be conditional~~
    * ~~the collapse should be graceful~~
    * ~~add bundle details to tooltip~~
8. ~~Update download button UI to use icon~~
9. ~~Uploaded Bundle Sidebar~~
11. ~~Make Bundle Sidebar~~
14. ~~Load `BundleDetail` in Bundle View~~
    * ~~Bundle view should be pre-expanded~~
    * ~~Bundle page link should be hidden~~
19. ~~Code documentation~~

leilenah:

This change optimizes the UI for viewing bundle data in the worksheet view as well as the standalone bundle view. Now both views leverage the same `BundleDetail` component, which has been improved from a UI perspective. 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4131

### Bundle Detail Demo

https://user-images.githubusercontent.com/25855750/180516127-4667175e-a5eb-4570-970f-1be428750f4e.mov

### Bundle View Demo

https://user-images.githubusercontent.com/25855750/180516888-a2b52eec-d37b-4e78-aeb7-17d3eeb2c722.mov

### Checklist


frontend/src/components/worksheets/BundleDetail/BundleDependencies.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+    render() {
+        const { bundleInfo } = this.props;
+        const dependencies = bundleInfo.dependencies;
+
+        if (!dependencies.length) {
+            return <div>None</div>;
+        }
+
+        const dependencies_table = [];
```
percyliang: use camelCase if possible?

frontend/src/components/worksheets/BundleDetail/BundlePageLink.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { withStyles } from '@material-ui/core/styles';
+
+/**
+ * This component renders a simple link out to a bungle view page.
```
percyliang: typo: bundle

frontend/src/components/worksheets/BundleDetail/BundleStateTable/BundleStateBox.jsx
```diff
@@ -0,0 +1,79 @@
...
+    constructor(props) {
+        super(props);
+    }
+
+    getActiveClass() {
+        const { classes, state, isActive } = this.props;
+        if (!isActive) {
+            return '';
+        }
+        if (state == 'ready') {
```
percyliang: ===

frontend/src/components/worksheets/BundleDetail/BundleStateTable/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+class BundleStateTable extends React.Component {
+    constructor(props) {
+        super(props);
+    }
+
+    roundTime(time) {
+        return Math.abs(Math.round(time));
+    }
+
+    getTimePassed() {
```
percyliang: Can we render the time field (which is time spent actually running)?  Time since created is also meaningful, but sometimes not the thing you want.  Maybe there's a way to render both?  In general, maybe the thing to do is to render a total time and the time spent in the current stage?

Also, if we have a function for rendering a duration, could we factor this out?  Ideally, it would match the one in `formatting.py`

frontend/src/__tests__/__snapshots__/Worksheet.test.js.snap
```diff
@@ -35,7 +35,7 @@ Object {
                       class="editable-field"
                       style="color: rgb(34, 94, 168);"
                     >
-                      &lt;none&gt;
+                      &lt;enter&gt;
```
percyliang: Pull this out into a separate function...  Not sure '<enter>' is more intuitive than '<none>'...

Either way, maybe we can italicize?

percyliang (commented):
This is great - much cleaner and more uniform and helpful!

Just a few minor comments:
1. The help icon next to each field is very faint.
2. UUID maybe should go at the top before name.
3. Bundle page could appear in the upper right corner like a NE arrow that looks like it'll expand things.
4. Store should be after Size, logically
5. Allowed Failed, Exclude Patterns can go after all the Request fields.
6. The fields could better match the rendering when you create a new run (especially the grouping of the resources) - could we share a component?
7. In the help text for State, we should write the state again (e.g., "staged: ...") so it's clear we're talking about the highlighted state.  Maybe when you mouse over the states, they should show a description too.

frontend/src/__tests__/__snapshots__/Worksheet.test.js.snap
leilenah: > Pull this out into a separate function... Not sure '' is more intuitive than ''...

@percyliang can you clarify what you mean by this?

epicfaace:
My feedback: https://github.com/codalab/codalab-worksheets/issues/4153

frontend/src/components/worksheets/BundleDetail/BundleStateTable/index.jsx
leilenah: I have removed the time-since-creation logic altogether. I feel like having that plus running time may be confusing to the user.

I have added times under the `running` state and `preparing` state. Appending times to the other states seemed like the logic required wouldn't be worth the payoff but plz lmk your thoughts.

**Note:** Times are also rendered in the `More Details` section for users to reference later if needed.

leilenah:
@percyliang 

> The fields could better match the rendering when you create a new run (especially the grouping of the resources) - could we share a component?

I have modified the labels to align with the labels in the new run form. I have also added field groupings. 

> In the help text for State, we should write the state again (e.g., "staged: ...") so it's clear we're talking about the highlighted state. Maybe when you mouse over the states, they should show a description too.

Note that since `state_details` are provided by the rest API for a single bundle, that bundle won't have the details for every state in the bundle lifecycle. It only has the details for the current state. I have added a tooltip for the current state in the lifecycle diagram, but the inactive states don't have tooltips. I could hardcode the definitions to achieve this, but I wanted to avoid that.

frontend/src/components/worksheets/BundleDetail/BundleStateTable/index.jsx
percyliang: Logically this is great.  Would it be possible to move the times to the (right) side of the states, so that the state flowchart doesn't jump around when times are added?

codalab/bundles/named_bundle.py
```diff
@@ -18,7 +18,7 @@ class NamedBundle(Bundle):
     # Don't format specs
     # fmt: off
     METADATA_SPECS = [
-        MetadataSpec('name', str, 'Short variable name (not necessarily unique); must conform to %s.' % spec_util.NAME_REGEX.pattern, short_key='n',),
+        MetadataSpec('name', str, 'Bundle names can only contain letters, digits, underscores, dots and dashes. Names can only start with a letter or an underscore.', short_key='n'),
```
percyliang: How about: Short name (not necessarily unique), which must start with a letter or underscore and can only contain letters, digits, underscores, periods, and dashes".

frontend/src/components/CollapseButton.jsx
```diff
@@ -0,0 +1,55 @@
...
+import KeyboardArrowUpIcon from '@material-ui/icons/KeyboardArrowUp';
+
+/**
+ * This button is optimized to be used alongside collapsable content.
+ *
+ * Example Usage:
+ *     label='Show More'
+ *     collapsedLabel='Show Less'
+ *
+ *     When collapsed=true, the button will be labeled 'Show More'
```
percyliang: Should be dedented?

percyliang (approved):
Great work!

frontend/src/components/worksheets/BundleDetail/BundleStateTable/index.jsx
leilenah: @percyliang certainly. Updated 👍 

https://user-images.githubusercontent.com/25855750/180668812-e2fcf78d-7f90-4f75-995d-3ccf5688374c.mov

PR #4161, opened 2022-07-25
Update size_str formatting to optionally include bytes string.

### LINKED ISSUE #4114: Improve run_status: upload "147 done" should render bytes properly

leilenah:

Currently, we are incorrectly rendering the number of bytes in a bundle's `run_status`, specifically during the uploading step. For example, this is what the uploading `run_status` looks like before this change:

before

Notice the wording, `120k bytes uploaded`. This is equivalent to saying `120 kilobytes bytes uploaded`, which is inaccurate. 

#### After this Change

This change updates the `size_str` helper to apply the word `bytes` to the result if and only if:
1. The `include_bytes` flag is set to `True`
2. The number being shown to the user is bytes (as opposed to kilobytes, megabytes, gigabytes or terabytes).

This is what the above `run_status` would look like after this change:

after

This is what it would look like if 160 bytes were being uploaded:

Screen Shot 2022-07-25 at 3 19 58 PM

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4114

### Checklist

### DIFF
--- a/codalab/lib/formatting.py
+++ b/codalab/lib/formatting.py
@@ -31,9 +31,10 @@ def verbose_contents_str(input_string):
     return contents_str(input_string, verbose=True)
 
 
-def size_str(size):
+def size_str(size, include_bytes=False):
     """
     size: number of bytes
+    include_bytes: whether or not to include 'bytes' string in the return value
     Return a human-readable string.
     """
     if size is None:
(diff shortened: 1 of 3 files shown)


epicfaace (commented):
Can you share a screenshot / description of what the end user sees differently with this PR? like messages before "before" and "after" this PR gets in?

leilenah:
@epicfaace screenshots added 👍

percyliang (approved):
Great!

PR #4168, opened 2022-08-01
Normalize worksheet item sort keys during get_worksheet_info.

### LINKED ISSUE #3989: New bundle is added at a random place

leilenah:

When worksheet items are created via CLI, they are assigned a `null` sort key. When a worksheet item has a `null` sort key, the web UI will always render worksheet items with non-null sort keys above it.

Thus if a user tries to create a new worksheet item directly below a worksheet item that has a `null` sort key, the web UI will render the new item somewhere above the item with the `null` sort key instead of below it as the user intended. 

This change checks worksheet items for `null` sort keys during `get_worksheet_info` and updates those items in the db if needed.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/3989

### Before this Change

https://user-images.githubusercontent.com/10134824/152871649-3a46f325-000f-41e6-853b-421011812961.mov

### After this Change

https://user-images.githubusercontent.com/25855750/182248487-ee2328f4-3e92-4aba-9004-b1d42a0f91a0.mov

### Checklist

### DIFF
--- a/codalab/rest/worksheets.py
+++ b/codalab/rest/worksheets.py
@@ -325,6 +325,19 @@ def get_worksheets_landing():
 def get_worksheet_info(uuid, fetch_items=False, fetch_permissions=True):
     """
     The returned info object contains items which are (bundle_info, subworksheet_info, value_obj, type).
+
+    Note that this helper scans worksheet items for null sort keys and updates
+    them in the database if needed.
+
+    Context around item sort keys:
+        When bundles are created via web, they are assigned a numeric sort_key.
+        When bundles are created via CLI, they are assigned a null sort_key.
+
+        This is due to the fact that, unlike the web UI, the CLI doesn't have
+        the appropriate after_sort_key readily availble.
+
+        Instead of fetching the appropriate after_sort_key via CLI every time a
+        user creates a new bundle, we normalize sort keys here if needed.
     """
     worksheet = local.model.get_worksheet(uuid, fetch_items=fetch_items)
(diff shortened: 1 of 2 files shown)


codalab/rest/worksheets.py
```diff
@@ -334,6 +347,19 @@ def get_worksheet_info(uuid, fetch_items=False, fetch_permissions=True):
 
     # Create the info by starting out with the metadata.
     result = worksheet.to_dict()
+    items = result.get('items', [])
+
+    # Update worksheet item sort keys if needed.
+    for item in items:
+        if item['sort_key'] is None:
```
percyliang: Could we write this as `if any(item["sort_key"] is None or item in items):`?

percyliang (approved):
Great!

PR #4184, opened 2022-08-04
Bundle detail ui cleanup

### LINKED ISSUE #4169: Bundle Detail UI Fixes (6)
### Tasks
- [x] Show an error message if bundle doesn't exist
    * **Example:**  https://codalab.stanford.edu/bundles/0x449b925399d547c0b2f5ee0150a5121a
- [x] Update the state box tooltips to use the same font size as the field name tooltips
- [x] Add proper field names to tooltips
- [x] Fix Permissions overflow
    * **Example:** https://worksheets.codalab.org/bundles/0x466051121f684894bdf59ac4a0567dfc
- [x] Confirm that each field has a tooltip
    * Run Bundles
    * Uploaded Bundles
    * Make Bundles
- [x] Confirm that tooltips are showing up when viewing a bundle that you don't own
    * **Example:** Staged bundles that I don't own https://worksheets.codalab.org/worksheets/0xa590fd1b68944a1a95c1c40c4931dc7b

leilenah:

This changes cleans up the bundle detail ui. It introduces the following changes:
1. The user will now see an error message if they try to access a bundle that is no longer available
2. All tooltips in the bundle detail sidebar now use the same font size
3. Field key names have been added to field tooltips
4. The permissions dropdown no longer overflows
5. I have refined the logic that determines what bundle fields to display 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4169

#### Fix 1: Bundle Error Handling

image

#### Fix 2: Tooltip Font Sizes - prod vs local

https://user-images.githubusercontent.com/25855750/182972475-ef70d106-6001-4036-8565-672cbabc8aa8.mov

#### Fix 3: Field key names in field tooltips

Screen Shot 2022-08-04 at 5 27 34 PM

#### Fix 4: Permissions overflow -- prod vs local

https://user-images.githubusercontent.com/25855750/182973067-95b09db8-4dbf-47b2-984b-cfe8d9f24516.mov

#### Fix 5: Only display relevant fields -- prod vs local

https://user-images.githubusercontent.com/25855750/182973415-aa47b168-d2ef-496b-9614-b263b3c0b82b.mov

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
@@ -27,6 +27,12 @@ class BundleFieldRow extends React.Component {
         super(props);
     }
 
+    /**
+     * Returns true if this field should be hidden.
+     * We should only show fields that either have values or are editable.
+     *
+     * @returns {bool}
+     */
     checkHideRow() {
         const field = this.props.field || {};
         const value = this.props.value || field?.value;
--- a/frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
@@ -25,6 +25,30 @@ import Copy from '../../../Copy';
 class BundleFieldRow extends React.Component {
     constructor(props) {
         super(props);
+        this.state = {
+            title: this.getTitle(),
+        };
+    }
+
+    getTitle() {
+        const field = this.props.field || {};
+        const description = this.props.description || field.description;
+        if (!description) {
+            return;
+        }
+        const name = field.name;
+        if (!name) {
+            return description;
+        }
+        const title = (
+            <>
+                {description}
+                <br />
+                <br />
+                Search Key: '{name}'
+            </>
+        );
+        return title;
     }
 
     render() {
@@ -37,7 +61,6 @@ class BundleFieldRow extends React.Component {
 
         // allow props to override field values
         const label = this.props.label || field.name;
-        const description = this.props.description || field.description;
         const value = this.props.value || field.value;
         const copyValue = this.props.copyValue || value;
 
@@ -56,9 +79,9 @@ class BundleFieldRow extends React.Component {
                     <Typography variant='subtitle2' inline>
                         {label}
                     </Typography>
-                    {description && (
+                    {this.state.title && (
                         <Tooltip
-                            title={description}
+                            title={this.state.title}
                             classes={{ tooltip: classes.tooltipContainer }}
                         >
                             <span className={classes.tooltipIcon}>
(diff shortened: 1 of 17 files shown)


frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
```diff
@@ -27,8 +27,45 @@ class BundleFieldRow extends React.Component {
...
+        const name = field.name;
+        if (!name) {
+            return description;
+        }
+        const title = (
+            <>
+                {description}
+                <br />
+                <br />
+                Search Key: '{name}'
```
percyliang: I'm not sure 'search key' is that intuitive, because it's not just used for search, and it's not clear how you'd use this to search in the UI.  Maybe we can just put it in brackets at the end?  I feel like it should be something that shouldn't draw attention because most people won't need it, but it should be easy for experts who already know about these fields to find it.

leilenah: Brackets makes sense. If we're going to go that route tho, I think it would be better to have the field names at the beginning of the description instead of at the end tho. If we place them at the end, even experts may not realize it's there.

Screen Shot 2022-08-04 at 5 27 34 PM

Lmk your thoughts

leilenah:
@teetone I have added the `remote` field to the bundle view.
@percyliang are you able to re-review / approve?

epicfaace:
> #### eld tooltips
> Screen Shot 2022-08-04 at 5 27 34 PM

@leilenah Should we instead put the field key names in parenthesis and at the end, for readability reasons? So "Amount of memory ..... (request_memory)"

frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
epicfaace: @leilenah Should we instead put the field key names in parenthesis and at the end, for readability reasons? So "Amount of memory ..... (request_memory)". I think that starting the tooltip with something that looks like code and isn't English (`[request_memory]`) may be intimidating / not look very nice from the user's perspective.

epicfaace:
> 5\. I have refined the logic that determines what bundle fields to display

Can you point me to the code that has this new logic?

frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
leilenah: @epicfaace fair, can do

leilenah:
> > 5. I have refined the logic that determines what bundle fields to display
> 
> Can you point me to the code that has this new logic?

@epicfaace that was in reference to the [checkHideRow() function here](https://github.com/codalab/codalab-worksheets/blob/622e5b602d0ca7cc212e8ecbf698dbfb795481c3/frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx#L30-L42). The main change is that if there is no value and the field is not editable, we don't render the field row at all.

epicfaace:
> [checkHideRow() function here](https://github.com/codalab/codalab-worksheets/blob/622e5b602d0ca7cc212e8ecbf698dbfb795481c3/frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx#L30-L42). The main change is that if there is no value and the field is not editable, we don't render the field row at all.

Thanks. Can you add a small comment in the code explaining this logic?

epicfaace (approved):

Issue #4185, opened 2022-08-05
Show toggles for true / false fields in the bundle detail sidebar


leilenah:
We currently show `true` and `false` strings for boolean bundle fields. We should show a toggle like we do for `Allow Failed Dependencies` in the bundle run UI.

leilenah:
cc @pranavjain

leilenah:
This may not be worth doing after https://github.com/codalab/codalab-worksheets/issues/4203

leilenah:
Closing. Not worth it since we're going to be locking most of the editable fields after bundles reach a certain state https://github.com/codalab/codalab-worksheets/issues/4203

PR #4187, opened 2022-08-05
Bundle dependency list ui redesign.

### LINKED ISSUE #4182: Optimize bundle detail dependency list UI
### Overview
In the bundle detail sidebar, dependencies with longer names are currently rendered in a way that is unpleasing to the eye. For example here is the dependencies list in [this worksheet](https://codalab.stanford.edu/bundles/0x3b15e291c825412680540b1805716627):

Screen Shot 2022-08-03 at 3 47 03 PM

### High-Level Goal
The goal of this issue is to explore the best way to render a list of bundle dependencies, and then implement your solution. UI options should be presented to the team prior to implementation.

leilenah:

The current bundle dependency list UI doesn't make great usage of the limited space available in the bundle detail sidebar:

Screen Shot 2022-08-03 at 3 47 03 PM

The reason for this change is to only show the user what they need to know about each dependency in the most clean and concise way possible.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4182
https://github.com/codalab/codalab-worksheets/issues/4192

### Screenshots

image

### Checklist

### DIFF
--- a/codalab/rest/schemas.py
+++ b/codalab/rest/schemas.py
@@ -155,10 +155,15 @@ class BundleDependencySchema(PlainSchema):
     parent_uuid = fields.String(validate=validate_uuid)
     parent_path = fields.String(missing="")
     parent_name = fields.Method('get_parent_name', dump_only=True)  # for convenience
+    parent_state = fields.Method('get_parent_state', dump_only=True)  # for convenience
 
     def get_parent_name(self, dep):
         uuid = dep['parent_uuid']
         return local.model.get_bundle_names([uuid]).get(uuid)
+
+    def get_parent_state(self, dep):
+        uuid = dep['parent_uuid']
+        return local.model.get_bundle_state(uuid)
 
 
 class BundlePermissionSchema(Schema):
--- a/codalab/rest/schemas.py
+++ b/codalab/rest/schemas.py
@@ -8,7 +8,7 @@ from marshmallow import Schema as PlainSchema, ValidationError, validate, valida
 from marshmallow_jsonapi import Schema, fields
 from marshmallow.fields import Field
 
-from codalab.common import UsageError
+from codalab.common import UsageError, NotFoundError
 from codalab.bundles import BUNDLE_SUBCLASSES
 from codalab.lib.bundle_action import BundleAction
 from codalab.lib.spec_util import SUB_PATH_REGEX, NAME_REGEX, UUID_REGEX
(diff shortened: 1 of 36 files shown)


percyliang:
Thanks!  Is it possible to use all that whitespace under 'Dependencies'?  It does break the key/value pattern, but dependencies are sort of special enough that we might want to have them look different anyway.

In a way, like `command`, it is central to defining the run, which is moved out.

leilenah:
@percyliang this is what I'm thinking for the dependency list UI. Plz lmk your thoughts.

I also want to update `Allow Failed`, as well as the other boolean fields to use a toggle. I have created a separate issue for this https://github.com/codalab/codalab-worksheets/issues/4185

leilenah:
@percyliang that's def possible -- will prepare something to show you

leilenah:
cc @pranavjain in case you have notes

leilenah:
cc @epicfaace

docs/REST-API-Reference.md
```diff
@@ -196,6 +196,7 @@ Name | Type
 `parent_uuid` | String
 `parent_path` | String
 `parent_name` | Method
+`parent_state` | Method
```
epicfaace, line 199: What does "Method" even mean? How can a REST API return a "method"?

epicfaace: Shouldn't this just be a "string" type (and similarly for parent_name)?

codalab/rest/schemas.py
```diff
@@ -155,11 +155,19 @@ class BundleDependencySchema(PlainSchema):
...
 
     def get_parent_name(self, dep):
         uuid = dep['parent_uuid']
         return local.model.get_bundle_names([uuid]).get(uuid)
 
+    def get_parent_state(self, dep):
+        uuid = dep['parent_uuid']
+        try:
+            return local.model.get_bundle_state(uuid)
+        except NotFoundError:
```
epicfaace, line 168: when would NotFoundError be thrown? Do we need this try-except?

epicfaace (approved):

docs/REST-API-Reference.md
leilenah: @epicfaace This was auto-generated. I think it signifies that some logic runs in the schema to get this value (specifically `get_parent_name()` and `get_parent_state()`.

leilenah: String probably makes more sense in terms of the docs tho....not sure how these are generated

codalab/rest/schemas.py
leilenah: @epicfaace I observed an edge case in which `NotFoundError` is thrown [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L1252) when we try to fetch a state for dependencies that were force-deleted.

PR #4198, opened 2022-08-10
Incompatible variable type fixed

luca-digrazia:
**"filename"**: "monitor.py"
**"warning_type"**: "Incompatible variable type [9]"
 **"warning_message"**: " last_sent is declared to have type `Dict[str, float]` but is used as type `typing.DefaultDict[Variable[collections._KT], int]`."
**"warning_line"**: 137
**"fix"**: defaultdict()


monitor.py
```diff
@@ -134,7 +134,7 @@ def logs(s):
 
 
 num_errors: Dict[str, int] = defaultdict(int)
-last_sent: Dict[str, float] = defaultdict(int)
+last_sent: Dict[str, float] = defaultdict()
```
percyliang: Should we make this `defaultdict(float)`?

epicfaace (approved):
Thanks!

PR #4205, opened 2022-08-12
Move time fields to top-level in bundle detail sidebar.

### LINKED ISSUE #4192: time and dependencies should be shown without having to click 'more detail'
Currently, you have to click 'more detail' to see time and dependencies.  Dependencies are special and can be shown closer to the top, although for some bundles, there could be a lot of dependencies, which could also make it bulky...

Time is currently grouped nicely with the other time that makes sense, so maybe we should leave that there, but put an additional timing next to the state or something?

In general, there is a complete verbose view where information is grouped accordingly (like right now), but the leaner, simpler view (less detail) should hit the basic things including time, dependencies, and remote (at least when the job is running).

leilenah:

Currently in the bundle detail sidebar, a user has to click **More Detail** in order to see the **Time** fields. This change moves the **Time** fields beneath the rest of the top-level bundle fields for better visibility.

This change also removes the **MoreDetails** component, and moves the extra fields directly into **BundleDetailSidebar** (which feels more intuitive).

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4192

### Screenshots

image

https://user-images.githubusercontent.com/25855750/184431340-d6c9aeba-3875-4f1f-925a-ce41b9cd8b9d.mov

### Checklist


percyliang:
I thought we were planning to just remove the 'more details' altogether to minimize clicks?

(Main downside is that it might get pretty long.)

frontend/src/components/worksheets/BundleDetail/BundleDetailSideBar.jsx
```diff
@@ -98,12 +125,138 @@ class BundleDetailSideBar extends React.Component {
...
+                                    label='Queue'
+                                    field={bundle.request_queue}
+                                    onChange={(request_queue) => onUpdate({ request_queue })}
+                                />
+                                <BundleFieldRow
+                                    label='Priority'
+                                    field={bundle.request_priority}
+                                    onChange={(request_priority) => onUpdate({ request_priority })}
+                                />
+                                <BundleFieldRow
```
percyliang, line 201: This is always true now, so maybe we can just remove?

percyliang (approved):

leilenah:
> I thought we were planning to just remove the 'more details' altogether to minimize clicks?
> 
> (Main downside is that it might get pretty long.)

@percyliang I considered that, but then I remembered seeing a bundle that had a very long list of host-worksheets, which gave me pause 😅

percyliang:
I think the 'show more' should be done at the level of individual fields.

PR #4207, opened 2022-08-12
Prominently display bundle status in the bundle detail sidebar

### LINKED ISSUE #4186: Show status of PREPARING
We used to display why a bundle was preparing: downloading dependency, downloading docker image, etc.

I don't see this information anymore on the Bundle View page.

### LINKED ISSUE #4195: Fix floating empty tooltip issue in bundle view sidebar
Image

leilenah:

Currently in the bundle detail sidebar, if you want to see a running bundle's status, you must hover over the bundle's current state in the dynamic bundle state diagram. 

This is unintuitive and cumbersome. This change makes a bundle's status more prominent when a bundle is running. 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4186
https://github.com/codalab/codalab-worksheets/issues/4195

### Screenshots

https://user-images.githubusercontent.com/25855750/184579038-579bc118-ec17-48d6-acc3-d005a0d69031.mov

### Checklist


percyliang:
Informationally, this is great.  Aesthetically, some comments:
- The white box where the text has padding looks a bit odd since it's not aligned.
- When we the status wraps and changes, the rest of the fields below jump around, which is not ideal.
I wonder if can get rid of 'Status' (since it's not an official field), put the status in italics and have it take up more width, centering it below the diagram?

epicfaace:
Looks good to me, my main concern is ensuring that the "Status" box doesn't get too large in height (in case the message is super long).

percyliang:
We can allocate enough space for the message to be 1-2 lines most of the time without things jumping around.  If it goes past that (hopefully it's rare), then it's okay if we wrap and push down the rest of the lines.

leilenah:
@percyliang 👍 centered / italics looked a bit off as the text changed, so instead I reduced the font size and lightened the text slightly to differentiate it. I've also added a min-height of 50. I tested this min-height with the current longest status and we should be good in terms of jumpiness.

Please see new screen recording in desc.

cc @epicfaace

percyliang:
Looks great!  The only thing I'd suggest is to remove the '[status]' prefix.

percyliang (approved):

PR #4212, opened 2022-08-16
POC: parallel file uploading and index creation

epicfaace:

For @wwwjn -- here's a POC that does both file uploading and index creation in parallel, from a single file stream.


epicfaace:
We need to first implement https://github.com/pauldmccarthy/indexed_gzip/issues/102 before this works, though.

epicfaace:
@wwwjn I'm closing this as we already have https://github.com/codalab/codalab-worksheets/pull/4370 now.

PR #4220, opened 2022-08-18
Show bundle `run_status` in `preparing` state

### LINKED ISSUE #4113: Improve run_status: show the percentage of docker image downloaded

### LINKED ISSUE #4186: Show status of PREPARING

leilenah:

Currently we show a static status for bundles that are in the `preparing` state. However there is useful information that is appended to `run_status` for bundles that are `preparing`, so we should show `run_status` for `preparing` bundles.

This change enables `run_status` to be shown to the user when a bundle is in the `preparing` state. It also refines the progress messaging that is appended to `run_status` when a docker image is being pulled.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4186
https://github.com/codalab/codalab-worksheets/issues/4113

#### Docker image is being downloaded

downloading-sample

#### Docker image is being extracted

extracting-sample

### Checklist

### DIFF
--- a/codalab/lib/bundle_util.py
+++ b/codalab/lib/bundle_util.py
@@ -371,9 +371,9 @@ def get_bundle_state_details(bundle):
     See Bundle-Lifecycle.md for public bundle state documentation.
     """
     metadata = bundle.get('metadata', {})
-    run_status = metadata.get('run_status', '')
-    type = bundle.get('bundle_type')
+    run_status = metadata.get('run_status')
     state = bundle.get('state')
+    type = bundle.get('bundle_type')
     state_details_by_type = {
         'dataset': {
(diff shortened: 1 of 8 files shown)


percyliang:
Nice!  Now that the message is getting pretty long, could we make it use the horizontal space more effectively (wrap less)?

percyliang (approved):

leilenah:
@percyliang I've been experimenting with breaking the key-value pattern for longer values like this so we can take advantage of the full width of the sidebar, but I haven't been able to get it to look quite right.

I will give this more thought as we iterate on the bundle view.

PR #4225, opened 2022-08-18
Bump to version 1.5.6

leilenah:
## Frontend

* Normalize worksheet item sort keys during get_worksheet_info. (#4168)

## CLI

None

## Backend

None

## Dev / docs / CI

None


leilenah:
@wwwjn not sure that we should be updating `codalab/lib/upload_manager.py` in this PR cc @epicfaace

wwwjn:
> codalab/lib/upload_manager.py

Should we only update the version number in "bump version" PR?

codalab/lib/upload_manager.py
```diff
@@ -227,7 +227,7 @@ def write_fileobj(
             os.environ['AZURE_STORAGE_CONNECTION_STRING'] = bundle_conn_str
         try:
             bytes_uploaded = 0
-            CHUNK_SIZE = 16 * 1024
+            CHUNK_SIZE = 4 * 1024
```
epicfaace: That's right, I think these changes look unintentional and should be reverted.

wwwjn: Thanks Ashwin and Leilenah! I will revert this (it's for local testing purpose)

wwwjn (approved):

PR #4229, opened 2022-08-22
Return empty state_details for deprecated bundle types.

### LINKED ISSUE #4228: [USER ISSUE] Unable to fetch data for worksheet that exists in the database
**Describe the bug**

**Note:** the user reported this issue to Percy via email.

Codalab is unable to load this worksheet: https://worksheets.codalab.org/worksheets/0x269ef752f8c344a28383240f7bb2be9c

However the worksheet is present in the user's worksheet list: https://worksheets.codalab.org/users/yushiw

It is also present in the prod database.
image

**Codalab Information**
- *Your username*: 
- *UUID of worksheet where problem occurred*:
 - *OS (if on CLI)*:
 - *Browser (if on Web)*;

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

leilenah:

There are old worksheets that contain `program` bundles. This bundle type has been deprecated. However, we should not throw an internal error if an old worksheet contains this type of bundle. 

This change updates the `get_bundle_state_details` function to return an empty string for deprecated bundle types.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4228
https://github.com/codalab/codalab-worksheets/issues/4235

### Checklist

### DIFF
--- a/codalab/lib/bundle_util.py
+++ b/codalab/lib/bundle_util.py
@@ -398,6 +398,9 @@ def get_bundle_state_details(bundle):
             'worker_offline': 'The worker where the bundle is running on is offline, and the worker might or might not come back online.',
         },
     }
+
+    # We can remove the defensive checks below once program bundles are converted to dataset bundles.
+    # Related Issue: https://github.com/codalab/codalab-worksheets/issues/4235
     state_details = state_details_by_type.get(type, {}).get(state, '')
 
     if state == 'preparing' or state == 'running':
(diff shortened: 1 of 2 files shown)


percyliang (commented):
Good catch!  Ideally, we would do a database migration to get rid of `program` so that the code can remain as simple as possible.  Otherwise, this change might lead to some silent failures.

percyliang (approved):

PR #4232, opened 2022-08-23
Truncate `remote` field and make it copyable

### LINKED ISSUE #4230: Make `remote` field copyable in bundle detail sidebar

leilenah:

The `remote` field is often long. The user doesn't need to be able to read the whole field, as it is usually a UUID. This changes truncates that uuid and makes it copyable (just like we do for the bundle's uuid).

This change also allows staged bundles to be killed from the bundle content CTAs.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4230

#### Before
Screen Shot 2022-08-22 at 5 49 32 PM

#### After

Screen Shot 2022-08-23 at 8 09 50 PM

Screen Shot 2022-08-23 at 8 08 59 PM

### Checklist

### DIFF
--- a/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
+++ b/frontend/src/components/worksheets/BundleDetail/BundleActions.jsx
@@ -87,14 +87,15 @@ class BundleActions extends React.Component<{
 
     render() {
         const { bundleInfo, classes, editPermission } = this.props;
+        const state = bundleInfo.state;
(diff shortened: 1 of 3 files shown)


percyliang:
Sometimes the remote is the hostname, so maybe we can show more of it (as much of it as you can without having it spill over)?  What do you think, @teetone?

leilenah:
@percyliang see revised screenshots in description

percyliang (approved):
Great!

PR #4119, opened 2022-05-31
Bypass server upload for GCS

wwwjn:
### Reasons for making this change

Issue #4074 

- Reformat Azure bypass server upload. 
- Add support for GCS bypass server

### Related issues

#4074

### Checklist

* [x] I've added a screenshot of the changes, if this is a frontend change
* [x] I've added and/or updated tests, if this is a backend change
* [x] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [x] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

PR #4239, opened 2022-08-27
[TEST] Fix dev environment upload timeout

wwwjn:
Solve #4119 

### Checklist


epicfaace (commented):
Great -- did you test and did it fix the issue on dev?

wwwjn:
> Great -- did you test and did it fix the issue on dev?

Yeah I tested it using dev environment, but it does not work for us unfortunately

wwwjn:
Close this because the timeout issue has been fixed.

PR #4242, opened 2022-09-01
Fix staged_status and add it to bundle state details

### LINKED ISSUE #4139: Fix `staged_status` and add it to bundle state details
### Implementation
- [x] Remove `staged` from `get_bundle_state_details()`
- [x] Update `get_bundle_state_details()` to return `staged_status` when the bundle is `staged`
- [x] Show generic `staged_status` when we transition to `staged`
    * "Bundle’s dependencies are all ready. Waiting for the bundle to be assigned to a worker to be run."
- [x] If we found an optimal worker for the bundle, update `staged_status` to reflect this
- [x] Comb through `_filter_and_sort_workers()` and update `staged_status` where applicable
- [x] Write `_generate_resource_recommendations` tests
- [x] Update `_filter_and_sort_workers()` tests
- [x] Update `get_bundle_state_details()` tests
- [x] QA

### Overview

`staged_status` isn't consistently getting added to bundle metadata. We need to figure out why that is occurring, resolve the issue. We also need to show `staged_status` to users as part of `state_details`.

### Acceptance Criteria
1. `staged_status` is present whenever a bundle is in a `staged` state
2. staged statuses are clear and accurate
3. `state_details` reflects `staged_status` when a bundle is in `staged` state
    * this logic will be similar to how we show `run_status` when a bundle is `running`
    * see `get_bundle_state_details` helper in `bundle_util.py` 

### Related Issue
https://github.com/codalab/codalab-worksheets/issues/4116

### Related PR
https://github.com/codalab/codalab-worksheets/pull/4127

leilenah:

We currently show a static message for bundles that are in the `staged` state. However if a bundle is in this state for a while, it is helpful for the user to know why their bundle is `staged`.

This change enables the user to see `staged_status` for bundles that are staged. It also adds useful information to `staged_status` during the worker filtering / sorting process.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4139

### Screenshots

Screen Shot 2022-09-05 at 3 35 32 PM

### Checklist


codalab/server/bundle_manager.py
```diff
@@ -120,7 +120,8 @@ def _run_iteration(self):
         self._fail_unresponsive_bundles()
 
     def _set_staged_status(self, bundle, staged_status):
-        self._model.update_bundle(bundle, {'metadata': {'staged_status': staged_status}})
+        if bundle.BUNDLE_TYPE == 'run':
```
percyliang: Shouldn't this function only be called when the bundle is a run?

codalab/server/bundle_manager.py
```diff
@@ -553,14 +561,49 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
...
+    def _create_bundle_resource_snapshot(bundle_resources):
+        """
+        Given a RunResources instance, returns a tuple of requested bundle
+        resources including CPUs, GPUs, memory and disk space.
+        """
+        cpus = bundle_resources.cpus
+        gpus = bundle_resources.gpus
+        memory = formatting.size_str(bundle_resources.memory)
+        disk = formatting.size_str(bundle_resources.disk)
+        return f'({cpus}, {gpus}, {memory}, {disk})'
```
percyliang: I think these tuples are not very easy to read...

codalab/server/bundle_manager.py
```diff
@@ -553,14 +561,49 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
                 worker['memory_bytes'] -= bundle_resources.memory
         return workers_list
 
+    @staticmethod
+    def _create_bundle_resource_snapshot(bundle_resources):
```
percyliang: Should also include the `tag`, which determines which queue a run is on.

codalab/server/bundle_manager.py
```diff
@@ -553,14 +561,49 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
...
+        resources including CPUs, GPUs, memory and disk space.
+        """
+        cpus = bundle_resources.cpus
+        gpus = bundle_resources.gpus
+        memory = formatting.size_str(bundle_resources.memory)
+        disk = formatting.size_str(bundle_resources.disk)
+        return f'({cpus}, {gpus}, {memory}, {disk})'
+
+    @staticmethod
+    def _create_workers_resource_snapshot(workers_list):
```
percyliang: Please add type annotations to everything.

codalab/server/bundle_manager.py
```diff
@@ -553,14 +561,49 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
...
+        """
+        Given a list of workers, returns a tuple of worker resource information
+        including CPUs, GPUs, memory and disk space for each worker.
```
percyliang: There's also the tag of the worker (corresponding to `--request-queue` on the job) that we should include.

codalab/server/bundle_manager.py
```diff
@@ -553,14 +561,49 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
...
+        why a suitable worker was not identified.
         """
         # Filter by tag.
-        if bundle.metadata.request_queue:
-            workers_list = self._get_matched_workers(bundle.metadata.request_queue, workers_list)
+        request_queue = bundle.metadata.request_queue
+        if request_queue:
+            workers_list = self._get_matched_workers(request_queue, workers_list)
+            if not workers_list:
+                staged_status = f"Worker {request_queue} not found. Run 'cl workers' to see information about workers that you have connected to the CodaLab instance."
```
percyliang: I think this logic should be handled as part of resources.

codalab/server/bundle_manager.py
```diff
@@ -570,6 +613,20 @@ def _filter_and_sort_workers(self, workers_list, bundle, bundle_resources):
                 for worker in workers_list
                 if not worker['tag_exclusive'] or not worker['tag']
             ]
+            if not workers_list:
+                staged_status = 'No workers available. Try connecting a worker to the CodaLab instance and using that to run this bundle.'
+                self._set_staged_status(bundle, staged_status)
+                return []
+
+        # Filter by the number of jobs allowed to run on this worker.
+        workers_list = [worker for worker in workers_list if worker['exit_after_num_runs'] > 0]
```
percyliang: Are there workers with exit_after_num_runs = 0?  This shouldn't be the case?

codalab/server/bundle_manager.py
```diff
@@ -570,6 +613,20 @@ def _filter_and_sort_workers(self, workers_list, bundle, bundle_resources):
...
+        if not workers_list:
+            staged_status = 'All workers have 0 runs left. This run cannot be accomodated.'
+            self._set_staged_status(bundle, staged_status)
+            return []
+
+        # Keep a copy of available workers in case we need to add this info to staged_status.
+        available_workers = workers_list.copy()
```
percyliang: Don't need to `copy`; better to make the subsequent logic immutable.

codalab/server/bundle_manager.py
```diff
@@ -977,25 +1039,19 @@ def _get_staged_bundles_to_run(self, workers, user_info_cache):
...
-                            bundle,
-                            {
-                                'metadata': {
-                                    'staged_status': 'Bundle is requested to run on a worker {} which has '
-                                    'not been connected to the CodaLab server yet.'.format(
-                                        bundle.metadata.request_queue
-                                    )
-                                }
-                            },
+                        staged_status = 'Bundle is requested to run on a worker {} which has not been connected to the CodaLab server yet.'.format(
```
percyliang: I feel like this logic shouldn't be here; if queue is handled as part of resources, this should be in the part of the same part of the code that reports when there aren't enough GPUs/CPUs, etc.

percyliang (changes requested):
I think it would be helpful first to define a dataclass `Resource` which has cpus, gpus, memory, disk, queue.  Then each `worker` object can be converted into a `Resource` and the bundle we're trying to schedule can be converted into a `Resource`.  Then we can define a function `def dominates(self, other: Resource)` which returns True if each dimension (cpus, gpus, etc.) for `self` is at least as great as `other`.  Then,
1) We compute the list of feasible workers (those that dominate the bundle).  If this is non-empty, we're done.
2) If not, then we take all the workers and keep only those workers that aren't dominated by another worker.  For each of these potential workers, we generate a string (`worker_resource.render_shortcomings(bundle_resource)`) which lists all the dimensions where worker_resource falls short.  For example, if worker_resource has gpus=2 and bundle_resource has gpus=4, then we'd generate something like "2 GPUs (4 requested)".  So then the final staged status would be something like "Insufficient resources, but the following workers are available: ...".

Let me know if this makes sense!

tests/unit/server/bundle_manager_test.py
```diff
@@ -257,3 +257,24 @@ def test_get_matched_workers_empty_tag(self):
...
+        workers_snapshot = BundleManager._create_workers_resource_snapshot(workers_list)
+        self.assertEqual(
+            workers_snapshot,
+            '(4, 2, 3.9k, 3.9k), (4, 1, 3.9k, 3.9k), (4, 0, 3.9k, 3.9k), (6, 0, 3.9k, 3.9k), (6, 0, 3.9k, 3.9k), (6, 0, 2.0k, 2.0k), (6, 0, 2.0k, 2.0k), (6, 1, 2.0k, 2.0k)',
+        )
+
+    def test_create_workers_resource_snapshot_one_worker(self):
+        workers_list = self.get_sample_workers_list()
+        workers_snapshot = BundleManager._create_workers_resource_snapshot([workers_list[0]])
+        self.assertEqual(workers_snapshot, '(4, 2, 3.9k, 3.9k)')
```
epicfaace: Let's also add a test in test_cli.py

codalab/lib/bundle_util.py
```diff
@@ -389,7 +390,6 @@ def get_bundle_state_details(bundle):
         },
         'run': {
             'created': 'Bundle has been created but its contents have not been populated yet.',
-            'staged': 'Bundle’s dependencies are all ready. Waiting for the bundle to be assigned to a worker to be run.',
```
epicfaace, line 392: Why removed?

codalab/lib/bundle_util.py
```diff
@@ -399,10 +399,13 @@ def get_bundle_state_details(bundle):
         },
     }
 
-    # We can remove the defensive checks below once program bundles are converted to dataset bundles.
-    # Related Issue: https://github.com/codalab/codalab-worksheets/issues/4235
-    state_details = state_details_by_type.get(type, {}).get(state, '')
```
epicfaace, line 404: Why move this part down?

codalab/server/bundle_manager.py
```diff
@@ -553,14 +561,49 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
                 worker['memory_bytes'] -= bundle_resources.memory
         return workers_list
 
+    @staticmethod
+    def _create_bundle_resource_snapshot(bundle_resources):
+        """
+        Given a RunResources instance, returns a tuple of requested bundle
```
epicfaace: This function doesn't return a tuple -- it returns a string...

codalab/lib/bundle_util.py
leilenah: whenever something is in a staged state, we show the `staged_status` now, so we don't need this static message

codalab/lib/bundle_util.py
leilenah: We only need to set `state_details` if we aren't returning `staged_status` or `run_status`. Also once the migration noted is complete, I'm gonna just `return state_details_by_type[type][state]` so `state_details` var will go away

codalab/server/bundle_manager.py
leilenah: There is an edge case in which `make` bundles are staged: https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L191-L196

I have removed this defensive check and updated the `MakeBundle` class to allow `staged_status` metadata.

codalab/server/bundle_manager.py
leilenah: Seems like a possible edge case: https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L513

tests/unit/server/bundle_manager_test.py
leilenah: Nothing is changing from a cli perspective. Is there a specific cli test that you think would be useful to add?

leilenah:
@percyliang updated. Instead of creating a new `Resource` class, I hijacked the existing `RunResources` class. Lmk your thoughts.

codalab/server/bundle_manager.py
```diff
@@ -553,47 +562,86 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
                 worker['memory_bytes'] -= bundle_resources.memory
         return workers_list
 
+    @staticmethod
+    def _worker_to_run_resources(worker):
```
percyliang, line 565: Could we type annotate?

codalab/server/bundle_manager.py
```diff
@@ -553,47 +562,86 @@ def _deduct_worker_resources(self, workers_list, running_bundles_info):
...
+            gpus=worker['gpus'],
+            memory=worker['memory_bytes'],
+            disk=worker['free_disk_bytes'],
+            runs_left=worker['exit_after_num_runs'],
+            docker_image=None,
+            time=None,
+            network=None,
+        )
+
+    def _get_dominating_workers(self, run_resources, workers_list, strict=False):
```
percyliang, line 584: Type annotate please

codalab/worker/bundle_state.py
```diff
@@ -160,8 +165,73 @@ def from_dict(cls, dct):
...
+
+        if self.cpus < run_resources.cpus:
+            return False
+        if self.gpus < run_resources.gpus:
+            return False
+        if self.memory < run_resources.memory:
+            return False
+        if self.disk < run_resources.disk:
+            return False
+        return True
```
percyliang: Put this in `else` to make it more parallel to the if clause

codalab/worker/bundle_state.py
```diff
@@ -160,8 +165,73 @@ def from_dict(cls, dct):
...
 
+    def dominates(self, run_resources, strict=False):
+        """
+        Accepts a RunResources instance and a strict bool.
+        Returns True if self's resources dominate resources in run_resources.
+
+        If strict = True, returns False if self's resources and resources in
+        run_resources are equivalent.
+        """
+        if self.runs_left == 0:
```
percyliang, line 205: Importantly we should check the request_queue/tag.

codalab/worker/bundle_state.py
```diff
@@ -160,8 +165,73 @@ def from_dict(cls, dct):
...
+        if self.disk < run_resources.disk:
+            return False
+        return True
+
+    def get_comparison(self, run_resources):
+        """
+        Accepts a RunResources instance.
+
+        Compares resources in self with resources in run_resources and returns
+        a comparison between the two.
```
percyliang: Be more specific about what constitutes a comparison?

percyliang (approved):
Logic looks very nice and clean!  Just a few minor comments (including one more major one about the queue, which is important).

PR #4243, opened 2022-09-02
Simplify `isLoading` logic in `MainContent` component

### LINKED ISSUE #4215: Simplify content loading logic in `MainContent` component
On the frontend, we fetch bundle data on an interval. This is by design, as bundle info is constantly changing as the bundle moves through various states. However, once a bundle is in a final state, we should stop refetching its data as the data should not change at that point.

Once we do ☝ we can also get rid of the wonky `isLoading()` logic [here](https://github.com/codalab/codalab-worksheets/blob/master/frontend/src/components/worksheets/BundleDetail/MainContent.jsx) and rely on `inFinalState` and `fetchingContent`.

leilenah:

A few weeks ago, I hastily added an `isLoading()` function to the `MainContent` component as a quick patch. This change simplifies that cumbersome logic. 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4215

### Screenshots

https://user-images.githubusercontent.com/25855750/188035967-67f134a7-adc3-44bd-b4b6-453923c40560.mp4

### Checklist

### DIFF
--- a/codalab/server/bundle_manager.py
+++ b/codalab/server/bundle_manager.py
@@ -167,11 +167,11 @@ class BundleManager(object):
                 ]
                 failure_message = ''
                 if failed_uuids:
-                    failure_message += ' Parent bundles failed: %s' % ', '.join(failed_uuids)
+                    failure_message += 'Parent bundles failed: %s ' % ', '.join(failed_uuids)
                 if killed_uuids:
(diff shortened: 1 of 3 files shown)


percyliang (approved):
Great, thanks!

PR #4246, opened 2022-09-07
Sync bundle row state and bundle detail state

### LINKED ISSUE #4174: State field doesn't refresh quickly in the table header view v/s bundle detail view.
Table Header View State: Uploading
Bundle View State: Ready
image

leilenah:

The state shown in the bundle row is currently not synced with the state shown in the expanded bundle details. This change ensures that the bundle row and the expanded bundle details will always be in sync.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4174

#### Before (prod)

https://user-images.githubusercontent.com/25855750/188984560-5a1ef10e-a1fc-42d7-b086-ebd5d9046577.mp4

#### After (local)

https://user-images.githubusercontent.com/25855750/188984600-2adff012-ed1c-446a-b1a5-e80a8e4e4c6b.mp4

### Checklist


percyliang (commented):
I think the `stateInfo` terminology might be confusing, because `state` is a specific field within a bundle...

percyliang:
Could you describe at a high-level what the fix was?

leilenah:
@percyliang the worksheet view fetches basic bundle info on an interval in order to keep the rows up-to-date. The BundleDetail view also fetches basic bundle info on an interval (as well as more detailed information that the worksheet view doesn't need to fetch).

The issue is that the fetching isn't synced, so the fix I implemented is to allow the worksheet to dictate what state to display in the bundle detail view if necessary. 

Note that I've updated some `stateInfo` variable names to make things more clear.

percyliang (commented):
Got it.  Would you mind updating the code with a comment that describes how the bundle information gets propagated from the table to the bundle detail page to keep things synchronized?

frontend/src/components/worksheets/items/TableItem/BundleRow.js
```diff
@@ -117,6 +118,7 @@ class BundleRow extends Component {
             focusIndex,
             ws,
         } = this.props;
+        const stateInfoFromWorksheet = getStateInfo(bundleInfo);
```
percyliang: Would it make more sense to pass the `bundleInfo` directly down into `BundleDetail`?  `stateInfo` right now is a helper that mixes in rendering logic.  I think perhaps we want to keep things more semantic and let `BundleDetail` deal with how it wants to render the information.  This way, we also don't have another type `StateInfo` to reason about.

leilenah:
@percyliang refactored a bit. lmk if this works

frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleStateRow.jsx
```diff
@@ -14,36 +20,45 @@ import BundleFieldRow from './BundleFieldRow';
...
         }
         if (bundleType === 'make') {
-            return ['created', 'making', 'ready'];
+            return MAKE_BUNDLE_STATES;
         }
         return [];
     }
 
+    showTime(time) {
+        return time && time !== '<none>' && time !== '0s';
```
percyliang: It seems like we should be able to access the raw value rather than the rendered one to avoid these rather brittle comparisons, so we could test whether `time == null` or something...

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
```diff
@@ -204,6 +204,25 @@ const BundleDetail = ({
...
     };
 
+    /**
+     * This helper syncs the state info that is used to render the bundle row
+     * with the state info that is passed down into the bundle detail sidebar.
+     *
+     * This enables the bundle row and the bundle detail sidebar to show the
+     * exact same state information.
+     */
+    const syncBundleStateInfo = () => {
```
percyliang, line 214: This only is called once, so why not just inline?

percyliang (approved):
A few minor comments, but overall looks good to me - thanks!

frontend/src/components/worksheets/BundleDetail/BundleDetail.jsx
leilenah: I feel like wrapping the logic in a named function makes it easier to figure out for someone else who might stumble across the code later.

frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleStateRow.jsx
leilenah: will update the `renderFormat` helper to allow for this 👍

PR #4250, opened 2022-09-10
DO NOT MERGE

AndrewJGaut:
Adds a new command to the CLI that prints out the name and UUID of the current bundle and its ancestors in the bundle dependency graph.

Users want an easy way to view the bundle dependency graph through the CLI.

### Related issues

This is a takehome assignment that is referenced [here](https://gist.github.com/percyliang/5bf88a5c054b9689b61c79b81bd16ff4).

### Checklist

* [x ] I've added a screenshot of the changes, if this is a frontend change
* [x ] I've added and/or updated tests, if this is a backend change
* [x ] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [x ] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -110,6 +110,7 @@ BUNDLE_COMMANDS = (
     'run',
     'edit',
     'detach',
+    'ancestors',
     'rm',
     'search',
     'ls',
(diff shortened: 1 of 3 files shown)


epicfaace:
Thanks!

epicfaace:
@AndrewJGaut not sure why tests are failing?

AndrewJGaut:
I think it should work now (if you approve re-running the CI tests). Basically, when running pre-commit.sh, I got an error when building the mysqlclient wheel (due to a missing ld library). So, I temporarily commented the line with mysqlclient out in the server-requirements.txt file and re-ran the pre-commit.sh script, intending to fix the issue before I submitted the pull request. However, I forgot about it.

I fixed that issue and verified that all test_runner.py default tests (including the new one I added for ancestors) pass for me locally. So, I believe it will work now.

PR #4252, opened 2022-09-12
Disable editing for certain fields once bundle reaches `starting` state

### LINKED ISSUE #4203: Disable editing for certain fields once bundle reaches `starting` state
### Implementation
- [x] Update get_editable_metadata_fields() to take state into consideration when determining if a field is editable
- [x] Determine what the web UI should show once fields are frozen
- [x] Allow metadata specs to be formatted
- [x] Get missing metadata descriptions somehow
- [x] Fix empty tags rendering for logged out user
- [x] Update the way we show 0 data size in bundle row
- [x] Fix staged status
- [x] QA web logged in
- [x] QA web logged out
- [x] Write tests for fill_missing_metadata()
- [x] Write tests for get_editable_metadata_fields()

All of the editable fields that are in the `More Details` section of the bundle detail sidebar should no longer be editable once a bundle reaches the `starting` state.  

All of the top-level editable fields that are above the `More Details` section can remain editable. This applies to the frontend UI as well as the CLI.

leilenah:

Currently, all bundle fields that aren't generated by CodaLab are editable by the user indefinitely. However, there are certain fields that don't make sense for the user to edit once a bundle has started running, e.g. `store`. 

That being said, there are some fields that do make sense for the user to edit indefinitely such as `name` and `description`. 

This change locks the following editable fields once a bundle has reached the `starting` state:

```
request_docker_image
request_time
request_memory
request_disk
request_cpus
request_gpus
request_queue
request_priority
request_network
exclude_patterns
store
allow_failed_dependencies
```

All other editable fields outside of the list above will remain editable indefinitely. 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4203

#### Bundle Before `starting` State

image

#### Bundle After `starting` State

image

### Checklist


codalab/lib/formatting.py
```diff
@@ -37,7 +37,7 @@ def size_str(size, include_bytes=False):
     include_bytes: whether or not to include 'bytes' string in the return value
     Return a human-readable string.
     """
-    if size is None:
+    if size is None or size == 0:
```
percyliang: Why add this?  0 bytes is different from not knowing the size.

codalab/lib/metadata_util.py
```diff
@@ -25,7 +25,12 @@ def fill_missing_metadata(bundle_subclass, args, initial_metadata):
     '''
     # Fill in default values for all unsupplied metadata keys.
     new_initial_metadata = {}
-    for spec in bundle_subclass.get_user_defined_metadata():
+    for spec in bundle_subclass.METADATA_SPECS:
```
percyliang: What's the motivation for this change?

codalab/lib/worksheet_util.py
```diff
@@ -213,18 +213,23 @@ def get_formatted_metadata(cls, metadata, raw=False, show_hidden=False):
     return result
 
 
-def get_editable_metadata_fields(cls):
+def get_editable_metadata_fields(cls, state):
```
percyliang: Type annotate

frontend/src/util/worksheet_utils.js
```diff
@@ -70,6 +70,9 @@ export function renderSize(size, includeBytes = false) {
     // size: number of bytes
     // includeBytes: whether or not to include 'bytes' string in the return value
     // Return a human-readable string.
+    if (!size) {
```
percyliang: Be careful about supporting 0 properly

codalab/lib/formatting.py
leilenah: From a formatting perspective, I feel that showing a size of `'0'` isn't useful information for the user. So instead of always checking for size `'0'` everywhere, I thought it would be better to just not return a size in that case.

codalab/lib/metadata_util.py
leilenah: There were certain metadata fields that weren't getting returned to the frontend because the bundle didn't have a value set for that field. E.g. there is no `data_size` util the bundle is in the `running` state. Since the metadata descriptions are pulled from the backend, this resulted in some fields not having a tooltip / description until that field was set for the bundle.

The fix that I came up with was to set default values for those missing fields so that they're always returned. This change enables that.

codalab/lib/worksheet_util.py
leilenah: Can you confirm what you mean by this plz? I thought this meant adding param docs like `:param cls:` and `:param state:`

Do you mean that I should set a default like `state = ''`?

frontend/src/util/worksheet_utils.js
leilenah: Not supporting `0` was intentional 👀 e.g. in the bundle row instead of showing `0` we just leave it blank. And in the detail sidebar, instead of showing `0` we show `<none>`

codalab/lib/formatting.py
percyliang: There is a meaningful difference between `None` and `0`.  You can have a job that outputs a file of `0` size, which is different from that file not existing.

leilenah:
@percyliang updated

frontend/src/components/worksheets/BundleDetail/BundleFieldTable/BundleFieldRow.jsx
```diff
@@ -27,43 +27,22 @@ class BundleFieldRow extends React.Component {
...
-        if (!field.editable) {
-            if (!value) {
-                return true;
-            }
-            if (field.type === 'list' && (!value.length || !value[0])) {
-                return true;
-            }
+    getValue(field) {
+        const value = this.props.value || field.value;
+        if (value || value === false) {
```
percyliang: This looks strange - should this be `[false, true].includes(value)`?

leilenah: mm not quite. We want to return anything truthy and also `false`.

leilenah: actually it looks like bools get changed to a string, so I can remove the `false` check 👍

leilenah:
@percyliang value check updated

codalab/bundles/run_bundle.py
```diff
@@ -21,65 +21,301 @@
 class RunBundle(DerivedBundle):
     BUNDLE_TYPE = 'run'
     METADATA_SPECS = list(DerivedBundle.METADATA_SPECS)  # type: List
-    # Note that these are strings, which need to be parsed
-    # Request a machine with this much resources and don't let run exceed these resources
-    # Don't format metadata specs
-    # fmt: off
-    METADATA_SPECS.append(MetadataSpec('request_docker_image', str, 'Which docker image (either tag or digest, e.g., codalab/default-cpu:latest) '
```
percyliang, line 28: I personally think that the original formatting is easier to read and see all the fields at one glance...what do you think?

frontend/src/util/worksheet_utils.js
```diff
@@ -66,27 +66,42 @@ export function renderDate(epochSeconds) {
...
 
-export function renderFormat(value, type, includeBytes = false) {
+/**
+ * Converts a list to a single string containing its values.
+ *
+ * @param {array} list
+ * @returns {string} joined list values
+ */
+export function renderList(list) {
+    if (!list?.length || !list[0]) {
```
percyliang, line 89: Why `!list[0]`?  In general, I'd like to be a bit more strict about the differences between falsy values...

leilenah: The `tags` field returns `['']` when there aren't any tags. Didn't want to change that in case doing so would break a bunch of stuff, but can change if you think I should.

codalab/bundles/run_bundle.py
leilenah: I generally agree, but things like this hurts my heart:

Screen Shot 2022-09-14 at 10 24 44 PM

leilenah: One option I've been vaguely thinking about is storing the definitions somewhere else. But I don't want to do that in the scope of this PR

frontend/src/util/worksheet_utils.js
percyliang: Oh, I see...maybe write that down as a comment - this is something we should fix.  No tags should be `[]`!

codalab/bundles/run_bundle.py
percyliang: I think putting the definitions here is nice so that everything is in one place.  I agree that long lines and weird wrapping is not-ideal either...  I only have a weak preference for the single lines, so feel free to wrap if you feel strongly about it.

percyliang (approved):

PR #4253, opened 2022-09-13
Prevent worksheet reloads during command execution.

### LINKED ISSUE #4176: DELETE Bundle UI Issues
- When deleting a bundle via UI, delete takes a long time (10+ sec) to process
- While delete is in process, a random bundle is removed temporarily from UI, and a copy of another bundle is created.

There is not a 100% repro of the above cases.

Recording of one repro (Restricted to team members): https://drive.google.com/file/d/1s2Z_UTaoJ8YdcxRDy501fv57hUJwRJZv/view?usp=sharing

leilenah:

Requesting worksheet reloads in the middle of worksheet command executions can result in unwanted UI side effects.

E.g. this unwanted side effect during bundle deletion: https://drive.google.com/file/d/1s2Z_UTaoJ8YdcxRDy501fv57hUJwRJZv/view?usp=sharing

This change prevents reloads during command execution. 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4176

### Screenshots

There are no changes from a UI perspective.

### Checklist


frontend/src/components/worksheets/Worksheet/Worksheet.js
```diff
@@ -287,7 +288,10 @@ class Worksheet extends React.Component {
         // Refreshes the checkbox after commands
         // If the action failed, the check will persist
         let force_delete = cmd === 'rm' && this.state.forceDelete ? '--force' : null;
-        this.setState({ updating: true });
+        this.setState({
+            updating: true,
```
leilenah, line 292: Note that the broad `updating` state is used outside of command execution, so I didn't want to rely on that state for this particular logic. The `executingCommand` state was introduced to specifically represent updates during command execution.

percyliang (approved):
Thanks!

PR #4260, opened 2022-09-27
Improve bundle view responsiveness.

### LINKED ISSUE #4216: Improve bundle view responsiveness
### Implementation
- [x] Prevent horizontal scroll in `CodeSnippet`
- [x] Reduce `CodeSnippet` font size (potentially)
- [x] Improve bundle view responsiveness 
- [x] QA

leilenah:

Unwanted horizontal scrolling can be observed when bundle content (like command, stdout, stderr, etc) has long lines (see sample screen recording below). This change prevents that unwanted scrolling. 

I have also added css that collapses the bundle view at the `600px` breakpoint, ordering the bundle sidebar above the bundle content. This prevents the bundle view from becoming squished.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4216

#### Before [prod]

https://user-images.githubusercontent.com/25855750/192422773-b58eec00-0a83-4bfe-9464-cd1b094b8184.mp4

#### After [local]

https://user-images.githubusercontent.com/25855750/192422788-13d3823f-b2c1-4d6c-a028-dff6ab90e09a.mp4

### Checklist

### DIFF
--- a/frontend/src/components/CodeSnippet.jsx
+++ b/frontend/src/components/CodeSnippet.jsx
@@ -22,7 +22,7 @@ class CodeSnippet extends React.Component {
         return (
             <Grid item xs={12}>
                 <div className={classes.snippet} style={{ maxHeight, marginBottom }}>
-                    <div>{code}</div>
+                    <div className={classes.codeContainer}>{code}</div>
                     {copyMessage && <Copy message={copyMessage} text={code} />}
                     {href && <NewWindowLink href={href} />}
                 </div>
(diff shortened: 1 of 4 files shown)


percyliang (commented):
I can imagine there are times when you don't want wrapping, because it might be confusing where lines start (e.g., for code).  I'm not sure - what do you think?

Can we look at the real worksheets from the gallery on prod (https://worksheets.codalab.org/worksheets/0x818930127c4d47de84c1ceaadf04d014/) and see how it feels?

leilenah:
@percyliang we basically have to choose between wrapping and the horizontal scroll. I was thinking that people who encounter wrapping can still expand their screen (to an extent of course) in order to un-wrap things. They're also able to open the output in a new window for more space.

As a user I think I would prefer wrapping to horizontal scroll, but I don't feel strongly. Lmk your preference 👍

percyliang:
I see the benefits of both.  Do you have any thoughts from the real worksheets?

leilenah:
@percyliang based on the worksheets on prod, I am leaning towards wrapping over having a horizontal scrollbar. I'm being tipped into that direction due to the fact that for the expanded bundle view, the horizontal scroll bar appears way at the bottom of the content. So the user may not instantly realize why some of their output is getting cut off. E.g.

image

With wrapping, it may seem more intuitive to expand your screen or open up the output in a new window.

percyliang:
Wrapping it is then!

percyliang (approved):

PR #4267, opened 2022-10-05
Modify error message output when input bundle prefix corresponds to more than one bundle.

### LINKED ISSUE #3747: Better error message for uuid collisions with `cl open`
image

AndrewJGaut:
Update error message displayed when user specifies a bundle prefix that belongs to more than one bundles

The previous error message was not descriptive or clear and may have caused confusion.

### Related issues

Fixes #3747


epicfaace (approved):

codalab/lib/canonicalize.py
```diff
@@ -78,7 +78,7 @@ def get_bundle_uuid(model, user, worksheet_uuid, bundle_spec):
         elif len(bundle_uuids) == 1:
             return bundle_uuids[0]
         else:
-            raise UsageError('uuid prefix %s more than one bundle' % bundle_spec)
+            raise UsageError('More than one bundle has uuid prefix %s' % bundle_spec)
```
epicfaace, line 81: Is this line covered by a test, btw? is there an easy way to add a test in test_cli.py that triggers this? If the setup is too complicated, let's not do it though.

epicfaace (approved):

codalab/lib/canonicalize.py
AndrewJGaut: I thought about it and I'm not sure how to do it in a particularly easy way. I tested it manually locally by repeatedly uploading bundles until two bundles had the same first letter in their UUID and then running a command with that first letter as the prefix. 
I could do this programmatically (upload 'a.txt' as a bundle in a while loop and save the UUID and break when you get two with the same first letter then run a command), but not sure if it's worth it?

PR #4268, opened 2022-10-05
Fix/3523 remove is superuser from user table

### LINKED ISSUE #3523: Remove is_superuser
is_superuser is a column in the User table that isn't used at all. We should delete this column from the code and remove it through a migration.

AndrewJGaut:

The is_superuser column in the User table was unused.

### Related issues

Fixes #3523


alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
```diff
@@ -0,0 +1,26 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('user', 'is_superuser')
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('is_superuser', mysql.TINYINT(display_width=1), autoincrement=False, nullable=False))
```
epicfaace: Should it just be `sa.Boolean()`? to match the migration where it was added -- https://github.com/codalab/codalab-worksheets/blob/master/alembic/versions/2016020401_augment_user_table_and_add__58eccccb346d.py

epicfaace (approved):

alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
AndrewJGaut: Done.

Issue #3787, opened 2021-09-15
Running "cl search .count" on codalab.stanford.edu times out


epicfaace:
Running "cl search .count" on codalab.stanford.edu times out. It should instead give us the correct result.

![image](https://user-images.githubusercontent.com/1689183/133358411-9c99188b-e517-4348-a83d-af03b93c8efa.png)

jzwang43:
@epicfaace I was looking at the query generated by `search_bundles`. Looks like it's already pretty optimized. Do you know if there's any way to improve the performance if there are a lot of bundles like on Stanford cluster?

```
SELECT DISTINCT count(*) AS count_1 
FROM bundle 
WHERE bundle.uuid IN (SELECT bundle_metadata.bundle_uuid 
FROM bundle_metadata 
WHERE bundle_metadata.metadata_key = :metadata_key_1 AND bundle_metadata.metadata_value >= :metadata_value_1)
 LIMIT -1 OFFSET :param_1
```

epicfaace:
Some optimizations we could do:
- we could potentially get rid of the outer query? What do `param_1` and `metadata_key_1` and `metadata_value_1` mean?
- add indexes for `metadata_key` and `metadata_value` in the `bundle_metadata` table, and an index for `uuid` in `bundle`

epicfaace:
maybe this will be faster

```
SELECT distinct bundle_uuid 
FROM bundle_metadata 
WHERE bundle_metadata.metadata_key = :metadata_key_1 AND bundle_metadata.metadata_value >= :metadata_value_1
```

pranavjain:
@jzwang43  Will start working on it next week (02/09)

pranavjain:
Removing this from p1.

This would require experimentation, moving to p2.

PR #4273, opened 2022-10-07
Fix/4196 speed up search bundles

### LINKED ISSUE #4196: `search_bundles` is extremely slow
Even uploading an empty file (via `cl upload`) takes 38 seconds.

-----

codalab.stanford.edu:

- Creating bundle (create an entry in DB): 0.19s
- Checking bundle location logic: 0.01s
- Sending data through (tiny file): 0.0s
- Reading the response: 21.11s
    - Upload to bundle store: 0.01s
     - [Update disk metadata](https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L1097): 21.034s
            - [_get_disk_used](https://github.com/codalab/codalab-worksheets/blob/master/codalab/model/bundle_model.py#L2709) causes the slowness
- **Total time: 24.453s**

`search_bundles` is slow: https://github.com/codalab/codalab-worksheets/blob/1b27806a419950695d6cb57d00e5bbd9068c83af/codalab/model/bundle_model.py#L354

Example query:

```
SELECT sum(q2.num) AS sum_1
FROM (SELECT DISTINCT bundle.uuid AS uuid, bundle_metadata.metadata_value * :metadata_value_1 AS num
FROM bundle, bundle_metadata
WHERE bundle.uuid = bundle_metadata.bundle_uuid AND bundle_metadata.metadata_key = :metadata_key_1 AND bundle.owner_id = :owner_id_1 AND bundle.data_hash LIKE :data_hash_1 AND (bundle.owner_id = :owner_id_2 OR bundle.uuid IN (SELECT group_bundle_permission.object_uuid
FROM group_bundle_permission
WHERE (group_bundle_permission.group_uuid = :group_uuid_1 OR group_bundle_permission.group_uuid IN (SELECT user_group.group_uuid
FROM user_group
WHERE user_group.user_id = :user_id_1)) AND group_bundle_permission.permission >= :permission_1))) AS q2
2022-08-24 22:24:46,454 query: SELECT DISTINCT bundle.uuid, bundle.id
FROM bundle ORDER BY bundle.id * :id_1 DESC
 LIMIT :param_1 OFFSET :param_2
```

Nested subqueries are really slow.

AndrewJGaut:

The search_bundles function was slow primarily because it relied on nested subqueries. We updated the function to use inner joins instead, which should be much faster.

### Related issues

Fixes #4196. Also addresses #3787 (see example in comment below for that function).


percyliang:
Do you have any before/after timings that you can share?

AndrewJGaut:
Sure thing!

Examples of speedup (run on prod):

Command: cl search .shared data_size=.sum

Original query:
SELECT sum(q1.num) AS sum_1 FROM (SELECT DISTINCT bundle.uuid AS uuid, bundle_metadata_1.metadata_value * 1 AS num FROM bundle_metadata AS bundle_metadata_1, bundle INNER JOIN group_bundle_permission ON bundle.uuid = group_bundle_permission.object_uuid INNER JOIN user_group ON group_bundle_permission.group_uuid = user_group.group_uuid WHERE user_group.user_id='0' AND group_bundle_permission.permission >= 1 AND bundle.uuid = bundle_metadata_1.bundle_uuid AND bundle_metadata_1.metadata_key='data_size') AS q1;
**7.52 seconds**

Updated query:
SELECT sum(q1.num) AS sum_1 FROM (SELECT DISTINCT bundle.uuid AS uuid, bundle_metadata_1.metadata_value * 1 AS num FROM bundle_metadata AS bundle_metadata_1, bundle INNER JOIN group_bundle_permission ON bundle.uuid = group_bundle_permission.object_uuid INNER JOIN user_group ON group_bundle_permission.group_uuid = user_group.group_uuid WHERE user_group.user_id=’0’ AND group_bundle_permission.permission >= 1 AND bundle.uuid = bundle_metadata_1.bundle_uuid AND bundle_metadata_1.metadata_key='data_size') AS q1;
**0.0 seconds**


Command: cl search .after=2021-08-20 .count

Original query:
SELECT count(q2.uuid) AS tbl_row_count FROM (SELECT DISTINCT bundle.uuid AS uuid FROM bundle WHERE bundle.uuid IN (SELECT bundle_metadata.bundle_uuid FROM bundle_metadata WHERE bundle_metadata.metadata_key = ‘created’ AND bundle_metadata.metadata_value >= 1629417600) LIMIT 0, 18446744073709551615) AS q2;
**27.74 seconds.**

Updated query:
SELECT count(q1.uuid) AS tbl_row_count FROM (SELECT DISTINCT bundle.uuid AS uuid FROM bundle INNER JOIN bundle_metadata AS bundle_metadata_1 ON bundle.uuid = bundle_metadata_1.bundle_uuid WHERE bundle_metadata_1.metadata_key = ‘created’ AND bundle_metadata_1.metadata_value >= 1629417600 LIMIT 0, 18446744073709551615) AS q1
**1.95 sec**

docker_config/compose_files/.docker/canary.json
```diff
@@ -0,0 +1,9 @@
+{
```
wwwjn: Do we need docker_config files? Or should we remove them?

AndrewJGaut: Just removed them. Sorry about that!

alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
```diff
@@ -0,0 +1,26 @@
+"""Remove is_superuser from User table
```
percyliang: This is not part of speeding up search bundles, right?  Ideally, would split the superuser thing into a different PR.

codalab/model/bundle_model.py
```diff
@@ -420,8 +420,16 @@ def make_condition(key, field, value):
                 return field == value
             return None
 
+        def add_join(join_table, join_condition):
+            join_tables.append(join_table)
+            join_conditions.append(join_condition)
+
         shortcuts = {'type': 'bundle_type', 'size': 'data_size', 'worksheet': 'host_worksheet'}
 
+        join_tables = []
```
percyliang: Could you add type annotations?

codalab/model/bundle_model.py
```diff
@@ -465,71 +473,67 @@ def make_condition(key, field, value):
...
-                                alias(
-                                    select([cl_user_group.c.group_uuid]).where(
-                                        cl_user_group.c.user_id == user_id
-                                    )
-                                )
-                            ),
-                            cl_group_bundle_permission.c.permission >= GROUP_OBJECT_PERMISSION_READ,
-                        )
-                    )
+                add_join(
```
percyliang, line 503: I remember there were some subtleties with joins when the joined tables (e.g., permissions) were empty, whereas for subqueries, it was much more straightforward to reason about, so would be good to double check the edge cases.

codalab/model/bundle_model.py
```diff
@@ -539,113 +543,114 @@ def make_condition(key, field, value):
...
-                                )
-                            )
-                        )
+                aliased_bundle_metadata = aliased(cl_bundle_metadata)
+                add_join(
+                    aliased_bundle_metadata,
+                    cl_bundle.c.uuid == aliased_bundle_metadata.c.bundle_uuid,
+                )
+
+                where_clause = []
```
percyliang: To be more accurate, I'd call this `where_clause_disjuncts`

codalab/model/bundle_model.py
```diff
@@ -539,113 +543,114 @@ def make_condition(key, field, value):
...
-                        )
-                    )
+                aliased_bundle_metadata = aliased(cl_bundle_metadata)
+                add_join(
+                    aliased_bundle_metadata,
+                    cl_bundle.c.uuid == aliased_bundle_metadata.c.bundle_uuid,
                 )
-                clause = or_(*clause)
+
+                where_clause = []
```
percyliang: seems like you can just write `or_(cl_bundle..., ..., ...)` directly instead of creating an array?

codalab/model/bundle_model.py
```diff
@@ -539,113 +543,114 @@ def make_condition(key, field, value):
...
 
-            if clause is not None:
-                clauses.append(clause)
+                    where_clause = and_(aliased_bundle_metadata.c.metadata_key == key, condition)
 
-        clause = and_(*clauses)
+            if where_clause is not None:
+                where_clauses.append(where_clause)
+
+        where_clause = and_(*where_clauses)
```
percyliang: Might be confusing to reuse `where_clause` for both the top-level clause and the conjuncts.  One might consider creating helper function(s) to break up this big function.

codalab/model/bundle_model.py
```diff
@@ -539,113 +543,114 @@ def make_condition(key, field, value):
...
-                        ),
-                        cl_group_bundle_permission.c.permission
-                        >= GROUP_OBJECT_PERMISSION_READ,  # Match the uuid of the parent
-                    )
-                )
+            access_via_group = and_(
+                or_(  # Join constraint (group)
+                    cl_group_bundle_permission.c.group_uuid
+                    == self.public_group_uuid,  # Public group
+                    cl_group_bundle_permission.c.group_uuid.in_(
```
percyliang: We can't replace this subquery with a join?

codalab/model/bundle_model.py
```diff
@@ -539,113 +543,114 @@ def make_condition(key, field, value):
...
+                ),
+                cl_group_bundle_permission.c.permission
+                >= GROUP_OBJECT_PERMISSION_READ,  # Match the uuid of the parent
             )
-            clause = and_(clause, or_(access_via_owner, access_via_group))
 
+            where_clause = and_(where_clause, or_(access_via_owner, access_via_group))
+
+        table = cl_bundle
+        for join_table, join_condition in zip(join_tables, join_conditions):
```
percyliang: Instead of using parallel arrays, which might be error prone, consider making a new `dataclass` with two fields (the table and the join).

percyliang (commented):
At a high-level, logic looks good, and it's great that we get such dramatic speedups!  I'd just be careful about the edge cases, and we should probably add some test cases for this.  Consider also splitting up the function into more modular pieces (something we've wanted to do for a while) if you think it makes sense.

alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
AndrewJGaut: Shoot. It is in a [different PR](https://github.com/codalab/codalab-worksheets/pull/4268) that I submitted Thursday (still hasn't been merged b/c needs review). I must have branched off from my local branch I used for that PR when I created the branch for this PR. Sorry about that! Perhaps I should revert those changes in this branch?

codalab/model/bundle_model.py
AndrewJGaut: Added for the two arrays I'm using (joins and where_clause_conjuncts). 

When I try to add type annotations for some of the other variables (like limit), I get a strange error from mypy:
Deferral trace:
    codalab.model.bundle_model:409
    codalab.model.bundle_model:412
    codalab.model.bundle_model:414
    codalab.model.bundle_model:417
    codalab.model.bundle_model:419
    codalab.model.bundle_model:433
    codalab.model.bundle_model:438
    codalab.model.bundle_model:438
    codalab.model.bundle_model:438
codalab/model/bundle_model.py: error: INTERNAL ERROR: maximum semantic analysis iteration count reached

So, I've left those un-typed for now, but I'll look into this.

codalab/model/bundle_model.py
AndrewJGaut: Yes, good point. So, the queries I converted to inner joins above are all OK in this regard; I verified each of them individually. Basically, the subqueries (aside from the one you highlighted in a comment above and one other which I didn’t convert) are used as conjuncts in the final WHERE clause, aren’t negated, and require the existence of something in the table in the subquery to be true. In other words, each of those subqueries is automatically false if the table is empty anyway.

This is _not_ true for the two other queries I didn't convert earlier (the one used for .floating and one used when the querying user is not root). For those, an inner join would not work for the reason you mentioned. For instance, for the negated query, it checks that the bundle_uuid is NOT in the Worksheet_Item table, and so using an INNER JOIN would be incorrect since the inner join would be empty even though the result we _really_ want is for every bundle_uuid to be in the result since none belong to a worksheet (since worksheet_item is empty). For this query (and for the other I mentioned), we remedy this by using a LEFT OUTER JOIN instead of an INNER JOIN.

I verified that my changes work by adding test cases. I added some for .floating, and I added some cases for cl search queries used by the non-root user. I verified these work and that, as expected, they all fail when I switch the LEFT OUTER JOIN to an INNER JOIN.

Let me know if I didn't address your comment sufficiently and/or if I missed something. Thanks!

alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
percyliang: Maybe you can get that PR in first, since it's pretty straightforward, and then you don't have to undo anything.

codalab/model/bundle_model.py
AndrewJGaut: Thanks! (I think they're actually conjuncts in this case since we use and_ to connect them in line 623, but please correct me if I'm wrong!)

codalab/model/bundle_model.py
AndrewJGaut: Done.

codalab/model/bundle_model.py
AndrewJGaut: Done.

codalab/model/bundle_model.py
AndrewJGaut: Done.

AndrewJGaut:
Thanks so much for the thorough review! I added more test cases to address the edge cases you mentioned (see my comment about INNER JOIN and LEFT OUTER JOIN above).

In terms of splitting the function, I do think that makes sense and gave it some thought throughout the day. However, I'm still not sure how to do it. I'll keep thinking, but in the meantime I addressed your other comments. Thanks!

codalab/model/bundle_model.py
AndrewJGaut: Done. But see the comment above about INNER JOIN vs LEFT OUTER JOIN. I left this and the other subquery in the .floating case because they were harder to convert and subsequently forgot to convert them. Sorry about that!

alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
AndrewJGaut: OK, I'll ping the chat for a review. Thanks!

epicfaace:
> **0.0 seconds**

Amazing!!

alembic/versions/2022100403_remove_is_superuser_from_user__22b6c41cd29c.py
epicfaace: I think it would be easiest to revert those changes in this branch, so it doesn't have to depend on that PR. Up to you though.

codalab/model/bundle_model.py
```diff
@@ -378,18 +379,19 @@ def search_bundles(self, user_id, keywords):
...
-        count = False
-        sort_key = [None]
-        sum_key = [None]
-        aux_fields = []  # Fields (e.g., sorting) that we need to include in the query
-
-        # Number nested subqueries
         subquery_index = [0]
 
+        @dataclass
+        class JoinTable:
```
epicfaace: Can we move this class declaration outside the `search_bundles` function? Eventually (in a later PR probably!) we'll need to do the same thing as this PR for the search_worksheets and will end up using `JoinTable` for that, too (https://github.com/codalab/codalab-worksheets/issues/4277).

codalab/model/bundle_model.py
```diff
@@ -378,18 +379,19 @@ def search_bundles(self, user_id, keywords):
...
+            """
+            Class that stores data for joins used in final SQL query.
+            By default, becomes an inner join. Becomes left outer join if left_outer_join is True.
+            """
+
+            table: Table
+            condition: Any
+            left_outer_join: bool = False
+
```
epicfaace: Great work! However, I think it's a bit difficult to understand how the code is working by just reading through it. Can you add a few sentences at the beginning of the method that describes the general approach this code is taking (and the role played by where_class_conjuncts, add_join, etc.)?

codalab/model/bundle_model.py
epicfaace: On that point, it might be worth thinking about whether it makes sense to abstract this code out into something more general, so that it could be implemented in `search_worksheets` (without having to copy and paste everything again). But not necessary for this PR, but if you do think of a good way to do it that works well in this PR, do add it in!

tests/cli/test_cli.py
```diff
@@ -1613,6 +1613,32 @@ def test_search(ctx):
...
+    check_equals('', _run_command([cl, 'search', '.floating', '-u']))
+    uuid3 = _run_command([cl, 'upload', test_path('a.txt')])
+    _run_command([cl, 'detach', uuid3], 0)
+    check_equals(uuid3, _run_command([cl, 'search', '.floating', '-u']))
+    _run_command([cl, 'rm', uuid3])  # need to remove since not on main worksheet
+    # Check search when groups empty
+    check_equals('', _run_command([cl, 'search', '.shared']))
+    # Check search with non-root user.
+    if not os.getenv('CODALAB_PROTECTED_MODE'):
+        # This test does not work when protected_mode is True.
```
epicfaace, line 1626: Why not?

AndrewJGaut: Running cl search with non-root user yields a 401 Unauthorized when in protected mode. I could solve this also by using 'cl uedit <non-root-username> --grant-access'  with the root user before switching to the non-root user instead, though. Perhaps that's better.

codalab/model/bundle_model.py
AndrewJGaut: Done. I have a potential idea for abstracting the logic, but I think it might be better to do it in a separate PR. The issue is that the there aren't many shared keys between search_worksheets and search_bundles (aside from the "special" keywords at the top of the function, which can be abstracted quite easily) and even some shared keys are dealt with a little differently (e.g. the key ''), which makes it hard to abstract.

codalab/model/bundle_model.py
AndrewJGaut: Done.

percyliang (approved):
Looks great!

Issue #4271, opened 2022-10-05
check the time quota while jobs are running


percyliang:
Prevent people from launching a ton of jobs in parallel and using more than their allotted time quota.

PR #4283, opened 2022-10-15
adding changes to check time quota for running jobs.

AndrewJGaut:

Currently, the time quota tracking for users is only updated when a job finishes and a user's remaining time quota is not checked while jobs are running. With this change, we update the time quota while the user's job runs and we kill jobs if the time quota is surpassed while the job is running.

This will help prevent people from abusing the free computation provided by Codalab (e.g. for cypro mining).

### Related issues

#4271


docker_config/compose_files/docker-compose.dev.yml
```diff
@@ -11,6 +11,8 @@ services:
     stdin_open: true
     tty: true
   rest-server:
+    stdin_open: true
```
AndrewJGaut, line 14: Note: I added these entries in the dev Docker compose so I could use pdb to debug code running in the containers locally. I thought others might want to use pdb in the future as well, so I've left this in. However, I can take it out if that's preferable!

tests/cli/test_cli.py
```diff
@@ -1715,6 +1721,12 @@ def test_search_time(ctx):
 
 @TestModule.register('run')
 def test_run(ctx):
+    # Test thaat bundle fails when run without sufficient time quota
```
```suggestion
    # Test that bundle fails when run without sufficient time quota
```

tests/cli/test_cli.py
```diff
@@ -1715,6 +1721,12 @@ def test_search_time(ctx):
...
+    _run_command([cl, 'uedit', 'codalab', '--time-quota', '2'])
+    uuid = _run_command([cl, 'run', 'sleep 100000'])
+    wait_until_state(uuid, State.KILLED)
```
epicfaace: Does `wait_until_state` take a timeout argument? if so, we might want to pass in a timeout of maybe ~1min

tests/cli/test_cli.py
```diff
@@ -393,6 +394,9 @@ def __enter__(self):
         self.worksheets.append(temp_worksheet)
         _run_command([cl, 'work', temp_worksheet])
         self.disk_quota = _run_command([cl, 'uinfo', '-f', 'disk']).split(' ')[2]
+        self.time_quota = (
+            _run_command([cl, 'uinfo', '-f', 'time']).split(' ')[2].split('y')[0] + 'y'
```
epicfaace, line 398: Why do we remove "y" and then add "y" to the end again?

docker_config/compose_files/docker-compose.dev.yml
epicfaace: Feel free to add it in, thanks. Maybe add a brief comment that this is meant for pdb (so we remember to add it to additional containers in the future).

codalab/server/bundle_manager.py
```diff
@@ -434,7 +434,6 @@ def _schedule_run_bundles_on_workers(self, workers, staged_bundles_to_run, user_
             )
             for queue_position, bundle in zip(queue_positions, sorted_user_staged_bundles):
                 staged_bundles_to_run[queue_position] = bundle
-
```
epicfaace: undo this change

codalab/rest/workers.py
```diff
@@ -42,16 +45,24 @@ def checkin(worker_id):
...
+                messages.append({'type': 'kill', 'uuid': bundle.uuid})
+
+        except Exception as e:
+            logger.info("Exception in REST checkin: {}".format(e))
 
     with closing(local.worker_model.start_listening(socket_id)) as sock:
-        return local.worker_model.get_json_message(sock, WAIT_TIME_SECS)
+        messages.append(local.worker_model.get_json_message(sock, WAIT_TIME_SECS))
+    response.content_type = 'application/json'
+    return json.dumps(messages)
```
epicfaace, line 68: Can you explain your reasoning for why we're changing messages to be a list (and thus have to add more logic to worker.py to handle it)?

AndrewJGaut: Yes. With how the worker code was written previously, it accepted just a single bundle action in the checkin response. In this case, we may have to kill several bundles (e.g. imagine an attacker starts up 200 bundles that run forever and goes over their time quota; ideally we could kill all of these in one checkin) and other messages aside from the kill messages may be present in that worker's socket.

For that reason, I thought adding the ability to have the worker do several bundle actions in one checkin response would be convenient. However, if you think there's a better way, I can definitely change it!

tests/cli/test_cli.py
AndrewJGaut: This is because of how the [parse duration](https://github.com/codalab/codalab-worksheets/blob/master/codalab/lib/formatting.py#L130) is written. Basically, it expects a _single_ string of the form "<number>[<s|m|h|d|y>]" and the default time quota for the root user is '100y0d'. Thus, I'm just removing the '0d' from the string. In general, this will just keep the number of years in the time quota.

It might be better to re-write parse_duration to take strings of the form "<number>[<s|m|h|d|y>]" repeated any number of times, which is easy enough, but I wasn't sure if that would be preferable?

codalab/server/bundle_manager.py
AndrewJGaut: Done

docker_config/compose_files/docker-compose.dev.yml
AndrewJGaut: Done.

tests/cli/test_cli.py
AndrewJGaut: Done.

tests/cli/test_cli.py
AndrewJGaut: Done.

wwwjn:
Update: add an error message to indicate the reason why the bundle is killed. Eg, "Time quota exceed. Apply for more time quota here: [some link]"

codalab/rest/workers.py
```diff
@@ -42,16 +45,27 @@ def checkin(worker_id):
...
             bundle = local.model.get_bundle(worker_run.uuid)
             local.model.bundle_checkin(bundle, worker_run, request.user.user_id, worker_id)
-        except Exception:
-            pass
+
+            if local.model.get_user_time_quota_left(bundle.owner_id) <= 0:
+                # Then, user has gone over their time quota and we kill the job.
+                kill_message = (
+                    'Kill requested: User time quota exceeded. To apply for more quota, please visit the following link: '
+                    'https://codalab-worksheets.readthedocs.io/en/latest/FAQ/#how-do-i-request-more-disk-quota-or-time-quota'
```
wwwjn, line 59: Thanks for adding this! LGTM!
A note for Jiani: this PR needs to be deployed at the same time as #4296

wwwjn (approved):

Issue #4263, opened 2022-10-01
Add optional arg `--account` for `SlurmBatchWorkerManager`


teetone:
Add arg and handle here: https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker_manager/slurm_batch_worker_manager.py#L413
 
From Action: 

```
- When submitting jobs on sc, you must now include the "--account" flag as followed (or inside your SBATCH script), if the person has multiple account (group) affiliations
srun --account=vision -p tibet --gres=gpu:1 --pty bash
 - To check your account affiliations, please simply issue "showaccount"
 ```

Issue #4261, opened 2022-10-01
get slurm worker manager working


percyliang:
sc was just upgraded and requires the `--account` flag.

percyliang:
In general, make sure that you can just run `nlp-cl-worker-manager nlp` and have run jobs and have them finish in a timely manner.

PR #4288, opened 2022-10-26
Fix/4263 update slurm

AndrewJGaut:

The Slurm instance used by Stanford got updated and people with multiple account affiliations need to specify an account with which to run a Slurm batch job. This PR fixes that issue so that these users can still launch a worker manager with nlp-cl-worker-manager.

### Related issues

[#4261](https://github.com/codalab/codalab-worksheets/issues/4261), #[4263](https://github.com/codalab/codalab-worksheets/issues/4263)


codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -100,6 +100,20 @@ def add_arguments_to_subparser(subparser):
...
         )
+        subparser.add_argument(
+            '--account',
+            type=str,
+            default='',
+            help='The account you want the Slurm batch job to be associated with.'
+            ' Required if submitter has multiple account (group) affiliations.',
+        )
+        subparser.add_argument(
+            '--account_file',
```
percyliang: Maybe we should generalize this to a config file (which could contain the password too)?

codalab/worker_manager/slurm_batch_worker_manager.py
```diff
@@ -428,6 +442,15 @@ def create_slurm_args(self, worker_id, slurm_worker_dir):
...
         if self.args.constraint:
             slurm_args['constraint'] = self.args.constraint
+        if self.args.account_file:
+            with open(self.args.account_file) as f:
+                slurm_args['account'] = f.readline().strip()
+        if self.args.account:
+            # override account file if both specified.
+            slurm_args['account'] = self.args.account
+        if 'account' not in slurm_args:
+            accounts = subprocess.check_output(["showaccount"]).decode('utf-8').split()[8::3]
```
percyliang: Can you document the format (8::3 is pretty cryptic)?

teetone (approved):
Looks good! Verified it works:

cl-workers

Issue #4286, opened 2022-10-24
Upload blobs to Azure storage in parallel


wwwjn:
Upload blobs to azure storage has 2 steps:
1. write/put blob
2. commit blob

We can do first step (write/put) in parallel to reduce the total upload time.

wwwjn:
Test result:
**TLDR: After using parallel, we can reduce the waiting time when some block is being committed.** 

In PR #4275, we minimize the size of MIN_WRITE_SIZE and MAX_WRITE_SIZE. This means when uploading a file, we need more put() operations and more commit() operations, which slow down the uploading process. So we need to parallel uploading process to speed up.

Measure the actual upload time of a file content:
1. without parallel: 17625s

![Image](https://user-images.githubusercontent.com/40016222/198908468-349c86c2-27e5-4036-9682-f6e8e384f2a7.png)



2. with parallel:  2690s

![Image](https://user-images.githubusercontent.com/40016222/198908457-99d799fd-e947-4d85-a6fc-11490e59d920.png)

PR #4291, opened 2022-10-26
Change to parallel uploading for Azure blob storage

wwwjn:
### Related issues

#4286 

### Checklist

### DIFF
--- a/codalab/lib/beam/blobstorageuploader.py
+++ b/codalab/lib/beam/blobstorageuploader.py
@@ -6,7 +6,7 @@ from azure.storage.blob import (
 from apache_beam.io.azure.blobstorageio import parse_azfs_path
 import base64
 from codalab.worker.un_gzip_stream import BytesBuffer
-from concurrent.futures import ThreadPoolExecutor, wait, ALL_COMPLETED
+from concurrent.futures import ThreadPoolExecutor
 
 class BlobStorageUploader(Uploader):
   """An improved version of apache_beam.io.azure.blobstorageio.BlobStorageUploader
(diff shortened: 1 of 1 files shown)


wwwjn:
Test result:
**TLDR: After using parallel, we can reduce the waiting time when some block is being committed.** 

In PR #4275, we minimize the size of MIN_WRITE_SIZE and MAX_WRITE_SIZE. This means when uploading a file, we need more put() operations and more commit() operations, which slow down the uploading process. So we need to parallel uploading process to speed up.

Measure the actual upload time of a file content:
1. without parallel: 17625s

Image



2. with parallel:  2690s

Image

codalab/lib/beam/blobstorageuploader.py
```diff
@@ -48,12 +51,13 @@ def put(self, data):
   def _write_to_blob(self, data):
     # block_id's have to be base-64 strings normalized to have the same length.
     block_id = base64.b64encode('{0:-32d}'.format(self.block_number).encode()).decode()
-    
-    self._blob_to_upload.stage_block(block_id, data)
+    # put the blob content to server, blob is uncommitted
```
epicfaace: Can you change this comment to mention that it's going to run in parallel?

codalab/lib/beam/blobstorageuploader.py
```diff
@@ -48,12 +51,13 @@ def put(self, data):
...
+    self.all_tasks.append(self.thread_pool.submit(self._blob_to_upload.stage_block, block_id, data))
     self.block_list.append(BlobBlock(block_id))
     self.block_number = self.block_number + 1
 
   def finish(self):
     # The buffer will have a size smaller than MIN_WRITE_SIZE, so its contents can fit into memory.
     self._write_to_blob(self.buffer.read())
+    wait(self.all_tasks, return_when=ALL_COMPLETED)
```
epicfaace: You should call `self.executor.shutdown()` here -- see https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

epicfaace (approved):

PR #4297, opened 2022-10-30
Add better error message for incorrect worksheet name.

### LINKED ISSUE #4188: [CLI] better error message for incorrect password
When user types in the incorrect password, it should say so rather than saying invalid spec:
```
cl work nlp::
Requesting access at https://codalab.stanford.edu
Username: pliang
Password: 
UsageError: Invalid spec: "nlp::"
```

leilenah:

This change allows the correct error to surface for users who type in an incorrect password.

Currently the "incorrect password" error is being masked by the CLI.

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4188

### Screenshots

https://user-images.githubusercontent.com/25855750/199907174-4d83efba-9755-4745-863c-0261a9a1a425.mp4

### Checklist

### DIFF
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -799,7 +799,11 @@ class BundleCLI(object):
                     client, base_worksheet_uuid, parsed_spec
                 )
             except ValueError:
-                raise UsageError('Invalid spec: "{}"'.format(spec))
+                raise UsageError(
+                    "'{}' worksheet could not be found. Use 'cl wls' command to search for worksheets.".format(
+                        spec
+                    )
+                )
         return client, worksheet_uuid
 
     @staticmethod
--- a/codalab/lib/bundle_cli.py
+++ b/codalab/lib/bundle_cli.py
@@ -800,7 +800,7 @@ class BundleCLI(object):
                 )
             except ValueError:
                 raise UsageError(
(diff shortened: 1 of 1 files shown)


leilenah:
@percyliang the [original request](https://github.com/codalab/codalab-worksheets/issues/4188) was to change the message that appears if a user types in an incorrect password. However, the incorrect password error handling seems to be functioning as expected (see below).

I think that the `nlp::` worksheet that you were requesting wasn't found, which was not clear in the error message that was surfaced to you as you were logging in. So I updated the error message that is shown when a worksheet is not found.

Lmk if I'm missing something.

image

percyliang:
`nlp::` is not the worksheet but the alias, which does exist in my case.  Run `cl alias nlp https://codalab.stanford.edu` first.

leilenah:
@percyliang this is what I'm seeing on my end while using my local `cl` (note that I typed in an incorrect password):

https://user-images.githubusercontent.com/25855750/199294890-a7a0322b-af86-4102-9f61-d3347ded3aa9.mp4

leilenah:
@percyliang can you re-review this please? This change will allow a `PermissionError` to be raised if the user has typed in an incorrect password.

Also I confirmed that if a worksheet is not found, an informative `NotFoundError` will still be surfaced to the user.

percyliang (approved):

PR #4299, opened 2022-11-01
Automatically reload REST server when file changes are detected

### LINKED ISSUE #1182: Live reload on backend
[This page](https://github.com/codalab/codalab-worksheets/wiki/Server-Setup) says:
"If you're making back end changes, the same functionality is not yet supported and you need to rebuild the relevant Docker image after each change."

@bkgoksel is this still the case? If so, what changes would we need in order to allow for live reload?

### LINKED ISSUE #4199: Improve the experience of having to restart backend docker containers during development
Having to restart the backend docker containers during dev can be cumbersome. Is there a workaround for this? If not, is there room to make the builds faster?

yifanmai:

The REST server will automatically reload when file changes are detected when the service is started in development mode i.e. `./codalab_service.py start -bd`. This eliminates the need to restart the REST server manually.

### Related issues

Fixes #1182
Fixes #4199

### Checklist

### DIFF
--- a/codalab/bin/server.py
+++ b/codalab/bin/server.py
@@ -35,7 +35,6 @@ class ClFileWatcherEventHandler(FileSystemEventHandler):
 
 def run_server_with_watch():
     modified_argv = list(sys.argv)
-    modified_argv[0] = 'cl'
     modified_argv.remove('--watch')
     event_handler = ClFileWatcherEventHandler(modified_argv)
     # Listen to root dir (/codalab/bin/../../)
--- a/docker_config/compose_files/docker-compose.dev.yml
+++ b/docker_config/compose_files/docker-compose.dev.yml
@@ -11,6 +11,7 @@ services:
     stdin_open: true
     tty: true
   rest-server:
+    command: cl-server --watch
     stdin_open: true
     tty: true
     volumes:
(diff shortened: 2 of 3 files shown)


percyliang (approved):
I remember that the reason we didn't activate this was that for the bundle manager, you would still have to manually restart.  But I guess we can try to do autorestart for that too?

Are there any downsides of doing this?

yifanmai:
Yes, the bundle manager still needs to be manually restarted; the updated docs does mention this.

The main disadvantage that when there's an error on worker startup (i.e. a Python exception when importing the modules, caused by a bug in the code), gunicorn shuts down entirely. gunicorn restarts automatically when the bug is fixed. This can be confusing for the developer because the website will appear to stop working, and they will have to look at Docker logs to find the exception. Other frameworks (i.e. Django, Rails, Play) keep the server running and return the exception in the browser instead. This seems to be an inherent limitation of gunicorn.

The other options I tried were:

- Setting `reloader=True` in [`bottle.run`](https://bottlepy.org/docs/dev/api.html#bottle.run) This reloads the gunicorn worker, but the code changes don't show for some reason.
- Setting `--reload` on the [gunicorn command line](https://docs.gunicorn.org/en/latest/settings.html?highlight=reload#reload) - The code changes show up, but an error on worker startup causes gunicorn to shutdown permanently, which means the developer has to manually restart the server; in contrast, `--watch` restarts the server automatically after the error is fixed.

The other disadvantage is that restarting the REST server causes it to lose state (e.g. in-memory global variables, persistent connections), but it is better for the REST server to be stateless anyway.

Issue #4302, opened 2022-11-06
Guard against miners: allow edu users have more default time/disk quota


wwwjn:
For normal users, they have 1h default time quota and 100m default disk quota.

For user with edu emails, they have 10g disk quota and 60d time quota.

leilenah:
PR ready, not merged

wwwjn:
Also see new changes in #4440

PR #4303, opened 2022-11-06
Allow edu users to have more time quota and disk quota

wwwjn:
To solve issue #4302.
Add two new environment variables to set edu email user's quota.  CODALAB_EDU_USER_DISK_QUOTA, CODALAB_EDU_USER_TIME_QUOTA.

```
  CODALAB_EDU_USER_DISK_QUOTA: 60g
  CODALAB_EDU_USER_TIME_QUOTA: 10d
```

### Related issues
#4302 

### Checklist

### DIFF
--- a/codalab/lib/codalab_manager.py
+++ b/codalab/lib/codalab_manager.py
@@ -326,11 +326,13 @@ class CodaLabManager(object):
     @cached
     def default_user_info(self):
         info = self.config['server'].get(
-            'default_user_info', {'time_quota': '1y', 'disk_quota': '1t', 'parallel_run_quota': 3}
+            'default_user_info', {'time_quota': '1y', 'disk_quota': '1t', 'edu_time_quota': '1y', 'edu_disk_quota': '1t', 'parallel_run_quota': 3}
         )
         return {
             'time_quota': formatting.parse_duration(info['time_quota']),
             'disk_quota': formatting.parse_size(info['disk_quota']),
+            'edu_time_quota': formatting.parse_duration(info['edu_time_quota']),
+            'edu_disk_quota': formatting.parse_size(info['edu_disk_quota']),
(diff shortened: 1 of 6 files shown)


codalab/lib/codalab_manager.py
```diff
@@ -326,11 +326,20 @@ def session(self):
     @cached
     def default_user_info(self):
         info = self.config['server'].get(
-            'default_user_info', {'time_quota': '1y', 'disk_quota': '1t', 'parallel_run_quota': 3}
+            'default_user_info',
+            {
+                'time_quota': '1y',
```
AndrewJGaut, line 338: Shouldn't time_quota and disk_quota be lowered (so that users without edu emails have less quota)?

codalab_service.py
```diff
@@ -249,6 +249,16 @@ def has_callable_default(self):
     # User
     CodalabArg(name='user_disk_quota', help='How much space a user can use', default='100g'),
```
AndrewJGaut, line 253: Same question as above: shouldn't this default be changed so that the time and disk quota for non-edu users is less?

AndrewJGaut (changes requested):

codalab/lib/codalab_manager.py
wwwjn: Yeah it should. But actually if we set environmental variables, these parameters here will not be used.

codalab_service.py
wwwjn: For our deployment, the actual disk / time quota is controlled by environmental variables `CODALAB_USER_DISK_QUOTA` and `CODALAB_USER_TIME_QUOTA`. But for other people who want to deploy their own Codalab instances, they might want edu users and normal users have same quota.

AndrewJGaut (approved):

PR #4304, opened 2022-11-06
Add error handle code for stale file handler when `cl upload`

wwwjn:
When uploading from nlp machine, we might meet **stale file handler**(https://www.baeldung.com/linux/stale-file-handles) generated by other process. Add some code to just ignore those stale file handlers.

For stale file handlers, we can not even `ls` them. 
image

### Checklist

### DIFF
--- a/codalab/worker/file_util.py
+++ b/codalab/worker/file_util.py
@@ -531,18 +531,17 @@ def get_path_size(path, exclude_names=[], ignore_nonexistent_path=False):
             return 0
         # Raise the FileNotFoundError
         raise
-    
+
     # Detect and ignore Stale file handler. Please see: https://www.baeldung.com/linux/stale-file-handles
     try:
         os.listdir(path)
     except OSError:
         if ignore_nonexistent_path:
-            #If we have trouble list the dir, return the size of this path as 0
+            # If we have trouble list the dir, return the size of this path as 0
             # Do not raise error and just ignore the stale file handler/
             return 0
         else:
             raise
-        
 
     if not os.path.islink(path) and os.path.isdir(path):
         for child in os.listdir(path):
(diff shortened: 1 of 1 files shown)


wwwjn:
Since this part of code is running by local CLI, and I can not create a stale file handler using local PC. Will keep an eye on this and test it out when CL version is updated.

codalab/worker/file_util.py
```diff
@@ -531,15 +531,26 @@ def get_path_size(path, exclude_names=[], ignore_nonexistent_path=False):
...
-                    full_child_path = os.path.join(path.decode('utf-8'), child.decode('utf-8'))
-                result += get_path_size(full_child_path, ignore_nonexistent_path=True)
-    return result
+        try:
+            child_paths = os.listdir(path)
+        except OSError:
+            if ignore_nonexistent_path:
+                # If we have trouble list the dir, return the size of this path as 0
+                # Do not raise error and just ignore the stale file handler/
+                return 0
```
percyliang, line 546: Could this potentially suppress other errors?  Maybe worth printing out a warning at least (assuming it doesn't result in too much output)?

wwwjn: Thank you! I will add a Warning print with the path, and test it out using the dev environment before getting it merged.

codalab/worker/file_util.py
```diff
@@ -540,7 +540,9 @@ def get_path_size(path, exclude_names=[], ignore_nonexistent_path=False):
             if ignore_nonexistent_path:
                 # If we have trouble list the dir, return the size of this path as 0
                 # Do not raise error and just ignore the stale file handler/
-                logging.warning(f"Error when list the child path. Ignore the files under path: {path}")
+                logging.warning(
+                    f"Error when list the child path. Ignore the files under path: {path}"
```
percyliang, line 544: "Error when listing files under {path}, ignoring..."

percyliang (approved):

Issue #4298, opened 2022-10-31
check the disk quota while jobs are running


AndrewJGaut:
As a counterpart to issue #4271, we also want to check user disk quota while jobs are running rather than checking only after the job has finished

PR #4307, opened 2022-11-09
Added changes to track disk quota usage

AndrewJGaut:

We want to ensure that disk quota usage can be tracked for running bundles so that attackers can't zip bomb the system.

### Related issues

Resolves issue #4298


tests/cli/test_cli.py
```diff
@@ -1771,6 +1776,24 @@ def test_run(ctx):
     )
     _run_command([cl, 'uedit', 'codalab', '--time-quota', ctx.time_quota])  # reset time quota
 
+    # Test that bundle fails when run without sufficient disk quota
+    _run_command([cl, 'uedit', 'codalab', '--disk-quota', f'{int(DISK_QUOTA_SLACK_BYTES)+1}'])
+    uuid = _run_command(
+        [cl, 'run', f'head -c {2*int(DISK_QUOTA_SLACK_BYTES)} /dev/zero > test.txt; sleep 100000',],
```
epicfaace: Add a comment here explaining we're creating a file twice the size of DISK_QUOTA_SLACK_BYTES

AndrewJGaut: I took that out to make the necessary file size smaller and to make it less confusing. Now, the bundle requests 1MB and the running bundle creates a file of size 1MB + 10 bytes, thereby going over the disk quota and causing the kill request.

Does this look a bit better?

AndrewJGaut: You can see the change here: https://github.com/codalab/codalab-worksheets/pull/4307/files/d12b0512d4935427f657a47bfb921075e45c8da2..6a8b2825e2d353d3f99a1e747acdaa8fe82ca6c1

tests/cli/test_cli.py
```diff
@@ -164,7 +164,7 @@ def get_info(uuid, key):
     return _run_command([cl, 'info', '-f', key, uuid])
 
 
-def wait_until_state(uuid, expected_state, timeout_seconds=1000):
+def wait_until_state(uuid, expected_state, timeout_seconds=1000, exclude_final_states=set()):
```
epicfaace: What exactly does "exclude_final_states" mean? Can you document in the docstring?

tests/cli/test_cli.py
```diff
@@ -1771,6 +1775,29 @@ def test_run(ctx):
...
+    disk_used = _run_command([cl, 'uinfo', 'codalab', '-f', 'disk_used'])
+    _run_command([cl, 'uedit', 'codalab', '--disk-quota', f'{int(disk_used) + 1000000}'])
+    uuid = _run_command(
+        [cl, 'run', f'head -c {1000010} /dev/zero > test.txt; sleep 100000',],
+        request_disk=None,
+        request_memory=None,
+    )
+    wait_until_state(
+        uuid, State.KILLED, timeout_seconds=500, exclude_final_states={State.FAILED}
+    )  # exclude State.FAILED because upon the initial upload attempt, the bundle state is set to FAILED
```
epicfaace: What does "exclude State.FAILED" mean?

codalab/model/bundle_model.py
```diff
@@ -1142,7 +1147,10 @@ def update_disk_metadata(self, bundle, bundle_location, enforce_disk_quota=False
 
         # TODO(Ashwin): make this non-fs specific
         data_hash = '0x%s' % (path_util.hash_directory(bundle_location, dirs_and_files))
-        data_size = path_util.get_size(bundle_location, dirs_and_files)
+        if hasattr(bundle.metadata, 'data_size'):
```
epicfaace: what does this change do? does it fix a separate bug?

codalab/model/bundle_model.py
```diff
@@ -1001,12 +1001,16 @@ def transition_bundle_running(self, bundle, worker_run, row, user_id, worker_id,
...
         }
 
-        # Increment user time as we go to ensure user doesn't go over time quota.
+        # Increment user time and disk as we go to ensure user doesn't go over quota.
         if user_id == self.root_user_id and hasattr(bundle.metadata, 'time'):
             time_increment = worker_run.container_time_total - bundle.metadata.time
             self.increment_user_time_used(bundle.owner_id, time_increment)
+        if hasattr(bundle.metadata, 'data_size'):
+            disk_increment = worker_run.disk_utilization - bundle.metadata.data_size
+            self.increment_user_disk_used(bundle.owner_id, disk_increment)
```
epicfaace: Can you explain a bit... where was `increment_user_disk_used` called previously? and are we still calling it in the previously-called place?

tests/cli/test_cli.py
AndrewJGaut: Yes. Essentially, I use this because some states aren't actually 'final'; in this case, the bundle enters State.FAILED before entering State.KILLED. The reason is that, after the bundle finishes, the worker attempts to upload the bundle to the bundle store, but that upload fails since the user doesn't have enough quota (so the bundle state is updated to State.FAILED). However, the bundle state is updated to State.KILLED soon after that.

Essentially, then, in this case, the State.FAILED is not the final state. There are a few ways to go about testing this:
- make it so we can test that the bundle state _eventually_ reaches State.KILLED (what I do here)
- change how the bundle state is actually updated (i.e. modify it so that it doesn't update to State.FAILED at all)
- just call wait_until_state and set expected_state to State.FAILED

I chose the first option because, by waiting for it to be killed, I could also check that the kill message is what I expected it to be.

tests/cli/test_cli.py
AndrewJGaut: It just means that wait_until_state should keep running even if the bundle reaches state State.FAILED. In this case, because of how the bundle is updated, that isn't actually a final state (it is updated to State.KILLED after being updated to State.FAILED as detailed in my reply to your comment right above this).

codalab/model/bundle_model.py
AndrewJGaut: It's not a bug, the data size was just being recomputed in this step so I skip the recomputation. I can add it back; it was just an optimization

codalab/model/bundle_model.py
AndrewJGaut: Sure thing!

Even after this change, the user disk quota is updated from scratch upon run bundle upload.

That update happens [in this function](https://github.com/codalab/codalab-worksheets/blob/e6e1e0eda7e7c10fcaa4ff0ae2bbf7b0694a873a/codalab/rest/bundles.py#L1083) [here](https://github.com/codalab/codalab-worksheets/blob/e6e1e0eda7e7c10fcaa4ff0ae2bbf7b0694a873a/codalab/rest/bundles.py#L1164). (And note that that function is only run if the 'filename' is included in the request query as seen [here](https://github.com/codalab/codalab-worksheets/blob/e6e1e0eda7e7c10fcaa4ff0ae2bbf7b0694a873a/codalab/rest/bundles.py#L1145), but that will always be the case since this API endpoint is called [here](https://github.com/codalab/codalab-worksheets/blob/d35148542d19cbab038b57524d15723aa887b327/codalab/worker/worker_run_state.py#L705) and the function used there is, by default, set [here](https://github.com/codalab/codalab-worksheets/blob/f2d4189fc6630858e144a67d648309953a05fdb7/codalab/worker/worker.py#L148) to be [this function](https://github.com/codalab/codalab-worksheets/blob/f2d4189fc6630858e144a67d648309953a05fdb7/codalab/worker/worker.py#L840), which uses the bundle_service_client function [here](https://github.com/codalab/codalab-worksheets/blob/f2d4189fc6630858e144a67d648309953a05fdb7/codalab/worker/bundle_service_client.py#L155) which sets the filename to be 'bundle.tar.gz', so in particular that field won't be empty.) That update_user_metadata function actually recomputes the user disk quota entirely from scratch (see the function call [here](https://github.com/codalab/codalab-worksheets/blob/94b5ca0dd17d195decea847669e70a38e806bc49/codalab/model/bundle_model.py#L1156) and its definition [here](https://github.com/codalab/codalab-worksheets/blob/94b5ca0dd17d195decea847669e70a38e806bc49/codalab/model/bundle_model.py#L2776)).

So, when the bundle finishes and its contents are uploaded, the user disk will be updated from scratch anyways (and by 'from scratch' I mean it will be completely recomputed to be the sum of the sizes of all the bundles currently owned by the user). This is how it's done before this change, and it will continue to be that way after this change.

Basically, then, increment_user_disk_used isn't currently being called anywhere except in bundles/rest.py; I actually added that function for a separate change I made last quarter. I use it here to just increment the user disk as the bundle runs, but the user disk quota update is done the exact same as before after the bundle finishes.

(By the way: overall, that the way the disk quota currently is updated and is managed is a bit confusing and oftentimes slows things down, and I actually think we could really speed up CodaLab by making our handling of the disk quota a bit tighter.)

codalab/model/bundle_model.py
```diff
@@ -1001,12 +1001,16 @@ def transition_bundle_running(self, bundle, worker_run, row, user_id, worker_id,
             'remote': worker_run.remote,
             'cpu_usage': cpu_usage,
             'memory_usage': memory_usage,
+            'data_size': worker_run.disk_utilization,
```
AndrewJGaut, line 1003: FYI, with regard to what we spoke about in the previous meeting, due to this change the size of the run bundle _is_ being updated on each bundle checkin, and so the updates to the user disk quota (which are made below by the increment_user_disk_used function) won't be overwritten when update_disk_metadata is called.

codalab/model/bundle_model.py
```diff
@@ -1001,12 +1001,16 @@ def transition_bundle_running(self, bundle, worker_run, row, user_id, worker_id,
...
         }
 
-        # Increment user time as we go to ensure user doesn't go over time quota.
+        # Increment user time and disk as we go to ensure user doesn't go over quota.
         if user_id == self.root_user_id and hasattr(bundle.metadata, 'time'):
             time_increment = worker_run.container_time_total - bundle.metadata.time
             self.increment_user_time_used(bundle.owner_id, time_increment)
+        if hasattr(bundle.metadata, 'data_size'):
+            disk_increment = worker_run.disk_utilization - bundle.metadata.data_size
```
AndrewJGaut: Note that this line works because the worker_run.disk_utilization is the new bundle data size and bundle.metadata.data_size is the old data size. The bundle metadata data_size isn't updated until this line [here](https://github.com/codalab/codalab-worksheets/blob/fix/4298-check-running-disk-quota/codalab/model/bundle_model.py#L1032).

```suggestion
            # disk_increment is the change in data_size from the previous cycle to the current one
            disk_increment = worker_run.disk_utilization - bundle.metadata.data_size
```

codalab/model/bundle_model.py
epicfaace, line 1003: Consider renaming disk_utilization to current_data_size

codalab/model/bundle_model.py
```suggestion
            # increment the user's disk_used as the bundle runs
                        self.increment_user_disk_used(bundle.owner_id, disk_increment)
```

codalab/model/bundle_model.py
epicfaace: maybe -- `data_size = getattr(bundle.metadata, "data_size", path_util.get_size(bundle_location, dirs_and_files))`

codalab/model/bundle_model.py
```diff
@@ -2719,9 +2727,9 @@ def increment_user_time_used(self, user_id, amount):
         user_info['time_used'] += amount
         self.update_user_info(user_info)
 
-    def increment_user_disk_used(self, user_id: str, amount: int):
+    def increment_user_disk_used(self, user_id: str, amount: int) -> None:
         """
-        Increment disk_used for user by amount
+        Increment number of bytes of disk used by user by amount.
```
```suggestion
        Increment disk_used (number of bytes of disk used) by user by amount.
```

codalab/model/bundle_model.py
```diff
@@ -2734,6 +2742,14 @@ def get_user_time_quota_left(self, user_id, user_info=None):
         time_used = user_info['time_used']
         return time_quota - time_used
 
+    def get_user_disk_quota_left(self, user_id: str, user_info=None) -> int:
+        if not user_info:
+            user_info = self.get_user_info(user_id)
+        user_info_dict = user_info
```
epicfaace: make this less verbose, revert to original implementation

codalab/rest/workers.py
```diff
@@ -59,6 +61,13 @@ def checkin(worker_id):
                     'https://codalab-worksheets.readthedocs.io/en/latest/FAQ/#how-do-i-request-more-disk-quota-or-time-quota'
                 )
                 messages.append({'type': 'kill', 'uuid': bundle.uuid, 'kill_message': kill_message})
+            elif local.model.get_user_disk_quota_left(bundle.owner_id) <= 0:
+                # Then, user has gone over their time quota and we kill the job.
+                kill_message = (
+                    'Kill requested: User disk quota exceeded. To apply for more quota, please visit the following link: '
+                    'https://codalab-worksheets.readthedocs.io/en/latest/FAQ/#how-do-i-request-more-disk-quota-or-time-quota'
```
epicfaace, line 68: refactor this to a constant

codalab/rest/workers.py
```diff
@@ -59,6 +61,13 @@ def checkin(worker_id):
                     'https://codalab-worksheets.readthedocs.io/en/latest/FAQ/#how-do-i-request-more-disk-quota-or-time-quota'
                 )
                 messages.append({'type': 'kill', 'uuid': bundle.uuid, 'kill_message': kill_message})
+            elif local.model.get_user_disk_quota_left(bundle.owner_id) <= 0:
+                # Then, user has gone over their time quota and we kill the job.
```
epicfaace:
```suggestion
                # Then, user has gone over their disk quota and we kill the job.
```

tests/cli/test_cli.py
epicfaace: let's simplify to a boolean

```suggestion
def wait_until_state(uuid, expected_state, timeout_seconds=1000, exclude_final_states=False):
```

epicfaace (approved):

epicfaace:
@AndrewJGaut what's the status of this PR?

AndrewJGaut:
> @AndrewJGaut what's the status of this PR?

Haven't had time to get to it. The current Actions run is failing because of a hasty change I made, but if you look at this prior one (https://github.com/codalab/codalab-worksheets/actions/runs/4171333091/jobs/7491776164) it almost worked. The only thing is I think there's a race condition that I've been trying to fix but have been unsuccessful thus far. I _don't_ think the issue really has to do with Kubernetes.

epicfaace:
@AndrewJGaut good to merge this now?

AndrewJGaut:
> @AndrewJGaut good to merge this now?

I was going to mention this in meeting today, but we could merge now or we could do one more quick review (since I made some changes since last review). I think it's OK either way

AndrewJGaut:
> > @AndrewJGaut good to merge this now?
> 
> I was going to mention this in meeting today, but we could merge now or we could do one more quick review (since I made some changes since last review). I think it's OK either way

In meeting, we decided to merge this now.

Issue #3965, opened 2022-01-21
Deleting a bundle that's on gcp doesn't delete the original file


epicfaace:

leilenah:
PR ready, needs review

PR #4308, opened 2022-11-09
Delete bundle content stored in Azure/GCS when deleting a bundle

wwwjn:
We need to delete the actual content when deleting the bundle, if the content is stored on cloud storage.

Test it using Azure and GCS. After user call `cl rm`, the remote content is deleted.

### Related issues
#3965 

### Checklist

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -21,6 +21,7 @@ from codalab.common import (
     NotFoundError,
 )
 from codalab.lib import canonicalize, spec_util, worksheet_util, bundle_util
+from codalab.lib.beam import filesystems
 from codalab.lib.beam.filesystems import LOCAL_USING_AZURITE, get_azure_bypass_conn_str
 from codalab.lib.server_util import (
     RequestSource,
@@ -1312,6 +1313,8 @@ def delete_bundles(uuids, force, recursive, data_only, dry_run):
         else:
             # Actually delete the bundle
             local.model.delete_bundles(relevant_uuids)
+        # TODO: get bundle location by uuiid
+        logging.info(f"relevant_uuids: {relevant_uuids}")
 
         # Update user statistics
         local.model.update_user_disk_used(request.user.user_id)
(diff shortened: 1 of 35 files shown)


wwwjn:
TODO: Test delete speed. create a new PR to make delete async.

codalab/rest/bundles.py
```diff
@@ -1310,6 +1314,16 @@ def delete_bundles(uuids, force, recursive, data_only, dry_run):
...
         else:
+            # If the bundle is stored on cloud, first delete data on cloud.
+            for uuid in relevant_uuids:
+                bundle_location = local.bundle_store.get_bundle_location(uuid)
+
+                file_location = '/'.join(bundle_location.split('/')[0:-1]) + "/"
+                if bundle_location.startswith(
+                    StorageURLScheme.AZURE_BLOB_STORAGE.value
+                ) or bundle_location.startswith(StorageURLScheme.GCS_STORAGE.value):
+                    FileSystems.delete([file_location])
```
AndrewJGaut, line 1325: Just want to make sure this works with both GCS and Azure, since I know in the upload case they had to be dealt with differently

AndrewJGaut (approved):

AndrewJGaut:
> TODO: Test delete speed. create a new PR to make delete async.

Yes, we should benchmark this. I need to do the same for the changes I'm trying to make to speed up cl rm anyways, so I can help with that.

codalab/rest/bundles.py
```diff
@@ -485,7 +486,10 @@ def _add_bundle_location(bundle_uuid: str):
         default_bundle_store = local.model.get_bundle_store(
             request.user.user_id, name=default_store_name
         )
-        if default_bundle_store['storage_type'] in (StorageType.AZURE_BLOB_STORAGE.value,):
+        if default_bundle_store['storage_type'] in (
+            StorageType.AZURE_BLOB_STORAGE.value,
+            StorageType.GCS_STORAGE.value,
```
epicfaace, line 491: what does this change do? Does it fix a bug?

codalab/rest/bundles.py
epicfaace, line 1325: possible to add a test? particularly for more complex code like `file_location = '/'.join(bundle_location.split('/')[0:-1]) + "/"`

codalab/rest/bundles.py
wwwjn: > what does this change do? Does it fix a bug?

Yeah this is a bug fixing. This `if` branch here is to handle when we set `CODALAB_DEFAULT_BUNDLE_STORE_NAME` is pointing to a cloud bundle store (Azure / GCS). However, I forgot to add GCS when get PR #4119 merged.  But we never set GCS as a default storage before, so it does not cause an actual error.

codalab/rest/bundles.py
wwwjn: > Just want to make sure this works with both GCS and Azure, since I know in the upload case they had to be dealt with differently


Yes it works for both Azure and GCS. Because this part of code is running on Codalab server, while bypass upload happens at the client side. At user client side, we need handle Azure and GCS differently.

wwwjn: This line is to get the folder path that continues content.gz and index.sqlite

Issue #4217, opened 2022-08-17
removing bundles on codalab.stanford.edu is extremely slow


epicfaace:

AndrewJGaut:
This still takes about 35-40 seconds with the root codalab user even after the search speedup. I'll look into this further.

AndrewJGaut:
Fixed with https://github.com/codalab/codalab-worksheets/pull/4312

PR #4312, opened 2022-11-17
Speed up rm bundles

AndrewJGaut:

Speed up rm bundles since it's very slow on the Stanford instance (and is slow in general on any instance that has lots of bundles).

### Related issues

#4217

### DIFF
--- a/codalab/lib/path_util.py
+++ b/codalab/lib/path_util.py
@@ -308,7 +308,7 @@ def remove(path):
     if parse_linked_bundle_url(path).uses_beam:
         from apache_beam.io.filesystems import FileSystems
 
-        if not FileSystems.exists(path):
+        if FileSystems.exists(path):
             FileSystems.delete([path])
         return True  # not sure about this one
     check_isvalid(path, 'remove')
--- a/codalab/lib/path_util.py
+++ b/codalab/lib/path_util.py
@@ -310,7 +310,7 @@ def remove(path):
 
         if not FileSystems.exists(path):
             FileSystems.delete([path])
-        return True # not sure about this one
+        return True  # not sure about this one
     check_isvalid(path, 'remove')
     set_write_permissions(path)  # Allow permissions
     if os.path.islink(path):
--- a/codalab/lib/path_util.py
+++ b/codalab/lib/path_util.py
@@ -310,24 +310,25 @@ def remove(path):
 
         if FileSystems.exists(path):
             FileSystems.delete([path])
-        return True  # not sure about this one
-    check_isvalid(path, 'remove')
-    set_write_permissions(path)  # Allow permissions
-    if os.path.islink(path):
-        os.unlink(path)
-        return False
-    elif os.path.isdir(path):
-        try:
-            shutil.rmtree(path)
             return True
-        except shutil.Error:
-            pass
     else:
-        os.remove(path)
-        return True
-    if os.path.exists(path):
-        print('Failed to remove %s' % path)
-        return False
+        check_isvalid(path, 'remove')
+        set_write_permissions(path)  # Allow permissions
+        if os.path.islink(path):
+            os.unlink(path)
+            return False
+        elif os.path.isdir(path):
+            try:
+                shutil.rmtree(path)
+                return True
+            except shutil.Error:
+                pass
+        else:
+            os.remove(path)
+            return True
+        if os.path.exists(path):
+            print('Failed to remove %s' % path)
+    return False
 
 
 def soft_link(source, path):
--- a/codalab/lib/path_util.py
+++ b/codalab/lib/path_util.py
@@ -310,20 +310,24 @@ def remove(path):
 
         if not FileSystems.exists(path):
             FileSystems.delete([path])
-        return
+        return True # not sure about this one
     check_isvalid(path, 'remove')
     set_write_permissions(path)  # Allow permissions
     if os.path.islink(path):
         os.unlink(path)
+        return False
     elif os.path.isdir(path):
         try:
             shutil.rmtree(path)
+            return True
         except shutil.Error:
             pass
     else:
         os.remove(path)
+        return True
     if os.path.exists(path):
         print('Failed to remove %s' % path)
+        return False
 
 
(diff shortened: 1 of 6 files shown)


AndrewJGaut:
Speedup on dev for deleting bundles:
Original code: ~2.48 seconds
Updated code: ~0.243 seconds

AndrewJGaut:
Note: the reason it was so slow prior is because it would recompute the user disk in entirety each time a bundle was removed (_get_disk_used(user_id) was called, which searches for all bundles owned by the user). To alleviate this, we now subtract the sum of the sizes of the deleted bundles from the user disk used instead, which is much faster.

codalab/model/bundle_model.py
```diff
@@ -2708,6 +2708,14 @@ def increment_user_time_used(self, user_id, amount):
         user_info['time_used'] += amount
         self.update_user_info(user_info)
 
+    def increment_user_disk_used(self, user_id, amount):
```
percyliang: Add type hints.

codalab/rest/bundles.py
```diff
@@ -1314,7 +1315,12 @@ def delete_bundles(uuids, force, recursive, data_only, dry_run):
             local.model.delete_bundles(relevant_uuids)
 
         # Update user statistics
-        local.model.update_user_disk_used(request.user.user_id)
+        # Just decrement the user disk used by the sum of sizes of bundles deleted
+        # OK, now let's add our change.
+        bundle_data_sizes = local.model.get_bundle_metadata(relevant_uuids, 'data_size')
+        local.model.increment_user_disk_used(
+            request.user.user_id, (-1) * sum(map(int, bundle_data_sizes.values()))
```
percyliang: Can just do `-sum...`

codalab/rest/bundles.py
```diff
@@ -1314,7 +1315,12 @@ def delete_bundles(uuids, force, recursive, data_only, dry_run):
             local.model.delete_bundles(relevant_uuids)
 
         # Update user statistics
-        local.model.update_user_disk_used(request.user.user_id)
+        # Just decrement the user disk used by the sum of sizes of bundles deleted
+        # OK, now let's add our change.
+        bundle_data_sizes = local.model.get_bundle_metadata(relevant_uuids, 'data_size')
```
percyliang: Does this diff produce the same answer as the above?  Just wondering if there are subtle differences?

codalab/rest/bundles.py
```diff
@@ -1314,7 +1315,12 @@ def delete_bundles(uuids, force, recursive, data_only, dry_run):
             local.model.delete_bundles(relevant_uuids)
 
         # Update user statistics
-        local.model.update_user_disk_used(request.user.user_id)
+        # Just decrement the user disk used by the sum of sizes of bundles deleted
+        # OK, now let's add our change.
```
percyliang: Delete this comment line?

percyliang (commented):
Ah, very nice!  Wonder if we could add a test for this.

codalab/rest/bundles.py
percyliang: Also check the case where some bundles that are being deleted don't have data (e.g., a bundle which has had `cl rm -d ...` done on it)

codalab/lib/path_util.py
```diff
@@ -310,20 +310,24 @@ def remove(path):
 
         if not FileSystems.exists(path):
             FileSystems.delete([path])
-        return
+        return True  # not sure about this one
```
wwwjn: `return True` look good to me

codalab/lib/path_util.py
```diff
@@ -310,20 +310,24 @@ def remove(path):
...
     check_isvalid(path, 'remove')
     set_write_permissions(path)  # Allow permissions
     if os.path.islink(path):
         os.unlink(path)
+        return False
     elif os.path.isdir(path):
         try:
             shutil.rmtree(path)
+            return True
         except shutil.Error:
```
wwwjn: If there is an error, it will have no return value here? Should we use `finally` here?

codalab/model/bundle_model.py
AndrewJGaut: Done.

codalab/rest/bundles.py
AndrewJGaut: Modified this to use something different anyways

codalab/rest/bundles.py
AndrewJGaut: It did not, it turns out, due to the following edge cases:
- cl rm -d: Here, the bundle contents are removed but not the bundle metadata. For a given bundle B, we must take care to decrement the user disk quota when *cl rm -d B* is run, but to not decrement it again when *cl rm B* is run (since that second *cl rm* just deletes the metadata and not anything from disk).
- Files uploaded with symlinks: These shouldn't affect user disk quota regardless, so when we remove them there should be no decrement to the user disk quota.

The new changes fix this.

codalab/rest/bundles.py
AndrewJGaut: Done.

codalab/rest/bundles.py
AndrewJGaut: Checked that case with the new tests.

codalab/lib/path_util.py
AndrewJGaut: Good catch! I added a return False statement at the end of the function to catch any other cases.

AndrewJGaut:
I made some changes to catch edge cases.

Previously, there were no test cases that I saw for verifying that disk quota was incremented and decremented properly upon file upload and deletion. So, I added test cases to check that disk quota is incremented and decremented properly upon:
- normal file upload and deletion
- normal file upload and deletion with rm -d before deletion with rm
- symlinked file upload and deletion

This helps to verify that my rm modification still runs correctly and to verify that our code handles disk quota usage correctly in various cases.

AndrewJGaut:
Now ready for final review @percyliang @wwwjn

Thanks for the reviews!

wwwjn (approved):

PR #4316, opened 2022-11-30
Support kubernetes runtime for launching jobs

epicfaace:
Continuation of https://github.com/codalab/codalab-worksheets/pull/4090 (it's the exact same as that PR).

@AndrewJGaut can you approve?


tests/cli/test_cli.py
```diff
@@ -1722,17 +1722,17 @@ def test_search_time(ctx):
...
-    wait_until_state(uuid, State.KILLED, timeout_seconds=60)
-    check_equals(
-        'Kill requested: User time quota exceeded. To apply for more quota,'
-        ' please visit the following link: '
-        'https://codalab-worksheets.readthedocs.io/en/latest/FAQ/'
-        '#how-do-i-request-more-disk-quota-or-time-quota',
-        get_info(uuid, 'failure_message'),
-    )
-    _run_command([cl, 'uedit', 'codalab', '--time-quota', ctx.time_quota])  # reset time quota
+    # _run_command([cl, 'uedit', 'codalab', '--time-quota', '2'])
```
AndrewJGaut: Uncomment this since it'll still work in Github Actions. Sorry again about the local testing issues!

.github/workflows/test.yml
```diff
@@ -150,14 +155,16 @@ jobs:
...
         if: matrix.runtime == 'kubernetes'
+        with:
+          go-version: '1.18.1'
       - name: Run tests using Kubernetes runtime
         if: matrix.runtime == 'kubernetes'
         run: |
           sh ./tests/test-setup.sh
           sh ./scripts/local-k8s/setup-ci.sh
-          #python3 test_runner.py --version ${VERSION} ${TEST}
+          # python3 test_runner.py --version ${VERSION} ${TEST}
```
AndrewJGaut: Shouldn't this be uncommented so that tests will use kubernetes runtime?

AndrewJGaut: When I look at the logs, it looks like the kubernetes runtime tests might only have executed the test setup (and not the actual tests) because this line is commented. But I could be wrong?

AndrewJGaut (commented):
Two minor changes, but we'll have to see if tests still pass after that test_runner line is uncommented.

tests/cli/test_cli.py
epicfaace: @AndrewJGaut I think the commented out part is actually pretty flaky. It seems to sometimes pass and other times fail. See https://github.com/codalab/codalab-worksheets/actions/runs/3587073727/jobs/6037163824 (latest commit on master branch).

.github/workflows/test.yml
epicfaace: Correct -- I commented this out to debug a CI issue -- and good catch! Re-added it now.

tests/cli/test_cli.py
AndrewJGaut: Yeah, that's happened for me too and the cause (in the cases I looked at) is that the timeout is a bit too aggressive. I'll increase the timeout in the PR I made to address the issues with local testing

codalab_service.py
```diff
@@ -326,7 +326,8 @@ def has_callable_default(self):
         name='worker_manager_worker_checkin_frequency_seconds',
         help='Number of seconds to wait between check-ins for a worker of the worker manager',
         type=int,
-        default=20,
+        # TODO(Ashwin): Change this back to 20 once we get websockets working with kubernetes workers.
+        default=5,
```
epicfaace, line 331: cc @AndrewJGaut @percyliang

docs/Development-Setup.md
```diff
@@ -7,15 +7,23 @@ If you want to test or develop with kubernetes locally, follow these steps to do
 
 ```
 # First, start codalab without a worker:
-codalab-service start -bds default no-worker
+codalab-service start -ds default no-worker
```
epicfaace: @AndrewJGaut instructions for running yourself

codalab/worker_manager/kubernetes_worker_manager.py
```diff
@@ -152,7 +168,8 @@ def start_worker_job(self) -> None:
         }
 
         # Start a worker pod on the k8s cluster
-        logger.debug('Starting worker {} with image {}'.format(worker_id, worker_image))
+        logger.error('Starting worker {} with image {}'.format(worker_id, worker_image))
+        print('starting...')
```
epicfaace: rm

codalab/worker/runtime/kubernetes_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+from codalab.common import BundleRuntime
+from codalab.worker.runtime import Runtime
+
+import urllib3
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+# https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#create-pod-v1-core
```
AndrewJGaut: Remove this?

codalab/worker/runtime/kubernetes_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+        Returns a Dict[index, UUID] of all NVIDIA devices available to docker
+
+        Arguments:
+            use_docker: whether or not to use a docker container to run nvidia-smi.
+
+        Raises docker.errors.ContainerError if GPUs are unreachable,
+            docker.errors.ImageNotFound if the CUDA image cannot be pulled
+            docker.errors.APIError if another server error occurs
+        """
+        return {}
```
AndrewJGaut, line 63: I assume you're planning on adding this in a later PR?

codalab/worker/runtime/kubernetes_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+            pod = utils.create_from_dict(self.k8_client, config)
+        except (client.ApiException, FailToCreateError, MaxRetryError, NewConnectionError) as e:
+            logger.error(f'Exception when calling Kubernetes utils->create_from_dict...: {e}')
+            raise e
+
+        return pod[0].metadata.name
+
+    def get_container_stats(self, pod_name: str):
+        # TODO: implement
+        return {}
```
AndrewJGaut, line 154: Also assuming you're saving this for a later PR?

codalab/worker/runtime/kubernetes_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def get_container_stats_with_docker_stats(self, pod_name: str):
+        """Returns the cpu usage and memory limit of a container using the Docker Stats API."""
+        # TODO: implement
+        return 0.0, 0
```
AndrewJGaut, line 159: Ditto

codalab_service.py
AndrewJGaut: Percy asked to keep this at 5 regardless, right? Just so we still have frequent checkins?

codalab_service.py
```diff
@@ -445,7 +447,7 @@ def has_callable_default(self):
         CodalabArg(
             name='worker_manager_{}_tag'.format(worker_manager_type),
             help='Tag of worker for {} jobs'.format(worker_manager_type),
-            default='codalab-{}'.format(worker_manager_type),
+            default='',
```
AndrewJGaut, line 450: Why remove this?

tests/unit/rest/bundles_test.py
```diff
@@ -42,6 +42,7 @@ def test_create(self):
         bundle_id = data[0]["id"]
         data[0]["attributes"].pop("state")
         data[0]["attributes"].pop("state_details")
+        data[0]["attributes"]["metadata"].pop("staged_status", None)
```
AndrewJGaut, line 45: Is this to fix that test that fails intermittently? Because I've had that issue, too

codalab/worker_manager/kubernetes_worker_manager.py
AndrewJGaut: Reminder to remove this

AndrewJGaut (changes requested):

tests/unit/rest/bundles_test.py
epicfaace: Yes, I think we should add this in. Because `state` is nondeterministic, we pop it. And though @leilenah added `staged_status` in an earlier PR, we didn't update this test, so we should also pop `staged_status` (since `staged_status` is linked to `state`).

codalab_service.py
epicfaace: I think otherwise you can't pass in an empty tag, or it will always use the default value (codalab-cpu or codalab-gpu).

codalab/worker/runtime/kubernetes_runtime.py
epicfaace: Yes :)

codalab/worker/runtime/kubernetes_runtime.py
epicfaace: Yes :)

codalab/worker/runtime/kubernetes_runtime.py
epicfaace: Yes. For GPU support

AndrewJGaut (approved):

PR #4320, opened 2022-12-05
Fix ci

AndrewJGaut:

CI is temporarily broken due to issue with Ubuntu-latest version not having support for python3.6 in python-setup Github Action. See issue here: https://github.com/actions/setup-python/issues/544

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   format:
     name: Format
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -392,7 +392,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.6
+          python-version: 3.7
```
wwwjn: Does this line means we are using Python3.7 as our default environment for all the tests?

AndrewJGaut: No that's just for a single test (Test backend - use azure as default bundle store) that was previously using python3.6. I could also change that test to use Ubuntu20.04 instead, but I didn't know why only that test was using Python3.6 so I changed it to Python3.7

wwwjn (approved):

PR #4329, opened 2022-12-07
Fix/4285 limit disk on parallel upload

### LINKED ISSUE #4285: make sure that users don't exceed disk quota with parallel uploads
Ideally, let people know that they don't have enough quota.
Subtlety: compressed versus uncompressed.

AndrewJGaut:

We want to make sure users can't bypass their disk quota limits by uploading files in parallel

### Related issues

Fixes #4285


codalab/lib/upload_manager.py
```diff
@@ -236,14 +241,32 @@ def write_fileobj(
...
+                    if self._client and iteration % ITERATIONS_PER_DISK_CHECK == 0:
+                        self._client.update(
+                            'user/increment_disk_used', {'disk_used_increment': len(to_send)}
+                        )
+                        user_info = self._client.fetch('user')
+                        if user_info['disk_used'] >= user_info['disk_quota']:
+                            raise Exception(
+                                'Upload aborted. User disk quota exceeded. '
+                                'To apply for more quota, please visit the following link: '
+                                'https://codalab-worksheets.readthedocs.io/en/latest/FAQ/'
```
epicfaace, line 266: can we refactor this link out into a constant?

codalab/lib/upload_manager.py
```diff
@@ -30,24 +30,29 @@
...
+        destination_bundle_store=None,
+        json_api_client=None,
     ):
         """
         params:
         bundle_model: Used on rest-server.
         bundle_store: Bundle store model, used on rest-server.
         destination_bundle_store: Indicate destination for bundle storage.
-        is_client: Whether this uploader is used on client side. Used on client.
+        json_api_client: A json API client. Only non-None if uploader is used on client side.
```
epicfaace:
```suggestion
        json_api_client: A json API client. Only set if uploader is used on client side; if the uploader is used on the server side, it is set to None.
```

codalab/worker/rest_client.py
```diff
@@ -114,7 +114,7 @@ def _make_request(
                     raise RestClientException('Invalid JSON: ' + response_data, False)
 
     def _upload_with_chunked_encoding(
-        self, method, url, query_params, fileobj, progress_callback=None
+        self, method, url, query_params, fileobj, pass_self=False, progress_callback=None
```
epicfaace, line 117: What does pass_self mean? Can you document here and in the other function where you add it

codalab/worker/upload_util.py
```diff
@@ -32,6 +33,11 @@ def upload_with_chunked_encoding(
         :param url: String. Location or sub url that indicating where the file object will be uploaded.
         :param need_response: Bool. Whether need to wait for the response.
         :param progress_callback: Function. Callback function indicating upload progress.
+        :param json_api_client: JsonApiClient. None when this function is run by the server.
+                                               Used to update disk usage from client.
+        :param check_disk: bool. True if the user's disk should be checked during the upload.
```
epicfaace: This param doesn't exist anymore

codalab/worker/upload_util.py
```diff
@@ -61,16 +67,33 @@ def upload_with_chunked_encoding(
```
epicfaace, line 64: Is this the exact same fn as the one in upload_manager.py? If so we might want to refactor / unify in a future PR

codalab/worker/upload_util.py
```diff
@@ -92,7 +115,10 @@ def upload_with_chunked_encoding(
                 logging.debug("Socket timeout, retrying url: %s", url)
                 conn.send(b'\0')
         logging.debug("Finished reading the response, url: %s", url)
+        print("Finished reading the response, url: %s", url)
```
epicfaace: remove these print statements

codalab/rest/users.py
```diff
@@ -239,3 +239,31 @@ def update_users():
...
+    Update the disk used for the user who makes the request to this endpoint.
+
+    This is required because users who are bypassing the server to upload
+    files straight to Azure will need their client to tell the server
+    to increment their disk used as file chunks are uploaded. They cannot
+    use the users/ PATCH endpoint since disk_used is in
+    USER_READ_ONLY_FIELDS. We make this special function (which only allows
+    positive disk increments so that users can't decrement their disk used) to ensure
+    that we can safely increment user disk used without introducing a
+    security flaw.
```
epicfaace, line 256: Add documentation as to what the request format should look like (i.e., the `disk_used_increment` key)

tests/cli/test_cli.py
```diff
@@ -805,23 +806,34 @@ def test_upload1(ctx):
...
+            # group_uuid = _run_command([cl, 'gnew', user_name])
+            switch_user(user_name)
+            worksheet_uuid = _run_command([cl, 'work', '-u'])
+            switch_user('codalab')
+            _run_command([cl, 'wperm', worksheet_uuid, 'public', 'a'])
+            switch_user(user_name)
+            _run_command([cl, 'work', worksheet_uuid])
+            # expect to fail when we upload something more than 2k bytes
+            check_contains(
+                'Upload aborted. User disk quota exceeded. '
```
epicfaace: Just check if it contains "User disk quota exceeded" -- this way we don't need to change this test if we slightly modify the error message in the future.

epicfaace (approved):

codalab/worker/rest_client.py
AndrewJGaut: Done.

codalab/worker/upload_util.py
AndrewJGaut: Done.

codalab/worker/upload_util.py
AndrewJGaut: Essentially. There are a few minor differences. I agree that it should be refactored.

codalab/worker/upload_util.py
AndrewJGaut: Done.

codalab/rest/users.py
AndrewJGaut: Done.

tests/cli/test_cli.py
AndrewJGaut: Done.

PR #4343, opened 2022-12-23
Worksheet View UI Optimizations

### LINKED ISSUE #4300: Optimize Worksheet View
## Implementation

- [x] Update worksheet view on dropdown 
    - Bundle details should be expanded 
- [x] Hide scroll down button when bundle is open
- [x] Back button
- [x] Disable header buttons when bundle is open
- [x] Fix Edit Source view
- [x] Fix Expand Shrink
- [x] Replace dropdown button with open button
- [x] Remove hover state from tables
    - only rows should have hover state
- [x] Update URL on bundle row click
    - e.g. ?bundle=abcd
- [x] Run modal
    - [x] Handle rub button click when modal is open
- [x] Show bundle CTA's
- [x] Remove unneeded code
    - [x] `showDetail` in BundleRow 
- [x] Documentation
- [x] Update tests
- [x] QA
    - [x] Worksheet with long content 

## Overview

https://worksheets.codalab.org/worksheets/0xb843335f02a9443995db2013fbe2e7f4

There is room for improvement in the worksheet view. Some comments from @percyliang:
* When you expand a bundle in the worksheet view, it sort of takes up the whole page, and is sort of weird to have it be still logically part of the table
    * we did this way in  the past to avoid having a modal dialogue, and the fact that you can have multiple expanded bundles, but this gets sort of confusing.
* It's hard to see the dependency structure of the bundles in the worksheet (we've talked about having a graph-based view of bundles).
* We've talked about having drag and drop on bundles

The scope of this issue is to determine what improvements can / should be made to the worksheet view. Once you've come up with a plan, present it to Percy for feedback before implementing.

leilenah:
### Major Changes

1. Bundle rows no longer expand in place when you click the down arrow icon on a bundle row in the worksheet view. If you click the [new] expand icon on a bundle row, the bundle detail view will be swapped into the worksheet (see demo).
2. The `/worksheets/:uuid` route now accepts an optional `:bundle_uuid` param. 
3. Initiating a new run from the worksheet view no longer opens the `NewRun` view in place. The `NewRun` view will now open in a modal.

### Minor Changes
1. A back button has been added to the worksheet view header.
2. Hover state has been removed from tables in worksheet view. Now, table hover styles only change when a user hovers over a specific bundle row.
3. Schema name field no longer defaults to an error state. 

### Related issues

https://github.com/codalab/codalab-worksheets/issues/4300

### Demo

https://user-images.githubusercontent.com/25855750/209410686-3b091e06-a5bb-4799-833d-48ec962f13cf.mp4

### Checklist


leilenah:
@percyliang ready for your eyes 🙇

percyliang (commented):
Overall, this is great - really nice work!  A few minor things that I ran into while playing with it:

1. Could we make the New Run modal take up a bit more of the screen?
1. If the user closes 'New Run', I wonder if it would make sense to keep the state around (in case they were typing a long command) so their work doesn't get lost?  It doesn't complicate things, so not super opinionated about this, but thought I'd bring it up.
1. 'Enter' doesn't work any more as a hot key to open up a bundleo
1. Could we have 'u' go back to the worksheet page from the bundle view like in Gmail?
1. When you go back to the worksheets page, it should keep the scroll exactly where you came from (currently, it scrolls to the top)
1. When you hit rerun, it takes you to a bundle page, but the scroll position is not at the top (it should be)
1. After hitting rerun, when hit 'Back', the cursor isn't anywhere (it should be on the bundle that we just came from)
1. When you load a worksheet, a blue bar should show at the top of the worksheet to denote that the cursor is before any bundle (this used to be there but is gone now)
1. When you hit 'gg' at the end of a worksheet that doesn't fit on one page, focus doesn't follow to the top of the worksheet (this was working before)

Some other bugs that I noticed (not related to this PR), but just putting them here for now:
1. Keyboard shortcuts scrolls off the end of the page
1. When you hold down 'j' or 'k' to scroll, the cursor position freezes instead of moving up/down the page gracefully as if you were in a text editor

leilenah:
@percyliang 

> Could we make the New Run modal take up a bit more of the screen?

done

> If the user closes 'New Run', I wonder if it would make sense to keep the state around (in case they were typing a long command) so their work doesn't get lost?  It doesn't complicate things, so not super opinionated about this, but thought I'd bring it up.

Currently, the `NewRun` data isn't stored anywhere when that component is removed from DOM. I would prefer to keep that as-is, but I can re-work that view later if this ends up being an annoyance for users. 

> 'Enter' doesn't work any more as a hot key to open up a bundle

fixed

> Could we have 'u' go back to the worksheet page from the bundle view like in Gmail?

You can now use `escape` to go back to the worksheet page. I prefer `escape` over `u` because `escape` is what is currently used to un-collapse bundle details and I don't want to confuse users who have already gotten used to this. `escape` also seems more intuitive.

> When you go back to the worksheets page, it should keep the scroll exactly where you came from (currently, it scrolls to the top)

fixed

> When you hit rerun, it takes you to a bundle page, but the scroll position is not at the top (it should be)

fixed

> After hitting rerun, when hit 'Back', the cursor isn't anywhere (it should be on the bundle that we just came from)

fixed

> When you load a worksheet, a blue bar should show at the top of the worksheet to denote that the cursor is before any bundle (this used to be there but is gone now)

fixed

> When you hit 'gg' at the end of a worksheet that doesn't fit on one page, focus doesn't follow to the top of the worksheet (this was working before)

fixed

> Keyboard shortcuts scrolls off the end of the page

Are you referring to the Keyboards Shortcuts modal that pops open when you click the shortcut info icon? If so, you can scroll to see the rest of the info. It appears that this implementation was intentional. 

> When you hold down 'j' or 'k' to scroll, the cursor position freezes instead of moving up/down the page gracefully as if you were in a text editor

Hm. Not sure how useful this would be.

percyliang:
This is great!  Two more things I noticed:

1. Now hitting 'G', which puts focus on the last bundle, doesn't scroll the page down to the bottom.
2. Hitting 'escape' from a worksheet page now goes back, which is new / jarring, so probably we should prevent that from happening.

leilenah:
@percyliang 

> Hitting 'escape' from a worksheet page now goes back, which is new / jarring, so probably we should prevent that from happening.

fixed

> Now hitting 'G', which puts focus on the last bundle, doesn't scroll the page down to the bottom.

This works on my end:

https://user-images.githubusercontent.com/25855750/209597158-6e4a44bb-b0d9-4153-8406-b452bc062e35.mp4

percyliang:
Interesting, that page works for me too, but if you create a new worksheet and create one bundle and then replicate it 100x times (probably easiest to do Raw Edit and copy/paste in the editor), then I think this should reproduce the bug.

leilenah:
@percyliang should be fixed now

percyliang:
'G' works, but when afterwards when I hit 'gg', I get:
image

leilenah:
@percyliang  yikes 😅  fixed.

percyliang (approved):
Seems to work now. :-)  Thanks for this great change - the interface feels so much more tight and natural!

leilenah:
@percyliang so glad you like it! 😃

Issue #4351, opened 2023-01-12
Disk quota computation redesign


AndrewJGaut:
Currently, the disk quota for a user is updated using the update_disk_metadata function, which uses the update_user_disk_used function, which re-computes the user's disk quota usage from scratch by just summing the size of all bundles owned by the user.

That's slow. Instead, we'd like to just increment and decrement when necessary. [ADD IN BENCHMARKING STATS -- Some can be seen in the "Speed Up Rm Bundles" PR, since the speedup implemented was to replace update_disk_metadata with more efficinet code.)

Issue #4350, opened 2023-01-12
Disk quota tests


AndrewJGaut:
Add in good testing to ensure that disk quota is updated and incremented correctly after various operations.

(The only tests we have for this were added in [this PR](https://github.com/codalab/codalab-worksheets/pull/4312), and they aren't currently sufficient for really testing the disk quota functionality.)

PR #4354, opened 2023-01-17
Speed up disk quota computation

AndrewJGaut:

Current method of updating disk quota is slow. Basically, currently the disk quota is updated using the following function:
```
def _get_disk_used(self, user_id):
        return (
            self.search_bundles(user_id, ['size=.sum', 'owner_id=' + user_id, 'data_hash=%'])[
                'result'
            ]
            or 0
        )
```
and then setting the disk quota to the returned value. However, doing such a search is quite slow, and so we want to do it faster. This PR speeds up the computation and just increments the user's disk quota each time by the correct amount.

This PR also adds a bunch of disk quota tests.

### Related issues

#4350 and #4351


AndrewJGaut:
TODO @AndrewJGaut : Add benchmarking

AndrewJGaut:
> TODO @AndrewJGaut : Add benchmarking

I might

codalab/lib/upload_manager.py
```diff
@@ -470,8 +472,6 @@ def upload_to_bundle_store(
                     params={'success': False, 'error_msg': f'Bypass server upload error. {err}',},
                 )
                 raise err
-            else:
```
epicfaace, line 496: why remove this?

tests/cli/test_cli.py
```diff
@@ -1354,6 +1387,56 @@ def test_rm(ctx):
...
+    uuid = _run_command([cl, 'run', 'head -c 50 /dev/zero > test.txt'])
+    wait_until_state(uuid, State.READY)
+    data_size = _run_command([cl, 'info', uuid, '-f', 'data_size'])
+    check_equals(
+        str(int(data_size) + int(disk_used)),
+        _run_command([cl, 'uinfo', 'codalab', '-f', 'disk_used']),
+    )
+    _run_command([cl, 'rm', uuid])
+    check_equals(disk_used, _run_command([cl, 'uinfo', 'codalab', '-f', 'disk_used']))
+
```
epicfaace, line 1425: Do these test code paths end up calling `cleanup_existing_contents`?

epicfaace (approved):

codalab/lib/upload_manager.py
AndrewJGaut: You can't see it in this abridged view, but if you look at the full file the code looks ilke: 
```
 try:
                ....
                self._client.update_bundle_state(bundle['id'], params={'success': True})
            except Exception as err:
                ...
            else:
                self._client.update_bundle_state(bundle['id'], params={'success': True})
```
What does this mean? Well, the 'else' branch triggers in a try-catch block when no error is thrown. In other words, when no errors was thrown, the ` self._client.update_bundle_state(bundle['id'], params={'success': True})` call was triggered twice. This wasn't an issue (except for efficiency reasons) before since the update wasn't incremental (it just reset the whole value), but it was for me since the disk increment happened twice.

tests/cli/test_cli.py
AndrewJGaut: Ah, shoot. No, I don't think so. Let me add one more test for that before I merge.

AndrewJGaut: Thanks for catching that!

Issue #4349, opened 2023-01-12
Benchmarking for Speed Improvements


AndrewJGaut:
We want to better understand what is slowing CodaLab down.

To do this, it would be helpful to be able to populate the Dev database with lots of bundles and then add in some timing code in various places to see what takes a long time during normal execution. (We can probably use the stress test script to do so.) We're then going to add timing code and play around with the system.

AndrewJGaut:
Focus on things that are slow in the life-cycle.

Do some analysis on what's slow!

PR #4362, opened 2023-01-25
Add in profiling

AndrewJGaut:
This addresses: #4349

Adds in code to take advantage of Sentry's profiling functionality (which is still in Beta) and to merge Yifan's Sentry bundle monitoring code.

This PR does a few things:
1. Merge Yifan's worker monitoring code in here. This gives a nice overview of how long each run stage is taking (see image below).
2. Add in profiling to the worker monitoring (so that we get profiling results for worker runs as well).
3. Add in profiling to the rest server.
4. Pass in a transaction sample rate from the command line. (We'll need to add an env variable for this and default value in the deployment repo as well).
5. Significantly beef up the timing tests so that we'll have a better idea of whether or not the code is slow.
Screen Shot 2023-02-28 at 11 34 43 AM


yifanmai:
I wonder if we could leave this in for production as well, but change the sampling rate accordingly? Something like a 1% sampling rate for production and 100% rate for dev.

AndrewJGaut:
> I wonder if we could leave this in for production as well, but change the sampling rate accordingly? Something like a 1% sampling rate for production and 100% rate for dev.

Agreed. I'll test what the performance hit is and see if that's viable. (Seems like a good idea to me, though.)

AndrewJGaut:
In terms of performance hit, it looks minimal when I'm doing tests on basic commands. However, one proposal we can discuss tomorrow would be to add another CI test that starts the codalab_service with CODALAB_SENTRY_TRANSACTION_RATE and CODALAB_SENTRY_PROFILES_RATE equivalent to the prod values (and the environment set to 'test') and then _only_ run the 'time' test for that setup to make sure it isn't slowing things down too much.

We can discuss tomorrow to see if that's a good idea?

codalab/lib/telemetry_util.py
```diff
@@ -25,9 +25,23 @@ def using_sentry():
 def initialize_sentry():
     """
     Initialize the Sentry SDK if it hasn't already been initialized.
+
+    Playing around with Sentry profiling
```
epicfaace: rm this comment?

codalab/lib/telemetry_util.py
```diff
@@ -25,9 +25,23 @@ def using_sentry():
...
     """
     if sentry_sdk.Hub.current.client is None:
-        sentry_sdk.init(dsn=CODALAB_SENTRY_INGEST, environment=CODALAB_SENTRY_ENVIRONMENT)
+        # Only do profiling in dev and test environments.
```
epicfaace: Which code restricts profiling to dev / test environments?

codalab/worker/worker_monitoring.py
```diff
@@ -0,0 +1,85 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Only do profiling in dev and test environments
+# And sample a lower percentage of transactions.
+transaction_sample_rate = float(os.getenv('CODALAB_SENTRY_TRANSACTION_RATE') or 0)
+profiles_sample_rate = float(os.getenv('CODALAB_SENTRY_PROFILES_RATE') or 0)
+assert 0 <= transaction_sample_rate <= 1
+assert 0 <= profiles_sample_rate <= 1
+sentry_sdk.init(
```
epicfaace, line 16: Should we move this profiling init code to a separate file (sentry_init.py or something) so we can reuse it?

codalab/worker/worker_monitoring.py
```diff
@@ -0,0 +1,85 @@
...
+    dsn=os.getenv('CODALAB_SENTRY_INGEST_URL'),
+    environment=os.getenv('CODALAB_SENTRY_ENVIRONMENT'),
+    traces_sample_rate=transaction_sample_rate,
+    _experiments={"profiles_sample_rate": profiles_sample_rate,},
+)
+
+
+class WorkerMonitoring(object):
+    def __init__(self):
+        # Each Hub can only have one open transaction at one time.
```
epicfaace, line 27: Explain what a Sentry Hub is in this comment.

codalab_service.py
```diff
@@ -293,6 +293,20 @@ def has_callable_default(self):
...
+        name='sentry_transaction_rate',
+        help=(
+            'Percentage of transactions we want to have Sentry sample. Only used if sentry_ingest_url and sentry_environment are provided.'
+        ),
+        default=0,
+    ),
+    CodalabArg(
+        name='sentry_profiles_rate',
+        help=(
+            'Percentage of transactions we want to have Sentry profile. Only used if sentry_ingest_url and sentry_environment are provided.'
```
```suggestion
            'Percentage of transactions we want to have Sentry profile. Only used if --sentry-ingest-url and --sentry-environment are provided.'
```

codalab_service.py
```diff
@@ -293,6 +293,20 @@ def has_callable_default(self):
         ),
         frontend=True,
     ),
+    CodalabArg(
+        name='sentry_transaction_rate',
+        help=(
+            'Percentage of transactions we want to have Sentry sample. Only used if sentry_ingest_url and sentry_environment are provided.'
```
```suggestion
            'Percentage of transactions we want to have Sentry sample. Only used if --sentry-ingest-url and --sentry-environment are provided.'
```

scripts/test_util.py
```diff
@@ -179,3 +181,46 @@ def cleanup(cl, tag, should_wait=True):
             run_command([cl, 'wrm', uuid, '--force'])
             worksheets_removed += 1
     print('Removed {} bundles and {} worksheets.'.format(bundles_removed, worksheets_removed))
+
+
+class timer:
```
epicfaace: uppercase?

epicfaace (approved):

Issue #4365, opened 2023-01-25
Run crfm-helm on CodaLab


yifanmai:
We need to be able to run crfm-helm end-to-end on CodaLab i.e. JSON assets are generated and hosted on CodaLab.

yifanmai:
Ran a [worksheet](https://worksheets.codalab.org/worksheets/0xc896f05df5c24dca95e4c581d9150340) with slightly less than 300 bundles. This seems to work reasonably well in terms of scheduling and UI.

CodaLab notes:

- When using Azure Batch Manager, it's possible for jobs to crash the worker, causing a `worker_offline` bundle state. This also causes the nodes in the Azure Batch pool to be permanently stuck in the "unusable" state. Unusable nodes have to be manually deleted. --> _we should look into this_
- Bundles with an upstream failed bundle get stuck in pending forever (should they get transitioned to failed, since it can never get scheduled?) --> _this is true only for worker-offline, so not an issue._
- How to throttle the number of bundles concurrently running? Less of a problem if we use our own worker queue (then we can control the number of workers using the worker manager). --> _worker parallel run quota_
- Would also be nice to the order of bundle scheduling / throttle the the number of bundles concurrently running per _HELM model_ (so we don't DDoS the model APIs) --> _we have --request-priority_

percyliang:
Not prioritizing.

PR #4367, opened 2023-01-26
Add Access-Control-Allow-Origin response header

yifanmai:

This allows websites not hosted CodaLab to make `XMLHttpRequests` to fetch files from CodaLab bundles.

### Related issues

Addresses #4365

### Checklist

### DIFF
--- a/codalab/rest/bundles.py
+++ b/codalab/rest/bundles.py
@@ -916,7 +916,7 @@ def _fetch_bundle_contents_blob(uuid, path=''):
     - `Content-Disposition: inline; filename=<bundle name or target filename>`
     - `Content-Type: <guess of mimetype based on file extension>`
     - `Content-Encoding: [gzip|identity]`
-    - `Access-Control-Allow-Origin: *` (only sent if the bundle is public)
+    - `Access-Control-Allow-Origin: *`
     - `Target-Type: file`
     - `X-CodaLab-Target-Size: <size of the target>`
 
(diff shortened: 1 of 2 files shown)


yifanmai:
My initial assessment that the header can only be added for public bundles is incorrect. It is safe to add the header to all responses, even for private bundles, because of the [how requests with credentials work](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#requests_with_credentials):

> By default, in cross-origin XMLHttpRequest or Fetch invocations, browsers will **not** send credentials. A specific flag has to be set on the XMLHttpRequest object or the Request constructor when it is invoked... the browser will reject any response that does not have the [Access-Control-Allow-Credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials): true header, and not make the response available to the invoking web content.

percyliang (approved):

codalab/rest/bundles.py
```diff
@@ -916,13 +916,15 @@ def _fetch_bundle_contents_blob(uuid, path=''):
...
     - `Content-Encoding: [gzip|identity]`
+    - `Access-Control-Allow-Origin: *` (only sent if the bundle is public)
     - `Target-Type: file`
     - `X-CodaLab-Target-Size: <size of the target>`
 
     HTTP Response headers (for directories):
     - `Content-Disposition: attachment; filename=<bundle or directory name>.tar.gz`
     - `Content-Type: application/gzip`
     - `Content-Encoding: identity`
+    - `Access-Control-Allow-Origin: *` (only sent if the bundle is public)
```
epicfaace: Please update this documentation to remove the "only sent if the bundle is public" part

epicfaace (approved):

PR #4400, opened 2023-02-14
Migration script that moves from disk storage to Azure

### LINKED ISSUE #4380: Create migration script that moves bundles from disk to blob
- The script should keep bundle id, but change the bundle storage type, and move it from disk to blob
- It can also do specific actions to clean up bundles (bundles which are not in database but are still on disk)

wwwjn:
We need to migrate from codalab disk store to Azure store, to reduce cost.

Run this script: `docker exec codalab_rest-server_1 python3 codalab/migration.py [worksheet_uuid]`

### Related issues

#4380 

### Checklist


codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+
+        return info
+
+    def upload_to_Azure(self, bundle_uuid, bundle_location, is_dir=False):
+        # generate target bundle path
+        file_name = "contents.tar.gz" if is_dir else "contents.gz"
+        target_location = f"{self.target_store_url}/{bundle_uuid}/{file_name}"
+
+        if FileSystems.exists(target_location):
+            path_util.remove(target_location)
```
AndrewJGaut, line 110: Why remove what already exists there? Could this potentially be dangerous?

It's seem to me this would only occur if this particular bundle had data stored both on disk and in Azure blob. Is there any scenario in which this would occur? Perhaps, instead of removing, we should just log that such a bundle exists and exit without doing anything for this particular bundle?

codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+            source_ext = ".tar.gz"
+            unpack = True
+        else:
+            # If it's a file, change it into GzipStream
+            source_fileobj = open(bundle_location, 'rb')
+            source_ext = ''
+            unpack = False
+
+        # Upload file content and generate index file
+        uploader.write_fileobj(source_ext, source_fileobj, target_location, unpack_archive=unpack)
```
AndrewJGaut, line 130: After my changes to the upload manager, this will modify the disk usage of the user on whose authority this function is called, just FYI. If it's the codalab user, we could just manually reset its disk quota after the migration. I could also add a flag to turn off disk quota checking, but that could potentially be unsafe.

codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+        try:
+            info = download_util.get_target_info(bundle_location, target, depth=0)
+            logging.info(f"[migration] {info}")
+        except Exception as e:
+            logging.info(f"[migration] Error: {str(e)}")
+            raise e
+
+        return info
+
+    def upload_to_Azure(self, bundle_uuid, bundle_location, is_dir=False):
```
AndrewJGaut: optional: call this upload_to_Azure_blob since we also have disk storage on Azure.

codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+        assert FileSystems.exists(target_location)
+
+    def modify_bundle_data(self, bundle, bundle_uuid, is_dir):
+        """
+        Change the bundle location in the database
+        ATTENTION: this function will modify codalab
+        """
+
+        # add bundle location
+        self.bundle_manager._model.add_bundle_location(bundle_uuid, self.target_store_uuid)
```
AndrewJGaut, line 141: I notice this calls an insert function. Will the bundle with this bundle_uuid not already have a bundle location in the DB? If so, should we delete that entry before adding another one (since the bundle's only location will be Azure blob).

codalab/migration.py
AndrewJGaut: One other idea is to decrement the user disk usage in origin_bundle_store, actually. That might even be better.

codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+                'is_dir': is_dir,
+                'metadata': {'store': self.target_store_name},
+            },
+        )
+
+    def delete_origin_bundle(self, bundle_uuid, bundle_location):
+        # Delete data from orginal bundle store
+        if os.path.exists(bundle_location):
+            path_util.remove(bundle_location)
+
```
AndrewJGaut, line 197: If you want to add the disk quota decrement here, you can simply do:
```
removed = path_util.remove(bundle_location)
if removed:
  data_size = self.bundle_manager._model.get_bundle_metadata(bundle_uuid, 'data_size')[bundle_uuid]
   self.bundle_manager._model.increment_user_disk_used(
                    user_id, -int(data_size)
                )
```
bundle_data_sizes = [bundle_uuid]

AndrewJGaut (commented):
Looks great! I love the code style. It looks really good. Just some nitpicks

codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+        if parse_linked_bundle_url(bundle_location).uses_beam:
+            # Do not migrate Azure / GCP bundles
+            continue
+
+        bundle_info = migration.get_bundle_info(bundle_uuid, bundle_location)
+
+        is_dir = bundle_info['type'] == 'directory'
+        migration.upload_to_Azure(bundle_uuid, bundle_location, is_dir)
+        migration.modify_bundle_data(bundle, bundle_uuid, is_dir)
+        migration.delete_origin_bundle(bundle_uuid, bundle_location)
```
epicfaace: Before we run delete_origin_bundle, should we first do a sanity check? such as listing the files in the Azure version, listing the files in the disk version, and comparing them?

codalab/migration.py
```diff
@@ -0,0 +1,216 @@
...
+        bundle_location = migration.get_bundle_location(bundle_uuid)
+
+        if parse_linked_bundle_url(bundle_location).uses_beam:
+            # Do not migrate Azure / GCP bundles
+            continue
+
+        bundle_info = migration.get_bundle_info(bundle_uuid, bundle_location)
+
+        is_dir = bundle_info['type'] == 'directory'
+        migration.upload_to_Azure(bundle_uuid, bundle_location, is_dir)
```
epicfaace: To make it easier to pause / resume this script or recover from errors:
- We should make sure upload_to_Azure, modify_bundle_data, and delete_origin_bundle are idempotent.
- Ideally, we should see if we could run upload_to_Azure, modify_bundle_data, and delete_origin_bundle together atomically.

codalab/migration.py
epicfaace: This could occur if the script failed in the past and we need to resume the script. So this line makes it idempotent.

codalab/migration.py
wwwjn: Thanks for reminding me this! I added a function call to decrement the user disk usage when deleting the bundle.

codalab/migration.py
wwwjn: First we check the bundle is stored in disk. For our current database design, if the bundle is stored in disk, it will not have a item in `bundle_location` table. This table is only used for blob storage.

codalab/migration.py
wwwjn: Wow thanks Andrew! Since the old bundle metadata is already modified here, so we can not use this size. But thanks a lot for reminding me function `increment_user_disk_used`!

codalab/migration.py
```diff
@@ -0,0 +1,253 @@
...
+        "azure-store-default" if args.target_store_name is None else args.target_store_name
+    )
+    if not spec_util.UUID_REGEX.match(worksheet_uuid):
+        raise Exception("Input worksheet uuid has wrong format. ")
+
+    # TODO: write output to log / log files
+    migration = Migration(target_store_name)
+    migration.setUp()
+
+    bundle_uuids = migration.get_bundle_uuids(worksheet_uuid)
```
epicfaace, line 231: Just get all bundles, not specific to a worksheet

codalab/migration.py
```diff
@@ -0,0 +1,253 @@
...
+            old_file_list.sort()
+
+            assert old_file_list == new_file_list
+        else:
+            # For files, check the file has same contents
+            old_content = read_file_section(bundle_location, 5, 10)
+            new_content = read_file_section(new_location, 5, 10)
+            assert old_content == new_content
+
+    def delete_origin_bundle(self, bundle_uuid, bundle_location):
```
epicfaace: Rename to delete_original_bundle

epicfaace (approved):

epicfaace:
@wwwjn we should be good to merge this

PR #4403, opened 2023-02-15
Kubernetes -- make runtime configurable

epicfaace:
Make runtime configurable for the kubernetes worker manager with the `--bundle-runtime` flag and associated environment variables. @yifanmai this would let you run the kubernetes worker manager with the docker runtime (as opposed to the kubernetes runtime).


AndrewJGaut:
@epicfaace Just to clarify, if you're using the Docker bundle_runtime with Kubernetes worker manager, this would imply that: your workers are scheduled as pods on the cluster, and the each worker pulls the Docker image and runs the specified bundle, right?

And then Kubernetes run-time with Kubernetes worker manager would imply that: the workers are run as pods in the cluster AND, to schedule the run bundles, the workers just use the KubeAPI to schedule those run bundles as pods in the cluster. Is this correct?

codalab/worker/main.py
```diff
@@ -133,13 +133,9 @@ def parse_args():
     )
     parser.add_argument(
         '--bundle-runtime',
-        choices=[
-            BundleRuntime.DOCKER.value,
-            BundleRuntime.KUBERNETES.value,
-            BundleRuntime.SINGULARITY.value,
-        ],
+        choices=[BundleRuntime.DOCKER.value, BundleRuntime.KUBERNETES.value,],
```
AndrewJGaut, line 135: Why remove singularity here? I see the referenced issue, but it doesn't detail why this should be removed.

AndrewJGaut (approved):

codalab/worker/main.py
epicfaace: Singularity support was something we had worked on but was unfinished, and we don't need it anymore, so we'd like to remove it.

epicfaace:
> @epicfaace Just to clarify, if you're using the Docker bundle_runtime with Kubernetes worker manager, this would imply that: your workers are scheduled as pods on the cluster, and the each worker pulls the Docker image and runs the specified bundle, right?
> 
> And then Kubernetes run-time with Kubernetes worker manager would imply that: the workers are run as pods in the cluster AND, to schedule the run bundles, the workers just use the KubeAPI to schedule those run bundles as pods in the cluster. Is this correct?

Correct!

Issue #4404, opened 2023-02-22
Stress tests running slow


AndrewJGaut:

AndrewJGaut:
Fully solved by PR #4406 and PR #4421.

PR #4406, opened 2023-02-27
Fix Stress Tests

AndrewJGaut:
This addresses: #4404

This PR does several things to alleviate issues brought to light by stress tests:
- **Reset ITERATIONS_PER_DISK_CHECK** to be 2000 rather than 1. When it was 1, upload speed was _significantly_ slower since the client was sending a request to server to update user_disk_used for every 16KB of data uploaded. With it now set to 2000, the user_disk_used will be updated every 32MB (and will be updated with the initial 16KB on the first iteration of the upload). Upload speeds are now approximately in line with previous speeds according to benchmarking tests.
- **Fix an SQL deadlock issue**. When user_disk_used was incremented, we were selecting the user disk used in one statement, incrementing it, and then updating it in another statement. This was not atomic and was leading to an SQL deadlock and possible race conditions. To alleviate this, we now use select_for_update rather than just select when getting user_disk_used.
- **Fix bundle finalization issue**. Stress tests were also taking significantly longer because bundles were staying in the "finalized" state for a long time. It turns out that the socket timeout set in the bundle_manager for the finalization message sent to workers was too low, and so we were frequently getting a socket timeout error for that message and the bundle_manager was taking a long time to progress to actual finalization of the bundle.
- **Update stress tests**: Stress tests now include timing for each individual stress test as well as overall timing. This gives us a better handle for _which_ tests are taking up the majority of the time.


AndrewJGaut:
I should note that stress tests are now finishing in around 4 hours as they were previously and there were no SQL deadlock errors that I saw. Once this PR is in, therefore, we should finally be able to deploy again!

tests/stress/stress_test.py
```diff
@@ -38,34 +46,38 @@ class TestFile:
 
     def __init__(self, file_name, size_mb=1, content=None):
         self._file_name = file_name
+        self._file_path = temp_path(file_name)
```
AndrewJGaut, line 49: I should note this oddity: suddenly, when running stress tests, I started getting an `Errno 13 Permission Denied` error when the stress tests would try to create a file. To get around this, instead of creating the file in the working directory, I create it /tmp/ and then update the rest of the code accordingly.

Not sure why this issue just started popping up, though.

codalab/model/bundle_model.py
```diff
@@ -2714,21 +2714,50 @@ def update_user_info(self, user_info):
                 cl_user.update().where(cl_user.c.user_id == user_info['user_id']).values(user_info)
             )
 
-    def increment_user_time_used(self, user_id, amount):
+    def increment_user_disk_used(self, user_id: str, amount: int):
         """
-        User used some time.
+        Increment disk_used for user by amount.
+        When incrementing values, we have to use a special query to ensure that we avoid
```
epicfaace, line 2720: can we shorten this docstring? probably a bit too much detail for posterity

codalab/model/worker_model.py
```diff
@@ -407,14 +407,15 @@ def send_json_message(self, socket_id, worker_id, message, timeout_secs, autoret
...
+                    logging.error(f"socket error when calling send_json_message: {e}")
 
                 if not success:
                     # Shouldn't be too expensive just to keep retrying.
                     # TODO: maybe exponential backoff
+                    logging.error("Sleeping for 0.1 seconds.")
                     time.sleep(
-                        0.3
+                        0.1
                     )  # changed from 0.003 to keep from rate-limiting due to dead workers
```
epicfaace:
```suggestion
                    )
```

scripts/test_util.py
```diff
@@ -179,3 +181,48 @@ def cleanup(cl, tag, should_wait=True):
             run_command([cl, 'wrm', uuid, '--force'])
             worksheets_removed += 1
     print('Removed {} bundles and {} worksheets.'.format(bundles_removed, worksheets_removed))
+
+
+class timer:
```
epicfaace: uppercase?

epicfaace (approved):
Looks great!

tests/stress/stress_test.py
```diff
@@ -393,6 +391,7 @@ def main():
     runner.run()
     duration_seconds = time.time() - start_time
     print("--- Completion Time: {} minutes---".format(duration_seconds / 60))
+    print(json.dumps(runner._runs))
```
epicfaace, line 394: let's print this before we print completion time.

epicfaace (approved):
Looks great!

PR #4411, opened 2023-03-01
Kubernetes: ensure only one worker pod can be scheduled per node

### LINKED ISSUE #4384: Kubernetes: localize one worker pod per node
We need to localize one worker pod per node. This way, each worker can control / launch only the pods available on each node.

epicfaace:
Fixes https://github.com/codalab/codalab-worksheets/issues/4384.


epicfaace:
@AndrewJGaut can you review?

epicfaace:
2023-04-26 19:33:52,976 Exception when calling Kubernetes utils->create_from_dict: Error from server (Unprocessable Entity): {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod \"cl-worker-c259b7deae854dffaa46c77a05d4ef40\" is invalid: [spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey: Required value: can not be empty, spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey: Invalid value: \"\": name part must be non-empty, spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey: Invalid value: \"\": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')]","reason":"Invalid","details":{"name":"cl-worker-c259b7deae854dffaa46c77a05d4ef40","kind":"Pod","causes":[{"reason":"FieldValueRequired","message":"Required value: can not be empty","field":"spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey"},{"reason":"FieldValueInvalid","message":"Invalid value: \"\": name part must be non-empty","field":"spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey"},{"reason":"FieldValueInvalid","message":"Invalid value: \"\": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')","field":"spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey"}]},"code":422}

codalab/worker_manager/kubernetes_worker_manager.py
```diff
@@ -138,7 +138,7 @@ def start_worker_job(self) -> None:
         config: Dict[str, Any] = {
             'apiVersion': 'v1',
             'kind': 'Pod',
-            'metadata': {'name': worker_name},
+            'metadata': {'name': worker_name, 'labels': {'type': 'cl-worker'}},
```
AndrewJGaut: Nit: change 'type' to 'app'. I see that a lot more often (e.g. [here](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)).

codalab/worker_manager/kubernetes_worker_manager.py
```diff
@@ -158,6 +158,27 @@ def start_worker_job(self) -> None:
...
+                                        "matchExpressions": [
+                                            {
+                                                "key": "type",
+                                                "operator": "In",
+                                                "values": ["cl-worker"],
+                                            }
+                                        ]
+                                    },
+                                },
+                                'topologyKey': 'topology.kubernetes.io/zone',
```
AndrewJGaut: I'm actually not sure this topology key is correct; from the example given [here](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone), it looks to me like this word only work if each of the VMs serving as nodes were spun up in different [zones](https://cloud.google.com/compute/docs/regions-zones#zones_and_clusters) on GCP. Will that be the case?

tests/cli/test_cli.py
```diff
@@ -1230,6 +1230,18 @@ def test_upload_default_bundle_store(ctx):
     check_contains(bundle_store_name, _run_command([cl, "info", uuid]))
```
AndrewJGaut, line 1229: We might want to add one more test where two nodes are spun up and we schedule two worker pods and make sure they both are scheduled on different nodes.

AndrewJGaut: Not sure how feasible this is, though.

AndrewJGaut (changes requested):

AndrewJGaut:
One other thing: I saw the linked issue, but can you give a little more detail?

I have two questions:
-  Why is it desirable to have one worker pod per node? Why is  "worker can control / launch only the pods available on each node." desirable?
- By my reading of the [Kubernetes Runtime](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/runtime/kubernetes_runtime.py#L103-L141), it looks like the workers will just schedule the jobs as pods in the cluster that could run on any node -- so I don't see how "worker can control / launch only the pods available on each node." will be true even after this PR.


Thanks!!

epicfaace:
> * Why is it desirable to have one worker pod per node? Why is  "worker can control / launch only the pods available on each node." desirable?

This is because each worker needs to report the amount of memory / CPU that is free for a given node. Because each worker launches pods in kubernetes (as opposed to pre-kubernetes when workers would launch docker images within the same worker), the worker queries the k8s api to get the amount of memory / CPU free for a given node. This only works if we have one worker per node, though. If we have multiple workers per node, then each worker may report the parent node's availability, which would overreport how much memory / CPU is actually available.




> * By my reading of the [Kubernetes Runtime](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/runtime/kubernetes_runtime.py#L103-L141), it looks like the workers will just schedule the jobs as pods in the cluster that could run on any node -- so I don't see how "worker can control / launch only the pods available on each node." will be true even after this PR.

We could certainly change that behavior as part of https://github.com/codalab/codalab-worksheets/pull/4413 so that it _does_ launch pods on the same node:

image

And after reading your comment, I initially thought that was an oversight of https://github.com/codalab/codalab-worksheets/pull/4413. But _then_ I realized that, actually, we don't need this to happen. It's completely fine if workers can launch pods that get run on other nodes instead. If a pod gets assigned to another node, it won't do any harm, because the worker in the other node will just report less CPU / memory available. In fact, allowing this behavior will be better because pods can be scheduled more efficiently. The **key thing we need to keep in constant** is that: each worker does not launch a pod when the cluster doesn't have capacity to run it; that's why keeping one worker per node, and reporting cpu/gpu/memory from parent node (https://github.com/codalab/codalab-worksheets/pull/4413), is good enough to ensure that.

codalab/worker_manager/kubernetes_worker_manager.py
epicfaace: Done

codalab/worker_manager/kubernetes_worker_manager.py
epicfaace: I'm not sure, these docs (https://blog.kubecost.com/blog/kubernetes-node-affinity) say "The topologyKey topology.kubernetes.io/zone ensures that the pod is scheduled only if the node is within the same zone as one of the existing pods with the matching key-value pairs. Note that some constraints are applied to the topology due to performance and security considerations found in the Kubernetes documentation."

Not exactly sure that that means though.

AndrewJGaut:
Ahhh I see. OK, this makes sense to me now. Yup, great point: if we did try to ensure workers only launched pods on the same nodes, then we'd effectively not be using Kubernetes's scheduler, which would be bad. That looks good to me.

codalab/worker_manager/kubernetes_worker_manager.py
AndrewJGaut: I see. I'll double-check this. I'm just concerned that if the zone really is the GCP zone (e.g. `us-west-1c`), then we'll only be able to have a single pod on the entire cluster of nodes within that GCP zone and so this won't work... but I'll double-check

AndrewJGaut (approved):
Thanks for making the relevant changes!

codalab/worker_manager/kubernetes_worker_manager.py
epicfaace: I believe we fixed this now

AndrewJGaut: Yeah, I believe we did by changing the topology key from zone to hostname.

Issue #4385, opened 2023-02-08
Kubernetes: worker pods CPU / GPU / disk report change


epicfaace:
We should change how available resources (CPU / GPU / disk available) are reported by worker pods. Currently, I think a worker pod will just report the amount of CPU / GPU / disk allocated by that pod. Instead, the worker pod should report the amount of CPU / GPU / disk available by its parent node -- this number will accordingly decrease as more job pods are allocated and run on the node.

PR #4412, opened 2023-03-01
Remove assign_cpu_and_gpu_sets

epicfaace:

Remove assign_cpu_and_gpu_sets as I don't think this check is needed. This is important as it simplifies https://github.com/codalab/codalab-worksheets/issues/4385, because workers won't need to do checks on their own cpu / gpusets (they can just request pods with specific GPUs). I could have just bypassed this code path for the kubernetes runtime, but it seems to be an unnecessary check we can just do away with -- because the bundle manager already handles ensuring the worker has enough CPUs / GPUs available to run a bundle. What do you think @AndrewJGaut @percyliang ?


codalab/worker/worker.py
```diff
@@ -584,44 +583,6 @@ def process_runs(self):
...
-                "Requested more CPUs (%d) than available (%d currently out of %d on the machine)"
-                % (request_cpus, len(cpuset), len(self.cpuset))
-            )
-        if len(gpuset) < request_gpus:
-            raise Exception(
-                "Requested more GPUs (%d) than available (%d currently out of %d on the machine)"
-                % (request_gpus, len(gpuset), len(self.gpuset))
-            )
-
-        def propose_set(resource_set, request_count):
```
epicfaace, line 626: Is there a chance that the user will actually get _less_ CPUs / GPUs than they request? If so, then we might need to keep this function.

epicfaace:
We should discuss this in a meeting.

codalab/worker/worker_run_state.py
```diff
@@ -237,19 +237,7 @@ def mount_dependency(dependency, shared_file_system):
             )
             return run_state._replace(stage=RunStage.CLEANING_UP)
 
-        # Check CPU and GPU availability
-        try:
-            cpuset, gpuset = self.assign_cpu_and_gpu_sets_fn(
```
epicfaace, line 242: Add an if statement -- consider just not doing this with kubernetes.

epicfaace:
@AndrewJGaut any thoughts on this?

AndrewJGaut:
> @AndrewJGaut any thoughts on this?

I believe you're correct @epicfaace and `assign_cpu_and_gpu_sets` can be removed.

Here's why: We do check if the worker has enough resources in the bundle-manager and keep a tally on the worker's cpus and gpus (e.g. see [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L552)).

Now, we could run into issues if, for instance, the bundle_manager was starting bundles in a multi-threaded fashion (and sent a start message to the same worker multiple times before decrementing the worker['cpus'] and/or worker['gpus'] in that function).However, the [condition](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L544) for entering the loop in which those resources are decremented will not return until the [rest-server has received the message](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L742) and sent it to the worker has received the message (or it returns False and no bundle is started anyway). Therefore, it will _never_ be the case as long as bundle_manager is single-threaded that there will be a race condition related to worker cpu and gpu counts.

Since the bundle-manager will _never_ schedule a run on the worker unless it has sufficient resources for _all_ of its running bundles, it should never be the case that there aren't enough cpus and/or gpus on the worker to run a bundle.

Therefore, the `assign_cpu_and_gpu_sets` check done in the worker is _not_ necessary and can be removed. However, it could become necessary if we wanted the worker and/or bundle-managers to be multi-threaded (with respect to transitioning bundle stages on the worker and run scheduling bundles on the bundle-manager). In that case, though, I'd recommend we keep a member variable in the worker keeping track of the number of cpus and gpus available rather than looping through the count for all run bundles as done [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L610-L613).

AndrewJGaut:
Some extra interesting tidbits:
-  The [beginning of the function](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/worker.py#L610-L613) is the only place where we check whether or not there are enough CPUs or GPUs on the worker currently to run that bundle.
- The scheduling algorithm on the bundle-manager makes more sense to me now. It l[oops over bundles it's trying to run](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L503). For each bundle, it calls a [function](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L541) that [checks](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L669) if any worker has the resources required to run that bundle. If so, it [tries to start the bundle](https://github.com/codalab/codalab-worksheets/blob/master/codalab/server/bundle_manager.py#L543-L555) on one of those workers and then deducts resources if that fails. So, in particular, it does each bundle individually and with a single thread, which is why we can be sure that no race conditions will arise.
- The bundle-manager only does one thing with multiple threads: make bundles
- I can't actually find where the cpu sets and gpu sets are incremented after a run bundle finishes... but I know that it does b/c otherwise workers wouldn't be able to schedule bundles ad infinitum

PR #4413, opened 2023-03-01
Report cpu/gpu/memory from parent node

### LINKED ISSUE #4385: Kubernetes: worker pods CPU / GPU / disk report change

epicfaace:

Report cpu/gpu/memory from parent node, fixes https://github.com/codalab/codalab-worksheets/issues/4385.


tests/cli/test_cli.py
```diff
@@ -480,6 +480,7 @@ def __exit__(self, exc_type, exc_value, tb):
 
         # Clean up and restore original worksheet
         print("[*][*] CLEANING UP")
+        return
```
AndrewJGaut: Make sure to delete this!

AndrewJGaut (changes requested):
Aside from needing to remove that one line it looks like you temporarily added in `test_cli.py`, looks great!

AndrewJGaut (approved):

PR #4416, opened 2023-03-01
Remove back button

### LINKED ISSUE #4414: Update "back" button behavior
The "back" button is pretty confusing. I also often accidentally click it to try to exit from the web CLI, but then the whole page goes to the previous page. It's basically the same as the browser back button too, so not needed.

epicfaace:
fixes #4414


epicfaace:
Change it to gray it out unless you're on the expanded bundle page.

Issue #4405, opened 2023-02-22
[No queue: nlp requested] error happening intermittently when running nlp queue with Stanford workers


AndrewJGaut:

AndrewJGaut:
I can't reproduce this anymore. I believe it's working correctly.

AndrewJGaut:
Reopening this as there is an issue that we've identified that occurs with bundles run on worksheets that are not publicly readable.

For instance, consider the following example, with the following steps:
![image](https://user-images.githubusercontent.com/35617203/225229021-bae0988d-7b91-4f0d-97fe-ef618625dada.png)
Here's me executing those commands:
<img width="468" alt="image" src="https://user-images.githubusercontent.com/35617203/225229153-83221ba3-6b55-4cf9-ac95-b9b192511c54.png">

Now, the weird thing is, I can't reproduce this locally! When I do something similar (with private worksheet test-ws) I get the following:
<img width="468" alt="image" src="https://user-images.githubusercontent.com/35617203/225229238-90b92f28-673f-4dd6-9397-0f50a85376e9.png">
We see that the staged bundle on the private worksheet does show up in the search output!
The only difference I've been able to identify thus far is that the staged_status fields differ. On my local, I don't get the `No queue (nlp requested) issue; see here:
<img width="468" alt="image" src="https://user-images.githubusercontent.com/35617203/225229398-bd7c0492-193c-43fb-86fd-4da368e1329e.png">


Now, the reason this is causing the issue that we see from the worker-manager is that the worker-manager uses a search call (equivalent to that made by the client when running `cl search`) that precisely runs `cl search .mine state=staged`.  Therefore, any bundles that don't show up in that output aren't picked up by the slurm worker-manager and so they remain in staged forever.

Note: we have verified that the slurm worker-manager and `cl search` work fine with bundles on _public_ worksheets.

The question: why are staged bundles on private worksheets not being picked up by `cl search` on prod, even though they are when I run the same commands locally? I'm investigating further...

AndrewJGaut:
Moreover, I don't see anything amiss in the query or the database entry.


The query looks as follows:
![image](https://user-images.githubusercontent.com/35617203/225237821-86727413-bdf5-47ef-a0a7-e0453312ad35.png)

Here's what the bundle that's staged in the private worksheet looks like in the prod database:
![image](https://user-images.githubusercontent.com/35617203/225231375-d2022df7-d0f8-4cde-85a6-8a439876ded4.png)

[...]

PR #4432, opened 2023-03-31
Fix search on private worksheets

AndrewJGaut:
We discovered an issue where bundles on private worksheets were not visible to `cl search` and would not appear in its output. Turns out there was a minor error in the `search_bundles` function when searching for private bundles with a non-root user (we used a JOIN instead of a LEFT OUTER JOIN in one place). This fixes that issue.

### Changes
- Fix issue for non-root user searching for private bundles.
- Beef up tests. Add in test for searching for bundles on private worksheets. Beef up non-root user tests significantly. (Before, we had a separate section for non-root user tests for `search` which weren't as rigorous as the tests used for the root user. Now, we apply the exact same tests for the root user and a non-root user.)

### Related issues
#4405


tests/cli/test_cli.py
```diff
@@ -1890,22 +1851,70 @@ def test_search(ctx):
...
+    if not os.getenv('CODALAB_PROTECTED_MODE'):
+        # This test does not work when protected_mode is True.
+        _, current_user_name = current_user()
+        user_name = 'non_root_user_' + random_name()
+        create_user(ctx, user_name, disk_quota='1000000')
+        switch_user(user_name)
+        new_wuuid = _run_command([cl, 'new', random_name()])
+        _run_command([cl, 'work', new_wuuid])
+        test_search_helper(ctx)
         switch_user(current_user_name)
```
AndrewJGaut, line 1917: Call helper function here to actually run the tests. Note we call the same helper for the root and non-root user, so the exact same tests are performed in both cases

tests/cli/test_cli.py
```diff
@@ -1837,50 +1837,11 @@ def test_perm(ctx):
...
-    check_equals('', _run_command([cl, 'search', '.shared']))
-    # Check search with non-root user.
-    if not os.getenv('CODALAB_PROTECTED_MODE'):
-        # This test does not work when protected_mode is True.
-        _, current_user_name = current_user()
-        user_name = 'non_root_user_' + random_name()
-        create_user(ctx, user_name, disk_quota='2000')
-        switch_user(user_name)
-        check_equals(uuid1, _run_command([cl, 'search', 'uuid=' + uuid1, '-u']))
+    def test_search_helper(ctx):
```
AndrewJGaut, line 1875: define a new helper function (so we can run the same tests multiple times)

codalab/model/bundle_model.py
```diff
@@ -634,19 +634,26 @@ def add_join(table: Table, condition: Any, left_outer_join: bool = False):
...
-                cl_bundle.c.uuid == cl_group_bundle_permission.c.object_uuid,
+                aliased_group_bundle_permission,
+                cl_bundle.c.uuid == aliased_group_bundle_permission.c.object_uuid,
+                left_outer_join=True,
+            )
+            aliased_cl_user_group = aliased(cl_user_group)
+            add_join(
+                aliased_cl_user_group,
+                aliased_cl_user_group.c.user_id == user_id,
+                left_outer_join=True,
```
AndrewJGaut, line 655: We use aliased here to account for the case when we've already done a join on the group_bundle_permission table or the user_group table. (If you do that twice without aliasing one of them, the query is invalid according to MySQL syntax rules.)

tests/cli/test_cli.py
```diff
@@ -1890,22 +1851,70 @@ def test_search(ctx):
...
+        _run_command([cl, 'rm', uuid3])  # need to remove since not on main worksheet
+
+        """Search bundles on private worksheets"""
+        wuuid = _run_command([cl, 'work', '-u'])
+        new_wuuid = _run_command([cl, 'new', random_name()])
+        _run_command([cl, 'wperm', new_wuuid, 'public', 'n'])  # make worksheet private
+        _run_command([cl, 'work', new_wuuid])  # switch to worksheet
+        uuid = _run_command([cl, 'upload', test_path('a.txt')])
+        check_equals(uuid, _run_command([cl, 'search', uuid, '-u']))
+        _run_command([cl, 'work', wuuid])
```
AndrewJGaut: this is the new test added for search on private worksheets

tests/cli/test_cli.py
```diff
@@ -1890,22 +1851,70 @@ def test_search(ctx):
...
+    # Test with non-root user.
+    if not os.getenv('CODALAB_PROTECTED_MODE'):
+        # This test does not work when protected_mode is True.
+        _, current_user_name = current_user()
+        user_name = 'non_root_user_' + random_name()
+        create_user(ctx, user_name, disk_quota='1000000')
+        switch_user(user_name)
+        new_wuuid = _run_command([cl, 'new', random_name()])
+        _run_command([cl, 'work', new_wuuid])
+        test_search_helper(ctx)
```
AndrewJGaut: Call helper function here to actually run the tests. Note we call the same helper for the root and non-root user, so the exact same tests are performed in both cases

codalab/model/bundle_model.py
yifanmai: Does `aliased` automatically pick a unique name, or do you need to explicitly pick a name?

In a different PR, we should probably add validation [when we do the joins](https://github.com/codalab/codalab-worksheets/blob/b0d285757036ebffc81c10a6dce245fc6759c21f/codalab/model/bundle_model.py#L656-L659) to make sure that we're not joining onto the same table twice.

tests/cli/test_cli.py
```diff
@@ -1890,22 +1851,70 @@ def test_search(ctx):
...
+            uuid2 + '\n' + uuid1, _run_command([cl, 'search', 'name=' + name, 'id=.sort-', '-u'])
+        )
+        check_equals('2', _run_command([cl, 'search', 'name=' + name, '.count']))
+        size1 = float(_run_command([cl, 'info', '-f', 'data_size', uuid1]))
+        size2 = float(_run_command([cl, 'info', '-f', 'data_size', uuid2]))
+        check_equals(
+            size1 + size2, float(_run_command([cl, 'search', 'name=' + name, 'data_size=.sum']))
+        )
+
+        """Search for floating bundles"""
```
yifanmai: nit: I think we use comment style `#` rather than docstrings for mid-method comments?

yifanmai (approved):
Great detective work!

codalab/model/bundle_model.py
```diff
@@ -633,19 +633,26 @@ def add_join(table: Table, condition: Any, left_outer_join: bool = False):
...
             )
-            add_join(cl_user_group, cl_user_group.c.user_id == user_id, left_outer_join=True)
             access_via_owner = cl_bundle.c.owner_id == user_id
             access_via_group = and_(
                 or_(  # Join constraint (group)
-                    cl_group_bundle_permission.c.group_uuid
+                    aliased_group_bundle_permission.c.group_uuid
```
yifanmai, line 660: Actually, is there a potential false positive case here (access_via_group is incorrectly true) if `aliased_group_bundle_permission.c.group_uuid` and `self.public_group_uuid` are both `None`?

yifanmai: Thinking about this a bit more, this logic seems incorrect?

- `aliased_group_bundle_permission` is all the group bundle permissions for the object
- `aliased_cl_user_group` is all the group memberships for the user
- `aliased_cl_user_group.c.user_id == user_id` doesn't do anything because this is already enforced by the join.
- So it seems like if the bundle grants access to some group A, and the user X belongs to some group B, then user X incorrectly gets access to the bundle?
- Seems like what we're missing here is we need to check that `aliased_group_bundle_permission.c.group_id == aliased_cl_user_group.c.group_id`?
- Alternatively you can push this constraint into the join condition i.e. when you in onto `aliased_group_bundle_permission`, join using the condition that the bundle permission group is the user's groups or the public group only?

yifanmai (changes requested):
So I understand, the failing case is when the bundle has no permissions, so the join is empty?

tests/cli/test_cli.py
AndrewJGaut: Aliased does automatically pick a unique name, so there's no issue with table name collisions or joining on the same table twice That's actually what I meant when I said the query was invalid otherwise; before I added aliased, sometimes it would join on the same table twice and MySQL actually throws an error when that happens. So we should be OK (if I'm understanding your comment correctly).

For the # comments style, I updated that.

codalab/model/bundle_model.py
AndrewJGaut: Yes, you're totally correct. Thanks for catching this! I updated it. I believe it should be correct now.

codalab/model/bundle_model.py
```diff
@@ -633,19 +633,27 @@ def add_join(table: Table, condition: Any, left_outer_join: bool = False):
...
                     == self.public_group_uuid,  # Public group
-                    cl_user_group.c.user_id == user_id,
+                    aliased_group_bundle_permission.c.group_uuid
+                    == aliased_cl_user_group.c.group_uuid,  # Private group
```
yifanmai, line 663: I believe you have to also check that `aliased_group_bundle_permission.c.group_uuid` is not NULL.

Otherwise this could be triggered if the user is not in any groups, and the bundles has no group permissions. Could you check if this is the case?

yifanmai (changes requested):

codalab/model/bundle_model.py
AndrewJGaut: OK, I added that check and some more tests to make sure that's not a problem anymore.

I think it looks good now?

PR #4435, opened 2023-04-05
Fix dev setup for now

### LINKED ISSUE #4429: Running codalab in dev mode doesn't work anymore
Running codalab in dev mode doesn't work anymore (`codalab-service start -bd`). It hangs at the end and then gives an error that the REST API isn't available. I can still run it in non-dev mode though (`codalab-service start -b`).

@wwwjn and I both faced this issue.

epicfaace:
Fixes https://github.com/codalab/codalab-worksheets/issues/4429 for now.


epicfaace:
Oops accidentally merged this without a review. But @wwwjn @AndrewJGaut let me know if you have any concerns / changes for this PR.

AndrewJGaut:
No worries, I would have approved this. It was not working for me either. We should add an issue to fix this, though; I'm not sure what happened.

PR #4440, opened 2023-04-07
Allow ".edu.cn" accounts to have more time and disk quota

wwwjn:

Allow ".edu.cn" accounts to have more time and disk quota

### Related issues

None

### Checklist

### DIFF
--- a/codalab/lib/spec_util.py
+++ b/codalab/lib/spec_util.py
@@ -23,6 +23,7 @@ HISTORY_RANGE_REGEX = re.compile(
 )  # Allow ranges foo^1-3 => foo^1 foo^2 foo^3
 BASIC_EMAIL_REGEX = re.compile(r'^[^@]+@[^@]+\.[^@]+$')
 SUB_PATH_REGEX = re.compile('^[a-zA-Z0-9_\-.]*\Z')
+EDU_USER_SUFFIX = [".edu", "edu.cn"]
 
 
 def expand_specs(specs):
@@ -106,3 +107,10 @@ def is_dashboard(name):
 
 def is_public_home(name):
     return name == 'home'
+
+
+def is_academic_email(email):
+    for s in EDU_USER_SUFFIX:
+        if email.endswith(s):
+            return True
+    return False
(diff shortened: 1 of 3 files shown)


codalab/model/bundle_model.py
```diff
@@ -2481,12 +2481,12 @@ def add_user(
 
             time_quota = (
                 self.default_user_info['edu_time_quota']
-                if email.endswith(".edu")
+                if email.endswith(".edu") or email.endswith(".edu.cn")
```
percyliang: Can we pull this out into a separate function `is_academic_email` and generalize to `.edu.*`?

codalab/lib/spec_util.py
```diff
@@ -106,3 +107,10 @@ def is_dashboard(name):
 
 def is_public_home(name):
     return name == 'home'
+
+
+def is_academic_email(email):
```
percyliang: This doesn't really belong in `spec_util`, which is the utility for specifications...you can just keep it in bundle_model for now.

Also, the logic we want is anything that ends with .edu or .edu.*, not just .edu.cn.

wwwjn: Thanks Percy! I'm really confused where to put this `is_academic_email ` function, and I will leave it in `bundle_model` for now.

codalab/model/bundle_model.py
```diff
@@ -68,6 +68,7 @@
 
 SEARCH_KEYWORD_REGEX = re.compile('^([\.\w/]*)=(.*)$')
 SEARCH_RESULTS_LIMIT = 10
+EDU_USER_REGEXS = re.compile('@[\w\.-]+\.(edu|edu\.[a-z]{2})$')
```
epicfaace: change to REGEXES

codalab/model/bundle_model.py
```diff
@@ -2475,18 +2476,27 @@ def add_user(
...
+            return len(email_suffix) > 0
+
         with self.engine.begin() as connection:
             now = datetime.datetime.utcnow()
             user_id = user_id or '0x%s' % uuid4().hex
 
             time_quota = (
                 self.default_user_info['edu_time_quota']
-                if email.endswith(".edu")
+                if is_academic_email(email)
```
epicfaace, line 2495: Can we add some unit tests for is_academic_email?

wwwjn: Thanks Ashwin! For consistency, I add this unit test to bundle_model_test for now.

epicfaace (approved):

epicfaace (changes requested):
We should also include ".ac.xx" domains -- see https://en.wikipedia.org/wiki/.ac_(second-level_domain)

wwwjn:
> We should also include ".ac.xx" domains -- see https://en.wikipedia.org/wiki/.ac_(second-level_domain)

That's a good catch!

epicfaace (approved):

PR #4370, opened 2023-02-01
Generate index without downloading the whole file from blob storage

wwwjn:
### Related issues
#4201 

### Checklist

* [ ] I've added a screenshot of the changes, if this is a frontend change
* [ ] I've added and/or updated tests, if this is a backend change
* [ ] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [ ] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

Issue #4201, opened 2022-08-10
Bypass server upload large file is slow because of generating index file


wwwjn:
According to #3370 and [test result](https://docs.google.com/document/d/1qFT6OWSUW5-2ZfnkK_q6P-Q-9sVe_fyBLX3UfEIWRiY/edit?pli=1#heading=h.nzsd7jt9zuoh), bypass server upload need a lot of time to calculate the index file.

epicfaace:
Ashwin made a POC here: https://github.com/codalab/codalab-worksheets/pull/4212

But we depend on fixing https://github.com/pauldmccarthy/indexed_gzip/issues/102 and https://github.com/mxmlnkn/ratarmount/issues/95 before it can work.

epicfaace:
Ashwin: get https://github.com/pauldmccarthy/indexed_gzip/pull/103 to work, Ashwin will also publish a fork of indexed_gzip (`indexed_gzip_fileobj_fork_epicfaace`)


Jiani: fixing up https://github.com/codalab/codalab-worksheets/pull/4212. Maybe try reverting to the simpler `peek` solution in https://github.com/codalab/codalab-worksheets/pull/4212/commits/ac5ed66cc35654537bfa2ae5b9bdfb8307f4409c. Once Ashwin publishes indexed_gzip fork, change it to use the indexed_gzip fork


```python
import indexed_gzip_fileobj_fork_epicfaace as indexed_gzip
```

wwwjn:
POC to fix this problem is here: [Using peek() instead of read() in SQLiteIndexedTar for gz files](https://github.com/codalab/codalab-worksheets/pull/4212/commits/ac5ed66cc35654537bfa2ae5b9bdfb8307f4409c#diff-ad5ad76eb55b6437b2e1aa24b324c6d11d176be1926ebea1950a006bfce4efbe) 

However, the current Indexed_gzip implementation will not call peek(), instead it calls read(). Simple script to reproduce this:  https://github.com/epicfaace/test-repro/pull/12

wwwjn:
There still some bugs in indexed_gzip, which blocks this issue: https://github.com/pauldmccarthy/indexed_gzip/pull/106

wwwjn:
Update: Related issue for indexed_gizp github repo: https://github.com/pauldmccarthy/indexed_gzip/issues/107

leilenah:
will sync with ashwin

epicfaace:
Potential solution:
[...]

PR #4444, opened 2023-04-18
Generate index without downloading the whole file from blob storage

wwwjn:

Similar to PR #4370 , but the previous PR was closed mistakenly, so I create this new one.

### Related issues
#4201 

### Checklist


epicfaace (approved):
Please remove the files in bin/__pycache__

PR #4451, opened 2023-04-23
Fix/4431 migrate communication to use websockets only

AndrewJGaut:
Still a WIP. Want to run CI to see what happens.

Still have to fix formatting and fix the code up a bit as well!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.8]
         os: [ubuntu-20.04, macos-latest]
         exclude:
           - os: macos-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8]
         os: [ubuntu-20.04, macos-latest]
         exclude:
           - os: macos-latest
--- a/codalab/bin/server.py
+++ b/codalab/bin/server.py
@@ -28,9 +28,9 @@ class ClFileWatcherEventHandler(FileSystemEventHandler):
         extensions_to_watch = ('.js', '.py', '.html', '.css', '.tpl')
         file_extension = os.path.splitext(event.src_path)[1]
 
-        if file_extension in extensions_to_watch:
-            print("Saw file change: %s -- restarting!" % (os.path.basename(event.src_path)))
-            self.restart()
+        # if file_extension in extensions_to_watch:
+        #     print("Saw file change: %s -- restarting!" % (os.path.basename(event.src_path)))
+        #     self.restart()
 
 
 def run_server_with_watch():
--- a/codalab/bin/ws_server.py
+++ b/codalab/bin/ws_server.py
@@ -1,75 +1,145 @@
+# This is the real ws-server, basically.
 # Main entry point for CodaLab cl-ws-server.
 import argparse
 import asyncio
+from collections import defaultdict
 import logging
 import re
-from typing import Any, Dict
+from typing import Any, Dict, List
 import websockets
+from dataclasses import dataclass
+import threading
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.WARNING)
 logging.basicConfig(format='%(asctime)s %(message)s %(pathname)s %(lineno)d')
 
-worker_to_ws: Dict[str, Any] = {}
+"""
+TODO!!! WE NEED TO ADD A SECRET OR SOME SORT OF AUTH FOR THE SERVER ENDPOITNS
+Otherwise, people could create a custom local worker build and wreak havoc on the ws server...
+"""
 
-
-async def rest_server_handler(websocket):
-    """Handles routes of the form: /main. This route is called by the rest-server
-    whenever a worker needs to be pinged (to ask it to check in). The body of the
-    message is the worker id to ping. This function sends a message to the worker
-    with that worker id through an appropriate websocket.
+@dataclass
+class WS:
     """
-    # Got a message from the rest server.
-    worker_id = await websocket.recv()
-    logger.warning(f"Got a message from the rest server, to ping worker: {worker_id}.")
+    Stores websocket object and whether or not the websocket is available.
+    TODO: give this a better, less confusing name.
+    """
+    _ws: Any = None
+    _is_available: bool = True
 
-    try:
-        worker_ws = worker_to_ws[worker_id]
-        await worker_ws.send(worker_id)
-    except KeyError:
-        logger.error(f"Websocket not found for worker: {worker_id}")
+    @property
+    def ws(self):
+        return self._ws
+    
+    @property
+    def is_available(self):
+        return self._is_available
+    
+    @is_available.setter
+    def is_available(self, value):
+        self._is_available = value
 
+worker_to_ws: Dict[str, Dict[str, WS]] = defaultdict(dict)  # Maps worker to list of its websockets (since each worker has a pool of connections)
+server_ws_to_worker_ws: Dict[Any, Any] = dict()  # Map the rest-server websocket connection to the corresponding worker socket connection.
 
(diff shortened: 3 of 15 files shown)


AndrewJGaut:
The client is getting a ConnectionClosed exception before it receives anything when hitting the connection endpoint sometimes. Moreover, we receive 1011, send 1011, and we receive the 1011 _before_ sending it (see the ConnectionClosed exception [here](https://github.com/python-websockets/websockets/blob/9c578a1f5a5e20b8901ac75f12a4dfec4877f6d2/src/websockets/exceptions.py#L90)).

I think it's _possible_ that arises from the following since these are 1011 codes sent without reasons (and elsewhere in the code, 1011 is sent with a reason):
- https://github.com/python-websockets/websockets/blob/9c578a1f5a5e20b8901ac75f12a4dfec4877f6d2/src/websockets/protocol.py#L570
- https://github.com/python-websockets/websockets/blob/9c578a1f5a5e20b8901ac75f12a4dfec4877f6d2/src/websockets/sync/server.py#L500
- https://github.com/python-websockets/websockets/blob/9c578a1f5a5e20b8901ac75f12a4dfec4877f6d2/src/websockets/legacy/protocol.py#L684 (and a few others in this file)
- https://github.com/python-websockets/websockets/blob/9c578a1f5a5e20b8901ac75f12a4dfec4877f6d2/src/websockets/legacy/server.py#L240

It is weird b/c each of these logs something before the 1011, but it doesn't show up in the logs I see...

[This](https://github.com/python-websockets/websockets/blob/9c578a1f5a5e20b8901ac75f12a4dfec4877f6d2/src/websockets/sync/server.py#L500) one would be a hander error. Perhaps the logging it does above just isn't logged since it's the protocol logger? Perhaps the handler really is failing?

AndrewJGaut:
A few weird things:

- We're getting the 1011 error from the _connect call from rest-server/bundle-manager (specifically when they try to recv the socket_id), despite the fact that NO request is ever logged on the ws-server side when this occurs. 
Notably, the error looks like such: 
```
> /opt/codalab-worksheets/codalab/model/worker_model.py(239)_connect()
-> logger.error(f"Socket id receive failed with {e}")
(Pdb) p e
ConnectionClosedError(Close(code=1011, reason=''), Close(code=1011, reason=''), True)
(Pdb) p vars(e)
{'rcvd': Close(code=1011, reason=''), 'sent': Close(code=1011, reason=''), 'rcvd_then_sent': True}
```
Indicating, in particular, that the 1011 was received by the client and then sent to the server (so, in particular, the client believes the server sent a 1011), but that's never logged on the server side.

- Some of the worker threads fail to connect (the ws-server logs the worker as having fewer connections than it the worker has threads). However, when I look at the worker logs for these threads, _every_ thread believes it has a connection and is within the `async with connect(<ws-server working endpoint>):` portion of the code calling recv_msg. So, a connection must have been established, but the ws-server somehow doesn't register all of the connections. It makes no sense...

AndrewJGaut:
OK, here's a new discovery: looks like if I just shut off the ws-server entirely and call _connect, I still get the same `ConnectionClosedError(Close(code=1011, reason=''), Close(code=1011, reason=''), True)`. It looks like `rcvd_then_sent` just isn't accurate and this library has REALLY poor error reporting... Why? Because there should be an error ON THE CONNECTION, THEN! I mean, look at the docs [here](https://websockets.readthedocs.io/en/stable/reference/sync/client.html)... it's very clear that either an `OSError` or `InvalidHandshake` should occur... I have no idea why it doesn't!

Regardless, this is progress. When this error occurs, given that the ws-server logs nothing, it looks like the request just isn't even reaching the ws-server.


Hmmm... Another weird thing. If I just `docker stop` the `ws-server` container, then the `websockets.connect(ws://ws-server:2091)` call works and then `websocket.recv()` throws the usual `1011` error. However, if I try `websockets.connect(ws://ws-server:2000)` (or any port besides `2901`, then I get a `ConnectionRefusedError: [Errno 111] Connection refused`. Since the container is stopped, shouldn't _any_ of these calls (including to port 2901) produce a `ConnectionRefusedError`?

AndrewJGaut:
Another issue: Dual receives.

Basically, the worker thread with ID _i_ sends information through its socket and then that thread loops back around to listen on the socket once again. The problem is it receives that message before the rest-server does and tries to process the rest-server's message! See the logs below for an example.

```
2023-05-13 23:49:57,627 Thread 1 processing messsage: b'{"type": "netcat", "socket_id": "1", "uuid": "0x921c10aac6c2481aafcc0fd9bf1fbf30", "port": 5005, "message": "hi patrick"}' /opt/codalab/worker/worker.py 355
2023-05-13 23:49:57,965 Received netcat message: {'type': 'netcat', 'socket_id': '1', 'uuid': '0x921c10aac6c2481aafcc0fd9bf1fbf30', 'port': 5005, 'message': 'hi patrick'} /opt/codalab/worker/worker.py 317
2023-05-13 23:49:57,970 @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ /opt/codalab/worker/bundle_service_client.py 139
2023-05-13 23:49:57,971 SENDING REPLY DATA!!!!!!!!!!!!! /opt/codalab/worker/bundle_service_client.py 140
2023-05-13 23:49:57,986 Thread 1 processing messsage: b'{}' /opt/codalab/worker/worker.py 355
2023-05-13 23:50:02,059 Finished run with UUID 0x921c10aac6c2481aafcc0fd9bf1fbf30, container_id 6a906230aca8b36a0eff10822e2069d59e2a76fd383546a23ac8d1c37b4a5a31, exitcode 0, failure_message None /opt/codalab/worker/worker_run_state.py 582
2023-05-13 23:50:03,964 Thread 1 processing messsage: b'No, this is dawg' /opt/codalab/worker/worker.py 355
```


Note: this issue would be alleviated if the `worker` communicated with the `rest-server` through websockets _directly_ rather than through the `ws-server`. Ugh...

AndrewJGaut:
```
2023-05-14 05:40:53,418 Just sent netcat message, about to read the response... /opt/codalab-worksheets/codalab/lib/download_manager.py 310
2023-05-14 05:40:53,418 recving header message /opt/codalab-worksheets/codalab/lib/download_manager.py 388
2023-05-14 05:40:53,419 <generator object WorkerModel.recv at 0x7f36ddb3cf48> /opt/codalab-worksheets/codalab/lib/download_manager.py 390
2023-05-14 05:40:53,419 got header message /opt/codalab-worksheets/codalab/lib/download_manager.py 391
Recv from worker DNa1c477f.SUNet through socket 1 failed with received 1011 (unexpected error); then sent 1011 (unexpected error)
2023-05-14 05:41:28,853 Recv from worker DNa1c477f.SUNet through socket 1 failed with received 1011 (unexpected error); then sent 1011 (unexpected error) /opt/codalab-worksheets/codalab/model/worker_model.py 321
2023-05-14 05:41:28,854 About to get fileobj /opt/codalab-worksheets/codalab/lib/download_manager.py 399
2023-05-14 05:41:28,854 Got fileobj /opt/codalab-worksheets/codalab/lib/download_manager.py 406
2023-05-14 05:41:28,855 about to read fileobjs /opt/codalab-worksheets/codalab/lib/download_manager.py 416
In recv STREAM
2023-05-14 05:41:28,861 In recv STREAM /opt/codalab-worksheets/codalab/model/worker_model.py 313
2023-05-14 05:41:28,862 @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ /opt/codalab-worksheets/codalab/lib/download_manager.py 312
2023-05-14 05:41:28,862 b'' /opt/codalab-worksheets/codalab/lib/download_manager.py 313
```

AndrewJGaut:
> Another issue: Dual receives.

Hold on: r.e. dual receives. Really, what we need is for send to not complete until the recipient has received the data. Period. So, we'll need something for either the websocket server to send back or some other way to make sure the worker_model send call does not complete until all data is received by the rest-server (or whoever you're `exchange`'-ing with on the websocket server)

AndrewJGaut:
Closing this in favor of #4490. This was causing too many issues, so we simplified things by JUST updating the JSON messaging mechanism.

PR #4452, opened 2023-04-25
Fix CI - upgrade apache beam to version 2.38.0

wwwjn:

CI broken with python version 3.9.0 on ubuntu-20.04. We need to update the version of apache-beam to fix this. Considering python 3.6 only have apache-beam=2.38.0 release, so I choose this version.

### Checklist


epicfaace (approved):
Thanks, nice fix 🥳

Issue #4433, opened 2023-03-31
Tests are extremely flaky


AndrewJGaut:
The CI tests are very flaky. Often, tests will fail and then pass when they are re-run.

AndrewJGaut:
After looking through some CI failures that occurred for commits to master (i.e. after the CI tests had already passed for that PR and still failed), I've identified the following tests that failed (and therefore are flaky):
-	Kill
o	Examples: [here](https://github.com/codalab/codalab-worksheets/actions/runs/4495011876/jobs/7908295990) 
o	Status: Unsure about problem (see below comment)
-	run
o	Examples: [shared fs](https://github.com/codalab/codalab-worksheets/actions/runs/4493408758/jobs/7904750706) and [Azure blob](https://github.com/codalab/codalab-worksheets/actions/runs/4431421114/jobs/7774483554)
o	Status: Same as kill (see below comment)
-	Mimic
o	Examples: [here](https://github.com/codalab/codalab-worksheets/actions/runs/4747386611/jobs/8433152532) and [here](https://github.com/codalab/codalab-worksheets/actions/runs/4619214014/jobs/8167843995) 
o	Status: PR ready for review.
-	Make
o	Examples: [here](https://github.com/codalab/codalab-worksheets/actions/runs/4623312902/jobs/8177104622)
o	Status: Found the problem. Want to chat about the solution.
-	Resources
o	Examples: [here](https://github.com/codalab/codalab-worksheets/actions/runs/4493408758/jobs/7904748770) and [here](https://github.com/codalab/codalab-worksheets/actions/runs/4305932181/jobs/7509225119)
o	Status: Unsure about problem.
-	General
o	Setup Tests: Sometimes we get setup tests failures. Not sure what causes this... e.g. [here](https://github.com/codalab/codalab-worksheets/actions/runs/4308206298/jobs/7514912278) 



Will now start looking through each of these in a PR...

AndrewJGaut:
**Kill/run**

First, note that the `run` tests that fail are due to failures in `kill`; see the examples above to see why.

This one perplexes me.

So, in the worker logs for the sharedFS run that failed [here](https://github.com/codalab/codalab-worksheets/actions/runs/4493408758/jobs/7904750706), we see the following:
```
[...]

PR #4454, opened 2023-04-27
Fix Flaky CI - Mimic

AndrewJGaut:
Addresses one part of issue #4433 by addressing the flakiness of the Mimic test.

What happened was, after we removed the data hash, we had to write code in the test_cli.py file to produce something equivalent for tests that relied on having a check that the contents of two bundles were the same. To do this, we download the bundle and then hash the files in the directories in a function in the test_cli.py file.

However, that function did not `wait` on the uuid of the bundle being data hashed, and so sometimes the downloaded contents wouldn't match the contents of the bundle being mimicked, thereby causing occasional failures.

(Note: This is still a hypothesis. I could not reproduce the issue locally but could on CI, but I have re-run the mimic CI test many times after this change and it has yet to fail.)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,6 +32,7 @@ selenium==3.141.0
 requests==2.25.0
 azure-storage-blob==12.4.0
 azure-core==1.8.0
+msrest==0.6.21
 sentry-sdk==1.11.1
 requests-oauthlib==1.1.0
 oauthlib==2.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,7 +32,6 @@ selenium==3.141.0
 requests==2.25.0
 azure-storage-blob==12.4.0
 azure-core==1.8.0
-msrest==0.6.21
 sentry-sdk==1.11.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -32,6 +32,7 @@ selenium==3.141.0
 requests==2.25.0
 azure-storage-blob==12.4.0
 azure-core==1.8.0
+msrest==0.6.21
```
AndrewJGaut: Note: this is a development issue I just discovered b/c I'm working on a new laptop. We need to pin msrest to 0.6.21 to get it to work with our version of azure-core. (see [here](https://github.com/Azure/azure-sdk-for-python/issues/24765)). We might want to add this to a different PR, though.

AndrewJGaut: Moved to new PR.

AndrewJGaut:
I've rerun the test several times and it seems to pass every time, so I believe the issue is fixed.

epicfaace (approved):
Nice fix!

Issue #4472, opened 2023-05-17
Disk usage not working in prod


AndrewJGaut:
Disk usage is not working in production; users' disk is not being accurately computed when deleting/creating bundles. for instance, see [here](https://github.com/codalab/codalab-worksheets/issues/4469).

It would appear the changes I made in [this PR](https://github.com/codalab/codalab-worksheets/pull/4354) are not working properly in production.

Luckily, I can revert back to the old method of computing disk quota and it will recompute their quota (see [here](https://github.com/codalab/codalab-worksheets/blob/7fc08ccda2d0c6a0208be8c327efbedf3ae7a53d/codalab/model/bundle_model.py#L1132-L1156) and, in particular, `update_user_disk_used` [here](https://github.com/codalab/codalab-worksheets/blob/7fc08ccda2d0c6a0208be8c327efbedf3ae7a53d/codalab/model/bundle_model.py#L2776-L2780), which uses a query [here](https://github.com/codalab/codalab-worksheets/blob/7fc08ccda2d0c6a0208be8c327efbedf3ae7a53d/codalab/model/bundle_model.py#L2762-L2769) to just recompute the user disk quota from scratch).

In short, I can effectively revert this change for the time being and the user's disk usage will automatically be reset to their true usage whenever they next make any changes that affect disk quota (e.g. upload a bundle or delete a bundle). I will investigate why this change passed the test and seems to work sometimes (e.g. when I tested it in dev) but not at other times. I suspect there may be a race condition somewhere.

PR #4473, opened 2023-05-17
Fix disk usage bug

AndrewJGaut:
Fix for #4472. This just reverts to using disk computation from scratch while I look for the issue with the prior change.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1009,9 +1009,11 @@ class BundleModel(object):
             'time_system': worker_run.container_time_system,
             'remote': worker_run.remote,
             'cpu_usage': cpu_usage,
-            'memory_usage': memory_usage,
-            'data_size': worker_run.disk_utilization,
+            'memory_usage': memory_usage
         }
+        if self.get_bundle_state(bundle.uuid) != State.FAILED:
+            # If the bundle state is failed, it means it failed on uploading_results and data_size was wiped.
+            metadata_update['data_size'] = worker_run.disk_utilization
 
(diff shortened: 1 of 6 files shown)


codalab/model/bundle_model.py
```diff
@@ -1010,8 +1010,10 @@ def transition_bundle_running(self, bundle, worker_run, row, user_id, worker_id,
             'remote': worker_run.remote,
             'cpu_usage': cpu_usage,
             'memory_usage': memory_usage,
-            'data_size': worker_run.disk_utilization,
         }
+        if self.get_bundle_state(bundle.uuid) != State.FAILED:
```
AndrewJGaut, line 1014: Note: This was another bug that occurred, but this time due to the [PR that tracks disk usage while a bundle is running](https://github.com/codalab/codalab-worksheets/pull/4307). Basically, in `bundles.py`, the bundle is failed during the `UPLOADING_RESULTS` phase [here](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/bundles.py#L1209-L1219) if the user doesn't have enough disk quota to upload the results. When this would happen, the worker isn't updated on the failure status, so it just goes and hits the bundle checkin and finalizes the bundle, and in `transition_bundle_finished` the bundle is marked as `FAILED` because it has a `failure_message` in its metdata from the code in bundles.py. But, when that bundle checkin occurs, the data size is updated again which is WRONG because the actual size of the bundle ON THE SERVER (or Azure blob) is zero even though on the worker it still has a real size (until it gets deleted later)

epicfaace (approved):

Issue #4478, opened 2023-05-29
Update PyPI process to use 2FA


epicfaace:
Given this announcement: https://blog.pypi.org/posts/2023-05-25-securing-pypi-with-2fa/

PR #4482, opened 2023-05-31
Upload to PyPI using trusted publishing

epicfaace:
For https://github.com/codalab/codalab-worksheets/issues/4478. Uses trusted publishers to upload to PyPI without needing credentials. Essentially, only the release.yml file on this github repo will have access to publish to PyPI. I configured it using instructions at https://docs.pypi.org/trusted-publishers/adding-a-publisher/ and https://docs.pypi.org/trusted-publishers/using-a-publisher/.

After we get this in, we should test it out for the next release, and once everything works, then enable 2FA on our PyPI account and delete TWINE_USERNAME and TWINE_PASSWORD from this repository's secrets.


epicfaace:
@AndrewJGaut @wwwjn  can you review?

AndrewJGaut (approved):

Issue #4431, opened 2023-03-22
Migrate websocket server to send json messages


AndrewJGaut:
send json messagse through websocket instead of having it tell worker to checkin again.

AndrewJGaut:
Note: I'll be producing a document on this in the coming days so we can discuss before our meeting this next week.

percyliang:
I assume we can close this?

PR #4490, opened 2023-06-14
Use websockets for json communication

AndrewJGaut:
This PR addresses #4431. It modifies JSON messaging between the REST server / bundle manager and workers.

An in-depth document describing these changes and their motivation is given [here](https://docs.google.com/document/d/1VkrkbAr_2mUAFRYT06RX0DdwK3O-zESCvO4MLMCh90w/edit?usp=sharing).

## Summary
Previously, messages were sent to workers in the HTTP response to worker checkins. However, this was inefficient; a server thread would try to send the message through an AF Unix socket corresponding to the worker using TCP, and so if the sending and receiving threads were not synchronized, the message would not be sent. The need for synchronization made server to worker communication highly inefficient. To make this slightly more efficient, a websocket server was added and would send a ping to the worker meant to receive the message to cause it to checkin immediately so that the receiving server thread would start listening as soon as possible on the AF Unix socket to receive the message. However, desynchronization and corresponding inefficiency persisted.

After this PR, JSON messages will be sent to workers immediately through the websocket server. This means that:
(1) Workers will no longer receive any messages through HTTP responses to the checkin.
(2) No synchronization is required between server threads for JSON messaging.
(3) The websocket server no longer uses the ping functionality.

We also add in authentication of workers and server to the websocket server. (The need for this is described [here](https://docs.google.com/document/d/14Ws2hpFmvNRlE-Lf9xa48Dh3PRFx82DE4NJZywlVUAs/edit#heading=h.8qfic7lguyks).)

## Speedup
When looking at the [`time` tests](https://github.com/codalab/codalab-worksheets/blob/master/tests/cli/test_cli.py#L2146), the only test that uses worker communication is the [`run` test](https://github.com/codalab/codalab-worksheets/blob/master/tests/cli/test_cli.py#L2165-L2170). For that test, before the websockets change, it would take over 11 seconds ([11.86 seconds here](https://github.com/codalab/codalab-worksheets/actions/runs/5405622761/jobs/10105043645), [11.44 seconds here](https://github.com/codalab/codalab-worksheets/actions/runs/5270447665/jobs/9530215242)). After the websockets change, we get [5.57 seconds](https://github.com/codalab/codalab-worksheets/actions/runs/5570884432/jobs/10175443323). That's a speedup of over 2x! (To see the time in those links, click on `Run tests using Docker runtime` and then search `cl info -f name` and it will be the number right above the line that comes up.)

## Future Work / TODO
I might do these in a future PR:
- Only release lock once worker is done with whatever it's doing
- More fine-grained locking for the worker (lock per bundle uuid)
- Dynamic resizing of number of websockets for worker depending on 'demand'


codalab/bin/ws_server.py
```diff
@@ -1,75 +1,204 @@
...
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.WARNING)
+logger.setLevel(logging.INFO)
 logging.basicConfig(format='%(asctime)s %(message)s %(pathname)s %(lineno)d')
 
-worker_to_ws: Dict[str, Any] = {}
 
+"""
+TODO!!! WE NEED TO ADD A SECRET OR SOME SORT OF AUTH FOR THE WS SERVER ENDPOITNS
```
epicfaace: https://websockets.readthedocs.io/en/stable/topics/authentication.html

epicfaace: Auth, we have two options:
1. merge the rest server and websocket server: either A) switch away from bottle, use flask, or B) maybe just run websocket in a separate process
2. set up some auth for websocket server (maybe it sends a secret initially, or query rest server -> DB to check credentials)

epicfaace:
"working quite well now" 👀 👀

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,143 @@
...
+from codalab.lib.codalab_manager import CodaLabManager
 
 
-async def rest_server_handler(websocket):
-    """Handles routes of the form: /main. This route is called by the rest-server
-    whenever a worker needs to be pinged (to ask it to check in). The body of the
-    message is the worker id to ping. This function sends a message to the worker
-    with that worker id through an appropriate websocket.
+@dataclass
+class WS:
```
AndrewJGaut: I might want to give this a better name. It basically just stores a lock associated with the websocket now (to indicate that that socket is already in use). Previously, it did more.

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,143 @@
...
-    """Handles routes of the form: /worker/{id}. This route is called when
-    a worker first connects to the ws-server, creating a connection that can
-    be used to ask the worker to check-in later.
+
+worker_to_ws: Dict[str, Dict[str, WS]] = defaultdict(
+    dict
+)  # Maps worker to list of its websockets (since each worker has a pool of connections)
+ACK = b'a'
+logger = logging.getLogger(__name__)
+manager = CodaLabManager()
```
AndrewJGaut, line 55: CodaLabManager is used to give access to the SQL database to websocket server.

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,143 @@
...
+bundle_model = manager.model()
+worker_model = manager.worker_model()
+server_secret = os.environ["CODALAB_SERVER_SECRET"]
```
AndrewJGaut: We use a server secret (some cryptographically secure string) that would be an environment variable added to the deployment repo

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,143 @@
...
+async def worker_handler(websocket: Any, worker_id: str, socket_id: str) -> None:
+    """Handles routes of the form: /worker/{worker_id}/{socket_id}.
+    This route is called when a worker first connects to the ws-server, creating
+    a connection that can be used to ask the worker to check-in later.
+    """
+    # Authenticate worker.
+    access_token = await websocket.recv()
+    user_id = worker_model.get_user_id_for_worker(worker_id=worker_id)
+    authenticated = bundle_model.access_token_exists_for_user(
+        'codalab_worker_client', user_id, access_token  # TODO: Avoid hard-coding this if possible.
```
AndrewJGaut, line 104: Maybe I should avoid hard-coding the `client-id` to be `codalab_worker_client`?

codalab/model/worker_model.py
```diff
@@ -35,10 +37,13 @@ class WorkerModel(object):
        listen on these sockets for messages and send messages to these sockets.
     """
 
+    ACK = b'a'
+
     def __init__(self, engine, socket_dir, ws_server):
         self._engine = engine
         self._socket_dir = socket_dir
         self._ws_server = ws_server
+        self._server_secret = os.environ["CODALAB_SERVER_SECRET"]
```
AndrewJGaut: Maybe I shouldn't access an environment variable here? Is there a better way to inject the server secret?

codalab/model/worker_model.py
```diff
@@ -451,3 +456,36 @@ def has_reply_permission(self, user_id, worker_id, socket_id):
             if row:
                 return True
             return False
+
+    def send_json(self, data: dict, worker_id: str, timeout_secs: int = 60):
```
AndrewJGaut: The new function that sends data through websockets.

The other functions (send_json_message_with_sock, send_stream) send data through the AF Unix sockets. (We still need those since file data is still send through AF Unix sockets and both those functions are used for file data. send_json_message_with_sock is also required for files since there's a [header message](https://github.com/codalab/codalab-worksheets/blob/master/codalab/rest/workers.py#L118-L128) pertaining to the file.)

codalab/worker/worker.py
```diff
@@ -90,6 +92,8 @@ def __init__(
         exit_on_exception=False,  # type: bool
         shared_memory_size_gb=1,  # type: int
         preemptible=False,  # type: bool
+        num_threads=10,  # type: int
```
AndrewJGaut: Maybe I should make this an environment variable and a command line variable in worker/main.py and inject it in the docker-compose?

AndrewJGaut:
Converted back to draft because of an intermittent issue I discovered in dev deployment that didn't pop up locally or in tests. Basically, occasionally the `bundle-manager` will get a `Send to worker <worker> failed with received 1000 (OK); then sent 1000 (OK). /opt/codalab-worksheets/codalab/model/worker_model.py 482` error when it tries to start a bundle and will continue to get that error as it keeps retrying and the bundle never starts.

Update: Moving back out of draft. Issue resolved. Basically, there was a bug where the `ws-server` didn't properly detect when a worker thread dropped its connection. In dev, workers start and stop, and so after the worker would stop and then a user tried to a run a new bundle and the worker was spun up again, the `ws-server` would no longer properly work since it would try to use a stale, no longer in use websocket.

One more update and general websocket programming advise: I made sure that I would never get `failed with received 1000 (OK); then sent 1000 (OK)` again by explicitly closing the server's websocket connection with an error code if an error were ever reached. Before, I would just let the websocket go out of scope, which causes the socket to close with a 1000 (OK) code. In general I now realize, you should NEVER do this since it makes it difficult to track down what went wrong. Always explicitly close the connection with an error code if anything goes wrong.

codalab/model/worker_model.py
```diff
@@ -451,3 +456,32 @@ def has_reply_permission(self, user_id, worker_id, socket_id):
             if row:
                 return True
             return False
+
+    def send_json(
```
epicfaace: Maybe rename to `send_json_message` for consistency with `send_json_message_with_sock`

codalab/model/worker_model.py
```diff
@@ -316,65 +362,24 @@ def get_stream(self, sock, timeout_secs):
...
-                await websocket.send(worker_id)
-
-        futures = [ping_ws()]
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
-        loop.run_until_complete(asyncio.wait(futures))
-        logging.warn(f"Pinged worker through websockets, worker id: {worker_id}")
-
-    def send_json_message(self, socket_id, worker_id, message, timeout_secs, autoretry=True):
+    def send_json_message_with_sock(
```
epicfaace: Keep the same order as before so we minimize diffs

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,148 @@
...
+ACK = b'a'
+logger = logging.getLogger(__name__)
+manager = CodaLabManager()
+bundle_model = manager.model()
+worker_model = manager.worker_model()
+server_secret = os.getenv("CODALAB_SERVER_SECRET")
+
+
+async def send_handler(server_websocket, worker_id):
+    """Handles routes of the form: /send/{worker_id}. This route is called by
```
epicfaace: maybe rename to `to_worker` or `send_to_worker` or `send_from_server`

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,148 @@
...
+    # Check if any websockets available
+    if worker_id not in worker_to_ws or len(worker_to_ws[worker_id]) == 0:
+        logger.warning(f"No websockets currently available for worker {worker_id}")
+        await server_websocket.close(
+            1013, f"No websockets currently available for worker {worker_id}"
+        )
+        return
+
+    # Authenticate server.
+    receieved_secret = await server_websocket.recv()
```
epicfaace: typo -- rename to `received_secret`

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,148 @@
...
+    ):
+        if ws_to_lock[worker_id][socket_id].acquire(blocking=False):
+            data = await server_websocket.recv()
+            await worker_websocket.send(data)
+            await server_websocket.send(ACK)
+            ws_to_lock[worker_id][socket_id].release()
+            return
+
+    logger.warning(f"All websockets for worker {worker_id} are currently busy.")
+    await server_websocket.close(1013, f"All websockets for worker {worker_id} are currently busy.")
```
epicfaace: what is 1013?

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,148 @@
...
+
+
+async def worker_handler(websocket: Any, worker_id: str, socket_id: str) -> None:
+    """Handles routes of the form: /worker/{worker_id}/{socket_id}.
```
epicfaace: maybe rename to `from_worker`?

codalab/lib/codalab_manager.py
```diff
@@ -248,6 +248,11 @@ def ws_server(self):
         ws_port = self.config['ws-server']['ws_port']
         return f"ws://ws-server:{ws_port}"
 
+    @property  # type: ignore
+    @cached
+    def server_secret(self):
+        return os.getenv("CODALAB_SERVER_SECRET")
```
epicfaace, line 254: should this be just an env variable? I think we should instead add it as a CodalabArg like the other CODALAB_ env vars

codalab/lib/download_manager.py
```diff
@@ -378,21 +376,19 @@ def _send_netcat_message(self, worker, response_socket_id, uuid, port, message):
...
-        if not self._worker_model.send_json_message(
-            worker['socket_id'], worker['worker_id'], message, 60
-        ):  # dead workers are a fact of life now
+        if not self._worker_model.send_json(message, worker['worker_id']):
             logging.info('Unable to reach worker')
 
     def _get_read_response_stream(self, response_socket_id):
         with closing(self._worker_model.start_listening(response_socket_id)) as sock:
-            header_message = self._worker_model.get_json_message(sock, 60)
+            header_message = self._worker_model.recv_json_message_with_sock(sock, 60)
```
epicfaace: rename to `get_json_message`?

codalab/lib/download_manager.py
```diff
@@ -378,21 +376,19 @@ def _send_netcat_message(self, worker, response_socket_id, uuid, port, message):
             'port': port,
             'message': message,
         }
-        if not self._worker_model.send_json_message(
-            worker['socket_id'], worker['worker_id'], message, 60
-        ):  # dead workers are a fact of life now
+        if not self._worker_model.send_json(message, worker['worker_id']):
```
epicfaace: rename to `send_json_message`?

codalab/model/bundle_model.py
```diff
@@ -2901,6 +2901,25 @@ def find_oauth2_token(self, client_id, user_id, expires_after):
 
         return OAuth2Token(self, **row)
 
+    def access_token_exists_for_user(self, client_id: str, user_id: str, access_token: str) -> bool:
```
epicfaace, line 2903: rename to `check_access_token`?

codalab/worker/main.py
```diff
@@ -116,6 +116,12 @@ def parse_args():
         default=None,
         help='Limit the amount of memory to a worker in bytes' '(e.g. 3, 3k, 3m, 3g, 3t).',
     )
+    parser.add_argument(
+        '--num-threads',
```
epicfaace: Rename to `num_coroutines`

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,148 @@
...
-    """Handles routes of the form: /worker/{id}. This route is called when
-    a worker first connects to the ws-server, creating a connection that can
-    be used to ask the worker to check-in later.
-    """
-    # runs on worker connect
-    worker_to_ws[worker_id] = websocket
-    logger.warning(f"Connected to worker {worker_id}!")
 
+worker_to_ws: Dict[str, Dict[str, Any]] = defaultdict(dict)  # Maps worker to socket ID to websocket
+ws_to_lock: Dict[str, Dict[str, TimedLock]] = defaultdict(dict)
```
epicfaace:
```suggestion
worker_to_lock: Dict[str, Dict[str, TimedLock]] = defaultdict(dict) # Maps worker ID to socket ID to lock
```

codalab/bin/ws_server.py
```diff
@@ -1,75 +1,148 @@
...
-async def worker_handler(websocket, worker_id):
-    """Handles routes of the form: /worker/{id}. This route is called when
-    a worker first connects to the ws-server, creating a connection that can
-    be used to ask the worker to check-in later.
-    """
-    # runs on worker connect
-    worker_to_ws[worker_id] = websocket
-    logger.warning(f"Connected to worker {worker_id}!")
 
+worker_to_ws: Dict[str, Dict[str, Any]] = defaultdict(dict)  # Maps worker to socket ID to websocket
```
epicfaace:
```suggestion
worker_to_ws: Dict[str, Dict[str, Any]] = defaultdict(dict)  # Maps worker ID to socket ID to websocket
```

codalab/model/bundle_model.py
epicfaace: maybe just replace with `find_oauth2_token(...).access_token == access_token`

codalab/model/worker_model.py
```diff
@@ -132,6 +136,80 @@ def _serialize_dependencies(dependencies):
     def _deserialize_dependencies(blob):
         return list(map(tuple, json.loads(blob)))
 
+    def update_workers(self, user_id, worker_id, update):
```
epicfaace: move it back where it came from

codalab/bin/ws_server.py
AndrewJGaut: Hmm... not sure that's the best name either. The worker is just connecting here to wait for messages, not sending anything. Maybe worker_connect?

codalab/lib/codalab_manager.py
AndrewJGaut: This is a CodalabArg. See [here](https://github.com/codalab/codalab-worksheets/blob/use-websockets-for-json-communication/codalab_service.py#L249-L252)

The purpose of that codalab manager property is (like the other properties that already do this) to abstract the env variable away from the WorkerModel (since the CodalabManager passes it in to its constructor). Let me know if this is bad stylistically, though; I certainly wasn't sure about it.

codalab/lib/download_manager.py
AndrewJGaut: I do want to keep the `with_sock` portion (I changed it to `with_unix_socket` in this newest version) since it helps make clear that it's using `AF_Unix` sockets. However, I could change `recv` back to `get` -- I thought `recv` was clearer since that's typically the API for getting data from sockets.

Let me know. I can definitely change this.

codalab/lib/download_manager.py
AndrewJGaut: done.

codalab/model/bundle_model.py
AndrewJGaut: Done.

codalab/worker/main.py
AndrewJGaut: Done.

codalab/bin/ws_server.py
AndrewJGaut: Done.

codalab/bin/ws_server.py
AndrewJGaut: Done.

codalab/model/worker_model.py
AndrewJGaut: Done.

codalab/model/worker_model.py
AndrewJGaut: Done.

codalab/model/worker_model.py
AndrewJGaut: Done.

codalab/bin/ws_server.py
AndrewJGaut: Done.

codalab/bin/ws_server.py
AndrewJGaut: Done.

codalab/bin/ws_server.py
AndrewJGaut: Hmm... weird. I see this listed in several other places (e.g. [here](https://github.com/Luka967/websocket-close-codes)) as `Try again later`, but it's not in [the RFC](https://www.rfc-editor.org/rfc/rfc6455#section-7.4).

Perhaps I should change this to a different status code, but none of the ones listed in the RFC are fitting... what do you think?

codalab/model/bundle_model.py
AndrewJGaut: OH, wait! I remember why I needed to do this. Basically, `find_oauth2_token` just fetches a single `oauth2` token, but the user may have several valid tokens active. For instance, here's my local `oauth2_token` database for the codalab user:
```
mysql> select * from oauth2_token where user_id=0
    -> ;
+-----+-----------------------+---------+---------+--------------------------------+--------------------------------+---------------------+
| id  | client_id             | user_id | scopes  | access_token                   | refresh_token                  | expires             |
+-----+-----------------------+---------+---------+--------------------------------+--------------------------------+---------------------+
| 123 | codalab_cli_client    | 0       | default | tDCP52XbDLP37nujTJOOjNzTS6Y68r | SVaPGWY9PpIA4JcLaHYfK0dvvJ48hD | 2023-08-16 06:37:25 |
| 124 | codalab_worker_client | 0       | default | F3tES3W9SEvpRfzW1d07CpzCzydYpe | 9fBYlZNpwmnH9TQTVmprH3YYae89v2 | 2023-08-16 06:37:41 |
| 125 | codalab_cli_client    | 0       | default | thgAVTJSNokta4g7EwZvWQNsmGDhZV | jmGC6hjBnB08EdujJ3qeH2Pt3Y5QOM | 2023-08-16 06:44:27 |
| 126 | codalab_cli_client    | 0       | default | ReKNdvycOdWjhae0PFubFRk8FAADxr | y5HiM5nIlxGpM4eSgZWJmOpX7RC1Lv | 2023-08-16 06:46:38 |
| 127 | codalab_worker_client | 0       | default | DNIvpmrPLmqK6o3ZEgYLYFF0HKkygO | VlOnzl1HfppROOhXQcDIKtW0hIy5rO | 2023-08-16 06:46:52 |
| 128 | codalab_cli_client    | 0       | default | UOYOfCkaU8y4dX8y3gczRaQmzbHTnQ | 0i9caPdfe5tJ2SU21tWIuZG9uFVfkO | 2023-08-16 06:47:11 |
| 129 | codalab_cli_client    | 0       | default | o36T83O7raUDwmDqywJijl2sLwwq5a | RL5MJ8W25UEgPiA23IWL8U0FWmc8Ob | 2023-08-16 06:49:26 |
| 130 | codalab_worker_client | 0       | default | qEiksslyxtsfuAxHJraqFR1gnjXhYT | MZug0gzJHv5uMwP1MhaSOlPRZdXjXg | 2023-08-16 06:49:59 |
| 131 | codalab_cli_client    | 0       | default | 5osMISohDoyVhDIOkjMO4VStyODgBU | Eb11XOj0pGhV6PzhKSZOKJSp2xFqfM | 2023-08-16 06:51:35 |
| 132 | codalab_worker_client | 0       | default | 9cttmuOkfXXUwawcmqXKu1g3Q3jOWJ | OO2Ey0ukIIN2Rm3zKUGLB5M4PMmbXy | 2023-08-16 06:51:59 |
| 133 | codalab_cli_client    | 0       | default | lDbuVCNK3Oee9Zlvv2n32aXjm5yI5V | dMJR0evB2fXUmkZgSotDLZxuKTBvqV | 2023-08-16 06:54:22 |
| 134 | codalab_worker_client | 0       | default | 8dq76r4bph03KKT4S06Gx9sMxAJyfL | OZChA69SQ470M84LH18w8fNg0sIkPi | 2023-08-16 06:54:46 |
| 135 | codalab_cli_client    | 0       | default | 43RJxmkngMqehMv4JyqMZUqif7ZSxi | xnerdfwqlyYEkuXiAm89FTPvoRjzjL | 2023-08-16 06:59:25 |
| 136 | codalab_worker_client | 0       | default | PDmOLXvJif5Rj6jdky3oC46CpMOnCi | BMQzvyzVbloRYauj2eVC7Xo7mC4YYE | 2023-08-16 06:59:38 |
+-----+-----------------------+---------+---------+--------------------------------+--------------------------------+---------------------+
14 rows in set (0.00 sec)
```
The current UTC time is `2023-08-16 6:04`, I believe, so these are all valid.

When I use `find_oauth2_token`, it returns the Oauth2Token with access token `F3tES3W9SEvpRfzW1d07CpzCzydYpe`, but the worker sends the access token `PDmOLXvJif5Rj6jdky3oC46CpMOnCi`. Both are valid, the auth is rejected when I use `find_oauth2_token`.

As such, we have two options: keep the function I made or modify `find_oauth2_token` to return all the valid tokens and then search through that list for the token we want.

For now, I'll leave in my earlier change and do that.

AndrewJGaut: I'm also not sure why the user has so many tokens active -- perhaps I should dig deeper into the Oauth2 code.

codalab/worker/worker.py
```diff
@@ -836,7 +852,7 @@ def netcat_fn():
                         break
                     total_data.append(data)
                 s.close()
-                reply(None, {}, b''.join(total_data))
+                reply(None, {}, io.BytesIO(b''.join(total_data)))
```
epicfaace: Do we need this change?

codalab/worker/worker.py
```diff
@@ -136,6 +140,11 @@ def __init__(
 
         self.ws_server = ws_server
 
+        assert (
+            num_coroutines > 0 and type(num_coroutines) is int
```
epicfaace: Do we need this?

epicfaace (approved):
Looks good to me. This new architecture is all documented in your doc, correct? Can you link to it? (maybe we also add a link to it as a comment in the code in `ws_server.py` so we never lose it)

epicfaace:
@AndrewJGaut is this ready to merge?

AndrewJGaut:
> Looks good to me. This new architecture is all documented in your doc, correct? Can you link to it? (maybe we also add a link to it as a comment in the code in `ws_server.py` so we never lose it)

I figured it was enough to link the doc in the PR description, since it's stated pretty clearly. Do you think we should in the code as well? I can do that.

AndrewJGaut:
> @AndrewJGaut is this ready to merge?

Let me test it out in dev one more time to be sure.

AndrewJGaut:
> > @AndrewJGaut is this ready to merge?
> 
> Let me test it out in dev one more time to be sure.

Confirmed it's good to merge.

PR #4501, opened 2023-07-03
Fix bugs in migration script

wwwjn:

After testing on dev environment, we find more bugs in migration script. 

### Checklist


epicfaace (approved):

epicfaace (approved):

wwwjn:
There is still another thing that needs to be fixed:

If a user first uploads a bundle to Azure Blob storage and changes the database pointing to Azure Blob storage, then the user re-runs the script to delete the original bundle storage, the script can not find the original bundle location. Since we migrate step by step, we need to store state information somewhere.

Possible solution: 
When changing the database, write (original_bundle_location, target_bundle_location) to the log file.  If the user wants to delete the original bundle, read through the log file and remove the original bundle files.

Working on this.

codalab/migration.py
```diff
@@ -195,56 +252,103 @@ def delete_original_bundle(self, bundle_uuid, bundle_location):
...
+            raise Exception("Input worksheet uuid has wrong format. ")
+        bundle_uuids = migration.get_bundle_uuids(worksheet_uuid)
 
     for bundle_uuid in bundle_uuids:
-        logging.info(bundle_uuid)
-
         bundle = migration.get_bundle(bundle_uuid)
-        if bundle.bundle_type != 'dataset' or bundle.state != 'ready':
+
+        # TODO: change this to allow migration of run bundles
```
AndrewJGaut: It looks like this change (removing bundle_type != 'dataset') allows for migration of run bundles, now, right?

Also, I just want to be sure: not migrating non-ready bundles is OK as long as, before we start the migration, the default upload bundle store is Azure blob (as referenced [here](https://github.com/codalab/codalab-worksheets/pull/4525#issuecomment-1699860312)), right? Although we may want to make sure uploading to disk isn't even possible as well...

wwwjn: > It looks like this change (removing bundle_type != 'dataset') allows for migration of run bundles, now, right?
> 

U r right.

> Also, I just want to be sure: not migrating non-ready bundles is OK as long as, before we start the migration, the default upload bundle store is Azure blob (as referenced [here](https://github.com/codalab/codalab-worksheets/pull/4525#issuecomment-1699860312)), right? Although we may want to make sure uploading to disk isn't even possible as well...

Even we set the Azure as default storage,  user could still set `--store [some-store-on-disk]` when `cl upload`, and upload to disk.

PR #4506, opened 2023-07-19
Fix metadata error - duplicate values by adding transactions

### LINKED ISSUE #4488: Metadata error - duplicate values
Not sure what triggers this error.

image

epicfaace:
Fix https://github.com/codalab/codalab-worksheets/issues/4488 by:
- adding transactions when bundle metadata is updated
- properly deal with existing duplicate values (pick the latest one) -- we no longer throw an error, we just show a warning now.

### DIFF
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1300,8 +1300,8 @@ class BundleModel(object):
                     logger.error(f"metadata_update_clause: {metadata_update_clause}")
                     logger.error(f"metadata_update_values: {metadata_update_values}")
                     logger.error(f"metadata_update: {metadata_update}")
-                    connection.execute(cl_bundle_metadata.insert(), metadata_update_values)
                     connection.execute(cl_bundle_metadata.delete().where(metadata_update_clause))
+                    self.do_multirow_insert(connection, cl_bundle_metadata, metadata_update_values)
                 if metadata_delete_keys:
                     connection.execute(cl_bundle_metadata.delete().where(metadata_delete_clause))
             except UnicodeError:
--- a/codalab/model/bundle_model.py
+++ b/codalab/model/bundle_model.py
@@ -1294,8 +1294,11 @@ class BundleModel(object):
                 if update:
                     connection.execute(cl_bundle.update().where(clause).values(update))
                 if metadata_update:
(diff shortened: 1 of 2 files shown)


AndrewJGaut:
Looks like there's a failure for the `competition` test:

```
codalab.client.json_api_client.JsonApiException: Unable to update bundles: Internal Server Error - Unexpected Internal Error ((MySQLdb._exceptions.IntegrityError) (1452, 'Cannot add or update a child row: a foreign key constraint fails (`codalab_bundles`.`bundle_metadata`, CONSTRAINT `bundle_metadata_ibfk_1` FOREIGN KEY (`bundle_uuid`) REFERENCES `bundle` (`uuid`))')
[SQL: INSERT INTO bundle_metadata () VALUES ()]
(Background on this error at: http://sqlalche.me/e/13/gkpj)). The administrators have been notified.
```

I'll help debug...

AndrewJGaut:
Looks like it's working now and I'm hopeful this will solve the kill/run issue as well! (I think a database transaction race condition is causing the issues we've been seeing based on what I saw).

codalab/model/bundle_model.py
```diff
@@ -157,8 +157,7 @@ def do_multirow_insert(connection, table, values):
         #   - Some dialects do not support multiple inserts in a single statement,
         #     which we deal with by using the DBAPI execute_many pattern.
         if values:
-            with connection.begin():
```
AndrewJGaut, line 160: This, to me, seems like a nested transaction (where `do_multi_row_insert` was called, the `connection` had always already had `begin()` called on it already). As such, I would actually think this would throw an error (see source code [here](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_19/lib/sqlalchemy/engine/base.py#L783-L857)), so I'm not sure why that wasn't happening.

Regardless, it would make sense this would introduce a race condition because this transaction becomes separate from the outer transaction that `do_multi_row_insert` is called in, and so other transactions can happen between this one and the 'outer' transaction.

AndrewJGaut (approved):

codalab/model/bundle_model.py
epicfaace: Great catch!

PR #4519, opened 2023-08-23
Fix kubernetes cert connections

epicfaace:
Fix kubernetes cert connections -- we are now passing certs in through an environment variable. This removes the issues that were happening with the newlines being parsed.


codalab/worker/main.py
```diff
@@ -222,7 +222,7 @@ def parse_args():
     parser.add_argument(
         '--kubernetes-cert',
         type=str,
-        help='Contents of the SSL cert for the Kubernetes cluster. Only applicable if --bundle-runtime is set to kubernetes and --kubernetes-cert-path is set to /dev/null.',
+        help='Base64 encoded contents of the SSL cert for the Kubernetes cluster. Only applicable if --bundle-runtime is set to kubernetes and --kubernetes-cert-path is set to /dev/null.',
```
percyliang: Is there not a way to use a path or environment variables?

epicfaace: This is set using an environment variable -- `CODALAB_WORKER_MANAGER_GPU_KUBERNETES_CERT` -- when starting up codalab_service.py.

We could use a path, but in that case we'd need to copy the cert file over to the worker pods on kubernetes when running them, which I think is a more complex solution.

percyliang: If it's already an environment variable, why do we need to pass it in as an argument?

codalab/worker/main.py
```diff
@@ -221,7 +222,7 @@ def parse_args():
...
```
epicfaace: @epicfaace Try -- quoting here (https://github.com/codalab/codalab-worksheets/blob/master/docker_config/compose_files/docker-compose.yml#L306C59-L306C59) to pass them as env variables

codalab/worker/main.py
epicfaace: Will try this

codalab/worker/main.py
```diff
@@ -221,7 +222,7 @@ def parse_args():
     parser.add_argument(
```
epicfaace: Delete --cert

docker_config/compose_files/docker-compose.yml
```diff
@@ -337,7 +337,7 @@ services:
       --cluster-host ${CODALAB_WORKER_MANAGER_GPU_KUBERNETES_CLUSTER_HOST}
       --auth-token ${CODALAB_WORKER_MANAGER_GPU_KUBERNETES_AUTH_TOKEN}
       --cert-path ${CODALAB_WORKER_MANAGER_GPU_KUBERNETES_CERT_PATH}
-      --cert ${CODALAB_WORKER_MANAGER_GPU_KUBERNETES_CERT}
+      --cert "${CODALAB_WORKER_MANAGER_GPU_KUBERNETES_CERT}"
```
epicfaace: Read this directly as an env variable CODALAB_WORKER_MANAGER_GPU_KUBERNETES_CERT

percyliang (approved):
This looks nice and clean.  So we ended up being able to pass the cert as a file (rather than doing some weird base64 or quoting thing)?  I thought we had trouble with this approach last time?

epicfaace:
> This looks nice and clean. So we ended up being able to pass the cert as a file (rather than doing some weird base64 or quoting thing)? I thought we had trouble with this approach last time?

We passed it as a file, but through an environment variable rather than a command-line argument (we were trying a command-line argument earlier, which didn't work).

PR #4525, opened 2023-08-29
Bump to version 1.7.1

yifanmai:
# Release notes

## Backend

- Fixed issues with GPU discovery (#4522)


wwwjn:
And we also want to change the deployment parameter, setting `always_use_azure_blob_beta` ENV variables to make Codalab always using  Azure as default storage. @yifanmai

epicfaace (approved):

PR #4530, opened 2023-09-14
PR

dma1dma1:
Implementing a new Codalab CLI command, `cl ancestors` that will recursively print out all ancestors/dependencies of some given bundle spec. 

### Related issues
* My local CLI was segfaulting incessantly on every CLI call after the functions had ran for some reason.
* Needed to update version to 1.7.1 to work with remote
* Needed to downgrade PyYAML to 5.3.1 for it to pip install correctly (Probably a me issue with python)
* I've added a test, but did not test it locally since I didn't have the local Docker containers running, so its just a rough outline of a test that could potentially not work.

### Screenshots
Screenshot 2023-09-13 at 11 01 56 PM

### Checklist

### DIFF
--- a/codalab/common.py
+++ b/codalab/common.py
@@ -29,7 +29,7 @@ from codalab.lib.beam.filesystems import (
 
 # Increment this on master when ready to cut a release.
 # http://semver.org/
-CODALAB_VERSION = '1.7.0'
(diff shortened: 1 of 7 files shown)


epicfaace:
Thanks!

PR #4536, opened 2023-09-24
Removes stale annotator exception.

### LINKED ISSUE #4533: Remove type:ignore mypy annotations in tar_subdir_stream
We had some annotations to silence the type checker on some lines due to a missing type. However, we should now be able to remove those. (See [here](https://github.com/codalab/codalab-worksheets/blob/v1.7.1/codalab/worker/tar_subdir_stream.py#L121-L122)).

Let's remove them if possible.

HidyHan:
### Related issues

Fixes https://github.com/codalab/codalab-worksheets/issues/4533.

### Checklist

* [NA] I've added a screenshot of the changes, if this is a frontend change
* [NA] I've added and/or updated tests, if this is a backend change
* [Y] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [NA] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

### DIFF
--- a/codalab/worker/tar_subdir_stream.py
+++ b/codalab/worker/tar_subdir_stream.py
@@ -118,9 +118,7 @@ class TarSubdirStream(BytesIO):
             assert self.output.fileobj is not None
             self.output.fileobj.write(chunk)
             self.current_desc.pos += len(chunk)
-            # We're ignoring types here because the TarFile.offset type is missing.
-            # TODO: Remove "# type: ignore" annotations once this PR is merged: https://github.com/python/typeshed/pull/5210
-            self.output.offset += len(chunk)  # type: ignore
+            self.output.offset += len(chunk)
         else:
             # We've finished reading the entire current descendant.
             # Write the remainder of the block, if needed, and then reset the descendant so it is empty.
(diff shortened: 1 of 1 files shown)


HidyHan:
@AndrewJGaut I don't think I can directly add you as a reviewer, so tagging you here.

AndrewJGaut:
@HidyHan Thanks for tagging me. I went ahead and kicked off CI, but it looks like the formatting test failed. Looking closer at it, it appears the MyPy typing check failed. So perhaps the ignore comments are still needed...

Either way, you can verify by running pre-commit again locally. (It may have failed when you ran it locally, but it's a bit hard to tell what failure looks like until you've run it a few times. Otherwise, maybe it's passing locally but failing when run on GHA, which would be surprising, but I've certainly seen weirder!)

HidyHan:
Taking a closer look.
1. I cannot reproduce the errors locally. The mypy version is specified in requirements.dev.txt, so the version difference may be some random place upstream. FWIW my python version is 3.9 and the python version in testing environment seems to be 3.7.
2. Our mypy requirement version was never upgraded after https://github.com/python/typeshed/pull/5210 is merged (so I am quite bewildered why I cannot reproduce the format error locally, given that my mypy version is 0.782 as required). Based on the release schedule in [1] we should upgrade to mypy 0.900+ to include the change. However, doing so triggers other errors, like this:

```
venv/bin/activate_this.py: error: Duplicate module named "activate_this" (also at "./codalab/lib/venv/bin/activate_this.py")
venv/bin/activate_this.py: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
Found 1 error in 1 file (errors prevented further checking)
```

@epicfaace any thoughts for quick fixes? If there are other benefits we can consider upgrading mypy version, but not sure whether this PR warrants it.

[1] https://mypy-lang.org/news.html

epicfaace:
Hmm @HidyHan the point of this issue was actually to upgrade mypy (https://github.com/codalab/codalab-worksheets/pull/4536). See my message here (https://github.com/python/typeshed/pull/5210). Basically I had contributed a missing type to mypy, and wanted to wait for a new mypy version so we can upgrade to it (so it can incorporate the new type).

PR #4556, opened 2023-10-18
Revert "Fix bugs in migration script"

epicfaace:
Reverts codalab/codalab-worksheets#4501


HidyHan:
Weirdly just reverting the migration script fixes the failed test. @AndrewJGaut Is it ok to revert, or should we debug the real cause further?

AndrewJGaut (approved):

PR #4578, opened 2023-11-15
Provides an option not to ignore patterns like .git, hidden files and…

### LINKED ISSUE #4575: Migration: don't have exclude_patterns
We want to be able to disable [these](https://github.com/codalab/codalab-worksheets/blob/master/codalab/worker/file_util.py#L29) so that we can migrate the _entire_ bundle when we need to.

HidyHan:
… MACOSX files.

Needed for the blob storage migration, because once upon a time these files were uploaded as part of bundles. Omitting them will cause discrepancy in bundles after the migration.

### Related issues

fixes #4575

### Checklist

* [NA] I've added a screenshot of the changes, if this is a frontend change
* [Y] I've added and/or updated tests, if this is a backend change
* [Y] I've run the [pre-commit.sh](https://github.com/codalab/codalab-worksheets/blob/master/pre-commit.sh) script
* [NA] I've updated [docs](https://github.com/codalab/codalab-worksheets/tree/master/docs), if needed

### DIFF
--- a/codalab/migration.py
+++ b/codalab/migration.py
@@ -118,8 +118,7 @@ class Migration:
         )
 
         if is_dir:
-            source_fileobj = zip_util.tar_gzip_directory(
-                bundle_location, exclude_patterns=None)
+            source_fileobj = zip_util.tar_gzip_directory(bundle_location, exclude_patterns=None)
             source_ext = ".tar.gz"
             unpack = True
         else:
--- a/codalab/migration.py
+++ b/codalab/migration.py
@@ -118,7 +118,8 @@ class Migration:
         )
 
         if is_dir:
-            source_fileobj = zip_util.tar_gzip_directory(bundle_location)
+            source_fileobj = zip_util.tar_gzip_directory(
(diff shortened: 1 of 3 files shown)


AndrewJGaut (commented):
This looks GREAT! Only thing is can we do the same thing for `zip_directory` [here](https://github.com/codalab/codalab-worksheets/blob/ignore-patterns/codalab/worker/file_util.py#L101) if possible? After that change, I think this is definitely good to go. Love the code styling for the test, too -- it's super clean!

AndrewJGaut (changes requested):
This looks GREAT! Only thing is can we do the same thing for `zip_directory` [here](https://github.com/codalab/codalab-worksheets/blob/ignore-patterns/codalab/worker/file_util.py#L101) if possible? After that change, I think this is definitely good to go. Love the code styling for the test, too -- it's super clean!

HidyHan:
Done. zip_directory appears to be dead code, but updated it to avoid confusion.

AndrewJGaut (approved):
Looks great!

PR #4583, opened 2023-12-06
Metadata only migration

### LINKED ISSUE #4581: Migration: Path '' in bundle <UUID> not found
Issue for the migration: "Error 2: Path '' in bundle <UUID> not found"

For these bundles, just don't migrate it. But change the database to say "blob"

(if you run "rm -d" for a bundle on blob, that same state should be reflected for these bundles.)

dma1dma1:

Bundles that have rm -d run on them have metadata but nothing on disk, breaking the migration script. This change avoids changing disk for those bundles and just runs a metadata change.

### Related issues

Fixes #4581 

### Checklist

### DIFF
--- a/codalab/migration.py
+++ b/codalab/migration.py
@@ -428,56 +428,59 @@ class Migration:
                 bundle_location = self.get_bundle_location(bundle_uuid)
 
                 # This is for handling cases where rm -d was run on the bundle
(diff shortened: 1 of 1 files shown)


percyliang:
What's the status? @dma1dma1

PR #4593, opened 2024-04-23
Updates the store field in codalab worksheet UI to display real bundle location.

HidyHan:

Previously the UI uses metadata.store, which is a parameter user optionally specifies when a bundle is created. However, this is not necessarily where the bundle is ultimately stored and can be empty, which is confusing.
Screenshot from 2024-04-23 11-56-38

### Checklist


wwwjn (approved):
LGTM! Thanks for identifying the problem here

wwwjn:
Pls address the failing CI checks first

PR #4595, opened 2024-05-23
Multireaderfilestream Redesign

dma1dma1:

The previous Multireaderfilestream design did not allow for backwards seeks, which causes uploads of directories above a certain size to break (~15MB). This was due to the process of reading directory bundles requiring seeks backwards up to a limit of 20MiB. The new design doesn't use BytesBuffers, but rather stores raw bytes, from 32MiB before the slowest reader, up to a maximum of 64MiB after the slowest reader. If a faster reader tries to read past this threshold, it will sleep until it is allowed to read more. 

#### Added Tests

**UPLOAD TESTS**

test_fileobj_tar_gz: basic directory bundle test, did not exist previously. Checks that a basic directory has the correct files and file sizes.

test_large_fileobj_tar_gz: Tests a large directory bundle past the size threshold limit of the previous Multireaderfilestream design. Checks that all files are uploaded.

test_large_fileobj_tar_gz2:  Tests a large directory bundle past the size threshold limit of the previous Multireaderfilestream design, with multiple large files. Checks that all files are uploaded and that sizes of files are correct.

test_upload_memory: Tests that the memory usage of the Multireaderfilestream uploading is expected at < 100MB

Manual speed test: Achieves faster than previous speeds with larger 1MB chunk sizes on upload, modified from 16KB

**MRFS TESTS**

test_reader_distance: Tests that 2 readers are always within the expected thresholds.

test_seek: Tests that backwards seeks within lookback range are working as expected.

test_toofar_seek: Tests that backwards seeks that go too far raise errors.

### Checklist

### DIFF
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -53,7 +53,8 @@ class MultiReaderFileStream(BytesIO):
         
         with self._lock:
             # Calculate how many new bytes need to be read
-            new_bytes_needed = num_bytes - (max(self._pos) - self._pos[index])
+            new_pos = self._pos[index] + num_bytes
+            new_bytes_needed = new_pos - max(self._pos)
             if new_bytes_needed > 0:
                 self._fill_buf_bytes(new_bytes_needed)
 
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -7,7 +7,7 @@ class MultiReaderFileStream(BytesIO):
     """
     NUM_READERS = 2
     LOOKBACK_LENGTH = 33554432
-    MAX_THRESHOLD = LOOKBACK_LENGTH * 4
+    MAX_THRESHOLD = LOOKBACK_LENGTH * 2
 
     def __init__(self, fileobj):
         self._buffer = bytes()
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -1,22 +1,21 @@
+import time
+
 from io import BytesIO, SEEK_SET, SEEK_END
 from threading import Lock
-import time
 class MultiReaderFileStream(BytesIO):
     """
-    FileStream that supports N readers with the following features and constraints:
+    FileStream that takes an input stream fileobj, and supports N readers with the following features and constraints:
         - Each reader's postion is tracked
         - A buffer of bytes() is stored which stores bytes from the position of the slowest reader
-          minus a lookback length of 32MiB to the fastest reader
-        - The fastest reader can be at most 64MiB ahead of the slowest reader, reads made
+          minus a LOOKBACK_LENGTH (default 32 MiB) to the fastest reader
+        - The fastest reader can be at most MAX_THRESHOLD (default 64 MiB) ahead of the slowest reader, reads made
           further than 64MiB will sleep until the slowest reader catches up
     """
     NUM_READERS = 2
-    LOOKBACK_LENGTH = 33554432 # 32 MiB
-    MAX_THRESHOLD = LOOKBACK_LENGTH * 2
 
-    def __init__(self, fileobj):
+    def __init__(self, fileobj, lookback_length=32*1024*1024):
         self._buffer = bytes() # Buffer of bytes read from the file object within the limits defined
-        self._buffer_pos = 0 # start position of buffer in the fileobj (min reader position - LOOKBACK LENGTH)
+        self._buffer_start_pos = 0 # start position of buffer in the fileobj (min reader position - LOOKBACK LENGTH)
         self._pos = [0 for _ in range(self.NUM_READERS)] # position of each reader in the fileobj
         self._fileobj = fileobj # The original file object the readers are reading from
         self._lock = Lock()  # lock to ensure one does not concurrently read self._fileobj / write to the buffer.
@@ -34,6 +33,8 @@ class MultiReaderFileStream(BytesIO):
                 return self.seek(s._index, offset, whence)
 
         self.readers = [FileStreamReader(i) for i in range(0, self.NUM_READERS)]
+        self.LOOKBACK_LENGTH = lookback_length
+        self.MAX_THRESHOLD = self.LOOKBACK_LENGTH * 2
 
     def _fill_buf_bytes(self, num_bytes=0):
         """
@@ -45,14 +46,10 @@ class MultiReaderFileStream(BytesIO):
         self._buffer += s
 
 
-    def read(self, index: int, num_bytes=None):  # type: ignore
+    def read(self, index: int, num_bytes: int):  # type: ignore
         """Read the specified number of bytes from the associated file.
         index: index that specifies which reader is reading.
         """
-        if num_bytes == None:
-            # Read remaining in buffer
-            num_bytes = (self._buffer_pos + len(self._buffer)) - self._pos[index]
-
         s = self.peek(index, num_bytes)
         with self._lock:
             # Modify reader position in fileobj
@@ -60,7 +57,7 @@ class MultiReaderFileStream(BytesIO):
 
             # If this reader is the minimum reader, we can remove some bytes from the beginning of the buffer
             # Calculated min position of buffer minus current min position of buffer
-            diff = (min(self._pos) - self.LOOKBACK_LENGTH) - self._buffer_pos 
+            diff = (min(self._pos) - self.LOOKBACK_LENGTH) - self._buffer_start_pos 
             # NOTE: it's possible for diff < 0 if seek backwards occur
             if diff > 0:
                 self._buffer = self._buffer[diff:]
@@ -69,7 +66,7 @@ class MultiReaderFileStream(BytesIO):
 
     def peek(self, index: int, num_bytes: int):   # type: ignore
         new_pos = self._pos[index] + num_bytes
-        while (new_pos) - self._buffer_pos > self.MAX_THRESHOLD:
+        while new_pos - self._buffer_start_pos > self.MAX_THRESHOLD:
             time.sleep(.1) # 100 ms
         
         with self._lock:
@@ -88,7 +85,7 @@ class MultiReaderFileStream(BytesIO):
         if whence == SEEK_END:
             super().seek(offset, whence)
         else:
-            assert offset >= self._buffer_pos
+            assert offset >= self._buffer_start_pos
             self._pos[index] = offset
             
     def close(self):
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -134,7 +134,10 @@ class MultiReaderFileStream(BytesIO):
         return s
 
     def seek(self, index: int, offset: int, whence=SEEK_SET):
-        pass
-
+        if whence == SEEK_END:
+            super().seek(offset, whence)
+        else:
+            pass
+            
     def close(self):
         self.__input.close()
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -61,7 +61,7 @@ class MultiReaderFileStream(BytesIO):
             # NOTE: it's possible for diff < 0 if seek backwards occur
             if diff > 0:
                 self._buffer = self._buffer[diff:]
-                self._buffer_pos += diff
+                self._buffer_start_pos += diff
         return s
 
     def peek(self, index: int, num_bytes: int):   # type: ignore
@@ -76,7 +76,7 @@ class MultiReaderFileStream(BytesIO):
                 self._fill_buf_bytes(new_bytes_needed)
 
             # Get the bytes in the buffer that correspond to the read function call
-            buffer_index = self._pos[index] - self._buffer_pos
+            buffer_index = self._pos[index] - self._buffer_start_pos
             s = self._buffer[buffer_index:buffer_index + num_bytes]
 
         return s
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -35,7 +35,7 @@ class MultiReaderFileStream(BytesIO):
 
         self.readers = [FileStreamReader(i) for i in range(0, self.NUM_READERS)]
 
-    def _fill_buf_bytes(self, num_bytes=None):
+    def _fill_buf_bytes(self, num_bytes=0):
         """
         Fills the buffer with bytes from the fileobj
         """
@@ -44,24 +44,17 @@ class MultiReaderFileStream(BytesIO):
             return
         self._buffer += s
 
-    def read(self, index: int, num_bytes=0):  # type: ignore
+
+    def read(self, index: int, num_bytes=None):  # type: ignore
         """Read the specified number of bytes from the associated file.
         index: index that specifies which reader is reading.
         """
-        while (self._pos[index] + num_bytes) - self._buffer_pos > self.MAX_THRESHOLD:
-            time.sleep(.1) # 100 ms
-        
+        if num_bytes == None:
+            # Read remaining in buffer
+            num_bytes = (self._buffer_pos + len(self._buffer)) - self._pos[index]
+
+        s = self.peek(index, num_bytes)
         with self._lock:
-            # Calculate how many new bytes need to be read
-            new_pos = self._pos[index] + num_bytes
-            new_bytes_needed = new_pos - max(self._pos)
-            if new_bytes_needed > 0:
-                self._fill_buf_bytes(new_bytes_needed)
-
-            # Get the bytes in the buffer that correspond to the read function call
-            buffer_index = self._pos[index] - self._buffer_pos
-            s = self._buffer[buffer_index:buffer_index + num_bytes]
-
             # Modify reader position in fileobj
             self._pos[index] += len(s)
 
@@ -74,20 +67,21 @@ class MultiReaderFileStream(BytesIO):
                 self._buffer_pos += diff
         return s
 
-    def peek(self, index: int, num_bytes):   # type: ignore
-        while (self._pos[index] + num_bytes) - self._buffer_pos > self.MAX_THRESHOLD:
+    def peek(self, index: int, num_bytes: int):   # type: ignore
+        new_pos = self._pos[index] + num_bytes
+        while (new_pos) - self._buffer_pos > self.MAX_THRESHOLD:
             time.sleep(.1) # 100 ms
         
         with self._lock:
             # Calculate how many new bytes need to be read
-            new_pos = self._pos[index] + num_bytes
             new_bytes_needed = new_pos - max(self._pos)
             if new_bytes_needed > 0:
                 self._fill_buf_bytes(new_bytes_needed)
-        
+
             # Get the bytes in the buffer that correspond to the read function call
             buffer_index = self._pos[index] - self._buffer_pos
             s = self._buffer[buffer_index:buffer_index + num_bytes]
+
         return s
 
     def seek(self, index: int, offset: int, whence=SEEK_SET):
--- a/codalab/lib/beam/MultiReaderFileStream.py
+++ b/codalab/lib/beam/MultiReaderFileStream.py
@@ -1,4 +1,4 @@
-from io import BytesIO
+from io import BytesIO, SEEK_SET, SEEK_END
 from threading import Lock
 import time
 
@@ -53,13 +53,13 @@ class MinMaxHeap:
         if self.heap:
             return self.item_index[self.max()]
         
-    def get_at_index(self, index):
+    def get_at_index(self, index: int):
         if index < len(self.heap):
             return self.heap[index]
         else:
             raise IndexError("Index out of range")
 
-class MultiReaderFileStream2(BytesIO):
+class MultiReaderFileStream(BytesIO):
     """
     FileStream that support multiple readers and seeks backwards
     """
@@ -69,7 +69,7 @@ class MultiReaderFileStream2(BytesIO):
 
     def __init__(self, fileobj):
         self._buffer = bytes()
-        self._buffer_pos = 0 # position in the fileobj (min reader position - LOOKBACK LENGTH)
+        self._buffer_pos = 0 # start position of buffer in the fileobj (min reader position - LOOKBACK LENGTH)
         self._size = 0 # size of bytes (for convenience)
         self._pos = MinMaxHeap() # position of each reader
         self._fileobj = fileobj
@@ -87,6 +87,9 @@ class MultiReaderFileStream2(BytesIO):
 
             def peek(s, num_bytes):
                 return self.peek(s._index, num_bytes)
+            
+            def seek(s, offset, whence):
+                return self.seek(s._index, offset, whence)
 
         self.readers = [FileStreamReader(i) for i in range(0, self.NUM_READERS)]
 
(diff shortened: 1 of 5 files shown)


codalab/lib/upload_manager.py
```diff
@@ -255,7 +255,7 @@ def write_fileobj(
             conn_str = os.environ.get('AZURE_STORAGE_CONNECTION_STRING', '')
             os.environ['AZURE_STORAGE_CONNECTION_STRING'] = bundle_conn_str
         try:
-            CHUNK_SIZE = 16 * 1024
+            CHUNK_SIZE = 1024 * 1024
```
epicfaace, line 259: Why increase?

dma1dma1: The upload speed with the smaller chunk size was too slow due to the sleep behavior that occurs on the faster reader, which is always the index reader.

dma1dma1: It seems like there's no super meaningful reason to keep chunk size smallish since the speed tradeoff is too large

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,22 @@
...
 from threading import Lock
-
-from codalab.worker.un_gzip_stream import BytesBuffer
-
-
+import time
 class MultiReaderFileStream(BytesIO):
     """
-    FileStream that support multiple readers
+    FileStream that support multiple readers and seeks backwards
```
percyliang: Could you flesh this out into a much longer description (like what you have in the PR description), talking about what is maintained and what operations are supported, and what assumptions are made.

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,22 @@
...
     NUM_READERS = 2
+    LOOKBACK_LENGTH = 33554432
+    MAX_THRESHOLD = LOOKBACK_LENGTH * 2
 
     def __init__(self, fileobj):
-        self._bufs = [BytesBuffer() for _ in range(0, self.NUM_READERS)]
-        self._pos = [0 for _ in range(0, self.NUM_READERS)]
+        self._buffer = bytes()
+        self._buffer_pos = 0 # start position of buffer in the fileobj (min reader position - LOOKBACK LENGTH)
+        self._size = 0 # size of bytes (for convenience)
```
percyliang: size of buffer?

codalab/lib/beam/MultiReaderFileStream.py
percyliang: Say things like: there are N readers, each reader has a current position...  buffer keeps bytes ... min reader position ... max reader position

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,22 @@
...
+    FileStream that support multiple readers and seeks backwards
     """
     NUM_READERS = 2
+    LOOKBACK_LENGTH = 33554432
+    MAX_THRESHOLD = LOOKBACK_LENGTH * 2
 
     def __init__(self, fileobj):
-        self._bufs = [BytesBuffer() for _ in range(0, self.NUM_READERS)]
-        self._pos = [0 for _ in range(0, self.NUM_READERS)]
+        self._buffer = bytes()
```
percyliang: Comment the variables like fileobj, _buffer, etc.

codalab/lib/beam/MultiReaderFileStream.py
percyliang: call it `_buffer_size` to be more unambiguous

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
-            while num_bytes is None or len(self._bufs[index]) < num_bytes:
-                s = self._fileobj.read(num_bytes)
-                if not s:
-                    break
-                for i in range(0, self.NUM_READERS):
-                    self._bufs[i].write(s)
-
-    def read(self, index: int, num_bytes=None):  # type: ignore
+    def _fill_buf_bytes(self, num_bytes=None):
+        # with self._lock:
```
percyliang: Add docstring of what this function does

codalab/lib/beam/MultiReaderFileStream.py
percyliang: Actually, can we get rid of this and just use `len(_buffer)`...avoid having extra state that you have to keep in sync

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
         index: index that specifies which reader is reading.
         """
-        self._fill_buf_bytes(index, num_bytes)
-        if num_bytes is None:
-            num_bytes = len(self._bufs[index])
-        s = self._bufs[index].read(num_bytes)
-        self._pos[index] += len(s)
+        # Calculate how many new bytes need to be read
+        with self._lock:
+            new_bytes_needed = num_bytes - (max(self._pos) - self._pos[index])
```
percyliang: I think it's clearer to define a new variable for `new_pos = self._pos[index] + num_bytes`, which can be re-used...

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
+            if new_bytes_needed > 0:
+                self._fill_buf_bytes(new_bytes_needed)
+        while (self._pos[index] + num_bytes) - self._buffer_pos > self.MAX_THRESHOLD:
+            time.sleep(10) # 100 ms
```
percyliang: time.sleep(10) does 10 seconds... you mean time.sleep(0.1)?

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
+
+        with self._lock:
+            old_position = self._pos[index] - self._buffer_pos
+            s = self._buffer[old_position:old_position + num_bytes]
+
+            # Modify position
+            self._pos[index] += len(s)
```
percyliang, line 56: I think one thing that is confusing about this code is that position refers to two things - absolute position in the file and position in the buffer.  I'd suggest using different names for the two...I'd try to use the file position for everything, and call that position, while using `index` for `_buffer`.

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
+
+            # Update buffer if this reader is the minimum reader
+            diff = (min(self._pos) - self.LOOKBACK_LENGTH) - self._buffer_pos # calculated min position of buffer minus current min position of buffer
```
percyliang: Put comment on own line

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
+            time.sleep(10) # 100 ms
+
+        with self._lock:
+            old_position = self._pos[index] - self._buffer_pos
+            s = self._buffer[old_position:old_position + num_bytes]
+
+            # Modify position
+            self._pos[index] += len(s)
+
+            # Update buffer if this reader is the minimum reader
```
percyliang: Be more specific - say that if we're the min reader, then we can remove some of the stuff in the buffer...

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +26,60 @@ def read(s, num_bytes=None):
...
+                self._buffer = self._buffer[diff:]
+                self._buffer_pos += diff
+                self._size -= diff
         return s
 
     def peek(self, index: int, num_bytes):   # type: ignore
-        self._fill_buf_bytes(index, num_bytes)
-        s = self._bufs[index].peek(num_bytes)
-        return s
+        pass
```
percyliang: Needs to be implemented...

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,142 @@
...
+                for _ in range(4):
+                    status = reader_2.read(CHUNKSIZE)
+
+            t1 = Thread(target=thread1)
+            t2 = Thread(target=thread2)
+
+            t1.start()
+
+            # Sleep a little for thread 1 to start reading
+            time.sleep(3)
```
percyliang: Can we reduce the times to < 1 sec to make the test run faster?

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+            # Assert that the first reader has not read past the Maximum threshold
```
percyliang: lowercase maximum

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+from threading import Thread
+
+from codalab.lib.beam.MultiReaderFileStream import MultiReaderFileStream
+
+FILESIZE = 100000000
+CHUNKSIZE = FILESIZE/10
+
+class MultiReaderFileStreamTest(unittest.TestCase):
+    def test_reader_distance(self):
```
percyliang, line 13: Describe at a high-level what this test does

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,142 @@
...
+            self.assertEqual(6445568, m_stream._buffer_pos)
+
+            # Assert that the buffer is length 100000000 - 6445568 
+            self.assertEqual(93554432, m_stream._size)
+
+            t1.join()
+            t2.join()
+    
+    def test_seek(self):
+        with tempfile.NamedTemporaryFile(delete=True) as f:
```
percyliang, line 73: Describe at a high-level what this test does

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+            self.assertIsNone(result)
+
+            # Check that reader 2 is at 50000000 and buffer position is correct
+            self.assertEqual(50000000, m_stream._pos[1])
+            self.assertEqual(16445568, m_stream._buffer_pos)
+
+
+    def test_toofar_seek(self):
+        with tempfile.NamedTemporaryFile(delete=True) as f:
```
percyliang, line 123: Describe at a high-level what this test does

codalab/lib/beam/MultiReaderFileStream.py
dma1dma1: Done!

codalab/lib/beam/MultiReaderFileStream.py
dma1dma1: Changed position for buffer usage to buffer_index

codalab/lib/beam/MultiReaderFileStream.py
dma1dma1: I'm a little bit unclear what this would be used for?

percyliang: You can then do `new_bytes_needed = new_pos - max(self._pos)`, etc.

dma1dma1: Ahh ok I see, that does make it clearer

dma1dma1: Made the changes

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,73 @@ def read(s, num_bytes=None):
...
-        with self._lock:
-            while num_bytes is None or len(self._bufs[index]) < num_bytes:
-                s = self._fileobj.read(num_bytes)
-                if not s:
-                    break
-                for i in range(0, self.NUM_READERS):
-                    self._bufs[i].write(s)
-
-    def read(self, index: int, num_bytes=None):  # type: ignore
+    def _fill_buf_bytes(self, num_bytes=None):
```
percyliang: Is None allowed?

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,73 @@ def read(s, num_bytes=None):
...
+        """
+        Fills the buffer with bytes from the fileobj
+        """
+        s = self._fileobj.read(num_bytes)
+        if not s:
+            return
+        self._buffer += s
+
+    def read(self, index: int, num_bytes=0):  # type: ignore
```
percyliang: Why 0?  If optional, use None, and add type hint `Optional[int]`

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,73 @@ def read(s, num_bytes=None):
...
         """Read the specified number of bytes from the associated file.
         index: index that specifies which reader is reading.
         """
-        self._fill_buf_bytes(index, num_bytes)
-        if num_bytes is None:
-            num_bytes = len(self._bufs[index])
-        s = self._bufs[index].read(num_bytes)
-        self._pos[index] += len(s)
+        while (self._pos[index] + num_bytes) - self._buffer_pos > self.MAX_THRESHOLD:
```
percyliang: Define `new_pos` up here to reuse

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,73 @@ def read(s, num_bytes=None):
...
+            # NOTE: it's possible for diff < 0 if seek backwards occur
+            if diff > 0:
+                self._buffer = self._buffer[diff:]
+                self._buffer_pos += diff
         return s
 
     def peek(self, index: int, num_bytes):   # type: ignore
-        self._fill_buf_bytes(index, num_bytes)
-        s = self._bufs[index].peek(num_bytes)
+        while (self._pos[index] + num_bytes) - self._buffer_pos > self.MAX_THRESHOLD:
```
percyliang: Can we avoid duplicate code with read?

codalab/lib/upload_manager.py
percyliang, line 259: Comment that we want 1MB for performance

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,155 @@
...
+                for _ in range(4):
+                    status = reader_2.read(CHUNKSIZE)
+
+            t1 = Thread(target=thread1)
+            t2 = Thread(target=thread2)
+
+            t1.start()
+
+            # Sleep a little for thread 1 to start reading
+            time.sleep(3)
```
percyliang: reduce time?

codalab/lib/beam/MultiReaderFileStream.py
dma1dma1: I made the change to its current iteration, logic is that read is just a peek and position change. My concern is that this requires the lock to be released, which could potentially mean another thread could interpose during this release and do another read or peek. From what I can tell this shouldn't affect any logic since the actual buffer reading occurs before the lock releases, but I could be wrong. Definitely might cause some context switches though, but with only 2 threads the effect should be minimal

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,25 @@
...
-    FileStream that support multiple readers
+    FileStream that supports N readers with the following features and constraints:
+        - Each reader's postion is tracked
+        - A buffer of bytes() is stored which stores bytes from the position of the slowest reader
+          minus a lookback length of 32MiB to the fastest reader
+        - The fastest reader can be at most 64MiB ahead of the slowest reader, reads made
+          further than 64MiB will sleep until the slowest reader catches up
     """
     NUM_READERS = 2
+    LOOKBACK_LENGTH = 33554432 # 32 MiB
```
percyliang: Write 32 * 1024 * 1024 for transparency

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,25 @@
...
     """
-    FileStream that support multiple readers
+    FileStream that supports N readers with the following features and constraints:
+        - Each reader's postion is tracked
+        - A buffer of bytes() is stored which stores bytes from the position of the slowest reader
+          minus a lookback length of 32MiB to the fastest reader
+        - The fastest reader can be at most 64MiB ahead of the slowest reader, reads made
+          further than 64MiB will sleep until the slowest reader catches up
     """
     NUM_READERS = 2
```
percyliang, line 14: Seems like this should be an argument that's passed into the constructor

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,25 @@
...
 from threading import Lock
-
-from codalab.worker.un_gzip_stream import BytesBuffer
-
-
+import time
 class MultiReaderFileStream(BytesIO):
     """
-    FileStream that support multiple readers
+    FileStream that supports N readers with the following features and constraints:
```
percyliang: Should say something even more general like: Given an input stream `fileobj`, copy this to multiple streams (readers) that have to be put in sync.

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,67 @@ def read(s, num_bytes=None):
...
     def read(self, index: int, num_bytes=None):  # type: ignore
         """Read the specified number of bytes from the associated file.
         index: index that specifies which reader is reading.
         """
-        self._fill_buf_bytes(index, num_bytes)
-        if num_bytes is None:
-            num_bytes = len(self._bufs[index])
-        s = self._bufs[index].read(num_bytes)
-        self._pos[index] += len(s)
+        if num_bytes == None:
```
percyliang: `num_bytes is None`

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,67 @@ def read(s, num_bytes=None):
...
+        """
+        Fills the buffer with bytes from the fileobj
+        """
+        s = self._fileobj.read(num_bytes)
+        if not s:
+            return
+        self._buffer += s
+
 
     def read(self, index: int, num_bytes=None):  # type: ignore
```
percyliang: `num_bytes: Optional[int] = None`

codalab/lib/beam/MultiReaderFileStream.py
percyliang: Do we ever call this with `None`?  If not, don't support it

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -25,33 +29,67 @@ def read(s, num_bytes=None):
...
+                self._buffer = self._buffer[diff:]
+                self._buffer_pos += diff
         return s
 
-    def peek(self, index: int, num_bytes):   # type: ignore
-        self._fill_buf_bytes(index, num_bytes)
-        s = self._bufs[index].peek(num_bytes)
+    def peek(self, index: int, num_bytes: int):   # type: ignore
+        new_pos = self._pos[index] + num_bytes
+        while (new_pos) - self._buffer_pos > self.MAX_THRESHOLD:
```
percyliang: remove `()`

codalab/lib/beam/MultiReaderFileStream.py
```diff
@@ -1,21 +1,25 @@
...
+    MAX_THRESHOLD = LOOKBACK_LENGTH * 2
 
     def __init__(self, fileobj):
-        self._bufs = [BytesBuffer() for _ in range(0, self.NUM_READERS)]
-        self._pos = [0 for _ in range(0, self.NUM_READERS)]
-        self._fileobj = fileobj
-        self._lock = Lock()  # lock to ensure one does not concurrently read self._fileobj / write to the buffers.
-
+        self._buffer = bytes() # Buffer of bytes read from the file object within the limits defined
+        self._buffer_pos = 0 # start position of buffer in the fileobj (min reader position - LOOKBACK LENGTH)
```
percyliang: to be more precise, call this `_buffer_start_pos`

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+            # Sleep a little for thread 2 to start reading
+            time.sleep(.5)
+
+            # Assert that the first reader is at 100000000, second reader is at 40000000
+            self.assertEqual(100000000, m_stream._pos[0])
+            self.assertEqual(40000000, m_stream._pos[1])
+
+            # Assert that the buffer is at 6445568 (40000000 - LOOKBACK_LENGTH)
+            self.assertEqual(6445568, m_stream._buffer_pos)
```
percyliang: Could we write the expression for these constants so it's more transparent?

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,155 @@
...
+from codalab.lib.beam.MultiReaderFileStream import MultiReaderFileStream
+
+FILESIZE = 100000000
+CHUNKSIZE = FILESIZE/10
+
+class MultiReaderFileStreamTest(unittest.TestCase):
+    def test_reader_distance(self):
+        """
+        This test verifies that both readers in the Multireaderfilestream
+        are within the limits defined in the class
```
percyliang: what does "within the limits" mean?

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+            # Assert that the first reader is at 100000000, second reader is at 40000000
+            self.assertEqual(100000000, m_stream._pos[0])
+            self.assertEqual(40000000, m_stream._pos[1])
+
+            # Assert that the buffer is at 6445568 (40000000 - LOOKBACK_LENGTH)
+            self.assertEqual(6445568, m_stream._buffer_pos)
+
+            # Assert that the buffer is length 100000000 - 6445568 
+            self.assertEqual(93554432, len(m_stream._buffer))
```
percyliang: same here

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+            t1.join()
+            t2.join()
+    
+    def test_backwards_seek(self):
+        """
+        This test verifies that a backwards seek within the lookback length
+        defined in the Multireaderfilestream class behaves as expected
```
percyliang: Fix case for Multi...

tests/unit/beam/multireaderfilestream_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+            # Check that reader 2 is at 50000000 and buffer position is correct
+            self.assertEqual(50000000, m_stream._pos[1])
+            self.assertEqual(16445568, m_stream._buffer_pos)
+
+
+    def test_toofar_seek(self):
+        """
+        This test verifies that a backwards seek past the lookback length
+        defined in the Multireaderfilestream class behaves as expected with
```
percyliang: Fix casing

percyliang (approved):
This is great overall.  Left a few comments so we can clean things up.

codalab/lib/beam/MultiReaderFileStream.py
dma1dma1: Don't think so, but this emulates default read() behavior for other io streams in python, where None means read to the end of the fileobj, or in our case the buffer.

percyliang: end of fileobj is different semantically from end of buffer (which is arbitrary), so I would avoid having `None` to set the wrong expectations of a behavior that we don't support

dma1dma1: Makes sense, I'll make the change.

